PR #139, opened 2019-06-25
Enable the updateRegistrarRdapBaseUrls endpoint

guyben13:
This endpoint will update the Registrar RDAP Base URLs from ICANN once a day.
    
Note that the endpoint needs a TLD to use for the query, since the ICANN MoSAPI endpoint we use needs a TLD to authenticate (even though the results don't depend on the TLD)
    
This is unfortunate in an open source project, and we might need to change it to a configured value.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/139)

### DIFF
--- a/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml
+++ b/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml
@@ -112,8 +112,6 @@
     <target>backend</target>
   </cron>
 
-  <!--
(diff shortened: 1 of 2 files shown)


gbrodman:
Wouldn't we just want to add the private config entry first, then use it in the public repo once that has merged in?

gbrodman:
I'm going to close this in favor of #141

PR #140, opened 2019-06-26
Stop provisioning RegistryEnvironment from Dagger

hstonec:
We found that some webdriver tests failed because RegistryEnvironment
was set to 'production' by other test and was carried over to the
webdriver test, and it was nontrivial to fix them because the instance
of RegistryEnvironment was injected into the testing web server.

Also, to eliminate this problem from potential victims, we stopped
provisioning RegistryEnvironment from Dagger. Instead, we just use
RegistryEnvironment.get() function to get the instance, which indeed
retrives the value from system property every time. If any test case
needs to run the test with other environment, it can just simply use
the existing SystemPropertyRule and RegistryEnvironment.${ENV}.setup
to change it.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/140)

### DIFF
--- a/core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java
+++ b/core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java
@@ -17,6 +17,7 @@ package google.registry.ui.server.registrar;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.net.HttpHeaders.LOCATION;
 import static com.google.common.net.HttpHeaders.X_FRAME_OPTIONS;
+import static google.registry.config.RegistryEnvironment.PRODUCTION;
 import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
 import static javax.servlet.http.HttpServletResponse.SC_MOVED_TEMPORARILY;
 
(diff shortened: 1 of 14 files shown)


core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java
```diff
@@ -70,53 +70,70 @@
...
-    response.setHeader("X-Ui-Compatible", "IE=edge");  // Ask IE not to be silly.
+    response.setHeader(X_FRAME_OPTIONS, "SAMEORIGIN"); // Disallow iframing.
+    response.setHeader("X-Ui-Compatible", "IE=edge"); // Ask IE not to be silly.
 
     logger.atInfo().log(
         "User %s is accessing the OT&E setup page. Method= %s",
         registrarAccessor.userIdForLogging(), method);
-    checkState(registryEnvironment != RegistryEnvironment.PRODUCTION, "Can't create OT&E in prod");
+    checkState(
+        RegistryEnvironment.get() != RegistryEnvironment.PRODUCTION, "Can't create OT&E in prod");
```
gbrodman: ditto hopefully using .equals here (and probably some other places as well)

PR #141, opened 2019-06-27
Attempt login to MosAPI via all available TLDs

gbrodman:
There's no reason why we should need a TLD as input here because it doesn't actually matter which one we use (they all have the same password).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/141)


gbrodman:
If we do this, we don't need to pass the TLD in when running the cron job like in #139

gbrodman:
Lai, I think we should be good for a review here. PTAL, thanks.

gbrodman:
Note: I will need to modify this to hard-code the password, deploy to alpha and test it, then remove the alpha-specific cron.xml file change before merging

gbrodman:
Testing is done and it worked, see https://pantheon.corp.google.com/logs/viewer?expandAll=false&project=domain-registry-alpha&minLogLevel=0&timestamp=2019-07-03T15:27:10.770000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=protoPayload.taskName%3D%2220641833e37c58d8071412cb23e44c54%22%0AprotoPayload.taskQueueName%3D%22__cron%22%0A&dateRangeStart=2019-07-03T14:27:12.193Z&dateRangeEnd=2019-07-03T15:27:12.193Z&interval=PT1H&scrollTimestamp=2019-07-03T15:26:29.552140000Z

PR #143, opened 2019-07-01
Move test resource files into src/test/resources

gbrodman:
Instead of living in the src/test/java/.../testdata/ folder, they should just live in src/test/resources/...

This only covers files that were in testdata/ folders but it's a start.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/143)


gbrodman:
Done

gbrodman:
Not sure why the Kokoro build hung but force-pushing after a rebase appears to have started it again.

PR #160, opened 2019-07-08
Don't extend expiration times for deleted domains

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/160)


gbrodman:
This may be a separate issue, but domains actually are never taken out of the PENDING_DELETE status when we delete them, so PENDING_DELETE is actually the status that all these domains are in.

For the grace period statuses, wouldn't we want to still skip extending the expiration if we're in a grace period? If we restore the domain, we reset the expiration time anyway to be the time of restoration + one year. 

What effect would a pending transfer have on the expiration time? That shouldn't have any effect on whether or not the domain has been deleted, right?

gbrodman:
I think I might have misread the code before -- while we do set the deletion time in the DomainDeleteFlow, we set it for the future like you say. To be safe, we can avoid situations where the deletion time is not the end of time-- this means that there's a pending deletion for some reason.

PR #163, opened 2019-07-10
Add a metric for EPP processing time regardless of ID/TLD

gbrodman:
Assuming this all goes well, we can remove the by-tld and by-id metrics since they're taking up a significantly larger amount of space / cost.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/163)


gbrodman:
Lai, do we the processing_time (that differentiates by registrar client ID) in the canary judge?

gbrodman:
Interesting idea. The logic could basically be something like:

if (tld.contains("canary")) return TrafficType.CANARY
else if (tld.contains("test")) return TrafficType.PROBER
return TrafficType.REAL

PR #166, opened 2019-07-11
Force update the cron jobs when deploying

gbrodman:
Also format the file to use two spaces instead of four as indentation

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/166)


gbrodman:
Good news: there exists such tasks. Bad news: they're not in the version of the appengine plugin we're on so we'll need to upgrade. #169

gbrodman:
Closing this since those Gradle tasks are going to be the better solution.

PR #169, opened 2019-07-11
Bump the version of the AppEngine Gradle plugin

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/169)

### DIFF
--- a/appengine_war.gradle
+++ b/appengine_war.gradle
@@ -83,6 +83,6 @@ dependencies {
     compile project(':core')
 }
 
-rootProject.deploy.dependsOn appengineDeploy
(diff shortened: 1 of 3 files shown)


jianglai:
Should we change [here](https://github.com/google/nomulus/blob/77590dcd8ea1490e85bcc0eb3c9696e6f9c5408d/appengine_war.gradle#L86) to have `rootProject.Deploy` depend on `appengineDeployAll` so that `gradle deploy` deploys the cron job as well?

gbrodman:
I was considering doing that in a separate PR but there's probably no reason not to do it here. The "deployAll" step takes more time I assume, but I don't believe it's significant.

PR #175, opened 2019-07-15
Spotless java fix

weiminyu:
Apply google-java-format to all files.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/175)


buildSrc/build.gradle
```diff
@@ -52,7 +52,7 @@ repositories {
 }
 
 apply from: '../dependencies.gradle'
-apply from: '../java_common.gradle'
+  apply from: '../java_common.gradle'
```
gbrodman: What's up with this formatting?

gbrodman:
Overall I think this is fine and I'm not bothered by the slight git blame corruption. If we want to enforce this at all, I think we need to have this change.

weiminyu:
Abandoned. Prefer diff-only formatting.

PR #180, opened 2019-07-18
Add full prober WebWHOIS sequence functionality

Sanger2000:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/180)


Sanger2000:
___
*[prober/src/main/java/google/registry/monitoring/blackbox/ProbingAction.java, line 84 at r4](https://reviewable.io/reviews/google/nomulus/180#-LlADiub3Uw8sQxLzA8g:-LlCS_Tc152dzMWSFY_k:b-313oj6) ([raw file](https://github.com/google/nomulus/blob/9c8af7eae6b4430a627dcf40dd99f15e6ffd3d21/prober/src/main/java/google/registry/monitoring/blackbox/ProbingAction.java#L84)):*

> ```
>   /** {@link ChannelFuture} object that is set to success after successful channel connection is established. */
>   public abstract ChannelFuture connectionFuture();
> ```
Is this really necessary? Shouldn't we just fail when the connection failed? My point is that after you call your bootstrap to make the connection, you immediately have access to the connection future and can do things with it (like register a failure). Why would the connection future be a property of this value class? Every getter should represent something that you explicitly set on the object. The connection future seems to a derived object that you don't explicitly give to the action, and does not need to be saved for other objects to access.
</blockquote></details>

Connection future is somewhat necessary because the connection is established when the ProbingAction is built for ephemeral connections. So after it is created, the ProbingStep calls the action.  As a result, within the call method, there needs to be a way to be informed when the connection is finished so it can know when the message should be passed down the channel pipeline.

One of the biggest points of refactoring ProbingAction is that a built ProbingAction has a channel associated with it and one of the tradeoffs is that we don't know for certain when that channel is successfully connected unless we use the connectionFuture.

Sanger2000:
___
*[prober/src/main/java/google/registry/monitoring/blackbox/tokens/WebWhoisToken.java, line 39 at r4](https://reviewable.io/reviews/google/nomulus/180#-Ll9sv421Au94VwcohV2:-LlCnkPEDWyW4X7Mo3Kt:bi6hi33) ([raw file](https://github.com/google/nomulus/blob/9c8af7eae6b4430a627dcf40dd99f15e6ffd3d21/prober/src/main/java/google/registry/monitoring/blackbox/tokens/WebWhoisToken.java#L39)):*

> ```
> private final ImmutableList<String> topLevelDomains;
> ```
Why don't we store them in a circular  `LinkedList<String>`? Seems cleaner than incrementing an index and doing modulus. 
</blockquote></details>

Is there a google implementation of a circular linked list or should I do my own custom class?

Sanger2000:
___
*[prober/src/main/java/google/registry/monitoring/blackbox/ProbingAction.java, line 84 at r4](https://reviewable.io/reviews/google/nomulus/180#-LlADiub3Uw8sQxLzA8g:-LlDTk4_9bM_8JXSTa1q:b-rbcgnu) ([raw file](https://github.com/google/nomulus/blob/9c8af7eae6b4430a627dcf40dd99f15e6ffd3d21/prober/src/main/java/google/registry/monitoring/blackbox/ProbingAction.java#L84)):*

> One of the biggest points of refactoring ProbingAction is that a built ProbingAction has a channel associated with it and one of the tradeoffs is that we don't know for certain when that channel is successfully connected unless we use the connectionFuture.

You can just call [`Channel.isActive`](https://netty.io/4.1/api/io/netty/channel/Channel.html#isActive--) on the channel, right?
</blockquote></details>

Looks like an issue is that the channel pipeline isn't established until the channel is active, so obtaining the future from the actionHandler and writing to the channel results in an immediate error since there is no actionHandler.

PR #182, opened 2019-07-19
Merge beam and GAE configs deployment to one GCB job

jianglai:
Deployment of GAE configs requires that the credential used by gcloud to
have GAE admin role of the project to be managed. We do not want to
grant the GCB service account that role, because it would all *any* GCB
job to deploy anything to GAE. Instead we use a dedicated credential
originally created to deploy beam pipelines. This credential is
encrypted by KMS and stored on GCS. Since the beam pipeline deployment
GCB job already does the decryption, it make sense to add the config
deployment step there as well. The beam deployment steps are tweaked to
use the nomulus tool docker image instead of the jar file.

Also moved the content of deploy_configs_to_env.sh to the GCB yaml file
itself because the shell script is not uploaded to GC Bat the same time
as the yaml file when the job is triggered by Spinnaker.

Lastly, due to b/137891685, using GCB to deploy cron jobs does not work
as we cannot use service account credential to deploy to projects under
google.com.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/182)


release/cloudbuild-deploy.yaml
```diff
@@ -13,39 +13,61 @@
...
+    set -e
+    gcloud auth activate-service-account --key-file=tool-credential.json
+    if [ ${_ENV} == production ]; then
+      project_id="domain-registry"
+    else
+      project_id="domain-registry-${_ENV}"
+    fi
+    gsutil cp gs://${PROJECT_ID}-deploy/${TAG_NAME}/${_ENV}.tar .
+    tar -xvf ${_ENV}.tar
+    for filename in cron dispatch dos index queue; do
```
gbrodman: Can we put a comment here linking the bug explaining why this won't work with a service account for the time being?

gbrodman: Also, if we use "set -e" and this won't work due to service account permissions, will this fail?

PR #188, opened 2019-07-23
Store only interesting failures, not login failures

gbrodman:
Login failures will happen any time that we aren't coming from a
whitelisted IP for that particular TLD. Since whitelists are out of date
(and we don't whitelist IPs for every TLD anyway) those failures aren't
interesting. Store and fully-log the interesting failures if one
happened.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/188)

### DIFF
--- a/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java
+++ b/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java
@@ -92,27 +92,23 @@ public final class UpdateRegistrarRdapBaseUrlsAction implements Runnable {
   @Inject
   UpdateRegistrarRdapBaseUrlsAction() {}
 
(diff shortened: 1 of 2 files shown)


gbrodman:
https://pantheon.corp.google.com/logs/viewer?expandAll=false&project=domain-registry&minLogLevel=0&timestamp=2019-07-23T15:22:25.029000000Z&customFacets=&limitCustomFacetWidth=true&advancedFilter=protoPayload.taskName%3D%226ea14b147898080f02329e13dec24f71%22%0AprotoPayload.taskQueueName%3D%22__cron%22%0A&dateRangeStart=2019-07-23T14:22:25.283Z&dateRangeEnd=2019-07-23T15:22:25.283Z&interval=PT1H&scrollTimestamp=2019-07-23T15:22:12.407780000Z


See here that all of the TLDs failed but most were because of whitelisting issues. The interesting one (.app) is unfortunately not logged in full because it's not the last one.

PR #191, opened 2019-07-23
Add customized .lgtm.yml

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/191)


gbrodman:
Do we have a doc on what LGTM does, out of curiosity?

PR #196, opened 2019-07-24
Rename dependency_license.gradle file

jianglai:
The file name confused the licensee app
(https://github.com/licensee/licensee) that GitHub uses to detect
license files. It thinks this file is also a license file and was not
able to determine is type.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/196)


gbrodman:
What if we made it dependency_license_report.gradle instead? Is that too long? Naming is tough.

PR #198, opened 2019-07-24
Fix deploying to alpha from workstation

hstonec:
appengineDeployAll requires appengine.deploy.version to be set
otherwiese the deployment would fail.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/198)

### DIFF
--- a/appengine_war.gradle
+++ b/appengine_war.gradle
@@ -75,6 +75,12 @@ if (project.path == ":services:default") {
 
 appengine {
     deploy {
+        // appengineDeployAll task requires the version to be set. So,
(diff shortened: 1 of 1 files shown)


appengine_war.gradle
```diff
@@ -75,6 +75,12 @@ if (project.path == ":services:default") {
 
 appengine {
     deploy {
+        // appengineDeployAll task requires the version to be set. So, 
+        // this config lets gcloud select a version name when deploying
```
jianglai: Isn't it better to always reuse the same version name for alpha? That way you don't have to deal with migrating traffic, or deleting old unused versions.

PR #200, opened 2019-07-25
Remove port to protocol map

jianglai:
There's no need for it. We can get the port from the protocol.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/200)


proxy/src/main/java/google/registry/proxy/ProxyModule.java
```diff
@@ -359,7 +351,7 @@ ProxyConfig provideProxyConfig(Environment env) {
       })
   interface ProxyComponent {
 
-    ImmutableMap<Integer, FrontendProtocol> portToProtocolMap();
+    Set<FrontendProtocol> protocols();
```
gbrodman: Can this be ImmutableSet?

PR #204, opened 2019-07-31
[wip] Registry lock mock action and UI implementation

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/204)

### DIFF
--- a/core/src/main/java/google/registry/env/common/default/WEB-INF/web.xml
+++ b/core/src/main/java/google/registry/env/common/default/WEB-INF/web.xml
@@ -55,6 +55,22 @@
     <url-pattern>/registrar-settings</url-pattern>
   </servlet-mapping>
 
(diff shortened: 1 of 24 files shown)


gbrodman:
Note for Guy and Shicong: The action setup is more or less what we'll want (at least to start) but obviously it just returns dummy data. There are a couple TODOs for when we can actually ask the question "do we support registry lock for this registrar / user?". 

There will also be a separate action and UI segment for the email that confirms the lock/unlock but it doesn't make sense to implement that yet since it'll likely be entirely separate.

gbrodman:
Oh also the build is failing due to the screenshot tests -- I didn't think it was worth updating those until we had a clearer picture.

gbrodman:
closing, #369 has the last of this content

PR #212, opened 2019-08-05
Clean up Gradle stylings and fix issues IDed in Prober

gbrodman:
We weren't applying the automatic style checks previously. Most of the Java changes are just autoformatting it (control - shift - L in IDEA defaults).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/212)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -236,9 +236,7 @@ subprojects {
     }
   }
 
-  if (project.name == 'util') return
-  if (project.name == 'proxy') return
-  if (project.name == 'core') return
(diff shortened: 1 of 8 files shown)


gbrodman:
This has the additional benefit of running the annotation processing so that IDEA can pick up the @AutoValue (et al.) generated classes.

gbrodman:
I don't know if y'all would do this anyway, but FYI this PR is way easier to review in Reviewable than in Github

PR #214, opened 2019-08-05
Add a Gradle task to deploy the proxy

jianglai:
Also refactored some common code regarding the mapping between
environments and projects.

The script to update proxy deployments and kill all pods are lifted from
https://cs.corp.google.com/piper///depot/google3/domain/registry/tools/bashrc?l=163

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/214)

### DIFF
--- a/proxy/build.gradle
+++ b/proxy/build.gradle
@@ -18,6 +18,20 @@ task buildProxyImage(dependsOn: deployJar, type: Exec) {
   commandLine 'docker', 'build', '-t', 'proxy', '.'
 }
 
+task deployProxy(dependsOn: [buildProxyImage, ':verifyDeployment']) {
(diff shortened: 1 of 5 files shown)


proxy/build.gradle
```diff
@@ -18,6 +18,20 @@ task buildProxyImage(dependsOn: deployJar, type: Exec) {
   commandLine 'docker', 'build', '-t', 'proxy', '.'
 }
 
+task deployProxy(dependsOn: [buildProxyImage, ':verifyDeployment']) {
+  doLast {
```
gbrodman: Is there any way to put these in to a single exec() command? Would that be the correct call?

PR #226, opened 2019-08-14
Add a registry lock password to contacts

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/226)


gbrodman:
I implemented what we discussed -- through the command line tool, we set a variable that allows users to set their password (and nulls out their previous password if one exists)

We don't enable the password-setting process yet. That'll be in a separate PR later.

PR #232, opened 2019-08-17
Update Gradle wrapper to 5.6

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/232)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,4 @@
+org.gradle.jvmargs=-Xmx1024m
 mavenUrl=
 pluginsUrl=
 uploaderDestination=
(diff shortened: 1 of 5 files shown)


gbrodman:
It seems like there are some issues with the Gradle upgrade and the shadow jar plugin that we use to build the Nomulus tool.

jianglai:
This is due to https://github.com/johnrengelman/shadow/issues/508. I'll push an update soon.

gradle.properties
```diff
@@ -1,3 +1,4 @@
+org.gradle.jvmargs=-Xmx1024m
```
gbrodman: Isn't this the default? Is specifying it necessary?

jianglai: Somehow it was giving me OOM error when I didn't specify this...

gbrodman: I was apparently wrong, or the posts online I found were wrong. Default is 512 MB apparently 

https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory

PR #235, opened 2019-08-19
Remove hardcoded Java home path

hstonec:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/235)


hstonec:
> The description may be confusing to non-Google readers. Better change it to "remove hardcoded Java home path".

Done.

PR #237, opened 2019-08-21
Add a "showErrorOutput" property

mindhog:
Add a property to let us dump test output and final status in real-time to the
console.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/237)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+
 buildscript {
   if (rootProject.enableDependencyLocking.toBoolean()) {
     // Lock buildscript dependencies.
@@ -254,6 +257,31 @@ subprojects {
       include "${project.relativePath}/"
     }
   }
+}
+
+// If "-P verboseTestOutput=true" is passed in, configure all subprojects to dump all of their
+// output and final test status (pass/fail, errors) for each test class.
+if (verboseTestOutput) {
+  gradle.projectsEvaluated({
+    subprojects {
+      tasks.withType(Test) {
+        testLogging {
+          events TestLogEvent.FAILED, TestLogEvent.PASSED, TestLogEvent.SKIPPED,
+              TestLogEvent.STANDARD_OUT, TestLogEvent.STANDARD_ERROR
+          exceptionFormat TestExceptionFormat.FULL
+          showExceptions true
+          showCauses true
+          showStackTraces true
+
+          afterSuite { desc, result ->
+            println "Results: ${result.resultType}, " +
+                "${result.successfulTestCount}/${result.testCount} tests " +
+                "passed, ${result.failedTestCount} failures.";
+          }
+        }
+      }
+    }
+  })
 }
 
 task checkDependenciesDotGradle {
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -3,5 +3,6 @@ pluginsUrl=
 uploaderDestination=
 uploaderCredentialsFile=
 uploaderMultithreadedUpload=
+verboseTestOutput=
```
gbrodman: Should we make this explicitly false?

build.gradle
```diff
@@ -256,6 +259,31 @@ subprojects {
   }
 }
 
+// If "-P verboseTestOutput=true" is passed in, configure all subprojects to dump all of their
+// output and final test status (pass/fail, errors) for each test class.
+if (verboseTestOutput) {
+  gradle.projectsEvaluated({
+    subprojects {
```
gbrodman: What's the reasoning for putting this here instead of in the subprojects block above?

gbrodman:
Is the thinking here that it'll be easier to debug remote build failures if the exceptions are printed to the console, rather than having to dig for remote files?

gradle.properties
mindhog: That actually doesn't work, I think gradle is just treating the property as a string and will treat it as false in a conditional only if it's empty.

build.gradle
mindhog: Yeah, I guess that merits a comment, which I've added.  The short answer is we have to apply these to all of the test targets, and they aren't defined when the main subprojects section is run.

mindhog:
> Is the thinking here that it'll be easier to debug remote build failures if the exceptions are printed to the console, rather than having to dig for remote files?

On the contrary, I added it for local builds.  I don't know how it is in an IDE, but on the shell it's a PITA to copy the link, switch over to the browser and dig down into the test failure.  Also, having test results scroll out in real-time means I can get to work on a fix faster and greatly reduces my cycle times.

mindhog:
> I see -- that's a good reason.
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/237#-:-LmtbtdM6M8QPVYBrr2T:bdtl5wk)_ status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @mindhog)
> 
> _[build.gradle, line 266 at r1](https://reviewable.io/reviews/google/nomulus/237#-Lmocg3rBIol7erYXweZ-r1-266:-Lmtbs7CCU8Eyd5njbkG:b-u5vc8w) ([raw file](https://github.com/google/nomulus/blob/75bce00a5002a65265022a813e1bf5fbec04ecb8/build.gradle#L266)):_
> 
> Previously, mindhog (Michael Muller) wrote…
> Sounds good, thanks for adding that
> 
> _[gradle.properties, line 6 at r1](https://reviewable.io/reviews/google/nomulus/237#-Lmocg3rBIol7erYXwe_-r1-6:-LmtWdU21GZDNC8pa5n4:b-u5htll) ([raw file](https://github.com/google/nomulus/blob/75bce00a5002a65265022a813e1bf5fbec04ecb8/gradle.properties#L6)):_
> 
> Previously, mindhog (Michael Muller) wrote…
> Looking at how we treat enableDependencyLocking, we'd have to say "if (verboseTestOutput.toBoolean()) {.."
> 
> Thoughts? I think either probably works

We should probably be consistent with enableDependencyLocking.  Done.

PR #238, opened 2019-08-22
Do not upload nomulus.jar file to GCS

jianglai:
We are using the docker image instead.


jianglai:
Does it? For me I didn't experience too much delay. For example:

```
$ time nomulus -e production get_tld goog > /dev/null

real    0m4.721s
user    0m0.252s
sys     0m0.165s
```

Can you give me an example command that takes too long?

My goal is to get rid of unused code. If we don't use the .jar file we shouldn't upload it. If indeed we find that the docker image is impeding productivity then we should consider using the .jar, but then again we are faced with the same problem that made us try docker in the first place -- how to fix the Java version.

PR #243, opened 2019-08-26
Add RegistryLock SQL schema

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/243)


gbrodman:
PTAL, should be good to go here thanks

PR #248, opened 2019-08-30
Output command test output as well as consuming it

mindhog:
CommandTestCase currently consumes stdout & stderr for the command being
tested.  Unfortunately, this results in us not being able to see the command
output.  Add an output splitter so that output gets written to the original
stream in addition to being captured.

A simpler approach would be to print the captured data after command
completion.  However, this won't work for tests that become hung and also
won't display output in real-time.

Tested: Ran a command test with verboseTestOutput=true, verified that standard
output was visible.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/248)

### DIFF
--- a/core/src/test/java/google/registry/tools/CommandTestCase.java
+++ b/core/src/test/java/google/registry/tools/CommandTestCase.java
@@ -36,6 +36,7 @@ import google.registry.tools.params.ParameterFactory;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.PrintStream;
 import java.util.List;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


core/src/test/java/google/registry/tools/CommandTestCase.java
```diff
@@ -75,8 +83,22 @@ public final void beforeCommandTestCase() throws Exception {
     // Ensure the UNITTEST environment has been set before constructing a new command instance.
     RegistryToolEnvironment.UNITTEST.setup(systemPropertyRule);
     command = newCommandInstance();
-    System.setOut(new PrintStream(stdout));
-    System.setErr(new PrintStream(stderr));
+
+    // Capture standard output/error.  This is problematic because gradle tests run in parallel in
```
jianglai: Remove the extra space between "output/error." and "This".

jianglai:
Does this mean that all `CommandTestCase`s are effectively serial..., at least the IO part?

jianglai:
LGTM.

PR #261, opened 2019-09-06
[wip] Use sql-formatter-cli to format SQL source files

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/261)


db/format_sql.sh
```diff
@@ -0,0 +1,54 @@
...
+VERB=$1
+GLOB=$2
+
+failed_files=""
+for sql_file in $(find ${GLOB} -type f -name "*.sql"); do
+  create_formatted_file ${sql_file}
+  if [ "$1" = "apply" ]; then
+    # Note: the formatter isn't necessarily idempotent -- the output of the first run might
+    # be different from the output of running it again. Run it a few times so it stabilizes.
+    for i in {1..10}; do
```
CydeWeys: 10 is more than "a few times".  How about 3?  Do we really think it might need more than that?

db/src/main/resources/sql/schema/claims_list.sql
```diff
@@ -11,13 +11,11 @@
 -- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
-
```
CydeWeys: It doesn't like blank lines between different statements, huh?  Is that an option?  I think having blank lines in schema files makes things more readable.

gbrodman: I agree that it'd make it more reasonable but there aren't really configurable options in this formatter.

Other formatters that I've seen that have configurable options still have issues like the extra spaces at the beginning of each line (this was a common theme among a few formatters, strangely enough)

db/format_sql.sh
gbrodman: 3 is probably fine

gbrodman:
After more consideration I'm not sure this is worth it -- it seems like there's no clean single standard for SQL files the way there is for Java, and we'd lose the flexibility to format SQL statements however we want for clarify. Thoughts on potentially closing this?

gbrodman:
Yeah, I'm going to close this. We can always reopen it if we want.

PR #262, opened 2019-09-06
Add a task to run the nomulus tool

jianglai:


jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/262)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)
> 
> _[core/build.gradle, line 596 at r1](https://reviewable.io/reviews/google/nomulus/262#-Lo6yX3l7micOr_v8i96:-Lo6yX3m9Vkocd6mOxXp:bdi9xiv) ([raw file](https://github.com/google/nomulus/blob/b12e97dbc48fec1378c878c3faa6fbef3ef750cd/core/build.gradle#L596)):_
> 
> > ```groovy-gradle
> > // To run the nomulus tool:
> > // gradle registryTool --args="foo --bar"
> > ```
> 
> There's no better way to handle args? Wrapping (and escaping) args into another arg is ugly.

See cl/267645678 as an internal solution. Using `--args` is the recommended way to pass args (https://docs.gradle.org/current/userguide/application_plugin.html#sec:application_usage).

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/262)_ status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)
> 
> _[core/build.gradle, line 600 at r1](https://reviewable.io/reviews/google/nomulus/262#-Lo6zK2M12IH0lt17cXn:-Lo6zK2M12IH0lt17cXo:b-jbai9s) ([raw file](https://github.com/google/nomulus/blob/b12e97dbc48fec1378c878c3faa6fbef3ef750cd/core/build.gradle#L600)):_
> 
> > ```groovy-gradle
> >   classpath = sourceSets.main.runtimeClasspath
> >   main = 'google.registry.tools.RegistryTool'
> > }
> > ```
> 
> Unless you anticipate using the tool from within the build, it might be better to do this with a (very small) shell script. Gradle has a lot of startup overhead and adds a lot of extraneous output, and this approach likely wouldn't work at all for "nomulus shell" or anything that consumes standard input.

The purpose of this task is to add the ability to easily test changes to the nomulus tool, not day-to-day operations.

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/262)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai)
> 
> _[core/build.gradle, line 596 at r1](https://reviewable.io/reviews/google/nomulus/262#-Lo6yX3l7micOr_v8i96:-LoQw2351ErvEv03jT9C:b547w8p) ([raw file](https://github.com/google/nomulus/blob/b12e97dbc48fec1378c878c3faa6fbef3ef750cd/core/build.gradle#L596)):_
> 
> Ugh. I don't like it, but if there's no alternative ... :/
> 
> So this doesn't solve the problem of replacing what we used to have, which is a simple target to run a locally built version of nomulus tool. Can we write a new bashrc target that handles --args and escaping so that we can run a locally built tool easily?

Isn’t cl/267645678 exactly what you described...?

PR #263, opened 2019-09-06
Create EntityManagerFactory for App Engine

hstonec:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/263)


core/src/main/java/google/registry/persistence/EntityManagerFactoryProvider.java
```diff
@@ -70,4 +85,57 @@ public static EntityManagerFactory create(String jdbcUrl, String username, Strin
...
+      }
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+
+    String instanceConnectionName = credentialFields[0];
+    String username = credentialFields[1];
+    String password = credentialFields[2];
+
+    ImmutableMap.Builder<String, String> overrides = ImmutableMap.builder();
```
CydeWeys: new ImmutableMap.Builder<>()

core/src/main/java/google/registry/persistence/EntityManagerFactoryProvider.java
```diff
@@ -70,4 +85,57 @@ public static EntityManagerFactory create(String jdbcUrl, String username, Strin
         "Persistence.createEntityManagerFactory() returns a null EntityManagerFactory");
     return emf;
   }
+
+  /** Constructs the {@link EntityManagerFactory} instance for the App Engine application. */
+  public static EntityManagerFactory createForAppEngine() {
```
CydeWeys: What are you envisioning would call this function?  Does this function have to be static?  If you could change it to an instance method then you could use Dagger to @Inject the various configuration options it depends on.

mindhog:
Seems like most of the work in this has moved on to other PRs.  Is it ok to close this one?

PR #267, opened 2019-09-11
Implement UserType for CreateAutoTimestamp

mindhog:
Implement CreateAutoTimestampType, which allows us to persist
CreateAutoTimestamp objects to hibernate in the same way that we do to
objectify.

The current implementation comes with caveats:
- Since CreateAutoTimestamp is immutable, it is currently necessary to evict
  the object from the cache immediately after persisting it to prevent future
  queries from using the cached version which does not have the correct
  timestamp.  An alternate approach would be to allow only the user type to
  mutate the underlying timestamp.
- We currently store the ZonedDateTime through OffsetDateTime, which is the
  only way that JDBC knows how to do it out of the box.  This strips the
  timezone data because an offset isn't adequate to capture that.  We've decided that this is
  acceptable.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/267)


CydeWeys:
We can close this now right?

PR #273, opened 2019-09-16
Update registryTool task to better handle command line arguments

jianglai:
Tokens are delimited by pipes and can be escaped. If we use -args
directly, we will not be able to escape the delimiter, if both the
double and single quotes happen to be present in the arguments.

See:

https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/JavaExec.html#setArgsString-java.lang.String-

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/273)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -593,10 +593,34 @@ task findGoldenImages(type: JavaExec) {
 }
 
 // To run the nomulus tool:
-// gradle registryTool --args="foo --bar"
+// gradle registryTool --PtoolArgs="foo|--bar|ba\\|z"
 task registryTool(type: JavaExec) {
   classpath = sourceSets.main.runtimeClasspath
   main = 'google.registry.tools.RegistryTool'
+
+  doFirst {
+    def delimiter = '|'
+    def toolArgs = rootProject.findProperty("toolArgs")
+    if (toolArgs == null) {
+      throw new GradleException("Must supply Nomulus tool arguments in " +
+          "-PtoolArgs=\"token1|token2|...\" format")
+    }
+    toolArgs += delimiter
+    def argsList = []
+    def currArg = ''
+    for (def i = 0; i < toolArgs.length(); i++) {
+      def currChar = toolArgs[i]
+      if (currChar != delimiter) {
+        currArg += currChar
+      } else if (i != 0 && toolArgs[i - 1] == '\\') {
+        currArg = currArg.substring(0, currArg.length() - 1) + currChar
+      } else {
+        argsList.add(currArg)
+        currArg = ''
+      }
+    }
+    args = argsList
+  }
 }
 
 task generateGoldenImages(type: Test) {
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -592,34 +592,39 @@ task findGoldenImages(type: JavaExec) {
   args arguments
 }
 
-// To run the nomulus tool:
-// gradle registryTool --PtoolArgs="foo|--bar|ba\\|z"
+// To run the nomulus tool with these command line tokens:
+// "--foo", "bar baz", "--qux=quz"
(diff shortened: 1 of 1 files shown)


jianglai:
> What's the game plan for supporting command line usage by humans? (I.e. not just dr-tool)
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/273#-:-Lp9uG_j-LjuGETrdgsm:b-yizd2c)_ status: :shipit: complete! all files reviewed, all discussions resolved

PTAL. I've changed it so that `dr-tool` can use the pipe delimited format whereas directly calling the task can use the simpler `--args=` format.

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/273)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)
> 
> _[core/build.gradle, line 596 at r3](https://reviewable.io/reviews/google/nomulus/273#-LpA4tZOEQnyK6YNWt8C:-LpA4tZOEQnyK6YNWt8D:brdexso) ([raw file](https://github.com/google/nomulus/blob/3e0631cff937901dcc366351ea36b863ddb582cd/core/build.gradle#L596)):_
> 
> > ```groovy-gradle
> > // To run the nomulus tool:
> > //   gradle registryTool --args="foo '--bar baz'"
> > ```
> 
> I'm not understanding why the single quoting is necessary in this example?
> 
> Wouldn't the following do the same thing?
> 
> --args="foo --bar baz"
> 
> This might be easier to understand using real parameters that might actually be used, such as `--tld foo` or `-c TheRegistrar`

OK This is not the best example that I can come up with... Updated with more sensible examples to show that the single quote is used to group `bar` and `baz` together as a single parameter.

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/273)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)
> 
> _[core/build.gradle, line 596 at r3](https://reviewable.io/reviews/google/nomulus/273#-LpA4tZOEQnyK6YNWt8C:-LpA7VKY7sWiA0aNc0SF:bdb2nv9) ([raw file](https://github.com/google/nomulus/blob/3e0631cff937901dcc366351ea36b863ddb582cd/core/build.gradle#L596)):_
> 
> Can you add another arg to the example?
> 
> Also, I wouldn't really say that "--foo" and "bar baz" are two parameters; rather, I'd say that --foo is the parameter name and 'bar baz' is the value being passed to it. Note that we generally configure JCommander to not require the use of an equal sign; a space is typically allowable as well.

Done.

They are really just tokens as parsed by the shell. The fact that JCommand can treat one token "--foo=bar" and two tokens "foo" "bar" the same doesn't really matter -- it happens after tokenization. Here we are trying to provide the Gradle task with these tokens, without regard to any semantics that might be significant to JCommander.

jianglai:
> > Reviewed 1 of 1 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/273)_ status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)
> 
> _[core/build.gradle, line 596 at r3](https://reviewable.io/reviews/google/nomulus/273#-LpA4tZOEQnyK6YNWt8C:-LpDnUchBfZGnuFGQBCE:bpuzy0b) ([raw file](https://github.com/google/nomulus/blob/3e0631cff937901dcc366351ea36b863ddb582cd/core/build.gradle#L596)):_
> 
> > ```
> > `--foo`, `bar baz`, `--qux=quz`
> > ```
> 
> ...nomulus tool *with these command...
> 
> (I was momentarily sent into a panic over the back-ticks, btw)

Changed to double quotes to ease Mike's panic attack :P

PR #290, opened 2019-09-27
Add a DAO for RegistryLock objects

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/290)


gbrodman:
Dependent on #282

core/src/main/java/google/registry/schema/domain/RegistryLock.java
```diff
@@ -57,11 +57,13 @@
...
-        @Index(
-            name = "idx_registry_lock_repo_id_revision_id",
-            columnList = "repo_id, revision_id",
-            unique = true))
+    indexes = {
+      @Index(
+          name = "idx_registry_lock_repo_id_revision_id",
+          columnList = "repo_id, revision_id",
+          unique = true),
+      @Index(name = "idx_registry_lock_verification_code", columnList = "verificationCode")
```
gbrodman: Of note: it seems like Hibernate actually requires "verificationCode" here, not "verification_code". It failed to create this index in the schema tests otherwise.

gbrodman: btw it's because until it gets finally written out, Hibernate apparently treats the column as "verificationCode" whereas the Column annotations are applied pretty much immediately (this is why repo_id and revision_id work in the previous index annotation)

gbrodman:
@weiminyu I added code that allows us to modify this in place, assuming we're in the same transaction. Would you be able to take a look? Thanks!

PR #293, opened 2019-09-30
Add persistence.xml to the war files

mindhog:
This fixes the App Engine build.  The runtime now needs access to this.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/293)

### DIFF
--- a/appengine_war.gradle
+++ b/appengine_war.gradle
@@ -38,6 +38,10 @@ war {
     from("${coreResourcesDir}/google/registry/ui/html") {
         include "*.html"
     }
+
+    from("${coreResourcesDir}/META-INF/persistence.xml") {
+        include "*.html"
+    }
 }
 
 if (project.path == ":services:default") {
--- a/appengine_war.gradle
+++ b/appengine_war.gradle
@@ -26,20 +26,21 @@ project.convention.plugins['war'].webAppDirName =
 
 apply plugin: 'com.google.cloud.tools.appengine'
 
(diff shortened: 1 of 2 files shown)


appengine_war.gradle
```diff
@@ -38,6 +38,10 @@ war {
     from("${coreResourcesDir}/google/registry/ui/html") {
         include "*.html"
     }
+
+    from("${coreResourcesDir}/META-INF/persistence.xml") {
+        include "*.html"
```
gbrodman: I think the .html part might be a remnant of a copy+paste?

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/293)_ status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)
> 
> _[core/src/main/java/google/registry/model/transaction/TransactionManagerFactory.java, line 30 at r2](https://reviewable.io/reviews/google/nomulus/293#-LqNUxVb-a3SHKs0LA1G:-LqNWyzC-qF6wEQafzic:bngyk3b) ([raw file](https://github.com/google/nomulus/blob/f8394a5188137d292f257bf264477c2960527ddc/core/src/main/java/google/registry/model/transaction/TransactionManagerFactory.java#L30)):_
> 
> :-) I figured he was more in the path of the dependencies. I had planned to communicate the details to him out of band, do you think that there should be more explanation in the code itself?

I mean if we are going to keep this code for a while (and it looks like we will), it think it is better to document the reasons in the code.

PR #300, opened 2019-10-07
Refactor environment -> project map out of build

mindhog:
Move the project mapping into its own file so that it can be more easily
overriden.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/300)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -119,13 +119,9 @@ task stage {
 
 def environments = ['production', 'sandbox', 'alpha', 'crash']
 
-// TODO(mmuller): Move this into internal specialization code.
-def projects = ['production': 'domain-registry',
-                'sandbox'   : 'domain-registry-sandbox',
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -119,14 +119,10 @@ task stage {
...
 
-// TODO(mmuller): Move this into internal specialization code.
-def projects = ['production': 'domain-registry',
-                'sandbox'   : 'domain-registry-sandbox',
-                'alpha'     : 'domain-registry-alpha',
-                'crash'     : 'domain-registry-crash']
-
 def gcpProject = null
 
+apply from: "${rootDir.path}/projects.gradle"
```
jianglai: Sometimes I felt this kind of `apply` is quite invasive. We are just importing everything in the other file into this namespace, unqualified. Can we explicitly only import the `projects`?

PR #316, opened 2019-10-21
Implement transactNew() in JpaTransactionManagerImpl

hstonec:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/316)


hstonec:
This PR is not needed any more, the current `transactNew` implementation is just the same as `transact` in `JpaTm`, and we will delete this method after DB migration.

PR #318, opened 2019-10-21
Fix security alerts by upgrading

gbrodman:
Note: this will be incomplete until Puppeteer releases a new version as we await their https-proxy-agent version bump

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/318)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,55 +30,14 @@
       }
     },
     "anymatch": {
-      "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


jianglai:
Don't you have to upgrade the package lock file as well?

gbrodman:
I did

jianglai:
Ah, sorry. The diff is so big so GitHub decides to skip showing it...

PR #320, opened 2019-10-21
Instrument proxy frontend latency metric

jianglai:
Backend latency only captures the time spent waiting for the GAE backend
to respond to a request, which is not representitive of what clients
experience. For instance it does not take into account the time the
proxy spends on processing the requests and the time it takes to send a
response to the client.

This PR adds a metric for frontend latency, which is the time spent on
answering a client request. This should serve as a better proxy for the
latency clients observe.

### DIFF
--- a/proxy/build.gradle
+++ b/proxy/build.gradle
@@ -18,7 +18,8 @@ task buildProxyImage(dependsOn: deployJar, type: Exec) {
   commandLine 'docker', 'build', '-t', 'proxy', '.'
 }
 
-task deployProxy(dependsOn: [buildProxyImage, ':verifyDeployment']) {
+task deployProxy(dependsOn: buildProxyImage) {
+  configure verifyDeploymentConfig
   doLast {
     exec {
       commandLine 'docker', 'tag', 'proxy', "gcr.io/${rootProject.gcpProject}/proxy"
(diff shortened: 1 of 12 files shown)


jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/320)_ status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)
> 
> _[proxy/src/main/java/google/registry/proxy/handler/FrontendMetricsHandler.java, line 113 at r1](https://reviewable.io/reviews/google/nomulus/320#-Lrua0ZX8seJtmbf5O_m:-LsIVBgIBfGhp-tS26Kq:b-w9ny8x) ([raw file](https://github.com/google/nomulus/blob/0565dde2195aa41ab6cb764a26320f75d2273d16/proxy/src/main/java/google/registry/proxy/handler/FrontendMetricsHandler.java#L113)):_
> 
> Previous metrics are not still ongoing though, right?
> 
> When a given command fails, all _ongoing_ metrics (and subsequent commands on the same connection i.e. pipeline) are jettisoned, right?

Correct. If one of the write fails the ones following them won't go through, so no metrics will be recorded for them. The ones before the failed write should be recorded.

PR #324, opened 2019-10-23
Add another test contact for Registry Lock testing

gbrodman:
Previously, we only had two contacts -- one per registrar. This PR adds
a second, registry-lock-enabled, contact to one registrar for two
reasons:

1. For registry-lock-related testing, we'd like to be able to test both
positively and negatively, making sure that the permissions work the way
they should
2. In general, the UI tests should include the case where we have
multiple contacts in the same registrar. Previously, this was never the
case in tests.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/324)


gbrodman:
dependent on https://github.com/google/nomulus/pull/317

PR #331, opened 2019-10-26
Make compilation work on Windows

jianglai:
The code should compile on Windows now. Note that it certainly wouldn't run or pass tests on Windows (I haven't tried), but it should make it possible to develop on Windows (like what I'm doing when I'm home).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/331)


mindhog:
Meta-question: how will we avoid breaking the Windows build?  Do we care?

PR #332, opened 2019-10-29
Save current deployment tag for every environment

weiminyu:
Store tag of the current deployment in each environment at then end of the deployment pipeline.
This tag is used by the server-sql compatibility test.


jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/332)_ status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai)
> 
> _[release/cloudbuild-deploy.yaml, line 60 at r1](https://reviewable.io/reviews/google/nomulus/332#-LsSFuNBDgKPwMbWEKwd:-LsSSaBy0QHYW_5td8UD:b-xauf28) ([raw file](https://github.com/google/nomulus/blob/638170caadbb08674bdc124b6e06e10faf3f4fbd/release/cloudbuild-deploy.yaml#L60)):_
> 
> Made this a separate step.
> 
> The other file only makes up for the lack of symbolic link, and tells us the version of whatever in the live folder.
> We can find out this info by looking into the content of the archives, but this file makes it simpler.
> 
> The files created here are per-environment and server as the authoritative records of what is currently deployed, which may be different the 'live' release during some of the deployment stages, or after stalled pushes or rollbacks.

My point is that with the `nomulus.${ENV}.tag` file, you already know what tag production uses, so saving the other file is redundant. But that's only one file, so I don't think it's a big deal.

PR #333, opened 2019-10-30
Upgrade org.apache.beam related packages to latest version

hstonec:


jianglai:
Looks like this update breaks beam pipeline deployment:

https://pantheon.corp.google.com/cloud-build/builds/9a44b5c8-2620-4776-923d-2a42ac3ca570;step=2?project=domain-registry-dev

This stage should only fail at the last step but now it fails at spec11 deployment. I think we probably should separate out the last step so that we don't mistake failures at other steps as expected.

hstonec:
I will take a look if there is a quick fix.

PR #336, opened 2019-10-31
Add verification action for registry locks

gbrodman:
This is the third part of the (not-in-order) backend PRs for Registry Lock. 

After a user requests a lock on a domain, they will receive an email link with a UUID to verify the lock -- they must click on this link within one hour. This PR introduces the action that accepts that verification request. We verify the request (valid UUID, has permissions, etc) and then perform the lock on the domain if it's appropriate. 

Note that if any step fails in the domain-lock process, we will back out of both the SQL and Datastore transactions and the user could try again. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/336)


gbrodman:
Thanks, Shicong. 

Ben, I've refactored this so we use the HtmlAction we created -- PTAL when you get a chance, thanks!

gbrodman:
old, doing #461 instead

PR #337, opened 2019-10-31
Prevent test from changing golden schema

hstonec:
When we add the extra test entity to the current `JpaTransactionManagerRule` by calling `jpaRule.withEntityClass(TestEntity.class)` and `jpaRule.withProperty(Environment.HBM2DDL_AUTO, "update")`, the rule would override the golden database scheme with the schema derived from the current entity class(This is an expected behavior by enabling `HBM2DDL_AUTO`). This behavior prevents us from detecting breaking changes in ORM classes.

This PR fixed the issue by adding `HibernateSchemaExporter` to export the DDL script for given extra entity class, and make `JpaTransactionManagerRule` execute the DDL script to create the corresponding table for the extra entity when initializing the database. This PR also simplified the code when adding extra entity class for testing. For now, you don't need to(and shouldn't) call `jpaRule.withProperty(Environment.HBM2DDL_AUTO, "update")`.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/337)

### DIFF
--- a/core/src/main/java/google/registry/persistence/HibernateSchemaExporter.java
+++ b/core/src/main/java/google/registry/persistence/HibernateSchemaExporter.java
@@ -1,0 +1,104 @@
+// Copyright 2019 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


mindhog:
>  the table name was not compatible with our name convention because of the override from HBM2DDL_AUTO.

Why not just change the table name?  It's explicitly specified in the Entity annotation for TestEntity.

Regarding removal of withProperty(), we might just have always done that in cases where withEntityClass() is called.

> if we want to test some change to the golden schema and need an extra entity, 

We don't really have an example of this case and it's difficult for me to envision one.  I expect you'd typically want to test only real classes against the golden schema.

On balance, I'm not opposed to this change.  I don't think it was necessary, but it's also not detrimental.

hstonec:
I think you are right, we probably won't gain as much as we thought before. Though, this is also a trivial change so I didn't spend too much time on it.

PR #340, opened 2019-11-01
Make Joda Money embeddable

jianglai:
This PR adds XML mappings to make Joda Money an embeddable field. When an entity contains a Money field is persisted, it will be stored in two fields (amount, currency). If the column names collide with other columns, they can be re-named on a case-by-base basis using `@AttributeOverrides`.

See `JodaMoneyConverterTest.java` for an example. If not all columns of `Money` need to be overridden, the `@AttirbuteOverrides` annotation can include just one `@AttributeOverride`.

`Money` can also be part of a collection (values in a map for instance) and to override the the columns the `name` filed in the `@AttributeOverride` annotation needs to be prefixed with `value.`. See examples in `JodaMoneyConverterTest.java`.

Note that when `Money` is a field persisted in an entity, a `@PostLoad` callback is necessary to set the amount to the correct scale according to the currency after the entity is loaded from the database. See the test file for examples.

Also updates Joda Money version to 1.0.1 which includes an no-op default constructor for `Money` and `BigMoney` so that Hibernate can instantiate them reflectively (see https://github.com/JodaOrg/joda-money/issues/23).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/340)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -534,8 +534,8 @@ task compileProdJS(type: JavaExec) {
 
 compileJava.dependsOn jaxbToJava
 compileJava.dependsOn soyToJava
-// The Closure JS compiler does not support Windows. It is fine to disable it if all we want to do
-// is to complile the Java code on Windows.
+// The Closure JS compiler does not support Windows. It is fine to disable it if
+// all we want to do is to complile the Java code on Windows.
 if (!System.properties['os.name'].toLowerCase().contains('windows')) {
   compileJava.dependsOn compileProdJS
   assemble.dependsOn compileProdJS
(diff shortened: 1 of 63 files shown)


jianglai:
OK there is one more problem -- `Money` will set the size of the `BigDecimal` in `BigMoney` depending on the currency. So JPY has size 0 and USD has size 2. If we have mixed currencies stored in the same table (e. g. Registry) the column for amount can only have one size. If we set it to size 2,  the `Money` loaded from DB will all have size 2 and breaks the contract for JPY. This would be solvable if we control the `Money` class. We can add a `@PostLoad` method to manually correct the size based on the currency.

We can either remember to add a `@EntityListener` to each entity that contains `Money` and correct the size, or load a default `EntityListener` for all entities, which reflectively iterate through all fields in any entity and fix the size.

Both of these are not ideal.

jianglai:
PTAL. I added an example on how to use the `@PostLoad` callback to correct the scale.

PR #342, opened 2019-11-01
Move soyutils_usegoog.js out of node_modules

jianglai:
Every time the npmInstall runs, it removes this file from node_modules.
Move it outside the folder to prevent this from happening.

Also moves the karma config file out of the root directory.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/342)


jianglai:
> Out of curiosity, why more the Karma configuration file?
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/342#-:-LsgrecM82OvXT_jJ5LV:b-jrlxre)_ status: 0 of 5 files reviewed, all discussions resolved

I just thought that we'd want to move all js files out of the root directory.

jianglai:
> Because the Karma config file is more of a testing configuration than actual source code I think we want to keep it in the root folder.
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/342#-:-LsqzFP35aRCQ-n0xoBT:bml6km)_ status: 0 of 5 files reviewed, all discussions resolved

Would it be better if it is in `core/src/test/`? It should definitely inside `core` because all Karma tests are in `core`, right?

PR #345, opened 2019-11-05
Clean up Karma config file

gbrodman:
soyutils_usegoog.js (the dependency) has been moved into core/src/main/javascript/ so it's covered by the other rules in the file -- this was a special-cased rule that we had before that is no longer necessary 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/345)

### DIFF
--- a/core/karma.conf.js
+++ b/core/karma.conf.js
@@ -43,10 +43,6 @@ module.exports = function(config) {
         included: false
       },
       {
-        pattern: 'node_modules/soyutils_usegoog.js',
-        included: false
-      },
-      {
         pattern: 'node_modules/google-closure-library/closure/goog/deps.js',
         included: false,
         served: false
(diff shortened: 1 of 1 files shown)


jianglai:
Why is the default sufficient? It is not obvious to me that we don’t need to explicitly reference that fine any more.

gbrodman:
We already include core/src/main/javascript/**/*.js in other parts of the file, which covers soyutils_usegoog.js

This part of the config should have been removed when we moved soyutils_usegoog.js; I should have noticed.

jianglai:
> We already include core/src/main/javascript/**/*.js in other parts of the file, which covers soyutils_usegoog.js
> 
> This part of the config should have been removed when we moved soyutils_usegoog.js; I should have noticed.

Aha, I see. Can you modify the PR description to include the reasoning as well?

PR #362, opened 2019-11-12
Add a POST action for registry locks

gbrodman:
This is the second of the three asynchronously-pushed Registry Lock actions. This one creates the lock object, saves it as pending to the DB, and sends the email to the contact with the verification link. It does not actually change the domain at all -- that occurs in the verification action 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/362)


gbrodman:
doing #459 instead since this is out of date

PR #366, opened 2019-11-14
Save an admin lock object during (un)lock domain commands

gbrodman:
We would like for both the command-line and the web console lock actions to create the lock object so that we have a single source of truth as to what lock/unlock actions have been performed on a domain. This also gets us the necessary benefit of non-admins not being allowed, through the web console, to unlock domains that were locked with the command-line tool

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/366)


gbrodman:
This PR is now moot thanks to #390. Thanks.

PR #371, opened 2019-11-15
Make dev project configurable

jianglai:
We should not hardcode our dev project in the public config file.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/371)

### DIFF
--- a/db/build.gradle
+++ b/db/build.gradle
@@ -73,12 +73,12 @@ ext {
   getCloudSqlCredential = { env, role ->
     def command =
         """gsutil cp \
-           gs://domain-registry-dev-deploy/cloudsql-credentials/${env}/${role}_credential.enc - | \
+           gs://${rootProject.ext.devProject}-deploy/cloudsql-credentials/${env}/${role}_credential.enc - | \
            base64 -d | \
            gcloud kms decrypt --location global --keyring nomulus-tool-keyring \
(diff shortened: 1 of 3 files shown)


jianglai:
Ping?

jianglai:
> > Reviewed 2 of 2 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/371)_ status: all files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)
> 
> _[db/build.gradle, line 76 at r1](https://reviewable.io/reviews/google/nomulus/371#-LuJi9-o3rs324v5LqIJ:-LuJi9-pAgGJRx7ympW2:b-4qqjha) ([raw file](https://github.com/google/nomulus/blob/49c58cecc157d25c54784ef626acda32d38aeb45/db/build.gradle#L76)):_
> 
> > ```groovy-gradle
> >     def command =
> >         """gsutil cp \
> >            gs://${rootProject.ext.devProject}-deploy/cloudsql-credentials/${env}/${role}_credential.enc - | \
> > ```
> 
> I don't think this will work. As I recall, we used the dev project for storing our credentials for all environments, and external users would have to do something similar.

I don't quite understand your comment. Isn't this precisely the point why we want to make this variable configurable? There's a corresponding CR in the Gerrit that sets this value for our project.

Issue #353, opened 2019-11-07
Schema publication fails in alpha


jianglai:
When building the release artifacts for alpha the schema publication [step](https://github.com/google/nomulus/blob/135ab66e55bb5804118358eedc0b7a976d339030/release/cloudbuild-nomulus.yaml#L68) failed because alpha cloud build service account does not have write access to domain-registry-maven-repository. We obviously don't want to write alpha schema to a repo shared by production.

The comment says this is tentative. So do we have a permanent, environment-agnostic solution yet?

jianglai:
@weiminyu Does it make sense to have the schema hosted on `gs://domain-registry-maven-repository`? Shouldn't it it be in a private repo? The Maven mirror is public and we don't want our schema to be public, do we?

PR #383, opened 2019-11-22
Stop publish Cloud SQL schema jar to maven repo

weiminyu:
The original purpose of the maven publication is for
use by the server/schema compatibility tests. A command line
flag can direct a test run to use difference versions of
the schema jar. However, this won't work due to dependency
locking.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/383)


jianglai:
Solves https://github.com/google/nomulus/issues/353

PR #390, opened 2019-11-26
Refactor lock/unlock commands to use Registry Locks

gbrodman:
This involves the creation of a shared utils class that will also be used for UI-created registry locks. This also means going away from lock-command-as-EPP-command which should be fine. The utils class will need to take care of history entry / billing event creation anyway and that's all we needed from the EPP action.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/390)


gbrodman:
Ben, thoughts on this approach of unifying the validation / application of RegistryLock objects into a single utils class? This means that we can simplify (Un)LockDomainCommand and have those commands use this class instead of EPP commands. 

When creating / verifying locks in requests from the UI, we'll mostly be able to just call these methods and be done, as well.

gbrodman:
note: i'd want to add tests to the utils class, just looking for thoughts on the approach

PR #399, opened 2019-12-05
Adding junit back into the runtime classpath

mindhog:
Unfortunately, GenerateSqlSchemaCommand depends on junit via testcontainers.
We should really move GenerateSqlSchemaCommand out of nomulus tool (we only
use it during development) but this gets nomulus tool working for the time
being.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/399)


mindhog:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/399)_ status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)
> 
> _[core/src/main/java/google/registry/tools/RegistryCli.java, line 109 at r1](https://reviewable.io/reviews/google/nomulus/399#-LvLpujJDz9ihX8TyA36:-LvLpujJDz9ihX8TyA37:b-co5h22) ([raw file](https://github.com/google/nomulus/blob/7de915abc51ba4eb3b1c6578c095ab76514899f3/core/src/main/java/google/registry/tools/RegistryCli.java#L109)):_
> 
> > ```
> > System.err.println("adding command " + entry.getKey());
> > ```
> 
> Do you want to keep this?

Nope :-) Removed.

PR #404, opened 2019-12-06
Add a set of dependent resource IDs to RegistryLock

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/404)


gbrodman:
punting since we don't want to do this (for now)

PR #405, opened 2019-12-09
Make devProject a project property

weiminyu:
Properties set in rootProject's ext block are not overridden by
commandline flags.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/405)


jianglai:
It should be overridden in a file. I don't think we need to change this in the command line. I had https://team-review.git.corp.google.com/c/domain-registry-eng/nomulus-internal/+/599522 out for review but it probably got lost.

PR #438, opened 2020-01-09
Get persistence of DomainBase actually working

mindhog:
Fix all of the existing problems with DomainBase persistence:
- Remove "final" keywords on getters that cause errors during startup.
- Remove Transient from creationTime (since there's a converter for
  CreateAutoTimestamp)
- Fix DesignatedContext persistence so that it only creates a single table.
  This is a lot more efficient given that these are many-to-one with their
  domains.
- Add a flyway script, update the golden schema.
- Create a unit test, add it to the integration test suite.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/438)

### DIFF
--- a/core/src/main/java/google/registry/model/EppResource.java
+++ b/core/src/main/java/google/registry/model/EppResource.java
@@ -51,6 +51,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import javax.persistence.MappedSuperclass;
+import javax.persistence.Transient;
 import org.hibernate.annotations.Type;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
@@ -126,6 +127,7 @@ public abstract class EppResource extends BackupGroupRoot implements Buildable {
    *
    * @see google.registry.model.translators.CommitLogRevisionsTranslatorFactory
    */
+  @Transient
   ImmutableSortedMap<DateTime, Key<CommitLogManifest>> revisions = ImmutableSortedMap.of();
 
   public final String getRepoId() {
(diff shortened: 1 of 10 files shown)


mindhog:
___
*[core/src/main/java/google/registry/model/eppcommon/StatusValue.java, line 170 at r3](https://reviewable.io/reviews/google/nomulus/438#-Lyovj8iDOysz1rkewbM:-LzO931j8OBsLKi-nnMB:b-jl55ot) ([raw file](https://github.com/google/nomulus/blob/9f3c8eaf1771d5576c2b2d1f404214267fc01b1a/core/src/main/java/google/registry/model/eppcommon/StatusValue.java#L170)):*

Enums.getIfPresent() doesn't do it?
</blockquote></details>

That still requires the runtime class.

mindhog:
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 94 at r3](https://reviewable.io/reviews/google/nomulus/438#-LyowDZCEN6GZLlzrQNr:-LzOArNz2P-5-QwdsKuy:b-8p1hgy) ([raw file](https://github.com/google/nomulus/blob/9f3c8eaf1771d5576c2b2d1f404214267fc01b1a/db/src/main/resources/sql/schema/nomulus.golden.sql#L94)):*

This file hasn't been updated yet?  A lot of these should no longer be here, but still are.
</blockquote></details>

I think you reviewed the next-to-last revision.

mindhog:
___
*[core/src/main/java/google/registry/model/eppcommon/StatusValue.java, line 170 at r3](https://reviewable.io/reviews/google/nomulus/438#-Lyovj8iDOysz1rkewbM:-LzODI0EBpphdrMlv9A7:b-v28du2) ([raw file](https://github.com/google/nomulus/blob/9f3c8eaf1771d5576c2b2d1f404214267fc01b1a/core/src/main/java/google/registry/model/eppcommon/StatusValue.java#L170)):*

Take a look at how it's handled in e.g. `EnumParameter.java`?  This is possible to get.
</blockquote></details>

It's possible in that case because we have an instance of the object so we can call getClass().  This is called from a generic with no instance to operate on.

PR #456, opened 2020-01-23
Require tests on all declared JPA entities

weiminyu:
The JPA entity classes declared in persistence.xml reflect
JPA's view of the database. The view must be compatible, though
not always consistent, with the actual schema. By defining and
using a new JpaIntegrationWithCoverageRule, this change verifies
that all JPA entities in persistence.xml are included in some
tests in the SqlIntegrationTestSuite.

Also removed one test from SqlIntegrationTestSuite since it
does not use Cloud SQL.

Lastly, global resources such as clock should be provided to
a test fixture. Having one fixture creating its own makes it
harder to use in a multi-system integration test. This change
refactored JpaTransactionManagerRule to take a user provided
clock.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/456)


mindhog:
___
*[core/src/test/java/google/registry/persistence/transaction/JpaEntityCoverage.java, line 30 at r1](https://reviewable.io/reviews/google/nomulus/456#-LzIpxAn2MxNnNwYDNhy:-LzIpxAn2MxNnNwYDNhz:byesbot) ([raw file](https://github.com/google/nomulus/blob/308210f79dacbb48692c4b251dd5d88947554a9e/core/src/test/java/google/registry/persistence/transaction/JpaEntityCoverage.java#L30)):*
> ```Java
> import org.junit.rules.ExternalResource;
> 
> /* Checks for JPA {@link Entity entities} that are declared in persistence.xml
> ```

Make this a  "/**" docstring?

mindhog:
___
*[core/src/test/java/google/registry/persistence/transaction/JpaEntityCoverage.java, line 52 at r1](https://reviewable.io/reviews/google/nomulus/456#-LzIszA81qDf9ee6YTE6:-LzIszA81qDf9ee6YTE7:b-yjc6z0) ([raw file](https://github.com/google/nomulus/blob/308210f79dacbb48692c4b251dd5d88947554a9e/core/src/test/java/google/registry/persistence/transaction/JpaEntityCoverage.java#L52)):*
> ```Java
>   private static final Set<Class> allCoveredJpaEntities = Sets.newHashSet();
>   // Map of test class name to boolean flag indicating if it tests any JPA entities.
>   private static final Map<String, Boolean> testClassRelevance = Maps.newHashMap();
> ```

Would "testsJpaEntities" be a better name for this?

mindhog:
___
*[core/src/test/java/google/registry/persistence/transaction/JpaEntityCoverage.java, line 108 at r1](https://reviewable.io/reviews/google/nomulus/456#-LzIv9P_5P8b3fXMoEz1:-LzIv9P_5P8b3fXMoEz2:bu01z63) ([raw file](https://github.com/google/nomulus/blob/308210f79dacbb48692c4b251dd5d88947554a9e/core/src/test/java/google/registry/persistence/transaction/JpaEntityCoverage.java#L108)):*
> ```Java
>                         .getEntityManager()
>                         .createQuery(
>                             String.format("SELECT e FROM %s e", entityType.getSimpleName()),
> ```

This appears to assume that the table name is the same as the class name.  That won't be the case after #438 in its current form (DomainBase -> Domain).

PR #459, opened 2020-01-23
Add a RegistryLockPostAction and tests

gbrodman:
This POST action takes as input a email-address/password combination, as well as a domain name and an action (lock or unlock). It will then verify those credentials and attempt to create a pending RegistryLock object for that domain (or modify the existing one, if an unlock is being requested). 

The action will then send an email to the email address provided that contains the link necessary to verify and apply the registry lock/unlock. if the email link isn't clicked within one hour, the action is no longer valid and cannot be applied. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/459)


gbrodman:
ben, PTAL and let me know if you have any thoughts, when you get a chance (no rush). Thanks!

PR #469, opened 2020-02-03
Auto-apply JPA converters for collection type

hstonec:
JPA's `AttributeConverter` can detect the generic type of Java class correctly at the runtime. So, this PR refactored the existing `UserType` for converting Java collections to use `AttributeConverter` instead. In that way, we don't need to add `@Type()` annotation for each collection field because `AttributeConverter` supports `autoApply`.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/469)


mindhog:
___
*[core/src/main/java/google/registry/persistence/StatusValueSetConverter.java, line 25 at r4](https://reviewable.io/reviews/google/nomulus/469#-M-QJJw1AvwCTHsqfAeR:-M-QMP3F9wJ0KDTqhYsq:bsq4glg) ([raw file](https://github.com/google/nomulus/blob/a3beebdd2b2204ffff8e844f6cae66d64d2a7144/core/src/main/java/google/registry/persistence/StatusValueSetConverter.java#L25)):*

Can't we make this better for Enums?  In another outstanding PR of yours, handling a Set of enums is as simple as a one-line inline static class as follows:

```
public static class RegistrarPocType extends EnumSetUserType<Type> {}
```

Can we make it that simple here too?
</blockquote></details>

Since there's currently only one persisted enum, I suggest that we just leave this for now and do code generation (see  https://github.com/mindhog/nomulus/tree/enum-set-ann)  if we ever need another one.

PR #476, opened 2020-02-07
Key to VKey conversion for Nameserver

mindhog:
This change illustrates the conversion of a single key in the system
(Key<HostResource> as used in the "nameservers" field of DomainBase) to a
VKey.

I don't expect that this will be satisfactory in its current form.  I'm mainly publishing this PR in order to gain feedback as to the overall approach, though it would be nice to ultimately merge some variation on it.  In principal, the approach shouldn't be extremely controversial as we've covered it in our internal design document.

So have at it :-)

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/476)

### DIFF
--- a/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
+++ b/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
@@ -52,6 +52,7 @@ import google.registry.mapreduce.inputs.NullInput;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.host.HostResource;
 import google.registry.model.server.Lock;
+import google.registry.persistence.VKey;
 import google.registry.request.Action;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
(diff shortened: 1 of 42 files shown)


core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
```diff
@@ -206,7 +207,9 @@ public final void map(@Nullable final DomainBase domain) {
       Key<HostResource> referencingHostKey = null;
       for (DnsRefreshRequest request : refreshRequests) {
         if (isActive(domain, request.lastUpdateTime())
-            && domain.getNameservers().contains(request.hostKey())) {
+            && domain
+                .getNameservers()
+                .contains(VKey.create(HostResource.class, request.hostKey()))) {
```
CydeWeys: What is VKey accomplishing for us here?  Ultimately the way this should work is that there is an array of repoIds on the domain, all of which are foreign keys to the hosts table which has the repoId as its id.  What is VKey getting us?

hstonec: I borrowed this idea when I was refactoring the `TransactionManager` to create a unified API but I made a little bit change to allow `VKey` to be able to carry both ofy key and sql key. Then, when `tm().load(VKey.create(clazz, ofyKey, sqlKey))` is invoked, `JpaTransactionManger` would use `vkey.getSqlKey()` while `DatastoreTransactionManager` can use `vkey.getOfyKey()` to get the key it needs.

PR #478, opened 2020-02-07
Add a scrap command to backfill registry locks

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/478)


gbrodman:
fyi we have about 50 locked domains right now, so this is doable

PR #490, opened 2020-02-18
Update dependency locks

CydeWeys:
The google repo is now required because the App Engine API is pulling in a
new dependency on android annotation.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/490)


CydeWeys:
Abandoning this.  It was subsumed into #488 once we figured out exactly what was going wrong.

PR #493, opened 2020-02-20
Use -t instead of main params for TLDs in nomulus count_domains command

CydeWeys:
This makes the command consistent with list_domains. I use both frequently and it
was annoying forgetting which one takes -t and which uses main parameters. Now
they both work the same way.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/493)


mindhog:
___
*[core/src/main/java/google/registry/tools/ListDomainsCommand.java, line 35 at r1](https://reviewable.io/reviews/google/nomulus/493#-M0Z1FpU0_h7iTT_iMAD:-M0Z1FpVBJQK0SLGb27m:ba49ees) ([raw file](https://github.com/google/nomulus/blob/21ef00c899f748b7962f98dc56682ff5bc32ab59/core/src/main/java/google/registry/tools/ListDomainsCommand.java#L35)):*
> ```
> "Space
> ```

Really??

PR #494, opened 2020-02-20
Create methods to administratively (un)lock domains

gbrodman:
Add a separate, simpler, one-stop method administratively locking or unlocking domains. This is used both for the Nomulus tool commands and also for relocking later.

Also refactor the utils methods to take as input a datetime instead of a clock, to make it easier to run these methods in transactions.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/494)


gbrodman:
All right, we now assert that we're in a transaction for any of the DAO methods. I added a SqlHelper similar to how we have the DatastoreHelper for test purposes. PTAL

PR #545, opened 2020-04-03
Actually run JUnit 5 tests

weiminyu:
In Gradle, JUnit 5 must be explicitly enabled with a call to
test.useJUnitPlatform().

The FilteringTest used in :core must also enable JUnit5 separately.

Fixed AppEngineRule to work with the few tests that have migrated
to JUnit5 (all in one of the two classes named PremiumListTest).

More work is needed with AppEngine before we can migrate tests
that actually use Cloud SQL.

For context, with @EnableRuleMigrationSupport, JUnit 5 runner calls
an external resource's before() and after() methods. TestRule.apply()
is not called, therefore any setup done their will be bypassed with
JUnit 5.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/545)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -646,6 +646,10 @@ artifacts {
  */
 class FilteringTest extends Test {
 
+  FilteringTest() {
+    useJUnitPlatform();
+  }
+
   private void applyTestFilter() {
     if (project.testFilter) {
       testNameIncludePatterns = project.testFilter.split(',')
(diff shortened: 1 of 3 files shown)


hstonec:
This is interesting because the `before` and `after` methods defined in  `ExternalResource` are eventually invoked in its own `apply` function. Any thoughts about that?

PR #547, opened 2020-04-06
Add JPA annotations to ContactResource and generate schema

hstonec:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/547)


mindhog:
I'm assuming this is for the draft schema?  What do you guys think of committing this on a branch?  (Or even just checking out from the existing PR branch)  This way, we don't have to iterate on changes in the prod schema.

hstonec:
> I'm assuming this is for the draft schema? What do you guys think of committing this on a branch? (Or even just checking out from the existing PR branch) This way, we don't have to iterate on changes in the prod schema.

Yes. I think if we commit this to another branch, it needs a lot of effort to keep that branch up to date and compatible with the master branch. Though we may need to come back to make change to the schema, we don't have any production data stored in it so I think it is ok to check this in. Thoughts?

PR #551, opened 2020-04-09
Add a listener to invoke entity callbacks

hstonec:
This PR added a listener class to invoke entity callbacks in cases where Hibernate doesn't invoke the callback as expected.

JPA defines a few annotations, e.g. `@PostLoad`, that we can use for the application to react to certain events that occur inside the persistence mechanism. However, Hibernate only supports a few basic use cases, e.g. defining a `@PostLoad` method directly in an `Entity` class or in an `Embeddable` class. If the annotated method is defined in an `Embeddable` class that is a property of another `Embeddable` class, or it is defined in a parent class of the `Embeddable` class, Hibernate doesn't invoke it.

See Also [HHH-13316](https://hibernate.atlassian.net/browse/HHH-13316)

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/551)

### DIFF
--- a/core/src/main/java/google/registry/persistence/EntityCallbacksListener.java
+++ b/core/src/main/java/google/registry/persistence/EntityCallbacksListener.java
@@ -43,6 +43,15 @@ import javax.persistence.PreUpdate;
  * defined in an {@link Embeddable} class that is a property of another {@link Embeddable} class, or
  * it is defined in a parent class of the {@link Embeddable} class, Hibernate doesn't invoke it.
  *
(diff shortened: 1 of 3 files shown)


mindhog:
Is there some place where you intend to use this?

hstonec:
> Is there some place where you intend to use this?

Yes, right now, it is used for the [postLoad method Address](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/eppcommon/Address.java).

PR #553, opened 2020-04-10
Add verification that domain labels aren't multi-level domains

gbrodman:
In addition, I did a bit of test refactoring because previously, the
CreateOrUpdateReserveListCommandTestCase test cases weren't actually
testing the proper things -- they were failing with
IllegalArgumentExceptions, but not the right ones.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/553)


gbrodman:
This is failing because of the issue fixed in https://team-review.git.corp.google.com/c/domain-registry-eng/nomulus-internal/+/715518, so it should be good to go otherwise

PR #566, opened 2020-04-23
Improve VKeyConverter

hstonec:
This PR added an annotation processor to generate `AttributeConoverter` Java class for the entity class annotated with `WithStringVKey` or `WithLongVKey`. The generated class will be in the same package as the entity class, and its name will be the `VKeyConverter_` concatenated with the type name of the VKey(or the value of `classNameSuffix` if set).

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/566)


mindhog:
___
*[core/src/main/java/google/registry/persistence/LongVKey.java, line 21 at r2](https://reviewable.io/reviews/google/nomulus/566#-M65TIVR2IHs1PdqSBqG:-M65TIVSCt8MG_9VJKd5:b-rdee9k) ([raw file](https://github.com/google/nomulus/blob/0543cde98ae6743168ec9ccd4dbdc84036281b4d/core/src/main/java/google/registry/persistence/LongVKey.java#L21)):*
> ```Java
> import javax.persistence.AttributeConverter;
> 
> /** Annotation for {@link VKey} that is stored as a long in the database. */
> ```

I had envisioned this being applied to the class that we're generating a VKeyConverter for, for example:

@WithStringVKey
class ContactResource ...

Did you try that?

If we have to annotate every VKey field it kind of defeats the purpose of the annotation, which is to reduce the amount of boilerplate.

mindhog:
___
*[processor/src/main/java/google/registry/processors/AbstractVKeyProcessor.java, line 130 at r2](https://reviewable.io/reviews/google/nomulus/566#-M65WWf39BrNZaaFcBrp:-M65WWf4Co_fuFS_G80r:bmco0i6) ([raw file](https://github.com/google/nomulus/blob/0543cde98ae6743168ec9ccd4dbdc84036281b4d/processor/src/main/java/google/registry/processors/AbstractVKeyProcessor.java#L130)):*
> ```Java
>       String converterClassName,
>       TypeMirror vKeyTypeMirror,
>       TypeMirror entityTypeMirror) {
> ```

This seems to be generating the converter - can we instead generate the specialization of the converter that just defines the getAttributeClass() method?

mindhog:
___
*[processor/src/main/java/google/registry/processors/AbstractVKeyProcessor.java, line 25 at r2](https://reviewable.io/reviews/google/nomulus/566#-M65YnCa7Z8p6JbgbtGa:-M65YnCa7Z8p6JbgbtGb:bkc8b96) ([raw file](https://github.com/google/nomulus/blob/0543cde98ae6743168ec9ccd4dbdc84036281b4d/processor/src/main/java/google/registry/processors/AbstractVKeyProcessor.java#L25)):*
> ```Java
> import com.squareup.javapoet.ParameterizedTypeName;
> import com.squareup.javapoet.TypeName;
> import com.squareup.javapoet.TypeSpec;
> ```

I know javapoet is generally the prescribed way to do this, but it seems like for very simple cases it just complicates the code.  Is there a reason you prefer it here?

mindhog:
___
*[processor/src/main/java/google/registry/processors/AbstractVKeyProcessor.java, line 130 at r2](https://reviewable.io/reviews/google/nomulus/566#-M65WWf39BrNZaaFcBrp:-M65oqtv0Re9StnLFB5U:b-nucsz5) ([raw file](https://github.com/google/nomulus/blob/0543cde98ae6743168ec9ccd4dbdc84036281b4d/processor/src/main/java/google/registry/processors/AbstractVKeyProcessor.java#L130)):*

I think you mean to make the generated converter extend the existing `VKeyConverter`? Hmm, my original idea is to not make the generated code depend on another piece of code.
</blockquote></details>

So, generally speaking, I've found that generated code is more difficult to deal with than normal code.  For this reason, it's often useful to generate only the minimal set of code that needs to be parameterized and move everything else into non-generated, common code.
Why is it that you wish to avoid dependencies?

PR #570, opened 2020-04-29
Implement DatastoreEntity and SqlEntity on more classes

gbrodman:
For classes that aren't going to transition to SQL, they should just
return an empty list of SqlEntities. When reading these in from the
commit log manifests, we just won't persist anything to SQL.

By having all Datastore entity classes implement DatastoreEntity, we can
avoid potential bugs where we forget to transition some entity to SQL,
or we forget to have the capability to read back from the commit logs.

Note: the EntityTest is still @Ignore'd because there are many SQL and
Datastore classes left -- ones that we are still in the process of
converting or adding, or ones that require more complicated transitions.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/570)


gbrodman:
more information on which classes can be / should be / shouldn't be converted in https://buganizer.corp.google.com/issues/155418579

PR #577, opened 2020-05-05
Add columns for TransferData in Domain and Contact

hstonec:
Altered table `Contact` and `Domain` to add columns for `TransferData`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/577)


hstonec:
___
*[db/src/main/resources/sql/flyway/V27__add_columns_for_transfer_data.sql, line 19 at r4](https://reviewable.io/reviews/google/nomulus/577#-M8Vg1ep8U6eVsyXDoRq:-M8w8RrK5vYSp3EslQsw:b1dwur1) ([raw file](https://github.com/google/nomulus/blob/533ad7f1913ec52356920897ced08a36ba7e51cb/db/src/main/resources/sql/flyway/V27__add_columns_for_transfer_data.sql#L19)):*

> ```
>     add column transfer_billing_cancellation_id int8,
>     add column transfer_billing_event_id int8,
>     add column transfer_billing_recurrence_id int8,
> ```
So, I ended up with adding these 3 columns and `transfer_autorenew_poll_message_id` in `Contact` which should not include them. The reason is `TransferData` object is embedded in both `DomainBase` and `ContactResource`, and when we add a `TransferData transferData` field into the entity class, each field defined in the embeddable class will become a column of the entity, there is no way to exclude certain column.

The recommended solution would be have an abstract `TransferData` class containing the common fields, and then have a `DomainTransferData` and `ContactTransferData` extending it and contain their own fields. Obviously, this will change a lot of things and we need to verify if it is feasible especially when we are migrating the database.

So, my proposal is:

1. We continue with the single `TransferData` class, and add all transfer related columns in both `Domain` and `Contact` table. This would work in any condition because we just add some unused columns in `Contact` and our code controls that we won't write anything in them. (Also, this is actually what we do in Datastore).

2. Finish the migration to Cloud SQL, and deprecate the Datastore solely fields and annotations in `TransferData`.

3. Refactor the `TransferData` class to have two subclasses `DomainTransferData` and `ContactTransferData`.

4. Delete unused fields in `Contact` table.

If we feel that we still want to have the final class version before migration, we can still at least check this PR in to unblock the other work, then start investigating the feasibility to refactor the classes to make them work for both Datastore and Cloud SQL. Any thoughts? (I can put this in a one-pager for further discussion)
</blockquote></details>

@cydeweys  Not sure if you noticed this but this answered your question why there are some fields in `Contact` but should not exist.

PR #578, opened 2020-05-06
Do not enqueue update snapshot task if import fails

jianglai:
If the import from Datastore to BigQuery fails, there is no point
enqueuing a job to update the snapshot view.

Also when there's an error updating the snapshot view, log it at severe
level. The HTTP exception thrown is logged at info and triggers a retry
implicitly. I'm not sure if we want this behavior though. Do we want to
retry upon snapshot updating failures? Unless the failurs are transient,
retrying doesn't help. In our case the failure (End of time out of range
in Standard SQL) is not transient.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/578)

### DIFF
--- a/core/src/main/java/google/registry/export/BigqueryPollJobAction.java
+++ b/core/src/main/java/google/registry/export/BigqueryPollJobAction.java
@@ -73,8 +73,10 @@ public class BigqueryPollJobAction implements Runnable {
 
   @Override
   public void run() {
-    checkJobOutcome();  // Throws a NotModifiedException if the job hasn't completed.
-    if (payload == null || payload.length == 0) {
+    boolean jobOutcome =
+        checkJobOutcome(); // Throws a NotModifiedException if the job hasn't completed.
+    // If the job failed, do not enqueue the next step.
+    if (!jobOutcome || payload == null || payload.length == 0) {
       return;
     }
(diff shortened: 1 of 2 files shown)


jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/578)_ status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)
> 
> _[core/src/main/java/google/registry/export/UpdateSnapshotViewAction.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/578#-M6ppAEuAvhVhd7LCIJm:-M79im6MESqb_K2ye70J:bql8ucr) ([raw file](https://github.com/google/nomulus/blob/1b187827b44180d8b955ec2f5f294f49df87962a/core/src/main/java/google/registry/export/UpdateSnapshotViewAction.java#L105)):_
> 
> I think the right thing to do is to add a logging level field to our `HttpException` base class that defaults to `INFO` in the base implementation but that is set to `SEVERE` for `InternalServerErrorException`. I think for now that's the `HttpException` subclass that merits logging at a higher level, but we can expand that in the future as necessary.
> 
> I just searched the logs for these and that's not a noisy exception at all. Anytime it's thrown I think we would legitimately want to know about it.

Sounds good. Updated in #585.

PR #584, opened 2020-05-12
Add a HostHistory and abstract EppHistory table

gbrodman:
We will have separate history tables for domain, contact, and host
histories. Each will subclass EppHistory, which contains common fields
about the change itself.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/584)


gbrodman:
The fields in EppHistory are mostly taken from HistoryEntry, and the additional fields in HostHistory are taken from the HostResource object 

Thoughts on this approach? 

In particular, let me know if you can think of better names :)

PR #608, opened 2020-06-04
Use the correct text VKey for HostResource's superordinateDomain

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/608)


gbrodman:
This should be all set for review now, thanks

PR #611, opened 2020-06-05
Only use asymmetric VKeys for EPP resources

gbrodman:
Given that we currently have no way of reconstituting a symmetric key
from the asymmetric key (or at least, we don't have a 100% reliable way
of doing so) it's best to keep keys as asymmetrical, referring to the
correct database. That way, we don't get situations where we cannot
    compare equality of two keys due to one being asymmetrical and one being
    symmetrical.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/611)


gbrodman:
whoops, accidentally pushed a commit to this branch that was meant for another PR. Fixed that.

PR #622, opened 2020-06-11
Log client certificate type and length

jianglai:
It appears that most client certificates are RSA certs, but we should
make sure that is indeed the case. Print out the strength of the cert
if it is RSA.

Also adds supports for TLS 1.3 and print out the supported cipher suites.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/622)

### DIFF
--- a/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
+++ b/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
@@ -75,6 +75,7 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
   // change when the artifacts on GCS changes.
   private final Supplier<PrivateKey> privateKeySupplier;
   private final Supplier<ImmutableList<X509Certificate>> certificatesSupplier;
+  private final ImmutableList<String> supportedSslVersions;
 
   public SslServerInitializer(
       boolean requireClientCert,
@@ -91,6 +92,12 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
     this.sslProvider = sslProvider;
     this.privateKeySupplier = privateKeySupplier;
     this.certificatesSupplier = certificatesSupplier;
+    this.supportedSslVersions =
+        sslProvider == SslProvider.OPENSSL
+            ? ImmutableList.of("TLSv1.3", "TLSv1.2", "TLSv1.1", "TLSv1")
+            // JDK support for TLS 1.3 won't be available until 2020-07-14 at the earliest.
+            // See: https://java.com/en/jre-jdk-cryptoroadmap.html
+            : ImmutableList.of("TLSv1.2", "TLSv1.1", "TLSv1");
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


jianglai:
> > Reviewed 1 of 1 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/622)_ status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)
> 
> _[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 141 at r1](https://reviewable.io/reviews/google/nomulus/622#-M9YVYCd8ya9SPN8fSOC:-M9YVYCd8ya9SPN8fSOD:bql6uww) ([raw file](https://github.com/google/nomulus/blob/f8403986f76d688536d19dc6d42a71f5c805169d/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L141)):_
> 
> > ```java
> >                           clientCertificate.getNotAfter(),
> >                           clientPublicKey.getClass().getName(),
> >                           clientCertificateLength);
> > ```
> 
> This is going to log a certificate length of zero for a non-rsa key, which is confusing. I suggest replacing clientCertificateLength with (clientCertificateLength > 0) ? Integer.toString(clientCertificateLength) : "unknown"

I intentionally left it to be 0 because it will make it easier to process in BQ to cast the string to int. I think it should be fine as the logs here are temporary and mostly for us to get study the characteristics of the connection, so we should make them easy to process downstream, as opposed to easy to understand for humans.

I've added a comment to that effect.

mindhog:
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 141 at r1](https://reviewable.io/reviews/google/nomulus/622#-M9YVYCd8ya9SPN8fSOC:-M9Yo38QBdlDKugYmFQs:b-o765bh) ([raw file](https://github.com/google/nomulus/blob/f8403986f76d688536d19dc6d42a71f5c805169d/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L141)):*

I intentionally left it to be 0 because it will make it easier to process in BQ to cast the string to int. I think it should be fine as the logs here are temporary and mostly for us to get study the characteristics of the connection, so we should make them easy to process downstream, as opposed to easy to understand for humans.

I've added a comment to that effect.  
</blockquote></details>

In that case, please at least use a value of -1 (which is clearly not a size) instead of 0 (which suggests an empty key)

jianglai:
> _[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 141 at r1](https://reviewable.io/reviews/google/nomulus/622#-M9YVYCd8ya9SPN8fSOC:-M9Yo38QBdlDKugYmFQs:b-o765bh) ([raw file](https://github.com/google/nomulus/blob/f8403986f76d688536d19dc6d42a71f5c805169d/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L141)):_
> 
> In that case, please at least use a value of -1 (which is clearly not a size) instead of 0 (which suggests an empty key)

Done.

PR #625, opened 2020-06-11
Allow class-specific creation of symmetric VKeys

mindhog:
When translating from a datastore Key to a VKey, see if the "kind" class
contains a createVKey(com.googlecode.objectify.Key) static method and if it
does, use it to construct a symmetric VKey instead of simply creating an
objectify-sided asymmetric VKey.

As a test case for this, implement the createVKey() static function for
DomainBase.  Also, create unit tests for VKeyTranslatorFactory, which
continues to house the functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/625)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -617,8 +617,11 @@ public class DomainBase extends EppResource
 
   @Override
   public VKey<DomainBase> createVKey() {
-    // TODO(mmuller): create symmetric keys if we can ever reload both sides.
-    return VKey.createOfy(DomainBase.class, Key.create(this));
+    return VKey.create(DomainBase.class, getRepoId(), Key.create(this));
+  }
+
+  public static VKey<DomainBase> createVKey(Key key) {
+    return VKey.create(DomainBase.class, key.getName(), key);
   }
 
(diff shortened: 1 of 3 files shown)


gbrodman:
Would we want to add an example (if one exists yet) of how this might work with a class whose Datastore key has a parent?

mindhog:
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 624 at r1](https://reviewable.io/reviews/google/nomulus/625#-M9cx4gdANUuar9oVWwV:-M9saJgpD2M2bDFoi9c2:b493wo5) ([raw file](https://github.com/google/nomulus/blob/439268fb95f234b0f73a98e1ba64eb7026d7c43d/core/src/main/java/google/registry/model/domain/DomainBase.java#L624)):*

I think we can go ahead with this and make the improvement later. BTW, which case in your mind that this doesn't work?
</blockquote></details>

VKey&lt;HistoryEntry&gt;.  They have demonstrably non-unique ids.  I'm not sure how we're going to deal with these, that's what I want to talk to Gus about.

gbrodman:
So actually I just made a change in my History PR that will allow us to use the same single sequence for all of the HistoryEntry subclass IDs. That way, there shouldn't be any collisions.

PR #631, opened 2020-06-16
Remove 'fullyQualified' from host and domain names

gbrodman:
We don't actually enforce that these are properly fully-qualified
(there's no dot at the end) and we specifically use the term "label
name" when talking about labels.

Note: this doesn't convert FQDN -> DN (et al) in at least two types of cases:
1. When the term is part of the XML schema
2. When the term is used by some external system, e.g. SafeBrowsing API

Note as well that this does not change the Datastore schema, to avoid being destructive. This only changes the SQL schema, which will eventually be authoritative. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/631)


core/src/main/java/google/registry/model/domain/DomainBase.java
```diff
@@ -137,6 +137,7 @@
    *
    * @invariant fullyQualifiedDomainName == fullyQualifiedDomainName.toLowerCase(Locale.ENGLISH)
    */
+  @Column(name = "domainName")
   @Index String fullyQualifiedDomainName;
```
CydeWeys, line 141: Add a TODO here to rename this once the Cloud SQL migration is complete.

gbrodman: Done (also in HostBase)

PR #632, opened 2020-06-16
Show price of reserved domains when using matching allocation token

CydeWeys:
When the registrar passes the fee extension, this shows the price of the domain
on a check command for reserved domains if the provided allocation token is a
match. Of course, the price is already always displayed on non-reserved names
(regardless of whether the specific provided token is a match or not).

This affects domain checks only; the price is already always displayed on domain
creates because you already by definition have access to register the domain in
question.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/632)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -163,7 +163,8 @@ public final class DomainCheckFlow implements Flow {
                     clientId,
                     now));
 
-    ImmutableList.Builder<DomainCheck> checks = new ImmutableList.Builder<>();
+    ImmutableList.Builder<DomainCheck> checksBuilder = new ImmutableList.Builder<>();
+    ImmutableSet.Builder<String> availableDomains = new ImmutableSet.Builder<>();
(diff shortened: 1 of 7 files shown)


core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
```diff
@@ -163,7 +163,8 @@ public EppResponse run() throws EppException {
                     clientId,
                     now));
 
-    ImmutableList.Builder<DomainCheck> checks = new ImmutableList.Builder<>();
+    ImmutableList.Builder<DomainCheck> checksBuilder = new ImmutableList.Builder<>();
+    ImmutableSet.Builder<String> availableDomains = new ImmutableSet.Builder<>();
```
CydeWeys, line 167: I tried several different approaches here before I settled on this one.  A Map of domain names to Checks doesn't work because there can be duplicate domains in the check.  A Multimap doesn't work because it no longer preserves strict insertion order (multiple values for the same key end up collated next to each other), while the checks themselves must be in strict order because they aren't associated with a domain name except by index.

So this ends up being the simplest thing that works.

PR #587, opened 2020-05-14
Add HostBase and HostHistory classes

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/587)

PR #634, opened 2020-06-17
Refactor ContactResource into ContactBase and create ContactHistory

gbrodman:
This is similar to #587, but with contacts instead of hosts. 

This also includes a couple cleanups for HostHistoryTest and RegistryLockDaoTest, just making code more proper (we shouldn't be referencing constant revision IDs when using a sequence that is used by multiple classes, and RLDT can extend EntityTest)

Note as well that we set ContactHistory to use the same revision ID sequence as HostHistory. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/634)


gbrodman:
The last couple commits just now didn't really change much of anything, just cleaning up the commit history. 

This is basically the same as what we did for HostHistory, with the exception of refactoring `cloneProjectedAtTime` into a separate, static method. This doesn't matter too much here, but in other places (DomainBase) this allows us to avoid duplicating a ton of code. Essentially, we'd like to have the logic for `cloneProjectedAtTime` to exist in a single place, but we'd like ContactBase.cloneProjectedAtime to return a ContactBase, and ContactResource.cloneProjectedAtTime to return a ContactResource. This is one way of doing that without duplicating logic.

PR #657, opened 2020-06-29
Disambiguate injected Cloud SQL parameter names

gbrodman:
This allows us to also inject the BeamJpaModule into RegistryTool, which
allows us to use the SocketJpaTransactionManager in Beam pipelines.

Some side effects of this include:
- duplication of KMS connections -- one standard, one Beam
- duplication of the creation of the partial Hibernate SQL configs
- removal of ambiguity between credentialFileName, credentialFilename,
and credentialFilePath -- we now use the latter.
- Performing the credential null check when instantiating the SQL
connection rather than when instantiating the module object. See the code
comments for more details on this.

I verified that this compiles and the tests run successfully when
injecting a @SocketFactoryJpaTm into a Beam pipeline.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/657)

### DIFF
--- a/core/src/main/java/google/registry/beam/initsql/BeamJpaModule.java
+++ b/core/src/main/java/google/registry/beam/initsql/BeamJpaModule.java
@@ -156,6 +156,12 @@ public class BeamJpaModule {
     return "nomulus-tool-keyring";
   }
 
+  @Provides
+  @Config("beamHibernateHikariMaximumPoolSize")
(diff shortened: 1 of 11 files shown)


gbrodman:
Note: the failing tests are seemingly unrelated to this issue (flaky WebDriver test(?) and javadoc generation) because it succeeds locally

PR #666, opened 2020-07-07
Upgrade flow test classes to JUnit 5

CydeWeys:
Most of the diffs are visibility changes.

Also deletes ShardableTestCase, which was only necessary because of Blaze (and
possible Bazel) limitations.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/666)


gbrodman:
oooooh, Satanic PR

CydeWeys:
I dunno, I think it's kind of an anti-pattern to use inheritance for something like this.  I think it's clearer to simply include these rules as necessary rather than inherit from a base test class you weren't already using just for the purposes of including these rules (while also hiding them).  Favor composition over inheritance.

You also don't always need both of these, and shouldn't be using both when you don't actually need both, so now you'd have a separate base class for InjectionTest and AppEngineTest, and then you couldn't combine them because Java doesn't allow multiple inheritance.  In another language this *might* work better, but not in Java.  We really do need free composition for these, which is not possible using inheritance.

PR #668, opened 2020-07-07
Refactor DomainBase into DomainContent and create DomainHistory

gbrodman:
This is similar to #587 and #634, but for domains.

One caveat is that we refactor some of the Domain* instance methods to
be static so that they can be called either on DomainBase or
DomainContent, returning the appropriate type each time.

Note that we set DomainHistory to use the same revision ID sequence as
HostHistory and ContactHistory.

In addition, we refactor the tests to the History objects a bit to
reduce duplicate code and because we cannot guarantee yet that the
SQL-stored VKeys are symmetrical -- the ofy keys are not persisted at
the moment.

In addition, rename the DomainHost table to the default `Domain_nsHosts` so that it automatically creates two separate nsHosts tables for us -- one foreign-keyed on the domain repo ID, and one foreign-keyed on the history revision ID 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/668)


gbrodman:
Sorry about all the pushes; I had to fix up some tests and there were a decent bit of merge conflicts (especially due to the file moving, since Github doesn't handle that as nicely as Critique does). 

PTAL, and thanks. Let me know if you have any questions on where to start; the main addition is the Transient annotation on nsHosts with the getter/setter method added in the subclasses for Hibernate's sake.

PR #674, opened 2020-07-09
Add domainRepoId to Subdomain class

leginachen:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/674)

In order for the ThreatMatch objects to be able to reference the foreign key of the Domain table, we need to have a domainRepoId field in the Subdomain Java class. This PR enables the sql code in subdomains.sql to pull those keys from that table and have them be retrievable from Subdomain.java.

### DIFF
--- a/core/src/main/java/google/registry/beam/spec11/Subdomain.java
+++ b/core/src/main/java/google/registry/beam/spec11/Subdomain.java
@@ -36,12 +36,14 @@ import org.apache.beam.sdk.io.gcp.bigquery.SchemaAndRecord;
 public abstract class Subdomain implements Serializable {
 
   private static final ImmutableList<String> FIELD_NAMES =
-      ImmutableList.of("fullyQualifiedDomainName", "registrarClientId", "registrarEmailAddress");
+      ImmutableList.of("domainName", "domainRepoId", "clientId", "registrarEmailAddress");
 
   /** Returns the fully qualified domain name. */
-  abstract String fullyQualifiedDomainName();
+  abstract String domainName();
+  /** Returns the domain repo ID (the primary key of the domain table). */
+  abstract String domainRepoId();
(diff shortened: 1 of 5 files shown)


core/src/main/java/google/registry/beam/spec11/Subdomain.java
```diff
@@ -36,12 +36,14 @@
 public abstract class Subdomain implements Serializable {
 
   private static final ImmutableList<String> FIELD_NAMES =
-      ImmutableList.of("fullyQualifiedDomainName", "registrarClientId", "registrarEmailAddress");
+      ImmutableList.of("domainName", "domainRepoId", "clientId", "registrarEmailAddress");
```
gbrodman, line 39: let's use "registrarId" instead of "clientId"

core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
```diff
@@ -81,21 +83,23 @@ public static void initializePipelineOptions() {
...
-        retrier
-    );
+    JpaTransactionManager jpaTm = jpaTm();
+    spec11Pipeline =
+        new Spec11Pipeline(
+            "test-project",
+            beamTempFolder.getAbsolutePath() + "/staging",
+            beamTempFolder.getAbsolutePath() + "/templates/invoicing",
+            tempFolder.getRoot().getAbsolutePath(),
+            jpaTm,
```
gbrodman, line 100: ah, we don't want <this> change in the file yet unfortunately

gbrodman:
nit on the title of the commit: it's ideal to have the first word of the verb phrase (in this case, "change") be as specific as possible. So here it could be something like "Add [repoId to Subdomain]" or something like that that describes what's going on

PR #678, opened 2020-07-10
Persist new threat matches

leginachen:
Creates Spec11ThreatMatch objects using Subdomains and ThreatMatches, and persists them using a JpaTransactionManager. Alter the existing test in Spec11PipelineTest to test for the persistence of them. This is so we can begin storing the new data from the daily runs using the SafeBrowsing API into our new SQL table.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/678)


gbrodman:
I think there are some conflicts; it seems like this isn't based on the most recent version of master (for instance, it shows the Subdomain changes that we already merged in)

PR #686, opened 2020-07-14
Persist ThreatMatches into the Spec11ThreatMatch table

leginachen:
Create Spec11ThreatMatches from Subdomains and ThreatMatches in the Spec11Pipeline, persist them, and test their persistence. This is so we can begin populating the database with the new ThreatMatches in addition to the usual JSON files, for future data analysis and advanced reporting.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/686)


leginachen:
PTAL @gbrodman @sarahcaseybot

PR #712, opened 2020-07-24
Use rearranged sql credentials in flyway task

weiminyu:
Let the flyway tasks use the sql credential files set up for BEAM
pipelines. 

Credential files have been created for each environment in GCS
at gs://${project}-beam/cloudsql/admin_credential.enc. They are
meant for the BEAM pipeline runner accounts in each project.

Alpha and crash accounts use the 'nomulus-tools-key' in their own project to
decrypt the credential file.

Sandbox and production accounts use the 'nomulus-tools-key' in
domain-registry-dev to decrypt the credential file.

Previously all flyway tasks use the 'nomulus-tools-key' in
domain-registry-dev, which was meant for the Cloud Build tasks.
It is a little cleaner to limit this key to sandbox and production.

Note that this setup is temporary. It will become obsolete once
we migrate to Cloud Secret Manager for secret storage.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/712)

### DIFF
--- a/db/build.gradle
+++ b/db/build.gradle
@@ -25,16 +25,11 @@ ext {
   Set allDbEnv =
       [ 'alpha', 'crash' ].plus(restrictedDbEnv).asUnmodifiable()
 
-  def appNameProperty = 'appName'
   def dbServerProperty = 'dbServer'
   def dbNameProperty = 'dbName'
 
   def dbServer = findProperty(dbServerProperty).toString().toLowerCase()
   def dbName = findProperty(dbNameProperty)
-  def appName = "domain-registry"
-  if (hasProperty(appNameProperty)) {
-    appName = findProperty(appNameProperty)
-  }
 
   isCloudSql = {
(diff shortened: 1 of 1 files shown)


weiminyu:
___
*[db/build.gradle, line 34 at r1](https://reviewable.io/reviews/google/nomulus/712#-MD-kxjQ3gCpXIil0pS_:-MD1JGKh-1gJMQfDmuRh:bwe97b8) ([raw file](https://github.com/google/nomulus/blob/d079897807a207bd68428f2ec42fb0f0808c6441/db/build.gradle#L34)):*

You don't actually need to do hasProperty() on it, it will always be set: https://github.com/google/nomulus/blob/master/build.gradle#L123
</blockquote></details>

This allows property override from command line, useful when working on the public repo alone. The -P arguments will not overrid
rootProject properties.

weiminyu:
No. We will keep BEAM and GCB credentials separate.

PR #716, opened 2020-07-25
Set up deployment of the Spec11 pipeline with JPA TM

gbrodman:
There are a few things going on here. 

1. We tell the Beam JPA module how to handle gs:// paths
2. We rename the credentialFilePath to make it clear that this should contain the SQL access info (as opposed to the JSON credential file that can be used for gcloud access)
3. Allow for configuration of the project that stores the keyring necessary to decrypt the SQL access info file
4. Refactor the DeploySpec11Pipeline command -- we'll need to create the SerializableSupplier<JpaTransactionManager> when we use the transaction manager in the Spec11Pipeline. As a result, we'll need to have the ability to create it there. 
5. Use an AppEngineEnvironment whenever we deal with ImmutableObject creation

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/716)


gbrodman:
Note: this doesn't actually inject the JPA transaction manager supplier into the pipeline yet but it sets up everything else

PR #723, opened 2020-07-28
Persist ThreatMatches into Spec11ThreatMatch

leginachen:
Build Spec11ThreatMatches from Subdomains and ThreatMatches in the Spec11Pipeline, persist them, and test their persistence. This is done so that we can begin populating the database with the new ThreatMatches in addition to the usual JSON files, for future data analysis and advanced reporting.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/723)


leginachen:
@sarahcaseybot  PTAL

PR #730, opened 2020-07-30
Remove empty <p> tag from generated javadoc

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/730)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -404,6 +404,7 @@ task jaxbToJava {
         }
       }
     }
+    execInBash('find . -name *.java -exec sed -i /\\<p\\>\\$/d {} +', outputDir)
   }
 }
 


jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/730)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)
> 
> _[core/build.gradle, line 407 at r1](https://reviewable.io/reviews/google/nomulus/730#-MDVe0xW4Y9A9WZbupCa:-MDVe0xX-1eWcGfKxr0L:b-jiqj5v) ([raw file](https://github.com/google/nomulus/blob/53a1db99555fcf7a5f6d82fc1aa6169bc798c68b/core/build.gradle#L407)):_
> 
> > ```groovy-gradle
> >       }
> >     }
> >     execInBash('find . -name *.java -exec sed -i /\\<p\\>\\$/d {} +', outputDir)
> > ```
> 
> Instead of running this against the current directory ("."), you want to run it against "build" where the generated files live. The in-place sed command will modify the timestamp for all files that it runs against, regardless of whether or not the file has actually been modified. This creates problems for some programs that rely on the timestamp (like my editor! ;-)
> 
> Also note that this pattern (<p>$) exists for at least one of our source files where we would not want to remove it, namely AppEngineInternalAuthenticationMechanism.java. I believe that this should be ok as long as we're running against the "build" directory, though.

I'm a bit confused. The `outputDir` is `${generatedDir}/google/registry/xjc`. It's only the jaxb compilation process that introduces the empty `<p>` tags, and all the problematic generated files are within that folder.

PR #735, opened 2020-08-01
Delete screenshots folder in fragile test first

jianglai:
It is not clear to me why this folder is deleted in standardTest first,
presumably this is done when the webdriver tests were part of
standardTest?

Also not clear to me is why screenshots are only deleted in presubmits
but not locally. At least on my workstation standardTest runs after
fragileTest, just like in kokoro, so it should have deleted the screenshots
folder, but it didn't.

Regardless, the correct thing to do is to delete this folder first in
fragileTest so that it doesn't interfere with the results.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/735)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -705,6 +705,10 @@ task fragileTest(type: FilteringTest) {
 
   // Run every test class in a freshly started process.
   forkEvery 1
+
+  doFirst {
+    new File(screenshotsDir).deleteDir()
+  }
 }
 
 task outcastTest(type: FilteringTest) {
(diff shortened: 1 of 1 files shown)


jianglai:
> Why are we deleting screenshots at all? Doesn't this make debugging things when something goes wrong harder?
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/735#-:-MDoqzEP9sb4K-v3aNuZ:b5urk6c)_ status: 0 of 1 files reviewed, all discussions resolved (waiting on @CydeWeys)

It's deleting the folder **before** the test runs, so presumably it is easier to find the last ones after the test runs.

PR #738, opened 2020-08-03
Replay Cloud SQL transactions against datastore

mindhog:
Implement the ReplicateToDatastore cron job that will apply all Cloud SQL
transactions to the datastore.  The last transaction id is stored in a
LastSqlTransaction entity in datastore.

Note that this will not be activate in production until a) the cron
configuration is updated and b) the cloudSql.replicateTransactions flag is set
to true in the nomulus config file.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/738)

### DIFF
--- a/core/src/main/java/google/registry/model/EntityClasses.java
+++ b/core/src/main/java/google/registry/model/EntityClasses.java
@@ -52,6 +52,7 @@ import google.registry.model.tmch.ClaimsListShard;
 import google.registry.model.tmch.ClaimsListShard.ClaimsListRevision;
 import google.registry.model.tmch.ClaimsListShard.ClaimsListSingleton;
 import google.registry.model.tmch.TmchCrl;
+import google.registry.schema.replay.LastSqlTransaction;
 
 /** Sets of classes of the Objectify-registered entities in use throughout the model. */
 public final class EntityClasses {
@@ -90,6 +91,7 @@ public final class EntityClasses {
           HostResource.class,
           KmsSecret.class,
           KmsSecretRevision.class,
+          LastSqlTransaction.class,
           Lock.class,
           PollMessage.class,
           PollMessage.Autorenew.class,
--- a/core/src/main/java/google/registry/model/ofy/Ofy.java
+++ b/core/src/main/java/google/registry/model/ofy/Ofy.java
@@ -116,7 +116,7 @@ public class Ofy {
     return ofy().getTransaction() != null;
   }
 
-  void assertInTransaction() {
+  public void assertInTransaction() {
     checkState(inTransaction(), "Must be called in a transaction");
   }
 
--- a/core/src/main/java/google/registry/persistence/transaction/Transaction.java
+++ b/core/src/main/java/google/registry/persistence/transaction/Transaction.java
@@ -86,7 +86,7 @@ public class Transaction extends ImmutableObject implements Buildable {
     }
   }
 
-  static Transaction deserialize(byte[] serializedTransaction) throws IOException {
+  public static Transaction deserialize(byte[] serializedTransaction) throws IOException {
     ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(serializedTransaction));
 
     // Verify that the data is what we expect.
--- a/core/src/main/java/google/registry/persistence/transaction/TransactionEntity.java
+++ b/core/src/main/java/google/registry/persistence/transaction/TransactionEntity.java
@@ -34,9 +34,9 @@ public class TransactionEntity implements SqlEntity {
 
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
-  long id;
+  private long id;
 
-  byte[] contents;
+  private byte[] contents;
 
   TransactionEntity() {}
 
@@ -47,5 +47,13 @@ public class TransactionEntity implements SqlEntity {
   @Override
   public Optional<DatastoreEntity> toDatastoreEntity() {
     return Optional.empty(); // Not persisted in Datastore per se
+  }
+
+  public long getId() {
+    return id;
+  }
+
+  public byte[] getContents() {
+    return contents;
   }
 }
--- a/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
+++ b/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
@@ -22,9 +22,7 @@ import com.googlecode.objectify.annotation.Entity;
 import com.googlecode.objectify.annotation.Id;
 import google.registry.model.ImmutableObject;
 
-/**
- * Datastore entity to keep track of the last SQL transaction imported into the datastore.
- */
+/** Datastore entity to keep track of the last SQL transaction imported into the datastore. */
 @Entity
 public class LastSqlTransaction extends ImmutableObject {
 
@@ -37,17 +35,19 @@ public class LastSqlTransaction extends ImmutableObject {
 
   private long transactionId;
 
-  LastSqlTransaction() {
-    transactionId = -1;
+  LastSqlTransaction() {}
+
+  private LastSqlTransaction(long newTransactionId) {
+    transactionId = newTransactionId;
   }
 
-  void setTransactionId(long transactionId) {
+  LastSqlTransaction withNewTransactionId(long transactionId) {
     checkArgument(
         transactionId > this.transactionId,
         "New transaction id (%s) must be greater than the current transaction id (%s)",
         transactionId,
         this.transactionId);
-    this.transactionId = transactionId;
+    return new LastSqlTransaction(transactionId);
   }
 
   long getTransactionId() {
@@ -66,11 +66,12 @@ public class LastSqlTransaction extends ImmutableObject {
     return result == null ? new LastSqlTransaction() : result;
   }
 
-  /** Stores the instance in datastore.
+  /**
+   * Stores the instance in datastore.
    *
    * <p>Must be called within the same transaction as load().
    */
   void store() {
-    ofy().saveWithoutBackup().entity(this);
+    ofy().save().entity(this);
   }
 }
--- a/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
+++ b/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
@@ -1,4 +1,4 @@
-// Copyright 2020 The Nomulus Authors. All Rights Reserved.
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -43,7 +43,7 @@ public class LastSqlTransaction extends ImmutableObject implements DatastoreOnly
     transactionId = newTransactionId;
   }
 
-  LastSqlTransaction withNewTransactionId(long transactionId) {
+  LastSqlTransaction cloneWithNewTransactionId(long transactionId) {
     checkArgument(
         transactionId > this.transactionId,
         "New transaction id (%s) must be greater than the current transaction id (%s)",
@@ -59,11 +59,12 @@ public class LastSqlTransaction extends ImmutableObject implements DatastoreOnly
   /**
    * Loads the instance.
    *
-   * <p>Must be called within a transaction.
+   * <p>Must be called within an Ofy transaction.
    *
-   * <p>Creates a new instance of the singleton if it is not already present in the datastore,
+   * <p>Creates a new instance of the singleton if it is not already present in Cloud Datastore,
    */
   static LastSqlTransaction load() {
+    ofy().assertInTransaction();
     LastSqlTransaction result = ofy().load().key(KEY).now();
     return result == null ? new LastSqlTransaction() : result;
   }
--- a/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
+++ b/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
@@ -17,6 +17,7 @@ package google.registry.schema.replay;
 import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.annotation.Entity;
 import com.googlecode.objectify.annotation.Id;
@@ -37,7 +38,8 @@ public class LastSqlTransaction extends ImmutableObject {
 
   LastSqlTransaction() {}
 
-  private LastSqlTransaction(long newTransactionId) {
+  @VisibleForTesting
+  LastSqlTransaction(long newTransactionId) {
     transactionId = newTransactionId;
   }
 
--- a/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
+++ b/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
@@ -25,7 +25,7 @@ import google.registry.model.ImmutableObject;
 
 /** Datastore entity to keep track of the last SQL transaction imported into the datastore. */
 @Entity
-public class LastSqlTransaction extends ImmutableObject {
+public class LastSqlTransaction extends ImmutableObject implements DatastoreOnlyEntity {
 
   /** The key for this singleton. */
   public static final Key<LastSqlTransaction> KEY = Key.create(LastSqlTransaction.class, 1);
--- a/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
+++ b/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
@@ -1,0 +1,76 @@
+// Copyright 2020 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.schema.replay;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.model.ofy.ObjectifyService.ofy;
+
+import com.googlecode.objectify.Key;
+import com.googlecode.objectify.annotation.Entity;
+import com.googlecode.objectify.annotation.Id;
+import google.registry.model.ImmutableObject;
+
+/**
+ * Datastore entity to keep track of the last SQL transaction imported into the datastore.
+ */
+@Entity
+public class LastSqlTransaction extends ImmutableObject {
+
+  /** The key for this singleton. */
+  public static final Key<LastSqlTransaction> KEY = Key.create(LastSqlTransaction.class, 1);
+
+  @SuppressWarnings("unused")
+  @Id
+  private long id = 1;
+
+  private long transactionId;
+
+  LastSqlTransaction() {
+    transactionId = -1;
+  }
+
+  void setTransactionId(long transactionId) {
+    checkArgument(
+        transactionId > this.transactionId,
+        "New transaction id (%s) must be greater than the current transaction id (%s)",
+        transactionId,
+        this.transactionId);
+    this.transactionId = transactionId;
+  }
+
+  long getTransactionId() {
+    return transactionId;
+  }
+
+  /**
+   * Loads the instance.
+   *
+   * <p>Must be called within a transaction.
+   *
+   * <p>Creates a new instance of the singleton if it is not already present in the datastore,
+   */
+  static LastSqlTransaction load() {
+    LastSqlTransaction result = ofy().load().key(KEY).now();
+    return result == null ? new LastSqlTransaction() : result;
+  }
+
+  /** Stores the instance in datastore.
+   *
+   * <p>Must be called within the same transaction as load().
+   */
+  void store() {
+    ofy().saveWithoutBackup().entity(this);
+  }
+}
--- a/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
+++ b/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java
@@ -65,13 +65,4 @@ public class LastSqlTransaction extends ImmutableObject {
     LastSqlTransaction result = ofy().load().key(KEY).now();
     return result == null ? new LastSqlTransaction() : result;
   }
-
-  /**
-   * Stores the instance in datastore.
-   *
-   * <p>Must be called within the same transaction as load().
-   */
-  void store() {
-    ofy().save().entity(this);
-  }
 }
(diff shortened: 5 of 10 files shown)


mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/738)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @hstonec)

___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/738#-MF2MIHx0NkCDWVNlFWF:-MUoljBtBtnZGPOpbQgY:b-z947og) ([raw file](https://github.com/google/nomulus/blob/bc6860de2bda19fa7509197b5c42518f39bd10f7/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L40)):*

The broader question is: how do we deal with structural differences between the datastore and SQL representations? And the answer is that this is still an area that we don't have a good handle on, AFAICT.

@gbrodman did some work on this a while back (e.g. #562), but AFAIK this is still not fully fleshed out.

As far as where it belongs in the replication mechanism, my inclination is to put it into transaction persistence (everything persisted in the Transaction table has to be a datastore entity).  But it gets weird because most of these entities are dealt with from the business practice code.  We may need to introduce an abstraction layer.

TL;DR -- someone needs to look into this.
</blockquote></details>

I haven't thought about this problem for quite a while, adding it to my TODO list.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 42 at r2](https://reviewable.io/reviews/google/nomulus/738#-MDu0sl58nfpqOQaknJJ:-MUolZ6P8rlitgFukCS5:byf846g) ([raw file](https://github.com/google/nomulus/blob/bc6860de2bda19fa7509197b5c42518f39bd10f7/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L42)):*

We should be using a `Lock`; this is exactly what it's for.  We use a `Lock` for all other cron jobs that we absolutely do not want to have multiple instances running simultaneously.  I don't think we can trust cron here.
</blockquote></details>

Discussed out-of-band.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 63 at r2](https://reviewable.io/reviews/google/nomulus/738#-MDu1TNa9Es-v5lYIx0G:-MUolfC70PHjef0D_hH6:byf846g) ([raw file](https://github.com/google/nomulus/blob/bc6860de2bda19fa7509197b5c42518f39bd10f7/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L63)):*

I respect your strong objection but I think we're out of sync on a couple of key points.  Let's break this down:

- GAE cron is documented as not allowing overrun.  If it does overrun, that's a bug.
- In general, actions can run outside of cron.  However, there is no reason to expect that _this_ action should ever be run outside of cron (we'll still be safe if it is, read on :-))
- The Lock approach is error prone because we're introducing two datastore mutations that must be sequential with respect to each other and their enclosed action.  If we did run into a situation where a lot of writes were happening concurrently (and short of a serious GAE outage, we won't), lock contention would likely be an issue.  Because we can never guarantee that a lock will be released by the owning action, Lock includes a lease length.  In the event of a such a failure, that Lock would be inaccessible until that lease expires, so in a high-error rate environment we would be losing a lot of processing time awaiting lease expiration. Lock is not a "good" solution, it's just the best we can do in certain circumstances and I wouldn't want to introduce it where we don't need it.
- From what I can see, the only places we use Lock are 1) in mapreduces to prevent overrun and 2) actions that _mutate external systems_.  Lock makes sense in this context because there is no other way to ensure that a concurrent thread is not attempting this mutation and no way of rolling the mutation back.
- The replication cron job does not mutate an external resource.  Its entire output is expressed in a datastore mutation that ultimately relies on the update of a single resource (LastSqlTransaction): if any other replication transaction were to run concurrently, it would necessarily collide on that resource and one of the transactions would have to roll back.  As we've said, this should not be a common occurrence. It is a _pathological_ case that would be introduced either by a bug in GAE's cron scheduler or by someone manually invoking the endpoint so there's no issue here with multiple actions repeatedly colliding and degrading datastore performance.
- LastSqlTransaction effectively serves the same purpose as Lock in this context, only within the scope of a single transaction.
- This cron job should be the only thing touching datastore during the timeframe in which it is executed, so datastore contention in general should be non-existent.

Am I missing something here?

Again, we don't need to resolve this now.  I intend to benchmark both approaches and it may well be that this approach doesn't perform well enough (though our current mutation rate is a pretty low bar).  But if it performs acceptably, I believe this is the simpler, more elegant approach and I honestly wouldn't feel right about complicating it.
</blockquote></details>

Discussed out-of-band.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/738)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/738#-MDph_lMBRN7OQzoAh3B:-MUp80Pp-hynTWTKnZWT:b-nunfiy) ([raw file](https://github.com/google/nomulus/blob/4196c1458b71e5b3a2e723937d721f0adb9a8352/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L44)):*

Oops, yeah.  This started out as a mutable object but then one our tests forbade it and I lost track of the interface change.
It is currently manipulated in a datastore transaction, which should address the concurrency concerns you have described.  Furthermore, it should only be manipulated by a cron job which should only have a single instance running at any given time.
The protection in the setter (or its new equivalent) is still appropriate, AFAICT.  Unless you know of a way to enforce the transaction id constraint from datastore/objectify?
</blockquote></details>

How about `cloneWithNewTransactionId` or similar.  Right now in this name there's a verb missing that doesn't make it 100% clear what's going on.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 1 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp7IIqEW68G_H-jZvf:-MUp7IIqEW68G_H-jZvg:b-gvqj8v) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L1)):*
> ```Java
> // Copyright 2020 The Nomulus Authors. All Rights Reserved.
> ```

Time to update the year on all of these files.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 48 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8DmZ3NCcl7tvpmx7:-MUp8DmZ3NCcl7tvpmx8:bgpvmo) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L48)):*
> ```Java
>   LastSqlTransaction withNewTransactionId(long transactionId) {
>     checkArgument(
>         transactionId > this.transactionId,
> ```

What exactly is this protecting against here?  Let's say that for some reason two versions of this action are running simultaneously; where is the protection that ensures that once one commits, the other is guaranteed to fail?  This check here is running against a version of the object in memory, not necessarily against the latest version loaded transactionally from the DB.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 62 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8UIp5ZuhBH7aoX4h:-MUp8UIp5ZuhBH7aoX4i:b-6m4giy) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L62)):*
> ```Java
>    * Loads the instance.
>    *
>    * <p>Must be called within a transaction.
> ```

Rather than leaving this comment, how about enforcing it by using the `assertInTransaction()` method?
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 64 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8d_WBjIw2TJlHKFn:-MUp8d_WBjIw2TJlHKFo:b-bzo3d6) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L64)):*
> ```Java
>    * <p>Must be called within a transaction.
>    *
>    * <p>Creates a new instance of the singleton if it is not already present in the datastore,
> ```

`in Cloud Datastore`?  "the datastore" is ambiguous ... we have several things that could plausibly be called that.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/738#-MF2MIHx0NkCDWVNlFWF:-MUp7cb06XV_AL77SRX3:b70o61v) ([raw file](https://github.com/google/nomulus/blob/bc6860de2bda19fa7509197b5c42518f39bd10f7/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L40)):*

I haven't thought about this problem for quite a while, adding it to my TODO list.
</blockquote></details>

Leave a TODO in the code then?  It sounds like, as of this PR, this action is not yet going to be complete.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 74 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8zSz6sqSUB9thZWb:-MUp8zSz6sqSUB9thZWc:b2s43o7) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L74)):*
> ```Java
> 
>   /**
>    * Apply a transaction to datastore, returns true if there was a fatel error and the batch should
> ```

`fatal`
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 110 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9YTsEScmog0PeOCf:-MUp9YTsEScmog0PeOCg:b-91dg7l) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L110)):*
> ```Java
>                 Transaction.deserialize(txnEntity.getContents()).writeToDatastore();
>               } catch (IOException e) {
>                 throw new RuntimeException(e);
> ```

Add an error message here?
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 122 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9meP9ZDC2uBXvugG:-MUp9meP9ZDC2uBXvugH:b-n1gzyc) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L122)):*
> ```Java
>   @Override
>   public void run() {
>     for (TransactionEntity txnEntity : getTransactionBatch()) {
> ```

Add some `INFO` logging here for when the action starts, ends, and for each batch loaded/applied.
___
*[core/src/test/java/google/registry/schema/replay/ReplicateToDatastoreActionTest.java, line 1 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9wChD5MQcjvCZR2Y:-MUp9wChD5MQcjvCZR2Z:bb6ht36) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/test/java/google/registry/schema/replay/ReplicateToDatastoreActionTest.java#L1)):*
> ```Java
> // Copyright 2020 The Nomulus Authors. All Rights Reserved.
> ```

2021

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/738)* status: 0 of 10 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @hstonec)

___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/738#-MDph_lMBRN7OQzoAh3B:-MUsH7XX-a74Zu4r8e2h:b-iqnuet) ([raw file](https://github.com/google/nomulus/blob/4196c1458b71e5b3a2e723937d721f0adb9a8352/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L44)):*

How about `cloneWithNewTransactionId` or similar.  Right now in this name there's a verb missing that doesn't make it 100% clear what's going on.
</blockquote></details>

Yeah, I was following the pattern of Builder methods.  But cloneWith is better in this case, and consistent with what we're doing elsewhere.  Done.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 1 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp7IIqEW68G_H-jZvf:-MUsGNc72WJAk9r7a6TS:b-896fix) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L1)):*

Time to update the year on all of these files.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 48 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8DmZ3NCcl7tvpmx7:-MUsHiGhA2gUpn3rhrCN:bssk6oh) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L48)):*

What exactly is this protecting against here?  Let's say that for some reason two versions of this action are running simultaneously; where is the protection that ensures that once one commits, the other is guaranteed to fail?  This check here is running against a version of the object in memory, not necessarily against the latest version loaded transactionally from the DB.
</blockquote></details>

This is purely an argument check: since the transaction id is set externally, it would be a bug to set it to a value <= the current transaction id as recorded in memory. 

Protection against collisions happens in `ReplicateToDatastoreAction.applyTransaction()`.  By virtue of the fact that the LastSqlTransaction is loaded, modified and saved in a single datastore transaction the transaction (which includes the replay of the _SQL transaction_) will fail if the transaction id has been modified in another transaction.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 62 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8UIp5ZuhBH7aoX4h:-MUsKAIMAFIXurbzCKYr:bpko8t3) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L62)):*

Rather than leaving this comment, how about enforcing it by using the `assertInTransaction()` method?
</blockquote></details>

Done. (Though I left it here, too)

Actually, this comment was a lie, since I was previously calling this method from a "doTransactionless" context.  From what I'm seeing online, "transactionless" for Ofy [doesn't seem to be materially different from "transactional"](https://github.com/objectify/objectify/wiki/Concepts#transactionless) for non-batch operations, so I've converted the aforementioned transactionless context to transactional.

Since this code is using ofy() directly, I've made Ofy.assertInTransaction() public (it was previously package-protected).  There's a case to be made for using ofyTm universally, and I considered just switching this code over to ofyTm, but we currently don't have a VKey for LastSqlTransaction and don't really need one so I was reluctant to add one just to hammer this square peg into a round hole.

We might want to reconsider this at some point, I don't feel strongly either way.
___
*[core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java, line 64 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8d_WBjIw2TJlHKFn:-MUsKTbr2ydn8nNMFCDb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/LastSqlTransaction.java#L64)):*

`in Cloud Datastore`?  "the datastore" is ambiguous ... we have several things that could plausibly be called that.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/738#-MF2MIHx0NkCDWVNlFWF:-MUsVg0K-V5VM4htzJ-Q:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bc6860de2bda19fa7509197b5c42518f39bd10f7/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L40)):*

Leave a TODO in the code then?  It sounds like, as of this PR, this action is not yet going to be complete.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 74 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp8zSz6sqSUB9thZWb:-MUsVltW1Enbi9akD0lF:b-896fix) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L74)):*

`fatal`
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 110 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9YTsEScmog0PeOCf:-MUsWJPf7tPyBasf657V:b-cj1u98) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L110)):*

Add an error message here?
</blockquote></details>

Done, though in general I'm not sure it's helpful to have additional text when converting from checked exception to RuntimeException (and it seems like we currently don't do this more often than we do)
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 122 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9meP9ZDC2uBXvugG:-MUsZCBZAJ0gnEeh8L4N:bz7ngai) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L122)):*

Add some `INFO` logging here for when the action starts, ends, and for each batch loaded/applied.
</blockquote></details>

Good call.  Done.
___
*[core/src/test/java/google/registry/schema/replay/ReplicateToDatastoreActionTest.java, line 1 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9wChD5MQcjvCZR2Y:-MUsZKYiD1uEjbSHUUD2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/test/java/google/registry/schema/replay/ReplicateToDatastoreActionTest.java#L1)):*

2021
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/738)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 110 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9YTsEScmog0PeOCf:-MUytExG38cIgy8LGhDr:b-ifi8yg) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L110)):*

Done, though in general I'm not sure it's helpful to have additional text when converting from checked exception to RuntimeException (and it seems like we currently don't do this more often than we do)
</blockquote></details>

The problem is that the checked exception is going to be some generic Java library exception message that doesn't actually tell us anything about what our own code was trying to do when the failure occurred.  Writing our own message about what our code is doing makes debugging a lot simpler.  Without it you just have some random generic Java exception in the logs and then need to go hunting by line numbers to figure out what was happening.

In the places where we don't currently do so, we're often wrapping our own exceptions, not generic Java library ones.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 1 at r4](https://reviewable.io/reviews/google/nomulus/738#-MUyt8-jALSV9bx_btzr:-MUyt8-jALSV9bx_btzs:b-vmzkye) ([raw file](https://github.com/google/nomulus/blob/22e88a0d0cf0d71f27df2ad0daa4955d1c1727e2/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L1)):*
> ```Java
> // Copyright 2020 The Nomulus Authors. All Rights Reserved.
> ```

Wrong year?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/738)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @hstonec)

___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 110 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9YTsEScmog0PeOCf:-MUyy_zf03bZDWZMzSpM:b-w4juhj) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L110)):*

The problem is that the checked exception is going to be some generic Java library exception message that doesn't actually tell us anything about what our own code was trying to do when the failure occurred.  Writing our own message about what our code is doing makes debugging a lot simpler.  Without it you just have some random generic Java exception in the logs and then need to go hunting by line numbers to figure out what was happening.

In the places where we don't currently do so, we're often wrapping our own exceptions, not generic Java library ones.
</blockquote></details>

Well, in addition to a line number you also get a class and method name -- I just don't see a text message describing context to be providing any additional value.
___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 1 at r4](https://reviewable.io/reviews/google/nomulus/738#-MUyt8-jALSV9bx_btzr:-MUz-v4h6hsQUg9UVWjq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/22e88a0d0cf0d71f27df2ad0daa4955d1c1727e2/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L1)):*

Wrong year?
</blockquote></details>

Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/738)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @hstonec)

___
*[core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java, line 110 at r3](https://reviewable.io/reviews/google/nomulus/738#-MUp9YTsEScmog0PeOCf:-MUzT6ee4HAwhKapiLzU:b-pwfe1) ([raw file](https://github.com/google/nomulus/blob/114dc405b21e2d6897b45ff177deb9d7b946cc00/core/src/main/java/google/registry/schema/replay/ReplicateToDatastoreAction.java#L110)):*

Well, in addition to a line number you also get a class and method name -- I just don't see a text message describing context to be providing any additional value.
</blockquote></details>

This method does multiple things, not just one thing.  Knowing immediately when looking at the error message in the logs exactly which thing went wrong is useful, rather than having to track it down by looking at the source and consulting the line number.  Also keep in mind that, depending on subsequent CLs, the line #s may no longer even match up, so you'd need to go back and look up what revision the live build is running off and then go view the source at that version.

Or, alternatively, just grep the file/codebase for this error message.  This really does make things easier to track down.

PR #744, opened 2020-08-04
Allow allocation token discounts on premiums and for multiple years

CydeWeys:
These are optional and default to off, but as of this PR we will have the ability to, when we want it, create an allocation token with a discount valid for premium names as well as standard names, and that will discount as many years of the initial registration as we want (rather than just the first).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/744)


CydeWeys:
Will get to the domain check flow tests as part of this PR next, but wanted to send out the bulk of it for review now.

PR #746, opened 2020-08-06
Add SQL schema for GracePeriodHistory

hstonec:
This PR added a `GracePeriodHistory` class to represent the historic `GracePeriod` entities. `GracePeriodHistory` has all the fields of `GracePeriod` plus `historyRevisionId` which is the foreign key to the `DomainHistory` table.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/746)


hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/flyway/V46__add_grace_period_history.sql, line 16 at r1](https://reviewable.io/reviews/google/nomulus/746#-MEOLpKu38_dRGUhyizK:-MIPFpGb5SestoOYLWR9:b-8nabtm) ([raw file](https://github.com/google/nomulus/blob/f52ac96e791377266d5cbbaa85e6e4675ad46451/db/src/main/resources/sql/flyway/V46__add_grace_period_history.sql#L16)):*
> How about grace_period_id then in both places for clarity?  That tends to be how the rest of our schema is named; we don't have bare id fields.

Done. We may want to review our schema as there are some tables that uses bare `id`.

> And more importantly, what exactly is this? Is this the same ID as in the GracePeriod entity? It seems not?

It is the same id now. We had the discussion about how we want to generate `long id` and decided to continue using `ObjectifyService.allocateId()` so it solves the problem.

> What does history_revision_id below refer to? Is that for the domain in question, or is there a separately tracked revision id for grace periods? If the former it should be domain_history_revision_id for clarity.

Yes, it is the foreign key, done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[db/src/main/resources/sql/flyway/V59__add_grace_period_history.sql, line 30 at r3](https://reviewable.io/reviews/google/nomulus/746#-MIViW4LDtm3HEZfkZK3:-MIViW4LDtm3HEZfkZK4:bqh8aaa) ([raw file](https://github.com/google/nomulus/blob/e02358b319f37d2e1a246b919ed0d161b277e452/db/src/main/resources/sql/flyway/V59__add_grace_period_history.sql#L30)):*
> ```TSQL
>     type text not null,
>     domain_history_revision_id int8,
>     history_revision_id int8,
> ```

I'm still struggling to understand this.  Why are there two separate revision ids?  Shouldn't there just be one?  Grace periods aren't revisioned separately from the domains they are attached to.  Shouldn't there just be one type of revision id, and it has the same value as the revision id in the `DomainHistory` table?

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @hstonec)

___
*[core/src/test/java/google/registry/tools/GenerateSqlSchemaCommandTest.java, line 89 at r3](https://reviewable.io/reviews/google/nomulus/746#-MI_H5vZAi2-z_tBxzdb:-MI_H5v_Bhkr4Acsl86q:b-gryzfe) ([raw file](https://github.com/google/nomulus/blob/e02358b319f37d2e1a246b919ed0d161b277e452/core/src/test/java/google/registry/tools/GenerateSqlSchemaCommandTest.java#L89)):*
> ```
> /tmp/schema.sql
> ```
Why this change?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/test/java/google/registry/tools/GenerateSqlSchemaCommandTest.java, line 89 at r3](https://reviewable.io/reviews/google/nomulus/746#-MI_H5vZAi2-z_tBxzdb:-MIdcz1jDIZKXLrgNiJ0:b-9e8ut9) ([raw file](https://github.com/google/nomulus/blob/e02358b319f37d2e1a246b919ed0d161b277e452/core/src/test/java/google/registry/tools/GenerateSqlSchemaCommandTest.java#L89)):*

> ```
> /tmp/schema.sql
> ```
Why this change?
</blockquote></details>

Reverted, that was a mistake.
___
*[db/src/main/resources/sql/flyway/V59__add_grace_period_history.sql, line 30 at r3](https://reviewable.io/reviews/google/nomulus/746#-MIViW4LDtm3HEZfkZK3:-MIdd3G4DHHLneXVcPg8:bwy3v89) ([raw file](https://github.com/google/nomulus/blob/e02358b319f37d2e1a246b919ed0d161b277e452/db/src/main/resources/sql/flyway/V59__add_grace_period_history.sql#L30)):*

I'm still struggling to understand this.  Why are there two separate revision ids?  Shouldn't there just be one?  Grace periods aren't revisioned separately from the domains they are attached to.  Shouldn't there just be one type of revision id, and it has the same value as the revision id in the `DomainHistory` table?
</blockquote></details>

This has been fixed. `history_revision_id` and `domain_history_revisoin_id` are the same thing, `history_revision_id` was created because I used the wrong annotation on `gracePeriodHistories` before.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/java/google/registry/model/domain/GracePeriodBase.java, line 36 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIdepR_0PsyEHpkDQFM:-MIdepR_0PsyEHpkDQFN:ba0n8ha) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/main/java/google/registry/model/domain/GracePeriodBase.java#L36)):*
> ```
> @javax.persistence.Id Long gracePeriodId;
> ```
Per our offline discussion, I added this id to Datastore schema to simplify the entity migration as this is the primary key in Cloud sql schema.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 11 unresolved discussions (waiting on @hstonec and @weiminyu)

___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 30 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyEgyg8JI7Sy-SrNxo:-MIyEgyg8JI7Sy-SrNxp:b-lj7qt4) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L30)):*
> ```Java
> import javax.annotation.Nullable;
> 
> public class GracePeriodCollectionSubject extends Subject {
> ```

"Collection" is an unnecessarily long word here.  Just name it `GracePeriodsSubject`.

Also needs a Javadoc.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 44 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFC8REtNNyyluSJLI:-MIyFC8REtNNyyluSJLJ:b2c8qi6) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L44)):*
> ```Java
>   }
> 
>   public static SimpleSubjectBuilder<
> ```

Needs Javadoc
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 46 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyEu2eAHN1AQlmxi_Z:-MIyEu2eAHN1AQlmxi__:b-zcb9mg) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L46)):*
> ```Java
>   public static SimpleSubjectBuilder<
>           GracePeriodCollectionSubject, Collection<? extends GracePeriodBase>>
>       assertAboutGracePeriodCollection() {
> ```

Simpler name: `assertAboutGracePeriods()`
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 50 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyEwn74h13MVaFE0gF:-MIyEwn74h13MVaFE0gG:bm6edlx) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L50)):*
> ```Java
>   }
> 
>   public void containsExactlyExceptId(@Nullable GracePeriodBase... expected) {
> ```

This is weird to see a `containsExactlyExceptId` method but no `containsExactly` method, which would seemingly be much more commonly used?  The latter should definitely exist.  It might need to be named `containsExactlyElementsIn` depending on its signature (for consistency with other such methods in Truth).
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 50 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFAqh9nwopoVFP2TT:-MIyFAqh9nwopoVFP2TU:b2c8qi6) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L50)):*
> ```Java
>   }
> 
>   public void containsExactlyExceptId(@Nullable GracePeriodBase... expected) {
> ```

Needs Javadoc
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFHh49Squztg8Oy92:-MIyFHh50IkMe1tIx3ii:beoo156) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L52)):*
> ```Java
>   public void containsExactlyExceptId(@Nullable GracePeriodBase... expected) {
>     if (actual == null) {
>       assertThat(expected).isNull();
> ```

I think these need messages for clarity.

Also there has to be a better way to do this than `if null { assert is null } else { assert is not null }`.  This pattern has cropped up repeatedly and should be extracted into a helper somewhere (and I suspect Truth might already provide one somewhere).
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 57 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFbkc2seAN9VwFAnr:-MIyFbkc2seAN9VwFAns:b-bewzdh) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L57)):*
> ```Java
>     }
>     if (actual != null) {
>       assertThat(actual).hasSize(expected.length);
> ```

This isn't a good way to do it because all you'll end up getting is an error message that the size of the collection isn't correct, whereas what you want is a full diff of the expected elements that are missing and the unexpected elements that are present (as a lot of other Truth subjects do).  See various `containsExactlyElementsIn()` methods.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 60 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyG5cMCMR271ft4aXk:-MIyG5cMCMR271ft4aXl:b-w35x2g) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L60)):*
> ```Java
>       Iterator<? extends GracePeriodBase> actualIte = actual.iterator();
>       Iterator<? extends GracePeriodBase> expectedIte = Arrays.asList(expected).iterator();
>       while (actualIte.hasNext()) {
> ```

This seems like it's ordering dependent?  That seems wrong, because collections aren't inherently ordered.  I don't think we care about grace period ordering anywhere else.
___
*[core/src/test/java/google/registry/testing/GracePeriodSubject.java, line 46 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyGGLw0B6jrPvfMIQB:-MIyGGLw0B6jrPvfMIQC:b-p7td3n) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodSubject.java#L46)):*
> ```Java
> 
>   public void isEqualExceptId(@Nullable GracePeriod expected) {
>     if (actual == null) {
> ```

Here's that pattern again.  And again, no messages.  A helper will definitely help us do this better, and with consistent messaging.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 25 files reviewed, 11 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 30 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyEgyg8JI7Sy-SrNxo:-MJ3W2K63K4s-VlrhQgA:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L30)):*

"Collection" is an unnecessarily long word here.  Just name it `GracePeriodsSubject`.

Also needs a Javadoc.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 44 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFC8REtNNyyluSJLI:-MJ3WN9q32YlznP6eUFs:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L44)):*

Needs Javadoc
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 46 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyEu2eAHN1AQlmxi_Z:-MJ3WPIKCt6tqVtogoqS:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L46)):*

Simpler name: `assertAboutGracePeriods()`
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 50 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyEwn74h13MVaFE0gF:-MJ3XsYE0AOojp_PeSBu:b-b73zm4) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L50)):*

This is weird to see a `containsExactlyExceptId` method but no `containsExactly` method, which would seemingly be much more commonly used?  The latter should definitely exist.  It might need to be named `containsExactlyElementsIn` depending on its signature (for consistency with other such methods in Truth).
</blockquote></details>

In practice, I don't think we would need `containsExactly` or `containsExactlyElementsIn` in this subject because we can just use the normal `assertThat` to achieve that .
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 50 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFAqh9nwopoVFP2TT:-MJ3imy18GL7TjNzEjsv:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L50)):*

Needs Javadoc
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFHh49Squztg8Oy92:-MJ5TFCI-FEIj1x6FrV_:bl4zbpb) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L52)):*

I think these need messages for clarity.

Also there has to be a better way to do this than `if null { assert is null } else { assert is not null }`.  This pattern has cropped up repeatedly and should be extracted into a helper somewhere (and I suspect Truth might already provide one somewhere).
</blockquote></details>

Done. Extracted this to a helper function.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 57 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyFbkc2seAN9VwFAnr:-MJ5TN0z-G-KUPrCf4kX:b-s51nwr) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L57)):*

This isn't a good way to do it because all you'll end up getting is an error message that the size of the collection isn't correct, whereas what you want is a full diff of the expected elements that are missing and the unexpected elements that are present (as a lot of other Truth subjects do).  See various `containsExactlyElementsIn()` methods.
</blockquote></details>

Refactored this code to compare two sets directly, so no need to have this any more.
___
*[core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java, line 60 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyG5cMCMR271ft4aXk:-MJ5TUO22if-JTRPetLb:bhztbhu) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodCollectionSubject.java#L60)):*

This seems like it's ordering dependent?  That seems wrong, because collections aren't inherently ordered.  I don't think we care about grace period ordering anywhere else.
</blockquote></details>

Good point, refactored the code to do order independent comparison.
___
*[core/src/test/java/google/registry/testing/GracePeriodSubject.java, line 46 at r4](https://reviewable.io/reviews/google/nomulus/746#-MIyGGLw0B6jrPvfMIQB:-MJ5THyf4fspLneEHMDR:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44796a0d243eeafe635600937803ad4454d9aa83/core/src/test/java/google/registry/testing/GracePeriodSubject.java#L46)):*

Here's that pattern again.  And again, no messages.  A helper will definitely help us do this better, and with consistent messaging.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @hstonec, and @weiminyu)

___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 28 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCvoSoCF2dnTgmyS7f:-MJCvoSoCF2dnTgmyS7g:b2jlvy4) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L28)):*
> ```Java
> 
>   /** Asserts that both of the given objects are either null or nonnull. */
>   public static void assertBothNullOrNonnull(@Nullable Object thisObj, @Nullable Object thatObj) {
> ```

I just realized that `assertThat(somethingThatIsNull).isEqualTo(somethingElseThatIsNull);` already works, so we can delete this entirely and just use the basic `assertThat` from call sites.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @hstonec, and @weiminyu)

___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 28 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCvoSoCF2dnTgmyS7f:-MJCwmstBoPssHkGVgjK:b-6fh0yi) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L28)):*

I just realized that `assertThat(somethingThatIsNull).isEqualTo(somethingElseThatIsNull);` already works, so we can delete this entirely and just use the basic `assertThat` from call sites.
</blockquote></details>

Wait never mind, that's not exactly what we're doing here.  We're not checking for strict equality at all.  Belay that.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @hstonec, and @weiminyu)

___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 15 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCxrRJ6hzrz-esNxQU:-MJCxrRK0_qHDcmDYiiC:b-oqiutl) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L15)):*
> ```Java
> 
> package google.registry.testing;
> ```

Can you put this in `common/src/testing/java/google/registry/testing/truth/` so that it can be used by all subprojects, not just core?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 25 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @hstonec, and @weiminyu)

___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 25 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCy4Qx64ceKm3QFf74:-MJCy4Qx64ceKm3QFf75:b-680xkz) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L25)):*
> ```Java
>  * covered.
>  */
> public class TruthHelper {
> ```

How about `TruthUtils`?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 25 files reviewed, 10 unresolved discussions (waiting on @hstonec and @weiminyu)

___
*[core/src/test/java/google/registry/testing/GracePeriodsSubject.java, line 35 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCzoDDC3xiB4Wg6nXy:-MJCzoDDC3xiB4Wg6nXz:bd0oq7z) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/GracePeriodsSubject.java#L35)):*
> ```Java
> 
> /**
>  * Truth subject for asserting things about a collection of {@link GracePeriodBase} that are not
> ```

"Truth Subject for assertions about collections of {@link GracePeriodBase} entities."

Shorter and sweeter.  "that are not built in" in particular is redundant, because of course Truth isn't going to come with out-of-the-box assertions for dealing with stuff that only exists in our data model.
___
*[core/src/test/java/google/registry/testing/GracePeriodsSubject.java, line 66 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJD-JaH1LzlfkQTGpj-:-MJD-JaH1LzlfkQTGpj0:btp8u1s) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/GracePeriodsSubject.java#L66)):*
> ```Java
>     if (actual != null) {
>       List<GracePeriodBase> expectedList = Arrays.asList(expected);
>       assertWithMessage("Missing GracePeriod in the actual collection: ")
> ```

If this fails then won't the output of the next assertWithMessage not occur, because an exception is already thrown?  We want to output both.

See how Truth's built-in `containsExactlyElementsIn()` method is implemented.  It's not trivial.  We're gonna have a better time here if we find a way to leverage that rather than trying to re-implement it from scratch.  Note how our own `ImmutableObjectSubject.isEqualExceptFields()` does successfully take this approach, so you can use that as a template.

Or maybe we should rethink this all entirely and just not even provide this method.  Generally we want to make exact assertions on what should be present.  What's the use case for making inexact assertions where we pass in some stuff but then tack on "oh but ignore this grace period"?  Shouldn't the test in question know exactly what grace periods should be present, and then specifically assert that those exist?  Or if I'm misreading the intent of this method and it's really just `ImmutableObjectSubject.isEqualExceptFields()`, then why not simply call that?
___
*[core/src/test/java/google/registry/testing/GracePeriodsSubject.java, line 75 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJD00gB5CwMY3RBE_Oq:-MJD00gB5CwMY3RBE_Or:b-1xdhzj) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/GracePeriodsSubject.java#L75)):*
> ```Java
>   }
> 
>   private static ImmutableList<GracePeriodBase> containsAll(
> ```

I don't think anything that requires implementing all this logic from scratch to determine what's missing is the right approach.  It should already be handled by Truth; if we're rewriting it ourselves then we're not so much using Truth as we are re-implementing it.
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 22 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCy6qP9fNRcCAH4rt-:-MJCy6qP9fNRcCAH4rt0:b-vwxvid) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L22)):*
> ```Java
> 
> /**
>  * Helper class to contain assertion functions that {@link com.google.common.truth.Truth} is not
> ```

"Utils class containing helper functions for {@link Truth}."  Something like that.
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 28 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCyI3u-w8Q-MN36Thi:-MJCyI3u-w8Q-MN36Thj:bs5v26) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L28)):*
> ```Java
> 
>   /** Asserts that both of the given objects are either null or nonnull. */
>   public static void assertBothNullOrNonnull(@Nullable Object thisObj, @Nullable Object thatObj) {
> ```

How about naming this `assertNullnessParity`?
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 28 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCyz_j0rXqQ9AoNDGd:-MJCyz_j0rXqQ9AoNDGe:bdgf5hm) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L28)):*
> ```Java
> 
>   /** Asserts that both of the given objects are either null or nonnull. */
>   public static void assertBothNullOrNonnull(@Nullable Object thisObj, @Nullable Object thatObj) {
> ```

Minor, but Truth seems to prefer `@NullableDecl` over `@Nullable` for parameters in its assertion functions so maybe we should do the same for consistency?  So long as it doesn't require adding an additional dependency on `org.checkerframerwork.checker` anyway (if it does, don't bother).
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 379 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCzZ7_CMUxGdKb4i_N:-MJCzZ7_CMUxGdKb4i_O:bec2ouk) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/db/src/main/resources/sql/schema/db-schema.sql.generated#L379)):*
> ```
> 
>     create table "GracePeriod" (
>        grace_period_id int8 not null,
> ```

Oh this is fun.  Both are 8 byte ints but `bigserial` is necessarily incremented from a sequence. Wheeee.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/test/java/google/registry/testing/GracePeriodsSubject.java, line 35 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCzoDDC3xiB4Wg6nXy:-MJE7QfDC3mtW-izdPN2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/GracePeriodsSubject.java#L35)):*

"Truth Subject for assertions about collections of {@link GracePeriodBase} entities."

Shorter and sweeter.  "that are not built in" in particular is redundant, because of course Truth isn't going to come with out-of-the-box assertions for dealing with stuff that only exists in our data model.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/GracePeriodsSubject.java, line 66 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJD-JaH1LzlfkQTGpj-:-MJgosgW1wkc19zDT7Zg:b-7ry2en) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/GracePeriodsSubject.java#L66)):*

If this fails then won't the output of the next assertWithMessage not occur, because an exception is already thrown?  We want to output both.

See how Truth's built-in `containsExactlyElementsIn()` method is implemented.  It's not trivial.  We're gonna have a better time here if we find a way to leverage that rather than trying to re-implement it from scratch.  Note how our own `ImmutableObjectSubject.isEqualExceptFields()` does successfully take this approach, so you can use that as a template.

Or maybe we should rethink this all entirely and just not even provide this method.  Generally we want to make exact assertions on what should be present.  What's the use case for making inexact assertions where we pass in some stuff but then tack on "oh but ignore this grace period"?  Shouldn't the test in question know exactly what grace periods should be present, and then specifically assert that those exist?  Or if I'm misreading the intent of this method and it's really just `ImmutableObjectSubject.isEqualExceptFields()`, then why not simply call that?
</blockquote></details>

Good point, I think removing these two subjects is probably the better solution here so did that.
___
*[core/src/test/java/google/registry/testing/GracePeriodsSubject.java, line 75 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJD00gB5CwMY3RBE_Oq:-MJgoosxDLkUpOYy76fR:bjrf556) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/GracePeriodsSubject.java#L75)):*

I don't think anything that requires implementing all this logic from scratch to determine what's missing is the right approach.  It should already be handled by Truth; if we're rewriting it ourselves then we're not so much using Truth as we are re-implementing it.
</blockquote></details>

This class was removed.
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 15 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCxrRJ6hzrz-esNxQU:-MJgqeQC-ftvR2ThdWLX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L15)):*

Can you put this in `common/src/testing/java/google/registry/testing/truth/` so that it can be used by all subprojects, not just core?
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 22 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCy6qP9fNRcCAH4rt-:-MJgqm74AktbGFQDrQDx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L22)):*

"Utils class containing helper functions for {@link Truth}."  Something like that.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 25 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCy4Qx64ceKm3QFf74:-MJgqmkR8CJI2OZsy7_r:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L25)):*

How about `TruthUtils`?
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/TruthHelper.java, line 28 at r5](https://reviewable.io/reviews/google/nomulus/746#-MJCyI3u-w8Q-MN36Thi:-MJgqqB436cVDwoNKdxm:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1305aa3d564d8fe0552b369ff4e959492c5e9275/core/src/test/java/google/registry/testing/TruthHelper.java#L28)):*

How about naming this `assertNullnessParity`?
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @hstonec and @weiminyu)

___
*[core/src/test/java/google/registry/model/domain/DomainBaseSqlTest.java, line 325 at r6](https://reviewable.io/reviews/google/nomulus/746#-MK0Rt5U2jz3yDFO4XYo:-MK0Rt5V9PxWnAxUgwZt:b6gxegz) ([raw file](https://github.com/google/nomulus/blob/a30d173e25ab5d5fe64fb729fbfb2bf795b434cd/core/src/test/java/google/registry/model/domain/DomainBaseSqlTest.java#L325)):*
> ```Java
>                   .containsExactly(
>                       GracePeriod.create(
>                           GracePeriodStatus.ADD, "4-COM", END_OF_TIME, "registrar1", null, 100L));
> ```

Yeah this is muuuuch nicer and simpler without all those more complicated helpers.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746#-:-MK0SKYe05t26Lr-Z271:bnfp4nl)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @hstonec and @weiminyu)

weiminyu (changes requested):
> Reviewed 1 of 16 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @hstonec)

___
*[common/src/testing/java/google/registry/testing/truth/TruthUtils.java, line 29 at r6](https://reviewable.io/reviews/google/nomulus/746#-MJm-tLg9GLanuILaoVy:-MJm-tLg9GLanuILaoVz:b-3s77fq) ([raw file](https://github.com/google/nomulus/blob/a30d173e25ab5d5fe64fb729fbfb2bf795b434cd/common/src/testing/java/google/registry/testing/truth/TruthUtils.java#L29)):*
> ```
> assertThat(thatObj)
> ```
We may want to use assertWithMessage to give a better description.
'thatObj is not null' can be misleading.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 23 files reviewed, 3 unresolved discussions (waiting on @hstonec)

___
*[common/src/testing/java/google/registry/testing/truth/TruthUtils.java, line 29 at r6](https://reviewable.io/reviews/google/nomulus/746#-MJm-tLg9GLanuILaoVy:-MKHh-rr4FARXk5e5YV1:bywryrj) ([raw file](https://github.com/google/nomulus/blob/a30d173e25ab5d5fe64fb729fbfb2bf795b434cd/common/src/testing/java/google/registry/testing/truth/TruthUtils.java#L29)):*

> ```
> assertThat(thatObj)
> ```
We may want to use assertWithMessage to give a better description.
'thatObj is not null' can be misleading.
</blockquote></details>

Actually we may want the caller to pass in a message for assertion failure.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 22 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[common/src/testing/java/google/registry/testing/truth/TruthUtils.java, line 29 at r6](https://reviewable.io/reviews/google/nomulus/746#-MJm-tLg9GLanuILaoVy:-MKu-GHnDI6FycDXDfMt:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a30d173e25ab5d5fe64fb729fbfb2bf795b434cd/common/src/testing/java/google/registry/testing/truth/TruthUtils.java#L29)):*

Actually we may want the caller to pass in a message for assertion failure.
</blockquote></details>

Done.
___
*[db/src/main/resources/sql/flyway/V70__add_grace_period_history.sql, line 18 at r7](https://reviewable.io/reviews/google/nomulus/746#-MKu1lqk5BkJRvVa4hUU:-MKu1lqk5BkJRvVa4hUV:b-57ro6k) ([raw file](https://github.com/google/nomulus/blob/fae5b19ab827673d29757f9a18e712744c10c893/db/src/main/resources/sql/flyway/V70__add_grace_period_history.sql#L18)):*
> ```
> create table "GracePeriodHistory" (
>     grace_period_history_revision_id int8 not null,
> ```
@cydeweys PTAL. 

So, the previous schema used `gracePeriodRevisionId` as the primary key for `GracePeriodHistory` table which is probably not sufficient because it is possible 2 `DomainHistory` entries have exact same `GracePeriod`s(the gracePeriodId are same) but we still need to store 2 separate entries in the table as they need to have different `domainHistoryRevisionId` for association.

Also, I removed two foreign keys for `BillingEvent` and `BillingRecurrence`, I just feel it is not very reasonable for a `*History` table to have reference to a non-history table.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 22 files reviewed, 2 unresolved discussions

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/746)* status: 0 of 22 files reviewed, 2 unresolved discussions (waiting on @hstonec)

___
*[db/src/main/resources/sql/flyway/V70__add_grace_period_history.sql, line 18 at r7](https://reviewable.io/reviews/google/nomulus/746#-MKu1lqk5BkJRvVa4hUU:-MLrXknE8AXyWZy1mPZb:b-81nlyc) ([raw file](https://github.com/google/nomulus/blob/fae5b19ab827673d29757f9a18e712744c10c893/db/src/main/resources/sql/flyway/V70__add_grace_period_history.sql#L18)):*

> ```
> create table "GracePeriodHistory" (
>     grace_period_history_revision_id int8 not null,
> ```
@cydeweys PTAL. 

So, the previous schema used `gracePeriodRevisionId` as the primary key for `GracePeriodHistory` table which is probably not sufficient because it is possible 2 `DomainHistory` entries have exact same `GracePeriod`s(the gracePeriodId are same) but we still need to store 2 separate entries in the table as they need to have different `domainHistoryRevisionId` for association.

Also, I removed two foreign keys for `BillingEvent` and `BillingRecurrence`, I just feel it is not very reasonable for a `*History` table to have reference to a non-history table.
</blockquote></details>

SGTM.

PR #747, opened 2020-08-06
Fix vkey construction in PollMessage and BillingEvent

hstonec:
This PR removed the code to construct a fake SQL key for VKey in PollMessage and BillingEvent
because it doesn't work with some entities having old key structure.

The reason why we construct the fake SQL key is because we discovered that there were some PollMessage
and BillingEvent entities having duplicate ids. So, we wanted the fake SQL key to be served as an alert to
remind us to fix the data problem before migrating to Cloud SQL. The problem has been fixed for 
PollMessage entities and we are actively working on the fix for BillingEvent entities.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/747)

### DIFF
--- a/core/src/main/java/google/registry/model/billing/BillingEvent.java
+++ b/core/src/main/java/google/registry/model/billing/BillingEvent.java
@@ -351,7 +351,7 @@ public abstract class BillingEvent extends ImmutableObject
     }
 
     public static VKey<OneTime> createVKey(Key<OneTime> key) {
-      return key == null ? null : VKey.create(OneTime.class, key.getId(), key);
+      return VKey.create(OneTime.class, key.getId(), key);
     }
 
     @Override
@@ -493,7 +493,7 @@ public abstract class BillingEvent extends ImmutableObject
     }
 
     public static VKey<Recurring> createVKey(Key<Recurring> key) {
-      return key == null ? null : VKey.create(Recurring.class, key.getId(), key);
+      return VKey.create(Recurring.class, key.getId(), key);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/google/registry/model/billing/BillingEvent.java
```diff
@@ -347,27 +347,11 @@ public DateTime getSyntheticCreationTime() {
...
     }
 
     public static VKey<OneTime> createVKey(Key<OneTime> key) {
-      // TODO(b/159207551): As it stands, the SQL key generated here doesn't mesh with the primary
-      // key type for the table, which is a single long integer.
-      if (key == null) {
-        return null;
-      }
-      Key parent = key.getParent();
-      Key grandparent = (parent != null) ? parent.getParent() : null;
```
CydeWeys, line 363: It seems like this code was trying to solve some kind of problem that will now not be solved by the code on the right side of this diff?  What is the long term plan here?

CydeWeys: Or will this be sufficient once the duplicate IDs are taken care of?

hstonec: 1. The old code was not trying to solve a problem, instead, it aimed to make any caller fail if they try to use the SQL key from the VKey. As it mentioned above, the SQL table for PollMessage and BillingEvent use a long type column as its primary key but the old code constructed a string for it. Ideally, this should work because none of the production code uses the SQL key from the VKey. However, the process to construct the SQL key has a bug in PollMessage(not the code you highlighted), i.e. `key.getParent().getParent()` would throw a NPE when `key` doesn't have its grandparent.
2. I assume you were asking if the new code, i.e. `key == null ? null : VKey.create(OneTime.class, key.getId(), key);` will be sufficient and the answer is yes, because our plan is to use the numeric id of the Ofy key as the primary key for its SQL table.

PR #764, opened 2020-08-11
Re-enable invoicing pipeline deployment

jianglai:
Now that beam deployment is compatible with Java 11. Re-enable this
step.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/764)


jianglai:
Tested with `cloud-build-local`.

PR #767, opened 2020-08-13
Use composite primary key for DomainHistory

hstonec:
Change to use a composite key of `domainRepoId` and `historyRevisionId` as the primary key for `DomainHistory` entity. The main reason behind this is we found that there are many `HistoryEntry` entities in Datastore having same value for its `id` field(becomes `historyRevisionId` column in Cloud SQL), which means we cannot migrate those entities to Cloud SQL if we only use `historyRevisionId` as the primary key. Given that the key of `HistoryEntry` in Datastore is actually a combination of `repoId` and `id`, and it is unique, we decided to make a similar change in the Cloud SQL schema to use those two fields as the composite primary key to avoid key conflict.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/767)


CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/767)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @hstonec)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 930 at r3](https://reviewable.io/reviews/google/nomulus/767#-MG5N9c96mKYNq5BvOJk:-MGFlI2QDaKF-oUJOpy3:b-2qs7f4) ([raw file](https://github.com/google/nomulus/blob/c9330b57a32dfcb1740623592e8dd2f709528e78/db/src/main/resources/sql/schema/nomulus.golden.sql#L930)):*

That is the default value. It seems no harm to increase it so I set it to 30.
</blockquote></details>

Or you could just not specify it, if that is indeed the default value?  At some point we are going to have to harmonize all of the sequences (maybe even only use one, per my most recent proposal).
___
*[db/src/main/resources/sql/flyway/V51__use_composite_primary_key_for_domain_history_table.sql, line 16 at r3](https://reviewable.io/reviews/google/nomulus/767#-MG5NSiYAVaYqjge5V7c:-MGFl3_r2cBTu4o7OClL:bk78k3) ([raw file](https://github.com/google/nomulus/blob/c9330b57a32dfcb1740623592e8dd2f709528e78/db/src/main/resources/sql/flyway/V51__use_composite_primary_key_for_domain_history_table.sql#L16)):*

That is a good question. Hibernate definitely creates a multi-column index for `(domain_repo_id, history_revision_id)` because they are the composite primary key right now. Since `domain_repo_id` is the leading column in the index, the index should be used when only `domain_repo_id` is used in the where clause. 
</blockquote></details>

At some point we'll use the query optimizer and catch all such errors in this class if there end up being any.  So hopefully this does actually work as envisioned.

CydeWeys (approved):
:lgtm:

It's too bad this change is being made for what essentially amounts to legacy data concerns, as I don't think it's necessarily ideal from a schema perspective, but fixing up all that data somehow does seem harder.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/767#-:-MGFlXfJ5vdLOZPWn7M5:b-av3hpa)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @hstonec)

PR #770, opened 2020-08-17
Hex dump the ICANN report response content for debugging.

jianglai:
For example if the content is a byte array of "Hello", the Hex dump will
be "48656c6c6f".

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/770)

### DIFF
--- a/core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java
+++ b/core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java
@@ -94,7 +94,7 @@ public class IcannHttpReporter {
         response.getContent().close();
       }
       logger.atInfo().log(
-          "Received response code %d with content %s\n\nResponse content in hex: %s",
+          "Received response code %d with content: %s\n\nResponse content in hex: %s",
           response.getStatusCode(),
           new String(content, UTF_8),
           BaseEncoding.base16().encode(content));
--- a/core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java
+++ b/core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java
@@ -28,6 +28,7 @@ import com.google.api.client.http.HttpTransport;
 import com.google.common.base.Ascii;
 import com.google.common.base.Splitter;
 import com.google.common.flogger.FluentLogger;
+import com.google.common.io.BaseEncoding;
 import com.google.common.io.ByteStreams;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.keyring.api.KeyModule.Key;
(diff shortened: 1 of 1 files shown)


jianglai:
> > Reviewed 14 of 23 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/770)_ status: 14 of 23 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)
> 
> _[dependencies.gradle, line 87 at r1](https://reviewable.io/reviews/google/nomulus/770#-MExolHCEu6ukfhz3AEW:-MExolHD2efc4NHOc80W:b-prefo6) ([raw file](https://github.com/google/nomulus/blob/2067e9e289d1e3828cc68e7c34c3c0a029561309/dependencies.gradle#L87)):_
> 
> > ```groovy-gradle
> >       'com.sun.xml.bind:jaxb-xjc:2.2.11',
> >       'com.thoughtworks.qdox:qdox:1.12.1',
> >       'commons-codec:commons-codec:1.14',
> > ```
> 
> It looks like you can do this with guava's com.google.common.io.BaseEncoding class. Does that work for you?

I think hex dump is easier/more straightforward to help identify the encoding used. It's not like we are concerned with space saving here. Any particular advantage with using base64?

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/770)_ status: 14 of 23 files reviewed, 1 unresolved discussion (waiting on @jianglai)
> 
> _[dependencies.gradle, line 87 at r1](https://reviewable.io/reviews/google/nomulus/770#-MExolHCEu6ukfhz3AEW:-MExxpGD-Axon9mBtUwZ:bj1ueec) ([raw file](https://github.com/google/nomulus/blob/2067e9e289d1e3828cc68e7c34c3c0a029561309/dependencies.gradle#L87)):_
> 
> Only to avoid pulling in the additional dependency.
> To be clear, I'm talking about doing this: https://stackoverflow.com/a/36477814
> 
> (basically just using guava to to a hex dump) Doesn't that do the same thing?

You are right. My brain short-circuited for a while when I saw base16.

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/770)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mindhog)
> 
> _[dependencies.gradle, line 87 at r1](https://reviewable.io/reviews/google/nomulus/770#-MExolHCEu6ukfhz3AEW:-MF0ok0S1jpKbKG2aHdS:bi01461) ([raw file](https://github.com/google/nomulus/blob/2067e9e289d1e3828cc68e7c34c3c0a029561309/dependencies.gradle#L87)):_
> 
> There's a difference in being a transitive dependency vs a direct dependency though.

Yes, my point being that we didn't save any space by not directly using it. I initially used it because the first SO post I found referred to it.

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/770)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mindhog)
> 
> _[dependencies.gradle, line 87 at r1](https://reviewable.io/reviews/google/nomulus/770#-MExolHCEu6ukfhz3AEW:-MF0q5EH55Du4VNHWlHw:b-dvqiqv) ([raw file](https://github.com/google/nomulus/blob/2067e9e289d1e3828cc68e7c34c3c0a029561309/dependencies.gradle#L87)):_
> 
> But it does add more noise to lockfiles, and the bigger those are the more unwieldy everything becomes. Anyway, you gotta look past just the first StackOverflow response!

Not to be pedantic, but transitive dependencies are locked in the lockfiles as well.

PR #775, opened 2020-08-20
Fix cascade issue for GracePeriod

hstonec:
This PR fixed the cascade issues when loading/saving `GracePeriod` entities through `Domain`:

1. We need to add `insertable = false, updatable = false` to `@JoinColumn` annotation attached to `Set<GracePeriod>`.
The exact reason is not clear but many Hibernate users encountered the same problem(See [here](https://stackoverflow.com/questions/27309075/hibernate-not-deleting-the-foreign-key-row-rather-setting-it-as-null)).
Without setting these two options to false, Hibernate would try to update the foreign key to null instead of executing
the deletion when an element is removed from the collection.

2. For the sake of `Datastore`, the builder of `EppResource` sets any empty collection to null. However, Hibernate would
complain about this because it expects a non-null collection. We have a few options to solve this problem and I applied the
one to make minimum change, which is to just set the null collection to be an empty collection after the new entity is created.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/775)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -81,6 +81,9 @@ public class DomainBase extends DomainContent
     return super.nsHosts;
   }
 
+  // Note that we need to set `insertable = false, updatable = false` for @JoinColumn, otherwise
+  // Hibernate would try to set the foreign key to null(through an UPDATE TABLE sql) instead of
+  // deleting the whole entry from the table when the GracePeriod is removed from the set.
   @Access(AccessType.PROPERTY)
   @OneToMany(
       cascade = {CascadeType.ALL},
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -14,7 +14,6 @@
 
 package google.registry.model.domain;
 
-
 import com.googlecode.objectify.Key;
 import google.registry.model.EppResource;
 import google.registry.model.EppResource.ForeignKeyedEppResource;
@@ -87,7 +86,11 @@ public class DomainBase extends DomainContent
       cascade = {CascadeType.ALL},
       fetch = FetchType.EAGER,
       orphanRemoval = true)
-  @JoinColumn(name = "domainRepoId", referencedColumnName = "repoId")
+  @JoinColumn(
+      name = "domainRepoId",
+      referencedColumnName = "repoId",
+      insertable = false,
+      updatable = false)
   @SuppressWarnings("UnusedMethod")
   private Set<GracePeriod> getInternalGracePeriods() {
     return gracePeriods;
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -81,9 +81,16 @@ public class DomainBase extends DomainContent
     return super.nsHosts;
   }
 
-  // Note that we need to set `insertable = false, updatable = false` for @JoinColumn, otherwise
-  // Hibernate would try to set the foreign key to null(through an UPDATE TABLE sql) instead of
-  // deleting the whole entry from the table when the GracePeriod is removed from the set.
+  /**
+   * Returns the set of {@link GracePeriod} associated with the domain.
+   *
+   * <p>This is the getter method specific for Hibernate to access the field so it is set to
+   * private. The caller can use the public {@link #getGracePeriods()} to get the grace periods.
+   *
+   * <p>Note that we need to set `insertable = false, updatable = false` for @JoinColumn, otherwise
+   * Hibernate would try to set the foreign key to null(through an UPDATE TABLE sql) instead of
+   * deleting the whole entry from the table when the {@link GracePeriod} is removed from the set.
+   */
   @Access(AccessType.PROPERTY)
   @OneToMany(
       cascade = {CascadeType.ALL},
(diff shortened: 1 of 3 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/775)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/775#-MFGcCyPALQ7VSimM0T3:-MGFNuSv1iKT3gceKXjK:bk3exqw) ([raw file](https://github.com/google/nomulus/blob/7cd7d602be30f13d2051269cba600363f714b5c8/core/src/main/java/google/registry/model/domain/DomainContent.java#L294)):*

I just removed the `@OnSave` method to keep the immutability.

Encoding is one thing given that `GracePeriod` has 7 fields and some of them is nullable. Also I think we will lose the foreign key constraint for some of the field like `registrarId`, `domainRepoId`.

After some cursory search I don't find anything about how to query array through HQL or JPQL, but I think since we have been using the postgresql dialect we should be able to use the query predicated provided by postgresql.
</blockquote></details>

We could deal with encoding the nulls, but I don't think we could support the kind of queryability we'd want (I don't think PSQL queries will let us query on a partial value, and there's a portability case for sticking to HQL/JPQL if possible) and it sounds like the foreign key constraints are a deal-breaker, in any case.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/775)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 706 at r3](https://reviewable.io/reviews/google/nomulus/775#-MGFkS_K5PGC0AHd2_Gc:-MGFkS_K5PGC0AHd2_Gd:b-k6hyj4) ([raw file](https://github.com/google/nomulus/blob/6297885687335469c94f428ae3f1b378ab2b0197/core/src/main/java/google/registry/model/domain/DomainContent.java#L706)):*
> ```Java
>       // operations and orphan removal.
>       newDomain.gracePeriods =
>           newDomain.gracePeriods == null ? ImmutableSet.of() : newDomain.gracePeriods;
> ```

Use `nullToEmptyImmutableSet` for this.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/775)* status: all files reviewed, 4 unresolved discussions (waiting on @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 84 at r3](https://reviewable.io/reviews/google/nomulus/775#-MGFkdYB5FMdUqHVQDGY:-MGFkdYB5FMdUqHVQDGZ:b-i3iv10) ([raw file](https://github.com/google/nomulus/blob/6297885687335469c94f428ae3f1b378ab2b0197/core/src/main/java/google/registry/model/domain/DomainBase.java#L84)):*
> ```Java
>   }
> 
>   // Note that we need to set `insertable = false, updatable = false` for @JoinColumn, otherwise
> ```

You may as well just write a Javadoc at this point.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/775#-:-MGFkhc9A-YGlD126d9U:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @hstonec, @mindhog, and @weiminyu)

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/775)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 84 at r3](https://reviewable.io/reviews/google/nomulus/775#-MGFkdYB5FMdUqHVQDGY:-MGFxFvjEdyAclN9BTxL:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6297885687335469c94f428ae3f1b378ab2b0197/core/src/main/java/google/registry/model/domain/DomainBase.java#L84)):*

You may as well just write a Javadoc at this point.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/775#-MFGcCyPALQ7VSimM0T3:-MGFxHKU5Y9ZuD3xruhz:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7cd7d602be30f13d2051269cba600363f714b5c8/core/src/main/java/google/registry/model/domain/DomainContent.java#L294)):*

We could deal with encoding the nulls, but I don't think we could support the kind of queryability we'd want (I don't think PSQL queries will let us query on a partial value, and there's a portability case for sticking to HQL/JPQL if possible) and it sounds like the foreign key constraints are a deal-breaker, in any case.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 706 at r3](https://reviewable.io/reviews/google/nomulus/775#-MGFkS_K5PGC0AHd2_Gc:-MGFxkqjAXZlAHEKfra7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6297885687335469c94f428ae3f1b378ab2b0197/core/src/main/java/google/registry/model/domain/DomainContent.java#L706)):*

Use `nullToEmptyImmutableSet` for this.
</blockquote></details>

Done.

PR #776, opened 2020-08-21
Add Vkeys for PremiumList in Registry Objects

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/776)

### DIFF
--- a/core/src/main/java/google/registry/export/ExportPremiumTermsAction.java
+++ b/core/src/main/java/google/registry/export/ExportPremiumTermsAction.java
@@ -113,7 +113,7 @@ public class ExportPremiumTermsAction implements Runnable {
           "Skipping premium terms export for TLD %s because Drive folder isn't specified", tld);
       return Optional.of("Skipping export because no Drive folder is associated with this TLD");
     }
-    if (registry.getPremiumList() == null) {
+    if (registry.getPremiumListName().isEmpty()) {
       logger.atInfo().log("No premium terms to export for TLD %s", tld);
       return Optional.of("No premium lists configured");
     }
@@ -137,8 +137,7 @@ public class ExportPremiumTermsAction implements Runnable {
   }
 
   private String getFormattedPremiumTerms(Registry registry) {
-    Optional<PremiumList> premiumList =
-        PremiumList.getCached(registry.getPremiumList().getOfyKey().getName());
+    Optional<PremiumList> premiumList = PremiumList.getCached(registry.getPremiumListName().get());
     checkState(premiumList.isPresent(), "Could not load premium list for " + tld);
     SortedSet<String> premiumTerms =
         Streams.stream(loadPremiumListEntries(premiumList.get()))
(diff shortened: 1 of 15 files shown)


mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/776)* status: 11 of 13 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 321 at r2](https://reviewable.io/reviews/google/nomulus/776#-MFbC9aY1qf8NwLsZ8tf:-MGFJGDF2SuCmSJmZydz:bsbd9e6) ([raw file](https://github.com/google/nomulus/blob/008fbd43b2b9a5299998a037e71c32e5f4b16c80/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L321)):*

>Oh, I originally thought this was just an issue with ReservedList, but it looks like it is an issue with PremiumList too?

Yes, it is an issue for both.

>Are you suggesting I abandon this PR entirely and not add VKeys to the Registry class and instead use only converters?

Yeah, I think adding converters for existing Objectify key is probably a simpler solution in this case.

>When creating a converter for Key what SQL datatype would you suggest converting to?

In this case, it should just be a string because we only need to store the `name` of the list. So when we convert the `Key` to the column, we can just do `key.getName()`; when we convert the column to `Key` we can do `Key.create(getCrossTldKey(), ReservedList.class, name)`.

>Would it just be converting to a String of the ReservedList Name?

Yes.

>I am not sure how we would store or retrieve the RevisionId.

We actually don't need to worry about that for `Registry` entity, revisionId is managed by `ReservedListDao` and I don't think we need to expose it externally. `Registry` just needs to keep the `name` of its associated `ReservedList`(through `Set<Key<ReservedList>>` right now, but we can easily change it to `Set<String>` after the migration).
</blockquote></details>

Ah, I see.  I didn't realize that it was the revision Id we were missing.

To expand on @hstonec 's point: it is a general assumption for VKey that all key material (i.e. both sides of a symmetric key) can be reconstituted when the key is loaded from either database. [as I think about it, this may prove problematic for hierarchical datastore keys, but that's not relevant here]  Furthermore, there's a good case for not using VKeys at all when we're not using async replication: we don't want people to use these keys with the tm() methods, which will dispatch to whichever database is canonical depending on the replication mode. 

I also spoke to @cydeweys a little while ago about the need for some documentation on how we're doing the migration for entities where we're not doing the full async replication and on specifically which entities those are, so hopefully we'll get a clearer picture on this soon.

I agree that our best bet is to use a converter here.  The alternative is to either revert to datastore's name-based indexing until after the migration or to upgrade datastore to revision based indexing prior to the migration.

mindhog (commented):
> Reviewed 2 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/776)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @hstonec, and @mindhog)

PR #779, opened 2020-08-24
Fix AllocationTokenTest

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/779)


gbrodman:
Ah, makes sense.

PR #782, opened 2020-08-26
Fix flaky web driver test

weiminyu:
In RegistrarConsoleScreenshotTest's password setting test,
wait until the transaction has happened instead of a fixed
amount of time.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/782)


jianglai:
While this is definitely better, does it completely eliminate the flakiness? The fact that datastore is updated doesn't indicate that the response has been sent to the user agent and rendered, correct?

jianglai:
> We can add a short delay to account for rendering. By the time the query can start the mutation has already committed. Marshalling and wire time shouldn't be a big concern.
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/782#-:-MFkJju43xwUt6Iivjkv:b-6p35oy)_ status: 0 of 2 files reviewed, all discussions resolved (waiting on @CydeWeys and @jianglai)

I mean it's still trading one possibility of flakiness for another. Is there a way to do this deterministically? For example waiting on an HTML element to finish rendering? i. e. calling the `waitForElement` method?

PR #783, opened 2020-08-27
Change disable invoicing flag to enable invoicing flag

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/783)

### DIFF
--- a/core/src/main/java/google/registry/beam/invoicing/sql/billing_events.sql
+++ b/core/src/main/java/google/registry/beam/invoicing/sql/billing_events.sql
@@ -55,7 +55,7 @@ FROM (
     FROM
       `%PROJECT_ID%.%DATASTORE_EXPORT_DATA_SET%.%REGISTRY_TABLE%`
     WHERE
-      enableInvoicing = TRUE) ) AS BillingEvent
+      enableInvoicing IS TRUE) ) AS BillingEvent
   -- Gather billing ID from registrar table
   -- This is a 'JOIN' as opposed to 'LEFT JOIN' to filter out
   -- non-billable registrars
--- a/core/src/main/java/google/registry/beam/invoicing/sql/billing_events.sql
+++ b/core/src/main/java/google/registry/beam/invoicing/sql/billing_events.sql
@@ -55,9 +55,7 @@ FROM (
     FROM
       `%PROJECT_ID%.%DATASTORE_EXPORT_DATA_SET%.%REGISTRY_TABLE%`
     WHERE
-    -- TODO(b/18092292): Add a filter for tldState (not PDT/PREDELEGATION)
-      tldType = 'REAL'
-    AND disableInvoicing is not TRUE) ) AS BillingEvent
+      enableInvoicing = TRUE) ) AS BillingEvent
   -- Gather billing ID from registrar table
   -- This is a 'JOIN' as opposed to 'LEFT JOIN' to filter out
   -- non-billable registrars
--- a/core/src/main/java/google/registry/model/registry/Registry.java
+++ b/core/src/main/java/google/registry/model/registry/Registry.java
@@ -345,7 +345,7 @@ public class Registry extends ImmutableObject implements Buildable {
    * <p>Note that this boolean is the sole determiner on whether invoices should be generated for a
    * TLD. This applies to {@link TldType#TEST} TLDs as well.
    */
-  boolean enableInvoicing = false;
+  boolean invoicingEnabled = false;
 
   /**
    * A property that transitions to different TldStates at different times. Stored as a list of
@@ -646,8 +646,8 @@ public class Registry extends ImmutableObject implements Buildable {
       return this;
     }
 
-    public Builder setEnableInvoicing(boolean enableInvoicing) {
-      getInstance().enableInvoicing = enableInvoicing;
+    public Builder setInvoicingEnabled(boolean invoicingEnabled) {
+      getInstance().invoicingEnabled = invoicingEnabled;
       return this;
     }
 
--- a/core/src/main/java/google/registry/model/registry/Registry.java
+++ b/core/src/main/java/google/registry/model/registry/Registry.java
@@ -340,12 +340,12 @@ public class Registry extends ImmutableObject implements Buildable {
   TldType tldType = TldType.REAL;
 
   /**
-   * Whether to disable invoicing for a {@link TldType#REAL} TLD.
+   * Whether to enable invoicing for this TLD.
    *
-   * <p>Note that invoicing is always disabled for {@link TldType#TEST} TLDs. Setting this field has
-   * no effect for {@link TldType#TEST} TLDs.
+   * <p>Note that this boolean is the sole determiner on whether invoices should be generated for a
+   * TLD. This applies to {@link TldType#TEST} TLDs as well.
    */
-  boolean disableInvoicing = false;
+  boolean enableInvoicing = false;
 
   /**
    * A property that transitions to different TldStates at different times. Stored as a list of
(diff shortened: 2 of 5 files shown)


CydeWeys (changes requested):
What's the migration plan here?  Wait until after this is deployed, then use the tool to set it to true on just the launched open TLDs afterwards?  Maybe the SQL change should go into a subsequent release to the new field and tooling to set it so that we're not time-dependent?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/783#-:-MGFdjtg--j8vAgRTust:biqd72y)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 348 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFdyj63t_wIekeQDYW:-MGFdyj63t_wIekeQDYX:b3j4zak) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/main/java/google/registry/model/registry/Registry.java#L348)):*
> ```Java
>    * TLD. This applies to {@link TldType#TEST} TLDs as well.
>    */
>   boolean enableInvoicing = false;
> ```

Should be `invoicingEnabled` for consistency with other fields.
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFe3kZ96hq9oA6Eu3Y:-MGFe3kZ96hq9oA6Eu3Z:b-t4ept2) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java#L121)):*
> ```Java
>   @Nullable
>   @Parameter(
>       names = "--enable_invoicing",
> ```

`invoicing_enabled` works better here because `invoicing_enabled=FALSE` makes sense but `enable_invoicing=FALSE` is a bit awkward semantically.  Basically, "enabled" works as an adjective but "enable" doesn't work here as a verb.
___
*[core/src/test/resources/google/registry/beam/invoicing/billing_events_test.sql, line 58 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFdbLo-TBj668qcrpT:-MGFdbLo-TBj668qcrpU:b-5f4zrn) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/test/resources/google/registry/beam/invoicing/billing_events_test.sql#L58)):*
> ```TSQL
>       `my-project-id.latest_datastore_export.Registry`
>     WHERE
>       enableInvoicing = TRUE) ) AS BillingEvent
> ```

Hrm, how about `IS` instead of `=`?  I guess they're functionally identical, but `IS` reads cleaner.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 622 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFdWydAQeidzpEMCwq:-MGFdWydAQeidzpEMCwr:bhxafig) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/test/resources/google/registry/model/schema.txt#L622)):*
> ```
>   @Parent com.googlecode.objectify.Key<google.registry.model.common.EntityGroupRoot> parent;
>   boolean dnsPaused;
>   boolean enableInvoicing;
> ```

This should be `invoicingEnabled` for consistency with the rest of these.

jianglai (commented):
I think it's fine to do it in one PR -- unless we want to be super conservative and keep both the fields during the transition and then remove `disableInvoicing`, as soon as the field name is changed the SQL is already broken. We just sent this month's invoice so we still have a long time to go before the next invoice is due.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/783#-:-MGG-nMV5jn6tE1vJXxU:bwbu4k4)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 348 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFdyj63t_wIekeQDYW:-MGG1zhx98H7MlzhOnur:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/main/java/google/registry/model/registry/Registry.java#L348)):*

Should be `invoicingEnabled` for consistency with other fields.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFe3kZ96hq9oA6Eu3Y:-MGG28R65W8L94u_ZKXY:b-49fc7z) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java#L121)):*

`invoicing_enabled` works better here because `invoicing_enabled=FALSE` makes sense but `enable_invoicing=FALSE` is a bit awkward semantically.  Basically, "enabled" works as an adjective but "enable" doesn't work here as a verb.
</blockquote></details>

Why doesn't "enable" in "enable invoicing" work as a verb?

Anyway changed it for consistency.
___
*[core/src/test/resources/google/registry/beam/invoicing/billing_events_test.sql, line 58 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFdbLo-TBj668qcrpT:-MGG6rAa3vCGhQUKo1Hc:bcxkd0u) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/test/resources/google/registry/beam/invoicing/billing_events_test.sql#L58)):*

Hrm, how about `IS` instead of `=`?  I guess they're functionally identical, but `IS` reads cleaner.
</blockquote></details>

Done. Previously we have to use `IS NOT true` because we need to include`null`. Here `IS` and `=` does the same thing for our purposes. 
___
*[core/src/test/resources/google/registry/model/schema.txt, line 622 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFdWydAQeidzpEMCwq:-MGG1ybT4SBvl-lfxsvl:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/test/resources/google/registry/model/schema.txt#L622)):*

This should be `invoicingEnabled` for consistency with the rest of these.
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/783)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/783#-MGFe3kZ96hq9oA6Eu3Y:-MGoM-T679nSpWhKOoc-:b-r542y8) ([raw file](https://github.com/google/nomulus/blob/8e88c43e6a7d150b651541d37aa50d5975cd8991/core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java#L121)):*

Why doesn't "enable" in "enable invoicing" work as a verb?

Anyway changed it for consistency.
</blockquote></details>

Because "enable invoicing" (as a verb) implies that it's actually being, you know, enabled, not disabled, yet what `enable_invoicing=FALSE` is doing is disabling, not invoicing.  If you want to use the verb it should be `--enable_invoicing` and `--disable_invoicing`, but that's probably worse because it's two separate flags that aren't sorted immediately next to each other in the alphabetical parameters list.  So the best compromise is `--invoicing_enabled=TRUE` and `--invoicing_enabled=FALSE`, which both work.  You can set an enabled status to true or false, but you can't enable something false; the verb for that is disable.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/783#-:-MGoMseK6ClhOxthMWAe:bnfp4nl)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @jianglai, and @weiminyu)

PR #790, opened 2020-09-02
Allow explicitly for null EPP resources in History objects

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/790)


CydeWeys (commented):
I don't understand how it's remotely possible to have a history entry that doesn't have an associated EPP resource, given the current Datastore schema.  This sounds like a data error that needs to be fixed, not an explicitly supported use case that should be allowed.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MGJukdXAoAShiJ2xvOH:bgtv0b0)* status: 0 of 13 files reviewed, all discussions resolved

gbrodman (commented):
Legacy objects

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MGJvvCVAdNWCfL6TqTq:b-umd04e)* status: 0 of 13 files reviewed, all discussions resolved

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[core/src/test/java/google/registry/model/history/ContactHistoryTest.java, line 73 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGjRU8p4lpl_iw7JxsU:-MGjRU8qB1KBnenHL1TQ:bxgn3dd) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/core/src/test/java/google/registry/model/history/ContactHistoryTest.java#L73)):*
> ```
> createContactHistory
> ```
Could you add a migration test for each of the three leaf history types:
- Create a legacy HistoryEntry instance with Objectify,
- Downcast it to a leaf history instance
- Store it in SQL.

CydeWeys (commented):
I need a lot more detail than that.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MGjVN5BCE3aOqLTeoOO:bfntdt4)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

gbrodman (commented):
Ben, this is related to the conversations that we had over email and in "person" about how we will treat the legacy history objects. Given that we don't want to have a migration task to fill the DomainContent/ContactBase/HostBase fields in legacy HistoryEntry objects when we shift over to using SQL as the primary database (and indeed it might not be possible to do so cleanly) then we must require that the EPP resource field in the new *History objects be nullable. When transitioning over, we will have these fields be null. 

Note that the "foreign key" repo ID will not be null, but rather the object fields itself. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MGjdwVL2SOJuWgX-tGT:b7ivzz3)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/model/history/ContactHistoryTest.java, line 73 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGjRU8p4lpl_iw7JxsU:-MGjec0U-taUdyY2HPEk:bebrt1w) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/core/src/test/java/google/registry/model/history/ContactHistoryTest.java#L73)):*

> ```
> createContactHistory
> ```
Could you add a migration test for each of the three leaf history types:
- Create a legacy HistoryEntry instance with Objectify,
- Downcast it to a leaf history instance
- Store it in SQL.
</blockquote></details>

Will do

CydeWeys (changes requested):
I'm talking about knowing what the parent EPP resource is at all, not knowing the full field contents of it.  I don't see how it's possible that `host_repo_id` or `contact_repo_id` could ever be null in the history tables.  See below.  There should be no migration task necessary because we must already know these as they're embedded in the very key of every single HistoryEntry already.  The data isn't missing.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MGmt_QGDI7ofr_Sixdp:b-lp5f2d)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 21 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtA5U8pGM5jCExyQ7:-MGmtA5U8pGM5jCExyQ8:b4bggsd) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L21)):*
> ```TSQL
> ALTER TABLE "ContactHistory" ALTER COLUMN creation_time DROP NOT NULL;
> ALTER TABLE "ContactHistory" ALTER COLUMN current_sponsor_registrar_id DROP NOT NULL;
> ALTER TABLE "ContactHistory" ALTER COLUMN contact_repo_id DROP NOT NULL;
> ```

Why is `contact_repo_id` allowed to be null?  How can there be a contact history row that doesn't know what contact it's associated with?  The repo id is embedded in the very key of the history entry itself in Datastore, so it's not possible that we wouldn't know this information.  This shouldn't be a "legacy problem".
___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 26 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtMUBAktjv7plSC7V:-MGmtMUBAktjv7plSC7W:b-cbvshx) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L26)):*
> ```TSQL
> ALTER TABLE "DomainHistory" ALTER COLUMN creation_registrar_id DROP NOT NULL;
> ALTER TABLE "DomainHistory" ALTER COLUMN creation_time DROP NOT NULL;
> ALTER TABLE "DomainHistory" ALTER COLUMN current_sponsor_registrar_id DROP NOT NULL;
> ```

Note how the `domain_repo_id` is NOT being made nullable here, which is what I'd expect.
___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 33 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtLdB1dcXJJhBVpf1:-MGmtLdB1dcXJJhBVpf2:b-r7rrqx) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L33)):*
> ```TSQL
> ALTER TABLE "HostHistory" ALTER COLUMN current_sponsor_registrar_id DROP NOT NULL;
> ALTER TABLE "HostHistory" ALTER COLUMN history_reason DROP NOT NULL;
> ALTER TABLE "HostHistory" ALTER COLUMN host_repo_id DROP NOT NULL;
> ```

Ditto for `host_repo_id`; how can it possibly be null?

gbrodman (commented):
Yes, the repo_id field should be always non-null (see the other comments). The concept of this PR is talking about the contents of the referenced object. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MGnZRRV5Wg2-Q2WFfpd:bs7l3o4)* status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 21 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtA5U8pGM5jCExyQ7:-MGnYj4V2Jwp9_hLcLq7:b9ukb0r) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L21)):*

Why is `contact_repo_id` allowed to be null?  How can there be a contact history row that doesn't know what contact it's associated with?  The repo id is embedded in the very key of the history entry itself in Datastore, so it's not possible that we wouldn't know this information.  This shouldn't be a "legacy problem".
</blockquote></details>

Yes, it should be non-null. FYI: I wasn't sure this was ready for review which is why I did not add you on this PR and why this part wasn't symmetrical / proper yet. 
___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 26 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtMUBAktjv7plSC7V:-MGnYoxZ18HTC6q1_nsd:b1hiebr) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L26)):*

Note how the `domain_repo_id` is NOT being made nullable here, which is what I'd expect.
</blockquote></details>

it should be non-null
___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 33 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtLdB1dcXJJhBVpf1:-MGnYqg21aIWAAHmX75t:b9ukb0r) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L33)):*

Ditto for `host_repo_id`; how can it possibly be null?
</blockquote></details>

Yes, it should be non-null. FYI: I wasn't sure this was ready for review which is why I did not add you on this PR and why this part wasn't symmetrical / proper yet.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[db/src/main/resources/sql/flyway/V52__null_content.sql, line 21 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGmtA5U8pGM5jCExyQ7:-MGoOcTx9cwggbZrumi8:bquctuy) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/db/src/main/resources/sql/flyway/V52__null_content.sql#L21)):*

Yes, it should be non-null. FYI: I wasn't sure this was ready for review which is why I did not add you on this PR and why this part wasn't symmetrical / proper yet. 
</blockquote></details>

Oh.  Whoops.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MGoOz8C0gS0NlHyzreQ:b-rgrba6) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*
> ```
>        repo_id text not null,
>         update_timestamp timestamptz,
>         creation_registrar_id text not null,
> ```

They should still be `NOT NULL` on the Domain/Host/Contact tables though.  It's only on the *History tables where they're allowed to be NULL, no?  Isn't there some way of accomplishing this?

(Or is this just an artifact of the PR still not being fully ready for review yet?)
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 398 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOo8z8jmpOmeis9xC:-MGoOo8z8jmpOmeis9xD:b70bmn0) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/db/src/main/resources/sql/schema/db-schema.sql.generated#L398)):*
> ```
>     );
> 
>     create table "HostResource" (
> ```

Why is this table named `HostResource` instead of `Host`??  I know it's not part of this PR, but did we screw up something big and fail to realize it?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MGoQbby0idj0o8iMoX-:b6bk5u3) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

They should still be `NOT NULL` on the Domain/Host/Contact tables though.  It's only on the *History tables where they're allowed to be NULL, no?  Isn't there some way of accomplishing this?

(Or is this just an artifact of the PR still not being fully ready for review yet?)
</blockquote></details>

You are correct, they should be NOT NULL and in the Flyway files / golden file, they are NOT NULL. 

However:

Because we pull, for instance, the creationRegistrarId field from the DomainContent class,  the generated files treat it exactly the same in both places that it shows up (in `Domain` and `DomainHistory`). Obviously it should never be null in `Domain` (and we keep that restriction in the golden) but because the we can't guarantee the existence of the `DomainContent` object stored within the `DomainHistory`, it <can> be null there. 

The generation isn't intelligent to know "this field can be null in some usages but not others"
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 398 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOo8z8jmpOmeis9xC:-MGoR_PF7rXysJar9kqe:be74xuy) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/db/src/main/resources/sql/schema/db-schema.sql.generated#L398)):*

Why is this table named `HostResource` instead of `Host`??  I know it's not part of this PR, but did we screw up something big and fail to realize it?
</blockquote></details>

I don't think I created it (maybe I did?) but I assume it was just an artifact of having the same name as the Java class because no one thought to change it.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MGoSEO73Tk47EG4XenS:b-bjh2jq) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

You are correct, they should be NOT NULL and in the Flyway files / golden file, they are NOT NULL. 

However:

Because we pull, for instance, the creationRegistrarId field from the DomainContent class,  the generated files treat it exactly the same in both places that it shows up (in `Domain` and `DomainHistory`). Obviously it should never be null in `Domain` (and we keep that restriction in the golden) but because the we can't guarantee the existence of the `DomainContent` object stored within the `DomainHistory`, it <can> be null there. 

The generation isn't intelligent to know "this field can be null in some usages but not others"


</blockquote></details>

*it can be null there, in the second to last sentence

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxYd2E0rUmUDz4itMd:-MGxYd2E0rUmUDz4itMe:b-cuc6sl) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L77)):*
> ```Java
>   /** The ID of the registrar that is currently sponsoring this resource. */
>   @Index
>   // This would be non-null if not for legacy History objects that have null resource fields
> ```

Put these in the Javadoc comments.  As-is they're not being exposed to other parts of the codebase in IDE hover-over tooltips.

As a general rule, if you're writing a comment specifically about a field, method, or class, it should be in a Javadoc comment.  The `//` style comments are for inline explanations inside functions about how code is actually working, and TODOs.  And you definitely shouldn't be mixing in inline `//` comments when there's already a `/**` Javadoc.
___
*[core/src/main/java/google/registry/model/EppResource.java, line 97 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxZqud1EjzymEELjg0:-MGxZqud1EjzymEELjg1:bkwzj2i) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L97)):*
> ```Java
> 
>   /** The time when this resource was created. */
>   // Map the method to XML, not the field, because if we map the field (with an adaptor class) it
> ```

Ditto.

Not gonna individually comment on the rest of these, but you get the gist.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_KNpCRI_kghJD50O:-MGx_KNpCRI_kghJD50P:bddbn) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/contact/ContactHistory.java#L68)):*
> ```Java
>   }
> 
>   /** The state of the {@link ContactBase} object at this point in time. */
> ```

Needs explanation in the Javadoc of what circumstances this would be absent for.  (I'm assuming pre-migration but I had to think about it for a bit.)

Also worth clarifying that it's not doing any projection, it's just returning the actual stored data with no modifications.  Because typically when we say "at a point in time" we're talking about taking the data currently stored in the DB and then projecting it to that specific time, but that's not what's going on here.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 80 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_hSf2bpCA4Y72QDo:-MGx_hSf2bpCA4Y72QDp:b-64g3ru) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/contact/ContactHistory.java#L80)):*
> ```Java
>   @PostLoad
>   void postLoad() {
>     // Normally Hibernate would see that the contact fields are all null and would fill contactBase
> ```

Fyi this feels like a code behavior explanation that can remain inline.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_zOO5KUjJTtGg66P:-MGx_zOP8vqjG_0pzZF9:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/domain/DomainHistory.java#L94)):*
> ```Java
>   }
> 
>   /** The state of the {@link DomainContent} object at this point in time. */
> ```

Ditto.
___
*[core/src/main/java/google/registry/model/host/HostHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxa1YkEAYM71P4yAHJ:-MGxa1YkEAYM71P4yAHK:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/host/HostHistory.java#L68)):*
> ```Java
>   }
> 
>   /** The state of the {@link HostBase} object at this point in time. */
> ```

Ditto.
___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 37 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxa3O076M9FgvuwPwS:-MGxa3O076M9FgvuwPwT:byxlumu) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L37)):*
> ```Java
>   public Timestamp convertToDatabaseColumn(@Nullable CreateAutoTimestamp entity) {
>     if (entity == null) {
>       return null;
> ```

Where's the `@Nullable` annotation on the return type?
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MGxauXn9IwmwlDp8gLJ:b2gmabg) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

*it can be null there, in the second to last sentence 
</blockquote></details>

Maybe we should make this generation smarter then?  It seems unexpected and wrong to me that the generated schema is different from the golden schema in such fundamental ways as whether columns are required to have values or not.  If they don't even agree on that, then what's even the purpose of having this generated file?
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 398 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOo8z8jmpOmeis9xC:-MGxakmtAVL2t56tVH-_:b-1jsxyg) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/db/src/main/resources/sql/schema/db-schema.sql.generated#L398)):*

I don't think I created it (maybe I did?) but I assume it was just an artifact of having the same name as the Java class because no one thought to change it. 
</blockquote></details>

Mike's gonna fix it.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 14 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxYd2E0rUmUDz4itMd:-MHIIDU9E501Jms52AHt:b-erhz9x) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L77)):*

Put these in the Javadoc comments.  As-is they're not being exposed to other parts of the codebase in IDE hover-over tooltips.

As a general rule, if you're writing a comment specifically about a field, method, or class, it should be in a Javadoc comment.  The `//` style comments are for inline explanations inside functions about how code is actually working, and TODOs.  And you definitely shouldn't be mixing in inline `//` comments when there's already a `/**` Javadoc.
</blockquote></details>

Done -- the reason why I did it previously was that it seemed like an implementation detail not relevant to anyone looking to reference this code from elsewhere in the codebase. 
___
*[core/src/main/java/google/registry/model/EppResource.java, line 97 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxZqud1EjzymEELjg0:-MHIINLy-B6oQRTzGs_t:b-ft0myk) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L97)):*

Ditto.

Not gonna individually comment on the rest of these, but you get the gist.
</blockquote></details>

hey, this was jart's code :p 
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_KNpCRI_kghJD50O:-MHILQtn5xK3HiutvKeD:b8mffxx) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/contact/ContactHistory.java#L68)):*

Needs explanation in the Javadoc of what circumstances this would be absent for.  (I'm assuming pre-migration but I had to think about it for a bit.)

Also worth clarifying that it's not doing any projection, it's just returning the actual stored data with no modifications.  Because typically when we say "at a point in time" we're talking about taking the data currently stored in the DB and then projecting it to that specific time, but that's not what's going on here.
</blockquote></details>

good point, clarified 
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_zOO5KUjJTtGg66P:-MHILd5D-Ph97Ry90er1:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/domain/DomainHistory.java#L94)):*

Ditto.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/host/HostHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxa1YkEAYM71P4yAHJ:-MHILdnH5k-Gxiog0KQK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/host/HostHistory.java#L68)):*

Ditto.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 37 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxa3O076M9FgvuwPwS:-MHILk-YDJCTXG7pJoub:b-6l3oe5) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L37)):*

Where's the `@Nullable` annotation on the return type?
</blockquote></details>

Done (this is auto-used anyway fortunately)
___
*[core/src/test/java/google/registry/model/history/ContactHistoryTest.java, line 73 at r1](https://reviewable.io/reviews/google/nomulus/790#-MGjRU8p4lpl_iw7JxsU:-MHILtjX9MlinQ37DCpk:bbe6jcf) ([raw file](https://github.com/google/nomulus/blob/4ed0d39298a7af50d981efd46f4b2710d4138c42/core/src/test/java/google/registry/model/history/ContactHistoryTest.java#L73)):*

Will do
</blockquote></details>

This should be all set to go now, in LegacyHistoryObjectTest

It might be a bit overkill to have a separate test for each of the three classes but it's probably better to have that than too less. 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MHIMXUf89Vcw0v82TbP:b-t45z8s) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

Maybe we should make this generation smarter then?  It seems unexpected and wrong to me that the generated schema is different from the golden schema in such fundamental ways as whether columns are required to have values or not.  If they don't even agree on that, then what's even the purpose of having this generated file?
</blockquote></details>

I think it serves as a sanity check to make sure that what we've coded in Java converts roughly to what we expect it to in SQL (and vice versa).

I'm not sure it's feasible to make the generation smarter. For one, we just hook in to the built-in 
Hibernate SchemaExport class (in our HibernateSchemaExporter class). For two, I'm not sure there's any way to annotate a Java object saying "this should be nullable in X table but non-null in Y table".

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 14 files reviewed, 7 unresolved discussions (waiting on @gbrodman and @weiminyu)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxYd2E0rUmUDz4itMd:-MHSxGlL7mjcsqAwXDj3:b-8dyw6v) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L77)):*

Done -- the reason why I did it previously was that it seemed like an implementation detail not relevant to anyone looking to reference this code from elsewhere in the codebase. 
</blockquote></details>

It's kind of relevant to other parts of the codebase, in that you're explicitly making it clear that this can have a null value (and why).  Otherwise one might assume that this would always be present.

You could phrase it a little bit more clearly to serve this purpose, e.g.:

```
  /**
   * The ID of the registrar that is currently sponsoring this resource.
   *
   * <p>This can be null in the case of pre-Registry-3.0-migration history objects with null resource fields.
   */
```

Same for the rest of these below.
___
*[core/src/main/java/google/registry/model/EppResource.java, line 97 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxZqud1EjzymEELjg0:-MHSxsaR0UxbugffOYbd:b-8xgsj6) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L97)):*

hey, this was jart's code :p 
</blockquote></details>

Wow that's going way back huh.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_KNpCRI_kghJD50O:-MHSyBFCBHjh3ypDEziZ:b-mae3gu) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/contact/ContactHistory.java#L68)):*

good point, clarified 
</blockquote></details>

Maybe say something like `The values of all the fields on the {@link ContactBase} object after the action represented by this history object was executed.`
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 70 at r3](https://reviewable.io/reviews/google/nomulus/790#-MHSy2bi18yKfVLmMe2q:-MHSy2bi18yKfVLmMe2r:bjg7yxq) ([raw file](https://github.com/google/nomulus/blob/7106fd8284bdf6eee0ae6f09f6bdc36cdc3e8fc4/core/src/main/java/google/registry/model/contact/ContactHistory.java#L70)):*
> ```Java
>    * The {@link ContactBase} object at the point in time this history object was created.
>    *
>    * <p>May be absent for objects created prior to the SQL migration.
> ```

Nit, but:  is this merely a "may", or is it a "most likely", or is it a certainty?
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_zOO5KUjJTtGg66P:-MHSyuobDqkLwNkWIedI:b2nreh6) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/domain/DomainHistory.java#L94)):*

Done.
</blockquote></details>

Same as above.
___
*[core/src/main/java/google/registry/model/host/HostHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxa1YkEAYM71P4yAHJ:-MHSyx4DAoOeqAbHGGPe:bio604c) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/host/HostHistory.java#L68)):*

Done.
</blockquote></details>

And again.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MHSzS8hEsHyWLYU3EPI:b-x1vxy9) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

I think it serves as a sanity check to make sure that what we've coded in Java converts roughly to what we expect it to in SQL (and vice versa).

I'm not sure it's feasible to make the generation smarter. For one, we just hook in to the built-in 
Hibernate SchemaExport class (in our HibernateSchemaExporter class). For two, I'm not sure there's any way to annotate a Java object saying "this should be nullable in X table but non-null in Y table". 

</blockquote></details>

But why is the `NOT NULL`ness of these columns correctly being determined for the golden SQL file but not for the generated file?  Are the NOT NULL restrictions only added by Flyway scripts, which the golden SQL file is summing up cumulatively to arrive at a golden, whereas this generated one isn't that smart?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 14 files reviewed, 7 unresolved discussions (waiting on @gbrodman and @weiminyu)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MHT-SA18bAxks90Divi:biocetp) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

But why is the `NOT NULL`ness of these columns correctly being determined for the golden SQL file but not for the generated file?  Are the NOT NULL restrictions only added by Flyway scripts, which the golden SQL file is summing up cumulatively to arrive at a golden, whereas this generated one isn't that smart?
</blockquote></details>

Is there any way to fix this by making the fields themselves ignored, and then having two separate getters in the respective places, with each one correctly using the Hibernate annotation to mark it nullable or not?  It worries me that if you just take the Java code at face value and generate a schema from it, you get the wrong result!  That proves that there's state that exists beyond just the Java code itself, which isn't what we want.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @transient, and @weiminyu)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGxYd2E0rUmUDz4itMd:-MHT7GVc6VE015fQjGZ7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/EppResource.java#L77)):*

It's kind of relevant to other parts of the codebase, in that you're explicitly making it clear that this can have a null value (and why).  Otherwise one might assume that this would always be present.

You could phrase it a little bit more clearly to serve this purpose, e.g.:

```
  /**
   * The ID of the registrar that is currently sponsoring this resource.
   *
   * <p>This can be null in the case of pre-Registry-3.0-migration history objects with null resource fields.
   */
```

Same for the rest of these below.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 68 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGx_KNpCRI_kghJD50O:-MHT6FIrB7pSKwb2lJrZ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9088a9365627b44671e4951de32e964c1e943224/core/src/main/java/google/registry/model/contact/ContactHistory.java#L68)):*

Maybe say something like `The values of all the fields on the {@link ContactBase} object after the action represented by this history object was executed.`
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 70 at r3](https://reviewable.io/reviews/google/nomulus/790#-MHSy2bi18yKfVLmMe2q:-MHT6NQtBYQz3j4zLmO5:b-2k1xib) ([raw file](https://github.com/google/nomulus/blob/7106fd8284bdf6eee0ae6f09f6bdc36cdc3e8fc4/core/src/main/java/google/registry/model/contact/ContactHistory.java#L70)):*

Nit, but:  is this merely a "may", or is it a "most likely", or is it a certainty?
</blockquote></details>

It's a "Will". Changed.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MHT7ke_1r64K4nRgXZ-:b-hrfpdf) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

Is there any way to fix this by making the fields themselves ignored, and then having two separate getters in the respective places, with each one correctly using the Hibernate annotation to mark it nullable or not?  It worries me that if you just take the Java code at face value and generate a schema from it, you get the wrong result!  That proves that there's state that exists beyond just the Java code itself, which isn't what we want.
</blockquote></details>

To answer the question in your first comment, you are correct. The `nomulus.golden.sql` file is created by essentially summing up all the VXX_*.sql Flyway scripts. 

We could theoretically do what you suggest in the second comment, I suppose. It'd be messy though, and as we add more special cases we begin to lose the benefits of embedding the EppResource inside the history object at all. 

It'd require adding a @Transient here, then adding 

```
  @Column(nullable = false, name = "creationTime")
  @Access(AccessType.PROPERTY)
  public CreateAutoTimestamp getCreationTimestamp() {
    return CreateAutoTimestamp.create(getCreationTime());
  }
  
  private void setCreationTime(CreateAutoTimestamp timestamp) {
    // actually set the timestamp here, which we can't do currently since it's immutable iirc
  }
```

in every class -- and that's just for the creation timestamp. We'd need to duplicate that for other fields as well

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MHXH22J6763tF9PF7Dw:b-79cfac) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

To answer the question in your first comment, you are correct. The `nomulus.golden.sql` file is created by essentially summing up all the VXX_*.sql Flyway scripts. 

We could theoretically do what you suggest in the second comment, I suppose. It'd be messy though, and as we add more special cases we begin to lose the benefits of embedding the EppResource inside the history object at all. 

It'd require adding a @Transient here, then adding 

```
  @Column(nullable = false, name = "creationTime")
  @Access(AccessType.PROPERTY)
  public CreateAutoTimestamp getCreationTimestamp() {
    return CreateAutoTimestamp.create(getCreationTime());
  }
  
  private void setCreationTime(CreateAutoTimestamp timestamp) {
    // actually set the timestamp here, which we can't do currently since it's immutable iirc
  }
```

in every class -- and that's just for the creation timestamp. We'd need to duplicate that for other fields as well 

</blockquote></details>

Ughhh.  3x that 3x times (for each EPP resource) would not be pleasant.  But the alternative, that our actual Java code as written does not properly reflect the actual schema either, is also not pleasant.  I guess as long as we have the golden file we're OK ...

We can leave this as-is for lack of better alternatives, but it's definitely quite sub-optimal, and I wish there were some better solution.  Maybe worth revisiting in the future.  Or maybe `NOT NULL` constraints on history tables just don't matter that much (also entirely possible).

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790#-:-MHXHfJI9cxUjsoF4OjM:bnfp4nl)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 241 at r2](https://reviewable.io/reviews/google/nomulus/790#-MGoOz8C0gS0NlHyzreP:-MHXq3GB9_IJIAOlHsGh:b2n9o79) ([raw file](https://github.com/google/nomulus/blob/cdf2c7f7cb14aaaae69d99dc9c15489c11ce2c20/db/src/main/resources/sql/schema/db-schema.sql.generated#L241)):*

Ughhh.  3x that 3x times (for each EPP resource) would not be pleasant.  But the alternative, that our actual Java code as written does not properly reflect the actual schema either, is also not pleasant.  I guess as long as we have the golden file we're OK ...

We can leave this as-is for lack of better alternatives, but it's definitely quite sub-optimal, and I wish there were some better solution.  Maybe worth revisiting in the future.  Or maybe `NOT NULL` constraints on history tables just don't matter that much (also entirely possible).
</blockquote></details>

yeah :/ it's not great but it's also not the only place where we do this. Many of the foreign key constraints that we define aren't autogenerated either, but rather are only defined in the Flyway files.

weiminyu (approved):
> Reviewed 2 of 13 files at r1, 3 of 9 files at r3, 9 of 9 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: :shipit: complete! all files reviewed, all discussions resolved

weiminyu (approved):
> Reviewed 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: :shipit: complete! all files reviewed, all discussions resolved

weiminyu (approved):
> Reviewed 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/790)* status: :shipit: complete! all files reviewed, all discussions resolved

gbrodman:
sorry for all the commits, just a lot of semantic merge conflicts + rebuilding based off a newly-not-broken master

PR #794, opened 2020-09-08
Add domain-specific history fields to DomainHistory objects

gbrodman:
We didn't include these at first (or in the HistoryEntry object) to make things simpler but we will want these fields to exist for DomainHistory objects (and only for them, since they're not applicable to contacts or hosts)

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/794)


gbrodman:
Ben, we punted on this before in order to lessen the size of previous PRs, but we need to add these in at some point. These three fields only exists in history entries corresponding to Domain changes. 

PTAL, thanks.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxU4KD70jci2WYdF7I:-MGxU4KD70jci2WYdF7J:b-6b91tl) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L92)):*
> ```Java
>           column = @Column(name = "historyPeriodValue"))
>   })
>   public Period getPeriod() {
> ```

Won't this often be null, because most domain operations don't have a period associated with them?  (E.g. updates.)  If so, this needs to return an `Optional`.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUF5YANjqJEJu6ROF:-MGxUF5YANjqJEJu6ROG:b-adakut) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L105)):*
> ```Java
>   @Access(AccessType.PROPERTY)
>   @Column(name = "historyOtherRegistrarId")
>   public String getOtherRegistrarId() {
> ```

Same; needs to be `Optional` because the vast majority of domain operations are not transfers.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUKuw00xhwR_OqyM_:-MGxUKuw00xhwR_OqyMa:b-181btj) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L112)):*
> ```Java
>    * Logging field for transaction reporting.
>    *
>    * <p>This will be empty for any HistoryEntry generated before this field was added.
> ```

Will it be empty on any history entries created post-migration?  (If so, which ones?)  Looking at the possible values of the `TransactionReportField` enum, it seems like these record billable events only, so updates and such won't have these?
___
*[core/src/main/java/google/registry/model/domain/Period.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUhrDEzVboMZHAbRg:-MGxUhrDEzVboMZHAbRh:b-4ul6x5) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/Period.java#L45)):*
> ```Java
>   }
> 
>   // for Hibernate
> ```

It's not an `ImmutableObject` anymore with these setters; why are they required again?  Can Hibernate use a constructor instead?  That at least would preserve immutability.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 258 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxVT4V7gcHm0IxxSPH:-MGxVT4WDl16criTUBpo:b-nkp3iv) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L258)):*
> ```Java
>   }
> 
>   // Hibernate needs this in order to populate the period but no one else should ever use it
> ```

Same comments on these setters re: immutability and whether we can replace this all with a constructor.

If this is really just the way that Hibernate has to work and we can't get around it, we need to have a larger conversation about what kind of immutability constraints we can actually guarantee (and maybe consider renaming `ImmutableObject`).
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 264 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxVmgE5Xnm91EweVhW:-MGxVmgE5Xnm91EweVhX:bu0kce) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L264)):*
> ```Java
>   }
> 
>   // Hibernate needs this in order to populate the other registrar ID
> ```

If these were Javadocs they'd be more accessible from other parts of the codebase in hover-over tooltips.  As is IDEs won't know these comments are specifically related to the methods.

If there really is no other way to do it than this, these comments should all be Javadocs explaining that this method exists solely to satisfy Hibernate and that we should never be using it, but rather, using the Builder/static create methods/constructor as applicable.  And include an `{@link ...}` to said construction method even.
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 20 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxX2jr6RcGNXgT9hYF:-MGxX2jr6RcGNXgT9hYG:bopamtr) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L20)):*
> ```TSQL
> ALTER TABLE "DomainHistory" ADD COLUMN history_period_value int4;
> 
> CREATE TABLE "DomainHistoryTransactionRecord" (
> ```

Do we actually need this table at all?  Do we ever go back and actually edit a previous transaction record, or do we only write out a new offsetting one?  I think it may be the latter.  So if transaction records are truly immutable as I suspect they might be (in the same way that one-time billing events are), then we don't need a history table for them at all.
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 27 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWmRKBUQxHPT8Fb_q:-MGxWmRKBUQxHPT8Fb_r:btzpzw0) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L27)):*
> ```TSQL
> );
> 
> CREATE TABLE  "DomainTransactionRecord" (
> ```

This is pretty long ... how about `DomainTxRecord`?
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 28 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWpWR-E1SATtAgy48:-MGxWpWR-E1SATtAgy49:b2deay5) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L28)):*
> ```TSQL
> 
> CREATE TABLE  "DomainTransactionRecord" (
>    id bigserial NOT NULL,
> ```

`domain_tx_record_id`?
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 29 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxXMEt-WAEOlVriYwd:-MGxXMEt-WAEOlVriYwe:bxtmcl0) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L29)):*
> ```TSQL
> CREATE TABLE  "DomainTransactionRecord" (
>    id bigserial NOT NULL,
>     report_amount int4 NOT NULL,
> ```

This could definitely be a smaller int type because I believe the only valid values for this field are ints in the range of -10 through 10.

But honestly we may not want to worry about such micro levels of DB optimization.  And we already have lots of other int4s (e.g. on the `DomainHistory` table for similarly small values).
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 534 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxY-ZP6fgK9JIecG7t:-MGxY-ZP6fgK9JIecG7u:b-i5d2uh) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/schema/nomulus.golden.sql#L534)):*
> ```TSQL
> 
> CREATE SEQUENCE public."DomainTransactionRecord_id_seq"
>     START WITH 1
> ```

It occurs to me that all of these sequences really need to be starting at large values (larger than the largest Datastore id we currently have).  But at the same time, I know that we already have a lot of existing sequences in the schema that also start at 1.  Is it a lot of effort to update a sequence's starting range?  If so, we may just want to get this right to begin with on all of these.  It merits broader discussion.  I know Shicong was looking into this recently too.
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 538 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxXusi4_pCY9CrL2Eu:-MGxXusi4_pCY9CrL2Ev:b-osxh3q) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/schema/nomulus.golden.sql#L538)):*
> ```TSQL
>     NO MINVALUE
>     NO MAXVALUE
>     CACHE 1;
> ```

Why a cache of 1?  Is this just the default or did you explicitly pick this value?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxU4KD70jci2WYdF7I:-MGxgBmyELK-k4rrtlrV:b-2ef46c) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L92)):*

Won't this often be null, because most domain operations don't have a period associated with them?  (E.g. updates.)  If so, this needs to return an `Optional`.
</blockquote></details>

Yes, however this getter is not for the sake of anyone actually using it but rather for the DB, which doesn't know how to handle Optionals. 
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUF5YANjqJEJu6ROF:-MGxgw5R52pCdkqdkAyt:b-jp5bfn) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L105)):*

Same; needs to be `Optional` because the vast majority of domain operations are not transfers.
</blockquote></details>

see other comment
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUKuw00xhwR_OqyM_:-MGxhN8eDjZ0IiTeehVi:b5fv5s3) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L112)):*

Will it be empty on any history entries created post-migration?  (If so, which ones?)  Looking at the possible values of the `TransactionReportField` enum, it seems like these record billable events only, so updates and such won't have these?
</blockquote></details>

Yes, I believe that's the case. This comment is actually referring to the original addition of this field which was some time in the past (before my time). 
___
*[core/src/main/java/google/registry/model/domain/Period.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUhrDEzVboMZHAbRg:-MGxe1y61u6b9uQj6mMp:b-7u8648) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/Period.java#L45)):*

It's not an `ImmutableObject` anymore with these setters; why are they required again?  Can Hibernate use a constructor instead?  That at least would preserve immutability.
</blockquote></details>

They're required because Hibernate fills out these fields reflectively after object creation. 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 258 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxVT4V7gcHm0IxxSPH:-MGxhaw95o9vKzzPB48a:b-q42rvj) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L258)):*

Same comments on these setters re: immutability and whether we can replace this all with a constructor.

If this is really just the way that Hibernate has to work and we can't get around it, we need to have a larger conversation about what kind of immutability constraints we can actually guarantee (and maybe consider renaming `ImmutableObject`).
</blockquote></details>

It seems that anything that Hibernate accesses via a getter method requires a corresponding (private, if you wish) setter method. 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 264 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxVmgE5Xnm91EweVhW:-MGxmCs02WEcusyMdL1m:b3x68nn) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L264)):*

If these were Javadocs they'd be more accessible from other parts of the codebase in hover-over tooltips.  As is IDEs won't know these comments are specifically related to the methods.

If there really is no other way to do it than this, these comments should all be Javadocs explaining that this method exists solely to satisfy Hibernate and that we should never be using it, but rather, using the Builder/static create methods/constructor as applicable.  And include an `{@link ...}` to said construction method even.
</blockquote></details>

Good call, I added those as javadoc lines. 
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 20 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxX2jr6RcGNXgT9hYF:-MGxkUSBClzgst11BRPF:babqzs6) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L20)):*

Do we actually need this table at all?  Do we ever go back and actually edit a previous transaction record, or do we only write out a new offsetting one?  I think it may be the latter.  So if transaction records are truly immutable as I suspect they might be (in the same way that one-time billing events are), then we don't need a history table for them at all.
</blockquote></details>

Yeah, we need to store these as part of the DomainHistory objects since we refer to them in a few places as part of business logic, e.g. when deleting or transferring a domain (example: DomainFlowUtils::createCancelingRecords)
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 27 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWmRKBUQxHPT8Fb_q:-MGxigvq9EOU35M2zk0R:b4dcdwo) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L27)):*

This is pretty long ... how about `DomainTxRecord`?
</blockquote></details>

Is it? I don't see it as a problem. 
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 28 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWpWR-E1SATtAgy48:-MGxjxyn2X51lQp1vBpg:b-hmvhcf) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L28)):*

`domain_tx_record_id`?
</blockquote></details>

Why? We usually just use `id` or `revision_id` elsewhere and if we use a longer name here, it makes the foreign key references incredibly unwieldy 
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 29 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxXMEt-WAEOlVriYwd:-MGxenjhBw2SUAUqtFk8:b-2nsg5a) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L29)):*

This could definitely be a smaller int type because I believe the only valid values for this field are ints in the range of -10 through 10.

But honestly we may not want to worry about such micro levels of DB optimization.  And we already have lots of other int4s (e.g. on the `DomainHistory` table for similarly small values).
</blockquote></details>

Yeah, I don't think this is worth worrying about since there's also value in keeping the handwritten SQL code the same as the generated SQL code whenever possible 
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 534 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxY-ZP6fgK9JIecG7t:-MGxe__Z2IPGoW56q8nl:bm1hizi) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/schema/nomulus.golden.sql#L534)):*

It occurs to me that all of these sequences really need to be starting at large values (larger than the largest Datastore id we currently have).  But at the same time, I know that we already have a lot of existing sequences in the schema that also start at 1.  Is it a lot of effort to update a sequence's starting range?  If so, we may just want to get this right to begin with on all of these.  It merits broader discussion.  I know Shicong was looking into this recently too.
</blockquote></details>

I believe it is trivially easy to set the point at which a sequence should next start (and we may get rid of this anyway if we wish to have all ID generation read from the same sequence) 
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 538 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxXusi4_pCY9CrL2Eu:-MGxeXhAC3bfffOsZcz3:b-cr3kad) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/schema/nomulus.golden.sql#L538)):*

Why a cache of 1?  Is this just the default or did you explicitly pick this value?
</blockquote></details>

This is the default.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @hstonec)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxU4KD70jci2WYdF7I:-MHNM7Jl9UZRFMPx8knK:b-6s71jb) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L92)):*

Yes, however this getter is not for the sake of anyone actually using it but rather for the DB, which doesn't know how to handle Optionals. 
</blockquote></details>

Can we change the visibility to package level or even private if they are not expected to be used by the caller other than Hibernate?
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 20 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxX2jr6RcGNXgT9hYF:-MHNXTxzDcLncnXtiEai:b8p7m41) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L20)):*

Yeah, we need to store these as part of the DomainHistory objects since we refer to them in a few places as part of business logic, e.g. when deleting or transferring a domain (example: DomainFlowUtils::createCancelingRecords)
</blockquote></details>

It looks like a many-to-many relationship between `DomainHistory` and `DomainTransactionRecord` so I think we probably need this join table to avoid duplicate entries, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxU4KD70jci2WYdF7I:-MHNe-hKEshGCnBHwrdL:b-ql2rrt) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L92)):*

Can we change the visibility to package level or even private if they are not expected to be used by the caller other than Hibernate?
</blockquote></details>

No, because the superclass (HistoryEntry) has the method as public, and you can't reduce the visibility of an overriden method 
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 20 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxX2jr6RcGNXgT9hYF:-MHNd_HX4jrmws8Xvj2y:bs06ewl) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L20)):*

It looks like a many-to-many relationship between `DomainHistory` and `DomainTransactionRecord` so I think we probably need this join table to avoid duplicate entries, right?
</blockquote></details>

I think it's one (HistoryEntry) to many (DomainTransactionRecord), no? I still think we need the join table though.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 20 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxX2jr6RcGNXgT9hYF:-MHNfwqzCW9Wfwn3Jokp:bvvygdv) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L20)):*

I think it's one (HistoryEntry) to many (DomainTransactionRecord), no? I still think we need the join table though. 
</blockquote></details>

If it is a one-to-many mapping, can we just add the foreign key to `DomainTransactionRecord` then use that for the join?

I just checked the use case for `DomainFlowUtils::createCancelingRecords` and it looks like it creates a new set of `DomainTransactionRecord` every time, so at least for that it is a one-to-many mapping.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 20 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxX2jr6RcGNXgT9hYF:-MHSJ9Jv6-X4egArIxNx:b-5vbtm6) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L20)):*

If it is a one-to-many mapping, can we just add the foreign key to `DomainTransactionRecord` then use that for the join?

I just checked the use case for `DomainFlowUtils::createCancelingRecords` and it looks like it creates a new set of `DomainTransactionRecord` every time, so at least for that it is a one-to-many mapping. 
</blockquote></details>

Yep, one-to-many. I realized that we can do this by specifying the @JoinColum annotations (multiple)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxU4KD70jci2WYdF7I:-MHSt4792-benBrjPaWp:bsjjgnw) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L92)):*

No, because the superclass (HistoryEntry) has the method as public, and you can't reduce the visibility of an overriden method 
</blockquote></details>

OK but this is still returning something that is nullable yet is not tagged with `@Nullable`.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUF5YANjqJEJu6ROF:-MHStJPv7to9zUC1sNWW:b-by9q0f) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L105)):*

see other comment
</blockquote></details>

See other comment.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUKuw00xhwR_OqyM_:-MHStOD61jqzdsPALsrA:ba5bqj0) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L112)):*

Yes, I believe that's the case. This comment is actually referring to the original addition of this field which was some time in the past (before my time). 
</blockquote></details>

Can you fix the comment then so it's correct?
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 534 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxY-ZP6fgK9JIecG7t:-MHSuTUxCfRoMHIEA24Y:bg7ithg) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/schema/nomulus.golden.sql#L534)):*

I believe it is trivially easy to set the point at which a sequence should next start (and we may get rid of this anyway if we wish to have all ID generation read from the same sequence) 
</blockquote></details>

Sounds like we're gonna be getting rid of these anyway.
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 27 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWmRKBUQxHPT8Fb_q:-MHSu5uv0XAXpsvB0sdC:bl14sah) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L27)):*

Is it? I don't see it as a problem. 
</blockquote></details>

Seems kinda long to me.  `DomainTransactionRecord` might well be the longest table name in our entire schema; if not the actual longest, definitely close to it.  `DomainTxRecord` is definitely shorter.  Do you not like the Tx abbreviation or something?
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 28 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWpWR-E1SATtAgy48:-MHStqNL0y8kPZljg5qk:bwlsgnd) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L28)):*

Why? We usually just use `id` or `revision_id` elsewhere and if we use a longer name here, it makes the foreign key references incredibly unwieldy 
</blockquote></details>

I think generally we don't actually just use short non-specific column names in other tables?  Do you have a count?  We should standardize this either way.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxU4KD70jci2WYdF7I:-MHSwMhG7qJ8pLjmAkXG:b6genq8) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L92)):*

OK but this is still returning something that is nullable yet is not tagged with `@Nullable`.
</blockquote></details>

Done, also added a nullable on getOtherClientId()
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUF5YANjqJEJu6ROF:-MHSwRDvBjRuTAijrUWU:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L105)):*

See other comment.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxUKuw00xhwR_OqyM_:-MHSxBh4Az7hRge6Yxbn:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/core/src/main/java/google/registry/model/domain/DomainHistory.java#L112)):*

Can you fix the comment then so it's correct?
</blockquote></details>

Done.
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 27 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWmRKBUQxHPT8Fb_q:-MHSxfm3Dq2gOK5BqwVz:b3nppil) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L27)):*

Seems kinda long to me.  `DomainTransactionRecord` might well be the longest table name in our entire schema; if not the actual longest, definitely close to it.  `DomainTxRecord` is definitely shorter.  Do you not like the Tx abbreviation or something?
</blockquote></details>

It is 23 characters; the longest existing table is `BillingCancellation` at 19. 

I just don't see a need for abbreviations in many cases; it's not like Java or SQL is particularly concerned with its verbosity. It's also the same name as the Java class, which is good for discoverability and means we don't need to manually specify a name.
___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 28 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWpWR-E1SATtAgy48:-MHSzGAX0XChXC131X9w:bpgvlch) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L28)):*

I think generally we don't actually just use short non-specific column names in other tables?  Do you have a count?  We should standardize this either way.
</blockquote></details>

It seems like we're roughly evenly split between 1. `id` 2. `revision_id`, and 3. `<some_descriptor>_id`, e.g. `billing_event_id`

We should definitely standardize it. Would you ever refer to this field outside of the namespace of the table? I'm not sure if `"DomainTransactionRecord".id` or `"DomainTxRecord".id` are worse than `"DomainTransactionRecord".domain_tx_record_id`or `"DomainTxRecord".domain_tx_record_id`

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794)* status: 0 of 9 files reviewed, all discussions resolved

___
*[db/src/main/resources/sql/flyway/V52__domain_history_fields.sql, line 28 at r1](https://reviewable.io/reviews/google/nomulus/794#-MGxWpWR-E1SATtAgy48:-MHWu9YADbRMq67hAmkA:bp4gbt5) ([raw file](https://github.com/google/nomulus/blob/3da3f7c4725cac701374f5dc32da04493656fc2a/db/src/main/resources/sql/flyway/V52__domain_history_fields.sql#L28)):*

It seems like we're roughly evenly split between 1. `id` 2. `revision_id`, and 3. `<some_descriptor>_id`, e.g. `billing_event_id`

We should definitely standardize it. Would you ever refer to this field outside of the namespace of the table? I'm not sure if `"DomainTransactionRecord".id` or `"DomainTxRecord".id` are worse than `"DomainTransactionRecord".domain_tx_record_id`or `"DomainTxRecord".domain_tx_record_id`
</blockquote></details>

We can leave this as-is for now, and then consider if we want to embark on some larger standardization scheme.  There are cases in which having unique names for all ID columns can be useful in SELECTs on JOINed tables, but I don't know how to weigh that benefit against the cost of having longer ids in all cases.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/794#-:-MHWzH3N5sEaLDANR8VM:bnfp4nl)* status: 0 of 9 files reviewed, all discussions resolved

PR #795, opened 2020-09-09
Add and use temp_history_id_sequence to avoid release error

hstonec:
We changed `history_id_sequence` to use `INCREMENT BY 50` in #767. However, this change cannot be applied correctly by our release process. Our release process always deploy the code change first then apply the schema change, after the code change is deployed, Hibernate throws the an exception that indicates it expects the sequence to be `INCREMENT BY 50` in the database  as well though at that time the schema change is not applied yet.

So, the work around I made in this PR is to create a temporary sequence for all history tables and also keep the previous change for `history_id_sequence`. After this PR is deployed, we can create another PR to switch back to use `history_id_sequence` for all history tables.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/795)

### DIFF
--- a/core/src/main/java/google/registry/model/contact/ContactHistory.java
+++ b/core/src/main/java/google/registry/model/contact/ContactHistory.java
@@ -53,7 +53,7 @@ public class ContactHistory extends HistoryEntry {
   VKey<ContactResource> contactRepoId;
 
   @Id
-  @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "HistorySequenceGenerator")
+  @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "TempHistorySequenceGenerator")
   @Column(name = "historyRevisionId")
   @Access(AccessType.PROPERTY)
   @Override
--- a/core/src/main/java/google/registry/model/domain/DomainHistory.java
+++ b/core/src/main/java/google/registry/model/domain/DomainHistory.java
@@ -76,7 +76,7 @@ public class DomainHistory extends HistoryEntry {
   Set<VKey<HostResource>> nsHosts;
 
   @Id
-  @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "HistorySequenceGenerator")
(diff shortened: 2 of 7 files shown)


CydeWeys (commented):
Ouch.  This is kinda messy.  I guess there isn't a better way to migrate over the DB schema though, not just as a fix for this particular issue, but in the general case?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/795#-:-MGoe_3VB1iBVU4l5u2t:b-7yn72i)* status: 0 of 7 files reviewed, all discussions resolved

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/795#-:-MGoefd9-QYN2LVl-FfE:bnfp4nl)* status: 0 of 7 files reviewed, all discussions resolved

hstonec (commented):
Yeah, though I'd hope this is just a special case. Actually, every time when we create a new table/column/sequence in the sql database we will run into the same situation, i.e. our code starts using the new thing while the schema is not deployed yet. Hibernate didn't throw any exception before(or we didn't notice?) but this time it did. My guess is Hibernate checked any change to existing sequence when it initializes `EntityManagerFactory` so it failed this time, though it didn't check creation of new table/column/sequence until the corresponding Java entity is actually used and most of our changes for schema is not used in production code path.

Anyway, I will test this PR in alpha to make sure it actually works.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/795#-:-MGohIEf05GWbjLRLkdU:b-262k16)* status: 0 of 7 files reviewed, all discussions resolved

hstonec (commented):
Tested this PR in crash before the schema in #767 is deployed and it worked; then I deployed the schema until this PR and tested again, it also worked.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/795#-:-MGxxyJiAyVFfBDs8-BT:broybvy)* status: 0 of 7 files reviewed, all discussions resolved

PR #800, opened 2020-09-11
Create a separate per-tld registry lock/unlock cost

gbrodman:
Currently we use the standard server status change cost for this, but
this might not be ideal at some point in the future if we wish to allow
manual forced updates outside of the standard registry lock system (we
would charge for these manual forced updates, even if we don't charge
for registry locks).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/800)

### DIFF
--- a/core/src/main/java/google/registry/model/registry/Registry.java
+++ b/core/src/main/java/google/registry/model/registry/Registry.java
@@ -116,6 +116,7 @@ public class Registry extends ImmutableObject implements Buildable {
   public static final Money DEFAULT_RENEW_BILLING_COST = Money.of(USD, 8);
   public static final Money DEFAULT_RESTORE_BILLING_COST = Money.of(USD, 100);
   public static final Money DEFAULT_SERVER_STATUS_CHANGE_BILLING_COST = Money.of(USD, 20);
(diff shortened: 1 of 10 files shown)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/800#-:-MHSx6Tv4ZJW33KTK3Wl:bnfp4nl)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 880 at r1](https://reviewable.io/reviews/google/nomulus/800#-MHSwkZR0P3BDqxh2cWX:-MHSwkZR0P3BDqxh2cWY:bgjxhgs) ([raw file](https://github.com/google/nomulus/blob/adf2b878a08eaa7b236dd843376d58c09c398c40/core/src/main/java/google/registry/model/registry/Registry.java#L880)):*
> ```Java
> 
>     public Builder setRegistryLockOrUnlockBillingCost(Money amount) {
>       checkArgument(amount.isPositiveOrZero(), "Registry lock/unlock cost cannot be negative.");
> ```

Super nit: exception messages don't have periods at the end of them

PR #808, opened 2020-09-17
Update ICANN activity reporting logging

jianglai:
1. It appears that when we have a 200 response, the response content is
   garbled, but we don't care since we know the request is successful.
   When we have a 400 response, the response is indeed UTF-8 encoded.
   Print the stack trace of the HTTP exception doesn't help anymore.

2. The result code is a complex type which includes the xml element with
   all its attributes, which don't care. We only want to print the
   value.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/808)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/808)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #814, opened 2020-09-22
Convert flyway scripts into a single file

mindhog:
Concatenate the flyway scripts into a single large file.  This has the
advantage of forcing a merge conflict if more than one person adds a script to
the same base revision, requiring that the change be merged manually.  This
also relieves us of the tedious responsibility of managing sequence numbers.

As part of this, also update run_schema_check.sh to ensure that we only ever
append to the existing concatenated file once it has been applied to sandbox.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/814)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 5 unresolved discussions (waiting on @mindhog and @weiminyu)

___
*[db/src/main/resources/sql/flyway.dat, line 1 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrk4cV6rc-5FfkaOIk:-MHrk4cV6rc-5FfkaOIl:b-2thpmf) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L1)):*
> ```
> Copyright 2020 The Nomulus Authors. All Rights Reserved.
> ```

Why is this file named .dat instead of .sql?
___
*[db/src/main/resources/sql/flyway.dat, line 31 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrk91Z3SxxvaDzSi2H:-MHrk91Z3SxxvaDzSi2I:ba2lpbp) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L31)):*
> ```
> Text in this file is ignored up to the first #file line below.
> 
> #file create_claims_list_and_entry
> ```

We're throwing away the version numbers?  Shouldn't we retain those?  Otherwise it'll be really hard to know what version any given change actually is, but knowing those version #s is important; that's how we version our schema.
___
*[db/src/main/resources/sql/flyway.dat, line 32 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrjpjAEQDTxF_JVFLl:-MHrjpjAEQDTxF_JVFLm:b9mslk9) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L32)):*
> ```
> 
> #file create_claims_list_and_entry
> -- Copyright 2019 The Nomulus Authors. All Rights Reserved.
> ```

I don't see a need to repeat the license header over and over.  It adds a lot of noise and makes it hard to read this file.
___
*[db/src/main/resources/sql/flyway.dat, line 62 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrjv4Q3fvgKOgpfBQx:-MHrjv4Q3fvgKOgpfBQy:be28eop) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L62)):*
> ```
>        add constraint FK6sc6at5hedffc0nhdcab6ivuq
>        foreign key (revision_id) 
>        references "ClaimsList";
> ```

There should be two blank lines before each individual entry header for readability reasons.  Right now it's hard to visually distinguish anything.
___
*[db/src/main/resources/sql/flyway.dat, line 895 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrkQsD7Vfg_rmy2QUu:-MHrkQsD7Vfg_rmy2QUv:b-rzcmgl) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L895)):*
> ```
>    foreign key (client_id)
>    references "Registrar";
> #file create_pollmessage
> ```

Why do all these headers start with `#file`?  How about `#version` or similar?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[db/src/main/resources/sql/flyway.dat, line 1 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrk4cV6rc-5FfkaOIk:-MHrllLK0NXrX_Q5Z48T:b56efls) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L1)):*

Why is this file named .dat instead of .sql?
</blockquote></details>

TBH, I wasn't sure what I wanted to name it.  It's not a pure sql file, I wouldn't want anyone to think that it was legal to use in that context.  sql.source, maybe?
___
*[db/src/main/resources/sql/flyway.dat, line 31 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrk91Z3SxxvaDzSi2H:-MHrm3ICCC0kX_iTW2U4:b5aiuqe) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L31)):*

We're throwing away the version numbers?  Shouldn't we retain those?  Otherwise it'll be really hard to know what version any given change actually is, but knowing those version #s is important; that's how we version our schema.
</blockquote></details>

We're not throwing them away, we're letting the script generate them.  This way merging just becomes a matter of ensuring the correct order.  I wasn't aware that these were used for anything but ordering.  Can you describe the use-case you're envisioning?

In any case, the sequence numbers can be determined from the resulting output files.  If we need to surface them for other purposes, let's consider how.  I really never want to have to manage a sequence number here again.
___
*[db/src/main/resources/sql/flyway.dat, line 32 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrjpjAEQDTxF_JVFLl:-MHrkqVb4XhojHRglENd:bfmr7if) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L32)):*

I don't see a need to repeat the license header over and over.  It adds a lot of noise and makes it hard to read this file.
</blockquote></details>

I'd love to remove them, but depending on the machinery of flyway, we may need these files to be identical to those we have already applied.  @weiminyu recommended this as well.
___
*[db/src/main/resources/sql/flyway.dat, line 62 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrjv4Q3fvgKOgpfBQx:-MHrl6YcDrsPEcGU0cNn:b-xfyj1v) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L62)):*

There should be two blank lines before each individual entry header for readability reasons.  Right now it's hard to visually distinguish anything.
</blockquote></details>

It complicates the extraction to do so.  Right now we can just process and write a line at a time.

Would it help to introduce a line of the form "#============================"?  That we can more easily do, and it should improve the distinction significantly.
___
*[db/src/main/resources/sql/flyway.dat, line 895 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrkQsD7Vfg_rmy2QUu:-MHrmhsT0ubevjlCEFBP:b-nsa8ke) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L895)):*

Why do all these headers start with `#file`?  How about `#version` or similar?
</blockquote></details>

We can use whatever we like :-)  If we're going to revisit it, I'll point out that @weiminyu  has advised me that there are two other kinds of files we may ultimately wish to accomodate: "R" files (which are not versioned and can be replayed multiple times) and "U" files (which undo other changes).

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[db/src/main/resources/sql/flyway.dat, line 895 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrkQsD7Vfg_rmy2QUu:-MHuqwIp8LwvGq8XGiHJ:b-ceplm0) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L895)):*

We can use whatever we like :-)  If we're going to revisit it, I'll point out that @weiminyu  has advised me that there are two other kinds of files we may ultimately wish to accomodate: "R" files (which are not versioned and can be replayed multiple times) and "U" files (which undo other changes).
</blockquote></details>

I should also add the the motivation for "#file" was that this keyword introduces the contents of a generated file, so I don't think it's a bad choice for the name.

CydeWeys (changes requested):
I have to say, I like the separate file approach better.  This just isn't working for me.  The one huge combined file is too hard to deal with.  I'd prefer that we stick with the separate files, one per revision, and then just have a generated file listing them to prevent merge conflicts.  It's clear from how this looks that we're hacking up Flyway and using it in a way that it isn't intended to work.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814#-:-MHv6OeuDLncBZ7rbmBP:b3k6okk)* status: 0 of 61 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/flyway.dat, line 31 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrk91Z3SxxvaDzSi2H:-MHv5WgCE6-6D1tFXO83:bveddnx) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L31)):*

We're not throwing them away, we're letting the script generate them.  This way merging just becomes a matter of ensuring the correct order.  I wasn't aware that these were used for anything but ordering.  Can you describe the use-case you're envisioning?

In any case, the sequence numbers can be determined from the resulting output files.  If we need to surface them for other purposes, let's consider how.  I really never want to have to manage a sequence number here again.
</blockquote></details>

Weimin can speak to the importance of the version numbers, but that's how we revision the DB.  "The codebase is on V45, but the prod DB is only on V42 right now and it'll be updated in the next deployment."  If we get rid of the version numbers entirely then we have no way to keep track of things.  The version numbers will still actually be there, but hidden, only in the generated files, and if something goes wrong when deploying a specific version script you won't even know how to find it in this file.
___
*[db/src/main/resources/sql/flyway.dat, line 62 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrjv4Q3fvgKOgpfBQx:-MHv67sM55PIVa_usXDe:b-9a6vc3) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L62)):*

It complicates the extraction to do so.  Right now we can just process and write a line at a time.

Would it help to introduce a line of the form "#============================"?  That we can more easily do, and it should improve the distinction significantly.
</blockquote></details>

Yeah something like that would definitely help to visually distinguish it.

Would blank lines *really* complicate extraction though?  That seems easy to strip out.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[db/src/main/resources/sql/flyway.dat, line 31 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrk91Z3SxxvaDzSi2H:-MHvMtd22iaF5ljRTV2B:brai3k6) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L31)):*

Weimin can speak to the importance of the version numbers, but that's how we revision the DB.  "The codebase is on V45, but the prod DB is only on V42 right now and it'll be updated in the next deployment."  If we get rid of the version numbers entirely then we have no way to keep track of things.  The version numbers will still actually be there, but hidden, only in the generated files, and if something goes wrong when deploying a specific version script you won't even know how to find it in this file.
</blockquote></details>

It's not clear to me how this would currently work in a debugging scenario.  Though I will point out that the generated files are not hidden in any sense: they're simply in a different directory.  I think they're also what gets deployed into the schema jar file?
___
*[db/src/main/resources/sql/flyway.dat, line 62 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHrjv4Q3fvgKOgpfBQx:-MHvDkBzE6sTPR9OB531:bd4x2zc) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/db/src/main/resources/sql/flyway.dat#L62)):*

Yeah something like that would definitely help to visually distinguish it.

Would blank lines *really* complicate extraction though?  That seems easy to strip out.
</blockquote></details>

It would be more complicated.  We'd have to collect all of the lines and then strip the two trailing ones, whereas now we can simply write a line to the current file.  We can't simply strip all trailing lines, either, because we have one case of a file that has a single trailing line that must be preserved.  And finally we'd need to either ensure that all new entries have two trailing lines or add an exception for the last file (probably the latter, as requiring folks to add two blank lines to the end of the file for a new entry would be extremely weird).
This might be somewhat palatable in python or java code, but it's currently implemented in gradle which is already difficult to comprehend.  We could conceivably move it out of the build.gradle file, but that has its own complications (for a java implementation, we'd need to add a new processor to buildSrc, for python we'd have to shell out to the script and introduce a build-time dependency on python).

gbrodman (commented):
Copying comment from email:

```
What is the point of adding them all to one single, large file? It doesn't seem to me like this adds anything over having the files be distinct, at the cost of increasing file size and generation complexity (since we split them up into a bunch of files later). 

I think the existing system works well for versioning, with the exception of merge conflicts, which can be fixed by generating / diffing a list of the Flyway files. The versioning might get a little bit silly as we get into the larger numbers, but I don't really see that as a problem and as Ben points out in the PR, it would be nice to easily be able to say that a particular DB is at version N. 
```

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814#-:-MHvZHRW89ohwCSH17uL:btlnd05)* status: 0 of 61 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @weiminyu)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[integration/run_schema_check.sh, line 69 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHxeumm8kW77NfeJYyA:-MHxeumm8kW77NfeJYyB:b-gh7xca) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/integration/run_schema_check.sh#L69)):*
> ```
> if git diff ${sandbox_tag} db/src/main/resources/sql/flyway.dat | \
> ```
It may be better to wait until my go/r3pr/815 is submitted.

I rewrote the change detection test. It can reliably verify that the generated flyway scripts are the same as the deleted ones.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[integration/run_schema_check.sh, line 69 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHxgDfTAq4YBZ5VeVIJ:-MHxgDfTAq4YBZ5VeVIK:b-gt3ozi) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/integration/run_schema_check.sh#L69)):*
> ```
> ${sandbox_tag}
> ```
I don't know when it happened but Kokoro no longer fetches all git tags. 
That's the source of my question about this script to you on Wednesday: git diff failed but pipestatus is not reflected in the return value so presubmit test on kokoro passed. 

We need to fetch sandbox_tag explicitly in our script.

weiminyu (changes requested):
With this all-in-one file, we ensure that version numbers are monotonously increasing without gap.
It is also easier to browse the history. We also don't have a golden filename list to edit. 

Downside is we don't know what the highest version number is in the database.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814#-:-MHxh68TAxMtrSzObhHa:b2og67w)* status: 0 of 61 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[integration/run_schema_check.sh, line 69 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHxgDfTAq4YBZ5VeVIJ:-MHxhx-c6kU6nFFav0td:bhrc8kc) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/integration/run_schema_check.sh#L69)):*

> ```
> ${sandbox_tag}
> ```
I don't know when it happened but Kokoro no longer fetches all git tags. 
That's the source of my question about this script to you on Wednesday: git diff failed but pipestatus is not reflected in the return value so presubmit test on kokoro passed. 

We need to fetch sandbox_tag explicitly in our script.
</blockquote></details>

Could you  document what this does? I'm not good at awk.

Also, would it be sufficient to run this check against the master? If yes it could be made a unit test and included in the regular build.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[integration/run_schema_check.sh, line 69 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHxeumm8kW77NfeJYyA:-MI-FJxE7usTIHUo60Kc:b2h8ak0) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/integration/run_schema_check.sh#L69)):*

> ```
> if git diff ${sandbox_tag} db/src/main/resources/sql/flyway.dat | \
> ```
It may be better to wait until my go/r3pr/815 is submitted.

I rewrote the change detection test. It can reliably verify that the generated flyway scripts are the same as the deleted ones.
</blockquote></details>

Yep, I just reviewed it. I definitely like the new approach better.

I think I'm going to add a presubmit (either here or in another PR that implements the golden list instead) that verifies against master, this way we can run it locally and fail-fast on kokoro-foss.
___
*[integration/run_schema_check.sh, line 69 at r1](https://reviewable.io/reviews/google/nomulus/814#-MHxgDfTAq4YBZ5VeVIJ:-MI-GcxJEZ6VfA5i6Xgj:bs5wtll) ([raw file](https://github.com/google/nomulus/blob/4dc0bc7ebea4f83e08864a6556cb6399d3c630a4/integration/run_schema_check.sh#L69)):*

Could you  document what this does? I'm not good at awk.

Also, would it be sufficient to run this check against the master? If yes it could be made a unit test and included in the regular build. 
</blockquote></details>

Done.  It was worth documenting even for those who know awk.

Re running against master, my thoughts exactly :-)

mindhog (commented):
WRT comparing against the database, I suggest that we add a script that queries the flyway table and also parses this file ("grep '^#file'  | wc db/src/main/resources/sql/flyway.dat" basically gives us this information).  If we care about database version comparison, that's useful regardless of how we structure the flyway files. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814#-:-MI-K56QE6himP3FTTzA:b6n8bfs)* status: 0 of 61 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @weiminyu)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814)* status: 0 of 61 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[db/build.gradle, line 223 at r2](https://reviewable.io/reviews/google/nomulus/814#-MI-eWPG8HX3QSQ-6mJZ:-MI-eWPG8HX3QSQ-6mJ_:b-27v3v6) ([raw file](https://github.com/google/nomulus/blob/327cd3e5f386975356c3e9835a13002dd8ee22bc/db/build.gradle#L223)):*
> ```
> test.dependsOn extractFlywayFiles
> ```
processResources.dependsOn would cover more cases, including 'schemaJar', and all flyway tasks can depend on buildNeeded only.
___
*[integration/run_schema_check.sh, line 102 at r2](https://reviewable.io/reviews/google/nomulus/814#-MI-ddaj3-r9KPcjtHkT:-MI-ddak7FrtC5GjwJqj:bb857ps) ([raw file](https://github.com/google/nomulus/blob/327cd3e5f386975356c3e9835a13002dd8ee22bc/integration/run_schema_check.sh#L102)):*
> ```
> "This change is not allows:"
> ```
allowed:

mindhog (commented):
Given that our two most prolific flyway file contributors prefer the current representation, I am retracting this PR and will instead implement the golden file approach.
Thanks, all, for your comments!

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/814#-:-MI0MW4OBv07sCEsLucY:b-h2e3ny)* status: 0 of 61 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

PR #820, opened 2020-09-28
Add VKey.restoreOfy() method for fixing ofy keys

mindhog:
Add a restoreOfy() instance method and a restoreOfyFrom() static method to
assist in restoring the objectify key for classes that have composite keys
that do not restore automatically.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/820)

### DIFF
--- a/core/src/main/java/google/registry/persistence/VKey.java
+++ b/core/src/main/java/google/registry/persistence/VKey.java
@@ -113,6 +113,83 @@ public class VKey<T> extends ImmutableObject implements Serializable {
     return new VKey<T>(kind, Key.create(kind, name), name);
   }
 
+  /**
+   * Returns a clone with an ofy key restored from {@code ancestors}.
+   *
+   * <p>The arguments should generally consist of pairs of Class and value, where the Class is the
+   * kind of the ancestor key and the value is either a String or a Long.
+   *
+   * <p>For example, to restore the objectify key for
+   * DomainBase("COM-1234")/HistoryEntry(123)/PollEvent(567), one might use:
+   *
+   * <pre>{@code
+   * pollEvent.restoreOfy(DomainBase.class, "COM-1234", HistoryEntry.class, 567)
+   * }</pre>
+   *
+   * <p>The final key id or name is obtained from the SQL key. It is assumed that this value must be
+   * either a long integer or a {@code String} and that this proper identifier for the objectify
+   * key.
+   *
+   * <p>As a special case, an objectify Key may be used as the first ancestor instead of a Class,
+   * value pair.
+   */
+  public VKey<T> restoreOfy(Object... ancestors) {
+    Class lastClass = null;
+    Key<?> lastKey = null;
+    for (Object ancestor : ancestors) {
+      if (ancestor instanceof Class) {
+        if (lastClass != null) {
+          throw new IllegalArgumentException(ancestor + " used as a key value.");
+        }
+        lastClass = (Class) ancestor;
+        continue;
+      } else if (ancestor instanceof Key) {
+        if (lastKey != null) {
+          throw new IllegalArgumentException(
+              "Objectify keys may only be used for the first argument");
+        }
+        lastKey = (Key) ancestor;
+        continue;
+      }
+
+      // The argument should be a value.
+      if (lastClass == null) {
+        throw new IllegalArgumentException("Argument " + ancestor + " should be a class.");
+      }
+      if (ancestor instanceof Long) {
+        lastKey = Key.create(lastKey, lastClass, (Long) ancestor);
+      } else if (ancestor instanceof String) {
+        lastKey = Key.create(lastKey, lastClass, (String) ancestor);
+      } else {
+        throw new IllegalArgumentException("Key value " + ancestor + " must be a string or long.");
+      }
+      lastClass = null;
+    }
+
+    // Make sure we didn't end up with a dangling class with no value.
+    if (lastClass != null) {
+      throw new IllegalArgumentException("Missing value for last key of type " + lastClass);
+    }
+
+    Object sqlKey = getSqlKey();
+    Key<T> ofyKey =
+        sqlKey instanceof Long
+            ? Key.create(lastKey, getKind(), (Long) sqlKey)
+            : Key.create(lastKey, getKind(), (String) sqlKey);
+
+    return VKey.create((Class<T>) getKind(), sqlKey, ofyKey);
+  }
+
+  /**
+   * Returns a clone of {@code key} with an ofy key restored from {@code ancestors}.
+   *
+   * <p>This is the static form of the method restoreOfy() above. If {@code key} is null, it returns
+   * null.
+   */
+  public static <T> VKey<T> restoreOfyFrom(@Nullable VKey<T> key, Object... ancestors) {
+    return key == null ? null : key.restoreOfy(ancestors);
+  }
+
   /** Returns the type of the entity. */
   public Class<? extends T> getKind() {
     return this.kind;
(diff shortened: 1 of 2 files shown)


hstonec (commented):
I was wondering how/where you'd expect to use these added methods? Will they be used in entity to help reconstruct symmetric vkey?

I am asking this because for entities using hierarchical ofy key, it must have a `parent` field in the entity. So, to make sure that we construct a symmetric key when the entity is loaded from Cloud SQL, we can backfill `parent` field in `@PostLoad` method and use `VKey.create(Entity.class, sqlKey, Key.create(this))` in `Entity.createVKey` to reconstruct vkey, note that `Key.create(this)` respects the ancestors as long as `parent` is set correctly.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/820#-:-MIMhfomABZJV97DHuSt:bxmn4i6)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @hstonec)

mindhog (commented):
This came about in the course of implementing VKey restoration in Domain, where we don't have access to the entity we're creating a vkey for (the BillingEvent and PollMessage objects) or the history object.  It was useful to have a helper function to "restoreOfyFrom(originalVKey, domainKey, HistoryEvent.class, historyId)"
We may be able to get away with something even more terse and specific, but for now it was easy enough to just solve for the general case. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/820#-:-MIPBEj7C-NJWYM9qa5U:b-8gr5gs)* status: 0 of 2 files reviewed, all discussions resolved

hstonec (changes requested):
Got it, makes sense.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/820#-:-MIPSiu5APOjUyNmZs1O:b-svvhpp)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/820#-MIQ9CMd4JzZToF10Cgo:-MIQ9CMd4JzZToF10Cgp:bz42e0w) ([raw file](https://github.com/google/nomulus/blob/9564cf515ceb740604c648c991d07980ba3daabb/core/src/main/java/google/registry/persistence/VKey.java#L136)):*
> ```
> public VKey<T> restoreOfy(Object... ancestors)
> ```
In practice, do you think if there is any case that a key has more that two ancestors? If not, should we just have `restoreOfy(Class parent, Object value)` and `restoreOfy(Class grandParent, Object grandParentValue, Class parent, Object parentValue)` to avoid the type checks below?
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 146 at r1](https://reviewable.io/reviews/google/nomulus/820#-MIPioC-5LHqDe6GVxcb:-MIPioC-5LHqDe6GVxcc:bbvz8v5) ([raw file](https://github.com/google/nomulus/blob/9564cf515ceb740604c648c991d07980ba3daabb/core/src/main/java/google/registry/persistence/VKey.java#L146)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
> } else if (ancestor instanceof Key) {
>         if (lastKey != null) {
>           throw new IllegalArgumentException(
>               "Objectify keys may only be used for the first argument");
>         }
>         lastKey = (Key) ancestor;
>         continue;
>       }
> ```
</details>

As we know the specific type of the parameter, maybe we should consider to just add a `restoreOfy(Key ofyKey, Object... ancestors)` to handle the case?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/820)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/820#-MIQ9CMd4JzZToF10Cgo:-MIQD8f79Aif_KaWJxQq:b7wjkjd) ([raw file](https://github.com/google/nomulus/blob/9564cf515ceb740604c648c991d07980ba3daabb/core/src/main/java/google/registry/persistence/VKey.java#L136)):*

> ```
> public VKey<T> restoreOfy(Object... ancestors)
> ```
In practice, do you think if there is any case that a key has more that two ancestors? If not, should we just have `restoreOfy(Class parent, Object value)` and `restoreOfy(Class grandParent, Object grandParentValue, Class parent, Object parentValue)` to avoid the type checks below?
</blockquote></details>

I doubt that we have more than two anywhere.

Yes, we can probably get away with some simpler statically typed versions of these.  My reason for doing this is that I'd just as soon not think about the cases in advance.  This approach is sufficiently general and easy enough to test that we can just turn it out up front and use it on anything.

We can retrofit to the more specific cases once we've got them all.
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 146 at r1](https://reviewable.io/reviews/google/nomulus/820#-MIPioC-5LHqDe6GVxcb:-MIQFn0D1rrrPD_1LMpB:b-706y90) ([raw file](https://github.com/google/nomulus/blob/9564cf515ceb740604c648c991d07980ba3daabb/core/src/main/java/google/registry/persistence/VKey.java#L146)):*

> ```
> } else if (ancestor instanceof Key) {
>         if (lastKey != null) {
>           throw new IllegalArgumentException(
>               "Objectify keys may only be used for the first argument");
>         }
>         lastKey = (Key) ancestor;
>         continue;
>       }
> ```
As we know the specific type of the parameter, maybe we should consider to just add a `restoreOfy(Key ofyKey, Object... ancestors)` to handle the case?
</blockquote></details>

Since the parameters are "Object..." you'd have to either still do the additional key check or accept the slightly less nice error checking imposed by the "class, String or Long" logic.

I'd be more inclined to just do the statically typed versions, TBH.

hstonec (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/820#-:-MIQrxp980Smgi1uzNkY:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/820#-MIQ9CMd4JzZToF10Cgo:-MIQeEt07eNPSo7Y-W0q:b-fd3vq5) ([raw file](https://github.com/google/nomulus/blob/9564cf515ceb740604c648c991d07980ba3daabb/core/src/main/java/google/registry/persistence/VKey.java#L136)):*

I doubt that we have more than two anywhere.

Yes, we can probably get away with some simpler statically typed versions of these.  My reason for doing this is that I'd just as soon not think about the cases in advance.  This approach is sufficiently general and easy enough to test that we can just turn it out up front and use it on anything.

We can retrofit to the more specific cases once we've got them all.
</blockquote></details>

sounds good to me. 
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 146 at r1](https://reviewable.io/reviews/google/nomulus/820#-MIPioC-5LHqDe6GVxcb:-MIQgpUu6xrFd-z67FYA:b-o0nmzc) ([raw file](https://github.com/google/nomulus/blob/9564cf515ceb740604c648c991d07980ba3daabb/core/src/main/java/google/registry/persistence/VKey.java#L146)):*

Since the parameters are "Object..." you'd have to either still do the additional key check or accept the slightly less nice error checking imposed by the "class, String or Long" logic.

I'd be more inclined to just do the statically typed versions, TBH. 
</blockquote></details>

ok, as you mentioned above I think we can go with this solution and review it later.

PR #821, opened 2020-09-29
Move "WithLongVKey" to BillingEvent subclasses

mindhog:
When loading the VKeys for the BillingEvents hierarchy, it is necessary to
restore the original concrete class for the type, otherwise we end up with a
different (and incompatible) VKey.

This change thus moves the WithLongVKey annotation from BillingEvent
to its derived classes, which fixes a problem in restoring the key in
DomainBase.

As part of this, convert the cancellation matching billing event to
VKey<Recurring>, which seems like the only thing it actually can be.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/821)


hstonec (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/821)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #826, opened 2020-10-02
Add more Datastore/Sql Entity annotations

gbrodman:
Note: EntityTest requires a 1-1 correspondence, meaning that while we could have HistoryEntry implement SqlEntity (to avoid the repetition of the same code in the child classes) the invariant of "javax.persistence.Entity classes === SqlEntity classes" means it is necessary and cleaner to implement it directly in the child classes

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/826)


hstonec (changes requested):
> Reviewed 4 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/826)* status: 4 of 11 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @hstonec)

___
*[core/src/main/java/google/registry/model/OteAccountBuilder.java, line 255 at r1](https://reviewable.io/reviews/google/nomulus/826#-MIt1Vvs60riWh4B18mu:-MIt1Vvs60riWh4B18mv:b-phiwq2) ([raw file](https://github.com/google/nomulus/blob/233b22d458bec6b72c24ebd7ba737cf409396cf0/core/src/main/java/google/registry/model/OteAccountBuilder.java#L255)):*
> ```
> // use ImmutableObject instead of Registry so that the Key generation doesn't break
> ```
I am curious why this is the case?
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 108 at r1](https://reviewable.io/reviews/google/nomulus/826#-MIt2liT4-x142wcJgMN:-MIt2liT4-x142wcJgMO:bvays0x) ([raw file](https://github.com/google/nomulus/blob/233b22d458bec6b72c24ebd7ba737cf409396cf0/core/src/main/java/google/registry/model/contact/ContactHistory.java#L108)):*
> ```
> @Override
>   // In Datastore, save as a HistoryEntry object regardless of this object's type
> ```
nit: put comment above `@Override` ?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/826)* status: 4 of 11 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/OteAccountBuilder.java, line 255 at r1](https://reviewable.io/reviews/google/nomulus/826#-MIt1Vvs60riWh4B18mu:-MIt7Glj6Jqrg0fIhStZ:bvcv3rs) ([raw file](https://github.com/google/nomulus/blob/233b22d458bec6b72c24ebd7ba737cf409396cf0/core/src/main/java/google/registry/model/OteAccountBuilder.java#L255)):*

> ```
> // use ImmutableObject instead of Registry so that the Key generation doesn't break
> ```
I am curious why this is the case?
</blockquote></details>

This has to do with the stream a few lines below this. Basically, we want to create a stream of `ImmutableObject`s, created by Stream.concat(...). Previously, the only interface that RegistrarContact, Registry, and Registrar share in common is `ImmutableObject`, so the stream becomes that type. However,  when we add DatastoreEntity to Registry in this PR, the compiler gets too smart for its own good and tries to create Key<ImmutableObject & DatastoreEntity> which results in issues when calling ofy(). 

Here, we're explicitly downcasting the collection to say it's just ImmutableObject, rather than `ImmutableObject & DatastoreEntity`

hstonec (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/826)* status: 4 of 11 files reviewed, all discussions resolved

PR #830, opened 2020-10-08
Use the parent to store the history repo ID and fill in the base object

gbrodman:
Storing the repo ID in the parent and in the base object has two primary
benefits.

First, it unifies the parent information in the HistoryEntry's `parent`
object. This simplifies the builders and the data flow.

Second, when possible (which should be always, post-migration) we fill
out the DomainContent's repo ID (similarly for the other EPP resources)
which means that when reconstituting the ofy keys we don't need to pass
the repo ID in from a separate object. This way, all the data are
encapsulated where they should be.

The primary downside here is that it further reduces the "immutability"
of the history objects (since we're using the Hibernate setter for the
parent repo ID) but we weren't immutable anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/830)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/830)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8o15_085LMepb8YMH:-MJ8o15_085LMepb8YMI:b-x23nqd) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/main/java/google/registry/model/contact/ContactHistory.java#L68)):*
> ```Java
>       return contactBase.getRepoId();
>     }
>     return parent.getName();
> ```

Would it be bad to simply always return parent.getName() for this?  Can it ever be missing or different from the contact repo id? (please add a comment describing how this works, if so)
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8tuWq-B_M8FKZGPxT:-MJ8okEp5wy7I8dIztv5:b-psbsso) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/main/java/google/registry/model/contact/ContactHistory.java#L101)):*
> ```Java
>         ContactResource.class,
>         getContactRepoId(),
>         Key.create(ContactResource.class, getContactRepoId()));
> ```

I'm a little concerned that we're reconstructing VKeys here instead of just storing the fields as VKeys, I haven't thought through the implications of that pattern.  Though one thing I can say is that it will make the final conversion a bit more complicated.

In any case, that's not something we should address in this PR. 
___
*[core/src/test/java/google/registry/model/history/ContactHistoryTest.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8tNK86O0RvOIyIiBq:-MJ8tNK97uSz5wzErmx0:b-25rfrt) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/test/java/google/registry/model/history/ContactHistoryTest.java#L59)):*
> ```Java
>               assertContactHistoriesEqual(fromDatabase, contactHistory);
>               assertThat(fromDatabase.getParentVKey().getSqlKey())
>                   .isEqualTo(contactHistory.getParentVKey().getSqlKey());
> ```

Don't we want to compare the two vkeys here and below?  They should be symmetric, right?
___
*[core/src/test/java/google/registry/model/history/DomainHistoryTest.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8te10D4LYO6i1aLln:-MJ8te10D4LYO6i1aLlo:b5tj20g) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/test/java/google/registry/model/history/DomainHistoryTest.java#L70)):*
> ```Java
>               assertDomainHistoriesEqual(fromDatabase, domainHistory);
>               assertThat(fromDatabase.getParentVKey().getSqlKey())
>                   .isEqualTo(domainHistory.getParentVKey().getSqlKey());
> ```

Same here and below, compare full vkeys.  Also in the following files.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/830)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8o15_085LMepb8YMH:-MJ8vz1jEAhx5FGmFrea:b-tku3az) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/main/java/google/registry/model/contact/ContactHistory.java#L68)):*

Would it be bad to simply always return parent.getName() for this?  Can it ever be missing or different from the contact repo id? (please add a comment describing how this works, if so)
</blockquote></details>

I guess since we'll always need to worry about the old objects we can't ever really remove the `if` statement at the beginning of this method, so we could probably get away with just returning the parent's name always. 

It makes a cleaner correspondence to the setter that way, too. 
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8tuWq-B_M8FKZGPxT:-MJ9-j-z3Av_WIFPQO_e:bkb6rrx) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/main/java/google/registry/model/contact/ContactHistory.java#L101)):*

I'm a little concerned that we're reconstructing VKeys here instead of just storing the fields as VKeys, I haven't thought through the implications of that pattern.  Though one thing I can say is that it will make the final conversion a bit more complicated.

In any case, that's not something we should address in this PR. 
</blockquote></details>

Yeah I think we can simplify this though, right? (see most recent commit) 
___
*[core/src/test/java/google/registry/model/history/ContactHistoryTest.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8tNK86O0RvOIyIiBq:-MJ8ze5w7t9VXFBhQzkx:bb77uci) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/test/java/google/registry/model/history/ContactHistoryTest.java#L59)):*

Don't we want to compare the two vkeys here and below?  They should be symmetric, right?
</blockquote></details>

Yep!
___
*[core/src/test/java/google/registry/model/history/DomainHistoryTest.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/830#-MJ8te10D4LYO6i1aLln:-MJ8zfdm10ImwgIXxPde:b-896fix) ([raw file](https://github.com/google/nomulus/blob/913ca09489ba71321e6334c9f98881f9e646ce5f/core/src/test/java/google/registry/model/history/DomainHistoryTest.java#L70)):*

Same here and below, compare full vkeys.  Also in the following files.
</blockquote></details>

Done.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/830)* status: :shipit: complete! all files reviewed, all discussions resolved

mindhog (approved):
> Reviewed 1 of 11 files at r1, 1 of 7 files at r2, 7 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/830)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #832, opened 2020-10-09
Minor python changes

weiminyu:
Use dataclasses instead of attrs. The former is part of the standard lib
while the latter may need to be installed separately. For now we are not
using advanced features in attrs. Ran nom_build_test by hand and it passes.

Also added python3 to the list of prerequisites in docs.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/832)

### DIFF
--- a/config/nom_build.py
+++ b/config/nom_build.py
@@ -16,7 +16,7 @@
 """
 
 import argparse
-import attr
+import dataclasses
 import io
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/832)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #835, opened 2020-10-09
Create SQL schema for RdeRevision

gbrodman:
This has the side effect of backing up the RdeRevision classes (previously we were using saveWithoutBackup) but that shouldn't be a problem. Now, we use `tm()` calls instead of the raw `ofy()` calls. 

We also change verify() to checkArgument() to be similar to the rest of the codebase. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/835)


gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbs0qjEVE08L-_mrXj:-MJc3dQ_7MAm36zGo38c:b-qvghd7) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/core/src/main/java/google/registry/model/rde/RdeRevision.java#L79)):*

Fundamentally this is the major problem with how RdeRevisions are being stored -- we're mashing together three separate pieces of data into a string in a way that makes them much less accessible.  The proper Cloud SQL schema for this would have these as 3 separate fields.  See `Cursor` for a similar example where the Cloud SQL schema was corrected to store the component parts separately.  Can we do that here?  Does this even need to exist as an entity type separate from `Cursor`?  Can it just be a `Cursor`, or is there additional data that doesn't fit in?  (And if so should we perhaps generalize the Cursor table in some manner to accommodate this use case?)
</blockquote></details>

I don't believe we can generalize the Cursor table or use it to store this, RdeRevision is a fairly separate concept. 

The Cursor table uses dual writing which we're trying to avoid, and it might not be necessary in this situation. One option here would be to continue using the Datastore schema as-is, but have a separate SQL schema that separates the fields out. Then, we could populate the Datastore field in a PostLoad method (called when loading from SQL) and we could populate the SQL fields in an OnLoad method (called when loading from Datastore). That way we get automatic dual writing. Thoughts?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbs0qjEVE08L-_mrXj:-MJcFA7jENrH1R-uhgjj:b-16jq5z) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/core/src/main/java/google/registry/model/rde/RdeRevision.java#L79)):*

I don't believe we can generalize the Cursor table or use it to store this, RdeRevision is a fairly separate concept. 

The Cursor table uses dual writing which we're trying to avoid, and it might not be necessary in this situation. One option here would be to continue using the Datastore schema as-is, but have a separate SQL schema that separates the fields out. Then, we could populate the Datastore field in a PostLoad method (called when loading from SQL) and we could populate the SQL fields in an OnLoad method (called when loading from Datastore). That way we get automatic dual writing. Thoughts? 
</blockquote></details>

eh, on the other hand since the only unique identifiers will be either the triplet or some assigned revision identifier, there will be no good way to use VKeys to select the appropriate entry without concern to which database we're referring to. Basically, we have to have different logic on a per-database level if we want to change the primary key, unfortunately.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 64 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbqGltAdUnF9MZnZkr:-MJgyeRWEclHzLE0n7e7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/core/src/main/java/google/registry/model/rde/RdeRevision.java#L64)):*

`maybeObject` isn't a great name.  How about `revisionOpt` or similar?  Or just `revision`?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbs0qjEVE08L-_mrXj:-MJgyLfUC5R8B0PyK_m4:b-btjk2y) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/core/src/main/java/google/registry/model/rde/RdeRevision.java#L79)):*

eh, on the other hand since the only unique identifiers will be either the triplet or some assigned revision identifier, there will be no good way to use VKeys to select the appropriate entry without concern to which database we're referring to. Basically, we have to have different logic on a per-database level if we want to change the primary key, unfortunately. 
</blockquote></details>

OK, adding an IdClass makes this a million times better.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 492 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbq1w_0ela8jv4or0W:-MJgyOllCpRgOGy-xw3U:b-gy9y0l) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/db/src/main/resources/sql/schema/db-schema.sql.generated#L492)):*

`revision_id`?
</blockquote></details>

irrelevant
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 493 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbq-1350TGPIHPzdz2:-MJgyPTdDojw5OTdWrYf:b-jfyl5e) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/db/src/main/resources/sql/schema/db-schema.sql.generated#L493)):*

`NOT NULL`?
</blockquote></details>

it's nullable everywhere else

gbrodman (commented):
OK should be better, PTAL thanks 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835#-:-MJhBDMh2YUl4eb8GinC:b-9r7iav)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJmSC64BuVPicAWSd3L:-MJmSC657ndOGz5DOISv:bsfsmjn) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L77)):*
> ```Java
>   private RdeRevision() {}
> 
>   public RdeRevision(String id, String tld, LocalDate date, RdeMode mode, int revision) {
> ```

In general, *Effective Java* prefers static construction methods over constructors.  That's what we tend towards in our codebase at large so we may as well not deviate here.
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 134 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnRfrQ5PxfDIV4TnW2:-MJnRfrQ5PxfDIV4TnW3:b2szy6b) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L134)):*
> ```Java
>           rdeRevision -> {
>             throw new IllegalArgumentException(
>                 String.format("RdeRevision object already created: %s", rdeRevision));
> ```

The error message should mention specifically that this is only an error when trying to save the first revision (i.e. revision 0).
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 139 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQshL0D5A0Td_RLKd:-MJnQshL0D5A0Td_RLKe:b9o6sc2) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L139)):*
> ```Java
>       checkArgument(
>           revisionOptional.isPresent(),
>           "RDE revision object missing for %s?! revision=%s",
> ```

Error message could be a little bit better.  Something like "Could not load existing RDE revision for %s when trying to save new RDE revision >=1: %s"
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 168 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQf-b9W5eCi-jYxk8:-MJnQf-b9W5eCi-jYxk9:b-dfbgfq) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L168)):*
> ```Java
>     private RdeRevisionId() {}
> 
>     RdeRevisionId(String tld, LocalDate date, RdeMode mode) {
> ```

Same constructor comment.
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 174 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQCvp2MSN1lZJxRr3:-MJnQCvp2MSN1lZJxRr4:bp1amfw) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L174)):*
> ```Java
>     }
> 
>     /** Private, only used by Hibernate. */
> ```

Ugh, having 6 of these "for Hibernate only" methods is pretty annoying.  I wonder if we can find anything better as a general-case solution.  This leads to lots of boilerplate code bloat.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 493 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbq-1350TGPIHPzdz2:-MJnPQoA9ydueVUkZ0Xs:b-ymwpoc) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/db/src/main/resources/sql/schema/db-schema.sql.generated#L493)):*

it's nullable everywhere else 
</blockquote></details>

Is that correct though?  How can this ever be null?  When it's first written, shouldn't this correspond to the creation time a.k.a. the first "update"?

Although, since that would require changes to `UpdateAutoTimestamp`, it seems like it's larger in scope to deal with than just this PR.  So we'll leave it alone for now.
___
*[db/src/main/resources/sql/flyway/V65__create_rde_revision.sql, line 18 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnPJxt3PKRmoJ5wp1i:-MJnPJxt3PKRmoJ5wp1j:b-sozxm7) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/db/src/main/resources/sql/flyway/V65__create_rde_revision.sql#L18)):*
> ```TSQL
>    tld TEXT NOT NULL,
>    mode TEXT NOT NULL,
>    date TEXT NOT NULL,
> ```

`date` is a `TEXT` data type?  That doesn't seem right.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJmSC64BuVPicAWSd3L:-MK1kefZ60f7sUfsqBxY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L77)):*

In general, *Effective Java* prefers static construction methods over constructors.  That's what we tend towards in our codebase at large so we may as well not deviate here.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 134 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnRfrQ5PxfDIV4TnW2:-MK1nb7nCeB9Qg68DQF4:b-xy1ivw) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L134)):*

The error message should mention specifically that this is only an error when trying to save the first revision (i.e. revision 0).
</blockquote></details>

Done. (this was just the old comment) 
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 139 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQshL0D5A0Td_RLKd:-MK1nkzN2kAgBj28vq5s:b-xy1ivw) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L139)):*

Error message could be a little bit better.  Something like "Could not load existing RDE revision for %s when trying to save new RDE revision >=1: %s"
</blockquote></details>

Done. (this was just the old comment) 
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 168 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQf-b9W5eCi-jYxk8:-MK1ktcL8_SdIPOtK6Us:b-ujct89) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L168)):*

Same constructor comment.
</blockquote></details>

Done, though we don't do that for the other *Id objects
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 174 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQCvp2MSN1lZJxRr3:-MK1m5Jv5u2NEzNeEQAb:b-rzwl2g) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L174)):*

Ugh, having 6 of these "for Hibernate only" methods is pretty annoying.  I wonder if we can find anything better as a general-case solution.  This leads to lots of boilerplate code bloat.
</blockquote></details>

Good news: I played around with this a bit and it seems like if we use field access we can get rid of these (phew) 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 493 at r1](https://reviewable.io/reviews/google/nomulus/835#-MJbq-1350TGPIHPzdz2:-MK1kQkg34z3qCqL_YGR:bevogeq) ([raw file](https://github.com/google/nomulus/blob/c19a53a6b8974c3e6a633b1e09ffdf6a9cf3fc3b/db/src/main/resources/sql/schema/db-schema.sql.generated#L493)):*

Is that correct though?  How can this ever be null?  When it's first written, shouldn't this correspond to the creation time a.k.a. the first "update"?

Although, since that would require changes to `UpdateAutoTimestamp`, it seems like it's larger in scope to deal with than just this PR.  So we'll leave it alone for now.
</blockquote></details>

Yeah, let's punt on this for now 
___
*[db/src/main/resources/sql/flyway/V65__create_rde_revision.sql, line 18 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnPJxt3PKRmoJ5wp1i:-MK1kN6G9R0JuSC9c-jC:b-4o0xmj) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/db/src/main/resources/sql/flyway/V65__create_rde_revision.sql#L18)):*

`date` is a `TEXT` data type?  That doesn't seem right.
</blockquote></details>

https://github.com/google/nomulus/pull/842

Leftover from the internship, fixed in separate PR (that one will likely go through before this one)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 11 files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 168 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQf-b9W5eCi-jYxk8:-MKFxj6n3LCNx58h2nNw:b90wbce) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L168)):*

Done, though we don't do that for the other *Id objects
</blockquote></details>

You know, you don't need to write a private constructor and then a public static create method.  That's a fair bit of duplication.  You only need one or the other.  Something like:

```
private RdeRevisionId() {}

static RdeRevisionId create(String tld, LocalDate date, RdeMode mode) {
  RdeRevisionId instance = new RdeRevisionId();
  instance.tld = tld;
  instance.date = date;
  instance.mode = mode;
  return instance;
}
```

Are there any compatibility problems with Hibernate?

(And the same changes could be made above too where you've duplicated both types of construction.)
___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 174 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQCvp2MSN1lZJxRr3:-MKFy9pF-zenYxo60xUX:bb77rlh) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L174)):*

Good news: I played around with this a bit and it seems like if we use field access we can get rid of these (phew) 
</blockquote></details>

Yay!
___
*[db/src/main/resources/sql/flyway/V65__create_rde_revision.sql, line 18 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnPJxt3PKRmoJ5wp1i:-MKFyGp13Cq5JjB6tnvN:b-wobp58) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/db/src/main/resources/sql/flyway/V65__create_rde_revision.sql#L18)):*

https://github.com/google/nomulus/pull/842

Leftover from the internship, fixed in separate PR (that one will likely go through before this one) 
</blockquote></details>

Looks like it's gone through now?

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835#-:-MKFyMoE-e_l_JT6aIG2:bnfp4nl)* status: 0 of 11 files reviewed, all discussions resolved

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/835)* status: 0 of 11 files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/model/rde/RdeRevision.java, line 168 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnQf-b9W5eCi-jYxk8:-MKG0tfl4fS1Fpm9RIlv:bl27uxt) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/core/src/main/java/google/registry/model/rde/RdeRevision.java#L168)):*

You know, you don't need to write a private constructor and then a public static create method.  That's a fair bit of duplication.  You only need one or the other.  Something like:

```
private RdeRevisionId() {}

static RdeRevisionId create(String tld, LocalDate date, RdeMode mode) {
  RdeRevisionId instance = new RdeRevisionId();
  instance.tld = tld;
  instance.date = date;
  instance.mode = mode;
  return instance;
}
```

Are there any compatibility problems with Hibernate?

(And the same changes could be made above too where you've duplicated both types of construction.)
</blockquote></details>

Discussed online, done. 
___
*[db/src/main/resources/sql/flyway/V65__create_rde_revision.sql, line 18 at r2](https://reviewable.io/reviews/google/nomulus/835#-MJnPJxt3PKRmoJ5wp1i:-MKFzHfA-HPc-h60__yE:b-g26rbm) ([raw file](https://github.com/google/nomulus/blob/47189493de4a5ebaf23168595efce0f52ae46ff5/db/src/main/resources/sql/flyway/V65__create_rde_revision.sql#L18)):*

Looks like it's gone through now?
</blockquote></details>

Yes, that PR already went through and I merged master into this branch

PR #837, opened 2020-10-13
Add certificate checks for create and update registrar commands

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/837)

### DIFF
--- a/core/src/main/java/google/registry/config/CertificateCheckerModule.java
+++ b/core/src/main/java/google/registry/config/CertificateCheckerModule.java
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package google.registry.tools;
+package google.registry.config;
 
 import com.google.common.collect.ImmutableSortedMap;
 import dagger.Module;
@@ -20,7 +20,6 @@ import dagger.Provides;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.util.CertificateChecker;
 import google.registry.util.SystemClock;
-import javax.inject.Named;
 import javax.inject.Singleton;
 import org.joda.time.DateTime;
 
@@ -29,7 +28,6 @@ public abstract class CertificateCheckerModule {
 
   @Provides
   @Singleton
-  @Named("certificateChecker")
   static CertificateChecker provideCertificateChecker(
       @Config("validityDaysMap") ImmutableSortedMap<DateTime, Integer> validityDaysMap,
       @Config("daysToExpiration") int daysToExpiration,
--- a/core/src/main/java/google/registry/config/CertificateCheckerModule.java
+++ b/core/src/main/java/google/registry/config/CertificateCheckerModule.java
@@ -24,6 +24,9 @@ import javax.inject.Singleton;
 import org.joda.time.DateTime;
 
 /** Dagger module that provides the {@link CertificateChecker} used in the application. */
+// TODO(sarahbot@): Move this module to a better location. Possibly flows/. If we decide to move
+// CertificateChecker.java to core/ delete this file and inject the CertificateChecker constructor
+// instead.
 @Module
 public abstract class CertificateCheckerModule {
 
--- a/core/src/main/java/google/registry/config/CertificateCheckerModule.java
+++ b/core/src/main/java/google/registry/config/CertificateCheckerModule.java
@@ -19,21 +19,22 @@ import dagger.Module;
 import dagger.Provides;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.util.CertificateChecker;
-import google.registry.util.SystemClock;
+import google.registry.util.Clock;
 import javax.inject.Singleton;
 import org.joda.time.DateTime;
 
+/** Dagger module that provides the {@link CertificateChecker} used in the application. */
 @Module
 public abstract class CertificateCheckerModule {
 
   @Provides
   @Singleton
   static CertificateChecker provideCertificateChecker(
-      @Config("validityDaysMap") ImmutableSortedMap<DateTime, Integer> validityDaysMap,
-      @Config("daysToExpiration") int daysToExpiration,
-      @Config("minimumRsaKeyLength") int minimumRsaKeyLength) {
-    return new CertificateChecker(
-        validityDaysMap, daysToExpiration, minimumRsaKeyLength, new SystemClock());
+      @Config("maxValidityDaysSchedule") ImmutableSortedMap<DateTime, Integer> validityDaysMap,
+      @Config("expirationWarningDays") int daysToExpiration,
+      @Config("minimumRsaKeyLength") int minimumRsaKeyLength,
+      Clock clock) {
+    return new CertificateChecker(validityDaysMap, daysToExpiration, minimumRsaKeyLength, clock);
   }
 
   private CertificateCheckerModule() {}
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -16,10 +16,12 @@ package google.registry.config;
 
 import static com.google.common.base.Suppliers.memoize;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableSortedMap.toImmutableSortedMap;
 import static google.registry.config.ConfigUtils.makeUrl;
 import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static google.registry.util.ResourceUtils.readResourceUtf8;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static java.util.Comparator.naturalOrder;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Ascii;
@@ -1350,29 +1352,30 @@ public final class RegistryConfig {
     }
 
     @Provides
-    @Config("validityDaysMap")
+    @Config("maxValidityDaysSchedule")
     public static ImmutableSortedMap<DateTime, Integer> provideValidityDaysMap(
         RegistryConfigSettings config) {
-      Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
-      ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
-          ImmutableSortedMap.naturalOrder();
-      stringMap.forEach(
-          (k, v) ->
-              validityDaysMap.put(
-                  k.equals("START_OF_TIME") ? START_OF_TIME : DateTime.parse(k), v));
-      return validityDaysMap.build();
+      return config.sslCertificateValidation.maxValidityDaysSchedule.entrySet().stream()
+          .collect(
+              toImmutableSortedMap(
+                  naturalOrder(),
+                  e ->
+                      e.getKey().equals("START_OF_TIME")
+                          ? START_OF_TIME
+                          : DateTime.parse(e.getKey()),
+                  e -> e.getValue()));
     }
 
     @Provides
-    @Config("daysToExpiration")
+    @Config("expirationWarningDays")
     public static int provideDaysToExpiration(RegistryConfigSettings config) {
-      return config.certChecker.daysToExpiration;
+      return config.sslCertificateValidation.expirationWarningDays;
     }
 
     @Provides
     @Config("minimumRsaKeyLength")
     public static int provideMinimumRsaKeyLength(RegistryConfigSettings config) {
-      return config.certChecker.minimumRsaKeyLength;
+      return config.sslCertificateValidation.minimumRsaKeyLength;
     }
   }
 
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -30,9 +30,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
 import dagger.Module;
 import dagger.Provides;
-import google.registry.util.CertificateChecker;
-import google.registry.util.Clock;
-import google.registry.util.SystemClock;
 import google.registry.util.TaskQueueUtils;
 import google.registry.util.YamlUtils;
 import java.lang.annotation.Documented;
@@ -1351,20 +1348,45 @@ public final class RegistryConfig {
       return config.registryPolicy.rdapTosStaticUrl;
     }
 
+    // @Provides
+    // @Config("certificateChecker")
+    // public static CertificateChecker provideCertificateChecker(RegistryConfigSettings config) {
+    //   Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
+    //   ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
+    //       ImmutableSortedMap.naturalOrder();
+    //   for (String key : stringMap.keySet()) {
+    //     validityDaysMap.put(DateTime.parse(key), stringMap.get(key));
+    //   }
+    //   return new CertificateChecker(
+    //       ImmutableSortedMap.copyOf(validityDaysMap.build()),
+    //       config.certChecker.daysToExpiration,
+    //       config.certChecker.minimumRsaKeyLength,
+    //       new SystemClock());
+    // }
+
     @Provides
-    @Config("certificateChecker")
-    public static CertificateChecker provideCertificateChecker(RegistryConfigSettings config) {
+    @Config("validityDaysMap")
+    public static ImmutableSortedMap<DateTime, Integer> provideValidityDaysMap(
+        RegistryConfigSettings config) {
       Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
       ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
           ImmutableSortedMap.naturalOrder();
       for (String key : stringMap.keySet()) {
         validityDaysMap.put(DateTime.parse(key), stringMap.get(key));
       }
-      return new CertificateChecker(
-          ImmutableSortedMap.copyOf(validityDaysMap.build()),
-          config.certChecker.daysToExpiration,
-          config.certChecker.minimumRsaKeyLength,
-          new SystemClock());
+      return validityDaysMap.build();
+    }
+
+    @Provides
+    @Config("daysToExpiration")
+    public static int provideDaysToExpiration(RegistryConfigSettings config) {
+      return config.certChecker.daysToExpiration;
+    }
+
+    @Provides
+    @Config("minimumRsaKeyLength")
+    public static int provideMinimumRsaKeyLength(RegistryConfigSettings config) {
+      return config.certChecker.minimumRsaKeyLength;
     }
   }
 
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -27,8 +27,12 @@ import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSortedMap;
 import dagger.Module;
 import dagger.Provides;
+import google.registry.util.CertificateChecker;
+import google.registry.util.Clock;
+import google.registry.util.SystemClock;
 import google.registry.util.TaskQueueUtils;
 import google.registry.util.YamlUtils;
 import java.lang.annotation.Documented;
@@ -46,6 +50,7 @@ import javax.inject.Qualifier;
 import javax.inject.Singleton;
 import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
+import org.joda.time.DateTime;
 import org.joda.time.DateTimeConstants;
 import org.joda.time.Duration;
 
@@ -1344,6 +1349,22 @@ public final class RegistryConfig {
     @Nullable
     public static String provideRdapTosStaticUrl(RegistryConfigSettings config) {
       return config.registryPolicy.rdapTosStaticUrl;
+    }
+
+    @Provides
+    @Config("certificateChecker")
+    public static CertificateChecker provideCertificateChecker(RegistryConfigSettings config) {
+      Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
+      ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
+          ImmutableSortedMap.naturalOrder();
+      for (String key : stringMap.keySet()) {
+        validityDaysMap.put(DateTime.parse(key), stringMap.get(key));
+      }
+      return new CertificateChecker(
+          ImmutableSortedMap.copyOf(validityDaysMap.build()),
+          config.certChecker.daysToExpiration,
+          config.certChecker.minimumRsaKeyLength,
+          new SystemClock());
     }
   }
 
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1348,22 +1348,6 @@ public final class RegistryConfig {
       return config.registryPolicy.rdapTosStaticUrl;
     }
 
-    // @Provides
-    // @Config("certificateChecker")
-    // public static CertificateChecker provideCertificateChecker(RegistryConfigSettings config) {
-    //   Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
-    //   ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
-    //       ImmutableSortedMap.naturalOrder();
-    //   for (String key : stringMap.keySet()) {
-    //     validityDaysMap.put(DateTime.parse(key), stringMap.get(key));
-    //   }
-    //   return new CertificateChecker(
-    //       ImmutableSortedMap.copyOf(validityDaysMap.build()),
-    //       config.certChecker.daysToExpiration,
-    //       config.certChecker.minimumRsaKeyLength,
-    //       new SystemClock());
-    // }
-
     @Provides
     @Config("validityDaysMap")
     public static ImmutableSortedMap<DateTime, Integer> provideValidityDaysMap(
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1356,10 +1356,10 @@ public final class RegistryConfig {
       Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
       ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
           ImmutableSortedMap.naturalOrder();
-      for (String key : stringMap.keySet()) {
-        validityDaysMap.put(
-            key.equals("START_OF_TIME") ? START_OF_TIME : DateTime.parse(key), stringMap.get(key));
-      }
+      stringMap.forEach(
+          (k, v) ->
+              validityDaysMap.put(
+                  k.equals("START_OF_TIME") ? START_OF_TIME : DateTime.parse(k), v));
       return validityDaysMap.build();
     }
 
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -17,6 +17,7 @@ package google.registry.config;
 import static com.google.common.base.Suppliers.memoize;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.config.ConfigUtils.makeUrl;
+import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static google.registry.util.ResourceUtils.readResourceUtf8;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
@@ -1356,7 +1357,8 @@ public final class RegistryConfig {
       ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
           ImmutableSortedMap.naturalOrder();
       for (String key : stringMap.keySet()) {
-        validityDaysMap.put(DateTime.parse(key), stringMap.get(key));
+        validityDaysMap.put(
+            key.equals("START_OF_TIME") ? START_OF_TIME : DateTime.parse(key), stringMap.get(key));
       }
       return validityDaysMap.build();
     }
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -15,6 +15,7 @@
 package google.registry.config;
 
 import java.util.List;
+import java.util.Map;
 
 /** The POJO that YAML config files are deserialized into. */
 public class RegistryConfigSettings {
@@ -38,6 +39,7 @@ public class RegistryConfigSettings {
   public Beam beam;
   public Keyring keyring;
   public RegistryTool registryTool;
+  public CertChecker certChecker;
 
   /** Configuration options that apply to the entire App Engine project. */
   public static class AppEngine {
@@ -217,5 +219,12 @@ public class RegistryConfigSettings {
     public String clientId;
     public String clientSecret;
     public String username;
+  }
+
+  /** Configuration for the certificate checker. */
+  public static class CertChecker {
+    public Map<String, Integer> maxValidityDays;
+    public int daysToExpiration;
+    public int minimumRsaKeyLength;
   }
 }
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -39,7 +39,7 @@ public class RegistryConfigSettings {
   public Beam beam;
   public Keyring keyring;
   public RegistryTool registryTool;
-  public CertChecker certChecker;
+  public SslCertificateValidation sslCertificateValidation;
 
   /** Configuration options that apply to the entire App Engine project. */
   public static class AppEngine {
@@ -222,9 +222,9 @@ public class RegistryConfigSettings {
   }
 
   /** Configuration for the certificate checker. */
-  public static class CertChecker {
-    public Map<String, Integer> maxValidityDays;
-    public int daysToExpiration;
+  public static class SslCertificateValidation {
+    public Map<String, Integer> maxValidityDaysSchedule;
+    public int expirationWarningDays;
     public int minimumRsaKeyLength;
   }
 }
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -447,9 +447,13 @@ registryTool:
   clientSecret: YOUR_CLIENT_SECRET
   username: toolusername
 
-certChecker:
-  maxValidityDays :
+# Configuration options for checking SSL certificates.
+sslCertificateValidation:
+  # A map specifying the maximum amount of days the certificate can be valid. The entry key is the date closest before the date the certificate was issued and the entry value is the applicable maximum validity days for that certificate.
+  maxValidityDaysSchedule:
     "START_OF_TIME": 825
     "2020-09-01T00:00:00Z": 398
-  daysToExpiration: 30
+  # The number of days before a certificate expires that indicates the certificate is nearing expiration and warnings should be sent.
+  expirationWarningDays: 30
+  # The minimum number of bits an RSA key must contain
   minimumRsaKeyLength: 2048
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -449,11 +449,14 @@ registryTool:
 
 # Configuration options for checking SSL certificates.
 sslCertificateValidation:
-  # A map specifying the maximum amount of days the certificate can be valid. The entry key is the date closest before the date the certificate was issued and the entry value is the applicable maximum validity days for that certificate.
+  # A map specifying the maximum amount of days the certificate can be valid.
+  # The entry key is the date closest before the date the certificate was issued
+  # and the entry value is the applicable maximum validity days for that certificate.
   maxValidityDaysSchedule:
     "START_OF_TIME": 825
     "2020-09-01T00:00:00Z": 398
-  # The number of days before a certificate expires that indicates the certificate is nearing expiration and warnings should be sent.
+  # The number of days before a certificate expires that indicates the
+  # certificate is nearing expiration and warnings should be sent.
   expirationWarningDays: 30
   # The minimum number of bits an RSA key must contain
   minimumRsaKeyLength: 2048
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -449,7 +449,7 @@ registryTool:
 
 certChecker:
   maxValidityDays :
-    "0000-01-01T00:00:00Z": 825
+    "START_OF_TIME": 825
     "2020-09-01T00:00:00Z": 398
   daysToExpiration: 30
   minimumRsaKeyLength: 2048
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -446,3 +446,10 @@ registryTool:
   # OAuth client secret used by the tool.
   clientSecret: YOUR_CLIENT_SECRET
   username: toolusername
+
+certChecker:
+  maxValidityDays :
+    "0000-01-01T00:00:00Z": 825
+    "2020-09-01T00:00:00Z": 398
+  daysToExpiration: 30
+  minimumRsaKeyLength: 2048
--- a/core/src/main/java/google/registry/tools/CertificateCheckerModule.java
+++ b/core/src/main/java/google/registry/tools/CertificateCheckerModule.java
@@ -21,14 +21,17 @@ import google.registry.config.RegistryConfig.Config;
 import google.registry.util.CertificateChecker;
 import google.registry.util.SystemClock;
 import javax.inject.Named;
+import javax.inject.Singleton;
+import org.joda.time.DateTime;
 
 @Module
 public abstract class CertificateCheckerModule {
 
   @Provides
+  @Singleton
   @Named("certificateChecker")
   static CertificateChecker provideCertificateChecker(
-      @Config("validityDaysMap") ImmutableSortedMap validityDaysMap,
+      @Config("validityDaysMap") ImmutableSortedMap<DateTime, Integer> validityDaysMap,
       @Config("daysToExpiration") int daysToExpiration,
       @Config("minimumRsaKeyLength") int minimumRsaKeyLength) {
     return new CertificateChecker(
--- a/core/src/main/java/google/registry/tools/CertificateCheckerModule.java
+++ b/core/src/main/java/google/registry/tools/CertificateCheckerModule.java
@@ -1,0 +1,39 @@
+// Copyright 2020 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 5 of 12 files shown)


jianglai (changes requested):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1359 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdngCGDGY0v8IKCGsW:-MJdngCH5W3oGR2vq6iy:bbxwrb6) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/config/RegistryConfig.java#L1359)):*
> ```
>  for (String key : stringMap.keySet()) {
> ```
Use `stringMap.forEach((k, v) -> ...)` for better readability.
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdpP1uBlN48YGSCwl-:-MJdpP1uBlN48YGSCwl0:bcsttt7) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L28)):*
> ```
> public abstract class CertificateCheckerModule 
> ```
We will need to use this module in other components (for example in the registrar console to check the uploaded certificate), so putting it in the tools folder is not really accurate (though it still works).

It feels too bad that we can't annotate the constructor because the `@Config` annotation is defined in core whereas `CertificateChecker` is defined in utils. If we just move the `@Config` definition to utils all the problems would be solved, but it doesn't feels semantically correct if we do this.

@cydeweys any thoughts?
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 32 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdqcJA-71kDyZyY-RY:-MJdqcJA-71kDyZyY-RZ:b51r9kh) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L32)):*
> ```
>   @Named("certificateChecker")
> ```
This is not necessary. There will only ever be one `CertificateChecker` at runtime.
___
*[core/src/main/java/google/registry/tools/RegistryToolComponent.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdrvgQ9KtabBq_XinI:-MJdrvgQ9KtabBq_XinJ:b-fr3h53) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/RegistryToolComponent.java#L94)):*
> ```Java
>   void inject(CreateContactCommand command);
>   void inject(CreateDomainCommand command);
> ```

Remove the extra spaces here and below.
___
*[core/src/test/java/google/registry/tools/CommandTestCase.java, line 177 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdsM0W57Icy6HWWsZ5:-MJdsM0W57Icy6HWWsZ6:b-2gi8m0) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/test/java/google/registry/tools/CommandTestCase.java#L177)):*
> ```
>   String getCertFilename() throws IOException {
>     return writeToNamedTmpFile("cert.pem", CertificateSamples.SAMPLE_CERT);
>   }
> ```
Call the new method instead. Also change the javadoc to distinguish the two.
___
*[core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java, line 390 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdsmDb0j2P64Vq9Wyu:-MJdsmDb0j2P64Vq9Wyv:b-n6qo6d) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java#L390)):*
> ```
> testFail_clientCertFileFlagWithViolations
> ```
Can we add a test to make sure multiple violations are handled correctly?
___
*[core/src/test/java/google/registry/tools/UpdateRegistrarCommandTest.java, line 264 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdstKrCanfPlcbug7P:-MJdstKrCanfPlcbug7Q:b-bzm1aj) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/test/java/google/registry/tools/UpdateRegistrarCommandTest.java#L264)):*
> ```
>  void testFail_certFileWithViolations
> ```
Same comment here.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 3 of 10 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1359 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdngCGDGY0v8IKCGsW:-MJhcKzSEmBzBuHeiPvv:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/config/RegistryConfig.java#L1359)):*

> ```
>  for (String key : stringMap.keySet()) {
> ```
Use `stringMap.forEach((k, v) -> ...)` for better readability.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/RegistryToolComponent.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdrvgQ9KtabBq_XinI:-MJheHWd4ova91KuAPUd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/RegistryToolComponent.java#L94)):*

Remove the extra spaces here and below.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/CommandTestCase.java, line 177 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdsM0W57Icy6HWWsZ5:-MJhfp2w8LJzI4VEbmW8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/test/java/google/registry/tools/CommandTestCase.java#L177)):*

> ```
>   String getCertFilename() throws IOException {
>     return writeToNamedTmpFile("cert.pem", CertificateSamples.SAMPLE_CERT);
>   }
> ```
Call the new method instead. Also change the javadoc to distinguish the two.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java, line 390 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdsmDb0j2P64Vq9Wyu:-MJhht1eAPie_7S2s-xw:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java#L390)):*

> ```
> testFail_clientCertFileFlagWithViolations
> ```
Can we add a test to make sure multiple violations are handled correctly?
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/UpdateRegistrarCommandTest.java, line 264 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdstKrCanfPlcbug7P:-MJhiTQxBdpNxXH3ra7u:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/test/java/google/registry/tools/UpdateRegistrarCommandTest.java#L264)):*

> ```
>  void testFail_certFileWithViolations
> ```
Same comment here.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdpP1uBlN48YGSCwl-:-MJhdmTJ7dIhPnqnuh4l:bjxqvlr) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L28)):*

> ```
> public abstract class CertificateCheckerModule 
> ```
We will need to use this module in other components (for example in the registrar console to check the uploaded certificate), so putting it in the tools folder is not really accurate (though it still works).

It feels too bad that we can't annotate the constructor because the `@Config` annotation is defined in core whereas `CertificateChecker` is defined in utils. If we just move the `@Config` definition to utils all the problems would be solved, but it doesn't feels semantically correct if we do this.

@cydeweys any thoughts?
</blockquote></details>

moved the module to the config folder
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 32 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdqcJA-71kDyZyY-RY:-MJhe7HQ44L2aVAaNPBO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L32)):*

> ```
>   @Named("certificateChecker")
> ```
This is not necessary. There will only ever be one `CertificateChecker` at runtime.
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/RegistryToolComponent.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdrvgQ9KtabBq_XinI:-MJm5_uT4_-O8uXCg1JD:bfii13z) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/RegistryToolComponent.java#L94)):*

Done.
</blockquote></details>

Is it? I think autoformatting added the newlines back.
___
*[core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java, line 444 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJm5q4LAoOufsWWwU7a:-MJm5q4M7TC6gd8zvIFU:biocv9g) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java#L444)):*
> ```
>             "Certificate is expired., Certificate validity period is too long; it must be less"
>                 + " than or equal to 398 days.\n");
> ```
Seems like maybe it is better to use newlines to separate the violations rather than commas?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 6 of 10 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java, line 444 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJm5q4LAoOufsWWwU7a:-MJmKXRj6Cd6ZD-gLrV8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/test/java/google/registry/tools/CreateRegistrarCommandTest.java#L444)):*

> ```
>             "Certificate is expired., Certificate validity period is too long; it must be less"
>                 + " than or equal to 398 days.\n");
> ```
Seems like maybe it is better to use newlines to separate the violations rather than commas?
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 6 of 10 files reviewed, 12 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 17 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmBLl49JTJuAGFBdUL:-MJmBLl49JTJuAGFBdUM:b-oiy8n0) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/RegistryConfig.java#L17)):*
> ```Java
> package google.registry.config;
> 
> import static com.google.common.base.Suppliers.memoize;
> ```

I don't know where to put this comment, but there's a TODO to add injection on `CertificateChecker` that should be removed since we are in fact not going to implement injection at that site if this is the approach we're taking.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 451 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmAFd968fjEJ0icrTL:-MJmAFd968fjEJ0icrTM:b-7rgms4) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/files/default-config.yaml#L451)):*
> ```YAML
> 
> certChecker:
>   maxValidityDays :
> ```

Extra space.

Also, how about `maxValidityDaysSchedule` or similar, because `maxValidityDays` sounds like it'd be a single int which this isn't.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 454 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmAOd8-kAQT6syZYB2:-MJmAOd8-kAQT6syZYB3:bnz5oik) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/files/default-config.yaml#L454)):*
> ```YAML
>     "START_OF_TIME": 825
>     "2020-09-01T00:00:00Z": 398
>   daysToExpiration: 30
> ```

It's not clear without context what this does exactly.  How about `expirationWarningDays`?  (And the comment will explain more in depth.)
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 450 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmLD9N7Fp9C5-0ftZ1:-MJmLD9N7Fp9C5-0ftZ2:b-ywjaya) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/files/default-config.yaml#L450)):*
> ```YAML
>   username: toolusername
> 
> certChecker:
> ```

How about naming this `sslCertificateValidation`
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 450 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmLl6W89SRseZOTkJ_:-MJmLl6W89SRseZOTkJa:bh739x8) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/files/default-config.yaml#L450)):*
> ```YAML
>   username: toolusername
> 
> certChecker:
> ```

Every line of this new section needs a comment explaining what that line is (see above config sections).
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 367 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmMVg55z-Hm1AzukcJ:-MJmMVg55z-Hm1AzukcK:b-y594s1) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L367)):*
> ```Java
>       if (clientCertificateFilename != null) {
>         String asciiCert = new String(Files.readAllBytes(clientCertificateFilename), US_ASCII);
>         if (!asciiCert.equals("")) {
> ```

Add a comment explaining that this exists to allow existing certs to be removed without providing a replacement (I assume that's the purpose).
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 380 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmMoN-0dMyGMdKa7n6:-MJmMoN-0dMyGMdKa7n7:b-nh6pja) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L380)):*
> ```Java
>                     .collect(toImmutableSet());
>             throw new CertificateException(
>                 displayMessages.toString().replace("[", "").replace("]", "\n").replace(", ", "\n"));
> ```

There's a better way to do this than mangling the `Set.toString()` output that you never wanted in the first place.  See `Collectors.joining()` -- you can turn the `Stream<String>` directly into a joined String (using delimiters) without ever having to make a `Set` in the first place.
___
*[core/src/main/java/google/registry/config/CertificateCheckerModule.java, line 27 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJm2UkG1EbunWFtHZdS:-MJm2UkG1EbunWFtHZdT:b1ierpc) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/CertificateCheckerModule.java#L27)):*
> ```Java
> 
> @Module
> public abstract class CertificateCheckerModule {
> ```

Needs Javadoc.
___
*[core/src/main/java/google/registry/config/CertificateCheckerModule.java, line 36 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmKdr-01XbvyEX_rOA:-MJmKdr-01XbvyEX_rOB:bjez9fv) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/CertificateCheckerModule.java#L36)):*
> ```Java
>       @Config("minimumRsaKeyLength") int minimumRsaKeyLength) {
>     return new CertificateChecker(
>         validityDaysMap, daysToExpiration, minimumRsaKeyLength, new SystemClock());
> ```

`Clock`s are themselves provided, so rather than constructing a new `SystemClock` here, it needs to be in the method parameter list as a dependency.
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdpP1uBlN48YGSCwl-:-MJmDiWyBGFeinmhjIhd:b-dchocb) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L28)):*

moved the module to the config folder
</blockquote></details>

`config` isn't the best package for this but it may be the only one that's workable depending on which packages exactly we plan on using `CertificateChecker` from.

Which packages exactly do we plan on using `CertificateChecker` from?  In this PR we have `tools`, and I'm sure we're gonna need it in `ui.server.registrar`.  Anything else?  In the future when we get to CA support we'll presumably have the proxy start passing along the cert's common name(s) in addition to the cert hash, and then we'll use one XOR the other of those to determine if the certificate is valid for the claimed registrar client ID.  So that validation will happen in `TlsCredentials.validateCertificate()` (what's already handling the cert hash checking), which is in `flows`.

So besides `tools.*`, `ui.*`, and `flows.*`, is there anywhere else that `CertificateChecker` will plausibly be needed from?  If not, given that all of these use are either `flows` or depend on it, why not put this module there?

Separately, is there a particular reason that `CertificateChecker` needs to be in the `utils` project at all?  Why not `core`?  That seems like it would make everything a bit easier (and the `CertificateChecker` could then simply `@Inject` itself via its own constructor, and have the relevant `@Config` annotations on the constructor's parameters).  Do we have any expectation of using `CertificateChecker` from outside of `core` in the future?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 6 of 10 files reviewed, 12 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 17 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmBLl49JTJuAGFBdUL:-MJmO8XCCMT8YT1muPdN:btregqu) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/RegistryConfig.java#L17)):*

I don't know where to put this comment, but there's a TODO to add injection on `CertificateChecker` that should be removed since we are in fact not going to implement injection at that site if this is the approach we're taking.
</blockquote></details>

If we do take this approach anyway (see my other comment).

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 6 of 10 files reviewed, 15 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1353 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmOEuD23IKccTds0n5:-MJmOEuD23IKccTds0n6:bju2bv8) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/RegistryConfig.java#L1353)):*
> ```Java
> 
>     @Provides
>     @Config("validityDaysMap")
> ```

How about `maxValidityDaysSchedule` or similar.
___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1356 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmOP564Hmk-u8CT65U:-MJmOP564Hmk-u8CT65V:bg49dia) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/RegistryConfig.java#L1356)):*
> ```Java
>     public static ImmutableSortedMap<DateTime, Integer> provideValidityDaysMap(
>         RegistryConfigSettings config) {
>       Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
> ```

You may as well just inline this since you're only using it in one place.
___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1357 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmOgJA5abQ7DpdJa1S:-MJmOgJA5abQ7DpdJa1T:b-f2oals) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/RegistryConfig.java#L1357)):*
> ```Java
>         RegistryConfigSettings config) {
>       Map<String, Integer> stringMap = config.certChecker.maxValidityDays;
>       ImmutableSortedMap.Builder<DateTime, Integer> validityDaysMap =
> ```

You should be able to do this entire thing in a single statement (without needing to declare a Builder) by just using a Stream and then using the `ImmutableSortedMap.toImmutableSortedMap` Collector.  It should look something like this:

```
config.certChecker.maxValidityDays.entrySet().stream()......collect(toImmutableSortedMap(naturalOrder(), ...)
```
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 450 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmLD9N7Fp9C5-0ftZ1:-MJmPTYg4obSihoEiLUo:b-kzlghu) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/files/default-config.yaml#L450)):*

How about naming this `sslCertificateValidation`
</blockquote></details>

The reasoning being that this is more informative from the context of someone using the system who isn't intimately familiar with its architecture (and shouldn't need to be).

Separately, we'll probably want to do a larger refactoring of these at some point because we're ballooning up to too many separate top-level categories that have natural groupings among them.  E.g. this should probably be lumped in together with several of the other sections such as `oAuth` under a broader `security` section.  But that's low priority and has been put off for years at this point.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 6 of 10 files reviewed, 15 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 17 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmBLl49JTJuAGFBdUL:-MJmgBKw55WXEWSNUF7d:byje5ay) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/RegistryConfig.java#L17)):*

If we do take this approach anyway (see my other comment).
</blockquote></details>

removed the TODO
___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1353 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmOEuD23IKccTds0n5:-MJmUh1E4nih6toaYADX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/RegistryConfig.java#L1353)):*

How about `maxValidityDaysSchedule` or similar.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1356 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmOP564Hmk-u8CT65U:-MJmUp3l2o9c153Go6KI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/RegistryConfig.java#L1356)):*

You may as well just inline this since you're only using it in one place.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 1357 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmOgJA5abQ7DpdJa1S:-MJmZFQ71PjFJNXUMBRA:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/RegistryConfig.java#L1357)):*

You should be able to do this entire thing in a single statement (without needing to declare a Builder) by just using a Stream and then using the `ImmutableSortedMap.toImmutableSortedMap` Collector.  It should look something like this:

```
config.certChecker.maxValidityDays.entrySet().stream()......collect(toImmutableSortedMap(naturalOrder(), ...)
```
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 451 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmAFd968fjEJ0icrTL:-MJm_dsS4VjFAcH0iQMp:b-896fix) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/files/default-config.yaml#L451)):*

Extra space.

Also, how about `maxValidityDaysSchedule` or similar, because `maxValidityDays` sounds like it'd be a single int which this isn't.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 454 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJmAOd8-kAQT6syZYB2:-MJmakSE6LIYyjkmU3jM:b-896fix) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/files/default-config.yaml#L454)):*

It's not clear without context what this does exactly.  How about `expirationWarningDays`?  (And the comment will explain more in depth.)
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 450 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmLD9N7Fp9C5-0ftZ1:-MJmZojm3UdE5T1Nehhd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/files/default-config.yaml#L450)):*

The reasoning being that this is more informative from the context of someone using the system who isn't intimately familiar with its architecture (and shouldn't need to be).

Separately, we'll probably want to do a larger refactoring of these at some point because we're ballooning up to too many separate top-level categories that have natural groupings among them.  E.g. this should probably be lumped in together with several of the other sections such as `oAuth` under a broader `security` section.  But that's low priority and has been put off for years at this point.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 450 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmLl6W89SRseZOTkJ_:-MJmc88l1Cz6RLztf5l_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/files/default-config.yaml#L450)):*

Every line of this new section needs a comment explaining what that line is (see above config sections).
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 367 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmMVg55z-Hm1AzukcJ:-MJmcrkk7Uqw_NWQ5UVX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L367)):*

Add a comment explaining that this exists to allow existing certs to be removed without providing a replacement (I assume that's the purpose).
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 380 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmMoN-0dMyGMdKa7n6:-MJmfJvaEvmGXTlxFbMD:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L380)):*

There's a better way to do this than mangling the `Set.toString()` output that you never wanted in the first place.  See `Collectors.joining()` -- you can turn the `Stream<String>` directly into a joined String (using delimiters) without ever having to make a `Set` in the first place.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/CertificateCheckerModule.java, line 27 at r2](https://reviewable.io/reviews/google/nomulus/837#-MJm2UkG1EbunWFtHZdS:-MJmS4v_2MK7jK0mJTyS:b-896fix) ([raw file](https://github.com/google/nomulus/blob/88ebe698feeaed5c08e564bbf95316598dcc6849/core/src/main/java/google/registry/config/CertificateCheckerModule.java#L27)):*

Needs Javadoc.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/config/CertificateCheckerModule.java, line 36 at r3](https://reviewable.io/reviews/google/nomulus/837#-MJmKdr-01XbvyEX_rOA:-MJmULlq6GndpHYf9RwU:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f1bee6cd2fb2a0ab77f282d35f547837f9670887/core/src/main/java/google/registry/config/CertificateCheckerModule.java#L36)):*

`Clock`s are themselves provided, so rather than constructing a new `SystemClock` here, it needs to be in the method parameter list as a dependency.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 2 of 11 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 452 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0drER--OOY6P0SihN:-MK0drER--OOY6P0SihO:b-8rbzs7) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/config/files/default-config.yaml#L452)):*
> ```YAML
> # Configuration options for checking SSL certificates.
> sslCertificateValidation:
>   # A map specifying the maximum amount of days the certificate can be valid. The entry key is the date closest before the date the certificate was issued and the entry value is the applicable maximum validity days for that certificate.
> ```

This line is too long.  (It's much longer than any existing line in the file.)  It needs to be wrapped to whatever the convention for this file is, which can be figured out from context. 
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 366 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0gGoqCw-x6nyQpJ27:-MK0gGor5KhQu0XhR4RY:b3hwkx2) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L366)):*
> ```Java
>         builder.setIpAddressAllowList(ipAllowListBuilder.build());
>       }
>       if (clientCertificateFilename != null) {
> ```

Does this even work?  It occurs to me that there might not be tests of this functionality below?  (There should be.)

Thinking about how you'd want to null out the client cert, wouldn't you be passing "" on the command line?  But there's no check for that?  It would try to load that file, which would have to have empty contents in order to null out the configured cert?  That seems unnecessarily circuitous.  I would expect to be able to do:

```
$ nomulus update_registrar fooreg --cert_file "" --failover_cert_file ""
```
 in order to turn off a registrar's ability to log in via EPP by removing all valid certs.  Instead, with this code as written, I'd have to do something like this?

```
$ echo "" > /tmp/blank_file
$ nomulus update_registrar fooreg --cert_file /tmp/blank_file --failover_cert_file /tmp/blank_file
```
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 393 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0eFdw35qAg1QTYEZa:-MK0eFdw35qAg1QTYEZb:b-xxth9y) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L393)):*
> ```Java
>             new String(Files.readAllBytes(failoverClientCertificateFilename), US_ASCII);
>         if (!asciiCert.equals("")) {
>           X509Certificate certificate =
> ```

You've duplicated this block of code here twice.  It should be extracted out to a helper method in this class.  Something that would make the two callsites look like:

```
if (clientCertificateFilename != null) {
  builder.setClientCertificate(loadAndValidateCert(clientCertificateFilename), now);
}
// [ ... ]
if (failoverClientCertificateFilename != null) {
  builder.setFailoverClientCertificate(loadAndValidateCert(failoverClientCertificateFilename), now);
}
```

(Though, per the comment above, the clause inside the conditional may need to be different from what it is currently.)
___
*[core/src/main/java/google/registry/tools/RegistryToolComponent.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdrvgQ9KtabBq_XinI:-MK0hRL2EhzLURHt6AkT:bzcb8i6) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/RegistryToolComponent.java#L94)):*

Is it? I think autoformatting added the newlines back.
</blockquote></details>

It may just be worth biting the bullet here and running the auto-formatter on the whole file.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 2 of 11 files reviewed, 6 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdpP1uBlN48YGSCwl-:-MK0pVs65w8YbtXfh7XY:b-bxz5hu) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L28)):*

Yes, it may make sense to move CertificateChecker to core/ if possible. I'm going to leave that though to another PR since I think it will require quite a bit of refactoring, and I want to make sure I can get to adding the UI part so It's live before November 1.
</blockquote></details>

Add a TODO to move it then?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: 2 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 452 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0drER--OOY6P0SihN:-MK0nHzR22u0H_JsYY5z:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/config/files/default-config.yaml#L452)):*

This line is too long.  (It's much longer than any existing line in the file.)  It needs to be wrapped to whatever the convention for this file is, which can be figured out from context. 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 393 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0eFdw35qAg1QTYEZa:-MK0s7G03sVve8C9bAZH:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L393)):*

You've duplicated this block of code here twice.  It should be extracted out to a helper method in this class.  Something that would make the two callsites look like:

```
if (clientCertificateFilename != null) {
  builder.setClientCertificate(loadAndValidateCert(clientCertificateFilename), now);
}
// [ ... ]
if (failoverClientCertificateFilename != null) {
  builder.setFailoverClientCertificate(loadAndValidateCert(failoverClientCertificateFilename), now);
}
```

(Though, per the comment above, the clause inside the conditional may need to be different from what it is currently.)
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/RegistryToolComponent.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdrvgQ9KtabBq_XinI:-MK0uwkI3-UvCr0NfZWn:b-896fjx) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/RegistryToolComponent.java#L94)):*

It may just be worth biting the bullet here and running the auto-formatter on the whole file.
</blockquote></details>

Done

jianglai (approved):
> Reviewed 5 of 8 files at r4, 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 366 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0gGoqCw-x6nyQpJ27:-MK2SRkp1bxF6gYv6X8T:bmyrh8t) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L366)):*

It seems how the tool works, it is not possible to use an empty string for an arg that looks for a file name. This is checked in the validate() method in PathParameter.java (https://sourcegraph.com/github.com/google/nomulus@3534a14/-/blob/core/src/main/java/google/registry/tools/params/PathParameter.java#L84). The test that tests certificate upload with an empty file is at https://sourcegraph.com/github.com/google/nomulus@3534a14/-/blob/core/src/test/java/google/registry/tools/UpdateRegistrarCommandTest.java#L255
</blockquote></details>

Hmm, so perhaps the way this flag is configured, if you don't specify it, the object here would be empty, but there's no way to unset it? I guess we never envisioned this as a way to remove a certificate? It is only meant to not specify a cert when we set up registrars in test environment?

Can you test locally to verify the behavior?
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdpP1uBlN48YGSCwl-:-MK2QYYQ6VLHtg3SzV0q:b-g8wjbd) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L28)):*

Add a TODO to move it then?
</blockquote></details>

I'm still not sold on the idea that when we start allowing CA certs we still need the registrars to provide us with the certs out of band, which defeats one of the major selling points of using CA certs -- that the registrars don't have to upload the certs periodically, which is not automated.

So if we don't store the certs ourselves, the check will have to be performed on the proxy, which means the checker need to be in `util` in order to not pull the entire `core` package as a dependency of `proxy`. Or alternatively we can have the proxy pass the entire cert to the backend, which we probably shouldn't do as a HTTP header on **every** request, but rather on initial EPP hello only.

We don't have to sort this out in this PR.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/837)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 366 at r4](https://reviewable.io/reviews/google/nomulus/837#-MK0gGoqCw-x6nyQpJ27:-MKFm_rP8F5cjm79ZO71:bseyyzs) ([raw file](https://github.com/google/nomulus/blob/fa4d4329f9ea8420e38c30832a883a4aece21286/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L366)):*

The way to unset it would be to use an empty file
</blockquote></details>

Huh, so this behavior is pre-existing?  I guess I'll add it onto my backlog to make it easier to unset a cert, because needing to pass along an empty file is circuitous.  Anyway, we can get this PR in as is.
___
*[core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java, line 367 at r5](https://reviewable.io/reviews/google/nomulus/837#-MKFlzft8PSNeBXJaLZW:-MKFlzft8PSNeBXJaLZX:bo5bdkw) ([raw file](https://github.com/google/nomulus/blob/3071d85c2bc7e9c8eeeee89c7c25cc9bb43cd5dc/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java#L367)):*
> ```Java
>       }
>       if (clientCertificateFilename != null) {
>         String asciiCert = new String(Files.readAllBytes(clientCertificateFilename), US_ASCII);
> ```

If you change `verifyCertificate()` below to take the file name instead of the certificate string per my previous suggestion, then you can extract out this line too.  And if it returns the loaded certificate instead of returning void, then you can inline it into the builder call.
___
*[core/src/main/java/google/registry/tools/CertificateCheckerModule.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/837#-MJdpP1uBlN48YGSCwl-:-MKFjP4l8rQznmcotczd:bmkq0jh) ([raw file](https://github.com/google/nomulus/blob/e6b120f4797b2e4ec96ca6317327b0047859bae5/core/src/main/java/google/registry/tools/CertificateCheckerModule.java#L28)):*

I'm still not sold on the idea that when we start allowing CA certs we still need the registrars to provide us with the certs out of band, which defeats one of the major selling points of using CA certs -- that the registrars don't have to upload the certs periodically, which is not automated.

So if we don't store the certs ourselves, the check will have to be performed on the proxy, which means the checker need to be in `util` in order to not pull the entire `core` package as a dependency of `proxy`. Or alternatively we can have the proxy pass the entire cert to the backend, which we probably shouldn't do as a HTTP header on **every** request, but rather on initial EPP hello only.

We don't have to sort this out in this PR.
</blockquote></details>

I'm definitely not proposing that registrars would have to provide the CA certs out-of-band.  That would of course defeat most of the purpose of it.

However, I'm not seeing how it's possible to perform the check on the proxy.  That would entail adding lots of stuff to the proxy that it doesn't yet have and that I don't think we want it to have, like needing to have access to our main Nomulus DB.  I'd much rather just have the proxy be as dumb as possible.

Why does the proxy need to validate the SSL cert at all?  Let the built-in SSL library handle the basics (like making sure that the cert is valid, not expired, etc.), and then pass along the rest of the info (either individually or just the entire cert) to the main nomulus app, on new connection instantiation only, for validation.  That seems a lot easier to me than having to put a lot more logic into the proxy in having it use the Nomulus configuration files and DB access to perform the full validation itself.  Ideally we'd like to be able to delete the proxy entirely some day once we're on GKE, so the validation is gonna have to be in core Nomulus eventually.

Agreed that we can get this PR in for now and then move things around later, but I don't see a reason why it needs to be in the `utils` project.  Everything gets much simpler if all of the certificate-specific stuff lives in `core...flows`.

PR #840, opened 2020-10-15
Add a SQL schema and DAO for KmsSecretRevision

gbrodman:
The dual-object nature of KmsSecret and KmsSecretRevision will not be
necessary once we have moved to SQL. In that world, the only object will
be the one now called KmsSecretRevision. KmsSecretRevision already
stores its parent so all we need to do is convert that key to the String
secretName (or from the secretName to the key, if loading from SQL) and
select the max revision ID for a given secret name.

In a future PR, we will add a dual-writing DAO to these objects and
perform the dual writes, similar to how ReservedList functions.

Note: we will dual-write these because we don't care about their history and we're slightly modifying the structure in which we store them anyway (removing the original KmsSecret class) 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/840)


hstonec:
I have two general questions:

1. As we are still using Datastore to store the password for SQL DB, it seems that we may not be able to migrate that part to a sql table right? 

2. Should we consider using Secret Manger instead of migrating this to cloud SQL? @weiminyu do you have any thoughts on this?

gbrodman:
Good questions. I pinged Weimin about the secret manager, but you're right in that his initial thoughts were basically to move passwords there and keep this around for crypto keys.

gbrodman:
OK I talked with Weimin and there's two main tasks we want to do

1. Move the password storage to Secret Manager
2. Add dual write / read to KmsSecret(Revision)
3. Make SQL primary for the dual write / read for KmsSecret(Revision)

Technically, 1 and (2/3) are entirely separate but in practice we'll want to do 1 before 3. We can do 1 and 2 concurrently though without them affecting each other

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @hstonec)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 443 at r1](https://reviewable.io/reviews/google/nomulus/840#-MKGJqEQ2vOIiU_ypB8K:-MKGJqEQ2vOIiU_ypB8L:b-fb1b2w) ([raw file](https://github.com/google/nomulus/blob/2976666843d89ad52418e985be0584cbb00f54bc/db/src/main/resources/sql/schema/db-schema.sql.generated#L443)):*
> ```
> 
>     create table "KmsSecret" (
>        revision_key int8 not null,
> ```

`revision_key` isn't the best name for this field.  `revision_id` maybe?
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 446 at r1](https://reviewable.io/reviews/google/nomulus/840#-MKGK5ra9zdHAS-RD2aa:-MKGK5ra9zdHAS-RD2ab:b7hstdt) ([raw file](https://github.com/google/nomulus/blob/2976666843d89ad52418e985be0584cbb00f54bc/db/src/main/resources/sql/schema/db-schema.sql.generated#L446)):*
> ```
>         creation_time timestamptz not null,
>         encrypted_value text not null,
>         kms_crypto_key_version_name text not null,
> ```

Can you think of a better/shorter name for this one?

hstonec:
> OK I talked with Weimin and there's three main tasks we want to do
> 
> 1. Move the password storage to Secret Manager
> 2. Add dual write / read to KmsSecret(Revision)
> 3. Make SQL primary for the dual write / read for KmsSecret(Revision)
> 
> Technically, 1 and (2/3) are entirely separate but in practice we'll want to do 1 before 3. We can do 1 and 2 concurrently though without them affecting each other

Sounds good, it looks like we will refactor this PR to use secret manager, right?

gbrodman:
> > OK I talked with Weimin and there's three main tasks we want to do
> > 
> > 1. Move the password storage to Secret Manager
> > 2. Add dual write / read to KmsSecret(Revision)
> > 3. Make SQL primary for the dual write / read for KmsSecret(Revision)
> > 
> > Technically, 1 and (2/3) are entirely separate but in practice we'll want to do 1 before 3. We can do 1 and 2 concurrently though without them affecting each other
> 
> Sounds good, it looks like we will refactor this PR to use secret manager, right?

No, this PR can happen concurrently with the secret manager efforts.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @hstonec)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 443 at r1](https://reviewable.io/reviews/google/nomulus/840#-MKGJqEQ2vOIiU_ypB8K:-MKLUz-cCZbaSRyEvf7u:b-a334ri) ([raw file](https://github.com/google/nomulus/blob/2976666843d89ad52418e985be0584cbb00f54bc/db/src/main/resources/sql/schema/db-schema.sql.generated#L443)):*

`revision_key` isn't the best name for this field.  `revision_id` maybe?
</blockquote></details>

Yeah, the downside that I originally noticed is that in the DAO you have to reference the Java variable name, not the SQL variable name. As a result, it looks a bit awkward to reference `revisionKey` in the quasi-SQL queries there, but that's probably fine (the tests will fail if we don't change the DAO text when we rename the field after getting rid of datastore) 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 446 at r1](https://reviewable.io/reviews/google/nomulus/840#-MKGK5ra9zdHAS-RD2aa:-MKLW1Ao6msBz063lKSt:bvrxxyh) ([raw file](https://github.com/google/nomulus/blob/2976666843d89ad52418e985be0584cbb00f54bc/db/src/main/resources/sql/schema/db-schema.sql.generated#L446)):*

Can you think of a better/shorter name for this one?
</blockquote></details>

we can remove the "kms" but other than that it seems fairly descriptive. 

`https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys.cryptoKeyVersions`

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840#-:-MKjoq3c4u2LkXWe1zPM:bnfp4nl)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @hstonec)

hstonec (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @hstonec)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 586 at r2](https://reviewable.io/reviews/google/nomulus/840#-MKkxWkg-wb17Ngu_BXT:-MKkxWkg-wb17Ngu_BXU:b-ltp33p) ([raw file](https://github.com/google/nomulus/blob/58562c07f1206c8c6213e171e6957be673977712/db/src/main/resources/sql/schema/nomulus.golden.sql#L586)):*
> ```
>     encrypted_value text NOT NULL,
>     crypto_key_version_name text NOT NULL,
>     secret_name text NOT NULL
> ```
I am still a little bit confused about these columns, my understanding is we will store the KMS encrypted password to Secret Manager, and store its identifier in this table right? If so, do we still need `encrypted_value` column?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 586 at r2](https://reviewable.io/reviews/google/nomulus/840#-MKkxWkg-wb17Ngu_BXT:-MKl-EJa89H_Jd1Sg6la:b-hzh8xm) ([raw file](https://github.com/google/nomulus/blob/58562c07f1206c8c6213e171e6957be673977712/db/src/main/resources/sql/schema/nomulus.golden.sql#L586)):*

> ```
>     encrypted_value text NOT NULL,
>     crypto_key_version_name text NOT NULL,
>     secret_name text NOT NULL
> ```
I am still a little bit confused about these columns, my understanding is we will store the KMS encrypted password to Secret Manager, and store its identifier in this table right? If so, do we still need `encrypted_value` column?
</blockquote></details>

There are two distinct type of things we currently store in the KmsSecretRevisions -- passwords, and crypto keys. We'll move the passwords to Secret Manager, and keep the crypto keys here. You can see the different types of passwords / keys in `KmsKeyring`

hstonec (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 586 at r2](https://reviewable.io/reviews/google/nomulus/840#-MKkxWkg-wb17Ngu_BXT:-MKl3ivr4nwal_TgBjuX:bdwkwik) ([raw file](https://github.com/google/nomulus/blob/58562c07f1206c8c6213e171e6957be673977712/db/src/main/resources/sql/schema/nomulus.golden.sql#L586)):*

There are two distinct type of things we currently store in the KmsSecretRevisions -- passwords, and crypto keys. We'll move the passwords to Secret Manager, and keep the crypto keys here. You can see the different types of passwords / keys in `KmsKeyring` 
</blockquote></details>

Why not just save all of them in Secret Manager then? The crypto keys themselves are also KMS encrypted right? Which indicates we do care about their security.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 586 at r2](https://reviewable.io/reviews/google/nomulus/840#-MKkxWkg-wb17Ngu_BXT:-MKlLy-n0cWFNOcBvJsO:bea0ydw) ([raw file](https://github.com/google/nomulus/blob/58562c07f1206c8c6213e171e6957be673977712/db/src/main/resources/sql/schema/nomulus.golden.sql#L586)):*

Why not just save all of them in Secret Manager then? The crypto keys themselves are also KMS encrypted right? Which indicates we do care about their security.
</blockquote></details>

Quoting Weimin from a while back, 

```I think we should take this opportunity to make a distinction between crypto keys and passwords. Revisions make sense to the former but not the latter.```

I think this makes sense

gbrodman (commented):
@hstonec did you have any more thoughts on this? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840#-:-MKvIF3s3Tniua8po-qc:bw4drao)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @hstonec)

hstonec (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/840)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 586 at r2](https://reviewable.io/reviews/google/nomulus/840#-MKkxWkg-wb17Ngu_BXT:-MKvJISxC4UkDZvXaBiV:bjx0uvw) ([raw file](https://github.com/google/nomulus/blob/58562c07f1206c8c6213e171e6957be673977712/db/src/main/resources/sql/schema/nomulus.golden.sql#L586)):*

Quoting Weimin from a while back, 

```I think we should take this opportunity to make a distinction between crypto keys and passwords. Revisions make sense to the former but not the latter.```

I think this makes sense
</blockquote></details>

sounds reasonable to me.

PR #842, opened 2020-10-19
Use a SQL date object for LocalDates

gbrodman:
The "date" definition in the Column annotation is because otherwise we don't have control over whether the SQL type ends up as a date or as a timestamptz (currently it seems like it'd end up as a timestamptz) 

We used Strings before during the internship as a way to simplify things (since it was time sensitive) but it is more correct to use a date.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/842)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/842)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/converter/LocalDateConverter.java, line 23 at r1](https://reviewable.io/reviews/google/nomulus/842#-MK65vvh1B4WfpZROM2A:-MK65vvi-kJmRno76kO0:b-mysmrb) ([raw file](https://github.com/google/nomulus/blob/9e6e7b21ff2953e5c8f93b295abfe96ee675da98/core/src/main/java/google/registry/persistence/converter/LocalDateConverter.java#L23)):*
> ```
> java.sql
> ```
Optional: omit the package, since it's an import (and external so it won't resolve correctly in any case)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/842)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/LocalDateConverter.java, line 23 at r1](https://reviewable.io/reviews/google/nomulus/842#-MK65vvh1B4WfpZROM2A:-MK6Cpmg48qTl_0LlGrS:b-f0s3gd) ([raw file](https://github.com/google/nomulus/blob/9e6e7b21ff2953e5c8f93b295abfe96ee675da98/core/src/main/java/google/registry/persistence/converter/LocalDateConverter.java#L23)):*

> ```
> java.sql
> ```
Optional: omit the package, since it's an import (and external so it won't resolve correctly in any case)
</blockquote></details>

Sure, since we use the imported Date elsewhere in this file (and technically, java.sql.Date extends java.time.Date)

PR #846, opened 2020-10-21
Restore ofy keys in GracePeriod objects

mindhog:
Restore the ofy keys when loading GracePeriod object from SQL.  There's no
clear way to do this using the normal approach (fix-up during a PostLoad
method) because fixups to these violate immutability after hibernate has                                                                        
already obtained their hash values.  Instead, we force reconstitution of the
ofy keys in all public methods that access them (including equals() and
hashCode()) so that they can be generated before an invalid hash is generated.                                                                  

As part of this change, convert the GracePeriod id from an autogenerated
sequence to a UUID allocated from ObjectifyService and enhance ImmutableObject
to allow it to exclude certain fields from hash/equals and print.

The ImmutableObject enhancements are necessary because we compare grace
periods against locally created test objects in a number of unit tests and
there's no way this can work with GracePeriods loaded from SQL currently, as
they will have an identifier field generated from the database and the test
objects will have an identifier field of null (or a new unique value, after
this change).

Removing autogeneration from GracePeriod ids ended up being likely not
strictly necessary for this change (it was a consequence of an earlier
iteration).  However, it does alleviate the problem of mutation of an
immutable object after creation and is more in line with how we've decided to
allocate other identifiers.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/846)


hstonec (changes requested):
> Reviewed 17 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/846)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/846#-MKGWM3ZBlu7uBFCEvgD:-MKGWM3ZBlu7uBFCEvgE:b-5fjid2) ([raw file](https://github.com/google/nomulus/blob/6e1fb1b9f53319b20ca0e1827c48754098ea6947/core/src/main/java/google/registry/model/domain/GracePeriod.java#L122)):*
> ```
> clone.id = ObjectifyService.allocateId();
> ```
I was wondering why we need to assign a new id for the cloned object?
___
*[core/src/main/java/google/registry/model/domain/GracePeriodBase.java, line 147 at r1](https://reviewable.io/reviews/google/nomulus/846#-MKGWkC8-bkM_It--MsQ:-MKGWkC8-bkM_It--MsR:b8f7b86) ([raw file](https://github.com/google/nomulus/blob/6e1fb1b9f53319b20ca0e1827c48754098ea6947/core/src/main/java/google/registry/model/domain/GracePeriodBase.java#L147)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>   /**
>    * Override {@link ImmutableObject#getSignificantFields()} to exclude "id", which breaks equality
>    * testing in the unit tests.
>    */
>   @Override
>   protected Map<Field, Object> getSignificantFields() {
> ```
</details>

We have this because when we clone the object, we assign a new id right?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/846)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/846#-MKGWM3ZBlu7uBFCEvgD:-MKKBoZ76wpNv-reKlQf:b-mchr9q) ([raw file](https://github.com/google/nomulus/blob/6e1fb1b9f53319b20ca0e1827c48754098ea6947/core/src/main/java/google/registry/model/domain/GracePeriod.java#L122)):*

> ```
> clone.id = ObjectifyService.allocateId();
> ```
I was wondering why we need to assign a new id for the cloned object?
</blockquote></details>

This is after loading from objectify, where we don't save the id field.  So the id field has not been resetored.
___
*[core/src/main/java/google/registry/model/domain/GracePeriodBase.java, line 147 at r1](https://reviewable.io/reviews/google/nomulus/846#-MKGWkC8-bkM_It--MsQ:-MKKBEz59ozOmTeV3HkL:bqeoa1x) ([raw file](https://github.com/google/nomulus/blob/6e1fb1b9f53319b20ca0e1827c48754098ea6947/core/src/main/java/google/registry/model/domain/GracePeriodBase.java#L147)):*

> ```
>   /**
>    * Override {@link ImmutableObject#getSignificantFields()} to exclude "id", which breaks equality
>    * testing in the unit tests.
>    */
>   @Override
>   protected Map<Field, Object> getSignificantFields() {
> ```
We have this because when we clone the object, we assign a new id right?
</blockquote></details>

it's more because there are numerous places where we create objects for comparison in tests. We can't do the usual trick of "compare excluding fields" because we compare the entire collection of grace period objects.

hstonec (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/846)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/GracePeriodBase.java, line 171 at r1](https://reviewable.io/reviews/google/nomulus/846#-MK_GjMSBst02J47wuZp:-MK_GjMSBst02J47wuZq:bu7z46a) ([raw file](https://github.com/google/nomulus/blob/6e1fb1b9f53319b20ca0e1827c48754098ea6947/core/src/main/java/google/registry/model/domain/GracePeriodBase.java#L171)):*
> ```
>  private final synchronized void restoreOfyKeys() {
> ```
nit: Is there any case we need synchronization?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/846)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/domain/GracePeriodBase.java, line 171 at r1](https://reviewable.io/reviews/google/nomulus/846#-MK_GjMSBst02J47wuZp:-MK_Lyaa72wmYLtxHmCm:b-1ergk1) ([raw file](https://github.com/google/nomulus/blob/6e1fb1b9f53319b20ca0e1827c48754098ea6947/core/src/main/java/google/registry/model/domain/GracePeriodBase.java#L171)):*

> ```
>  private final synchronized void restoreOfyKeys() {
> ```
nit: Is there any case we need synchronization?
</blockquote></details>

Since the objects are immutable, it is at least theoretically possible that the fields that are reconstituted here could be accessed from multiple threads (in fact, we rely on the thread-safety of immutable objects in a number of contexts).  I'm not sure how much of an issue this is in practice, but I wanted to do the right thing :-)

PR #852, opened 2020-10-27
Move CertificateChecker to core/

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/852)


jianglai (commented):
Have we thought this through? Proxy cannot depend on core. So we'll pass the certificate from the proxy to GAE to validate?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKgTqef30RXZM5Uq_KE:b-s7fsxq)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @CydeWeys and @jianglai)

CydeWeys (commented):
I'm not seeing how proxy is ever going to be able to independently run the full validation.  Here are all the reasons it would be hard to do:

* It would require us having access to the nomulus configuration for stuff like min key length, max validity days schedule, etc.  Practically speaking we'd need to duplicate this configuration from nomulus into the proxy, and then we'd have these values stored in two separate places.  That's not great.
* Once we add CA support it would require access to the DB as well so we could load the registrar-configured allowed cert common names persisted on the `Registrar` entity.  That we *definitely* don't want to do.

How exactly is the proxy ever going to be able to run the full range of validation?  I don't see how it's possible unless we do all sorts of things that seem clearly worse than simply passing along the SSL cert when new connections are being opened so that Nomulus can handle it.

There's also the issue that our long-term plan is to eventually get rid of the proxy entirely in Registry 3.0 phase 3.5 (full migration to GKE w/ elimination of the proxy), in which case all SSL cert validation will happen inside Nomulus anyway.  If that's the endgame anyway, why not just get there sooner and avoid having to implement validation in the proxy which we're just gonna delete later anyway?  Given that nomulus already has access to its own configuration and its own DB, it's always gonna be easier to implement validation there anyway, which we can see because SSL certificate validation is already happening there anyway!!  We're already passing along the cert hash so that nomulus can validate that it's the right one for the Registrar, because the proxy already can't do these things without a DB connection.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKjkCp47O3-BolaP8hY:b4z451a)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @CydeWeys and @jianglai)

CydeWeys (commented):
I posed these same questions about the feasibility of doing it in the proxy on PR #837, but it was submitted not longer after that so I suppose Lai that you never saw them.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKjnJmRBZEdm5FwhEZ7:bpqiq1f)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (approved):
Yeah, I certainly wasn't implying that we should do this in the proxy, but asking whether we've considered the pros and cons. It's great that you have given it much thought and I agree that it is best that Nomulus handle the cert validation for reasons you outlined. My only concern is that a hash's size is finite whereas a cert's isn't. It mostly likely won't get out of hand, but passing it in its entirety in **every** request (which we currently do for the hashes) probably is not the best choice. 

So I think to err on the side of caution and efficiency, we should only pass the cert in the header in the very first request from the proxy, which must be login. The login flow is where we validate the credentials anyway. This way the proxy still can be "dumb" (in that it doesn't need to understand the content of the EPP request, but just know which one is the first in a connection).

Actually we can do this now with the hashes. I'd suggest that the first order of business to modify the proxy is to only pass the hash in the login flow and make sure that it still works.

> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKmHOSx9X0850jj5wsY:b-erqckn)* status: :shipit: complete! all files reviewed, all discussions resolved

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/certificates/CertificateChecker.java, line 15 at r1](https://reviewable.io/reviews/google/nomulus/852#-MKoVNZn7i05oXfvmzad:-MKoVNZn7i05oXfvmzae:b-m4zaz7) ([raw file](https://github.com/google/nomulus/blob/0a76df0cb04e6c11ec774ca66f81be1161b9d6d0/core/src/main/java/google/registry/flows/certificates/CertificateChecker.java#L15)):*
> ```Java
> 
> package google.registry.flows.certificates;
> ```

`google.registry.flows.certificates` is a fairly long package.  What do you think about `certs` instead?  I don't think it loses any clarity, but it does lose 7 chars.  And everything you're gonna be importing from it will look like `import google.registry.flows.certs.CertificateChecker;` anyway, so the full "Certificate" string is definitely in there and thus doesn't need to be duplicated twice in full in the path.

(And I didn't do an exhaustive check, but near as I can tell, "certificates" would be the single longest label out of any package name in the entire core package.)
___
*[core/src/test/java/google/registry/flows/certificates/CertificateCheckerTest.java, line 209 at r1](https://reviewable.io/reviews/google/nomulus/852#-MKoVDjsBAbgTHgpr38M:-MKoVDjsBAbgTHgpr38N:bz1yxfe) ([raw file](https://github.com/google/nomulus/blob/0a76df0cb04e6c11ec774ca66f81be1161b9d6d0/core/src/test/java/google/registry/flows/certificates/CertificateCheckerTest.java#L209)):*
> ```Java
>     fakeClock.setTo(DateTime.parse("2020-11-01T00:00:00Z"));
>     IllegalArgumentException thrown =
>         Assertions.assertThrows(
> ```

`assertThrows` and `assertThat` should be statically imported (they are in the rest of our codebase).

sarahcaseybot (commented):
ok, I will do a PR to modify the proxy to only send the hash during login

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKpZD-eDgNOrOcZL41h:bbu2noy)* status: 3 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/certificates/CertificateCheckerTest.java, line 209 at r1](https://reviewable.io/reviews/google/nomulus/852#-MKoVDjsBAbgTHgpr38M:-MKpFB1L2uWNCA8T1iJH:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0a76df0cb04e6c11ec774ca66f81be1161b9d6d0/core/src/test/java/google/registry/flows/certificates/CertificateCheckerTest.java#L209)):*

`assertThrows` and `assertThat` should be statically imported (they are in the rest of our codebase).
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKv4Eq15oYQe-zXghdJ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/flows/certificates/CertificateChecker.java, line 15 at r1](https://reviewable.io/reviews/google/nomulus/852#-MKoVNZn7i05oXfvmzad:-MKv4Awv1ZJ1zad1LJ9s:b-ucfkn4) ([raw file](https://github.com/google/nomulus/blob/0a76df0cb04e6c11ec774ca66f81be1161b9d6d0/core/src/main/java/google/registry/flows/certificates/CertificateChecker.java#L15)):*

`google.registry.flows.certificates` is a fairly long package.  What do you think about `certs` instead?  I don't think it loses any clarity, but it does lose 7 chars.  And everything you're gonna be importing from it will look like `import google.registry.flows.certs.CertificateChecker;` anyway, so the full "Certificate" string is definitely in there and thus doesn't need to be duplicated twice in full in the path.

(And I didn't do an exhaustive check, but near as I can tell, "certificates" would be the single longest label out of any package name in the entire core package.)
</blockquote></details>

Thanks.

CydeWeys (commented):
And I don't see why we'd want to modify the proxy to only send the hash during login.  What we need to do is have it pass the full certificate during login, in addition to leaving the hash passing unchanged.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/852#-:-MKv4HVX6G-3HeqdSA2_:bl44evt)* status: :shipit: complete! all files reviewed, all discussions resolved

CydeWeys:
More than a little risky, a lot risky.

Also, it's true that we use the cert hash for validating the registrar via the certificate being used at login.  However, it's also true that we have in the past used the cert hash associated with an ongoing open connection for some purposes (maybe to identify the owner of the connection; I forget the details).  It's in the request past from the proxy and thus it gets logged by GAE.  Before removing it I'd want to make sure that we definitely aren't using it for anything and that we're 100% sure we can always identify the full state of an ongoing connection without it.  I think in particular it was useful on failed login attempts maybe?

We know we want the full certificate in addition to what we already have on login only, and that should be low risk to add (though we should verify that first).  We're not yet sure that we can/should remove the cert hash, but that'll be long in the future, only after we're successfully using the certificate itself for login purposes, not even using the cert hash as a fallback, and after we've ensured that we will always have some other mechanism of determining the owner of an open connection after the cert hash is gone.  Then and only then do we remove the cert hash.

Note that when we switch fully over to just using the full certificate for verification there's gonna be some related functionality that should be removed as well (basically anything that allows setting the cert hash only; I believe there's some in nomulus commands, and the `Registrar` entity itself still has two fields for cert hash that we'd want to get rid of too).

PR #856, opened 2020-10-30
Fix entity annotations for a few classes

gbrodman:
- Abstract classes shouldn't implement DatastoreEntity/SqlEntity
- We aren't persisting Modification in SQL
- Because DelegationSignerData is embedded, we don't need to worry about
converting and persisting it
- DomainDsDataHistory isn't persisted in Datastore

After this, there are only three classes left to fix before we can enable EntityTest

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/856)


hstonec (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/856)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #857, opened 2020-11-02
Add an extension to verify transaction replay

mindhog:
Add ReplayExtension, which can be applied to test suites to verify that
transactions committed to datastore can be replayed to SQL.

This introduces a ReplayQueue class, which serves as a stand-in for the
current lack of replay-from-commit-logs.  It also includes replay logic in
TransactionInfo which introduces the concept of "entity class weights."
Entity weighting allows us store and delete objects in an order that is
consistent with the direction of foreign key and deferred foreign key
relationships.  As a general rule, lower weight classes must have no direct or
indirect non-deferred foreign key relationships on higher weight classes.

It is expected that much of this code will change when the final replay
mechanism is implemented.

NOTE: this code currently breaks the one test that it is applied to
(DomainCreateFlowTest) due to a few outstanding persistence bugs.
Merge after these are resolved.  Also, we'll want to revisit this if the
final replay mechanism is implemented soon.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/857)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -57,6 +57,7 @@ import javax.persistence.Column;
 import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
 import javax.persistence.Table;
+import javax.persistence.Transient;
 import org.joda.time.DateTime;
 
 /** An entity representing an allocation token. */
@@ -105,7 +106,8 @@ public class AllocationToken extends BackupGroupRoot implements Buildable, Datas
   @javax.persistence.Id @Id String token;
 
   /** The key of the history entry for which the token was used. Null if not yet used. */
-  @Nullable @Index VKey<HistoryEntry> redemptionHistoryEntry;
+  // TODO(b/172848495): Remove the "Transient" when we can finally persist and restore this.
+  @Transient @Nullable @Index VKey<HistoryEntry> redemptionHistoryEntry;
 
   /** The fully-qualified domain name that this token is limited to, if any. */
   @Nullable @Index String domainName;
--- a/core/src/main/java/google/registry/model/ofy/CommitLoggedWork.java
+++ b/core/src/main/java/google/registry/model/ofy/CommitLoggedWork.java
@@ -161,6 +161,7 @@ class CommitLoggedWork<R> implements Runnable {
           .addAll(untouchedRootsWithTouchedChildren)
           .build())
       .now();
+    ReplayQueue.addInTests(info);
   }
 
   /** Check that the timestamp of each BackupGroupRoot is in the past. */
--- a/core/src/main/java/google/registry/model/ofy/ReplayQueue.java
+++ b/core/src/main/java/google/registry/model/ofy/ReplayQueue.java
@@ -1,0 +1,46 @@
+// Copyright 2020 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.model.ofy;
+
+import google.registry.config.RegistryEnvironment;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * Implements simplified datastore to SQL transaction replay.
+ *
+ * <p>This code is to be removed when the actual replay cron job is implemented.
+ */
+public class ReplayQueue {
+
+  static ConcurrentLinkedQueue<TransactionInfo> queue =
+      new ConcurrentLinkedQueue<TransactionInfo>();
+
+  static void addInTests(TransactionInfo info) {
+    if (RegistryEnvironment.get() == RegistryEnvironment.UNITTEST) {
+      queue.add(info);
+    }
+  }
+
+  public static void replay() {
+    TransactionInfo info;
+    while ((info = queue.poll()) != null) {
+      info.saveToJpa();
+    }
+  }
+
+  public static void clear() {
+    queue.clear();
+  }
+}
--- a/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
+++ b/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
@@ -21,10 +21,14 @@ import static com.google.common.collect.Maps.filterValues;
 import static com.google.common.collect.Maps.toMap;
 import static google.registry.model.ofy.CommitLogBucket.getArbitraryBucketId;
 import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.googlecode.objectify.Key;
+import google.registry.persistence.VKey;
+import google.registry.schema.replay.DatastoreEntity;
+import google.registry.schema.replay.SqlEntity;
 import java.util.Map;
 import org.joda.time.DateTime;
 
@@ -91,5 +95,50 @@ class TransactionInfo {
         .stream()
         .filter(not(Delete.SENTINEL::equals))
         .collect(toImmutableSet());
+  }
+
+  // Mapping from class name to "weight" (which in this case is the order in which the class must
+  // be "put" in a transaction with respect to instances of other classes).  Lower weight classes
+  // are put first, by default all classes have a weight of zero.
+  static final ImmutableMap<String, Integer> CLASS_WEIGHTS =
+      ImmutableMap.of(
+          "HistoryEntry", -1,
+          "DomainBase", 1);
+
+  // The beginning of the range of weights reserved for delete.  This must be greater than any of
+  // the values in CLASS_WEIGHTS.
+  static final int DELETE_RANGE = 100;
+
+  /** Returns the weight of the entity type in the map entry. */
+  private static int getWeight(ImmutableMap.Entry<Key<?>, Object> entry) {
+    Integer weightObject = CLASS_WEIGHTS.get(entry.getKey().getKind());
+    int weight = weightObject == null ? 0 : weightObject;
+    return entry.getValue().equals(Delete.SENTINEL) ? DELETE_RANGE + -weight : weight;
+  }
+
+  private static int compareByWeight(
+      ImmutableMap.Entry<Key<?>, Object> a, ImmutableMap.Entry<Key<?>, Object> b) {
+    return getWeight(a) - getWeight(b);
+  }
+
+  void saveToJpa() {
+    // Sort the changes into an order that will work for insertion into the database.
+    jpaTm()
+        .transact(
+            () -> {
+              changesBuilder.build().entrySet().stream()
+                  .sorted(TransactionInfo::compareByWeight)
+                  .forEach(
+                      entry -> {
+                        if (entry.getValue().equals(Delete.SENTINEL)) {
+                          jpaTm().delete(VKey.from(entry.getKey()));
+                        } else {
+                          for (SqlEntity entity :
+                              ((DatastoreEntity) entry.getValue()).toSqlEntities()) {
+                            jpaTm().put(entity);
+                          }
+                        }
+                      });
+            });
   }
 }
--- a/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
+++ b/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
@@ -107,7 +107,8 @@ class TransactionInfo {
 
   // The beginning of the range of weights reserved for delete.  This must be greater than any of
   // the values in CLASS_WEIGHTS.
-  static final int DELETE_RANGE = 100;
+  static final int DELETE_RANGE =
+      CLASS_WEIGHTS.values().stream().mapToInt(Integer::intValue).max().getAsInt() + 1;
 
   /** Returns the weight of the entity type in the map entry. */
   private static int getWeight(ImmutableMap.Entry<Key<?>, Object> entry) {
--- a/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
+++ b/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
@@ -111,8 +111,7 @@ class TransactionInfo {
 
   /** Returns the weight of the entity type in the map entry. */
   private static int getWeight(ImmutableMap.Entry<Key<?>, Object> entry) {
-    Integer weightObject = CLASS_WEIGHTS.get(entry.getKey().getKind());
-    int weight = weightObject == null ? 0 : weightObject;
+    int weight = CLASS_WEIGHTS.getOrDefault(entry.getKey().getKind(), 0);
     return entry.getValue().equals(Delete.SENTINEL) ? DELETE_RANGE + -weight : weight;
   }
 
--- a/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
+++ b/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
@@ -23,19 +23,27 @@ import static google.registry.model.ofy.CommitLogBucket.getArbitraryBucketId;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Streams;
 import com.googlecode.objectify.Key;
 import google.registry.persistence.VKey;
 import google.registry.schema.replay.DatastoreEntity;
 import google.registry.schema.replay.SqlEntity;
+import java.util.IntSummaryStatistics;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.joda.time.DateTime;
 
 /** Metadata for an {@link Ofy} transaction that saves commit logs. */
 class TransactionInfo {
 
-  private enum Delete { SENTINEL }
+  @VisibleForTesting
+  enum Delete {
+    SENTINEL
+  }
 
   /** Logical "now" of the transaction. */
   DateTime transactionTime;
@@ -100,20 +108,29 @@ class TransactionInfo {
   // Mapping from class name to "weight" (which in this case is the order in which the class must
   // be "put" in a transaction with respect to instances of other classes).  Lower weight classes
   // are put first, by default all classes have a weight of zero.
+  @VisibleForTesting
   static final ImmutableMap<String, Integer> CLASS_WEIGHTS =
       ImmutableMap.of(
           "HistoryEntry", -1,
           "DomainBase", 1);
 
   // The beginning of the range of weights reserved for delete.  This must be greater than any of
-  // the values in CLASS_WEIGHTS.
-  static final int DELETE_RANGE =
-      CLASS_WEIGHTS.values().stream().mapToInt(Integer::intValue).max().getAsInt() + 1;
+  // the values in CLASS_WEIGHTS by enough overhead to accomodate any negative values in it.
+  static final int DELETE_RANGE = calculateDeleteRangeStart(CLASS_WEIGHTS);
+
+  @VisibleForTesting
+  static int calculateDeleteRangeStart(ImmutableMap<String, Integer> map) {
+    IntSummaryStatistics stats =
+        Streams.concat(map.values().stream(), Stream.of(0))
+            .collect(Collectors.summarizingInt(Integer::intValue));
+    return stats.getMax() - stats.getMin() + 1;
+  }
 
   /** Returns the weight of the entity type in the map entry. */
-  private static int getWeight(ImmutableMap.Entry<Key<?>, Object> entry) {
+  @VisibleForTesting
+  static int getWeight(ImmutableMap.Entry<Key<?>, Object> entry) {
     int weight = CLASS_WEIGHTS.getOrDefault(entry.getKey().getKind(), 0);
-    return entry.getValue().equals(Delete.SENTINEL) ? DELETE_RANGE + -weight : weight;
+    return entry.getValue().equals(Delete.SENTINEL) ? DELETE_RANGE - weight : weight;
   }
 
   private static int compareByWeight(
--- a/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
+++ b/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
@@ -26,15 +26,11 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Streams;
 import com.googlecode.objectify.Key;
 import google.registry.persistence.VKey;
 import google.registry.schema.replay.DatastoreEntity;
 import google.registry.schema.replay.SqlEntity;
-import java.util.IntSummaryStatistics;
 import java.util.Map;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.joda.time.DateTime;
 
 /** Metadata for an {@link Ofy} transaction that saves commit logs. */
@@ -108,7 +104,6 @@ class TransactionInfo {
   // Mapping from class name to "weight" (which in this case is the order in which the class must
   // be "put" in a transaction with respect to instances of other classes).  Lower weight classes
   // are put first, by default all classes have a weight of zero.
-  @VisibleForTesting
   static final ImmutableMap<String, Integer> CLASS_WEIGHTS =
       ImmutableMap.of(
           "HistoryEntry", -1,
@@ -116,15 +111,7 @@ class TransactionInfo {
 
   // The beginning of the range of weights reserved for delete.  This must be greater than any of
   // the values in CLASS_WEIGHTS by enough overhead to accomodate any negative values in it.
-  static final int DELETE_RANGE = calculateDeleteRangeStart(CLASS_WEIGHTS);
-
-  @VisibleForTesting
-  static int calculateDeleteRangeStart(ImmutableMap<String, Integer> map) {
-    IntSummaryStatistics stats =
-        Streams.concat(map.values().stream(), Stream.of(0))
-            .collect(Collectors.summarizingInt(Integer::intValue));
-    return stats.getMax() * 2 + 1;
-  }
+  @VisibleForTesting static final int DELETE_RANGE = Integer.MAX_VALUE / 2;
 
   /** Returns the weight of the entity type in the map entry. */
   @VisibleForTesting
--- a/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
+++ b/core/src/main/java/google/registry/model/ofy/TransactionInfo.java
@@ -123,7 +123,7 @@ class TransactionInfo {
     IntSummaryStatistics stats =
         Streams.concat(map.values().stream(), Stream.of(0))
             .collect(Collectors.summarizingInt(Integer::intValue));
-    return stats.getMax() - stats.getMin() + 1;
+    return stats.getMax() * 2 + 1;
   }
 
   /** Returns the weight of the entity type in the map entry. */
--- a/core/src/main/java/google/registry/model/registry/label/BaseDomainLabelList.java
+++ b/core/src/main/java/google/registry/model/registry/label/BaseDomainLabelList.java
@@ -79,7 +79,7 @@ public abstract class BaseDomainLabelList<T extends Comparable<?>, R extends Dom
   // set to the timestamp when the list is created. In Datastore, we have two fields and the
   // lastUpdateTime is set to the current timestamp when creating and updating a list. So, we use
   // lastUpdateTime as the creation_timestamp column during the dual-write phase for compatibility.
-  @Column(name = "creation_timestamp", nullable = false)
+  @Column(name = "creation_timestamp")
   DateTime lastUpdateTime;
 
   /** Returns the ID of this revision, or throws if null. */
--- a/core/src/test/java/google/registry/flows/FlowTestCase.java
+++ b/core/src/test/java/google/registry/flows/FlowTestCase.java
@@ -101,7 +101,6 @@ public abstract class FlowTestCase<F extends Flow> {
   @RegisterExtension
   final AppEngineExtension appEngine =
       AppEngineExtension.builder()
-          //          .withClock(clock)
           .withDatastoreAndCloudSql()
           .withTaskQueue()
           .build();
--- a/core/src/test/java/google/registry/flows/FlowTestCase.java
+++ b/core/src/test/java/google/registry/flows/FlowTestCase.java
@@ -59,6 +59,7 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import org.joda.time.DateTime;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 /**
@@ -80,12 +81,6 @@ public abstract class FlowTestCase<F extends Flow> {
     SUPERUSER
   }
 
-  @RegisterExtension
-  final AppEngineExtension appEngine =
-      AppEngineExtension.builder().withDatastoreAndCloudSql().withTaskQueue().build();
-
-  @RegisterExtension final InjectExtension inject = new InjectExtension();
-
   protected EppLoader eppLoader;
   protected SessionMetadata sessionMetadata;
   protected FakeClock clock = new FakeClock(DateTime.now(UTC));
@@ -95,14 +90,28 @@ public abstract class FlowTestCase<F extends Flow> {
 
   private EppMetric.Builder eppMetricBuilder;
 
+  // Set the clock for transactional flows.  We have to order this before the AppEngineExtension
+  // which populates data (and may do so with clock-dependent commit logs if mixed with
+  // ReplayExtension).
+  @Order(value = Order.DEFAULT - 1)
+  @RegisterExtension
+  final InjectExtension inject =
+      new InjectExtension().withStaticFieldOverride(Ofy.class, "clock", clock);
+
+  @RegisterExtension
+  final AppEngineExtension appEngine =
+      AppEngineExtension.builder()
+          //          .withClock(clock)
+          .withDatastoreAndCloudSql()
+          .withTaskQueue()
+          .build();
+
   @BeforeEach
   public void beforeEachFlowTestCase() {
     sessionMetadata = new HttpSessionMetadata(new FakeHttpSession());
     sessionMetadata.setClientId("TheRegistrar");
     sessionMetadata.setServiceExtensionUris(ProtocolDefinition.getVisibleServiceExtensionUris());
     ofy().saveWithoutBackup().entity(new ClaimsListSingleton()).now();
-    // For transactional flows
-    inject.setStaticField(Ofy.class, "clock", clock);
  }
 
   protected void removeServiceExtensionUri(String uri) {
--- a/core/src/test/java/google/registry/flows/FlowTestCase.java
+++ b/core/src/test/java/google/registry/flows/FlowTestCase.java
@@ -100,10 +100,7 @@ public abstract class FlowTestCase<F extends Flow> {
 
   @RegisterExtension
   final AppEngineExtension appEngine =
-      AppEngineExtension.builder()
-          .withDatastoreAndCloudSql()
-          .withTaskQueue()
-          .build();
+      AppEngineExtension.builder().withDatastoreAndCloudSql().withTaskQueue().build();
 
   @BeforeEach
   public void beforeEachFlowTestCase() {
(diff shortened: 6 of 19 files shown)


hstonec (changes requested):
> Reviewed 4 of 6 files at r1, 15 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuKSii1Lm7ues0-SZb:-MLuKSii1Lm7ues0-SZc:bhymo3e) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*
> ```
>     Integer weightObject = CLASS_WEIGHTS.get(entry.getKey().getKind());
>     int weight = weightObject == null ? 0 : weightObject;
> ```
Use `Map.getOrDefault()` instead?
___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 134 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuLRyt424pWGtFZOlb:-MLuLRyt424pWGtFZOlc:b-y1nkkh) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L134)):*
> ```
> jpaTm().delete(VKey.from(entry.getKey()));
> ```
There is a `TransactionManager.delete(entity)` method you may want to try here.
___
*[core/src/test/java/google/registry/flows/FlowTestCase.java, line 104 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuM5mJCWGu7i4zThYk:-MLuM5mJCWGu7i4zThYl:b-mkt04b) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/flows/FlowTestCase.java#L104)):*
> ```
> //          .withClock(clock)
> ```
remove this line?
___
*[core/src/test/java/google/registry/testing/AppEngineExtension.java, line 485 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuMLU7BCbgRxsoYP5C:-MLuMLU7BCbgRxsoYP5D:b-mgusrg) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/AppEngineExtension.java#L485)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
>       // If there is a replay extension, we'll want to call its replayToSql() method.
>       ReplayExtension replayer =
>           (ReplayExtension)
>               context.getStore(ExtensionContext.Namespace.GLOBAL).get(ReplayExtension.class);
>       if (replayer != null) {
>         replayer.replayToSql();
>       }
> ```
</details>

As we call `replay` inside the `ReplayExtension` anyway, do we still need this?
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuNbtF2QiFeN-GhtCV:-MLuNbtF2QiFeN-GhtCW:btqzdsv) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/ReplayExtension.java#L40)):*
> ```
>     DatastoreHelper.setClock(clock);
>     DatastoreHelper.setAlwaysSaveWithBackup(true);
> ```
If those fields/states are only associated with `ReplayExtension`, can we put them here instead of having a static fields in `DatastoreHelper`?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman and @hstonec)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuKSii1Lm7ues0-SZb:-MLxi7XoAJPGN0oBf85w:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*

> ```
>     Integer weightObject = CLASS_WEIGHTS.get(entry.getKey().getKind());
>     int weight = weightObject == null ? 0 : weightObject;
> ```
Use `Map.getOrDefault()` instead?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 134 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuLRyt424pWGtFZOlb:-MLxiGjN0tWSdBUpfoHt:bubx7jp) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L134)):*

> ```
> jpaTm().delete(VKey.from(entry.getKey()));
> ```
There is a `TransactionManager.delete(entity)` method you may want to try here.
</blockquote></details>

In this case, we don't have an entity.  Just a key and the Delete.SENTINEL.
___
*[core/src/test/java/google/registry/flows/FlowTestCase.java, line 104 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuM5mJCWGu7i4zThYk:-MLxikgy5PGWB5qnmClS:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/flows/FlowTestCase.java#L104)):*

> ```
> //          .withClock(clock)
> ```
remove this line?
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/AppEngineExtension.java, line 485 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuMLU7BCbgRxsoYP5C:-MLxqZ1X3oxfoptnCR-n:b-aj424m) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/AppEngineExtension.java#L485)):*

> ```
>       // If there is a replay extension, we'll want to call its replayToSql() method.
>       ReplayExtension replayer =
>           (ReplayExtension)
>               context.getStore(ExtensionContext.Namespace.GLOBAL).get(ReplayExtension.class);
>       if (replayer != null) {
>         replayer.replayToSql();
>       }
> ```
As we call `replay` inside the `ReplayExtension` anyway, do we still need this?
</blockquote></details>

Unfortunately, yes.  I've added more comments that hopefully clarify the situation, PTAL?
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuNbtF2QiFeN-GhtCV:-MLxquVN0k9JvXAsCAab:b-b27s1c) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/ReplayExtension.java#L40)):*

> ```
>     DatastoreHelper.setClock(clock);
>     DatastoreHelper.setAlwaysSaveWithBackup(true);
> ```
If those fields/states are only associated with `ReplayExtension`, can we put them here instead of having a static fields in `DatastoreHelper`?
</blockquote></details>

We could, although we would need to access them from DatastoreHelper somehow.

We could:
1) Make them static fields of ReplayExtension with accessors to allow DatastoreHelper to read them.
2) Make them instance variables of ReplayExtension and inject the ReplayExtension instance into DatastoreHelper
3) Leave everything as it is.
4) Something else?

What would you prefer?  (It all seems equally ugly to me, so no preference here :-)

hstonec (changes requested):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 17 of 19 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/test/java/google/registry/testing/AppEngineExtension.java, line 485 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuMLU7BCbgRxsoYP5C:-MM1A0SfEU_65ZU1y951:b-uyasw) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/AppEngineExtension.java#L485)):*

Unfortunately, yes.  I've added more comments that hopefully clarify the situation, PTAL?
</blockquote></details>

makes sense, thanks for the explanation.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuNbtF2QiFeN-GhtCV:-MM1BChz6P6jc_TmGvxI:b-od0496) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/ReplayExtension.java#L40)):*

We could, although we would need to access them from DatastoreHelper somehow.

We could:
1) Make them static fields of ReplayExtension with accessors to allow DatastoreHelper to read them.
2) Make them instance variables of ReplayExtension and inject the ReplayExtension instance into DatastoreHelper
3) Leave everything as it is.
4) Something else?

What would you prefer?  (It all seems equally ugly to me, so no preference here :-)
</blockquote></details>

hmm, that is unfortunate, I think we can leave them in DatastoreHelper and add some comments to remind us that they are actually only used by this extension.

hstonec (commented):
> Reviewed 2 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 11 of 19 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuNbtF2QiFeN-GhtCV:-MM1SQvt18adnmgT1BXB:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/ReplayExtension.java#L40)):*

hmm, that is unfortunate, I think we can leave them in DatastoreHelper and add some comments to remind us that they are actually only used by this extension.
</blockquote></details>

Done.

hstonec (approved):
:lgtm:

> Reviewed 8 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857#-:-MM1Sy4SDd6fchVNfoB3:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 8 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/ReplayQueue.java, line 23 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2V77d0W3LvqrFQsQ9:-MM2V77e3biAY-1Vfjda:bbrkabd) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/ofy/ReplayQueue.java#L23)):*
> ```Java
>  * Implements simplified datastore to SQL transaction replay.
>  *
>  * <p>This code is to be removed when the actual replay cron job is implemented.
> ```

:) we shall see
___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Y3TT6GZ5MTudHzj5:-MM2Y3TT6GZ5MTudHzj6:bj5z8pq) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L105)):*
> ```Java
>   static final ImmutableMap<String, Integer> CLASS_WEIGHTS =
>       ImmutableMap.of(
>           "HistoryEntry", -1,
> ```

Why not use Class objects here? Does the Key object not allow that? 
___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 110 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Y5qS37hufzWFI_66:-MM2Y5qS37hufzWFI_67:b-o6misg) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L110)):*
> ```Java
>   // The beginning of the range of weights reserved for delete.  This must be greater than any of
>   // the values in CLASS_WEIGHTS.
>   static final int DELETE_RANGE = 100;
> ```

it's unlikely but maybe it might be good to have this be significantly higher? No risk of conflict, however unlikely 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 186 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Yow64pjBPPeaD5am:-MM2Yow64pjBPPeaD5an:boxtrq6) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L186)):*
> ```Java
>   /** Whether this change was requested by a registrar. */
>   @Column(nullable = false, name = "historyRequestedByRegistrar")
>   Boolean requestedByRegistrar = false;
> ```

I think we wanted this to be null by default? At least I thought that's what Ben said, but he might have been referring to xmlBytes 
___
*[core/src/test/java/google/registry/testing/DatastoreHelper.java, line 124 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2ditPDZNK1MBYnb0n:-MM2ditPDZNK1MBYnb0o:b-cw9uu1) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/DatastoreHelper.java#L124)):*
> ```Java
> /** Static utils for setting up test resources. */
> public class DatastoreHelper {
> ```

Does this PR cover all the changes that would be necessary to make to DatastoreHelper?

(also we shouldn't do it here ,but we should rename the class at some point :) )
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 25 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2dDrQ-mupiyUjhdc3:-MM2dDrRAQI_hd_mj2rj:b8moxf0) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/ReplayExtension.java#L25)):*
> ```Java
>  * A JUnit extension that replays datastore transactions against postgresql.
>  *
>  * <p>This extension must be ordered before AppEngineExtension. If AppEngineExtension is not used,
> ```

To be clear, this is because we want objects set up in the AppEngineExtension to be replayed as well? Assuming so, we may wish to explain that here. 

Can we also explain why it should be after the JPATME as well? To someone reading the code in a month or two it might not be clear, might be helpful to avoid having them dig 
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 26 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2dKpkDNf4ELbN1vTr:-MM2dKpkDNf4ELbN1vTs:bgwcmn5) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/ReplayExtension.java#L26)):*
> ```Java
>  *
>  * <p>This extension must be ordered before AppEngineExtension. If AppEngineExtension is not used,
>  * JpaTransactionManagerException must be, and this extension should be ordered _after_
> ```

nit: s/Exception/Extension/g
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 54 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2daPN5Rt4B-ZINXEx:-MM2daPN5Rt4B-ZINXEy:b-uvj7w1) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/ReplayExtension.java#L54)):*
> ```Java
> 
>   public void replayToSql() {
>     DatastoreHelper.setAlwaysSaveWithBackup(false);
> ```

Out of curiosity, what is the purpose of setting this back to false after the tests?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 7 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Y3TT6GZ5MTudHzj5:-MMG8gfs9bKm4ZJYZXF2:bx85f6s) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L105)):*

Why not use Class objects here? Does the Key object not allow that? 
</blockquote></details>

We don't get the class object in the objectify key, just the unqualified class name.
___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 110 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Y5qS37hufzWFI_66:-MMG9ziM7AshUgvSZQB5:bome0up) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L110)):*

it's unlikely but maybe it might be good to have this be significantly higher? No risk of conflict, however unlikely 
</blockquote></details>

Actually, it occurs to me that the cleanest way to do this is to compute it from the values, so done.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 186 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Yow64pjBPPeaD5am:-MMGA9Op2iG5alAug_8K:b-80spgx) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L186)):*

I think we wanted this to be null by default? At least I thought that's what Ben said, but he might have been referring to xmlBytes 
</blockquote></details>

We were last discussing it in the context of xmlBytes.  The case for this field being null is AFAIK just datastore, it's cheaper to store a null field than a false value.  This isn't likely to be a big deal for new history entries, though. 
___
*[core/src/test/java/google/registry/testing/DatastoreHelper.java, line 124 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2ditPDZNK1MBYnb0n:-MMGAlcO7-ZvEueFIQM7:b-xkxqwx) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/DatastoreHelper.java#L124)):*

Does this PR cover all the changes that would be necessary to make to DatastoreHelper?

(also we shouldn't do it here ,but we should rename the class at some point :) )
</blockquote></details>

I believe so, yes.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 25 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2dDrQ-mupiyUjhdc3:-MMGDjL01R6RuzGWoohV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/ReplayExtension.java#L25)):*

To be clear, this is because we want objects set up in the AppEngineExtension to be replayed as well? Assuming so, we may wish to explain that here. 

Can we also explain why it should be after the JPATME as well? To someone reading the code in a month or two it might not be clear, might be helpful to avoid having them dig 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 26 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2dKpkDNf4ELbN1vTr:-MMGDqJy5EQhJVTx07HT:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/ReplayExtension.java#L26)):*

nit: s/Exception/Extension/g
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 54 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2daPN5Rt4B-ZINXEx:-MMGDxyv0eu_c82Wc6ZY:b-ym1nvo) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/test/java/google/registry/testing/ReplayExtension.java#L54)):*

Out of curiosity, what is the purpose of setting this back to false after the tests? 
</blockquote></details>

IIUIC, all of the database tests are run in the same process and most don't use the replay extension at this time.  Therefore, we want to turn off the functionality when the test is complete, otherwise it would break for tests that didn't have the other necessary extensions set up correctly.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/857#-MM2Y3TT6GZ5MTudHzj5:-MMGr-Sg03kL5wk3ipi7:b7afptu) ([raw file](https://github.com/google/nomulus/blob/7d8c51fcf8d40545eb6df362e0e5fd7a3a571f5d/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L105)):*

We don't get the class object in the objectify key, just the unqualified class name.
</blockquote></details>

annoying, but all right. We shouldn't have conflicts anyway.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 11 of 18 files reviewed, 2 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 116 at r6](https://reviewable.io/reviews/google/nomulus/857#-MMHqStX4pyNLfaDyK1R:-MMHqStX4pyNLfaDyK1S:be746nm) ([raw file](https://github.com/google/nomulus/blob/87e2c7fbc46003a025df4cee445e070ea915e26e/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L116)):*
> ```Java
>   private static int getWeight(ImmutableMap.Entry<Key<?>, Object> entry) {
>     int weight = CLASS_WEIGHTS.getOrDefault(entry.getKey().getKind(), 0);
>     return entry.getValue().equals(Delete.SENTINEL) ? DELETE_RANGE + -weight : weight;
> ```

Actually, two questions:
1. i think we can remove the + in  DELETE_RANGE + -weight 
2. shouldn't we make the delete range higher? If we're subtracting weights from the DELETE_RANGE for deletes, we'd want to make DELETE_RANGE start at the (max weight * 2) + 1, right?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 11 of 18 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 116 at r6](https://reviewable.io/reviews/google/nomulus/857#-MMHqStX4pyNLfaDyK1R:-MMIJ4rw-9o1GDqbZ2qX:brqrvcv) ([raw file](https://github.com/google/nomulus/blob/87e2c7fbc46003a025df4cee445e070ea915e26e/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L116)):*

Actually, two questions:
1. i think we can remove the + in  DELETE_RANGE + -weight 
2. shouldn't we make the delete range higher? If we're subtracting weights from the DELETE_RANGE for deletes, we'd want to make DELETE_RANGE start at the (max weight * 2) + 1, right? 
</blockquote></details>

1) done
2) Good catch!  Fixed the calculation and added a couple of tests (since the code clearly merits it).

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 116 at r6](https://reviewable.io/reviews/google/nomulus/857#-MMHqStX4pyNLfaDyK1R:-MMKxRPV9_7IQpcnl-RM:b-896fix) ([raw file](https://github.com/google/nomulus/blob/87e2c7fbc46003a025df4cee445e070ea915e26e/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L116)):*

1) done
2) Good catch!  Fixed the calculation and added a couple of tests (since the code clearly merits it).
</blockquote></details>

Done.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 3 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMLsdxb1qjp7G8IyyNP:-MMLsdxc-lPOYGWlb_5H:b8kh025) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*

> ```
>   static final ImmutableMap<String, Integer> CLASS_WEIGHTS =
>       ImmutableMap.of(
>           "HistoryEntry", -1,
>           "DomainBase", 1);
> ```
</details>

We should probably create a schema inspector that builds an ordered list from the declared foreign key constraints.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 3 unresolved discussions (waiting on @gbrodman, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMLsdxb1qjp7G8IyyNP:-MMLvt5z5GrRb4TvU9HS:b-xae38d) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*

> ```
>   static final ImmutableMap<String, Integer> CLASS_WEIGHTS =
>       ImmutableMap.of(
>           "HistoryEntry", -1,
>           "DomainBase", 1);
> ```
We should probably create a schema inspector that builds an ordered list from the declared foreign key constraints.
</blockquote></details>

That would be awesome!  Can we get programmatic access to the schema or would we have to parse it?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 3 unresolved discussions (waiting on @gbrodman, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMLsdxb1qjp7G8IyyNP:-MMLw_GU3Vk5SO4F8IcO:bfid9om) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*

That would be awesome!  Can we get programmatic access to the schema or would we have to parse it?
</blockquote></details>

I think SchemaCrawler should be able to do that, see [here](https://www.schemacrawler.com/how-to.html#api). FYI, we use SchemaCrawler to genereate the ER diagram so it is already added.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 3 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMLsdxb1qjp7G8IyyNP:-MMLxr0d5YXTfUT3hfRA:biie3ur) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*

I think SchemaCrawler should be able to do that, see [here](https://www.schemacrawler.com/how-to.html#api). FYI, we use SchemaCrawler to genereate the ER diagram so it is already added.
</blockquote></details>

JDBC's Connection.getMetaData() returns a java.sql.DatabaseMetadata instance. We can start from there.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 3 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 116 at r6](https://reviewable.io/reviews/google/nomulus/857#-MMHqStX4pyNLfaDyK1R:-MMM-6JA31C9H-EQ9g9w:b8z5m1p) ([raw file](https://github.com/google/nomulus/blob/87e2c7fbc46003a025df4cee445e070ea915e26e/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L116)):*

Done.
</blockquote></details>

I'm not sure this fixes the issue -- in the test, for example, we have classes with weights 10 and 20 and a delete range start of 21. That means that a delete of object 10 would happen before the addition of object 20

weiminyu (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 115 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMLsdxb1qjp7G8IyyNP:-MMM0iFEB48u5xv1LYF9:b-ubeiw4) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L115)):*

JDBC's Connection.getMetaData() returns a java.sql.DatabaseMetadata instance. We can start from there.
</blockquote></details>

Also, if we build an list we won't need (negative) weights.
___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 120 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMM-O8o2__O-W6aYbeA:-MMM-O8o2__O-W6aYbeB:bqd18i0) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L120)):*
> ```
>   static final int DELETE_RANGE = calculateDeleteRangeStart(CLASS_WEIGHTS);
> ```
How about (Integer.MAX / 2)?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @hstonec, and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 116 at r6](https://reviewable.io/reviews/google/nomulus/857#-MMHqStX4pyNLfaDyK1R:-MMM-YPsApXHl5YBZH41:b-lh8f46) ([raw file](https://github.com/google/nomulus/blob/87e2c7fbc46003a025df4cee445e070ea915e26e/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L116)):*

I'm not sure this fixes the issue -- in the test, for example, we have classes with weights 10 and 20 and a delete range start of 21. That means that a delete of object 10 would happen before the addition of object 20 
</blockquote></details>

You're right.  I keep forgetting to flip the delete range.  PTAL?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 120 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMM-O8o2__O-W6aYbeA:-MMM5HaZ8T3V1xDTkXCG:b-boshqn) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L120)):*

> ```
>   static final int DELETE_RANGE = calculateDeleteRangeStart(CLASS_WEIGHTS);
> ```
How about (Integer.MAX / 2)?
</blockquote></details>

Now what fun would that be?  ;-)  Done.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 120 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMM-O8o2__O-W6aYbeA:-MMMAS4R5yJEOUtosN1Z:b-1mxygf) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L120)):*

Now what fun would that be?  ;-)  Done.
</blockquote></details>

haha wasn't this the original code?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 120 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMM-O8o2__O-W6aYbeA:-MMMAYG4Di01QvIWj55f:b7w54bt) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L120)):*

haha wasn't this the original code? 
</blockquote></details>

wait, never mind it wasn't, I must be thinking of some other thing

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: 10 of 19 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 120 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMM-O8o2__O-W6aYbeA:-MMMJWwE2-wPedjA9NhM:bv6dpn7) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L120)):*

wait, never mind it wasn't, I must be thinking of some other thing 
</blockquote></details>

Gus, are you good?  (I think your comments retracted your LGTM)

gbrodman (approved):
> Reviewed 7 of 8 files at r6, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec, @mindhog, and @weiminyu)

___
*[core/src/main/java/google/registry/model/ofy/TransactionInfo.java, line 120 at r7](https://reviewable.io/reviews/google/nomulus/857#-MMM-O8o2__O-W6aYbeA:-MMMPID_DtYs4tRIM7qL:b-z9gy74) ([raw file](https://github.com/google/nomulus/blob/8103092ede619c9e58b63ccb68adad0cb207a1f0/core/src/main/java/google/registry/model/ofy/TransactionInfo.java#L120)):*

Gus, are you good?  (I think your comments retracted your LGTM)
</blockquote></details>

yeah, we should be good to go

mindhog (commented):
> Dismissed @weiminyu from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: all files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/test/java/google/registry/testing/AppEngineExtension.java, line 485 at r2](https://reviewable.io/reviews/google/nomulus/857#-MLuMLU7BCbgRxsoYP5C:-MMMRql50suxxQHLL-bK:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/d67bcc7d7221a8d64469d999e588d995d3b735fe/core/src/test/java/google/registry/testing/AppEngineExtension.java#L485)):*

makes sense, thanks for the explanation.
</blockquote></details>

Ack.

mindhog (commented):
> Dismissed @hstonec from a discussion.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/857)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #860, opened 2020-11-03
Persist two singleton entities in SQL tables

gbrodman:
A table might not be the cleanest place to store singleton entities, but by
doing this we ensure we can easily inspect them later and use the same
sort of persistence logic for these that we do elsewhere.

ServerSecret is stored upon retrieval so that we make sure that the same
secret is used in both Datastore and SQL (we wouldn't want to change
it).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/860)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860)* status: 0 of 15 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[db/src/main/resources/sql/flyway/V72__singleton_entities.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJJlQN6DiWR-vk96VG:-MLJJlQN6DiWR-vk96VH:b-5i6quv) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/flyway/V72__singleton_entities.sql#L15)):*
> ```TSQL
> 
> CREATE TABLE "ServerSecret" (
> ```

This feels so dumb, having two tables that exist solely to contain one row each, but I honestly can't think of a better alternative in a DB that uses schema like SQL ones do.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 630 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJMls23OzXTB9H3__f:-MLJMls23OzXTB9H3__g:bd39jtc) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L630)):*
> ```
> 
>     create table "ServerSecret" (
>        id int8 not null,
> ```

Is there any point of having the ID at all?  Why not just have the UUID field (see next comment) be the primary key?  The reason Datastore had a fixed id value was because Datastore works on a load-by-key basis, whereas with SQL it's trivial to say "get me the only row in this table"  (`SELECT * FROM tablename LIMIT 1`)
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 631 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJLzoYDVM8qZbJ-bMV:-MLJLzoYDVM8qZbJ-bMW:bgzmm78) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L631)):*
> ```
>     create table "ServerSecret" (
>        id int8 not null,
>         least_significant int8 not null,
> ```

This is a hack left over from Datastore not having suitable types for us to use.  We don't have that problem now, as PostgreSQL has the perfect-for-this-use-case `UUID` type (it's 16 bytes big).  Let's use that.  https://www.postgresql.org/docs/current/datatype-uuid.html
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 703 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNB6B7KmNNG676RdA:-MLJNB6B7KmNNG676RdB:by0kcgx) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L703)):*
> ```
> 
>     create table "TmchCrl" (
>        id int8 not null,
> ```

Same comment regarding `id` likely not being necessary.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 704 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNIdh2x8R5OvdjNV2:-MLJNIdh2x8R5OvdjNV3:b-2618i8) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L704)):*
> ```
>     create table "TmchCrl" (
>        id int8 not null,
>         crl text not null,
> ```

Can this be given a better name that's more indicative of the kind of data it actually contains?
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 705 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNERPBG7yFn0zQ_4E:-MLJNERPBG7yFn0zQ_4F:b-a785um) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L705)):*
> ```
>        id int8 not null,
>         crl text not null,
>         updated timestamptz not null,
> ```

This is inconsistent with how the last updated timestamp fields are named on other tables.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860)* status: 0 of 15 files reviewed, 6 unresolved discussions (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/flyway/V72__singleton_entities.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJJlQN6DiWR-vk96VG:-MLJd6bm1GRVboTjzkyC:b-dfpfxl) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/flyway/V72__singleton_entities.sql#L15)):*

This feels so dumb, having two tables that exist solely to contain one row each, but I honestly can't think of a better alternative in a DB that uses schema like SQL ones do.
</blockquote></details>

I know :( 

Shicong pointed out that it's also probably beneficial to have as much business-data-storage in the same place as well. 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 630 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJMls23OzXTB9H3__f:-MLJeeduDNbU3BKiP8uR:b-9qxm06) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L630)):*

Is there any point of having the ID at all?  Why not just have the UUID field (see next comment) be the primary key?  The reason Datastore had a fixed id value was because Datastore works on a load-by-key basis, whereas with SQL it's trivial to say "get me the only row in this table"  (`SELECT * FROM tablename LIMIT 1`)
</blockquote></details>

The primary benefit to having a single ID as the key is that we can use the same TransactionManager call to retrieve the row regardless of which database is currently the primary -- we can just call `tm().maybeLoad(key)` and it'll just work regardless of whether we're on DS or SQL 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 631 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJLzoYDVM8qZbJ-bMV:-MLJf3nh9B2Lq9Uq0Zhb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L631)):*

This is a hack left over from Datastore not having suitable types for us to use.  We don't have that problem now, as PostgreSQL has the perfect-for-this-use-case `UUID` type (it's 16 bytes big).  Let's use that.  https://www.postgresql.org/docs/current/datatype-uuid.html
</blockquote></details>

Done.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 703 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNB6B7KmNNG676RdA:-MLJisQaCcXnFpRMik7P:b-s95if0) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L703)):*

Same comment regarding `id` likely not being necessary.
</blockquote></details>

See the other comment -- I'm not sure if it's worth having to create a separate DAO object for this when we can just use a single vkey regardless of which DB we're in?
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 704 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNIdh2x8R5OvdjNV2:-MLJjbU0CNfJV6OLOIAO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L704)):*

Can this be given a better name that's more indicative of the kind of data it actually contains?
</blockquote></details>

Done.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 705 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNERPBG7yFn0zQ_4E:-MLJj3qI4ZCp7tblnVr5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L705)):*

This is inconsistent with how the last updated timestamp fields are named on other tables.
</blockquote></details>

Done.

gbrodman (commented):
Removed the IDs, PTAL thanks 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860#-:-MLOvmFG7JV6a1le847L:bdlay2m)* status: 0 of 15 files reviewed, 6 unresolved discussions (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 630 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJMls23OzXTB9H3__f:-MLOvguu-cb11L8O8k5H:b-bvlta) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L630)):*

The primary benefit to having a single ID as the key is that we can use the same TransactionManager call to retrieve the row regardless of which database is currently the primary -- we can just call `tm().maybeLoad(key)` and it'll just work regardless of whether we're on DS or SQL 
</blockquote></details>

OK, I tried it without the ID. I think this may end up being cleaner in the end once the migration is done, even if it requires a bit of cruft now

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 703 at r1](https://reviewable.io/reviews/google/nomulus/860#-MLJNB6B7KmNNG676RdA:-MLU6Nh0E4lzjU8wpSYo:br8rjcf) ([raw file](https://github.com/google/nomulus/blob/3ad3093ba4f91ae6c5f1eeeb45aa053999f8ab52/db/src/main/resources/sql/schema/db-schema.sql.generated#L703)):*

See the other comment -- I'm not sure if it's worth having to create a separate DAO object for this when we can just use a single vkey regardless of which DB we're in?
</blockquote></details>

Looks like it's gone now, so resolving.
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 907 at r2](https://reviewable.io/reviews/google/nomulus/860#-MLU6TzmBuYCkAWaDxAf:-MLU6TzmBuYCkAWaDxAg:b-o7k29c) ([raw file](https://github.com/google/nomulus/blob/6f7034dc3085dbf0558f883650cc1580b2745c76/db/src/main/resources/sql/schema/nomulus.golden.sql#L907)):*
> ```TSQL
> 
> CREATE TABLE public."ServerSecret" (
>     uuid uuid NOT NULL
> ```

Wait what, the column is named `uuid` which is also the same as its data type `uuid`?  And that's even allowed by SQL???  You can't name an int `int` in Java!  This seems unnecessarily confusing.  Why not just name the column `secret` or something similar?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 907 at r2](https://reviewable.io/reviews/google/nomulus/860#-MLU6TzmBuYCkAWaDxAf:-MLU8CGHEgr1eYjX8spo:b-sg5ksg) ([raw file](https://github.com/google/nomulus/blob/6f7034dc3085dbf0558f883650cc1580b2745c76/db/src/main/resources/sql/schema/nomulus.golden.sql#L907)):*

Wait what, the column is named `uuid` which is also the same as its data type `uuid`?  And that's even allowed by SQL???  You can't name an int `int` in Java!  This seems unnecessarily confusing.  Why not just name the column `secret` or something similar?
</blockquote></details>

oh yes, it is most definitely allowed and somewhat frequently done. We don't have to do it though.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860)* status: 0 of 15 files reviewed, all discussions resolved

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 907 at r2](https://reviewable.io/reviews/google/nomulus/860#-MLU6TzmBuYCkAWaDxAf:-MLhwioz7STGrmebuqvR:b-bdxip0) ([raw file](https://github.com/google/nomulus/blob/6f7034dc3085dbf0558f883650cc1580b2745c76/db/src/main/resources/sql/schema/nomulus.golden.sql#L907)):*

oh yes, it is most definitely allowed and somewhat frequently done. We don't have to do it though. 
</blockquote></details>

Yeah, we should make a habit of not doing it, because while it would work in this particular case in Java, it won't work generally for columns that map onto Java primitive types.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/860#-:-MLhx0087ZaaPASfcZdY:bnfp4nl)* status: 0 of 15 files reviewed, all discussions resolved

PR #863, opened 2020-11-06
Small SQL persistence fixes to model classes

mindhog:
- Add a createVKey() method to Registry (Registry vkeys are composite)
- Add/fix toSqlEntities() methods in premium and reserved list classes.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/863)

### DIFF
--- a/core/src/main/java/google/registry/model/registry/Registry.java
+++ b/core/src/main/java/google/registry/model/registry/Registry.java
@@ -285,6 +285,10 @@ public class Registry extends ImmutableObject implements Buildable, DatastoreAnd
     return VKey.create(Registry.class, tld, Key.create(getCrossTldKey(), Registry.class, tld));
   }
 
+  public static VKey<Registry> createVKey(Key<Registry> key) {
+    return createVKey(key.getName());
+  }
+
   /**
    * The name of the pricing engine that this TLD uses.
    *
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -444,11 +444,6 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
     }
   }
 
-  @Override
-  public ImmutableList<SqlEntity> toSqlEntities() {
-    return ImmutableList.of();
-  }
-
   @PrePersist
   void prePersist() {
     lastUpdateTime = creationTime;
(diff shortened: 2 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/863)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 320 at r1](https://reviewable.io/reviews/google/nomulus/863#-MLTwK0s9pZxV1gnelRO:-MLTwK0s9pZxV1gnelRP:b1ecjjv) ([raw file](https://github.com/google/nomulus/blob/1aa54ebd1c6ca4bb8859df8c19db2ada06415b96/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L320)):*
> ```Java
>   }
> 
>   @Override
> ```

nit: this should be above the Builder 

non-nit: We should have this implement DatastoreEntity and SqlEntity separately, then have both methods return `ImmutableList.of()`. Currently, it implements `DatastoreAndSqlEntity` which is meant to signify that replay should save the same object, which shouldn't be the case for dually-written objects. 

We could also create a `DuallyWrittenEntity` interface which implements `DatastoreEntity` and `SqlEntity` and returns empty lists for both. Then all dually-written entities could just extend that.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/863)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/863)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 320 at r1](https://reviewable.io/reviews/google/nomulus/863#-MLTwK0s9pZxV1gnelRO:-MLioas0Bze1hC8dhk7L:bcpzgf6) ([raw file](https://github.com/google/nomulus/blob/1aa54ebd1c6ca4bb8859df8c19db2ada06415b96/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L320)):*

nit: this should be above the Builder 

non-nit: We should have this implement DatastoreEntity and SqlEntity separately, then have both methods return `ImmutableList.of()`. Currently, it implements `DatastoreAndSqlEntity` which is meant to signify that replay should save the same object, which shouldn't be the case for dually-written objects. 

We could also create a `DuallyWrittenEntity` interface which implements `DatastoreEntity` and `SqlEntity` and returns empty lists for both. Then all dually-written entities could just extend that. 


</blockquote></details>

Dealt with in #866 (thanks for that!)

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/863)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #864, opened 2020-11-06
Allow InjectExtension to apply in "before each"

mindhog:
InjectExtension is non-standard in that, while it cleans up the changes it has
applied in its "after each" method, those changes must be applied by the test
driver or by another extension.  This breaks extension ordering, which is
something we're going to need to make heavy use of for the ReplayExtension.

Allow changes to be specified (but not activated) after construction and apply
all unactivated changes during "before each."

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/864)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/864)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #868, opened 2020-11-10
Add SQL replay checkpoint object to SQL

gbrodman:
This will be part of the asynchronous commit-log replay to SQL. Whenever
we successfully export commits up to a particular time, we should
persist that time so we don't replay the same commits again (it is not
idempotent)

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/868)


mindhog (commented):
Is there a reason not to persist the timestamp in SQL rather than datastore?  I think that would simplify things considerably because then you could do it as part of the SQL transaction and not have to deal with the failure mode of "replay transaction succeeds, datastore write fails"

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868#-:-MLnLBGF8mjBj8wDaM8s:b-b0tqcg)* status: :shipit: complete! all files reviewed, all discussions resolved

gbrodman (commented):
My doubts were because this will only be "useful" during the stage of the migration where Datastore is still the primary DB. As a result, I figured we'd want to store everything important there (rather than a potentially-unreliable SQL instance). Thoughts? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868#-:-MLnN5yWEgu33hPrT1vv:b-x45wd8)* status: :shipit: complete! all files reviewed, all discussions resolved

mindhog (commented):
Given that this is a essentially pointer to how far we've gotten in replaying to SQL, I don't see the case for keeping it in datastore.  If we discovered that our replay code was writing an incorrect representation to SQL, I expect that after fixing the bugs we'd want to just delete the database and start with a fresh migration.  In that scenario, having to clear the datastore record would just be an extra step.
I can't think of a scenario where we'd want to update this pointer in the absence of a successful transaction replay, can you? [we can chat about this out of band, if you like] 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868#-:-MLnO1vD3llYpGQRjl8-:b-bxit0i)* status: :shipit: complete! all files reviewed, all discussions resolved

gbrodman (commented):
Discussed out of band, moving this to SQL because the thinking is that if SQL gets messed up, this datetime becomes pointless anyway 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868#-:-MLngDkN-iZnxJRd0Ujl:b4rdcua)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @mindhog)

mindhog (changes requested):
> Reviewed 16 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java, line 41 at r2](https://reviewable.io/reviews/google/nomulus/868#-MLnmdAP6LrtcF64meUe:-MLnmdAP6LrtcF64meUf:b-duhtsc) ([raw file](https://github.com/google/nomulus/blob/d7af79260e488b7043e73c5056dcc25cff8f4b2a/core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java#L41)):*
> ```Java
>   public static DateTime get() {
>     return jpaTm()
>         .transact(
> ```

These should always be called from the replay transaction (so it succeeds or fails atomically with that transaction).  While this code should still work in that case, I think it would be better to not wrap it in an explicit transaction here so that if it does get called outside of a transaction we'll get a runtime error.
___
*[core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java, line 63 at r2](https://reviewable.io/reviews/google/nomulus/868#-MLnpJPOAk-ognKztU4y:-MLnpJPOAk-ognKztU4z:b-as27hs) ([raw file](https://github.com/google/nomulus/blob/d7af79260e488b7043e73c5056dcc25cff8f4b2a/core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java#L63)):*
> ```Java
>               SqlReplayCheckpoint checkpoint = new SqlReplayCheckpoint();
>               checkpoint.lastReplayTime = lastReplayTime;
>               jpaTm().insert(checkpoint);
> ```

I think that you could just do a jpaTm().put(...) here and omit the delete, would that work?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java, line 41 at r2](https://reviewable.io/reviews/google/nomulus/868#-MLnmdAP6LrtcF64meUe:-MLntuGL9VE7bUSw7-H-:b-48vilu) ([raw file](https://github.com/google/nomulus/blob/d7af79260e488b7043e73c5056dcc25cff8f4b2a/core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java#L41)):*

These should always be called from the replay transaction (so it succeeds or fails atomically with that transaction).  While this code should still work in that case, I think it would be better to not wrap it in an explicit transaction here so that if it does get called outside of a transaction we'll get a runtime error.
</blockquote></details>

Good call, done 
___
*[core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java, line 63 at r2](https://reviewable.io/reviews/google/nomulus/868#-MLnpJPOAk-ognKztU4y:-MLnsv3WEx3htw7ZOuwM:b-8y2ebk) ([raw file](https://github.com/google/nomulus/blob/d7af79260e488b7043e73c5056dcc25cff8f4b2a/core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java#L63)):*

I think that you could just do a jpaTm().put(...) here and omit the delete, would that work?
</blockquote></details>

yeah you're right, i wrote this before creating the explicit long id (while trying to use the DateTime as the ID, in which case it would be necessary)

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/868)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #870, opened 2020-11-10
Replace ofyOrJpaTm with tm().isOfy() conditionals

mindhog:
Replace existing ofyOrJpaTm() calls with conditionals (either "if" statements
or ternary expressions) gated on tm().isOfy().

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/870)

### DIFF
--- a/core/src/main/java/google/registry/model/registry/Registries.java
+++ b/core/src/main/java/google/registry/model/registry/Registries.java
@@ -25,7 +25,6 @@ import static google.registry.model.CacheUtils.memoizeWithShortExpiration;
 import static google.registry.model.common.EntityGroupRoot.getCrossTldKey;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
-import static google.registry.persistence.transaction.TransactionManagerUtil.ofyOrJpaTm;
 import static google.registry.util.CollectionUtils.entriesToImmutableMap;
 import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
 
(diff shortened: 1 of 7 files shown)


hstonec (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/870)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 252 at r1](https://reviewable.io/reviews/google/nomulus/870#-MLo8diB-613uCCSxQTd:-MLo8diB-613uCCSxQTe:b-63ka2t) ([raw file](https://github.com/google/nomulus/blob/befee975838e68cc1d3f44f2bbeeef39f8446c72/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L252)):*
> ```
>                   return tm().isOfy()
>                       ? ReservedListDualWriteDao.getLatestRevision(listName).orElse(null)
>                       : ReservedListSqlDao.getLatestRevision(listName).orElse(null);
> ```
Not sure if this is overthinking but if we make this change, maybe we should add a test to assert that we only have two implementations of `TransactionManager`, i.e. `DatastoreTransactionManager` and `JpaTransactionManager`, the original `ofyOrJpaTm` does throw exception if the TM is neither `Jpa` nor `Ofy`. Though it would be very likely we won't add another implementation so I am not too worried about it.

hstonec (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/870#-:-MLoAniX7QvwBw54b4Ae:bnfp4nl)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @hstonec and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/870)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 252 at r1](https://reviewable.io/reviews/google/nomulus/870#-MLo8diB-613uCCSxQTd:-MLrHB8m2y2TjHzIRJaa:b-55ep5l) ([raw file](https://github.com/google/nomulus/blob/befee975838e68cc1d3f44f2bbeeef39f8446c72/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L252)):*

> ```
>                   return tm().isOfy()
>                       ? ReservedListDualWriteDao.getLatestRevision(listName).orElse(null)
>                       : ReservedListSqlDao.getLatestRevision(listName).orElse(null);
> ```
Not sure if this is overthinking but if we make this change, maybe we should add a test to assert that we only have two implementations of `TransactionManager`, i.e. `DatastoreTransactionManager` and `JpaTransactionManager`, the original `ofyOrJpaTm` does throw exception if the TM is neither `Jpa` nor `Ofy`. Though it would be very likely we won't add another implementation so I am not too worried about it.
</blockquote></details>

Yeah, it seemed unlikely that we would add another implementation and I figured that if we did, we'd want to replace isOfy() with something that returned an enum (or maybe do something else entirely), requiring us to address the callsites.

PR #871, opened 2020-11-11
Add dual write for RegistrarPoc

hstonec:
This PR added the dual-write for RegistrarPoc and also converted `RegistrarContactTest` to working with Cloud SQL.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/871)

### DIFF
--- a/core/src/main/java/google/registry/model/registrar/RegistrarContact.java
+++ b/core/src/main/java/google/registry/model/registrar/RegistrarContact.java
@@ -363,9 +363,20 @@ public class RegistrarContact extends ImmutableObject
     parent = Key.create(getCrossTldKey(), Registrar.class, registrarId);
   }
 
+  public String getRegistrarId() {
+    return registrarId;
+  }
+
   public VKey<RegistrarContact> createVKey() {
     return VKey.create(
         RegistrarContact.class, new RegistrarPocId(emailAddress, registrarId), Key.create(this));
+  }
+
+  public static VKey<RegistrarContact> createVKey(Registrar registrar, String emailAddress) {
+    return VKey.create(
+        RegistrarContact.class,
+        new RegistrarPocId(emailAddress, registrar.getClientId()),
+        Key.create(Key.create(registrar), RegistrarContact.class, emailAddress));
   }
 
   /** Class to represent the composite primary key for {@link RegistrarContact} entity. */
--- a/core/src/main/java/google/registry/tools/RegistrarContactCommand.java
+++ b/core/src/main/java/google/registry/tools/RegistrarContactCommand.java
@@ -18,6 +18,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.util.CollectionUtils.nullToEmpty;
 import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
 import static google.registry.util.PreconditionsUtils.checkArgumentPresent;
@@ -29,6 +30,7 @@ import com.google.common.base.Enums;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.flogger.FluentLogger;
 import google.registry.model.common.GaeUserIdConverter;
 import google.registry.model.registrar.Registrar;
 import google.registry.model.registrar.RegistrarContact;
@@ -52,6 +54,8 @@ import javax.annotation.Nullable;
     separators = " =",
     commandDescription = "Create/read/update/delete the various contact lists for a Registrar.")
 final class RegistrarContactCommand extends MutatingCommand {
+
+  static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
   @Parameter(
       description = "Client identifier of the registrar account.",
@@ -158,6 +162,9 @@ final class RegistrarContactCommand extends MutatingCommand {
   @Nullable
   private ImmutableSet<RegistrarContact.Type> contactTypes;
 
+  private RegistrarContact oldContact;
+  private RegistrarContact newContact;
+
   @Override
   protected void init() throws Exception {
     checkArgument(mainParameters.size() == 1,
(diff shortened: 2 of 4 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/871)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @hstonec)

___
*[core/src/main/java/google/registry/tools/RegistrarContactCommand.java, line 249 at r1](https://reviewable.io/reviews/google/nomulus/871#-MLsIBXg4SjYqN3XG3xK:-MLsIBXhEP9FbZ_Sg_Z6:byh33rx) ([raw file](https://github.com/google/nomulus/blob/20b405819e65f8217b85d0ea3cc53bf628c63c6d/core/src/main/java/google/registry/tools/RegistrarContactCommand.java#L249)):*
> ```Java
>   protected String execute() throws Exception {
>     // Save registrarPoc to Datastore and output its response
>     logger.atInfo().log(super.execute());
> ```

RegistrarContact is currently a replicated entity.  When we save it to datastore, if replication is enabled, it will be written to SQL during the replay, meaning that this will actually get "triple written" if this code is called during the replication phase.
I think we need a more principled way to deal with these kinds of entities.  I suspect that if we just rely on the TransactionManager, the "right thing" will happen by itself.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/871)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/tools/RegistrarContactCommand.java, line 249 at r1](https://reviewable.io/reviews/google/nomulus/871#-MLsIBXg4SjYqN3XG3xK:-MLy68ZOB_OgyKJdmOPC:bp5d7w8) ([raw file](https://github.com/google/nomulus/blob/20b405819e65f8217b85d0ea3cc53bf628c63c6d/core/src/main/java/google/registry/tools/RegistrarContactCommand.java#L249)):*

RegistrarContact is currently a replicated entity.  When we save it to datastore, if replication is enabled, it will be written to SQL during the replay, meaning that this will actually get "triple written" if this code is called during the replication phase.
I think we need a more principled way to deal with these kinds of entities.  I suspect that if we just rely on the TransactionManager, the "right thing" will happen by itself.
</blockquote></details>

The correct solution here might actually be that `RegistrarPoc` shouldn't be a replicated entity.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/871)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/RegistrarContactCommand.java, line 249 at r1](https://reviewable.io/reviews/google/nomulus/871#-MLsIBXg4SjYqN3XG3xK:-MLyFKDo1ylMIUmakR0N:b-m137h6) ([raw file](https://github.com/google/nomulus/blob/20b405819e65f8217b85d0ea3cc53bf628c63c6d/core/src/main/java/google/registry/tools/RegistrarContactCommand.java#L249)):*

The correct solution here might actually be that `RegistrarPoc` shouldn't be a replicated entity.
</blockquote></details>

I wouldn't rule that out, but let's be careful with this.  I think the question of whether an entity can be non-replicated might come down to one of foreign key relationships.  IOW, if there is an FK relationship between the entity and a replicated entity, then that entity must also be a replicated entity because otherwise those key constraints may break if the entities in the relationship are not added/deleted in order.  RegistrarPoc currently has an FK constraint referencing Registrar.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/871)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/RegistrarContactCommand.java, line 249 at r1](https://reviewable.io/reviews/google/nomulus/871#-MLsIBXg4SjYqN3XG3xK:-MLyR6wP2Bt6gzisj03a:b6lsskv) ([raw file](https://github.com/google/nomulus/blob/20b405819e65f8217b85d0ea3cc53bf628c63c6d/core/src/main/java/google/registry/tools/RegistrarContactCommand.java#L249)):*

I wouldn't rule that out, but let's be careful with this.  I think the question of whether an entity can be non-replicated might come down to one of foreign key relationships.  IOW, if there is an FK relationship between the entity and a replicated entity, then that entity must also be a replicated entity because otherwise those key constraints may break if the entities in the relationship are not added/deleted in order.  RegistrarPoc currently has an FK constraint referencing Registrar. 
</blockquote></details>

Another thing to consider is that (assuming we can use TransactionManager in nomulus tool in the same way that we do server-side) for replicated entities we don't need to do any of this kind of thing at all -- we get asynchronous dual-write, behavior changes during the phase switches, and probably a few other benefits I can't even come up with off of the top of my head for free.

I think at this point the default choice should be replicated and that we should only do non-replicated if there is a good reason to.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/871)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/tools/RegistrarContactCommand.java, line 249 at r1](https://reviewable.io/reviews/google/nomulus/871#-MLsIBXg4SjYqN3XG3xK:-MMv9wgO2OMAm_JLTiOn:b6nxah2) ([raw file](https://github.com/google/nomulus/blob/20b405819e65f8217b85d0ea3cc53bf628c63c6d/core/src/main/java/google/registry/tools/RegistrarContactCommand.java#L249)):*

Another thing to consider is that (assuming we can use TransactionManager in nomulus tool in the same way that we do server-side) for replicated entities we don't need to do any of this kind of thing at all -- we get asynchronous dual-write, behavior changes during the phase switches, and probably a few other benefits I can't even come up with off of the top of my head for free.

I think at this point the default choice should be replicated and that we should only do non-replicated if there is a good reason to. 
</blockquote></details>

We need a larger team-wide discussion than can happen here in these comments.  But I think we all can agree on the fact that, whatever approach `Registrar` and `RegistrarPoc` take, they should both use the same one.  Currently that's not the case.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/871)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/RegistrarContactCommand.java, line 249 at r1](https://reviewable.io/reviews/google/nomulus/871#-MLsIBXg4SjYqN3XG3xK:-MMvCrcN8r9JpvmaVnnZ:b-xsqhre) ([raw file](https://github.com/google/nomulus/blob/20b405819e65f8217b85d0ea3cc53bf628c63c6d/core/src/main/java/google/registry/tools/RegistrarContactCommand.java#L249)):*

We need a larger team-wide discussion than can happen here in these comments.  But I think we all can agree on the fact that, whatever approach `Registrar` and `RegistrarPoc` take, they should both use the same one.  Currently that's not the case.
</blockquote></details>

I agree that we need a larger discussion on this (we should probably do that early next week) and also that Registrar and RegistrarContact should use the same approach.
However, I would point out that both Registrar and RegistrarContact currently both implement DatastoreAndSqlEntity so they are, in fact, both replicated as far as the existing system is concerned. I think the replay tests I've been working on currently rely on this behavior for initialization of the test registrars.  The dual-write of Registrar in nomulus tool is currently the discrepancy, and if we want to retain that and make Registrar and RegistrarContact both non-replicated, we're going to have to deal with populating these entities both in the unit tests and during initialization testing.

PR #874, opened 2020-11-14
Use DomainHistoryVKey to restore symmetric VKey

hstonec:
This PR introduced the `DomainHistoryVKey` class to replace the usage of `VKey<HistoryEntry>` with parent of `DomainBase` as an entity field, i.e. replace `VKey<HistoryEntry> redemptionHistoryEntry` with `DomainHistoryVKey redemptionHistoryEntry` in `AllocationToken` class.

The main issue of using `VKey<HistoryEntry>` is there is no easy way for us to restore a symmetric VKey when a `VKey<HistoryEntry>` field is loaded from Postgresql. This is because the `ofyKey` of `VKey<HistoryEntry>` has two columns, i.e. `repoId`(a string) and `historyEntryId` (an integer), but the JPA's `AttributeConverter` doesn't support mapping two table columns to one Java class.

So, to solve this problem, `DomainHistoryVKey` is marked as `@Embeddable` so its `domainRepoId` and `historyEntryId` field will become two columns in the SQL table, and when a `DomainHistoryVKey` is loaded both fields can get assigned correctly and we are able to use a `@PostLoad` method to reconstruct a symmetric VKey from them.

Here are some questions and answers may help the reviewer to understand what happens:

1. Why do we only add `DomainHistoryVKey`? What if we need to convert `VKey<HistoryEntry>` for `HostResource` and `ContactResource`?
    * Surprisingly, `DomainHisotryVKey` is probably the only thing we need for `VKey<HistoryEntry>`. `VKey<HistoryEntry> redemptionHistoryEntry` in `AllocationToken` class is the only use case I can find in the production code, and it is a `VKey<HistoryEntry>` with parent of `DomainBase`.

2. Why do we need to add `DomainHistoryVKey` instead of making `VKey` `@Embeddable`?
    * `VKey` works fine for keys having only one column, e.g. `DomainBase`, so we still want to keep it.

Also, in the subsequent PRs, I will try to apply the similar pattern to `VKey<BillingEvent.OneTime>`, `VKey<PollMessage>`, etc., which are not handled very well in current solution.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/874)


mindhog (changes requested):
> Reviewed 27 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/874)* status: all files reviewed, 2 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/persistence/DomainHistoryVKey.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/874#-MMHoPcs9AVp1afNAFKk:-MMHoPcs9AVp1afNAFKl:bjbrwuu) ([raw file](https://github.com/google/nomulus/blob/62ec977b739aa2836820e6bf45854796f2b2bd49/core/src/main/java/google/registry/persistence/DomainHistoryVKey.java#L49)):*
> ```Java
>     this.kind = HistoryEntry.class;
>     this.ofyKey =
>         Key.create(Key.create(DomainBase.class, domainRepoId), HistoryEntry.class, domainHistoryId);
> ```

Don't we also need to deal with the case of registry 1.0 keys here? (cross-tld key/history id, IIRC)
___
*[core/src/test/resources/google/registry/model/schema.txt, line 917 at r1](https://reviewable.io/reviews/google/nomulus/874#-MMHrM7rBZVDLyjhHRKd:-MMHrM7t0lw3MzwasJwc:b-j8vs5o) ([raw file](https://github.com/google/nomulus/blob/62ec977b739aa2836820e6bf45854796f2b2bd49/core/src/test/resources/google/registry/model/schema.txt#L917)):*
> ```vimhelp
>   java.lang.Object sqlKey;
>   java.lang.String domainRepoId;
> }
> ```

This is a little disturbing because a VKey has to load and save as an ofy.Key in datastore (though I get why the schema generation would get confused by this, ).

Can you verify that this is still the case against keys on crash?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/874)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/DomainHistoryVKey.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/874#-MMHoPcs9AVp1afNAFKk:-MMMZ3ip3bvUMfno-ppu:b-6835v2) ([raw file](https://github.com/google/nomulus/blob/62ec977b739aa2836820e6bf45854796f2b2bd49/core/src/main/java/google/registry/persistence/DomainHistoryVKey.java#L49)):*

Don't we also need to deal with the case of registry 1.0 keys here? (cross-tld key/history id, IIRC)
</blockquote></details>

It seems that we no longer(or never) have registry 1.0 keys for `HistoryEntry`? I tried [this query](https://console.cloud.google.com/bigquery?sq=1080941367941:44c2889f970a4cca82caf4cec2ab5eea) in production but didn't find any one with registry 1.0 key.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 917 at r1](https://reviewable.io/reviews/google/nomulus/874#-MMHrM7rBZVDLyjhHRKd:-MMMZsJzEw1b1tUxrqyf:b-ma5git) ([raw file](https://github.com/google/nomulus/blob/62ec977b739aa2836820e6bf45854796f2b2bd49/core/src/test/resources/google/registry/model/schema.txt#L917)):*

This is a little disturbing because a VKey has to load and save as an ofy.Key in datastore (though I get why the schema generation would get confused by this, ).

Can you verify that this is still the case against keys on crash?
</blockquote></details>

Yeah, I deployed the change to crash and created an allocation token and used it to create a domain successfully, also verified its `redemptionHistoryEntry` field is set as expected through this process.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/874)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #887, opened 2020-11-24
Add an action to replay commit logs to SQL

gbrodman:
- We import from the commit-log GCS files so that we are sure that we
have a consistent snapshot.
- We use the object weighting (moved to ObjectWeights) to verify that we
are replaying objects in the correct order.
- We add a config setting (default false) of whether or not to replay
the logs.
- The action is triggered after the export if the aforementioned config
setting is on.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/887)

### DIFF
--- a/core/src/main/java/google/registry/backup/CommitLogImports.java
+++ b/core/src/main/java/google/registry/backup/CommitLogImports.java
@@ -68,7 +68,8 @@ public final class CommitLogImports {
       ImmutableList.Builder<VersionedEntity> currentTransactionBuilder =
           new ImmutableList.Builder<>();
 
-      for (ImmutableObject currentObject : ImmutableList.copyOf(commitLogs)) {
+      while (commitLogs.hasNext()) {
+        ImmutableObject currentObject = commitLogs.next();
         if (currentObject instanceof CommitLogManifest) {
           // CommitLogManifest means we are starting a new transaction
           addIfNonempty(resultBuilder, currentTransactionBuilder);
--- a/core/src/main/java/google/registry/backup/CommitLogImports.java
+++ b/core/src/main/java/google/registry/backup/CommitLogImports.java
@@ -15,10 +15,10 @@
 package google.registry.backup;
 
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.backup.BackupUtils.createDeserializingIterator;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Streams;
 import google.registry.model.ImmutableObject;
 import google.registry.model.ofy.CommitLogCheckpoint;
 import google.registry.model.ofy.CommitLogManifest;
@@ -31,7 +31,6 @@ import java.io.InputStream;
 import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.util.Iterator;
-import java.util.stream.Stream;
 
 /**
  * Helpers for reading CommitLog records from a file.
@@ -42,6 +41,55 @@ import java.util.stream.Stream;
 public final class CommitLogImports {
 
   private CommitLogImports() {}
+
+  /**
+   * Returns entities in an {@code inputStream} (from a single CommitLog file) as an {@link
+   * ImmutableList} of {@link ImmutableList}s of {@link VersionedEntity} records where the inner
+   * lists each consist of one transaction. Upon completion the {@code inputStream} is closed.
+   *
+   * <p>The returned list may be empty, since CommitLogs are written at fixed intervals regardless
+   * if actual changes exist. Each sublist, however, will not be empty.
+   *
+   * <p>A CommitLog file starts with a {@link CommitLogCheckpoint}, followed by (repeated)
+   * subsequences of [{@link CommitLogManifest}, [{@link CommitLogMutation}] ...]. Each subsequence
+   * represents the changes in one transaction. The {@code CommitLogManifest} contains deleted
+   * entity keys, whereas each {@code CommitLogMutation} contains one whole entity.
+   */
+  public static ImmutableList<ImmutableList<VersionedEntity>> loadEntitiesByTransaction(
+      InputStream inputStream) {
+    try (AppEngineEnvironment appEngineEnvironment = new AppEngineEnvironment();
+        InputStream input = new BufferedInputStream(inputStream)) {
+      Iterator<ImmutableObject> commitLogs = createDeserializingIterator(input);
+      checkState(commitLogs.hasNext());
+      checkState(commitLogs.next() instanceof CommitLogCheckpoint);
+
+      ImmutableList.Builder<ImmutableList<VersionedEntity>> resultBuilder =
+          new ImmutableList.Builder<>();
+      ImmutableList.Builder<VersionedEntity> currentTransactionBuilder =
+          new ImmutableList.Builder<>();
+
+      for (ImmutableObject currentObject : ImmutableList.copyOf(commitLogs)) {
+        if (currentObject instanceof CommitLogManifest) {
+          // CommitLogManifest means we are starting a new transaction
+          addIfNonempty(resultBuilder, currentTransactionBuilder);
+          currentTransactionBuilder = new ImmutableList.Builder<>();
+          VersionedEntity.fromManifest((CommitLogManifest) currentObject)
+              .forEach(currentTransactionBuilder::add);
+        } else if (currentObject instanceof CommitLogMutation) {
+          currentTransactionBuilder.add(
+              VersionedEntity.fromMutation((CommitLogMutation) currentObject));
+        } else {
+          throw new IllegalStateException(
+              String.format("Unknown entity type %s in commit logs", currentObject.getClass()));
+        }
+      }
+      // Add the last transaction in (if it's not empty)
+      addIfNonempty(resultBuilder, currentTransactionBuilder);
+      return resultBuilder.build();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
 
   /**
    * Returns entities in an {@code inputStream} (from a single CommitLog file) as an {@link
@@ -57,23 +105,9 @@ public final class CommitLogImports {
    * entity keys, whereas each {@code CommitLogMutation} contains one whole entity.
    */
   public static ImmutableList<VersionedEntity> loadEntities(InputStream inputStream) {
-    try (AppEngineEnvironment appEngineEnvironment = new AppEngineEnvironment();
-        InputStream input = new BufferedInputStream(inputStream)) {
-      Iterator<ImmutableObject> commitLogs = createDeserializingIterator(input);
-      checkState(commitLogs.hasNext());
-      checkState(commitLogs.next() instanceof CommitLogCheckpoint);
-
-      return Streams.stream(commitLogs)
-          .map(
-              e ->
-                  e instanceof CommitLogManifest
-                      ? VersionedEntity.fromManifest((CommitLogManifest) e)
-                      : Stream.of(VersionedEntity.fromMutation((CommitLogMutation) e)))
-          .flatMap(s -> s)
-          .collect(ImmutableList.toImmutableList());
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+    return loadEntitiesByTransaction(inputStream).stream()
+        .flatMap(ImmutableList::stream)
+        .collect(toImmutableList());
   }
 
   /** Covenience method that adapts {@link #loadEntities(InputStream)} to a {@link File}. */
@@ -91,5 +125,14 @@ public final class CommitLogImports {
    */
   public static ImmutableList<VersionedEntity> loadEntities(ReadableByteChannel channel) {
     return loadEntities(Channels.newInputStream(channel));
+  }
+
+  private static void addIfNonempty(
+      ImmutableList.Builder<ImmutableList<VersionedEntity>> resultBuilder,
+      ImmutableList.Builder<VersionedEntity> currentTransactionBuilder) {
+    ImmutableList<VersionedEntity> currentTransaction = currentTransactionBuilder.build();
+    if (!currentTransaction.isEmpty()) {
+      resultBuilder.add(currentTransaction);
+    }
   }
 }
--- a/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java
+++ b/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java
@@ -79,7 +79,6 @@ public final class ExportCommitLogDiffAction implements Runnable {
   @Inject @Config("commitLogDiffExportBatchSize") int batchSize;
   @Inject @Parameter(LOWER_CHECKPOINT_TIME_PARAM) DateTime lowerCheckpointTime;
   @Inject @Parameter(UPPER_CHECKPOINT_TIME_PARAM) DateTime upperCheckpointTime;
-
   @Inject ExportCommitLogDiffAction() {}
 
   @Override
--- a/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java
+++ b/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java
@@ -14,6 +14,8 @@
 
 package google.registry.backup;
 
+import static com.google.appengine.api.taskqueue.QueueFactory.getQueue;
+import static com.google.appengine.api.taskqueue.TaskOptions.Builder.withUrl;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Verify.verifyNotNull;
@@ -39,6 +41,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 import com.google.common.flogger.FluentLogger;
 import com.googlecode.objectify.Key;
+import google.registry.config.RegistryConfig;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.ImmutableObject;
 import google.registry.model.ofy.CommitLogBucket;
@@ -48,6 +51,7 @@ import google.registry.model.ofy.CommitLogMutation;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
+import google.registry.util.TaskQueueUtils;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Collection;
@@ -79,6 +83,9 @@ public final class ExportCommitLogDiffAction implements Runnable {
   @Inject @Config("commitLogDiffExportBatchSize") int batchSize;
   @Inject @Parameter(LOWER_CHECKPOINT_TIME_PARAM) DateTime lowerCheckpointTime;
   @Inject @Parameter(UPPER_CHECKPOINT_TIME_PARAM) DateTime upperCheckpointTime;
+
+  @Inject TaskQueueUtils taskQueueUtils;
+
   @Inject ExportCommitLogDiffAction() {}
 
   @Override
@@ -137,6 +144,13 @@ public final class ExportCommitLogDiffAction implements Runnable {
       throw new RuntimeException(e);
     }
     logger.atInfo().log("Exported %d manifests in total", sortedKeys.size());
+    if (RegistryConfig.getCloudSqlReplayCommitLogs()) {
+      // If enabled, trigger the SQL replay
+      logger.atInfo().log("Triggering ReplayCommitLogsToSqlAction");
+      taskQueueUtils.enqueue(
+          getQueue(ReplayCommitLogsToSqlAction.QUEUE_NAME),
+          withUrl(ReplayCommitLogsToSqlAction.PATH));
+    }
   }
 
   /**
--- a/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java
+++ b/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java
@@ -14,8 +14,6 @@
 
 package google.registry.backup;
 
-import static com.google.appengine.api.taskqueue.QueueFactory.getQueue;
-import static com.google.appengine.api.taskqueue.TaskOptions.Builder.withUrl;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Verify.verifyNotNull;
@@ -41,7 +39,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 import com.google.common.flogger.FluentLogger;
 import com.googlecode.objectify.Key;
-import google.registry.config.RegistryConfig;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.ImmutableObject;
 import google.registry.model.ofy.CommitLogBucket;
@@ -51,7 +48,6 @@ import google.registry.model.ofy.CommitLogMutation;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
-import google.registry.util.TaskQueueUtils;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Collection;
@@ -83,8 +79,6 @@ public final class ExportCommitLogDiffAction implements Runnable {
   @Inject @Config("commitLogDiffExportBatchSize") int batchSize;
   @Inject @Parameter(LOWER_CHECKPOINT_TIME_PARAM) DateTime lowerCheckpointTime;
   @Inject @Parameter(UPPER_CHECKPOINT_TIME_PARAM) DateTime upperCheckpointTime;
-
-  @Inject TaskQueueUtils taskQueueUtils;
 
   @Inject ExportCommitLogDiffAction() {}
 
@@ -144,13 +138,6 @@ public final class ExportCommitLogDiffAction implements Runnable {
       throw new RuntimeException(e);
     }
     logger.atInfo().log("Exported %d manifests in total", sortedKeys.size());
-    if (RegistryConfig.getCloudSqlReplayCommitLogs()) {
-      // If enabled, trigger the SQL replay
-      logger.atInfo().log("Triggering ReplayCommitLogsToSqlAction");
-      taskQueueUtils.enqueue(
-          getQueue(ReplayCommitLogsToSqlAction.QUEUE_NAME),
-          withUrl(ReplayCommitLogsToSqlAction.PATH));
-    }
   }
 
   /**
--- a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java
+++ b/core/src/main/java/google/registry/backup/GcsDiffFileLister.java
@@ -122,7 +122,7 @@ class GcsDiffFileLister {
 
     // Reconstruct the sequence of files by traversing backwards from "lastUpperBoundTime" (i.e. the
     // last file that we found) and finding its previous file until we either run out of files or
-    // get to one that preceeds "fromTime".
+    // get to one that precedes "fromTime".
     //
     // GCS file listing is eventually consistent, so it's possible that we are missing a file. The
     // metadata of a file is sufficient to identify the preceding file, so if we start from the
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -1,0 +1,186 @@
+// Copyright 2020 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.backup;
+
+import static google.registry.backup.ExportCommitLogDiffAction.DIFF_FILE_PREFIX;
+import static google.registry.model.ofy.ObjectWeights.getObjectWeight;
+import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
+import static org.joda.time.Duration.standardHours;
+
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.tools.cloudstorage.GcsFileMetadata;
+import com.google.appengine.tools.cloudstorage.GcsService;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.flogger.FluentLogger;
+import google.registry.config.RegistryConfig;
+import google.registry.model.server.Lock;
+import google.registry.model.translators.VKeyTranslatorFactory;
+import google.registry.persistence.VKey;
+import google.registry.request.Action;
+import google.registry.request.Response;
+import google.registry.request.auth.Auth;
+import google.registry.schema.replay.DatastoreEntity;
+import google.registry.schema.replay.DatastoreOnlyEntity;
+import google.registry.schema.replay.NonReplicatedEntity;
+import google.registry.schema.replay.SqlEntity;
+import google.registry.schema.replay.SqlReplayCheckpoint;
+import google.registry.util.RequestStatusChecker;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.channels.Channels;
+import java.util.Optional;
+import javax.inject.Inject;
+import javax.servlet.http.HttpServletResponse;
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
+
+/** Action that replays commit logs to SQL to keep it up to date. */
+@Action(
+    service = Action.Service.BACKEND,
+    path = ReplayCommitLogsToSqlAction.PATH,
+    method = Action.Method.POST,
+    automaticallyPrintOk = true,
+    auth = Auth.AUTH_INTERNAL_OR_ADMIN)
+public class ReplayCommitLogsToSqlAction implements Runnable {
+
+  static final String QUEUE_NAME = "replay-commit-logs-to-sql";
+  static final String PATH = "/_dr/task/replayCommitLogsToSql";
+
+  private static final int BLOCK_SIZE =
+      1024 * 1024; // Buffer 1mb at a time, for no particular reason.
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+  private static final Duration LEASE_LENGTH = standardHours(1);
+
+  @Inject GcsService gcsService;
+  @Inject Response response;
+  @Inject RequestStatusChecker requestStatusChecker;
+  @Inject GcsDiffFileLister diffLister;
+
+  @Inject
+  ReplayCommitLogsToSqlAction() {}
+
+  @Override
+  public void run() {
+    if (!RegistryConfig.getCloudSqlReplayCommitLogs()) {
+      String message = "ReplayCommitLogsToSqlAction was called but disabled in the config.";
+      logger.atWarning().log(message);
+      // AppEngine will retry on any non-2xx status code, which we don't want in this case.
+      response.setStatus(SC_NO_CONTENT);
+      response.setPayload(message);
+      return;
+    }
+    Optional<Lock> lock =
+        Lock.acquire(
+            this.getClass().getSimpleName(), null, LEASE_LENGTH, requestStatusChecker, false);
+    if (lock.isEmpty()) {
+      String message = "Can't acquire SQL commit log replay lock, aborting";
+      logger.atSevere().log(message);
+      // AppEngine will retry on any non-2xx status code, which we don't want in this case.
+      // Let the next run after the next export happen naturally.
+      response.setStatus(SC_NO_CONTENT);
+      response.setPayload(message);
+      return;
+    }
+    try {
+      jpaTm().transact(this::replayFiles);
+      response.setStatus(HttpServletResponse.SC_OK);
+    } finally {
+      lock.ifPresent(Lock::release);
+    }
+  }
+
+  private void replayFiles() {
+    // Start at the first millisecond we haven't seen yet
+    DateTime fromTime = SqlReplayCheckpoint.get().plusMillis(1);
+    // the diff lister is inclusive, so use everything before the current txn
+    DateTime toTime = jpaTm().getTransactionTime().minusMillis(1);
+    // If there's an inconsistent file set, this will throw IllegalStateException and the job
+    // will try later -- this is likely because an export hasn't finished yet.
+    ImmutableList<GcsFileMetadata> commitLogFiles = diffLister.listDiffFiles(fromTime, toTime);
+    for (GcsFileMetadata metadata : commitLogFiles) {
+      try (InputStream input =
+          Channels.newInputStream(
+              gcsService.openPrefetchingReadChannel(metadata.getFilename(), 0, BLOCK_SIZE))) {
+        // Save the corresponding SQL entities or delete the key if it was a delete
+        CommitLogImports.loadEntities(input).stream()
+            .sorted(ReplayCommitLogsToSqlAction::compareByWeight)
+            .forEach(
+                versionedEntity ->
+                    versionedEntity
+                        .getEntity()
+                        .ifPresentOrElse(
+                            this::handleEntityPut, () -> handleEntityDelete(versionedEntity)));
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+    if (!commitLogFiles.isEmpty()) {
+      // The last checkpoint time isn't necessarily right now, but rather the upper bound of the
+      // last file we processed
+      DateTime upperBoundTime =
+          DateTime.parse(
+              Iterables.getLast(commitLogFiles)
+                  .getFilename()
+                  .getObjectName()
+                  .substring(DIFF_FILE_PREFIX.length()));
+      // if we succeeded, set the last-seen time
+      SqlReplayCheckpoint.set(upperBoundTime);
+    }
+  }
+
+  private void handleEntityPut(Entity entity) {
+    Object ofyPojo = ofy().toPojo(entity);
+    if (ofyPojo instanceof DatastoreEntity) {
+      DatastoreEntity datastoreEntity = (DatastoreEntity) ofyPojo;
+      ImmutableList<SqlEntity> convertedEntities = datastoreEntity.toSqlEntities();
+      jpaTm().putAll(convertedEntities);
+    } else {
+      // this should never happen, but we shouldn't fail on it
+      logger.atSevere().log(
+          "%s does not implement DatastoreEntity, which is necessary for SQL replay.",
+          ofyPojo.getClass());
+    }
+  }
+
+  private void handleEntityDelete(VersionedEntity entityToDelete) {
+    Key key = entityToDelete.key();
+    VKey<?> entityVKey;
+    try {
+      entityVKey = VKeyTranslatorFactory.createVKey(key);
+    } catch (RuntimeException e) {
+      // This means that the key wasn't convertible to VKey through the standard methods or via
+      // a createVKey method. This means that the object isn't persisted in SQL so we ignore it.
+      logger.atInfo().log(
+          "Skipping SQL delete for kind %s since it is not convertible", key.getKind());
+      return;
+    }
+    Class<?> entityClass = entityVKey.getKind();
+    // Delete the key iff the class represents a JPA entity that is replicated
+    if (!NonReplicatedEntity.class.isAssignableFrom(entityClass)
+        && !DatastoreOnlyEntity.class.isAssignableFrom(entityClass)
+        && entityClass.getAnnotation(javax.persistence.Entity.class) != null) {
+      jpaTm().delete(entityVKey);
+    }
+  }
+
+  private static int compareByWeight(VersionedEntity a, VersionedEntity b) {
+    return getObjectWeight(a.key().getKind(), a.getEntity().isEmpty())
+        - getObjectWeight(b.key().getKind(), b.getEntity().isEmpty());
+  }
+}
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -37,7 +37,6 @@ import google.registry.request.auth.Auth;
 import google.registry.schema.replay.DatastoreEntity;
 import google.registry.schema.replay.DatastoreOnlyEntity;
 import google.registry.schema.replay.NonReplicatedEntity;
-import google.registry.schema.replay.SqlEntity;
 import google.registry.schema.replay.SqlReplayCheckpoint;
 import google.registry.util.RequestStatusChecker;
 import java.io.IOException;
@@ -152,8 +151,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
     Object ofyPojo = ofy().toPojo(entity);
     if (ofyPojo instanceof DatastoreEntity) {
       DatastoreEntity datastoreEntity = (DatastoreEntity) ofyPojo;
-      ImmutableList<SqlEntity> convertedEntities = datastoreEntity.toSqlEntities();
-      jpaTm().putAll(convertedEntities);
+      datastoreEntity.toSqlEntity().ifPresent(jpaTm()::put);
     } else {
       // this should never happen, but we shouldn't fail on it
       logger.atSevere().log(
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -26,7 +26,6 @@ import com.google.appengine.api.datastore.Key;
 import com.google.appengine.tools.cloudstorage.GcsFileMetadata;
 import com.google.appengine.tools.cloudstorage.GcsService;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig;
 import google.registry.model.server.Lock;
@@ -59,7 +58,6 @@ import org.joda.time.Duration;
     auth = Auth.AUTH_INTERNAL_OR_ADMIN)
 public class ReplayCommitLogsToSqlAction implements Runnable {
 
-  static final String QUEUE_NAME = "replay-commit-logs-to-sql";
   static final String PATH = "/_dr/task/replayCommitLogsToSql";
 
   private static final int BLOCK_SIZE =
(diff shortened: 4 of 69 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 0 of 14 files reviewed, 10 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 147 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0PXvP0s6mrmQSZgj8:-MN0PXvP0s6mrmQSZgj9:b-k2bscx) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L147)):*
> ```Java
>     }
>     logger.atInfo().log("Exported %d manifests in total", sortedKeys.size());
>     if (RegistryConfig.getCloudSqlReplayCommitLogs()) {
> ```

We need to be able to configure this at runtime rather than having to push a new release to enable it (and then roll back said release to disable it).  Could be as simple as a Datastore entity with a boolean in it and a tool to update said boolean.
___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0Pvcj5wfKg35i-0Ro:-MN0Pvcj5wfKg35i-0Rp:b-qtyd83) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L150)):*
> ```Java
>       // If enabled, trigger the SQL replay
>       logger.atInfo().log("Triggering ReplayCommitLogsToSqlAction");
>       taskQueueUtils.enqueue(
> ```

What's the point of doing this here?  Why not just have this be a separate cron job?  The other action could be running for a long time, right?  So why keep piling up queue items on top of that?  It seems like doing so might lead to a degenerate state.  If we use a cron job directly then cron itself can handle the detection of whether an existing task is already running, and if so, it won't enqueue a new one.  That seems to be a better solution.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 83 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0QjsJ3ZSUUNbWcL_x:-MN0QjsJ3ZSUUNbWcL_y:b7hoeg6) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L83)):*
> ```Java
>       String message = "ReplayCommitLogsToSqlAction was called but disabled in the config.";
>       logger.atWarning().log(message);
>       // AppEngine will retry on any non-2xx status code, which we don't want in this case.
> ```

App Engine
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0QlD6B00EAlgSko5M:-MN0QlD6B00EAlgSko5N:b7hoeg6) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L94)):*
> ```Java
>       String message = "Can't acquire SQL commit log replay lock, aborting";
>       logger.atSevere().log(message);
>       // AppEngine will retry on any non-2xx status code, which we don't want in this case.
> ```

App Engine
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MN0SnboDrlCNUXA1yxH:b-e8r0l9) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*
> ```Java
>     }
>     try {
>       jpaTm().transact(this::replayFiles);
> ```

Does the approach of having a single large PostgreSQL transaction to write *all* changes actually work?  What if, for whatever reason, there's hours of Datastore writes to catch up on?  Do we require exclusive access to all of these tables in PostgreSQL in order for this to work, i.e. if there's anything else anywhere writing to any of them then this will just conflict out and always fail?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0R-Yq1BOzCj44H6Jo:-MN0R-Yq1BOzCj44H6Jp:bgxwgy9) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L102)):*
> ```Java
>     try {
>       jpaTm().transact(this::replayFiles);
>       response.setStatus(HttpServletResponse.SC_OK);
> ```

Log a success message at INFO here after the transaction completes!
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SQ-g5lnvXuPGR04k:-MN0SQ-g5lnvXuPGR04l:b-44jttj) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L112)):*
> ```Java
>     DateTime fromTime = SqlReplayCheckpoint.get().plusMillis(1);
>     // the diff lister is inclusive, so use everything before the current txn
>     DateTime toTime = jpaTm().getTransactionTime().minusMillis(1);
> ```

Is a`toTime` even necessary?  Why not just ingest everything that exists?  Are you trying to protect against time inconsistencies between PostgreSQL and Datastore or something?  Protecting against transactions traveling backwards in time that might cause cyclic reference errors?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TcPM3GTKtzUkWzpN:-MN0TcPM3GTKtzUkWzpO:bcfcbg3) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*
> ```Java
>         // Save the corresponding SQL entities or delete the key if it was a delete
>         CommitLogImports.loadEntities(input).stream()
>             .sorted(ReplayCommitLogsToSqlAction::compareByWeight)
> ```

What if there are multiple updates to a single entity in a single GCS export?  Is that even possible?  If there are, wouldn't the order that they're applied in matter?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TL0d6DzDCFf8jJSx:-MN0TL0d6DzDCFf8jJSy:b5kcbxs) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L161)):*
> ```Java
>   }
> 
>   private void handleEntityDelete(VersionedEntity entityToDelete) {
> ```

Very few entities in our system are actually deleted as a matter of course.  This affects, what, only PollMessages?
___
*[core/src/main/java/google/registry/model/ofy/ObjectWeights.java, line 20 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0RGdoBTWmpvvxwwCE:-MN0RGdoBTWmpvvxwwCF:b-wi5i5t) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/model/ofy/ObjectWeights.java#L20)):*
> ```Java
> import com.google.common.collect.ImmutableMap;
> 
> public class ObjectWeights {
> ```

Needs a Javadoc.

In fact it _really_ needs a Javadoc because I have no idea what any of this is supposed to be.

Is this intended to prevent foreign key reference errors?  If so, "weights" isn't a great name to properly convey that concept.  How about priority or similar.

Also Object is super-generic and has a well-established meaning in Java.  you probably mean Entity, DbEntity, etc.

`DbEntityWritePriorities`?

mindhog (changes requested):
> Reviewed 3 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 3 of 14 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 147 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0PXvP0s6mrmQSZgj8:-MNOlTCS-A55Rzvog0DT:b-vv76ce) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L147)):*

We need to be able to configure this at runtime rather than having to push a new release to enable it (and then roll back said release to disable it).  Could be as simple as a Datastore entity with a boolean in it and a tool to update said boolean.
</blockquote></details>

There's a design doc on this: go/r3.0-migration-config-changes

TL;DR -- put a parameter in datastore or cloud SQL that is an enum corresponding to the migration phase. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 90 at r1](https://reviewable.io/reviews/google/nomulus/887#-MNOp5_d3sRHZQrJvGMZ:-MNOp5_d3sRHZQrJvGM_:bv042py) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L90)):*
> ```Java
>     Optional<Lock> lock =
>         Lock.acquire(
>             this.getClass().getSimpleName(), null, LEASE_LENGTH, requestStatusChecker, false);
> ```

I've pointed out my objections to the use of Lock in these cases, but we should probably discuss this out-of-band.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MNOnzjk6glIdqLnlZ5R:bd3uatx) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

Does the approach of having a single large PostgreSQL transaction to write *all* changes actually work?  What if, for whatever reason, there's hours of Datastore writes to catch up on?  Do we require exclusive access to all of these tables in PostgreSQL in order for this to work, i.e. if there's anything else anywhere writing to any of them then this will just conflict out and always fail?
</blockquote></details>

+1.  I'm not worried about exclusive access to all tables, during this phase the async replication code is the only thing touching those (at least for all _replicated_ entities).  But I do agree with Ben's other points, and, fundamentally, having one JPA transaction for each datastore transaction is just easier to reason about.  
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TcPM3GTKtzUkWzpN:-MNOvb939u3B1eVmxami:b8m548b) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*

What if there are multiple updates to a single entity in a single GCS export?  Is that even possible?  If there are, wouldn't the order that they're applied in matter?
</blockquote></details>

Yeah, this would be a problem.  loadEntities() basically gives you back a list of VersionedEntity in the order that they are saved/deleted based on the transaction ordering, and sorting by weight (IIUIC) will ignore that.
You could conceivably convert the list to a map by Key, in which case the final state of an entity would be recorded in the map.  But once again, it's probably better to just deal with one datastore transaction at a time.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 165 at r1](https://reviewable.io/reviews/google/nomulus/887#-MNOzLK53lw364rTIF3U:-MNOzLK53lw364rTIF3V:b-cidexo) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L165)):*
> ```Java
>     VKey<?> entityVKey;
>     try {
>       entityVKey = VKeyTranslatorFactory.createVKey(key);
> ```

It occurs to me that this wouldn't work in the case of a DatastoreEntity that mapped to multiple SQL entities.  I'm guessing that we don't currently have any of those?  If not, we should probably enforce that by changing toSqlEntities to toSqlEntity.  If we do, I'm not sure how best to deal with this.

mindhog (commented):
> Reviewed 11 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: all files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/ObjectWeights.java, line 20 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0RGdoBTWmpvvxwwCE:-MNP19nU7f8diQgxLGnJ:b25kdqz) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/model/ofy/ObjectWeights.java#L20)):*

Needs a Javadoc.

In fact it _really_ needs a Javadoc because I have no idea what any of this is supposed to be.

Is this intended to prevent foreign key reference errors?  If so, "weights" isn't a great name to properly convey that concept.  How about priority or similar.

Also Object is super-generic and has a well-established meaning in Java.  you probably mean Entity, DbEntity, etc.

`DbEntityWritePriorities`?
</blockquote></details>

I take full responsibility for "weights" as I introduced it, but "priority" works for me, too.

And yes, it is to ensure that entities are stored/deleted in an order that won't cause errors due to foreign key relations.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 7 of 14 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 147 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0PXvP0s6mrmQSZgj8:-MNPH-Cs1upHxarMm1Ey:b-kvwic9) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L147)):*

There's a design doc on this: go/r3.0-migration-config-changes

TL;DR -- put a parameter in datastore or cloud SQL that is an enum corresponding to the migration phase. 
</blockquote></details>

Yeah, we'll need to figure out when that'll be implemented but that'd work 
___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0Pvcj5wfKg35i-0Ro:-MNPH5FA62wQTCauIy0B:b-ab6f3k) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L150)):*

What's the point of doing this here?  Why not just have this be a separate cron job?  The other action could be running for a long time, right?  So why keep piling up queue items on top of that?  It seems like doing so might lead to a degenerate state.  If we use a cron job directly then cron itself can handle the detection of whether an existing task is already running, and if so, it won't enqueue a new one.  That seems to be a better solution.
</blockquote></details>

It shouldn't be running for a while--at least, not any longer than this job.  

For some reason I thought that the export action could export multiple files -- this does not appear to be the case, meaning that we <should> be fine if we just launch the replay action every N minutes. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 83 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0QjsJ3ZSUUNbWcL_x:-MN0X1fC0lqGcWxyua0Z:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L83)):*

App Engine
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 90 at r1](https://reviewable.io/reviews/google/nomulus/887#-MNOp5_d3sRHZQrJvGMZ:-MNPT9SIDxa_Faid1f53:b-qtvofj) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L90)):*

I've pointed out my objections to the use of Lock in these cases, but we should probably discuss this out-of-band.
</blockquote></details>

yes, let's
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0QlD6B00EAlgSko5M:-MN0X1yRAc7s6UoGAwQ2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L94)):*

App Engine
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MN0YkWyC0t2rdlBnNS9:blkswyk) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

+1.  I'm not worried about exclusive access to all tables, during this phase the async replication code is the only thing touching those (at least for all _replicated_ entities).  But I do agree with Ben's other points, and, fundamentally, having one JPA transaction for each datastore transaction is just easier to reason about.  
</blockquote></details>

I believe anything that requires writes would fail / queue up / retry (whatever the normal behavior is for transaction contention which I'm not sure about) 

If we do one JPA transaction per one DS transaction then we'll definitely need the locking mechanism. 

It probably does make the most sense to have one transaction correspond to one transaction, assuming that we won't be writing any other versions of these items 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0R-Yq1BOzCj44H6Jo:-MNPHnEXCWWD8BkVW_Ny:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L102)):*

Log a success message at INFO here after the transaction completes!
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SQ-g5lnvXuPGR04k:-MN0Xi8WCAkARtjLCXhq:be231ew) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L112)):*

Is a`toTime` even necessary?  Why not just ingest everything that exists?  Are you trying to protect against time inconsistencies between PostgreSQL and Datastore or something?  Protecting against transactions traveling backwards in time that might cause cyclic reference errors?
</blockquote></details>

Practically speaking I suppose not, as long as we set the checkpoint time to be the latest file seen (which we do). Removed.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TcPM3GTKtzUkWzpN:-MNPILmL4EwHLRLnaNRQ:b-d117cy) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*

Yeah, this would be a problem.  loadEntities() basically gives you back a list of VersionedEntity in the order that they are saved/deleted based on the transaction ordering, and sorting by weight (IIUIC) will ignore that.
You could conceivably convert the list to a map by Key, in which case the final state of an entity would be recorded in the map.  But once again, it's probably better to just deal with one datastore transaction at a time.
</blockquote></details>

Yeah, we should probably modify CommitLogImports to contain a method that returns something like  ImmutableList<ImmutableList<VersionedEntity>> that is basically this list partitioned by transaction. This is done
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TL0d6DzDCFf8jJSx:-MN0WojE0Tc8DL7M7GFz:bpvkl8d) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L161)):*

Very few entities in our system are actually deleted as a matter of course.  This affects, what, only PollMessages?
</blockquote></details>

Good question, I don't actually know.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 165 at r1](https://reviewable.io/reviews/google/nomulus/887#-MNOzLK53lw364rTIF3U:-MNPSwU3EGar3ux8HZVN:b-9mabuf) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L165)):*

It occurs to me that this wouldn't work in the case of a DatastoreEntity that mapped to multiple SQL entities.  I'm guessing that we don't currently have any of those?  If not, we should probably enforce that by changing toSqlEntities to toSqlEntity.  If we do, I'm not sure how best to deal with this.
</blockquote></details>

AFAIK we don't have any entities that map to multiple entities (which is good). 

Yeah, we could convert the signature of DatastoreEntity to be something like `public Optional<SqlEntity> toSqlEntity() {...}`
___
*[core/src/main/java/google/registry/model/ofy/ObjectWeights.java, line 20 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0RGdoBTWmpvvxwwCE:-MNPnSynDHIxn7eQWUly:b-lsplri) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/model/ofy/ObjectWeights.java#L20)):*

Needs a Javadoc.

In fact it _really_ needs a Javadoc because I have no idea what any of this is supposed to be.

Is this intended to prevent foreign key reference errors?  If so, "weights" isn't a great name to properly convey that concept.  How about priority or similar.

Also Object is super-generic and has a well-established meaning in Java.  you probably mean Entity, DbEntity, etc.

`DbEntityWritePriorities`?
</blockquote></details>

Yeah, that sounds good. PTAL?

mindhog (changes requested):
> Reviewed 7 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: all files reviewed, 16 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/CommitLogImports.java, line 71 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTSTgD48KZjaVah4Fn:-MNTSTgD48KZjaVah4Fo:b-ngxmim) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/CommitLogImports.java#L71)):*
> ```Java
>           new ImmutableList.Builder<>();
> 
>       for (ImmutableObject currentObject : ImmutableList.copyOf(commitLogs)) {
> ```

It's probably more efficient to just use the commitLogs iterator directly (as opposed to copying the entire contents into an immutable list) and likely with little impact to readability.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MNTYM0FAJGWxvSzLonN:b2akyue) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

I believe anything that requires writes would fail / queue up / retry (whatever the normal behavior is for transaction contention which I'm not sure about) 

If we do one JPA transaction per one DS transaction then we'll definitely need the locking mechanism. 

It probably does make the most sense to have one transaction correspond to one transaction, assuming that we won't be writing any other versions of these items 
</blockquote></details>

You can still omit the lock if you retrieve the last datastore transaction time from within the SQL transaction. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTmUAp1TWnU4RXiUYd:-MNTmUAp1TWnU4RXiUYe:bag9czh) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*
> ```Java
>         for (ImmutableList<VersionedEntity> transaction :
>             CommitLogImports.loadEntitiesByTransaction(input)) {
>           jpaTm().transact(() -> replayTransaction(transaction));
> ```

So, the problem with replaying each transaction as opposed to the entire commit log is that you need to record when each transaction was replayed as opposed to just recording that the entire commit log was replayed at the end.  As it stands, if there were an error during replay this code would double-replay the transactions prior to the error, and I don't think that can be made to work in all cases because of foreign key dependencies.

I don't think we get a per-transaction timestamp in the manifest, do we?  If we did, you could persist that timestamp and check against it during the replay.  If not, you'll probably want to record the index of the current transaction as well as the commit log timestamp.
___
*[core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java, line 26 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTWoI04OgMDsBDl-3P:-MNTWoI04OgMDsBDl-3Q:b-l914wv) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java#L26)):*
> ```Java
>    *
>    * <p>Here, "priority" means the order in which the class should be inserted / updated in a
>    * transaction with respect to instances of other classes. By default, all classes have a weight
> ```

"priority"
___
*[core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java, line 41 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTWrOW20JbutP4Uqcb:-MNTWrOW20JbutP4Uqcc:bpi9gna) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java#L41)):*
> ```
> CLASS_WEIGHTS
> ```
CLASS_PRIORITIES (also "priorities" instead of "weights" in the line above.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 11 of 14 files reviewed, 16 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/CommitLogImports.java, line 71 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTSTgD48KZjaVah4Fn:-MNTwvE40PlB6zJN0X1i:b-vmrddi) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/CommitLogImports.java#L71)):*

It's probably more efficient to just use the commitLogs iterator directly (as opposed to copying the entire contents into an immutable list) and likely with little impact to readability.
</blockquote></details>

eh yeah i was hoping to have to have the two lines calling hasNext() and next() in an effort to avoid possible errors in the future, but it turns out intellij yells at you if you have a hasNext loop without invoking `next` at any point inside it 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTmUAp1TWnU4RXiUYd:-MNUn-gk0IC2MKO9LnNO:bjuqytu) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*

So, the problem with replaying each transaction as opposed to the entire commit log is that you need to record when each transaction was replayed as opposed to just recording that the entire commit log was replayed at the end.  As it stands, if there were an error during replay this code would double-replay the transactions prior to the error, and I don't think that can be made to work in all cases because of foreign key dependencies.

I don't think we get a per-transaction timestamp in the manifest, do we?  If we did, you could persist that timestamp and check against it during the replay.  If not, you'll probably want to record the index of the current transaction as well as the commit log timestamp.
</blockquote></details>

So we discussed this offline, and we basically have two choices.

1. Have one SQL transaction per commit log file. This means that we only need to store the timestamp for each file we see, and if we replay the transactions in order within each file we will end up with the consistent result. The primary downside to this is the commit log file size -- median ~410 KB, but we have had periods of time up to a couple hours where each commit log file (one every ~3 minutes) can be 100-200 MB in size. That might be a large Postgres transaction. 

2. Have one SQL transaction per Datastore transaction. While this would have smaller transactions, we'd need to store and use an additional field, marking essentially how far we've gotten in each commit log file (in case replay of a transaction fails partway through a diff file). By storing a commit-log-file checkpoint and a transaction-index checkpoint we could guarantee that we start where we left off in the event of failure. The downside of this is that the logic is more complicated (and thus more bug-prone) and it involves creating an additional field. 


The currently implemented system in this PR is (1)
___
*[core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java, line 26 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTWoI04OgMDsBDl-3P:-MNTwjqF1qkzxCkz1bwR:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java#L26)):*

"priority"
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java, line 41 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTWrOW20JbutP4Uqcb:-MNTwjSL4hAvXFVXVf_M:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/model/ofy/EntityWritePriorities.java#L41)):*

> ```
> CLASS_WEIGHTS
> ```
CLASS_PRIORITIES (also "priorities" instead of "weights" in the line above.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 11 of 14 files reviewed, 13 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0R-Yq1BOzCj44H6Jo:-MNVIoNj6nLho0mQUN1G:bit4zi1) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L102)):*

Done.
</blockquote></details>

Is it done?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TL0d6DzDCFf8jJSx:-MNVJ68HCPs5CgdYSTbe:b-mygq27) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L161)):*

Good question, I don't actually know.
</blockquote></details>

Worth looking into just so you know what's going on.
___
*[core/src/main/java/google/registry/model/ofy/ObjectWeights.java, line 20 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0RGdoBTWmpvvxwwCE:-MNVJBzG9gbu4fhamALN:b-pia5tq) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/model/ofy/ObjectWeights.java#L20)):*

Yeah, that sounds good. PTAL?
</blockquote></details>

It still needs a Javadoc.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 11 of 14 files reviewed, 13 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0R-Yq1BOzCj44H6Jo:-MNZrK2D-rhSaz8o1wrp:bc2e6iw) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L102)):*

Is it done?
</blockquote></details>

Yes. We log an info message after every commit log file (end of `processFile` and we log the number of files processed at the end of `replayFiles`
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TL0d6DzDCFf8jJSx:-MNZsS0V387yFNGEi1mb:b-tfq6k8) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L161)):*

Worth looking into just so you know what's going on.
</blockquote></details>

Seems like 

- poll messages in various ack or domain restore scenarios
- in an explicit resource transfer (e.g. DomainTransferApproveFlow) we delete the billing event and poll messages associated with a server-approved transfer

I don't see any other usages of `TransactionManager::delete*` (any of the delete methods) 
___
*[core/src/main/java/google/registry/model/ofy/ObjectWeights.java, line 20 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0RGdoBTWmpvvxwwCE:-MNZs22OCo8CwkoFPGhO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/model/ofy/ObjectWeights.java#L20)):*

It still needs a Javadoc.
</blockquote></details>

Done.

mindhog (approved):
I'm good with this change in it's current form.  We're going to want to run it against actual commit logs on an environment once we can import a full dataset, at that point hopefully we'll be able to get a better sense of whether we can live with the XXL transactions.

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887#-:-MN_-dxR6looX7FFYEy5:bhv0i2m)* status: all files reviewed, 11 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 165 at r1](https://reviewable.io/reviews/google/nomulus/887#-MNOzLK53lw364rTIF3U:-MN_-MESDJGbaH9Ru2t4:bw6me8p) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L165)):*

AFAIK we don't have any entities that map to multiple entities (which is good). 

Yeah, we could convert the signature of DatastoreEntity to be something like `public Optional<SqlEntity> toSqlEntity() {...}`
</blockquote></details>

Opened b/174680822

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: all files reviewed, 11 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MN_1veN7ZlTZxjQVZd9:bfi042j) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

You can still omit the lock if you retrieve the last datastore transaction time from within the SQL transaction. 
</blockquote></details>

I think we probably still want the lock here, especially since we're not storing per-transaction-time replay checkpoints 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 165 at r1](https://reviewable.io/reviews/google/nomulus/887#-MNOzLK53lw364rTIF3U:-MN_22Sp2Zd5v3mFtQd4:bi8intp) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L165)):*

Opened b/174680822
</blockquote></details>

Hah, opened https://github.com/google/nomulus/pull/894 a few minutes ago

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTmUAp1TWnU4RXiUYd:-MN_CKgm6diqFr9OFxqe:b-sy508b) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*

So we discussed this offline, and we basically have two choices.

1. Have one SQL transaction per commit log file. This means that we only need to store the timestamp for each file we see, and if we replay the transactions in order within each file we will end up with the consistent result. The primary downside to this is the commit log file size -- median ~410 KB, but we have had periods of time up to a couple hours where each commit log file (one every ~3 minutes) can be 100-200 MB in size. That might be a large Postgres transaction. 

2. Have one SQL transaction per Datastore transaction. While this would have smaller transactions, we'd need to store and use an additional field, marking essentially how far we've gotten in each commit log file (in case replay of a transaction fails partway through a diff file). By storing a commit-log-file checkpoint and a transaction-index checkpoint we could guarantee that we start where we left off in the event of failure. The downside of this is that the logic is more complicated (and thus more bug-prone) and it involves creating an additional field. 


The currently implemented system in this PR is (1)
</blockquote></details>

It doesn't seem to me like we're getting the order of the writes within the transaction though?  The GCS import spans 3 minutes of changes.  There are often going to be many cases where multiple writes are affecting the same entity, and need to be replayed in Cloud SQL in the same order that they were originally written to end up in the same correct end state.  Where is that happening here?  All I see is ordering by weight, which would be incorrect.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: all files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0R-Yq1BOzCj44H6Jo:-MN_CqAv9GopJ4HuCq0f:b-fo1ogi) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L102)):*

Yes. We log an info message after every commit log file (end of `processFile` and we log the number of files processed at the end of `replayFiles`
</blockquote></details>

"Replayed 26892476 commit log files to SQL" isn't *exactly* a success message though.  It doesn't actually have any wording in it indicating overall success.  If you didn't know anything about this action you'd have to dig into the code to realize that that is the final desired end logging message.  Can you add something indicating success of the overall action, or adjust the existing wording, or such?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 90 at r3](https://reviewable.io/reviews/google/nomulus/887#-MN_DI5pBg9pToAR8l8q:-MN_DI5pBg9pToAR8l8r:bg0sfc0) ([raw file](https://github.com/google/nomulus/blob/ab64f4255c66f687140b5820ed39e4a8c829a4ea/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L90)):*
> ```Java
>             this.getClass().getSimpleName(), null, LEASE_LENGTH, requestStatusChecker, false);
>     if (lock.isEmpty()) {
>       String message = "Can't acquire SQL commit log replay lock, aborting";
> ```

Nit:  Should end with a period.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 117 at r3](https://reviewable.io/reviews/google/nomulus/887#-MN_DDx7CZ45LEQlvY0n:-MN_DDx7CZ45LEQlvY0o:bg0sfc0) ([raw file](https://github.com/google/nomulus/blob/ab64f4255c66f687140b5820ed39e4a8c829a4ea/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L117)):*
> ```Java
>       jpaTm().transact(() -> processFile(metadata));
>     }
>     logger.atInfo().log("Replayed %d commit log files to SQL", commitLogFiles.size());
> ```

Nit:  Should end with a period.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 133 at r3](https://reviewable.io/reviews/google/nomulus/887#-MN_DLcU4dgLK9v382uD:-MN_DLcU4dgLK9v382uE:bg0sfc0) ([raw file](https://github.com/google/nomulus/blob/ab64f4255c66f687140b5820ed39e4a8c829a4ea/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L133)):*
> ```Java
>               metadata.getFilename().getObjectName().substring(DIFF_FILE_PREFIX.length()));
>       SqlReplayCheckpoint.set(checkpoint);
>       logger.atInfo().log("Replayed %d transactions from commit log file", allTransactions.size());
> ```

Nit:  Should end with a period.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 173 at r3](https://reviewable.io/reviews/google/nomulus/887#-MN_DP5k8hn156-jLm4W:-MN_DP5l4YUTQ48Qg53s:bg0sfc0) ([raw file](https://github.com/google/nomulus/blob/ab64f4255c66f687140b5820ed39e4a8c829a4ea/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L173)):*
> ```Java
>       // a createVKey method. This means that the object isn't persisted in SQL so we ignore it.
>       logger.atInfo().log(
>           "Skipping SQL delete for kind %s since it is not convertible", key.getKind());
> ```

Nit:  Should end with a period.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTmUAp1TWnU4RXiUYd:-MN_G4Qc6RaMQxFr3eu9:b-vdat0v) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*

It doesn't seem to me like we're getting the order of the writes within the transaction though?  The GCS import spans 3 minutes of changes.  There are often going to be many cases where multiple writes are affecting the same entity, and need to be replayed in Cloud SQL in the same order that they were originally written to end up in the same correct end state.  Where is that happening here?  All I see is ordering by weight, which would be incorrect.
</blockquote></details>

I'm not sure what you're seeing but we definitely do this. We load the entities by transaction, then only apply the weights per transaction.  It's in the `replayTransaction` method.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 12 of 14 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 82 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LLjL4KQIlsWKarJU:-MN_LLjL4KQIlsWKarJV:b32vmt) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L82)):*
> ```Java
>   @Inject @Parameter(LOWER_CHECKPOINT_TIME_PARAM) DateTime lowerCheckpointTime;
>   @Inject @Parameter(UPPER_CHECKPOINT_TIME_PARAM) DateTime upperCheckpointTime;
> ```

??
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MN_L_QkDVIEUgSHrYhF:b-5iqutg) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

I think we probably still want the lock here, especially since we're not storing per-transaction-time replay checkpoints 
</blockquote></details>

It sounds like we want to slow down `ResaveAllEppResources` (i.e. have it run on fewer shards) at least for the duration of the migration, so that we mitigate this problem.  No need to commit 250 MB transactions to Cloud SQL if we don't need to.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0TL0d6DzDCFf8jJSx:-MN_KG9_0G_HBTESddlR:b-of4brp) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L161)):*

Seems like 

- poll messages in various ack or domain restore scenarios
- in an explicit resource transfer (e.g. DomainTransferApproveFlow) we delete the billing event and poll messages associated with a server-approved transfer

I don't see any other usages of `TransactionManager::delete*` (any of the delete methods) 
</blockquote></details>

OK, so poll messages always end their lifecycle by being deleted, along with speculative billing events (which exist only for transfers, and which become counterfactual i.e. deleted if the transfer doesn't end up happening).  Makes sense.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 122 at r2](https://reviewable.io/reviews/google/nomulus/887#-MNTmUAp1TWnU4RXiUYd:-MN_K7PL6iqLw1uESJv2:bm5m5b0) ([raw file](https://github.com/google/nomulus/blob/fb29bc145a596de41d75136837e9cf3ec8cff0ac/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L122)):*

I'm not sure what you're seeing but we definitely do this. We load the entities by transaction, then only apply the weights per transaction.  It's in the `replayTransaction` method. 
</blockquote></details>

OK, good.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MN_LOX56ElG9DKO7rES:b279gge) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*
> ```Java
> import org.joda.time.Duration;
> 
> /** Action that replays commit logs to SQL to keep it up to date. */
> ```

s/SQL/Cloud SQL/

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 12 of 14 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java, line 82 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LLjL4KQIlsWKarJU:-MN_N-PsCNdOX4rHNqPN:b-fmseb1) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ExportCommitLogDiffAction.java#L82)):*

??
</blockquote></details>

i blame autoformat
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0SnboDrlCNUXA1yxG:-MN_Oj3q0e5vJsNeG1-Z:b6o4g6a) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

It sounds like we want to slow down `ResaveAllEppResources` (i.e. have it run on fewer shards) at least for the duration of the migration, so that we mitigate this problem.  No need to commit 250 MB transactions to Cloud SQL if we don't need to.
</blockquote></details>

Yeah, definitely. We can potentially experiment with setting a particular number of shards. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/887#-MN0R-Yq1BOzCj44H6Jo:-MN_Otg3CB2f7TGCdg_w:bb7oryy) ([raw file](https://github.com/google/nomulus/blob/2621a7863267428c3a9e1380fe93256edef503b6/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L102)):*

"Replayed 26892476 commit log files to SQL" isn't *exactly* a success message though.  It doesn't actually have any wording in it indicating overall success.  If you didn't know anything about this action you'd have to dig into the code to realize that that is the final desired end logging message.  Can you add something indicating success of the overall action, or adjust the existing wording, or such?
</blockquote></details>

sure
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MN_N4cXE9tArYtm8waj:b-896fix) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*

s/SQL/Cloud SQL/
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MNeOIq4DKpVw3lKHJnI:bkzb8ep) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*

Done.
</blockquote></details>

Doesn't look done to me.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MNeOTQGBNHVABCzRM92:bxcmmyg) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*

Doesn't look done to me.
</blockquote></details>



??

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MNeOfoO-tRYfh6qJkL4:bi94moc) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*



??
</blockquote></details>

Unless you mean in the class name itself?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 10 of 13 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MNeQ3_b1FpFuHc2Fcea:bgxw5jy) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*

Unless you mean in the class name itself? 
</blockquote></details>

It wasn't showing the diff across all commits, despite reloading.  Closing and reopening the tab did the trick though.  I don't know why it keeps somehow only showing some of the commits' diffs rather than all of them; I only ever want the latter.  Using a commit --amend and git push -f workflow is one way to completely avoid this problem.  (I've done that for a lot of my PRs.)  They all get squashed upon merge anyway.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887#-:-MNeQuMjE2zcb9kPKYL3:bnfp4nl)* status: 10 of 13 files reviewed, 4 unresolved discussions (waiting on @mindhog)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/887)* status: 10 of 13 files reviewed, 4 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/887#-MN_LOX56ElG9DKO7rER:-MNeRKSO9kfje1O_VQ5o:bawjur5) ([raw file](https://github.com/google/nomulus/blob/695fb322d4630eae55fa282498675c6afdb07ce4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L52)):*

It wasn't showing the diff across all commits, despite reloading.  Closing and reopening the tab did the trick though.  I don't know why it keeps somehow only showing some of the commits' diffs rather than all of them; I only ever want the latter.  Using a commit --amend and git push -f workflow is one way to completely avoid this problem.  (I've done that for a lot of my PRs.)  They all get squashed upon merge anyway.
</blockquote></details>

I've found myself using workflow where I add separate commits and merge with master because I was under the impression that it's often better to use this style so that reviewers can easily see <just> the changes that you made in response to their comments in that particular commit, e.g. https://github.com/google/nomulus/pull/887/commits/b0ee5ae733332990355027e90656a5ceb4670e3a 

I've asked the group if this is the case -- if not, then yeah amending and rebasing is totally fine.

PR #888, opened 2020-11-25
Script to rolling-start Nomulus

weiminyu:
Add a script to restart Nomulus non-disruptively. This can be used after
a configuration change to external resources (e.g.,  Cloud SQL
credential) to make Nomulus pick up the latest config.

Also added proper support to paging based List api methods, replacing the
current hack that forces the server to return everything in one response.
The List method for instances has a lower limit on page size than others
which is not sufficient for our project.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/888)

### DIFF
--- a/release/rollback/appengine.py
+++ b/release/rollback/appengine.py
@@ -74,10 +74,9 @@ class AppEngineAdmin:
         Returns: An immutable collection of the serving versions grouped by
             service.
         """
-        services = common.list_all_pages(
-            'services',
-            lambda page_token: self._services.list(appsId=self._project,
-                                                   pageToken=page_token))
+        services = common.list_all_pages(self._services.list,
+                                         'services',
+                                         appsId=self._project)
 
         # Response format is specified at
         # http://googleapis.github.io/google-api-python-client/docs/dyn/appengine_v1beta.apps.services.html#list.
@@ -114,18 +113,14 @@ class AppEngineAdmin:
         """
         requested_services = {version.service_id for version in versions}
 
-        def request_factory(service_id: str):
-            return lambda page_token: self._versions.list(appsId=self._project,
-                                                          servicesId=
-                                                          service_id,
-                                                          pageToken=page_token)
-
         version_configs = []
         # Sort the requested services for ease of testing. For now the mocked
         # AppEngine admin in appengine_test can only respond in a fixed order.
         for service_id in sorted(requested_services):
-            response = common.list_all_pages('versions',
-                                             request_factory(service_id))
+            response = common.list_all_pages(self._versions.list,
+                                             'versions',
+                                             appsId=self._project,
+                                             servicesId=service_id)
 
             # Format of version_list is defined at
             # https://googleapis.github.io/google-api-python-client/docs/dyn/appengine_v1beta.apps.services.versions.html#list.
@@ -156,12 +151,11 @@ class AppEngineAdmin:
     def list_instances(
             self,
             version: common.VersionKey) -> Tuple[common.VmInstanceInfo, ...]:
-        instances = common.list_all_pages(
-            'instances', lambda page_token: self._versions.instances().list(
-                appsId=self._project,
-                servicesId=version.service_id,
-                versionsId=version.version_id,
-                pageToken=page_token))
+        instances = common.list_all_pages(self._versions.instances().list,
+                                          'instances',
+                                          appsId=self._project,
+                                          servicesId=version.service_id,
+                                          versionsId=version.version_id)
 
         # Format of version_list is defined at
         # https://googleapis.github.io/google-api-python-client/docs/dyn/appengine_v1beta.apps.services.versions.instances.html#list
--- a/release/rollback/appengine.py
+++ b/release/rollback/appengine.py
@@ -14,37 +14,18 @@
 """Helper for using the AppEngine Admin REST API."""
 
 import time
-from typing import Any, Dict, FrozenSet, Set
+from typing import FrozenSet, Optional, Set, Tuple
 
 from googleapiclient import discovery
-from googleapiclient import http
 
 import common
 
 # AppEngine services under management.
 SERVICES = frozenset(['backend', 'default', 'pubapi', 'tools'])
-# Forces 'list' calls (for services and versions) to return all
-# results in one shot, to avoid having to handle pagination. This values
-# should be greater than the maximum allowed services and versions in any
-# project (
-# https://cloud.google.com/appengine/docs/standard/python/an-overview-of-app-engine#limits).
-_PAGE_SIZE = 250
 # Number of times to check the status of an operation before timing out.
 _STATUS_CHECK_TIMES = 5
 # Delay between status checks of a long-running operation, in seconds
 _STATUS_CHECK_INTERVAL = 5
-
-
-class PagingError(Exception):
-    """Error for unexpected partial results.
-
-    List calls in this module do not handle pagination. This error is raised
-    when a partial result is received.
-    """
-    def __init__(self, uri: str):
-        super().__init__(
-            self, f'Received paged response unexpectedly when calling {uri}. '
-            'Consider increasing _PAGE_SIZE.')
 
 
 class AppEngineAdmin:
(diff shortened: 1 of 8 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/888)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @mindhog and @weiminyu)

___
*[release/rollback/appengine.py, line 121 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNOIzlW0ITaCpSyqbvx:-MNOIzlX-ycWAPsI_Lva:b7ohtj1) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/appengine.py#L121)):*
> ```Python
>                                                           servicesId=
>                                                           service_id,
>                                                           pageToken=page_token)
> ```

Optional: An alternate approach to wrapping the call in a lambda would be to pass the function objects in to a modified version of your "list_all_pages" function.  So the the function would be implemented like (omitting the type annotations and most of the implementation):

    def list_all_pages(func,  *args, **kwargs):
      ...
      result = func(*args, pageToken=page_token, **kwargs)
      ...

At which point a call-site would be:

    common.list_all_pages(self._versions.list, appsId=self._project, servicesId=service_id)

The advantage to this is that you wouldn't need to define all of these lambdas.  The disadvantage is that you'd lose type-checking, though I'm not sure you're getting very much from it as I think that the underlying cloud api functions are all dynamic and therefore not type-checked anyway.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/888)* status: all files reviewed, 4 unresolved discussions (waiting on @weiminyu)

___
*[release/rollback/common.py, line 130 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNOYPzr-El80l2FYXUV:-MNOYPzr-El80l2FYXUW:b9ver5y) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/common.py#L130)):*
> ```Python
> # yapf: disable
> ListRequestFactory = NewType(
>     'ListRequestFactory', Callable[[str], http.HttpRequest])
> ```

I don't think the NewType is what you want.  This creates ListRequestFactory which is derived from Callable[...] which would require that the functions you pass into list_all_pages be derived from ListRequestFactory, when in fact they are derived from Callable[...].  And, in fact, this does show up as an error under "mypy".   Doing "ListRequestFactory = Callable[...]" should be just fine.
___
*[release/rollback/common.py, line 155 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNO_wZ195G2MmFLhodF:-MNO_wZ26fPqBW_Ua1ke:b-ad9ky7) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/common.py#L155)):*
> ```Python
>         result_collector.extend(response.get(data_field, []))
>         page_token = response.get('nextPageToken')
>         if page_token:
> ```

I think that

    if not page_token:
       return tuple(...)

would be clearer and terser.
___
*[release/rollback/rolling_restart.py, line 91 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNOeA9n8xRz5cbeUsxy:-MNOeA9o4tQ67z3ch2Yc:b-u2lnm) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/rolling_restart.py#L91)):*
> ```Python
>             if cmd.instance_name in running_instances:
>                 print('Waiting for VM to shut down...')
>                 continue
> ```

I recall at least one outage where GAE wasn't restarting instances.  It seems like it would be a pretty simple addition to just also verify that the number of instances is what we started with, no?

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/888)* status: 2 of 8 files reviewed, 4 unresolved discussions (waiting on @mindhog)

___
*[release/rollback/appengine.py, line 121 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNOIzlW0ITaCpSyqbvx:-MNQm_bvCqanqU6GtHt7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/appengine.py#L121)):*

Optional: An alternate approach to wrapping the call in a lambda would be to pass the function objects in to a modified version of your "list_all_pages" function.  So the the function would be implemented like (omitting the type annotations and most of the implementation):

    def list_all_pages(func,  *args, **kwargs):
      ...
      result = func(*args, pageToken=page_token, **kwargs)
      ...

At which point a call-site would be:

    common.list_all_pages(self._versions.list, appsId=self._project, servicesId=service_id)

The advantage to this is that you wouldn't need to define all of these lambdas.  The disadvantage is that you'd lose type-checking, though I'm not sure you're getting very much from it as I think that the underlying cloud api functions are all dynamic and therefore not type-checked anyway.
</blockquote></details>

Done.
___
*[release/rollback/common.py, line 130 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNOYPzr-El80l2FYXUV:-MNQVGWe5elnclN3Z69l:b3cfmo6) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/common.py#L130)):*

I don't think the NewType is what you want.  This creates ListRequestFactory which is derived from Callable[...] which would require that the functions you pass into list_all_pages be derived from ListRequestFactory, when in fact they are derived from Callable[...].  And, in fact, this does show up as an error under "mypy".   Doing "ListRequestFactory = Callable[...]" should be just fine.
</blockquote></details>

Realized that the named type is redundant. Dropped it and using Callable directly on the parameter.
___
*[release/rollback/common.py, line 155 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNO_wZ195G2MmFLhodF:-MNQQxuD83_oxntTDQDK:bmfxpif) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/common.py#L155)):*

I think that

    if not page_token:
       return tuple(...)

would be clearer and terser.
</blockquote></details>

Done.
___
*[release/rollback/rolling_restart.py, line 91 at r1](https://reviewable.io/reviews/google/nomulus/888#-MNOeA9n8xRz5cbeUsxy:-MNPj5Q6CWqUlBSmTqX7:b-gnax39) ([raw file](https://github.com/google/nomulus/blob/274aec8efa216818928f678a79c524bee728657a/release/rollback/rolling_restart.py#L91)):*

I recall at least one outage where GAE wasn't restarting instances.  It seems like it would be a pretty simple addition to just also verify that the number of instances is what we started with, no? 
</blockquote></details>

Added the check for manual scaling versions. Basic and auto scaling are load driven and do not always replace deleted instances.

mindhog (approved):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/888)* status: all files reviewed, 3 unresolved discussions (waiting on @mindhog and @weiminyu)

___
*[release/rollback/common.py, line 132 at r2](https://reviewable.io/reviews/google/nomulus/888#-MNTL6OF7oCthpWqoh9-:-MNTL6OF7oCthpWqoh90:bo5nzpd) ([raw file](https://github.com/google/nomulus/blob/e9733a894e6c6ad261200d3f05fabfb15e1798a5/release/rollback/common.py#L132)):*
> ```Python
>             items to be returned. This is guaranteed to be an Iterable
>             type.
> ```

We would typically document *args and **kwargs as well, in this case with something like:

    *args: Positional arguments passed to "func"

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/888)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @mindhog)

___
*[release/rollback/common.py, line 132 at r2](https://reviewable.io/reviews/google/nomulus/888#-MNTL6OF7oCthpWqoh9-:-MNTrSD7EfNCYIg40Xwc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e9733a894e6c6ad261200d3f05fabfb15e1798a5/release/rollback/common.py#L132)):*

We would typically document *args and **kwargs as well, in this case with something like:

    *args: Positional arguments passed to "func"
</blockquote></details>

Done.

PR #889, opened 2020-11-25
Add an action to resave all Domain entities

hstonec:
We added a new field `gracePeriodId` to `GracePeriod` class but it cannot get assigned until the Domain entity is loaded. So, using this action can load all existing Domain entities and resave them to set `gracePeriodId`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/889)

### DIFF
--- a/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java
+++ b/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java
@@ -1,0 +1,89 @@
+// Copyright 2020 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.batch;
+
+import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+
+import com.google.appengine.tools.mapreduce.Mapper;
+import com.google.common.collect.ImmutableList;
+import com.googlecode.objectify.Key;
+import google.registry.mapreduce.MapreduceRunner;
+import google.registry.mapreduce.inputs.EppResourceInputs;
+import google.registry.model.domain.DomainBase;
+import google.registry.model.domain.GracePeriod;
+import google.registry.request.Action;
+import google.registry.request.Response;
+import google.registry.request.auth.Auth;
+import javax.inject.Inject;
+
+/**
+ * A mapreduce that re-saves all Domain entities so all {@link GracePeriod#gracePeriodId}.
+ *
+ * <p>We added the {@link GracePeriod#gracePeriodId} to help migrate Domain entity to Cloud SQL. As
+ * it is a newly added field, all existing {@link DomainBase#gracePeriods} won't have it until its
+ * domain gets loaded. Ths mapreduce job loads all Domain entities and resave them so the {@link
+ * GracePeriod#gracePeriodId} can get assigned correctly.
+ */
+@Action(
+    service = Action.Service.BACKEND,
+    path = "/_dr/task/resaveAllDomains",
+    auth = Auth.AUTH_INTERNAL_OR_ADMIN)
+public class ResaveAllDomainsAction implements Runnable {
+
+  @Inject MapreduceRunner mrRunner;
+  @Inject Response response;
+
+  @Inject
+  ResaveAllDomainsAction() {}
+
+  @Override
+  public void run() {
+    mrRunner
+        .setJobName("Re-save all Domain entities")
+        .setModuleName("backend")
+        .runMapOnly(
+            new ResaveAllDomainsActionMapper(),
+            ImmutableList.of(EppResourceInputs.createKeyInput(DomainBase.class)))
+        .sendLinkToMapreduceConsole(response);
+  }
+
+  /** Mapper to re-save all Domain entities. */
+  public static class ResaveAllDomainsActionMapper extends Mapper<Key<DomainBase>, Void, Void> {
+
+    private static final long serialVersionUID = 1L;
+
+    public ResaveAllDomainsActionMapper() {}
+
+    @Override
+    public final void map(final Key<DomainBase> domainKey) {
+      tm().transact(
+              () -> {
+                DomainBase domain = ofy().load().key(domainKey).now();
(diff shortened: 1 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/889)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/889#-MN0b4TU1XLWBIKoUac9:-MN0b4TV9FBNEBwh8ynT:b-jbx045) ([raw file](https://github.com/google/nomulus/blob/7d10c88104970b4dae33ed95087013eaaba20d89/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java#L44)):*
> ```Java
>     path = "/_dr/task/resaveAllDomains",
>     auth = Auth.AUTH_INTERNAL_OR_ADMIN)
> public class ResaveAllDomainsAction implements Runnable {
> ```

Why is this new action necessary when `ResaveAllEppResourcesAction` already exists?

It seems to me we could just run that, and then use BQ the next day to verify that everything is set?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/889)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/889#-MN0b4TU1XLWBIKoUac9:-MN0kVTrDhYEpW_lLUtx:bo0bowm) ([raw file](https://github.com/google/nomulus/blob/7d10c88104970b4dae33ed95087013eaaba20d89/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java#L44)):*

Why is this new action necessary when `ResaveAllEppResourcesAction` already exists?

It seems to me we could just run that, and then use BQ the next day to verify that everything is set?
</blockquote></details>

Because that will resave all entities but we only need to do it for domains.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/889)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/889#-MN0b4TU1XLWBIKoUac9:-MN5EG6yEdwpTsI7bzty:b-a3hv52) ([raw file](https://github.com/google/nomulus/blob/7d10c88104970b4dae33ed95087013eaaba20d89/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java#L44)):*

Because that will resave all entities but we only need to do it for domains.
</blockquote></details>

So?  Re-save more entities then (we already run it at least monthly anyway).  It's certainly not worth writing/reviewing all this extra code just to avoid doing some extra database re-saves.  Just run what we already have.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/889)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/889#-MN0b4TU1XLWBIKoUac9:-MN5FOY68k5uziikoBgr:blqf1m3) ([raw file](https://github.com/google/nomulus/blob/7d10c88104970b4dae33ed95087013eaaba20d89/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java#L44)):*

So?  Re-save more entities then (we already run it at least monthly anyway).  It's certainly not worth writing/reviewing all this extra code just to avoid doing some extra database re-saves.  Just run what we already have.
</blockquote></details>

The advantage of using something we already have is that we know it's good and working because it's been running successfully for years.  This is new untested code.  We should always opt for what we already have if possible.

For example, there's one fairly significant error here, which is that it's not calling `cloneProjectedAtTime` on the EPP resources before re-saving them.  `ResaveAllEppResources` does that.  There may be more, but it's not worth looking into in depth; just use what we already have.

Another option you could have gone with is adding a URL parameter to be able to filter the EPP resources that `ResaveAllEppResources` runs on (with it continuing to default to all 3).  I still don't think that's necessary, but that would be a better option than writing an entirely new mapreduce if the only issue is that the existing one runs over too much data.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/889)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/889#-MN0b4TU1XLWBIKoUac9:-MN5Ioe5CSsoU9iLWrf4:b-bf99ez) ([raw file](https://github.com/google/nomulus/blob/7d10c88104970b4dae33ed95087013eaaba20d89/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java#L44)):*

The advantage of using something we already have is that we know it's good and working because it's been running successfully for years.  This is new untested code.  We should always opt for what we already have if possible.

For example, there's one fairly significant error here, which is that it's not calling `cloneProjectedAtTime` on the EPP resources before re-saving them.  `ResaveAllEppResources` does that.  There may be more, but it's not worth looking into in depth; just use what we already have.

Another option you could have gone with is adding a URL parameter to be able to filter the EPP resources that `ResaveAllEppResources` runs on (with it continuing to default to all 3).  I still don't think that's necessary, but that would be a better option than writing an entirely new mapreduce if the only issue is that the existing one runs over too much data.
</blockquote></details>

Oh, and another advantage of that approach (if we even think it's necessary -- I'm still not sold) is that if we have some kind of migration to do on contacts or hosts, then we get it for free, rather than having to write entirely new actions for them as well.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/889)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/889#-MN0b4TU1XLWBIKoUac9:-MNOuO0G4OFvjZ_Mhfmq:b-hhibt9) ([raw file](https://github.com/google/nomulus/blob/7d10c88104970b4dae33ed95087013eaaba20d89/core/src/main/java/google/registry/batch/ResaveAllDomainsAction.java#L44)):*

Oh, and another advantage of that approach (if we even think it's necessary -- I'm still not sold) is that if we have some kind of migration to do on contacts or hosts, then we get it for free, rather than having to write entirely new actions for them as well.
</blockquote></details>

Makes sense, I just verified that we run `resaveAllEppResources` monthly, so I think we can just wait until 12/2 or 12/3 to verify all GracePeriod in DomainBase have a valid id. Will close this PR.

PR #890, opened 2020-11-30
Make Domain -> BillingEvent FK deferred

mindhog:
It appears that Hibernate can sporadically introduce FK constraint failures
when updating a Domain to reference a new BillingEvent and then deleting the
old BillingEvent, causing a flakey test failure in DomainDeleteFlowTest.  This
may be due to the fact that this FK relationships is not known to hibernate.

An alternate solution appears to be to flush after every update, but that
likely has some pretty serious performance implications.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/890)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/890)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #892, opened 2020-12-01
Parameterize the serialization of objects being written to SQL

gbrodman:
We shouldn't require that objects written to SQL during a Beam pipeline
be VersionedEntity objects -- they may be non-Objectify entities. As a
result, we should allow the user to specify what the objects are that
should be written to SQL.

Note: we will need to clean up the Spec11PipelineTest more but that can
be out of the scope of this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/892)

### DIFF
--- a/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
+++ b/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
@@ -15,6 +15,7 @@
 package google.registry.beam.initsql;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.model.ofy.ObjectifyService.ofy;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
@@ -47,6 +48,7 @@ import org.apache.beam.sdk.transforms.Wait;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionTuple;
 import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.sdk.values.TypeDescriptor;
 import org.joda.time.DateTime;
 
 /**
@@ -226,7 +228,9 @@ public class InitSqlPipeline implements Serializable {
             transformId,
             options.getMaxConcurrentSqlWriters(),
             options.getSqlWriteBatchSize(),
-            new JpaSupplierFactory(credentialFileUrl, options.getCloudKmsProjectId(), jpaGetter)));
+            new JpaSupplierFactory(credentialFileUrl, options.getCloudKmsProjectId(), jpaGetter),
+            (entity) -> ofy().toPojo(entity.getEntity().get()),
+            TypeDescriptor.of(VersionedEntity.class)));
   }
 
   private static ImmutableList<String> toKindStrings(Collection<Class<?>> entityClasses) {
--- a/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
+++ b/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
@@ -15,7 +15,6 @@
 package google.registry.beam.initsql;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
@@ -48,7 +47,6 @@ import org.apache.beam.sdk.transforms.Wait;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionTuple;
 import org.apache.beam.sdk.values.TupleTag;
-import org.apache.beam.sdk.values.TypeDescriptor;
 import org.joda.time.DateTime;
 
 /**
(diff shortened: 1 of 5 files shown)


gbrodman:
Weimin, thoughts on this approach? The primary point of this is to be able to write any type of entity to SQL with this transform, rather than specifically the VersionedEntity objects. The Spec11Pipeline has an example of how this could be used in a non-VersionedEntity context.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/892)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @gbrodman and @weiminyu)

___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 281 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO35rayEPQnJNgj4eQZ:-MO35rayEPQnJNgj4eQ_:b-vf6ceo) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L281)):*
> ```
> <T>
> ```
Let's make it <T extends Serializable> for now, so that we can get rid of the objectDescriptor. 
___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 286 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO36Lsq-Dp-OXNWv0s8:-MO36Lsq-Dp-OXNWv0s9:bnkx567) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L286)):*
> ```
> SerializableFunction<T, Object> jpaConverter
> ```
For a general purpose transform, this should be optional. Alternatively , we can overload the method without this param.

Please update the javadoc.
___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 287 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO36bXf7qAGUYmiiyvM:-MO36bXf7qAGUYmiiyvN:b9kfrkx) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L287)):*
> ```
> TypeDescriptor<T> objectDescriptor
> ```
We don't need this if T extends serializable.
___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO36hJP2jcWq9NTwEjl:-MO36hJP2jcWq9NTwEjm:b-gbb0yq) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L294)):*
> ```
> objectDescriptor
> ```
Serializable.class

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/892)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 281 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO35rayEPQnJNgj4eQZ:-MO7Hngm94_lByA8GgfN:bf3ms5z) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L281)):*

> ```
> <T>
> ```
Let's make it <T extends Serializable> for now, so that we can get rid of the objectDescriptor. 
</blockquote></details>

(i'm assuming you meant `Serializable` here, as in the other comments)

Unfortunately we cannot. If we use `TypeDescriptor.of(Serializable.class)` here, then that means we lose any possible references to the specific type T for the rest of the method. If we do that, then we would just remove the generic-ization all together and we would just use `Serializable` everywhere, not `T`

This means we would lose a <ton> of flexibility when calling this, since the converter that takes the object from VersionedEntity (for instance) to Object (e.g. `(versionedEntity) -> ofy().toPojo(versionedEntity.getEntity().get())`) requires that we know the type of the object being converted (and presumably any other converters would also require that information). So we kind of need the <T>. 
___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 286 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO36Lsq-Dp-OXNWv0s8:-MO7KuyLC04ZIetz_Vmq:b-a25ywj) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L286)):*

> ```
> SerializableFunction<T, Object> jpaConverter
> ```
For a general purpose transform, this should be optional. Alternatively , we can overload the method without this param.

Please update the javadoc.
</blockquote></details>

Yep, that sounds good 
___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 287 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO36bXf7qAGUYmiiyvM:-MO7Qpq6C6VXNJ-VSXTH:bogthd6) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L287)):*

> ```
> TypeDescriptor<T> objectDescriptor
> ```
We don't need this if T extends serializable.
</blockquote></details>

(see other comment)
___
*[core/src/main/java/google/registry/beam/initsql/Transforms.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/892#-MO36hJP2jcWq9NTwEjl:-MO7QqkICPcrDB7btGsd:bogthd6) ([raw file](https://github.com/google/nomulus/blob/c5c8f315d825b582755bd7d71d51c6e6c983c5ad/core/src/main/java/google/registry/beam/initsql/Transforms.java#L294)):*

> ```
> objectDescriptor
> ```
Serializable.class
</blockquote></details>

(see other comment)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/892)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

PR #894, opened 2020-12-02
Convert DatastoreEntity/SqlEntity to return Optional, not List

gbrodman:
We don't have any entities that convert to more than one entity, so we
can use an Optional instead for clarity and simplicity.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/894)


mindhog (approved):
> Reviewed 19 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/894)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #895, opened 2020-12-02
Add SetSqlReplayCheckpoint command for SQL replay

gbrodman:
We should set this to the same time that we initially populate the SQL
database from Datastore.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/895)

### DIFF
--- a/core/src/main/java/google/registry/tools/RegistryTool.java
+++ b/core/src/main/java/google/registry/tools/RegistryTool.java
@@ -108,6 +108,7 @@ public final class RegistryTool {
           .put("save_sql_credential", SaveSqlCredentialCommand.class)
           .put("send_escrow_report_to_icann", SendEscrowReportToIcannCommand.class)
           .put("set_num_instances", SetNumInstancesCommand.class)
+          .put("set_sql_replay_checkpoint", SetSqlReplayCheckpointCommand.class)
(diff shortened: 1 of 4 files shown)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/895#-:-MOIuuUu2IBXyfnsX-jp:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/tools/SetSqlReplayCheckpointCommand.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/895#-MOIup86BS3N0aRjHsVt:-MOIup86BS3N0aRjHsVu:bj5p021) ([raw file](https://github.com/google/nomulus/blob/bcee463afef9fe179ff8825f715c3b2b2150069d/core/src/main/java/google/registry/tools/SetSqlReplayCheckpointCommand.java#L37)):*
> ```Java
>   @Override
>   protected String prompt() {
>     checkArgument(mainParameters.size() == 1, "Must provide exactly one DateTime to set");
> ```

Ugh, I can't believe there isn't a better way to do this, but there really isn't.

PR #897, opened 2020-12-07
Add a scrap command to backfill Spec11 threats

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/897)


sarahcaseybot (changes requested):
> Reviewed 9 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 9 of 11 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz2tTN3iSP3s3Kr2Oj:-MNz2tTN3iSP3s3Kr2Ok:b-bvm7g9) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L63)):*
> ```Java
>   protected String prompt() {
>     return String.format(
>         "We will attempt to backfill Spec11 results from %d files.", getDatesToBackfill().size());
> ```

This prompt seems a little awkward. Isn't it usually a question? "Attempt to backfill Spec11 results from %d files?"
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 106 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz4j-h4EQnb_XQwjvu:-MNz4j-i0GWXfj9XRFth:byvbbyp) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L106)):*
> ```Java
> 
>   /** Converts the previous {@link RegistrarThreatMatches} objects to {@link Spec11ThreatMatch}. */
>   private ImmutableList<Spec11ThreatMatch> registrarThreatMatchesToNewObjects(
> ```

Would a better name for this be registrarThreatMatchesToCloudSqlObjects?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[db/src/main/resources/sql/flyway/V81__drop_spec11_fkeys.sql, line 16 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz8Vfm16oAmPAsCgp7:-MNz8Vfm16oAmPAsCgp8:b-ohbgwe) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/db/src/main/resources/sql/flyway/V81__drop_spec11_fkeys.sql#L16)):*
> ```TSQL
> 
> -- Some objects haven't been fully populated in SQL yet, so don't depend on them.
> ALTER TABLE "Spec11ThreatMatch" DROP CONSTRAINT "fk_spec11_threat_match_domain_repo_id";
> ```

TODO to re-add these somewhere?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz2tTN3iSP3s3Kr2Oj:-MNzRVpK7twjvAI6ghZW:b3w0eno) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L63)):*

This prompt seems a little awkward. Isn't it usually a question? "Attempt to backfill Spec11 results from %d files?"
</blockquote></details>

Yeah, that sounds good 
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 106 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz4j-h4EQnb_XQwjvu:-MNzRnGqAng3JCMDUXHl:b-w6u10q) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L106)):*
> registrarThreatMatchesToCloudSqlObjects
Sounds good. Such a long name :) 
___
*[db/src/main/resources/sql/flyway/V81__drop_spec11_fkeys.sql, line 16 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz8Vfm16oAmPAsCgp7:-MNzUpEWElDjhoFuuzCQ:bplaolg) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/db/src/main/resources/sql/flyway/V81__drop_spec11_fkeys.sql#L16)):*

TODO to re-add these somewhere?
</blockquote></details>

Filed b/175059784

sarahcaseybot (approved):
> Reviewed 2 of 11 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 10 of 11 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 71 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2po-T5ESiEpNmMzJs:-MO2po-T5ESiEpNmMzJt:b-902t6c) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L71)):*
> ```Java
>     try (InputStream in = gcsUtils.openInputStream(spec11ReportFilename);
>         InputStreamReader isr = new InputStreamReader(in, UTF_8)) {
>     ImmutableList<String> reportLines =
> ```

Formatting here is off?
___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 74 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2pttU8Co-Sm-dJqnt:-MO2pttU8Co-Sm-dJqnu:bvtqpcr) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L74)):*
> ```Java
>           ImmutableList.copyOf(CharStreams.toString(isr).split("\n"));
>       // Iterate from 1 to size() to skip the header at line 0.
>       for (int i = 1; i < reportLines.size(); i++) {
> ```

You can do this easily with a Java Stream.  `reportLines.stream().skip(1).map(...)...`
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz2tTN3iSP3s3Kr2Oj:-MO2qBetEY4zkAi0Q56_:b-dbcj9n) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L63)):*

Yeah, that sounds good 
</blockquote></details>

I would get rid of the "Attempt to" part of it; that's implied.  "Backfill Spec11 results .... ?"
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2qOdkDyDex-nLgwl5:-MO2qOdkDyDex-nLgwl6:bqjel2l) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L124)):*
> ```Java
> 
>   private String getDomainRepoId(String domainName, DateTime now) {
>     return loadByForeignKeyCached(DomainBase.class, domainName, now)
> ```

This is going to individually load up each domain, which will be inefficient.  Consider something like a `loadByForeignKeys` on batches of ~20 domains instead (note that something like this exists as `loadByForeignKeysCached` as a private method on another class).

How much are we operating over anyway?  How much total data is there and how many total domains are there to load?  We're backfilling years, right?  This could actually make a meaningful difference.
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2s4oq932Fwo6Ylyys:-MO2s4oq932Fwo6Ylyyt:b-4tuq5g) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L124)):*
> ```Java
> 
>   private String getDomainRepoId(String domainName, DateTime now) {
>     return loadByForeignKeyCached(DomainBase.class, domainName, now)
> ```

Using the cached version doesn't seem like the right call here.  See the following warning text on what you're calling:

```
<p>Do not call this cached version for anything that needs transactional consistency. It should
   * only be used when it's OK if the data is potentially being out of date, e.g. WHOIS.
```
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2s8z52PvMCE7lvEup:-MO2s8z52PvMCE7lvEuq:b-uynv5r) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L124)):*
> ```Java
> 
>   private String getDomainRepoId(String domainName, DateTime now) {
>     return loadByForeignKeyCached(DomainBase.class, domainName, now)
> ```

I worry that this `now` might be hugely problematic.  How far back is this backfilling going on?  If you're backfilling data from 2017 and that domain was deleted in 2018 but then re-created in 2019, aren't you now using entirely the wrong data?  `loadByForeignKey` is only suitable for a live view of currently active domains, which I'm not sure is what you're actually doing here.  This needs a rethink.  I don't think it's valid to call `loadByForeignKey` with a date in the past either; we only project forwards in time, never backwards.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 10 of 11 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 74 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2pttU8Co-Sm-dJqnt:-MO2tHf26xHDgmOod8FA:bdrl25v) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L74)):*

You can do this easily with a Java Stream.  `reportLines.stream().skip(1).map(...)...`
</blockquote></details>

Ahhh, I see you didn't actually write this because you moved the function around for some reason (grouping it with the other publics?) and I misinterpreted the diff.  Still, so long as you're making changes ... we can fix this :)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 71 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2po-T5ESiEpNmMzJs:-MO2u9VvC0puu_B6bpsg:b45teym) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L71)):*

Formatting here is off?
</blockquote></details>

Very odd, how did that not get caught? 
___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 74 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2pttU8Co-Sm-dJqnt:-MO2uBwg83IPEJ069ALq:bn6s8c) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L74)):*

Ahhh, I see you didn't actually write this because you moved the function around for some reason (grouping it with the other publics?) and I misinterpreted the diff.  Still, so long as you're making changes ... we can fix this :) 
</blockquote></details>

yeah, moved it to group it with the other publics (and to add the return-empty-if-file-doesn't-exist logic). 

Stream-skip is definitely better. 
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/897#-MNz2tTN3iSP3s3Kr2Oj:-MO2uj_ZC2G9ICqN3C6E:bs7wumx) ([raw file](https://github.com/google/nomulus/blob/ffc0f197358653bfab9b4825016cbb5ff7369317/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L63)):*

I would get rid of the "Attempt to" part of it; that's implied.  "Backfill Spec11 results .... ?"
</blockquote></details>

Done. 
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2s8z52PvMCE7lvEup:-MO2sx9r2ACWn-uNFAWo:b-oj3tnv) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L124)):*

I worry that this `now` might be hugely problematic.  How far back is this backfilling going on?  If you're backfilling data from 2017 and that domain was deleted in 2018 but then re-created in 2019, aren't you now using entirely the wrong data?  `loadByForeignKey` is only suitable for a live view of currently active domains, which I'm not sure is what you're actually doing here.  This needs a rethink.  I don't think it's valid to call `loadByForeignKey` with a date in the past either; we only project forwards in time, never backwards.
</blockquote></details>

Ah, we can't project backwards in time? That is problematic. Yeah, let me think about it. 

See other comments on the start date. 
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2s4oq932Fwo6Ylyys:-MO2t31Z5-5YxRIsNnND:b-9okfgp) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L124)):*

Using the cached version doesn't seem like the right call here.  See the following warning text on what you're calling:

```
<p>Do not call this cached version for anything that needs transactional consistency. It should
   * only be used when it's OK if the data is potentially being out of date, e.g. WHOIS.
```
</blockquote></details>

I'm not sure this matters -- we don't need transactional consistency, this is just (supposed to be) a best-effort "what was this domain's repo ID on date X?" 
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/897#-MO2qOdkDyDex-nLgwl5:-MO2tBd36Z1D6E5IVjbW:bgyaarq) ([raw file](https://github.com/google/nomulus/blob/1ffc35616f530eab5a02980e0491815ad906e9f1/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L124)):*

This is going to individually load up each domain, which will be inefficient.  Consider something like a `loadByForeignKeys` on batches of ~20 domains instead (note that something like this exists as `loadByForeignKeysCached` as a private method on another class).

How much are we operating over anyway?  How much total data is there and how many total domains are there to load?  We're backfilling years, right?  This could actually make a meaningful difference.
</blockquote></details>

The start date (at the top of the file) is 2019-01-01, so there are roughly 730 days (two years, give or take) of data. Each day has maybe 200-300 domains on average, so maybe around 150k total entries. We may need to batch this in some way.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 75 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3Hk7v7esb3sIdd2Uv:-MO3Hk7v7esb3sIdd2Uw:bnhbtzw) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L75)):*
> ```Java
>           ImmutableList.copyOf(CharStreams.toString(isr).split("\n"));
>       // Skip the header at line 0
>       return reportLines.stream()
> ```

I think you can just delete the previous line and then have this be something like

```
  CharStreams.toString(isr).split("\n").stream().skip(1)...
```

No need to even make an immutable list copy when you can just work off a stream of the Strings directly.  Saves memory too.
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 66 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3IL41AVqrDBDRza3N:-MO3IL41AVqrDBDRza3O:b-l5czjn) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L66)):*
> ```Java
> 
>   @Override
>   protected String execute() {
> ```

What does resume behavior look like with this command?  If it runs part way then fails with a fatal error for some reason (could just be as simple as a network flake), what then?  When you run it again does it do all that work over?  Restart from where it left off somehow?  Given how much data we're expecting to process over we expect this to run for a minimum of hours, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 75 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3Hk7v7esb3sIdd2Uv:-MO3J_7uA1oYBK_vlr7Y:b-iwrcz4) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L75)):*

I think you can just delete the previous line and then have this be something like

```
  CharStreams.toString(isr).split("\n").stream().skip(1)...
```

No need to even make an immutable list copy when you can just work off a stream of the Strings directly.  Saves memory too.
</blockquote></details>

derp of course (with Arrays.stream since `split` returns an array) 
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 66 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3IL41AVqrDBDRza3N:-MO3JCYv-K0hkmj80qX_:b-y1usw7) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L66)):*

What does resume behavior look like with this command?  If it runs part way then fails with a fatal error for some reason (could just be as simple as a network flake), what then?  When you run it again does it do all that work over?  Restart from where it left off somehow?  Given how much data we're expecting to process over we expect this to run for a minimum of hours, right?
</blockquote></details>

At the moment, it starts over (idempotently -- it deletes entries from a day before refilling it). There are going to be many rows in the SQL table (maybe upwards of 100k) but only sub-2k individual domains.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 75 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3Hk7v7esb3sIdd2Uv:-MODNBL27OK5_XHRT0YE:b-rw1vmd) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L75)):*

derp of course (with Arrays.stream since `split` returns an array) 
</blockquote></details>

We're really getting into the weeds here, but built-in `split()` kinda sucking is why we tend to prefer using Guava's `Splitter`.  E.g. something like this should work and be nicer, because everything happens in a linear procession without as much mixing up of parentheses-wrapping and fluent method calling:

```
Splitter.on("\n").splitToList(CharStreams.toString(isr)).stream()....
```

Generally, we'd prefer never to have to deal with arrays, when possible anyway.  That is possible in almost all circumstances except when dealing with Java method varargs.
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 66 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3IL41AVqrDBDRza3N:-MODORu92bpmOmdBqK5g:bvqpwiw) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L66)):*

At the moment, it starts over (idempotently -- it deletes entries from a day before refilling it). There are going to be many rows in the SQL table (maybe upwards of 100k) but only sub-2k individual domains. 
</blockquote></details>

We should probably be smarter about this then, because it seems quite likely that some kind of error will happen far into the process, yet there's no mechanism here for resuming from where we left off, or only retrying the specified dates.  Is there some way to easily detect which dates have been successfully imported and skip over those, rather than overwriting everything and starting over?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 75 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3Hk7v7esb3sIdd2Uv:-MODPisd2D4MgQ8EWOVw:b-5xq191) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L75)):*

We're really getting into the weeds here, but built-in `split()` kinda sucking is why we tend to prefer using Guava's `Splitter`.  E.g. something like this should work and be nicer, because everything happens in a linear procession without as much mixing up of parentheses-wrapping and fluent method calling:

```
Splitter.on("\n").splitToList(CharStreams.toString(isr)).stream()....
```

Generally, we'd prefer never to have to deal with arrays, when possible anyway.  That is possible in almost all circumstances except when dealing with Java method varargs.
</blockquote></details>

Sure, will do
___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 66 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3IL41AVqrDBDRza3N:-MODPbRI278RRma6INZH:b-yzcbwk) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L66)):*

We should probably be smarter about this then, because it seems quite likely that some kind of error will happen far into the process, yet there's no mechanism here for resuming from where we left off, or only retrying the specified dates.  Is there some way to easily detect which dates have been successfully imported and skip over those, rather than overwriting everything and starting over?
</blockquote></details>

We could always query SQL and skip days for which we have data (possibly with an `--overwrite` parameter to not do that). Sound good?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 66 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3IL41AVqrDBDRza3N:-MODQXiWCIfQzCEPZWex:byaor9p) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L66)):*

We could always query SQL and skip days for which we have data (possibly with an `--overwrite` parameter to not do that). Sound good? 
</blockquote></details>

or `--overwrite_existing_dates`

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 66 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3IL41AVqrDBDRza3N:-MODk1L459cx9dLbV5wB:b1b8n5) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L66)):*

or `--overwrite_existing_dates` 
</blockquote></details>

I think `--overwrite_existing_dates` seems good. PTAL?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java, line 75 at r4](https://reviewable.io/reviews/google/nomulus/897#-MO3Hk7v7esb3sIdd2Uv:-MOHgTkcDnrZHlG-tcut:b-rljovm) ([raw file](https://github.com/google/nomulus/blob/04172d7415abe0df95253c257b1b2715c34890b6/core/src/main/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParser.java#L75)):*

Sure, will do
</blockquote></details>

fun fact: `Splitter`'s default behavior is not the same  as `String::split` in that the former (by default) does not omit tailing empty strings but the latter does. There's an option for that though.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 142 at r5](https://reviewable.io/reviews/google/nomulus/897#-MOIwmaBDjwrGF3Sldce:-MOIwmaBDjwrGF3Sldcf:b-hau88g) ([raw file](https://github.com/google/nomulus/blob/07ef5d382a8b5a4c1c4efbdf41f6ab34dbe440e5/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L142)):*
> ```Java
>                           .load()
>                           .type(DomainBase.class)
>                           .filter("fullyQualifiedDomainName", domainName)
> ```

So just to confirm, each domain is only loaded once, each in an individual eventually consistent query, with expected order of many hundreds of domains?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: 8 of 11 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java, line 142 at r5](https://reviewable.io/reviews/google/nomulus/897#-MOIwmaBDjwrGF3Sldce:-MOIxNyJ2Pd-SnkZGc11:bhuzt2w) ([raw file](https://github.com/google/nomulus/blob/07ef5d382a8b5a4c1c4efbdf41f6ab34dbe440e5/core/src/main/java/google/registry/tools/javascrap/BackfillSpec11ThreatMatchesCommand.java#L142)):*

So just to confirm, each domain is only loaded once, each in an individual eventually consistent query, with expected order of many hundreds of domains?
</blockquote></details>

Yes. The `distinct()` above is filtering domain names so we only query once per domain name, then the query returns N domains, all with that domain name (N is probably very low here due to the fact that there isn't that much domain churn) 

The total number of domain names since 2019-01 was roughly 1500 domains IIRC (it was above 1000 and below 2000). So even if there are multiple domain objects per domain name we should be ok. 

I ran a test where I chose 10-20 of the malicious domains at random and ran a `select count(*) from DomainBase where fqdn =....` and most only have one DomainBase instance, a few had two, and one had three.

gbrodman:
ptal?

sarahcaseybot (approved):
> Reviewed 2 of 3 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/897)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #902, opened 2020-12-09
Add BillingVKey to restore symmetric VKey in GracePeriodBase

hstonec:
This PR introduces `BillingVKey` and its two subclasses to replace the usage of `VKey<BillingEvent>` in `GracePeriodBase`. It helps restore a symmetric VKey without adding the extra field, e.g. `billingEventOneTimeHistoryId `, in the entity.

This PR also refactored `AllocationToken.redemptionHistoryEntry` related code to not expose `DomainHistoryVKey` externally, i.e. `DomainHistoryVKey redemptionHistoryEntry` is only used inside `AllocationToken`, its setter and getter will convert it from/to `VKey<HistoryEntry>`.

(The original PR was to apply the new `PollMessageVKey` to vkey fields in `DomainContent` but we decided to not make that change because the current implementation works and the improvement of using `PollMessageVKey` is marginal and it comes with a disadvantage that it adds a few redundant `*repo_id` columns in the table) 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/902)


mindhog (changes requested):
I haven't finished reviewing all of this yet, but it seems to be adding a substantial amount of infrastructure in order to marginally simplify DomainContent.  Are you sure this is worth doing?  It looks like it could also be applied to DomainTransferData, but even so.  There's not a lot of places where we have PollMessage VKeys.

We definitely need the parts that break out PollMessage into OneTime and Autorenew keys, but right now I'm not feeling too good about the rest of it -- especially if we have to store extra copies of the domain repo id.

> Reviewed 6 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MO7rnNQDXZscnhYuT-X:bbe59ff)* status: 6 of 17 files reviewed, 1 unresolved discussion (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 255 at r1](https://reviewable.io/reviews/google/nomulus/902#-MO7hkMF3QQPDXqKBnUv:-MO7hkMG9hnqw8cGt3-5:b-n892ac) ([raw file](https://github.com/google/nomulus/blob/ac6aaea1c56c5e6a58de581f4da45cd8b35cac43/core/src/main/java/google/registry/model/domain/DomainContent.java#L255)):*
> ```Java
>     @AttributeOverride(
>         name = "repoId",
>         column = @Column(name = "autorenew_poll_message_domain_repo_id")),
> ```

Doesn't this mean we're going to be storing the domain's repo id 3 times in the domain record and the domain history record?

hstonec (commented):
I think it worths. Let's think about it this way, for each `objecitfy.Key<PollMessage>`, it has three kind of information that we need to know in order to restore the symmetric key, i.e. the repo id, the HistoryEntry id, and its own id. So, it makes sense to map it to three columns in a sql table because this is how it works. Given that, the natural way to do it in Hibernate is to make the class `@Embeddable` because `AttributeConverter` can only do 1-to-1 mapping. 

Yeah, the only downside of this approach is to have the extra column for the repo id, and I don't find a way to reuse the entity's repoId while allowing the key nullable. However, we need to add the extra HistoryEntry id for each key anyway, and it will be removed after the migration, so I think it is ok to add one more thing for similar purpose. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MO7zY2cCLgREzCm2UnF:b4k3n61)* status: 6 of 17 files reviewed, 1 unresolved discussion (waiting on @hstonec and @mindhog)

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902)* status: 6 of 17 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 255 at r1](https://reviewable.io/reviews/google/nomulus/902#-MO7hkMF3QQPDXqKBnUv:-MO81Ggd4nOAI5LoJWSs:brukcox) ([raw file](https://github.com/google/nomulus/blob/ac6aaea1c56c5e6a58de581f4da45cd8b35cac43/core/src/main/java/google/registry/model/domain/DomainContent.java#L255)):*

Doesn't this mean we're going to be storing the domain's repo id 3 times in the domain record and the domain history record?
</blockquote></details>

Yes, this is unfortunate.  As I explained above, if we reuse the entity's repo id, Hibernate will always create an instance and set the repo id even other fields, i.e. HistoryEntry id, are null, which makes an invalid key.

mindhog (changes requested):
> Reviewed 11 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902)* status: all files reviewed, 2 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/persistence/PollMessageVKey.java, line 31 at r1](https://reviewable.io/reviews/google/nomulus/902#-MOBdY221iVPWUprEZ7V:-MOBdY221iVPWUprEZ7W:b-djh3zt) ([raw file](https://github.com/google/nomulus/blob/ac6aaea1c56c5e6a58de581f4da45cd8b35cac43/core/src/main/java/google/registry/persistence/PollMessageVKey.java#L31)):*
> ```Java
> @MappedSuperclass
> public abstract class PollMessageVKey<E extends EppResource>
>     extends EppHistoryVKey<PollMessage, E> {
> ```

I realize that this code gets exercised by DomainBase and (eventually) by ContactResource, but it really deserves its own unit test.  That includes EppHistoryVKey, which does most of the work.

mindhog (commented):
I agree that this more closely maps to the prescribed approach for Hibernate, but is there any material advantage?

The material disadvantages are that it requires a lot more code and introduces redundant columns in our database.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MOCMCJx9qWzyQsp0StC:bqy1meb)* status: all files reviewed, 2 unresolved discussions (waiting on @hstonec and @mindhog)

hstonec (commented):
We still have some other places, e.g. TransferData, that we need to implement the symmetric vkey and I think this approach should be simpler. And also, we should be able to remove the `synchronized void restoreOfyKeys()` from GracePeriodBase by applying this approach.

If the redundant column is a concern, I can try to reuse the entity's repo id, then have some check in `@PostLoad` and set the field to null. Thoughts?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MOCrAip21_0ZS6wM-3g:b-6itz9j)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/PollMessageVKey.java, line 31 at r1](https://reviewable.io/reviews/google/nomulus/902#-MOBdY221iVPWUprEZ7V:-MOCjuLRAuzp7__PJS-U:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ac6aaea1c56c5e6a58de581f4da45cd8b35cac43/core/src/main/java/google/registry/persistence/PollMessageVKey.java#L31)):*

I realize that this code gets exercised by DomainBase and (eventually) by ContactResource, but it really deserves its own unit test.  That includes EppHistoryVKey, which does most of the work.
</blockquote></details>

Done.

mindhog (changes requested):
I'm still not completely convinced, but the possibility of removing those contortions from GracePeriodBase definitely makes for a stronger case.

I guess I prefer the redundant column to the specialized logic in the containing classes.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MOHFIk-E9LEIDuUx8ox:b1a4wek)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/test/java/google/registry/persistence/PollMessageVKeyTest.java, line 86 at r2](https://reviewable.io/reviews/google/nomulus/902#-MOHHNgw5c6JhWYreIwo:-MOHHNgw5c6JhWYreIwp:b-gi40ml) ([raw file](https://github.com/google/nomulus/blob/3e0cc9337227559648c07a36767aa66f3853e07c/core/src/test/java/google/registry/persistence/PollMessageVKeyTest.java#L86)):*
> ```Java
> 
>     assertThat(persisted).isEqualTo(original);
>   }
> ```

Can you also add a test for an unknown kind path?

hstonec (commented):
That's a good point, let me use this to GracePeriodBase first to see if it can remove that ad-hoc method, we can defer applying this to everywhere as long as the current implementation works well.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MOHJLfQ-NRbtRPbMIPg:bcsx3ek)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec and @mindhog)

hstonec (commented):
I made the following changes:

1. Reverted all vkey changes in `DomainContent` as we discussed before, so I also deleted the `PollMessageVKey` class as it is not used right now.

2. Added a `BillingVKey` class and two subclasses to be used in `GracePeriodBase`, and refactored `GracePeriodBase` by removing `private final synchronized void restoreOfyKeys` method and two `*HistoryId` fields because the `BillingVKey` handles them properly.

3. Refactored `AllocationToken.redemptionHistoryEntry` related code to not expose `DomainHistoryVKey` externally, i.e. `DomainHistoryVKey redemptionHistoryEntry` is only used inside `AllocationToken`, its setter and getter will convert it from/to `VKey<HistoryEntry>`.

4. Make `EppHistoryVKey` not extend `VKey` because 1) it simplifies the constructor and removes `@PostLoad` method 2) it is not necessary as we should always convert `EppHistoryVKey` to a VKey instance in the getter, and convert a VKey instance to `EppHistoryVKey` in the setter 3) we don't need to worry about that the new `EppHistoryVKeyTranslatorFactory` interferes the existing VKey serde, as they are not parent and subclass any more.  

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902#-:-MObFqYGAeGtUOv-qMWP:bhqhfjq)* status: 2 of 32 files reviewed, 3 unresolved discussions (waiting on @mindhog)

___
*[core/src/test/java/google/registry/persistence/PollMessageVKeyTest.java, line 86 at r2](https://reviewable.io/reviews/google/nomulus/902#-MOHHNgw5c6JhWYreIwo:-MObEU4fEpeuAyisZ6Kj:b437qoy) ([raw file](https://github.com/google/nomulus/blob/3e0cc9337227559648c07a36767aa66f3853e07c/core/src/test/java/google/registry/persistence/PollMessageVKeyTest.java#L86)):*

Can you also add a test for an unknown kind path?
</blockquote></details>

I thought about this but it seems unnecessary to do that because only the subclass of `EppHistoryVKey` will be added to `EppHistoryVKeyTranslatorFactory.kindPathToVKeyClass`, and if it is a subclass of `EppHistoryVKey`, it gets a valid kind path through `EppHistoryVKey.getKindPath` anyway. If we add a new subclass of `EppHistoryVKey` but forget to add it to `EppHistoryVKeyTranslatorFactory.kindPathToVKeyClass`, the Objectify's initialization process complains that it is missing a translator factory.

mindhog (changes requested):
> Reviewed 1 of 17 files at r1, 32 of 32 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902)* status: all files reviewed, 9 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 88 at r3](https://reviewable.io/reviews/google/nomulus/902#-MObyszBCpMFtrOSMbqj:-MObyszBCpMFtrOSMbqk:b2my1aj) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/model/domain/GracePeriod.java#L88)):*
> ```Java
>     instance.clientId = checkArgumentNotNull(clientId);
>     instance.billingEventOneTime =
>         billingEventOneTime == null ? null : BillingEventVKey.create(billingEventOneTime);
> ```

Can we move the null check into the create() methods for {BillingEvent,BillingRecurrence}VKey?  (I think that's what we're doing elsewhere)
___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 194 at r3](https://reviewable.io/reviews/google/nomulus/902#-MObz9xI5HxRCSNql2aj:-MObz9xI5HxRCSNql2ak:b-lwlvsv) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/model/domain/GracePeriod.java#L194)):*
> ```Java
>   public GracePeriod cloneWithRecurringBillingEvent(VKey<BillingEvent.Recurring> recurring) {
>     GracePeriod clone = clone(this);
>     clone.billingEventRecurring = BillingRecurrenceVKey.create(recurring);
> ```

If not I presume we'll want to do the null check here?
___
*[core/src/main/java/google/registry/model/translators/EppHistoryVKeyTranslatorFactory.java, line 45 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOc2Y3-Ef-i9tFGknh0:-MOc26Fa3CetfTxNSO4b:bdwwwzj) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/model/translators/EppHistoryVKeyTranslatorFactory.java#L45)):*
> ```
> final
> ```
Add at-VisibleForTesting or protected?
___
*[core/src/main/java/google/registry/persistence/BillingVKey.java, line 42 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOc3YOa-KlzsjPttT1G:-MOc3YOa-KlzsjPttT1H:b-lrpvnf) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/persistence/BillingVKey.java#L42)):*
> ```Java
>     this.repoId = repoId;
>     this.historyRevisionId = historyRevisionId;
>     this.billingId = billingId;
> ```

Is there a base class constructor that you could call instead for repoId, historyRevisionId?
___
*[core/src/main/java/google/registry/persistence/BillingVKey.java, line 82 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOc5JMp-qyorlsDpUr8:-MOc5JMp-qyorlsDpUr9:b-lkcdyw) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/persistence/BillingVKey.java#L82)):*
> ```Java
>       long billingEventId = ofyKey.getId();
>       long historyRevisionId = ofyKey.getParent().getId();
>       String repoId = ofyKey.getParent().getParent().getName();
> ```

I know you've gone to great pains to remove the Registry 1.0 keys from the system, but given that this "parent of a parent" broke on map-reduce I'm wondering if there's any better way we can deal with a malformed key?  Like maybe we want to relax the "no asymmetric VKeys" rule in this case?
___
*[core/src/test/java/google/registry/model/domain/GracePeriodTest.java, line 95 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOcBcHn5-QWdj0iFJAL:-MOcBcHoEdnPTHOz_NFf:blq36u5) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/test/java/google/registry/model/domain/GracePeriodTest.java#L95)):*
> ```Java
>     if (!gracePeriod.getRecurringBillingEvent().equals(recurringKey)) {
>       throw new RuntimeException("not equal");
>     }
> ```

Does assertThat(...).isEqualTo(...) not work here?  If not, please do "assertThat(gracePeriod...equals(recurringKey)).isFalse()".

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902)* status: 20 of 32 files reviewed, 9 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 88 at r3](https://reviewable.io/reviews/google/nomulus/902#-MObyszBCpMFtrOSMbqj:-MOgKDOuC3AItA2knFfI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/model/domain/GracePeriod.java#L88)):*

Can we move the null check into the create() methods for {BillingEvent,BillingRecurrence}VKey?  (I think that's what we're doing elsewhere)
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 194 at r3](https://reviewable.io/reviews/google/nomulus/902#-MObz9xI5HxRCSNql2aj:-MOgKREeD5oXsIN3Eec4:b6o9e06) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/model/domain/GracePeriod.java#L194)):*

If not I presume we'll want to do the null check here?
</blockquote></details>

Yeah, I added the null check inside the create function.
___
*[core/src/main/java/google/registry/model/translators/EppHistoryVKeyTranslatorFactory.java, line 45 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOc2Y3-Ef-i9tFGknh0:-MOgKewY7bpzh3b8Wsz5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/model/translators/EppHistoryVKeyTranslatorFactory.java#L45)):*

> ```
> final
> ```
Add at-VisibleForTesting or protected?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/BillingVKey.java, line 42 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOc3YOa-KlzsjPttT1G:-MOgLE7o4eovnFimp_1I:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/persistence/BillingVKey.java#L42)):*

Is there a base class constructor that you could call instead for repoId, historyRevisionId?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/BillingVKey.java, line 82 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOc5JMp-qyorlsDpUr8:-MOgRpLeEzAsV1O7Udeo:b-p49abw) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/main/java/google/registry/persistence/BillingVKey.java#L82)):*

I know you've gone to great pains to remove the Registry 1.0 keys from the system, but given that this "parent of a parent" broke on map-reduce I'm wondering if there's any better way we can deal with a malformed key?  Like maybe we want to relax the "no asymmetric VKeys" rule in this case?
</blockquote></details>

I think for `GracePeriod` it should be safe to assume that there is no registry 1.0 key and all keys have 3 tiers([query1](https://console.cloud.google.com/bigquery?sq=1080941367941:ce9a647c2c2b4fe5bbb090bccab55de4) and [query2](https://console.cloud.google.com/bigquery?sq=1080941367941:c81dc810d4404d3f95c3ce8045fad4b3) verified that, these two queries strictly checked that if there are any field that is not 3-tier)
___
*[core/src/test/java/google/registry/model/domain/GracePeriodTest.java, line 95 at r3](https://reviewable.io/reviews/google/nomulus/902#-MOcBcHn5-QWdj0iFJAL:-MOgLKUe5H_A9EZnI_Jr:bovl7o9) ([raw file](https://github.com/google/nomulus/blob/b48f4e80ac54b08e465c68b204c2c33a5d0a48ca/core/src/test/java/google/registry/model/domain/GracePeriodTest.java#L95)):*

Does assertThat(...).isEqualTo(...) not work here?  If not, please do "assertThat(gracePeriod...equals(recurringKey)).isFalse()".
</blockquote></details>

haha, this is some debug code I added to investigate the test failure. BTW, it failed because we passed `12345` instead of `12345L` to `VKey.create()` as the `sqlKey` for `recurringKey`. However, when the persisted key loaded from DB, it is always `12345L`, and the check failed because the type doesn't match.

mindhog (approved):
> Reviewed 12 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/902)* status: all files reviewed, 3 unresolved discussions (waiting on @mindhog)

PR #904, opened 2020-12-10
Use 10 workers instead of the default 100 for re-save all EPP resources

CydeWeys:
The intended/desired effect is to have a larger number of GCS commit log diffs
spread out over a longer period of time, with each diff itself being
significantly smaller. This should retain roughly the same amount of total work
for the async Cloud SQL replication action to have to deal with, but spread
across 10X as much time.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/904)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/904)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #910, opened 2020-12-17
Convert HostCreateFlow and HostCheckFlow to tm()

hstonec:
This is the first PR that tried to convert EPP flows to use `tm()` APIs in its application and test code. So, there are
also some corresponding changes in `EppResourceUtils` and `ForeignKeyIndex`.

For the changes in `ForeignKeyIndex`, as we no longer maintain the entity in Cloud SQL, we need to read the original
entity table to reconstruct the instance. This approach helps keep a single code path as `ForeignKeyIndex` are used in
many EPP flows.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/910)

### DIFF
--- a/core/src/main/java/google/registry/flows/host/HostCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/host/HostCreateFlow.java
@@ -21,10 +21,8 @@ import static google.registry.flows.host.HostFlowUtils.validateHostName;
 import static google.registry.flows.host.HostFlowUtils.verifySuperordinateDomainNotInPendingDelete;
 import static google.registry.flows.host.HostFlowUtils.verifySuperordinateDomainOwnership;
 import static google.registry.model.EppResourceUtils.createRepoId;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.CollectionUtils.isNullOrEmpty;
-import static google.registry.util.CollectionUtils.union;
 
 import com.google.common.collect.ImmutableSet;
 import com.googlecode.objectify.Key;
@@ -137,13 +135,11 @@ public final class HostCreateFlow implements TransactionalFlow {
     ImmutableSet<ImmutableObject> entitiesToSave =
         ImmutableSet.of(
             newHost,
-            historyBuilder.build(),
+            historyBuilder.build().toChildHistoryEntity(),
             ForeignKeyIndex.create(newHost, newHost.getDeletionTime()),
             EppResourceIndex.create(Key.create(newHost)));
     if (superordinateDomain.isPresent()) {
-      entitiesToSave =
-          union(
-              entitiesToSave,
+      tm().update(
               superordinateDomain
                   .get()
                   .asBuilder()
@@ -153,7 +149,7 @@ public final class HostCreateFlow implements TransactionalFlow {
       // they are only written as NS records from the referencing domain.
       dnsQueue.addHostRefreshTask(targetId);
     }
-    ofy().save().entities(entitiesToSave);
+    tm().insertAll(entitiesToSave);
     return responseBuilder.setResData(HostCreateData.create(targetId, now)).build();
   }
 
--- a/core/src/main/java/google/registry/model/EppResourceUtils.java
+++ b/core/src/main/java/google/registry/model/EppResourceUtils.java
@@ -18,6 +18,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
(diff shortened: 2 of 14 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/910)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 96 at r1](https://reviewable.io/reviews/google/nomulus/910#-MP5QdM8E6qXLMaa0Kf3:-MP5QdM90q7jI6WAnLgO:b-tvgyx8) ([raw file](https://github.com/google/nomulus/blob/59e80f0e534825efeef356245313df2b85befc69/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L96)):*
> ```Java
>           ImmutableMap.of(
>               ContactResource.class, "contactId",
>               DomainBase.class, "fullyQualifiedDomainName",
> ```

Will ForeignKeyIndex go away entirely after the complete switch to SQL? If so, cool. If not, we may run in to issues if we ever want to rename these fields in the Java classes (I believe the SQL fields are already renamed) 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 64 at r1](https://reviewable.io/reviews/google/nomulus/910#-MP5XNim2SsbkmeDpwPq:-MP5XNin1h9FYqEQslsM:b6q7qvm) ([raw file](https://github.com/google/nomulus/blob/59e80f0e534825efeef356245313df2b85befc69/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L64)):*
> ```Java
>   private static final Retrier retrier = new Retrier(new SystemSleeper(), 3);
> 
>   // The entity of classes in this set will be simply ignored when passed to modification
> ```

Do we have a backlog / doc of TODOs / bugs that should be fixed post-migration? 
___
*[core/src/test/java/google/registry/testing/AppEngineExtension.java, line 457 at r1](https://reviewable.io/reviews/google/nomulus/910#-MP5XsRFEA_0QbMXc3lX:-MP5XsRGASjJfU-hRquF:b-n98j9r) ([raw file](https://github.com/google/nomulus/blob/d7e2b24468a76a06580fa5abebbab6818817575a/core/src/test/java/google/registry/testing/AppEngineExtension.java#L457)):*
> ```Java
>     }
> 
>     if (clock != null) {
> ```

Out of curiosity, why was this removed?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/910)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 96 at r1](https://reviewable.io/reviews/google/nomulus/910#-MP5QdM8E6qXLMaa0Kf3:-MP5r9Kt3wJ5Xw6InkNf:b-gkrrmj) ([raw file](https://github.com/google/nomulus/blob/59e80f0e534825efeef356245313df2b85befc69/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L96)):*
> Will ForeignKeyIndex go away entirely after the complete switch to SQL?

Yes, it will need some effort to remove it entirely from the code base, however, it is still better than doing it right now because it is hard to make the change while we are still using Datastore.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 64 at r1](https://reviewable.io/reviews/google/nomulus/910#-MP5XNim2SsbkmeDpwPq:-MP63bSdDHOGJApy6h9P:b-896fix) ([raw file](https://github.com/google/nomulus/blob/59e80f0e534825efeef356245313df2b85befc69/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L64)):*

Do we have a backlog / doc of TODOs / bugs that should be fixed post-migration? 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/AppEngineExtension.java, line 457 at r1](https://reviewable.io/reviews/google/nomulus/910#-MP5XsRFEA_0QbMXc3lX:-MP5s5LR4lnWlQxY1LHw:b-bezxkl) ([raw file](https://github.com/google/nomulus/blob/d7e2b24468a76a06580fa5abebbab6818817575a/core/src/test/java/google/registry/testing/AppEngineExtension.java#L457)):*

Out of curiosity, why was this removed? 
</blockquote></details>

This is due to the change in `FlowTestCase`, where I added `.withClock(clock)` to AppEngineExtension builder for successfully running the test against SQL db. Though, I found that if I keep setting the clock here, the `TaskQueue` complaint something like `ETA cannot be reached` for some tests that the `clock` is set to year 1999, so I guess the `LocalServiceTestHelper` checks the time if it is set. And I also noticed that none of our test code using `TastQueue` actually requires an injected clock, so I removed it and all tests still passed and the problem is also solved.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/910)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #912, opened 2020-12-21
Add a fast mode to the ResaveAllEppResourcesAction mapreduce

CydeWeys:
This new mode avoids writing no-op mutations for entities that don't actually
have any changes to write. The cronjobs use fast mode by default, but manual
invocations do not, as manual invocations are often used to trigger @OnLoad
migrations, and fast mode won't pick up on those changes.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/912)

### DIFF
--- a/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java
+++ b/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java
@@ -14,6 +14,7 @@
 
 package google.registry.batch;
 
+import static google.registry.mapreduce.MapreduceRunner.PARAM_FAST;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 
@@ -24,6 +25,7 @@ import google.registry.mapreduce.MapreduceRunner;
 import google.registry.mapreduce.inputs.EppResourceInputs;
 import google.registry.model.EppResource;
 import google.registry.request.Action;
+import google.registry.request.Parameter;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import javax.inject.Inject;
@@ -39,6 +41,14 @@ import javax.inject.Inject;
  * <p>Because there are no auth settings in the {@link Action} annotation, this command can only be
  * run internally, or by pretending to be internal by setting the X-AppEngine-QueueName header,
  * which only admin users can do.
+ *
+ * <p>If the <code>?fast=true</code> querystring parameter is passed, then entities that are not
+ * changed by {@link EppResource#cloneProjectedAtTime} will not be re-saved. This helps prevent
+ * mutation load on the DB and has the beneficial side effect of writing out smaller commit logs.
+ * Note that this does NOT pick up mutations caused by migrations using the {@link
+ * com.googlecode.objectify.annotation.OnLoad} annotation, so if you are running a one-off schema
+ * migration, do not use fast mode. Fast mode defaults to false for this reason, but is used by the
+ * monthly invocation of the mapreduce.
  */
 @Action(
     service = Action.Service.BACKEND,
@@ -48,7 +58,13 @@ public class ResaveAllEppResourcesAction implements Runnable {
 
   @Inject MapreduceRunner mrRunner;
   @Inject Response response;
-  @Inject ResaveAllEppResourcesAction() {}
+
+  @Inject
+  @Parameter(PARAM_FAST)
+  boolean isFast;
+
+  @Inject
+  ResaveAllEppResourcesAction() {}
 
   /**
    * The number of shards to run the map-only mapreduce on.
@@ -66,7 +82,7 @@ public class ResaveAllEppResourcesAction implements Runnable {
         .setModuleName("backend")
         .setDefaultMapShards(NUM_SHARDS)
         .runMapOnly(
-            new ResaveAllEppResourcesActionMapper(),
+            new ResaveAllEppResourcesActionMapper(isFast),
             ImmutableList.of(EppResourceInputs.createKeyInput(EppResource.class)))
         .sendLinkToMapreduceConsole(response);
   }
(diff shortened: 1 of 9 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/912)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6Eocm8iTQzY6mCuZb:-MP6Eocm8iTQzY6mCuZc:b-3t8wqh) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java#L50)):*
> ```Java
>  * Note that this does NOT pick up mutations caused by migrations using the {@link
>  * com.googlecode.objectify.annotation.OnLoad} annotation, so if you are running a one-off schema
>  * migration, do not use fast mode. Fast mode defaults to false for this reason, but is used by the
> ```

Are we planning on doing any migrations in the nearish future? Relying on this `fast` parameter might possibly mean that it we do run any schema migrations, we might run in to issues with the sizes of the commit logs after that time (or we might not have issues, we don't have test data) 
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6EYFG9SeMGmY9Xlhi:-MP6EYFG9SeMGmY9Xlhj:buzte7s) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java#L121)):*
> ```Java
>               String.format(
>                   "%s entities %s",
>                   resourceKey.getKind(), resaved ? "re-saved" : "with no changes skipped"));
> ```

the nittiest: period at end of sentence?
___
*[core/src/main/java/google/registry/env/alpha/default/WEB-INF/cron.xml, line 81 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6F9qR-rLGn6IY54YU:-MP6F9qR-rLGn6IY54YV:bypc33a) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cron.xml#L81)):*
> ```XML
>     <target>backend</target>
>   </cron>
> ```

huh, I wonder why we don't run it in crash
___
*[core/src/test/java/google/registry/batch/ResaveAllEppResourcesActionTest.java, line 64 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6FVDgCExwjOas41Wm:-MP6FVDgCExwjOas41Wn:b-n57t5s) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/test/java/google/registry/batch/ResaveAllEppResourcesActionTest.java#L64)):*
> ```Java
>     assertThat(ofy().load().entity(contact).now().getUpdateTimestamp().getTimestamp())
>         .isEqualTo(creationTime);
>     ofy().clearSessionCache();
> ```

Does clearing the session cache increase the clock that ofy uses? 

edit: Just making sure that the time would otherwise have increased -- but I actually see that in the first test of the class, it does appear to increase the update timestamp otherwise.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/912)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6Eocm8iTQzY6mCuZb:-MP6I64o-rNCY0wmbI32:b-s8999b) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java#L50)):*

Are we planning on doing any migrations in the nearish future? Relying on this `fast` parameter might possibly mean that it we do run any schema migrations, we might run in to issues with the sizes of the commit logs after that time (or we might not have issues, we don't have test data) 
</blockquote></details>

I suspect we'll do at least one in the future.

Note that, with 10 shards, it shouldn't be too bad (not much worse than the 15-30 MB you already saw).  The real problem would be if we ran it with 100 shards.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6EYFG9SeMGmY9Xlhi:-MP6IFc_1Pfut_V8Qduz:bvs2g1t) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java#L121)):*

the nittiest: period at end of sentence?
</blockquote></details>

Nope.  These are GAE MapReduce counters, not logging statements.  Here's what they look like in use: https://sookocheff.com/post/appengine/mapreduce/mapreduce-yaml/
___
*[core/src/main/java/google/registry/env/alpha/default/WEB-INF/cron.xml, line 81 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6F9qR-rLGn6IY54YU:-MP6IYqtEWezF5PeNxuv:bqiqffe) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cron.xml#L81)):*

huh, I wonder why we don't run it in crash
</blockquote></details>

Oversight?  To be honest there's little reason to run it in alpha/qa either.
___
*[core/src/test/java/google/registry/batch/ResaveAllEppResourcesActionTest.java, line 64 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6FVDgCExwjOas41Wm:-MP6Ik8BEIKzbtf1kpAy:bphyf0k) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/test/java/google/registry/batch/ResaveAllEppResourcesActionTest.java#L64)):*

Does clearing the session cache increase the clock that ofy uses? 

edit: Just making sure that the time would otherwise have increased -- but I actually see that in the first test of the class, it does appear to increase the update timestamp otherwise.
</blockquote></details>

If you drill down into how `MapreduceTestCase` executes the fake mapreduces, you'll see where the `FakeClock` gets advanced if one is in use.

However, one isn't in use here; like most other mapreduce tests, this one uses the real system clock, so the actual update time is likely to be seconds later than the initial creation time.

gbrodman (approved):
> Reviewed 1 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/912)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6Eocm8iTQzY6mCuZb:-MP6Xhh_3QSmXreYUwn2:bpltx55) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java#L50)):*

I suspect we'll do at least one in the future.

Note that, with 10 shards, it shouldn't be too bad (not much worse than the 15-30 MB you already saw).  The real problem would be if we ran it with 100 shards.
</blockquote></details>

Right, so we shouldn't increase back to 100 shards any time soon.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/912)* status: 1 of 9 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @hstonec)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/912#-MP6Eocm8iTQzY6mCuZb:-MP6Yc3YCiFhR3UhZHhe:brzmnd0) ([raw file](https://github.com/google/nomulus/blob/a6afdc8aa0b5074c319ee88dcf6d5cd74673ce38/core/src/main/java/google/registry/batch/ResaveAllEppResourcesAction.java#L50)):*

Right, so we shouldn't increase back to 100 shards any time soon. 
</blockquote></details>

Good thing that's not in this PR then.

PR #914, opened 2020-12-22
Enforce existence (or explicitly don't) in txn manager bulk loads

gbrodman:
The contract of load(Iterable<VKey>) specifies that the method will
throw a NoSuchElementException if any of the specified keys don't exist.
We don't do that before this PR.

This also adds loadExisting and loadAllExisting methods that explicitly do not throw a NSEE in these cases, but rather only return the elements that we could find. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/914)

### DIFF
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -16,6 +16,7 @@ package google.registry.model.ofy;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
 
@@ -23,6 +24,8 @@ import com.google.common.base.Functions;
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.Result;
@@ -197,6 +200,21 @@ public class DatastoreTransactionManager implements TransactionManager {
 
   @Override
   public <T> ImmutableMap<VKey<? extends T>, T> load(Iterable<? extends VKey<? extends T>> keys) {
+    ImmutableMap<VKey<? extends T>, T> result = loadExisting(keys);
+    ImmutableSet<? extends VKey<? extends T>> missingKeys =
+        Streams.stream(keys).filter(k -> !result.containsKey(k)).collect(toImmutableSet());
+    if (!missingKeys.isEmpty()) {
+      // Ofy ignores nonexistent keys but the method contract specifies to throw if nonexistent
+      throw new NoSuchElementException(
+          String.format(
+              "Expected to find the following VKeys but they were missing: %s.", missingKeys));
+    }
+    return result;
+  }
+
+  @Override
+  public <T> ImmutableMap<VKey<? extends T>, T> loadExisting(
+      Iterable<? extends VKey<? extends T>> keys) {
     // Keep track of the Key -> VKey mapping so we can translate them back.
     ImmutableMap<Key<T>, VKey<? extends T>> keyMap =
         StreamSupport.stream(keys.spliterator(), false)
@@ -217,6 +235,16 @@ public class DatastoreTransactionManager implements TransactionManager {
 
   @Override
   public <T> ImmutableList<T> loadAll(Iterable<T> entities) {
+    ImmutableList<T> result = loadAllExisting(entities);
+    if (result.size() != Iterables.size(entities)) {
+      throw new NoSuchElementException(
+          String.format("Attempted to load entities, some of which are missing: %s", entities));
+    }
+    return result;
+  }
+
+  @Override
+  public <T> ImmutableList<T> loadAllExisting(Iterable<T> entities) {
     return ImmutableList.copyOf(getOfy().load().entities(entities).values());
   }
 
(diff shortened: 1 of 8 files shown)


CydeWeys (changes requested):
> The contract of load(Iterable<VKey>) specifies that the method will
> throw a NoSuchElementException if any of the specified keys don't exist.

There's more work to be done here to make this contract super explicit through better Javadoc documentation on `TransactionManager` (see below).

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/914#-:-MPFl9l2EoJG3rJyLxxI:bfjz5jd)* status: 0 of 8 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/914#-MPFv1Vb73MPpLR4jcwa:-MPFv1Vb73MPpLR4jcwb:bm8fd5e) ([raw file](https://github.com/google/nomulus/blob/f3b61b606a798a5390fb757fe61d81309f24ac6c/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L210)):*
> ```Java
>       throw new NoSuchElementException(
>           String.format(
>               "Expected to find the following VKeys but they were missing: %s.", missingKeys));
> ```

Shouldn't have a period at the end.  Also I'd do something more terse and explicit like `Failed loading non-existent entities: %s`
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 216 at r1](https://reviewable.io/reviews/google/nomulus/914#-MPFpJnj7Jch_iPfOkbK:-MPFpJnj7Jch_iPfOkbL:b-lmjafn) ([raw file](https://github.com/google/nomulus/blob/f3b61b606a798a5390fb757fe61d81309f24ac6c/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L216)):*
> ```Java
> 
>   @Override
>   public <T> ImmutableMap<VKey<? extends T>, T> loadExisting(
> ```

I don't think the name of this method is properly reconciled with the name of `maybeLoad`.  It seems to me that `doFoobar(X)` and `doFoobar(List<X>)` should have similar method names, maybe even identical method names, just relying on function overloading.

More broadly there are some naming issues in the `TransactionManager` API that need to be fixed, e.g. `loadAll(Class<T> clazz)` seems properly named whereas `loadAll(Iterable<T> entities)` does NOT -- what the latter is doing is actually much more akin to `load(Iterable<? extends VKey<? extends T>> keys)` and should be named similarly.

We kind of have multiple different axes we're dealing with here.  There's one situation, whether we're loading by key or by entity, and then there's a cross-product with whether we return a List of results and throw when non-existent or return a Map of results and not throw, and then separately from all of that, we also have loading of all entities of a given entity class.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 406 at r1](https://reviewable.io/reviews/google/nomulus/914#-MPFwylg6-w7gFJKbOY7:-MPFwylh2J1m8gD0Go0o:b-2lwdrt) ([raw file](https://github.com/google/nomulus/blob/f3b61b606a798a5390fb757fe61d81309f24ac6c/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L406)):*
> ```Java
>   @Override
>   public <T> ImmutableList<T> loadAllExisting(Iterable<T> entities) {
>     return Streams.stream(entities).filter(this::exists).map(this::load).collect(toImmutableList());
> ```

Are we sure that this is performant?  It doesn't seem like we're actually performing a batch operation against the DB?
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 203 at r1](https://reviewable.io/reviews/google/nomulus/914#-MPFmKwQA2uiVzgKXwmf:-MPFmKwQA2uiVzgKXwmg:bh24fix) ([raw file](https://github.com/google/nomulus/blob/f3b61b606a798a5390fb757fe61d81309f24ac6c/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L203)):*
> ```Java
>   <T> ImmutableMap<VKey<? extends T>, T> load(Iterable<? extends VKey<? extends T>> keys);
> 
>   /** Loads the set of entities by their key id, nonexistent entiites are ignored. */
> ```

Typo, bad use of comma.

Also not clear what "ignored" means.  Be more explicit about exactly what happens if the given key is present -- I assume what happens is that the returned Map omits the corresponding key?  If so that should be mentioned.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/914#-MPFm2N107H65THv_6XA:-MPFm2N107H65THv_6XB:b1c2hkh) ([raw file](https://github.com/google/nomulus/blob/f3b61b606a798a5390fb757fe61d81309f24ac6c/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L210)):*
> ```Java
> 
>   /**
>    * Loads all given entities from the database, throws NoSuchElementException if it doesn't exist.
> ```

Use the same style of Javadoc as what `load` has above (using the `@throws` row).  Having that as metadata for IDEs is useful given that we're using unchecked exceptions that aren't explicitly declared in `throws` clauses.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 214 at r1](https://reviewable.io/reviews/google/nomulus/914#-MPFmMmT3DeyAOwXXctf:-MPFmMmT3DeyAOwXXctg:bh24fix) ([raw file](https://github.com/google/nomulus/blob/f3b61b606a798a5390fb757fe61d81309f24ac6c/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L214)):*
> ```Java
>   <T> ImmutableList<T> loadAll(Iterable<T> entities);
> 
>   /** Loads all given entities from the database if possible, nonexistent entiites are ignored. */
> ```

Typo, bad use of comma.

Also not clear what "ignored" means.  Be more explicit about exactly what happens if the given key is present -- I assume what happens is that the returned Map omits the corresponding key?  If so that should be mentioned.

gbrodman:
Closing this in favor of a more substantial renaming

PR #918, opened 2020-12-30
Refactor naming and behavior of bulk load methods in TransactionManager

gbrodman:
The contract of loadByKeys(Iterable<VKey>) specifies that the method will
throw a NoSuchElementException if any of the specified keys don't exist.
We don't do that before this PR, but now we do.

Existing calls (when necessary) were converted to the new load*
methods, which have the same behavior as the previous methods.

Existing methods were also renamed to be more clear -- see b/176239831
for more details and discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/918)


gbrodman:
Most of the changes are just due to the refactor -- the only meaningful differences are in TransactionManager, JpaTransactionManagerImpl, and DatastoreTransactionManager

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/918)* status: 0 of 80 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/918#-MQNWEbLADGIoI_RBUgm:-MQNWEbLADGIoI_RBUgn:b-7vc9jv) ([raw file](https://github.com/google/nomulus/blob/f87f8912d41bc128c8a973e0da9efeed8c0098ba/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L385)):*
> ```Java
>     return Streams.stream(entities)
>         .filter(this::exists)
>         .map(this::loadByEntity)
> ```

I'm still questioning whether this is actually as performant as running a query that is explicitly loading everything in a single SQL statement -- but it doesn't need to be addressed now.  We need to run some performance profiling (and this is already how stuff is in SQL already anyway so it's not a regression).

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/918#-:-MQNWkNm-1OPAJnRDVd6:bnfp4nl)* status: 0 of 80 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/918)* status: 0 of 80 files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/918#-MQNWEbLADGIoI_RBUgm:-MQNbxOs7LGvLl1rYJgk:b-h8x9jc) ([raw file](https://github.com/google/nomulus/blob/f87f8912d41bc128c8a973e0da9efeed8c0098ba/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L385)):*

I'm still questioning whether this is actually as performant as running a query that is explicitly loading everything in a single SQL statement -- but it doesn't need to be addressed now.  We need to run some performance profiling (and this is already how stuff is in SQL already anyway so it's not a regression).
</blockquote></details>

I think you're almost certainly correct, and I would imagine this is a problem that we'll deal with in the future.

PR #921, opened 2021-01-06
Convert some Host and Contact flows to tm()

hstonec:
This PR converted `HostDeleteFlow`, `ContactDeleteFlow` and `ContactInfoFlow` to use `tm()`. The main reason we put them in the same PR is because they all use some common methods in `ResourceFlowUtils` and `EppResourceUtils` that need to be converted together as well.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/921)

### DIFF
--- a/core/src/main/java/google/registry/flows/ResourceFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/ResourceFlowUtils.java
@@ -14,14 +14,11 @@
 
 package google.registry.flows;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.Sets.intersection;
+import static google.registry.model.EppResourceUtils.getLinkedDomainKeys;
 import static google.registry.model.EppResourceUtils.loadByForeignKey;
-import static google.registry.model.EppResourceUtils.queryForLinkedDomains;
 import static google.registry.model.index.ForeignKeyIndex.loadAndGetKey;
 import static google.registry.model.ofy.ObjectifyService.ofy;
-import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 
 import com.google.common.collect.ImmutableSet;
@@ -50,7 +47,6 @@ import google.registry.model.domain.Period;
 import google.registry.model.domain.rgp.GracePeriodStatus;
 import google.registry.model.eppcommon.AuthInfo;
 import google.registry.model.eppcommon.StatusValue;
-import google.registry.model.host.HostResource;
 import google.registry.model.index.ForeignKeyIndex;
 import google.registry.model.transfer.TransferStatus;
 import google.registry.persistence.VKey;
@@ -60,7 +56,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.function.Predicate;
-import javax.persistence.Query;
 import org.joda.time.DateTime;
 
 /** Static utility functions for resource flows. */
@@ -70,14 +65,6 @@ public final class ResourceFlowUtils {
 
   /** In {@link #failfastForAsyncDelete}, check this (arbitrary) number of query results. */
   private static final int FAILFAST_CHECK_COUNT = 5;
-
-  // We have to use the native SQL query here because DomainHost table doesn't have its entity
-  // class so we cannot reference its property like domainHost.hostRepoId in a JPQL query.
-  private static final String HOST_LINKED_DOMAIN_QUERY =
-      "SELECT d.repo_id FROM \"Domain\" d "
-          + "JOIN \"DomainHost\" dh ON dh.domain_repo_id = d.repo_id "
-          + "WHERE d.deletion_time > :now "
-          + "AND dh.host_repo_id = :fkRepoId";
 
   /** Check that the given clientId corresponds to the owner of given resource. */
   public static void verifyResourceOwnership(String myClientId, EppResource resource)
@@ -93,10 +80,6 @@ public final class ResourceFlowUtils {
       final DateTime now,
       final Class<R> resourceClass,
       final Function<DomainBase, ImmutableSet<?>> getPotentialReferences) throws EppException {
-    checkArgument(
-        resourceClass.equals(ContactResource.class) || resourceClass.equals(HostResource.class),
-        "resourceClass must be either ContactResource or HostResource, but it is %s",
-        resourceClass);
     // Enter a transactionless context briefly.
     EppException failfastException =
         tm().doTransactionless(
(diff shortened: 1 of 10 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/921)* status: all files reviewed, 3 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSn_Q8CGODkBWnCZep:-MQSn_Q9CAhJZ_sSAzNZ:bwj50dq) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L129)):*
> ```Java
>                       // TODO(shicong): Add a query to get contact's linked domain.
>                       throw new UnsupportedOperationException(
>                           "Query contact's linked domain is not supported");
> ```

In a vacuum, this exception message might not make a lot of sense.  Can we explain that in SQL, we have no way to grab domain links from contacts at the moment? 

Of course this might go away soon if you're implementing contact flow conversions soon anyway. 
___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSnpmd9dWPTKClaJ8K:-MQSnpmd9dWPTKClaJ8L:bhk5eob) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L135)):*
> ```Java
>                     }
>                     keys =
>                         (Iterable<VKey<DomainBase>>)
> ```

Is this cast necessary? What fails without it? 
___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 144 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSnuCh7ug_fWFjjG6m:-MQSnuCh7ug_fWFjjG6n:b-5gt693) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L144)):*
> ```Java
>                                     repoId ->
>                                         DomainBase.createVKey(
>                                             Key.create(DomainBase.class, (String) repoId)))
> ```

Can you not set the result type to be String somewhere so that we can avoid this cast? I thought that was possible but I might be wrong.

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/921)* status: 1 of 10 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSn_Q8CGODkBWnCZep:-MQXrE-T0oBg16suaR5Q:bfgpoa7) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L129)):*

In a vacuum, this exception message might not make a lot of sense.  Can we explain that in SQL, we have no way to grab domain links from contacts at the moment? 

Of course this might go away soon if you're implementing contact flow conversions soon anyway. 
</blockquote></details>

Yeah, my original idea was to keep the scope of the change smaller to help review but I think it also makes sense to make the change for both use cases together so I did that.

It turns out that it is better to refactor the `queryForLinkedDomains` in `EppResourceUtils` in stead of changing it here because `EppResourceUtils.isLinked` also uses `queryForLinkedDomains`.
___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSnpmd9dWPTKClaJ8K:-MQXs2Nd-N98FkirAIJ1:bxb8ajw) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L135)):*

Is this cast necessary? What fails without it? 
</blockquote></details>

Yeah, I think it mainly because `javax.persistence.Query` (returned by `createNativeQuery`) doesn't include the type information, it can only return the raw `List` or `Stream` without type information so we have to make the cast here.
___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 144 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSnuCh7ug_fWFjjG6m:-MQXt6_R8EkTiLSMK4o9:bgbyk6c) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L144)):*

Can you not set the result type to be String somewhere so that we can avoid this cast? I thought that was possible but I might be wrong. 
</blockquote></details>

The compiler complains the type mismatch if we don't do the cast, I guess it is also caused by the native `Query` object like the above case.

gbrodman (changes requested):
> Reviewed 9 of 9 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/921)* status: all files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/flows/ResourceFlowUtils.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/921#-MQSn_Q8CGODkBWnCZep:-MQYGWpu5v_v2conDTie:bqog80t) ([raw file](https://github.com/google/nomulus/blob/44a80c2049c9a373ae731a38f43f9db0363b8dc2/core/src/main/java/google/registry/flows/ResourceFlowUtils.java#L129)):*

Yeah, my original idea was to keep the scope of the change smaller to help review but I think it also makes sense to make the change for both use cases together so I did that.

It turns out that it is better to refactor the `queryForLinkedDomains` in `EppResourceUtils` in stead of changing it here because `EppResourceUtils.isLinked` also uses `queryForLinkedDomains`.
</blockquote></details>

Agreed
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 423 at r2](https://reviewable.io/reviews/google/nomulus/921#-MQYFpu95kVwjKXNe0uN:-MQYFpu95kVwjKXNe0uO:b-koa27h) ([raw file](https://github.com/google/nomulus/blob/70e555552a7469d1f1f3cc26975499ddf05b9002/core/src/main/java/google/registry/model/EppResourceUtils.java#L423)):*
> ```Java
>                           .getEntityManager()
>                           .createQuery(CONTACT_LINKED_DOMAIN_QUERY, String.class)
>                           .setParameter("fkRepoId", key)
> ```

key.getSqlKey() ?
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 433 at r2](https://reviewable.io/reviews/google/nomulus/921#-MQYFw4XCBKdiDMAO42Q:-MQYFw4XCBKdiDMAO42R:be99v42) ([raw file](https://github.com/google/nomulus/blob/70e555552a7469d1f1f3cc26975499ddf05b9002/core/src/main/java/google/registry/model/EppResourceUtils.java#L433)):*
> ```Java
>                           .setParameter("now", now.toDate());
>                 }
>                 return (ImmutableSet<VKey<DomainBase>>)
> ```

casting is gross but hopefully temporary in the long run?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/921)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 423 at r2](https://reviewable.io/reviews/google/nomulus/921#-MQYFpu95kVwjKXNe0uN:-MQYjzzC4nWjzqrQ7_mM:bfh2tfz) ([raw file](https://github.com/google/nomulus/blob/70e555552a7469d1f1f3cc26975499ddf05b9002/core/src/main/java/google/registry/model/EppResourceUtils.java#L423)):*

key.getSqlKey() ?
</blockquote></details>

This is a JPQL query which requires the parameter to be the same type as the property type(as you can see 1 line below that I passed `now` instead of `now.toDate()` because it requires jodaTime).
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 433 at r2](https://reviewable.io/reviews/google/nomulus/921#-MQYFw4XCBKdiDMAO42Q:-MQYl8Kq8IvgPiX-LVtV:b-jk0emd) ([raw file](https://github.com/google/nomulus/blob/70e555552a7469d1f1f3cc26975499ddf05b9002/core/src/main/java/google/registry/model/EppResourceUtils.java#L433)):*

casting is gross but hopefully temporary in the long run? 
</blockquote></details>

Yeah, I think after the migration we may be able to refactor this or get rid of the query at all.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/921)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #924, opened 2021-01-08
Convert remaining Contact flows to tm()

hstonec:
Converted the remaining Contact flows, i.e. `ContactCreateFlow`, `ContactUpdateFlow` and `ContactCheckFlow` to use `tm()`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/924)

### DIFF
--- a/core/src/main/java/google/registry/flows/contact/ContactCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/contact/ContactCreateFlow.java
@@ -19,9 +19,9 @@ import static google.registry.flows.ResourceFlowUtils.verifyResourceDoesNotExist
 import static google.registry.flows.contact.ContactFlowUtils.validateAsciiPostalInfo;
 import static google.registry.flows.contact.ContactFlowUtils.validateContactAgainstPolicy;
 import static google.registry.model.EppResourceUtils.createRepoId;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 
+import com.google.common.collect.ImmutableSet;
 import com.googlecode.objectify.Key;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.flows.EppException;
@@ -95,11 +95,12 @@ public final class ContactCreateFlow implements TransactionalFlow {
         .setModificationTime(now)
         .setXmlBytes(null)  // We don't want to store contact details in the history entry.
         .setParent(Key.create(newContact));
-    ofy().save().entities(
-        newContact,
-        historyBuilder.build(),
-        ForeignKeyIndex.create(newContact, newContact.getDeletionTime()),
-        EppResourceIndex.create(Key.create(newContact)));
+    tm().insertAll(
+            ImmutableSet.of(
+                newContact,
+                historyBuilder.build().toChildHistoryEntity(),
+                ForeignKeyIndex.create(newContact, newContact.getDeletionTime()),
+                EppResourceIndex.create(Key.create(newContact))));
     return responseBuilder
         .setResData(ContactCreateData.create(newContact.getContactId(), now))
         .build();
--- a/core/src/main/java/google/registry/flows/contact/ContactUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/contact/ContactUpdateFlow.java
@@ -24,7 +24,6 @@ import static google.registry.flows.ResourceFlowUtils.verifyOptionalAuthInfo;
 import static google.registry.flows.ResourceFlowUtils.verifyResourceOwnership;
 import static google.registry.flows.contact.ContactFlowUtils.validateAsciiPostalInfo;
 import static google.registry.flows.contact.ContactFlowUtils.validateContactAgainstPolicy;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 
 import com.google.common.collect.ImmutableSet;
@@ -151,7 +150,8 @@ public final class ContactUpdateFlow implements TransactionalFlow {
     }
     validateAsciiPostalInfo(newContact.getInternationalizedPostalInfo());
     validateContactAgainstPolicy(newContact);
-    ofy().save().<Object>entities(newContact, historyBuilder.build());
+    tm().insert(historyBuilder.build().toChildHistoryEntity());
+    tm().update(newContact);
     return responseBuilder.build();
   }
 
--- a/core/src/main/java/google/registry/model/eppcommon/Address.java
+++ b/core/src/main/java/google/registry/model/eppcommon/Address.java
@@ -35,6 +35,7 @@ import javax.persistence.Embeddable;
 import javax.persistence.MappedSuperclass;
 import javax.persistence.PostLoad;
 import javax.persistence.Transient;
+import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlTransient;
 import javax.xml.bind.annotation.adapters.CollapsedStringAdapter;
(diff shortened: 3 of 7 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/924)* status: all files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/eppcommon/Address.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/924#-MQYHdbN96L_h18AAsxT:-MQYHdbN96L_h18AAsxU:bwxix5w) ([raw file](https://github.com/google/nomulus/blob/5a419e081d99d4ae046a4a061d5b85b934d86145/core/src/main/java/google/registry/model/eppcommon/Address.java#L206)):*
> ```Java
>    * reconstructed from XML message.
>    *
>    * <p>This is a callback function that JAXB invokes after unmarshalling the XML message.
> ```

Is there really no annotation or anything for this? That's lame on the part of JAXB. Why is this change necessary here, out of curiosity?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/924)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/eppcommon/Address.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/924#-MQYHdbN96L_h18AAsxT:-MQYdPg39e5Alq-oSGfY:bd6pkd) ([raw file](https://github.com/google/nomulus/blob/5a419e081d99d4ae046a4a061d5b85b934d86145/core/src/main/java/google/registry/model/eppcommon/Address.java#L206)):*

Is there really no annotation or anything for this? That's lame on the part of JAXB. Why is this change necessary here, out of curiosity? 
</blockquote></details>

I couldn't find any annotation for this specific use case. So, in `ContactUpdateFlow`, the update XML request with address change will be unmarshalled to `Address` instance then set to the `ContactResource` if there is any difference. (See [here](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/contact/ContactUpdateFlow.java;l=120-133?q=contactupdateflow&ss=nomulus%2Fnomulus)). 

Right now, we need to keep both `List<String> street`(used in Datastore) and `streetLine1, streetLine2, streetLine3` (used in Cloud SQL) for backwards compatibility. So, when the `Address` instance is unmarshalled from the XML request, we also need to make sure all these street fields are set correctly, otherwise the street information will be lost when it is stored to Cloud SQL(because it uses streetLine1, streetLine2 and streetLine3 but only the List<String> street is set).

It would be nice if JAXB provides annotation for readonly elements like JPA does but I cannot find any, also I am not sure if there is a way to map multiple `<contact:street>` tags to `streetLine1, streetLine2, streetLine3` in a correct order. Below is part of the update XML request for your reference:

```xml
<contact:postalInfo type="int">
    <contact:org/>
    <contact:addr>
        <contact:street>124 Example Dr.</contact:street>
        <contact:street>Suite 200</contact:street>
        <contact:city>Dulles</contact:city>
        <contact:sp>VA</contact:sp>
        <contact:pc>20166-6503</contact:pc>
        <contact:cc>US</contact:cc>
    </contact:addr>
</contact:postalInfo>
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/924)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/model/eppcommon/Address.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/924#-MQYHdbN96L_h18AAsxT:-MQsUzSx7VCrQnpx5Jgk:bv8c6gd) ([raw file](https://github.com/google/nomulus/blob/5a419e081d99d4ae046a4a061d5b85b934d86145/core/src/main/java/google/registry/model/eppcommon/Address.java#L206)):*

I couldn't find any annotation for this specific use case. So, in `ContactUpdateFlow`, the update XML request with address change will be unmarshalled to `Address` instance then set to the `ContactResource` if there is any difference. (See [here](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/contact/ContactUpdateFlow.java;l=120-133?q=contactupdateflow&ss=nomulus%2Fnomulus)). 

Right now, we need to keep both `List<String> street`(used in Datastore) and `streetLine1, streetLine2, streetLine3` (used in Cloud SQL) for backwards compatibility. So, when the `Address` instance is unmarshalled from the XML request, we also need to make sure all these street fields are set correctly, otherwise the street information will be lost when it is stored to Cloud SQL(because it uses streetLine1, streetLine2 and streetLine3 but only the List<String> street is set).

It would be nice if JAXB provides annotation for readonly elements like JPA does but I cannot find any, also I am not sure if there is a way to map multiple `<contact:street>` tags to `streetLine1, streetLine2, streetLine3` in a correct order. Below is part of the update XML request for your reference:

```xml
<contact:postalInfo type="int">
    <contact:org/>
    <contact:addr>
        <contact:street>124 Example Dr.</contact:street>
        <contact:street>Suite 200</contact:street>
        <contact:city>Dulles</contact:city>
        <contact:sp>VA</contact:sp>
        <contact:pc>20166-6503</contact:pc>
        <contact:cc>US</contact:cc>
    </contact:addr>
</contact:postalInfo>
```
</blockquote></details>

OK cool, so this is basically to make sure that the SQL fields get populated correctly from XML requests. So this isn't directly related to the tm() conversions themselves -- the objects that were being created from XML weren't populated correctly anyway before this. 

Out of curiosity do you know if we have any tests that make sure that all the fields are populated correctly from the XML request? If so (and they include the `streetLine1` etc fields), they should have failed before this right?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/924)* status: 6 of 7 files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/model/eppcommon/Address.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/924#-MQYHdbN96L_h18AAsxT:-MQsbVpiDT3SAJwOgsPs:bk5ehxk) ([raw file](https://github.com/google/nomulus/blob/5a419e081d99d4ae046a4a061d5b85b934d86145/core/src/main/java/google/registry/model/eppcommon/Address.java#L206)):*
>OK cool, so this is basically to make sure that the SQL fields get populated correctly from XML requests. So this isn't directly related to the tm() conversions themselves -- the objects that were being created from XML weren't populated correctly anyway before this.

Correct, more accurately, the `Address` instance from XML was not created correctly to be used for Cloud SQL, because we set `@Transient` to `Address.street` because we don't want to deal with `List` in this case, instead we added three individual `Address.streetLine${num}` fields as the table columns to store the street information, and these `Address.streetLine${num}` fields are not populated from XML until this PR.

>Out of curiosity do you know if we have any tests that make sure that all the fields are populated correctly from the XML request? If so (and they include the streetLine1 etc fields), they should have failed before this right?

It seems that we just don't have such test so I added one. See `ContactCommandTest`.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/924)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/model/eppcommon/Address.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/924#-MQYHdbN96L_h18AAsxT:-MQwfNY-B6OeKeCfZmly:b-hkagas) ([raw file](https://github.com/google/nomulus/blob/5a419e081d99d4ae046a4a061d5b85b934d86145/core/src/main/java/google/registry/model/eppcommon/Address.java#L206)):*

>OK cool, so this is basically to make sure that the SQL fields get populated correctly from XML requests. So this isn't directly related to the tm() conversions themselves -- the objects that were being created from XML weren't populated correctly anyway before this.

Correct, more accurately, the `Address` instance from XML was not created correctly to be used for Cloud SQL, because we set `@Transient` to `Address.street` because we don't want to deal with `List` in this case, instead we added three individual `Address.streetLine${num}` fields as the table columns to store the street information, and these `Address.streetLine${num}` fields are not populated from XML until this PR.

>Out of curiosity do you know if we have any tests that make sure that all the fields are populated correctly from the XML request? If so (and they include the streetLine1 etc fields), they should have failed before this right?

It seems that we just don't have such test so I added one. See `ContactCommandTest`.
</blockquote></details>

great, thank you very much

PR #925, opened 2021-01-11
Add object comparison to replay tests

mindhog:
Allow optional object comparison in the replay test extension and enable it
for the DomainCreateFlow test.

To faciliate this, add two new field annotations to ImmutableObject:
DoNotCompare, to be used for fields that are not relevant to either database,
and Insignificant, to be used for fields that are mutated after they have been                                                                    
accessed and therefore violate immutability (there is currently only one of                                                                       
these, however we might discover more in the course of adding more comparisons
to the replay test.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/925)

### DIFF
--- a/core/src/main/java/google/registry/model/EppResource.java
+++ b/core/src/main/java/google/registry/model/EppResource.java
@@ -148,7 +148,7 @@ public abstract class EppResource extends BackupGroupRoot implements Buildable {
    *
    * @see google.registry.model.translators.CommitLogRevisionsTranslatorFactory
    */
-  @Transient
+  @Transient @DoNotCompare
   ImmutableSortedMap<DateTime, Key<CommitLogManifest>> revisions = ImmutableSortedMap.of();
 
   public String getRepoId() {
--- a/core/src/main/java/google/registry/model/ImmutableObject.java
+++ b/core/src/main/java/google/registry/model/ImmutableObject.java
@@ -54,9 +54,37 @@ public abstract class ImmutableObject implements Cloneable {
   @Target(FIELD)
   public @interface DoNotHydrate {}
 
-  @Ignore
-  @XmlTransient
-  Integer hashCode;
+  /**
+   * Indicates that the field should be ignored when comparing an object in the datastore to the
+   * corresponding object in Cloud SQL.
+   */
+  @Documented
+  @Retention(RUNTIME)
+  @Target(FIELD)
+  public @interface DoNotCompare {}
+
+  /**
+   * Indicates that the field stores a null value to indicate an empty set. This is also used in
+   * object comparison.
+   */
+  @Documented
+  @Retention(RUNTIME)
+  @Target(FIELD)
+  public @interface EmptySetToNull {}
+
+  /**
+   * Indicates that the field does not take part in the immutability contract.
+   *
+   * <p>Certain fields currently get modified by hibernate and there is nothing we can do about it.
+   * As well as violating immutability, this breaks hashing and equality comparisons, so we mark
+   * these fields with this annotation to exclude them from most operations.
+   */
+  @Documented
+  @Retention(RUNTIME)
+  @Target(FIELD)
+  public @interface Insignificant {}
+
+  @Ignore @XmlTransient protected Integer hashCode;
 
   private boolean equalsImmutableObject(ImmutableObject other) {
     return getClass().equals(other.getClass())
@@ -71,7 +99,14 @@ public abstract class ImmutableObject implements Cloneable {
    * <p>Isolated into a method so that derived classes can override it.
    */
   protected Map<Field, Object> getSignificantFields() {
-    return ModelUtils.getFieldValues(this);
+    // Can't use streams or ImmutableMap because we can have null values.
+    Map<Field, Object> result = new LinkedHashMap();
+    for (Map.Entry<Field, Object> entry : ModelUtils.getFieldValues(this).entrySet()) {
+      if (!entry.getKey().isAnnotationPresent(Insignificant.class)) {
+        result.put(entry.getKey(), entry.getValue());
+      }
+    }
+    return result;
   }
 
   @Override
--- a/core/src/main/java/google/registry/model/domain/DomainContent.java
+++ b/core/src/main/java/google/registry/model/domain/DomainContent.java
@@ -50,6 +50,7 @@ import com.googlecode.objectify.condition.IfNull;
 import google.registry.flows.ResourceFlowUtils;
 import google.registry.model.EppResource;
 import google.registry.model.EppResource.ResourceWithTransferData;
+import google.registry.model.ImmutableObject.EmptySetToNull;
 import google.registry.model.billing.BillingEvent;
 import google.registry.model.common.EntityGroupRoot;
 import google.registry.model.contact.ContactResource;
@@ -132,7 +133,7 @@ public class DomainContent extends EppResource
   @Index String tld;
 
   /** References to hosts that are the nameservers for the domain. */
-  @Index @Transient Set<VKey<HostResource>> nsHosts;
+  @EmptySetToNull @Index @Transient Set<VKey<HostResource>> nsHosts;
 
   /**
    * The union of the contacts visible via {@link #getContacts} and {@link #getRegistrant}.
@@ -319,6 +320,11 @@ public class DomainContent extends EppResource
     autorenewPollMessageHistoryId = getHistoryId(autorenewPollMessage);
     autorenewBillingEventHistoryId = getHistoryId(autorenewBillingEvent);
     deletePollMessageHistoryId = getHistoryId(deletePollMessage);
+
+    // Fix PollMessage VKeys.
+    autorenewPollMessage = PollMessage.Autorenew.convertVKey(autorenewPollMessage);
+    deletePollMessage = PollMessage.OneTime.convertVKey(deletePollMessage);
+
     dsData =
         nullToEmptyImmutableCopy(dsData).stream()
             .map(dsData -> dsData.cloneWithDomainRepoId(getRepoId()))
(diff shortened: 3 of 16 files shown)


hstonec (changes requested):
> Reviewed 6 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/925)* status: 6 of 16 files reviewed, 3 unresolved discussions (waiting on @hstonec and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumListUtils.java, line 116 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrTf1f4Pn3NySoXEvC:-MQrTf1f4Pn3NySoXEvD:b9aae0r) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/main/java/google/registry/model/registry/label/PremiumListUtils.java#L116)):*
> ```
>       //      logger.atSevere().withCause(t).log(
>       //          "Error loading price of domain %s.%s from Cloud SQL.", label,
>       // registry.getTldStr());
> ```
Why comment out these logs?
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 43 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrcMoe0BZhA9RlNZke:-MQrcMof8TIOFq-Ep0Px:b207p4f) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L43)):*
> ```
>  * <p>If the "compare" flag is set in the constructor, this will also compare all touched objects in
>  * both databases after performing the replay.
> ```
How do we determine if we need to enable the comparison? I guess I am just worried if it is safe to not compare everything as ideally they are supposed to be same.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrUQt7AQyzlWeOCELi:-MQrUQt7AQyzlWeOCELj:b8oy1ra) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L50)):*

> ```
>   public ReplayExtension(FakeClock clock, boolean compare) {
>     this.clock = clock;
>     this.compare = compare;
>   }
> ```
</details>

How about having some static factory methods to avoid the boolean parameter? Like `createComparable(clock)` and `createNonComparable(clock)`.

Also, why is the `@DoNotCompare` not sufficient in the case that `compare = false`? If certain mutable fields are excluded from the comparison, I'd expect the rest of the fields are supposed to be same.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/925)* status: 6 of 15 files reviewed, 3 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumListUtils.java, line 116 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrTf1f4Pn3NySoXEvC:-MR0JMDY3K5jxQObTwB9:bopyedo) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/main/java/google/registry/model/registry/label/PremiumListUtils.java#L116)):*

> ```
>       //      logger.atSevere().withCause(t).log(
>       //          "Error loading price of domain %s.%s from Cloud SQL.", label,
>       // registry.getTldStr());
> ```
Why comment out these logs?
</blockquote></details>

Gah!  I knew this was going to happen.  I've been commenting it out locally to remove the noisy severe error stack trace (as Ben suggested).  It was only a matter of time before it ended up in a CL.

Thanks for catching it!
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 43 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrcMoe0BZhA9RlNZke:-MR0JsmD2ZFCN5G_q4KO:blhnkf7) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L43)):*

> ```
>  * <p>If the "compare" flag is set in the constructor, this will also compare all touched objects in
>  * both databases after performing the replay.
> ```
How do we determine if we need to enable the comparison? I guess I am just worried if it is safe to not compare everything as ideally they are supposed to be same.
</blockquote></details>

We want to enable it for all replay tests.  I've put it on a flag because it likely requires fixes and I didn't want to have to deal with fixing everything all at once.

I'm going to work through all of the replay tests with this.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrUQt7AQyzlWeOCELi:-MR0L_TwD5U8_sjssCIH:bx9zifl) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L50)):*

> ```
>   public ReplayExtension(FakeClock clock, boolean compare) {
>     this.clock = clock;
>     this.compare = compare;
>   }
> ```
How about having some static factory methods to avoid the boolean parameter? Like `createComparable(clock)` and `createNonComparable(clock)`.

Also, why is the `@DoNotCompare` not sufficient in the case that `compare = false`? If certain mutable fields are excluded from the comparison, I'd expect the rest of the fields are supposed to be same.
</blockquote></details>

Added the static factory methods (used createWith/Without as that seemed more consistent with extension naming).

Regarding DoNotCompare, I'm not sure I understand the question?

If you're asking why we also need "Insignificant" , the semantics are different: "DoNotCompare" affects only isEqualAcrossDatabases(), which is done specifically for cross database comparison and does not break immutability.  "Insignificant" affects the memoized hash value,  A difference in those fields affects equality checks throughout the codebase.

hstonec (changes requested):
> Reviewed 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/925)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrUQt7AQyzlWeOCELi:-MR5zXFoDkpjzbRgLdRm:b-19ik13) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L50)):*

Added the static factory methods (used createWith/Without as that seemed more consistent with extension naming).

Regarding DoNotCompare, I'm not sure I understand the question?

If you're asking why we also need "Insignificant" , the semantics are different: "DoNotCompare" affects only isEqualAcrossDatabases(), which is done specifically for cross database comparison and does not break immutability.  "Insignificant" affects the memoized hash value,  A difference in those fields affects equality checks throughout the codebase.
 
</blockquote></details>

I see, thanks for the explanation. So, I think what I am confused is what is the problem that prevents us from doing the cross database comparison even we exclude certain fields with `@DoNotCompare`?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/925)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @hstonec)

___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrUQt7AQyzlWeOCELi:-MR6Fd-i1zoB2t_H1Ji0:b-t09p4e) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L50)):*

I see, thanks for the explanation. So, I think what I am confused is what is the problem that prevents us from doing the cross database comparison even we exclude certain fields with `@DoNotCompare`?
</blockquote></details>

I believe it is because the one case where we're using the Insignificant field is DomainTransactionRecord.id, which is not currently a top-level entity object and we're only checking the DoNotCompare label on top-level objects (by "top-level" I mean "the object is an Entity in Datastore").  That is to say, isEqualAcrossDatabases() is not currently recursive. 

This in itself is problematic, and I need to do something about it before comparisons will work for DomainRenewFlowTest (which needs a DoNotCompare for the serverApproveEntities field on a TransferData object that is accessed through a DomainBase object).  However, I think in conjunction with that we'll probably also want to do something to ensure that the final hash values of our persisted objects are correct -- basically, enforcing the use of the Insignificant annotation where appropriate.

hstonec (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/925)* status: 7 of 15 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 50 at r1](https://reviewable.io/reviews/google/nomulus/925#-MQrUQt7AQyzlWeOCELi:-MR6U5UPDv9rPe5GUXzz:b9y6xnt) ([raw file](https://github.com/google/nomulus/blob/00dd911ecb395ebe714f2331cd8b4f79b881b182/core/src/test/java/google/registry/testing/ReplayExtension.java#L50)):*

I believe it is because the one case where we're using the Insignificant field is DomainTransactionRecord.id, which is not currently a top-level entity object and we're only checking the DoNotCompare label on top-level objects (by "top-level" I mean "the object is an Entity in Datastore").  That is to say, isEqualAcrossDatabases() is not currently recursive. 

This in itself is problematic, and I need to do something about it before comparisons will work for DomainRenewFlowTest (which needs a DoNotCompare for the serverApproveEntities field on a TransferData object that is accessed through a DomainBase object).  However, I think in conjunction with that we'll probably also want to do something to ensure that the final hash values of our persisted objects are correct -- basically, enforcing the use of the Insignificant annotation where appropriate.  
</blockquote></details>

I see, that makes sense.

PR #926, opened 2021-01-11
Add databaseTransitionSchedule entity and tool for updating

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/926)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRZwKkDQ5JdKyEiXZw:-MRRZwKkDQ5JdKyEiXZx:b-m6nobk) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L37)):*
> ```Java
>     separators = " =",
>     commandDescription = "Add a new entry to the database transition schedule.")
> public class UpdateDatabaseTransitionScheduleCommand extends MutatingCommand {
> ```

I hope we don't ever need a way to remove an entry from the transition schedule, as so far one isn't provided.  Yes it may be a little bit clunkier from the point of view of the user to always have to specify the schedule in full (as we see with the other commands operating on these, e.g. the `--tld_state_transitions` parameter to the `create|update_tld` commands), but it seems to me that the implementation is simpler and the flexibility at point of use is greater.  It also eliminates the `START_OF_TIME` problem you're seeing below.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 40 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRR_T89Emki8TSC2L1L:-MRR_T8ACJD0vLgI6OLP:b3cqrqz) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L40)):*
> ```Java
> 
>   @Parameter(names = "--start_time", description = "Datetime for database to become primary.")
>   DateTime start;
> ```

Should be named `startTime`, not just `start`, so that it matches its parameter name.

Also I think we want validation somewhere that the time is at least an hour in the future (or however long the in-memory cache for `DatabaseTransitionSchedule` is going to last).
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRY_QP1BeiUNTxC8Kk:-MRRY_QP1BeiUNTxC8Kl:b-ivfjrr) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L59)):*
> ```Java
>         ofyTm().transact(() -> ofyTm().loadByKeyIfPresent(key));
> 
>     boolean scheduleWasNull = false;
> ```

Is this boolean necessary?  Why not just use `!currentSchedule.isPresent()` in all places instead?  This simplifies things a bit.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 66 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRYzLE4lCS45Hpq1tH:-MRRYzLE4lCS45Hpq1tI:b-c2q2mh) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L66)):*
> ```Java
>     if (!currentSchedule.isPresent()) {
>       scheduleWasNull = true;
>       currentSchedule =
> ```

Why re-initialize`currentSchedule` here instead of initializing `newSchedule` with this single value?  Just treat `currentSchedule` as immutable and leave it alone.  This makes the logic in `stageEntityChange` go away to boot.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 83 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRYiUe6K61GZnL8N9d:-MRRYiUe6K61GZnL8N9e:bhbluzh) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L83)):*
> ```Java
>     DatabaseTransitionSchedule.Builder newSchedule = new DatabaseTransitionSchedule.Builder();
>     newSchedule.setDatabaseTransitions(newDatabaseTransitions.build());
>     stageEntityChange(scheduleWasNull ? null : currentSchedule.get(), newSchedule.build());
> ```

This then just becomes `stageEntityChange(currentSchedule.orElse(null), ...);`

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRZwKkDQ5JdKyEiXZw:-MRuRofLDSb3g5Gcvt_m:b-896fix) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L37)):*

I hope we don't ever need a way to remove an entry from the transition schedule, as so far one isn't provided.  Yes it may be a little bit clunkier from the point of view of the user to always have to specify the schedule in full (as we see with the other commands operating on these, e.g. the `--tld_state_transitions` parameter to the `create|update_tld` commands), but it seems to me that the implementation is simpler and the flexibility at point of use is greater.  It also eliminates the `START_OF_TIME` problem you're seeing below.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 40 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRR_T89Emki8TSC2L1L:-MRgFKDmB8cXE07m72Df:bpuxphv) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L40)):*

Should be named `startTime`, not just `start`, so that it matches its parameter name.

Also I think we want validation somewhere that the time is at least an hour in the future (or however long the in-memory cache for `DatabaseTransitionSchedule` is going to last).
</blockquote></details>

I'm not sure how to do this now that I switched the logic to update the whole schedule
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRY_QP1BeiUNTxC8Kk:-MRgNAa4AiwNqM4oDkPv:b-896fix) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L59)):*

Is this boolean necessary?  Why not just use `!currentSchedule.isPresent()` in all places instead?  This simplifies things a bit.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 66 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRYzLE4lCS45Hpq1tH:-MRgNBbU7nAevgUg-J6U:b-896fix) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L66)):*

Why re-initialize`currentSchedule` here instead of initializing `newSchedule` with this single value?  Just treat `currentSchedule` as immutable and leave it alone.  This makes the logic in `stageEntityChange` go away to boot.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 83 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRYiUe6K61GZnL8N9d:-MRgNB5SAFhxhEWIDZRO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L83)):*

This then just becomes `stageEntityChange(currentSchedule.orElse(null), ...);`
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 0 of 11 files reviewed, 7 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 115 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9sEv5-hTRu69NbYIs:-MS9sEv5-hTRu69NbYIt:bl7u8di) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L115)):*
> ```Java
> 
>   /** Returns the database that is indicated as primary at the given time. */
>   public PrimaryDatabase getPrimaryDatabase(DateTime now) {
> ```

What "now" is intended to be passed in here?  Is this a transaction time?  If so, for what database?  If we don't 100% use the same clock in all situations then bad things can happen.
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 36 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9vSm07TfrRok1bXc6:-MS9vSm07TfrRok1bXc7:bvtnd3h) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L36)):*
> ```Java
>       DatabaseTransitionSchedule schedule =
>           checkArgumentPresent(
>               DatabaseTransitionSchedule.get(id),
> ```

See comment on set command for more info, but this shouldn't be using the cached path.
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tEbSDcGU7MmpYQK1:-MS9tEbT2XMUqfUYjDsF:b-uso38d) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L40)):*
> ```Java
>               id);
> 
>       System.out.println(schedule.getDatabaseTransitions());
> ```

This is cleaner if you implicitly use (or implement as necessary) `toString()`, in particular because you're going to want to output the id so that the output in the multiple schedule case is legible.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/926#-MRRY_QP1BeiUNTxC8Kk:-MS9vn6263NEWL0-Wahp:b-8uz25r) ([raw file](https://github.com/google/nomulus/blob/216f1aaf32cee94f110204042851d16591675234/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L59)):*

Done.
</blockquote></details>

Thanks, this is overall simpler and easier to follow now.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 31 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9t_Ry29IE4KFj8cex:-MS9t_Ry29IE4KFj8cey:b46yy22) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L31)):*
> ```Java
> @Parameters(
>     separators = " =",
>     commandDescription = "Add a new entry to the database transition schedule.")
> ```

This command description is a little off.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 32 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tgYv6DZ0hBYswV9p:-MS9tgYv6DZ0hBYswV9q:b-r1j0gs) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L32)):*
> ```Java
>     separators = " =",
>     commandDescription = "Add a new entry to the database transition schedule.")
> public class UpdateDatabaseTransitionScheduleCommand extends MutatingCommand {
> ```

I don't think `Update` is quite the right verb anymore, given that this command is also used for the initial create of a schedule.  `Set` feels like a better fit.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 43 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tvCq6RZm9MeO8HVi:-MS9tvCq6RZm9MeO8HVj:bvkktuf) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L43)):*
> ```Java
>   ImmutableSortedMap<DateTime, PrimaryDatabase> transitionSchedule;
> 
>   @Parameter(names = "--schedule_id", description = "ID string for the schedule being updated")
> ```

May as well just make this `--id`; it's completely unambiguous, and `nomulus set_database_transition_schedule --schedule_id foobar` is a bit superfluous.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 49 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9v7qI7DkzrkWvWAOb:-MS9v7qI7DkzrkWvWAOc:b-b8whxz) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L49)):*
> ```Java
>   protected void init() {
>     Optional<DatabaseTransitionSchedule> currentSchedule =
>         DatabaseTransitionSchedule.get(scheduleId);
> ```

This is dangerous.  You have to actually transactionally get the current schedule as it currently is in the DB, not a potentially stale version from a cache that could be up to 10 minutes out of date.  The cache should only be used for non-mutating uses.  `get()` should be renamed to `getCached()` to make this potential pitfall obvious, and it needs a warning on the method about when to use it.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 0 of 11 files reviewed, 7 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 115 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9sEv5-hTRu69NbYIs:-MSF7gNDDAYZVxjc2nAT:b5k0p5h) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L115)):*

What "now" is intended to be passed in here?  Is this a transaction time?  If so, for what database?  If we don't 100% use the same clock in all situations then bad things can happen.
</blockquote></details>

I was thinking the transaction time, and callers would use tm().getTransactionTime
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 36 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9vSm07TfrRok1bXc6:-MSF61f497OEO8It209-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L36)):*

See comment on set command for more info, but this shouldn't be using the cached path.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tEbSDcGU7MmpYQK1:-MSEO7a57mBoGzitxc7Q:b-896fix) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L40)):*

This is cleaner if you implicitly use (or implement as necessary) `toString()`, in particular because you're going to want to output the id so that the output in the multiple schedule case is legible.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 31 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9t_Ry29IE4KFj8cex:-MSExR8J-Wng-4oLQs7p:b-896fix) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L31)):*

This command description is a little off.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 32 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tgYv6DZ0hBYswV9p:-MSExRRb8tbdVlX2IC1K:b-896fix) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L32)):*

I don't think `Update` is quite the right verb anymore, given that this command is also used for the initial create of a schedule.  `Set` feels like a better fit.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 43 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tvCq6RZm9MeO8HVi:-MSEz6ciBiQIGILTtdzU:bs7wumx) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L43)):*

May as well just make this `--id`; it's completely unambiguous, and `nomulus set_database_transition_schedule --schedule_id foobar` is a bit superfluous.
</blockquote></details>

Done. 
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 49 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9v7qI7DkzrkWvWAOb:-MSF5wdSCYwTHIicuC2D:b-896fix) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L49)):*

This is dangerous.  You have to actually transactionally get the current schedule as it currently is in the DB, not a potentially stale version from a cache that could be up to 10 minutes out of date.  The cache should only be used for non-mutating uses.  `get()` should be renamed to `getCached()` to make this potential pitfall obvious, and it needs a warning on the method about when to use it.
</blockquote></details>

Done.

mindhog (changes requested):
> Reviewed 2 of 8 files at r1, 5 of 9 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: all files reviewed, 10 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 118 at r3](https://reviewable.io/reviews/google/nomulus/926#-MSZLMnp6buXUnzuk02M:-MSZLMnq-U1REbjWU7GN:b-i0rm0p) ([raw file](https://github.com/google/nomulus/blob/b540f5e2488e2cb8609e2fe69932454032325ec7/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L118)):*
> ```Java
> 
>   /**
>    * Returns the current cached schedule for the given id.
> ```

You should probably document, here and above, that "id" is an entity.  In fact, instead of id, you might want to make the variable "entityName" and document the expectation ("unqualified entity name", presumably.)
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 43 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tvCq6RZm9MeO8HVi:-MSZM5AuDJAhBHQJjMTm:bfsg2nu) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L43)):*

Done. 
</blockquote></details>

So, once again, if this is going to be used exclusively for entity types, you should indicate that this is expected to be the "unqualified entity class name" in the help text.

If you're planning something more general than this, you should remove the "transition schedule for an entity" language in the command name.

If we're going to do groups of entities in a single conversion, it might be better to drop this notion entirely and replace the free-form string with an enum -- we're going to need to modify specific pieces of code to accommodate specific transitions, so it's not unreasonable to constrain this to an enum defining the specific transition.

In fact, in that case you should probably just replace the identifier with an enum all the way down.
___
*[core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java, line 35 at r3](https://reviewable.io/reviews/google/nomulus/926#-MSZPFii77LZqjhcbZzp:-MSZPFijE1DGuGBDaOMy:b-162m1w) ([raw file](https://github.com/google/nomulus/blob/b540f5e2488e2cb8609e2fe69932454032325ec7/core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java#L35)):*
> ```Java
> 
>   @Test
>   void testSuccess() throws Exception {
> ```

This test should really verify output, here and below.  See, for example, [assertStdoutIs()](https://github.com/google/nomulus/blob/98283a67ac603a2ac07007827c54441530b3bb81/core/src/test/java/google/registry/tools/CommandTestCase.java#L206)
___
*[core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java, line 47 at r3](https://reviewable.io/reviews/google/nomulus/926#-MSZQVE3AkVKrNfaCyIs:-MSZQVE3AkVKrNfaCyIt:b-7dizgh) ([raw file](https://github.com/google/nomulus/blob/b540f5e2488e2cb8609e2fe69932454032325ec7/core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java#L47)):*
> ```Java
> 
>   @Test
>   void testSuccess_multipleArguments() throws Exception {
> ```

Same here, verify output.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 4 of 11 files reviewed, 10 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 118 at r3](https://reviewable.io/reviews/google/nomulus/926#-MSZLMnp6buXUnzuk02M:-MSZqjITBHol9gh9-6Ib:bud7i65) ([raw file](https://github.com/google/nomulus/blob/b540f5e2488e2cb8609e2fe69932454032325ec7/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L118)):*

You should probably document, here and above, that "id" is an entity.  In fact, instead of id, you might want to make the variable "entityName" and document the expectation ("unqualified entity name", presumably.)
</blockquote></details>

changed to TransitionId enum
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 43 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tvCq6RZm9MeO8HVi:-MSZn9Sw7mIPoDu8o_sm:b-q344w0) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L43)):*

So, once again, if this is going to be used exclusively for entity types, you should indicate that this is expected to be the "unqualified entity class name" in the help text.

If you're planning something more general than this, you should remove the "transition schedule for an entity" language in the command name.

If we're going to do groups of entities in a single conversion, it might be better to drop this notion entirely and replace the free-form string with an enum -- we're going to need to modify specific pieces of code to accommodate specific transitions, so it's not unreasonable to constrain this to an enum defining the specific transition.

In fact, in that case you should probably just replace the identifier with an enum all the way down.
</blockquote></details>

Changed to an enum
___
*[core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java, line 35 at r3](https://reviewable.io/reviews/google/nomulus/926#-MSZPFii77LZqjhcbZzp:-MSZqe9CDA9xH1fCVi_j:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b540f5e2488e2cb8609e2fe69932454032325ec7/core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java#L35)):*

This test should really verify output, here and below.  See, for example, [assertStdoutIs()](https://github.com/google/nomulus/blob/98283a67ac603a2ac07007827c54441530b3bb81/core/src/test/java/google/registry/tools/CommandTestCase.java#L206)
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java, line 47 at r3](https://reviewable.io/reviews/google/nomulus/926#-MSZQVE3AkVKrNfaCyIs:-MSZqel12LhZRX8KpO_y:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b540f5e2488e2cb8609e2fe69932454032325ec7/core/src/test/java/google/registry/tools/GetDatabaseTransitionScheduleCommandTest.java#L47)):*

Same here, verify output.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 4 of 11 files reviewed, 10 unresolved discussions (waiting on @CydeWeys, @mindhog, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 29 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_-g1A99EN01JBkvkd:-MS_-g1A99EN01JBkvke:bwsw6ds) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L29)):*
> ```Java
> final class GetDatabaseTransitionScheduleCommand implements CommandWithRemoteApi {
> 
>   @Parameter(description = "Transition id string for the schedule to get", required = true)
> ```

Description doesn't handle possible plurality correctly; should be more like "Transition id(s) ..."
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 37 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_1Dg3-F6gR9lcddBw:-MS_1Dg3-F6gR9lcddBx:bllo5i) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L37)):*
> ```Java
>       DatabaseTransitionSchedule schedule =
>           checkArgumentPresent(
>               DatabaseTransitionSchedule.get(TransitionId.valueOf(id)),
> ```

Better to let JCommander's extensibility handle the conversion for you and have the main parameters just be typed correctly.  See `EnumParameter` with an example implementation at `TldStateParameter`; ignoring the header lines/imports it's just a 1-liner, then you'll be able to define `mainParameters` as correctly being the `List<TransitionId>` that it actually is.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 4 of 11 files reviewed, 12 unresolved discussions (waiting on @mindhog and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 115 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9sEv5-hTRu69NbYIs:-MS_3Ndm2JMHN9pbXu3N:bqierm2) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L115)):*

I was thinking the transaction time, and callers would use tm().getTransactionTime
</blockquote></details>

If the callers are always using `getTransactionTime()` then why not just inline that here, which adds the side benefit that callers can now no longer call the method incorrectly?
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 55 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_34uUCzI_iszbhPs3:-MS_34uV1A7UpiwSnYhe:bp6418f) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L55)):*
> ```Java
> 
>   /** The id of the transition schedule. */
>   public enum TransitionId {
> ```

These values need Javadocs explaining what they are.  You might also consider a display field that could be used in output.
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 56 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_35Cu5y7lrdf5_xeZ:-MS_35Cu5y7lrdf5_xe_:bvm5wzq) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L56)):*
> ```Java
>   /** The id of the transition schedule. */
>   public enum TransitionId {
>     TEST,
> ```

What is `TEST` exactly?
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 57 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_36It3qS7WUmrcgOi:-MS_36It3qS7WUmrcgOj:b-an3vqz) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L57)):*
> ```Java
>   public enum TransitionId {
>     TEST,
>     SMD
> ```

I don't think this name is quite right.
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 81 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_3CeW7kBusQ2eu2zw:-MS_3CeW7kBusQ2eu2zx:b6htjcg) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L81)):*
> ```Java
>   @Parent Key<EntityGroupRoot> parent = getCrossTldKey();
> 
>   @Id String transitionId;
> ```

Ooof, this can't be typed as `TransitionId` while still being a Datastore `@Id`?  Did you try?
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 145 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_4GCU63pmHepxxO-1:-MS_4GCU63pmHepxxO-2:b-siur9p) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L145)):*
> ```Java
>   @Override
>   public String toString() {
>     return String.format("%s : %s", transitionId, databaseTransitions.toValueMap());
> ```

Nit, but, Java, like the English language, tends to use a space after a colon but not before it.  I can see situations where the bit before the colon might be really long and complicated and that added space could be helpful for clarity, but in this particular case with the possible transition ID values just being short uppercased words (and no punctuation!), it's not necessary.
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 40 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tEbSDcGU7MmpYQK1:-MS_3t1f9kz_wCcjGbkM:b-e84459) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L40)):*

Done.
</blockquote></details>

So you're currently explicitly using `toString()`, but you could be implicitly using it, which I think might look cleaner.  `System.out.println(_____.toString())` is a little bit redundant.
___
*[core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java, line 32 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9tgYv6DZ0hBYswV9p:-MS_2g-Z-DJ0ltdO_tuv:bulygo3) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/tools/UpdateDatabaseTransitionScheduleCommand.java#L32)):*

Done.
</blockquote></details>

Is it?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 3 of 12 files reviewed, 11 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 115 at r2](https://reviewable.io/reviews/google/nomulus/926#-MS9sEv5-hTRu69NbYIs:-MS_PE14-yKqXBH9GY3g:b-896fix) ([raw file](https://github.com/google/nomulus/blob/60e3dc15614e6fc3bf6fd2b147e2351540ed73d2/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L115)):*

If the callers are always using `getTransactionTime()` then why not just inline that here, which adds the side benefit that callers can now no longer call the method incorrectly?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 55 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_34uUCzI_iszbhPs3:-MS_bOCe4purswuRU17m:bn6lh9z) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L55)):*

These values need Javadocs explaining what they are.  You might also consider a display field that could be used in output.
</blockquote></details>

Added javadocs. Did not add display fields because I am not sure why they would be necessary. I fear that if the output prints out a name other than the Enum value, it could be confusing to know what value actually needs to be passed into the nomulus tool. 
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 56 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_35Cu5y7lrdf5_xeZ:-MS_bBh3CcoIMdKNcoyY:bm04lrm) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L56)):*

What is `TEST` exactly?
</blockquote></details>

I included a TEST id for creating a schedule for tests
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 57 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_36It3qS7WUmrcgOi:-MS_bIO-9clP5Io3RmXX:b-y4z1pg) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L57)):*

I don't think this name is quite right.
</blockquote></details>

Changed to SIGNED_MARK_REVOCATION_LIST
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 81 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_3CeW7kBusQ2eu2zw:-MS_LtgW1Y0JSSmvTEhI:b-x0k41g) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L81)):*

Ooof, this can't be typed as `TransitionId` while still being a Datastore `@Id`?  Did you try?
</blockquote></details>

Yes, I tried. Objectify did not allow it.
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 29 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_-g1A99EN01JBkvkd:-MS_UkS9Bt2TQDkWxNU9:b-896fix) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L29)):*

Description doesn't handle possible plurality correctly; should be more like "Transition id(s) ..."
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java, line 37 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_1Dg3-F6gR9lcddBw:-MS_WmFhESjKN7CFNVc1:b-896fix) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/tools/GetDatabaseTransitionScheduleCommand.java#L37)):*

Better to let JCommander's extensibility handle the conversion for you and have the main parameters just be typed correctly.  See `EnumParameter` with an example implementation at `TldStateParameter`; ignoring the header lines/imports it's just a 1-liner, then you'll be able to define `mainParameters` as correctly being the `List<TransitionId>` that it actually is.
</blockquote></details>

Done.

mindhog (approved):
> Reviewed 1 of 7 files at r4, 8 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @mindhog)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 81 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_3CeW7kBusQ2eu2zw:-MSisWSD9lToeOqH6R_w:bl16iim) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L81)):*

Yes, I tried. Objectify did not allow it.
</blockquote></details>

Oh well.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: all files reviewed, 7 unresolved discussions (waiting on @mindhog and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 56 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_35Cu5y7lrdf5_xeZ:-MSiu72u9h2MOAcuSZTv:b-pphxtz) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L56)):*

I included a TEST id for creating a schedule for tests
</blockquote></details>

Why not just use `SIGNED_MARK_REVOCATION_LIST` for tests?  It's not common for test data to be a possible valid value in production code paths.  As far as I can tell, the tests just need a value (any value) to work; it doesn't matter what it is since they're not actually writing that data for real anywhere.  See e.g. this enum, which is widely used and doesn't have a type dedicated for testing:   https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/reporting/HistoryEntry.java;l=119;bpv=1;bpt=1?q=historyentry&ss=nomulus%2Fnomulus
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 43 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSiwEAr7l569zqoQ7Dz:-MSiwEAr7l569zqoQ7E-:bv9f0e1) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L43)):*
> ```Java
> 
>   Key<DatabaseTransitionSchedule> key;
>   private final TransactionManager tm = new DatastoreTransactionManager(new Ofy(fakeClock));
> ```

`CommandTestCase` already has a `fakeClock`, which is accessible to you here in this class, and the `AppEngineExtension` is already being initialized with said fake clock.  So I don't think you need this here at all, and indeed, it could be causing potential problems.

If you do need to inject that fake clock into ofy, the best way to do so is by using the `InjectExtension to run `inject.setStaticField(Ofy.class, "clock", clock);` in the setup method; that way you can use `tm()` at each call site (the way it's meant to be used, and more consistent with production code) than calling `tm()` once and saving it to a local variable (which is a pattern discouraged in production code because `TransactionManager` itself is supposed to be doing the instance-handling).
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 55 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSiulqeERaJ8chTS4_j:-MSiulqfBbJhhq5Y9YzX:beghf1p) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L55)):*
> ```Java
>     assertThat(ofy().load().key(key).now()).isNull();
>     runCommandForced(
>         "--transition_id=TEST", String.format("--transition_schedule=%s=DATASTORE", START_OF_TIME));
> ```

It's our convention to hard-code parameters in tests so that they're more readable.  The less logic in tests the better.  See http://go/tott/510 and http://go/tott/319
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 60 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSivSSj5ryQ3CMzrn6A:-MSivSSk9I_9d9GSFzgu:b-t1xbnr) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L60)):*
> ```Java
>                 () -> DatabaseTransitionSchedule.get(TransitionId.TEST).get().getPrimaryDatabase()))
>         .isEqualTo(PrimaryDatabase.DATASTORE);
>     String changes = command.prompt();
> ```

There's no need to have this on a separate line.  The test is actually slightly more readable if it's all inlined, as it's then immediately clear what's being asserted on (the output of the command's prompt) without having to back-scan the test to figure out what `changes` is.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 82 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSixS8z6v8AwA6M3k9f:-MSixS8z6v8AwA6M3k9g:b66srq7) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L82)):*
> ```Java
>         "--transition_id=TEST",
>         String.format(
>             "--transition_schedule=%s=DATASTORE,%s=CLOUD_SQL,%s=DATASTORE",
> ```

Same issue as above -- don't use String formatting, and take advantage of the varargs of `runCommandForced()` rather than concatenating all the parameters into a single string here.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSixOwv3o1AoTh0gre2:-MSixOwv3o1AoTh0gre3:b9h63v1) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L90)):*
> ```Java
>                         .get()
>                         .getDatabaseTransitions()))
>         .hasSize(3);
> ```

Should make an actual assertion on the contents of the Map.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 8 of 12 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 56 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_35Cu5y7lrdf5_xeZ:-MSj3OxH5kVrc5yP5MFp:bo67ox5) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L56)):*

Why not just use `SIGNED_MARK_REVOCATION_LIST` for tests?  It's not common for test data to be a possible valid value in production code paths.  As far as I can tell, the tests just need a value (any value) to work; it doesn't matter what it is since they're not actually writing that data for real anywhere.  See e.g. this enum, which is widely used and doesn't have a type dedicated for testing:   https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/reporting/HistoryEntry.java;l=119;bpv=1;bpt=1?q=historyentry&ss=nomulus%2Fnomulus
</blockquote></details>

Ok, I needed a second enum value for some tests, so I added PREMIUM_RESERVED_LIST since I plan on migrating those two entities together.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 43 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSiwEAr7l569zqoQ7Dz:-MSj6eHLEXwBwchuRTfm:brlf3dc) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L43)):*

`CommandTestCase` already has a `fakeClock`, which is accessible to you here in this class, and the `AppEngineExtension` is already being initialized with said fake clock.  So I don't think you need this here at all, and indeed, it could be causing potential problems.

If you do need to inject that fake clock into ofy, the best way to do so is by using the `InjectExtension to run `inject.setStaticField(Ofy.class, "clock", clock);` in the setup method; that way you can use `tm()` at each call site (the way it's meant to be used, and more consistent with production code) than calling `tm()` once and saving it to a local variable (which is a pattern discouraged in production code because `TransactionManager` itself is supposed to be doing the instance-handling).
</blockquote></details>

Removed it
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 55 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSiulqeERaJ8chTS4_j:-MSjH98CBftYwwag4Xk2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L55)):*

It's our convention to hard-code parameters in tests so that they're more readable.  The less logic in tests the better.  See http://go/tott/510 and http://go/tott/319
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 60 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSivSSj5ryQ3CMzrn6A:-MSj6qVJEeRaLjv_zpmz:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L60)):*

There's no need to have this on a separate line.  The test is actually slightly more readable if it's all inlined, as it's then immediately clear what's being asserted on (the output of the command's prompt) without having to back-scan the test to figure out what `changes` is.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 82 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSixS8z6v8AwA6M3k9f:-MSjHWPN4CGOobZAWs_x:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L82)):*

Same issue as above -- don't use String formatting, and take advantage of the varargs of `runCommandForced()` rather than concatenating all the parameters into a single string here.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSixOwv3o1AoTh0gre2:-MSj7Wts3PhAENxma0tM:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L90)):*

Should make an actual assertion on the contents of the Map.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 8 of 12 files reviewed, 6 unresolved discussions (waiting on @mindhog and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 56 at r4](https://reviewable.io/reviews/google/nomulus/926#-MS_35Cu5y7lrdf5_xeZ:-MSmog6IDiV-rplRegCg:b-ao56ix) ([raw file](https://github.com/google/nomulus/blob/658e54e857a6acfc28bdafe72655f99f95e5b9e1/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L56)):*

Ok, I needed a second enum value for some tests, so I added PREMIUM_RESERVED_LIST since I plan on migrating those two entities together.
</blockquote></details>

A better name for this new one miiiight be `DOMAIN_LABEL_LISTS`.  Or at least make the existing one plural, e.g. `PREMIUM_AND_RESERVED_LISTS`.
___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 57 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmsMz2BTaMlK7qS8Xs:-MSmsMz2BTaMlK7qS8Xt:b-2w4kpw) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L57)):*
> ```Java
>     DATASTORE
>   }
> ```

A comment I was leaving below just made me realize that we want an `UpdateAutoTimestamp` field on this entity so we know the last time it was saved (this should be super simple to add, and there's lots of examples throughout the rest of the codebase).
___
*[core/src/test/java/google/registry/model/common/DatabaseTransitionScheduleTest.java, line 36 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmtfmF-Z8rU0WX150E:-MSmtfmGBpDOTMG2yNdH:bg2nj4x) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/test/java/google/registry/model/common/DatabaseTransitionScheduleTest.java#L36)):*
> ```Java
> public class DatabaseTransitionScheduleTest extends EntityTestCase {
> 
>   private final TransactionManager tm = new DatastoreTransactionManager(new Ofy(fakeClock));
> ```

Same comment as on the command tests -- this isn't really the correct way to set up a test transaction manager, or to use one generally, and conveniently here `EntityTestCase` is already setting the transaction manager using a proper `@RegisterExtension`, and injecting a fake clock into `Ofy` for you, so this line shouldn't be needed at all (and its presence may cause things to perform unexpectedly, since your tests here are not actually using the transaction manager and fake clock that the rest of the testing framework expects that you are).

Below you're inconsistently calling either `ofyTm()` (which uses the one set up by the test framework and is the more correct way to go) and `tm` (the one you've created here), and that's a recipe for confusion as the tests grow more complicated.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/926#-MSixOwv3o1AoTh0gre2:-MSmsn3wDRrv7aqIhqc1:b-8ywpi) ([raw file](https://github.com/google/nomulus/blob/6476cf2237f5549e6d7057fb0c1aa88bc300aa9f/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L90)):*

Done.
</blockquote></details>

I meant along the lines of something like:

```
assertThat(______.getDatabaseTransitions()).containsExactly(.....);
```

You added an assertion on a single state of the map, not all of them, but it makes sense to assert on the entire contents of the map.

Also once you've done so you can remove the `hasSize(3)` call as that's superfluous.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 76 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmqQp_CAZQX57QzEtG:-MSmqQp_CAZQX57QzEtH:bbjs7ys) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L76)):*
> ```Java
>             TransitionId.SIGNED_MARK_REVOCATION_LIST,
>             TimedTransitionProperty.fromValueMap(transitionMap, PrimaryDatabaseTransition.class));
>     persistResource(schedule);
> ```

Inline this into the previous line and then set `schedule` to the return value.  More importantly than saving a line (which it does), it also means that `schedule` is getting the correct value, as `persistResource()` returns the value of the item after it has been stored and then retrieved from the DB, which might be different from what the object looked like prior to being saved into the DB e.g. for reasons of autogenerated IDs being assigned, create/update auto timestamps being assigned, any `@OnSave` methods being triggered, etc.  I don't think any of them apply here (for now anyway), but it's a good habit to get into.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 102 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmtKxJ6MDfd7ngtDh0:-MSmtKxJ6MDfd7ngtDh1:bozne13) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L102)):*
> ```Java
>                             .getPrimaryDatabase()))
>         .isEqualTo(PrimaryDatabase.DATASTORE);
>     String changes = command.prompt();
> ```

You can inline this for clarity per previous content.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926)* status: 7 of 12 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java, line 57 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmsMz2BTaMlK7qS8Xs:-MSn9Cy44Zb7RBMy5BL3:b-896fix) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/main/java/google/registry/model/common/DatabaseTransitionSchedule.java#L57)):*

A comment I was leaving below just made me realize that we want an `UpdateAutoTimestamp` field on this entity so we know the last time it was saved (this should be super simple to add, and there's lots of examples throughout the rest of the codebase).
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/common/DatabaseTransitionScheduleTest.java, line 36 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmtfmF-Z8rU0WX150E:-MSnE6qs44e8dbFXZjC6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/test/java/google/registry/model/common/DatabaseTransitionScheduleTest.java#L36)):*

Same comment as on the command tests -- this isn't really the correct way to set up a test transaction manager, or to use one generally, and conveniently here `EntityTestCase` is already setting the transaction manager using a proper `@RegisterExtension`, and injecting a fake clock into `Ofy` for you, so this line shouldn't be needed at all (and its presence may cause things to perform unexpectedly, since your tests here are not actually using the transaction manager and fake clock that the rest of the testing framework expects that you are).

Below you're inconsistently calling either `ofyTm()` (which uses the one set up by the test framework and is the more correct way to go) and `tm` (the one you've created here), and that's a recipe for confusion as the tests grow more complicated.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 76 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmqQp_CAZQX57QzEtG:-MSnEbKm7-LcuVlo3J0d:b-896fix) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L76)):*

Inline this into the previous line and then set `schedule` to the return value.  More importantly than saving a line (which it does), it also means that `schedule` is getting the correct value, as `persistResource()` returns the value of the item after it has been stored and then retrieved from the DB, which might be different from what the object looked like prior to being saved into the DB e.g. for reasons of autogenerated IDs being assigned, create/update auto timestamps being assigned, any `@OnSave` methods being triggered, etc.  I don't think any of them apply here (for now anyway), but it's a good habit to get into.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java, line 102 at r6](https://reviewable.io/reviews/google/nomulus/926#-MSmtKxJ6MDfd7ngtDh0:-MSnEk_uDDxEwZkk2CdW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/275906edda578588f523b019c5af96089d42b4aa/core/src/test/java/google/registry/tools/SetDatabaseTransitionScheduleCommandTest.java#L102)):*

You can inline this for clarity per previous content.
</blockquote></details>

Done.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/926#-:-MT2XVtN6Z8JxPup1f0t:bnfp4nl)* status: 7 of 12 files reviewed, 1 unresolved discussion (waiting on @mindhog)

PR #936, opened 2021-01-20
Use CertificateChecker on login

sarahcaseybot:
Use the CertificateChecker to check if login certificate contains any violations. If it does, the violations are logged, but the login still succeeds.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/936)

### DIFF
--- a/core/src/main/java/google/registry/flows/TlsCredentials.java
+++ b/core/src/main/java/google/registry/flows/TlsCredentials.java
@@ -26,6 +26,7 @@ import dagger.Module;
 import dagger.Provides;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.flows.EppException.AuthenticationErrorException;
+import google.registry.flows.certs.CertificateChecker;
 import google.registry.model.registrar.Registrar;
 import google.registry.request.Header;
 import google.registry.util.CidrAddressBlock;
@@ -72,6 +73,8 @@ public class TlsCredentials implements TransportCredentials {
     this.clientCertificate = clientCertificate;
     this.clientInetAddr = clientAddress.map(TlsCredentials::parseInetAddress);
   }
+
+  @Inject public CertificateChecker certificateChecker;
 
   static InetAddress parseInetAddress(String asciiAddr) {
     try {
@@ -146,6 +149,20 @@ public class TlsCredentials implements TransportCredentials {
       // Check if the certificate is equal to the one on file for the registrar.
       if (clientCertificate.equals(registrar.getClientCertificate())
           || clientCertificate.equals(registrar.getFailoverClientCertificate())) {
+        // Check certificate for any requirement violations
+        // TODO(Sarahbot@): Throw exceptions instead of just logging once requirement enforcement
+        // begins
+        try {
+          certificateChecker.validateCertificate(clientCertificate.get());
+        } catch (Exception e) {
+          if (e instanceof IllegalArgumentException) {
+            logger.atWarning().log(
+                "Registrar certificate used for %s does not meet certificate requirements: %s",
+                registrar.getClientId(), e.getMessage());
+          } else {
+            logger.atWarning().log("Error validating certificate for %s", registrar.getClientId());
+          }
+        }
         // successfully validated, return here since hash validation is not necessary
         return;
       }
--- a/core/src/main/java/google/registry/flows/TlsCredentials.java
+++ b/core/src/main/java/google/registry/flows/TlsCredentials.java
@@ -235,13 +235,16 @@ public class TlsCredentials implements TransportCredentials {
     }
   }
 
-  /** Registrar certificate contains security violations. */
+  /** Registrar certificate contains the following security violations: ... */
   public static class CertificateContainsSecurityViolationsException
       extends AuthenticationErrorException {
     InsecureCertificateException exception;
 
     CertificateContainsSecurityViolationsException(InsecureCertificateException exception) {
-      super(exception.getMessage());
+      super(
+          String.format(
+              "Registrar certificate contains the following security violations:\n%s",
+              exception.getMessage()));
       this.exception = exception;
     }
   }
--- a/core/src/main/java/google/registry/flows/TlsCredentials.java
+++ b/core/src/main/java/google/registry/flows/TlsCredentials.java
@@ -25,6 +25,7 @@ import com.google.common.net.InetAddresses;
 import dagger.Module;
 import dagger.Provides;
 import google.registry.config.RegistryConfig.Config;
+import google.registry.config.RegistryEnvironment;
 import google.registry.flows.EppException.AuthenticationErrorException;
 import google.registry.flows.certs.CertificateChecker;
 import google.registry.model.registrar.Registrar;
@@ -155,6 +156,11 @@ public class TlsCredentials implements TransportCredentials {
         try {
           certificateChecker.validateCertificate(clientCertificate.get());
         } catch (Exception e) {
+          // throw exception in unit tests and Sandbox
+          if (RegistryEnvironment.get().equals(RegistryEnvironment.UNITTEST)
+              || RegistryEnvironment.get().equals(RegistryEnvironment.SANDBOX)) {
+            throw new CertificateContainsSecurityViolationsException(e.getMessage());
+          }
           if (e instanceof IllegalArgumentException) {
             logger.atWarning().log(
                 "Registrar certificate used for %s does not meet certificate requirements: %s",
@@ -225,6 +231,14 @@ public class TlsCredentials implements TransportCredentials {
   public static class BadRegistrarCertificateException extends AuthenticationErrorException {
     BadRegistrarCertificateException() {
       super("Registrar certificate does not match stored certificate");
+    }
+  }
+
+  /** Registrar certificate contains security violations. */
+  public static class CertificateContainsSecurityViolationsException
+      extends AuthenticationErrorException {
+    CertificateContainsSecurityViolationsException(String message) {
+      super(message);
     }
   }
 
--- a/core/src/main/java/google/registry/flows/TlsCredentials.java
+++ b/core/src/main/java/google/registry/flows/TlsCredentials.java
@@ -28,6 +28,7 @@ import google.registry.config.RegistryConfig.Config;
 import google.registry.config.RegistryEnvironment;
 import google.registry.flows.EppException.AuthenticationErrorException;
 import google.registry.flows.certs.CertificateChecker;
+import google.registry.flows.certs.CertificateChecker.InsecureCertificateException;
 import google.registry.model.registrar.Registrar;
 import google.registry.request.Header;
 import google.registry.util.CidrAddressBlock;
@@ -62,20 +63,21 @@ public class TlsCredentials implements TransportCredentials {
   private final Optional<String> clientCertificateHash;
   private final Optional<String> clientCertificate;
   private final Optional<InetAddress> clientInetAddr;
+  private final CertificateChecker certificateChecker;
 
   @Inject
   public TlsCredentials(
       @Config("requireSslCertificates") boolean requireSslCertificates,
       @Header("X-SSL-Certificate") Optional<String> clientCertificateHash,
       @Header("X-SSL-Full-Certificate") Optional<String> clientCertificate,
-      @Header("X-Forwarded-For") Optional<String> clientAddress) {
+      @Header("X-Forwarded-For") Optional<String> clientAddress,
+      CertificateChecker certificateChecker) {
     this.requireSslCertificates = requireSslCertificates;
     this.clientCertificateHash = clientCertificateHash;
     this.clientCertificate = clientCertificate;
     this.clientInetAddr = clientAddress.map(TlsCredentials::parseInetAddress);
+    this.certificateChecker = certificateChecker;
   }
-
-  @Inject public CertificateChecker certificateChecker;
 
   static InetAddress parseInetAddress(String asciiAddr) {
     try {
@@ -155,19 +157,18 @@ public class TlsCredentials implements TransportCredentials {
         // begins
         try {
           certificateChecker.validateCertificate(clientCertificate.get());
-        } catch (Exception e) {
+        } catch (InsecureCertificateException e) {
           // throw exception in unit tests and Sandbox
           if (RegistryEnvironment.get().equals(RegistryEnvironment.UNITTEST)
               || RegistryEnvironment.get().equals(RegistryEnvironment.SANDBOX)) {
-            throw new CertificateContainsSecurityViolationsException(e.getMessage());
+            throw new CertificateContainsSecurityViolationsException(e);
           }
-          if (e instanceof IllegalArgumentException) {
-            logger.atWarning().log(
-                "Registrar certificate used for %s does not meet certificate requirements: %s",
-                registrar.getClientId(), e.getMessage());
-          } else {
-            logger.atWarning().log("Error validating certificate for %s", registrar.getClientId());
-          }
+          logger.atWarning().log(
+              "Registrar certificate used for %s does not meet certificate requirements: %s",
+              registrar.getClientId(), e.getMessage());
+        } catch (Exception e) {
+          logger.atWarning().log(
+              "Error validating certificate for %s: %s", registrar.getClientId(), e);
         }
         // successfully validated, return here since hash validation is not necessary
         return;
@@ -237,8 +238,11 @@ public class TlsCredentials implements TransportCredentials {
   /** Registrar certificate contains security violations. */
   public static class CertificateContainsSecurityViolationsException
       extends AuthenticationErrorException {
-    CertificateContainsSecurityViolationsException(String message) {
-      super(message);
+    InsecureCertificateException exception;
+
+    CertificateContainsSecurityViolationsException(InsecureCertificateException exception) {
+      super(exception.getMessage());
+      this.exception = exception;
     }
   }
 
--- a/core/src/main/java/google/registry/flows/certs/CertificateChecker.java
+++ b/core/src/main/java/google/registry/flows/certs/CertificateChecker.java
@@ -89,14 +89,14 @@ public class CertificateChecker {
    * Checks the given certificate string for violations and throws an exception if any violations
    * exist.
    */
-  public void validateCertificate(String certificateString) {
+  public void validateCertificate(String certificateString) throws InsecureCertificateException {
     ImmutableSet<CertificateViolation> violations = checkCertificate(certificateString);
     if (!violations.isEmpty()) {
       String displayMessages =
           violations.stream()
               .map(violation -> getViolationDisplayMessage(violation))
               .collect(Collectors.joining("\n"));
-      throw new IllegalArgumentException(displayMessages);
+      throw new InsecureCertificateException(violations, displayMessages);
     }
   }
 
@@ -256,6 +256,16 @@ public class CertificateChecker {
      */
     public String getDisplayMessage(CertificateChecker certificateChecker) {
       return certificateChecker.getViolationDisplayMessage(this);
+    }
+  }
+
+  /** Exception to throw when a certificate has security violations. */
+  public static class InsecureCertificateException extends Exception {
+    ImmutableSet<CertificateViolation> violations;
+
+    InsecureCertificateException(ImmutableSet<CertificateViolation> violations, String message) {
+      super(message);
+      this.violations = violations;
     }
   }
 }
--- a/core/src/main/java/google/registry/tools/RegistryToolComponent.java
+++ b/core/src/main/java/google/registry/tools/RegistryToolComponent.java
@@ -168,6 +168,8 @@ interface RegistryToolComponent {
 
   void inject(ValidateEscrowDepositCommand command);
 
+  void inject(ValidateLoginCredentialsCommand command);
+
   void inject(WhoisQueryCommand command);
 
   AppEngineConnection appEngineConnection();
--- a/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java
+++ b/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java
@@ -25,12 +25,14 @@ import static java.nio.charset.StandardCharsets.US_ASCII;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import google.registry.flows.TlsCredentials;
+import google.registry.flows.certs.CertificateChecker;
 import google.registry.model.registrar.Registrar;
 import google.registry.tools.params.PathParameter;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Optional;
 import javax.annotation.Nullable;
+import javax.inject.Inject;
 
 /** A command to test registrar login credentials. */
 @Parameters(separators = " =", commandDescription = "Test registrar login credentials")
@@ -68,6 +70,8 @@ final class ValidateLoginCredentialsCommand implements CommandWithRemoteApi {
       description = "Client ip address to pretend to use")
   private String clientIpAddress = "10.0.0.1";
 
+  @Inject CertificateChecker certificateChecker;
+
   @Override
   public void run() throws Exception {
     checkArgument(
@@ -81,12 +85,14 @@ final class ValidateLoginCredentialsCommand implements CommandWithRemoteApi {
     Registrar registrar =
         checkArgumentPresent(
             Registrar.loadByClientId(clientId), "Registrar %s not found", clientId);
-    new TlsCredentials(
+    TlsCredentials credentials =
+        new TlsCredentials(
             true,
             Optional.ofNullable(clientCertificateHash),
             Optional.ofNullable(clientCertificate),
-            Optional.ofNullable(clientIpAddress))
-        .validate(registrar, password);
+            Optional.ofNullable(clientIpAddress));
+    credentials.certificateChecker = certificateChecker;
+    credentials.validate(registrar, password);
     checkState(
         registrar.isLive(), "Registrar %s has non-live state: %s", clientId, registrar.getState());
   }
--- a/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java
+++ b/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java
@@ -85,14 +85,13 @@ final class ValidateLoginCredentialsCommand implements CommandWithRemoteApi {
     Registrar registrar =
         checkArgumentPresent(
             Registrar.loadByClientId(clientId), "Registrar %s not found", clientId);
-    TlsCredentials credentials =
-        new TlsCredentials(
+    new TlsCredentials(
             true,
             Optional.ofNullable(clientCertificateHash),
             Optional.ofNullable(clientCertificate),
-            Optional.ofNullable(clientIpAddress));
-    credentials.certificateChecker = certificateChecker;
-    credentials.validate(registrar, password);
+            Optional.ofNullable(clientIpAddress),
+            certificateChecker)
+        .validate(registrar, password);
     checkState(
         registrar.isLive(), "Registrar %s has non-live state: %s", clientId, registrar.getState());
   }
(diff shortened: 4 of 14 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 5 files reviewed, 8 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 78 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_e-I49fauCNiB4DgJ:-MR_e-I49fauCNiB4DgK:bk5y7b0) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L78)):*
> ```Java
>   }
> 
>   @Inject public CertificateChecker certificateChecker;
> ```

This definitely shouldn't be public.

Also, you're mixing field injection with constructor injection here, which is a no-no (and which we'd have a presubmit to guard against if said presubmit were easy to write, but it's not).

Add the `CertificateChecker` as a parameter to the `@Inject`ed constructor instead, and make the field private like all the others.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 159 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_fEBIEexTR7XJPxT9:-MR_fEBIEexTR7XJPxTA:bz3k9xt) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L159)):*
> ```Java
>         try {
>           certificateChecker.validateCertificate(clientCertificate.get());
>         } catch (Exception e) {
> ```

This is an overly broad `catch` statement that makes it difficult to reason about what's actually going on here.

Also, looking into how the `validateCertificate()` method works, I think it needs a bit of a reworking.  Taking a bunch of string messages and concatenating them into a generic `IllegalArgumentException` to throw is not ideal.  Given that the entire point of the `validateCertificate()` method is to verify a certificate, and that the desired return value to check is whether or not it throws an Exception, it absolutely makes sense to used a checked (not runtime) exception for that purpose to enforce that the call-site is actually inspecting the necessary returned "value".  And we should be using a custom Exception class for that purpose rather than a generic one, so that the `catch` statement at callsites like here can be specific.  And now that we'll be using a custom Exception type, we can actually have an `ImmutableSet<String>` field on it to store the individual errors in a more accessible way than merely concatenating them all into a newline-delimited string (which is fine to continue doing for the message).
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 163 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_ewUL0Vx2f3_Yh1Wv:-MR_ewUL0Vx2f3_Yh1Ww:b-nt3a3u) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L163)):*
> ```Java
>           if (RegistryEnvironment.get().equals(RegistryEnvironment.UNITTEST)
>               || RegistryEnvironment.get().equals(RegistryEnvironment.SANDBOX)) {
>             throw new CertificateContainsSecurityViolationsException(e.getMessage());
> ```

This is an antipattern.  You should be wrapping Exceptions as inner exceptions by using the Exception constructor that takes the inner exception as an additional argument.  Don't just take the message from another Exception (while discarding all of its other important data, like its stack trace) and rethrow it.  This makes this harder to debug because it's throwing away all the information associated with the actual real Exception.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 165 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_gC36DkXaO9qJ8COT:-MR_gC372Xz5ZBnOhZK9:b-hamd7u) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L165)):*
> ```Java
>             throw new CertificateContainsSecurityViolationsException(e.getMessage());
>           }
>           if (e instanceof IllegalArgumentException) {
> ```

Big code smell here.  This should be handled in the type in the `catch` statement except.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 167 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_gQk9DkWkCt_zNZdX:-MR_gQk9DkWkCt_zNZdY:b-k3zohl) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L167)):*
> ```Java
>           if (e instanceof IllegalArgumentException) {
>             logger.atWarning().log(
>                 "Registrar certificate used for %s does not meet certificate requirements: %s",
> ```

Log statements and periods again :/
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 238 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_effMA_Rx0UMu1yD7:-MR_effMA_Rx0UMu1yD8:bo24cia) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L238)):*
> ```Java
>   }
> 
>   /** Registrar certificate contains security violations. */
> ```

I don't think this Javadoc is actually correct, as the message for the exception is passed in and thus can be anything, not necessarily `Registrar certificate contains security violations`.  Where would that message even come from anyway?  All I see is some inner Exception's message being rewrapped?
___
*[core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java, line 73 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_dqFUCUR_H4xiTRD0:-MR_dqFUCUR_H4xiTRD1:b-41wa54) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java#L73)):*
> ```Java
>   private String clientIpAddress = "10.0.0.1";
> 
>   @Inject CertificateChecker certificateChecker;
> ```

Have you tested this change by running it locally?  I don't see how it could work without listing this command in `RegistryToolComponent`.
___
*[core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_e4e64ospLueVmLi0:-MR_e4e64ospLueVmLi1:b4pjtnl) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java#L94)):*
> ```Java
>             Optional.ofNullable(clientCertificate),
>             Optional.ofNullable(clientIpAddress));
>     credentials.certificateChecker = certificateChecker;
> ```

Calling a constructor, then immediately overwriting a field, is a code smell.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 78 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_e-I49fauCNiB4DgJ:-MRaFGluD606MBbUxAvE:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L78)):*

This definitely shouldn't be public.

Also, you're mixing field injection with constructor injection here, which is a no-no (and which we'd have a presubmit to guard against if said presubmit were easy to write, but it's not).

Add the `CertificateChecker` as a parameter to the `@Inject`ed constructor instead, and make the field private like all the others.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 159 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_fEBIEexTR7XJPxT9:-MRaUpVqAwdfhq9jcnlA:b4pzqf6) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L159)):*

This is an overly broad `catch` statement that makes it difficult to reason about what's actually going on here.

Also, looking into how the `validateCertificate()` method works, I think it needs a bit of a reworking.  Taking a bunch of string messages and concatenating them into a generic `IllegalArgumentException` to throw is not ideal.  Given that the entire point of the `validateCertificate()` method is to verify a certificate, and that the desired return value to check is whether or not it throws an Exception, it absolutely makes sense to used a checked (not runtime) exception for that purpose to enforce that the call-site is actually inspecting the necessary returned "value".  And we should be using a custom Exception class for that purpose rather than a generic one, so that the `catch` statement at callsites like here can be specific.  And now that we'll be using a custom Exception type, we can actually have an `ImmutableSet<String>` field on it to store the individual errors in a more accessible way than merely concatenating them all into a newline-delimited string (which is fine to continue doing for the message).
</blockquote></details>

Added an InsecureCertificateException
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 163 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_ewUL0Vx2f3_Yh1Wv:-MRaQ3D8-kfS8F4vPmoa:b-lo08a) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L163)):*

This is an antipattern.  You should be wrapping Exceptions as inner exceptions by using the Exception constructor that takes the inner exception as an additional argument.  Don't just take the message from another Exception (while discarding all of its other important data, like its stack trace) and rethrow it.  This makes this harder to debug because it's throwing away all the information associated with the actual real Exception.
</blockquote></details>

ok, I'm not sure if how I did this correctly saves all of the inner exception data since CertificateContainsSecurityViolationsException is not a runtimeException, but an EPPException where the message is used to create the EPP result.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 165 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_gC36DkXaO9qJ8COT:-MRaHVivCZBq_eZxVzHx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L165)):*

Big code smell here.  This should be handled in the type in the `catch` statement except.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 167 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_gQk9DkWkCt_zNZdX:-MRaGWXAC55YAec2C0S8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L167)):*

Log statements and periods again :/
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 238 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_effMA_Rx0UMu1yD7:-MRaaXyN1hWfPnT0WV2Y:bzeaap2) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L238)):*

I don't think this Javadoc is actually correct, as the message for the exception is passed in and thus can be anything, not necessarily `Registrar certificate contains security violations`.  Where would that message even come from anyway?  All I see is some inner Exception's message being rewrapped?
</blockquote></details>

I think this should be fixed now that I added InsecureCertificateException
___
*[core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_e4e64ospLueVmLi0:-MRaFuxx5I3d_YQppUa5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/tools/ValidateLoginCredentialsCommand.java#L94)):*

Calling a constructor, then immediately overwriting a field, is a code smell.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 238 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_effMA_Rx0UMu1yD7:-MRay68i7zulyGm1XmjR:b-gr8w4x) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L238)):*

I think this should be fixed now that I added InsecureCertificateException
</blockquote></details>

Is anything anywhere actually ever writing the string `Registrar certificate contains security violations`?

Once we start throwing these in EPP flows, what does the EPP response actually look like?  As of this PR the new behavior will be happening in sandbox; where are the tests with EPP XMLs showing what happens when a certificate doesn't meet the security requirements?  (This would go a long way towards alleviating fears we might have of things not working for the deployment on Tuesday).  Note that you can set the Nomulus environment for a specific test, i.e. to make it think it's running in sandbox.

The larger point I'm making is that, without exception, the Javadoc string on an `EppException` is exactly the error message that would actually be visible in EPP.  But I'm not seeing anywhere in this code that would actually put this string into the EPP.  See above.
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 234 at r2](https://reviewable.io/reviews/google/nomulus/936#-MRaz_VX5HNv9fUSuX76:-MRaz_VX5HNv9fUSuX77:b-lmp25u) ([raw file](https://github.com/google/nomulus/blob/9f399f83c9969332bc5245bb0e1e99d03938eb7e/core/src/main/java/google/registry/flows/TlsCredentials.java#L234)):*
> ```Java
>   public static class BadRegistrarCertificateException extends AuthenticationErrorException {
>     BadRegistrarCertificateException() {
>       super("Registrar certificate does not match stored certificate");
> ```

Note how this String matches the Javadoc above?
___
*[core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java, line 396 at r2](https://reviewable.io/reviews/google/nomulus/936#-MRayMRX1fpc-xmpbDxh:-MRayMRX1fpc-xmpbDxi:b4uit1p) ([raw file](https://github.com/google/nomulus/blob/9f399f83c9969332bc5245bb0e1e99d03938eb7e/core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java#L396)):*
> ```Java
>             "ERROR",
>             "message",
>             "google.registry.flows.certs.CertificateChecker$InsecureCertificateException:"
> ```

Why is this in here?  This seems like something we shouldn't be exposing to the end-user.  The previous message was good.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 238 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_effMA_Rx0UMu1yD7:-MRb-DNZEtwnDfdzl6P5:b-v2hvyk) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L238)):*

Is anything anywhere actually ever writing the string `Registrar certificate contains security violations`?

Once we start throwing these in EPP flows, what does the EPP response actually look like?  As of this PR the new behavior will be happening in sandbox; where are the tests with EPP XMLs showing what happens when a certificate doesn't meet the security requirements?  (This would go a long way towards alleviating fears we might have of things not working for the deployment on Tuesday).  Note that you can set the Nomulus environment for a specific test, i.e. to make it think it's running in sandbox.

The larger point I'm making is that, without exception, the Javadoc string on an `EppException` is exactly the error message that would actually be visible in EPP.  But I'm not seeing anywhere in this code that would actually put this string into the EPP.  See above.
</blockquote></details>

See `EppLoginTlsTest` -- there needs to be more tests in there, that emulate sandbox for now (but that in the near future don't), that test the new security validation failure modes.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 238 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_effMA_Rx0UMu1yD7:-MRbGSP-8WUpYi3JfEoJ:b7upfel) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L238)):*

Is the last test in EppLoginTlsTest correctly doing what you're describing?

The message it is currently returning is "Certificate validity period is too long;....

Right now this message is able to detail in the epp response exactly why the certificate is failing. I could add the line "Registrar certificate contains security violations." to the front of that message. Or are you suggesting that I should just return that line in the message instead of listing the violations?
</blockquote></details>

Added another test to `EppLoginTlsTest` to test epp response for certificate with multiple violations
___
*[core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java, line 396 at r2](https://reviewable.io/reviews/google/nomulus/936#-MRayMRX1fpc-xmpbDxh:-MRb6ZKJAEWtbQVxAu6O:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9f399f83c9969332bc5245bb0e1e99d03938eb7e/core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java#L396)):*

Why is this in here?  This seems like something we shouldn't be exposing to the end-user.  The previous message was good.
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 238 at r1](https://reviewable.io/reviews/google/nomulus/936#-MR_effMA_Rx0UMu1yD7:-MReuyEX87UGAOeb4AWg:b5883p) ([raw file](https://github.com/google/nomulus/blob/d1220d9b14d4111b36cb872075ac44cf234cfee9/core/src/main/java/google/registry/flows/TlsCredentials.java#L238)):*

Added another test to `EppLoginTlsTest` to test epp response for certificate with multiple violations
</blockquote></details>

Yes, those tests are exactly what I'm looking for.  Thanks.  And as expected, they revealed that we weren't sending a fully proper error message back through EPP.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 194 at r3](https://reviewable.io/reviews/google/nomulus/936#-MRez-Lp4pJKjImGtxoU:-MRez-Lp4pJKjImGtxoV:b-42rk7a) ([raw file](https://github.com/google/nomulus/blob/b26b12032c87ee4321edfdb54ff059597460472b/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L194)):*
> ```Java
>             ImmutableMap.of("CODE", "2200", "MSG", "Registrar certificate is not configured"));
>   }
> ```

There needs to be a test that's overridden to the `PRODUCTION` environment that verifies that this enforcement isn't yet throwing actual exceptions and causing EPP failures (but that it is logged).  I think that's the last missing test of important functionality.

Then obviously this test will be removed in the coming weeks once we are enforcing the requirements in production for real.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 194 at r3](https://reviewable.io/reviews/google/nomulus/936#-MRez-Lp4pJKjImGtxoU:-MRfH4Bw3B3HYUTtTFgW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b26b12032c87ee4321edfdb54ff059597460472b/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L194)):*

There needs to be a test that's overridden to the `PRODUCTION` environment that verifies that this enforcement isn't yet throwing actual exceptions and causing EPP failures (but that it is logged).  I think that's the last missing test of important functionality.

Then obviously this test will be removed in the coming weeks once we are enforcing the requirements in production for real.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 273 at r4](https://reviewable.io/reviews/google/nomulus/936#-MRfO9163uzU7CUlcA5W:-MRfO9163uzU7CUlcA5X:b-f2veuh) ([raw file](https://github.com/google/nomulus/blob/3a760529f9d57badafcb5ce93f6f8f246cc972bc/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L273)):*
> ```Java
>     // Even though the certificate contains security violations, the login will succeed in
>     // production
>     assertThatLoginSucceeds("NewRegistrar", "foo-BAR2");
> ```

Needs verification that the right thing is logged.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 194 at r3](https://reviewable.io/reviews/google/nomulus/936#-MRez-Lp4pJKjImGtxoU:-MRfOFS6-3HAmcahw2Zx:bhugc7p) ([raw file](https://github.com/google/nomulus/blob/b26b12032c87ee4321edfdb54ff059597460472b/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L194)):*

Done.
</blockquote></details>

I'm not seeing the logging verification.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 273 at r4](https://reviewable.io/reviews/google/nomulus/936#-MRfO9163uzU7CUlcA5W:-MRfaPbO3QbHFLF0CHuZ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3a760529f9d57badafcb5ce93f6f8f246cc972bc/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L273)):*

Needs verification that the right thing is logged.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 67 at r5](https://reviewable.io/reviews/google/nomulus/936#-MRffEpfCmbRkvKMAPXF:-MRffEpg3NBmbXSIqJR2:bq2gk0b) ([raw file](https://github.com/google/nomulus/blob/f146a667bc1c535993cdf9501e8726a86ed6b0c2/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L67)):*
> ```Java
> 
>   private final Logger loggerToIntercept =
>       Logger.getLogger(TlsCredentials.class.getCanonicalName());
> ```

I think you're missing some plumbing here in connecting this to the `handler`?  See e.g. https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/flows/FlowRunnerTest.java

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 67 at r5](https://reviewable.io/reviews/google/nomulus/936#-MRffEpfCmbRkvKMAPXF:-MRffUDe2XEvuN2cPtHm:b-pr0ss3) ([raw file](https://github.com/google/nomulus/blob/f146a667bc1c535993cdf9501e8726a86ed6b0c2/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L67)):*

I think you're missing some plumbing here in connecting this to the `handler`?  See e.g. https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/flows/FlowRunnerTest.java
</blockquote></details>

Meant to link to this specific line: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/flows/FlowRunnerTest.java;l=66

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 67 at r5](https://reviewable.io/reviews/google/nomulus/936#-MRffEpfCmbRkvKMAPXF:-MRffZVmDozBLGmCilVM:bjkaldy) ([raw file](https://github.com/google/nomulus/blob/f146a667bc1c535993cdf9501e8726a86ed6b0c2/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L67)):*

Meant to link to this specific line: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/flows/FlowRunnerTest.java;l=66
</blockquote></details>

Ohhh never mind, I see you did it individually at the method level.  Probably better to do it in the `beforeEach()` method so that if anyone else needs to make assertions about logging in tests, then this will be plumbed up correctly.  As-is it's a bit of a trap.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936#-:-MRffo7wC_vy6-_SZD8l:bnfp4nl)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/936)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 67 at r5](https://reviewable.io/reviews/google/nomulus/936#-MRffEpfCmbRkvKMAPXF:-MRffr-T8doThtObL0rO:b-ca9hh) ([raw file](https://github.com/google/nomulus/blob/f146a667bc1c535993cdf9501e8726a86ed6b0c2/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L67)):*

Ohhh never mind, I see you did it individually at the method level.  Probably better to do it in the `beforeEach()` method so that if anyone else needs to make assertions about logging in tests, then this will be plumbed up correctly.  As-is it's a bit of a trap.
</blockquote></details>

(And that's the way we do it in other tests, so it should be the same here for consistency.)

PR #937, opened 2021-01-21
Convert ofy() to tm() for all contact transfer flows

hstonec:
There are a few changes in this PR that help convert ofy() calls to tm() for contact transfer flows:

1. In `TransferData`, renamed `gainingTransferPollMessageId` and `losingTransferPollMessageId` to `pollMessageId1` and `pollMessageId2`.
  * `gainingTransferPollMessageId` and `losingTransferPollMessageId` were added for Cloud SQL to replace its counterpart in `serverApproveEntities` used by Datastore. However, when we reconstruct the `serverApproveEntities` from Datastore, we cannot distinguish whether a poll message VKey is a gaining poll message or losing poll message based on the information in the vkey alone. Also, we don't make this distinction for Datastore because the only way we use `serverApproveEntities` is to delete all the entities in the set. So, changing the name to `pollMessageId${num}` is probably the easiest way we can do for the conversion.

2. In `TransferData`, added `repoId` and `historyEntryId` fields.
  * These two fields are used to help reconstruct the symmetric VKey in `serverApproveEntities` when the entity is loaded from Cloud SQL. `serverApproveEntites` contain VKey for PollMessage and BillingEvent, and they all have 3-tier OfyKey, i.e. they have a HistoryEntry key as its parent and a EPP resource key as its grandparent. In addition, all keys in `serverApproveEntites` share the same HistoryEntry key and EPP resource key. So, we just need to add two more fields here to avoid duplicate table columns.

3. Added a few callback functions in `PollMessage.OneTime` class to reconstruct `ResponseData` when the entity is loaded from Cloud SQL.
  * Similar to the above changes, we also added a few fields in `PollMessage.OneTime` to store response data for Cloud SQL, as the existing fields are dedicated for Datastore. We added those fields a while ago though don't actually use them until this PR(as some tests failed when running against Cloud SQL).

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/937)


gbrodman (commented):
It's crazy that we have never (?) had a way of distinguishing the gaining ID from the losing ID for poll messages (#1 above).  Reviewing now. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/937#-:-MRaFKc4-YrC2JS9KgHM:bus13p0)* status: 0 of 20 files reviewed, all discussions resolved

hstonec (commented):
Thanks, take your time.

Yeah, I do know which is which when constructing the `TransferData`(like [this](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/contact/ContactTransferRequestFlow.java;l=135;drc=6794c6fbd759343d12a36e876af40b15f8548de5;bpv=0;bpt=1)), however, when the entity is loaded from Datastore, we lose that information because the type of `serverApproveEntities` is `Set<VKey<? extends TransferServerApproveEntity>>` which is just a set of `VKey`. So, there is no way to distinguish them.

Fortunately, we actually don't need to do so because they only way we use `serverApproveEntities` is just to delete all of the underlying entities(check the usage of [getServerApproveEntities](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/transfer/TransferData.java;l=88;drc=6794c6fbd759343d12a36e876af40b15f8548de5;bpv=1;bpt=1)).

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/937#-:-MRaGSyi00fF7X55j_UQ:b-f90en)* status: 0 of 20 files reviewed, all discussions resolved

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/937)* status: all files reviewed, 8 unresolved discussions (waiting on @hstonec)

___
*[core/src/main/java/google/registry/model/poll/PollMessage.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaUKNPEt7brprpyBJJ:-MRaUKNPEt7brprpyBJK:bgmzk16) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/poll/PollMessage.java#L385)):*
> ```Java
>     void onLoad() {
>       super.onLoad();
>       if (contactPendingActionNotificationResponses != null) {
> ```

any chance it could be not-null but empty (causing the next line to fail)? 

(same for the next if statement)
___
*[core/src/main/java/google/registry/model/transfer/ContactTransferData.java, line 17 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaQwDCESjU53Syq7K-:-MRaQwDD6_i7vISDEUId:be90z3) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/ContactTransferData.java#L17)):*
> ```Java
> package google.registry.model.transfer;
> ```

nit: i think this newline is unnecessary
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 88 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaSt167Bu0Gz6HxAuY:-MRaSt167Bu0Gz6HxAuZ:bxdityw) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L88)):*
> ```Java
> 
>   @Ignore
>   @Column(name = "transfer_poll_message_id_1")
> ```

somewhat nitty (and i'm not sure if this is the right place to do this), but can we document somewhere the reason for having these just have IDs 1 and 2?  (the reason being that we can't differentiate them in Datastore since they're just a set of keys) 
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 176 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTNQoA3EGPc-tLHVP:-MRaTNQoA3EGPc-tLHVQ:b-v7zazq) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L176)):*
> ```Java
>       Set<VKey<? extends TransferServerApproveEntity>> serverApproveEntities,
>       TransferData transferData) {
>     if (serverApproveEntities == null || serverApproveEntities.isEmpty()) {
> ```

CollectionUtils.isNullOrEmpty? 
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 183 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTS7S81eAIaj1NrF4:-MRaTS7S81eAIaj1NrF5:b-3wa24e) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L183)):*
> ```Java
>       return;
>     }
>     // Element in serverApproveEntities should have the exact same Key<HistoryEntry> as its parent.
> ```

grammar nit: perhaps `Each element in...`
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 190 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTcH63Ax_qWvgS-VO:-MRaTcH63Ax_qWvgS-VP:b-4eoshc) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L190)):*
> ```Java
> 
>     ImmutableList<Long> sortedPollMessageIds = getSortedPollMessageIds(serverApproveEntities);
>     if (sortedPollMessageIds.size() > 1) {
> ```

Should this be `> 0`? 
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTjABCNhuIxZWSsWC:-MRaTjABCNhuIxZWSsWD:bzrl01) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L198)):*
> ```Java
>   }
> 
>   /**
> ```

it might be possible to do this more simply with something like 

`nullToEmpty(serverApproveEntities).stream().filter(is poll message).map(get sql key).sorted().collect(toImmutableSet())`
___
*[core/src/test/java/google/registry/flows/contact/ContactTransferRequestFlowTest.java, line 107 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaRGJ_6PUjFwDoKiHx:-MRaRGJ_6PUjFwDoKiHy:b-wokf3x) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/test/java/google/registry/flows/contact/ContactTransferRequestFlowTest.java#L107)):*
> ```Java
>                 // do this comparison separately below.
>                 .setServerApproveEntities(
>                     contact.getTransferData().getServerApproveEntities().isEmpty()
> ```

CollectionUtils.forceEmptyToNull ?

hstonec (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/937)* status: all files reviewed, 7 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/poll/PollMessage.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaUKNPEt7brprpyBJJ:-MRaw2jd-RseUcLEDdY8:bp04y4k) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/poll/PollMessage.java#L385)):*

any chance it could be not-null but empty (causing the next line to fail)? 

(same for the next if statement)
</blockquote></details>

Actually, we call `forceEmptyToNull` before this gets saved so there is no chance it can be empty. However, I think people may have the same question when reviewing this code later, so I still make the change.
___
*[core/src/main/java/google/registry/model/transfer/ContactTransferData.java, line 17 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaQwDCESjU53Syq7K-:-MRawxWt8XwRKR4UtiG7:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/ContactTransferData.java#L17)):*

nit: i think this newline is unnecessary
</blockquote></details>

Done
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 88 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaSt167Bu0Gz6HxAuY:-MRb09aV39cPOlvvSQKb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L88)):*

somewhat nitty (and i'm not sure if this is the right place to do this), but can we document somewhere the reason for having these just have IDs 1 and 2?  (the reason being that we can't differentiate them in Datastore since they're just a set of keys) 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 176 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTNQoA3EGPc-tLHVP:-MRb0G_01srEGTXNwwoJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L176)):*

CollectionUtils.isNullOrEmpty? 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 183 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTS7S81eAIaj1NrF4:-MRb0LB-CudOC2ztZuTI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L183)):*

grammar nit: perhaps `Each element in...`
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 190 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTcH63Ax_qWvgS-VO:-MRb0Zxn54e99hxoxTFn:b1jxzi3) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L190)):*

Should this be `> 0`? 
</blockquote></details>

good catch, done.
___
*[core/src/main/java/google/registry/model/transfer/TransferData.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaTjABCNhuIxZWSsWC:-MRb1Enn10xPP3-meV2A:b9ow35o) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/main/java/google/registry/model/transfer/TransferData.java#L198)):*

it might be possible to do this more simply with something like 

`nullToEmpty(serverApproveEntities).stream().filter(is poll message).map(get sql key).sorted().collect(toImmutableSet())`
</blockquote></details>

good call, done.
___
*[core/src/test/java/google/registry/flows/contact/ContactTransferRequestFlowTest.java, line 107 at r1](https://reviewable.io/reviews/google/nomulus/937#-MRaRGJ_6PUjFwDoKiHx:-MRb1zC3B0NSrN6cEOns:b-896fix) ([raw file](https://github.com/google/nomulus/blob/499282865c01f78dee269011a12a0328d950abf7/core/src/test/java/google/registry/flows/contact/ContactTransferRequestFlowTest.java#L107)):*

CollectionUtils.forceEmptyToNull ?
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/937)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #942, opened 2021-01-25
Make cross database comparison recursive

mindhog:
Cross-database comparison was previously just a shallow check: fields marked
with DoNotCompare on nested objects were still compared.  This causes problems
in some cases where there are nested immutable objects.

This change introduces recursive comparison.  It also provides a
hasCorrectHashCode() method that verifies that an object has not been mutated
since the hash code was calculated, which has been a problem in certain cases.

Finally, this also fixes the problem of objects that are mutated in multiple
transactions: we were previously comparing against the value in datastore, but
this doesn't work in these cases because the object in datastore may have
changed since the transaction that we are verifying.  Instead, check against
the value that we would have persisted in the original transaction.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/942)

### DIFF
--- a/core/src/test/java/google/registry/model/ImmutableObjectSubject.java
+++ b/core/src/test/java/google/registry/model/ImmutableObjectSubject.java
@@ -19,6 +19,7 @@ import static com.google.common.truth.Truth.assertThat;
 import static google.registry.testing.truth.TruthUtils.assertNullnessParity;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.truth.Correspondence;
 import com.google.common.truth.Correspondence.BinaryPredicate;
@@ -62,21 +63,7 @@ public final class ImmutableObjectSubject extends Subject {
   }
 
   private static String formatItems(String prefix, Iterator<?> iter) {
-    StringBuilder message = new StringBuilder(prefix);
-    boolean first = true;
-    while (iter.hasNext()) {
-      if (!first) {
-        message.append(", ");
-      }
-      message.append(iter.next());
-      first = false;
-    }
-    return message.toString();
-  }
-
-  /** Null-safe function to get the class name. */
-  private static String getClassName(Object object) {
-    return object != null ? object.getClass().getName() : "null";
+    return prefix + Joiner.on(", ").join(iter);
   }
 
   /**
@@ -86,121 +73,131 @@ public final class ImmutableObjectSubject extends Subject {
    * <p>This is used to verify that entities stored in both cloud SQL and Datastore are identical.
    */
   public void isEqualAcrossDatabases(@Nullable ImmutableObject expected) {
-    checkObjectAcrossDatabases(actual, expected, getClassName(actual));
+    checkObjectAcrossDatabases(
+        actual, expected, actual != null ? actual.getClass().getName() : "null");
   }
 
   // The following "check" methods implement a recursive check of immutable object equality across
-  // databases.  All of them function in both assertive and predicate modes: if "{@code path}" is
-  // provided (not null) then they throw {@link AssertionError}'s with detailed error messages.  If
+  // databases.  All of them function in both assertive and predicate modes: if "path" is
+  // provided (not null) then they throw AssertionError's with detailed error messages.  If
   // it is null, they return true for equal objects and false for inequal ones.
+  //
+  // The reason for this dual-mode behavior is that all of these methods can either be used in the
+  // context of a test assertion (in which case we want a detailed error message describing exactly
+  // the location in a complex object where a difference was discovered) or in the context of a
+  // membership check in a set (in which case we don't care about the specific location of the first
+  // difference, we just want to be able to determine if the object "is equal to" another object as
+  // efficiently as possible -- see checkSetAcrossDatabase()).
 
   @VisibleForTesting
   static boolean checkObjectAcrossDatabases(
       @Nullable Object actual, @Nullable Object expected, @Nullable String path) {
-    if (actual != expected && (actual == null || !actual.equals(expected))) {
-      // They're different, do a more detailed comparison.
+    if (Objects.equals(actual, expected)) {
+      return true;
+    }
 
-      // Check for null first
-      if (actual == null && expected != null) {
-        if (path != null) {
-          throw new AssertionError("At " + path + ": expected " + expected + "got null.");
-        } else {
-          return false;
-        }
-      } else if (actual != null && expected == null) {
-        if (path != null) {
-          throw new AssertionError("At " + path + ": expected null, got " + actual);
-        } else {
-          return false;
-        }
+    // They're different, do a more detailed comparison.
 
-        // For immutable objects, we have to recurse since the contained
-        // object could have DoNotCompare fields, too.
-      } else if (expected instanceof ImmutableObject) {
-        // We only verify that actual is an ImmutableObject so we get a good error message instead
-        // of a context-less ClassCastException.
-        if (!(actual instanceof ImmutableObject)) {
-          if (path != null) {
-            throw new AssertionError("At " + path + ": " + actual + " is not an immutable object.");
-          } else {
-            return false;
-          }
-        }
-
-        if (!checkImmutableAcrossDatabases(
-            (ImmutableObject) actual, (ImmutableObject) expected, path)) {
-          return false;
-        }
-      } else if (expected instanceof Map) {
-        if (!(actual instanceof Map)) {
-          if (path != null) {
-            throw new AssertionError("At " + path + ": " + actual + " is not a Map.");
-          } else {
-            return false;
-          }
-        }
-
-        // We could do better performance-wise by assuming that keys can be compared across
-        // databases using .equals() -- then we could just iterate over the entries and verify that
-        // the key has the same value in the other set.  However, there is currently no way for us
-        // to guard against the invalidation of this assumption, and it's less code to simply reuse
-        // the set comparison.  As long as the performance is adequate in the context of a test, it
-        // doesn't seem like a good idea to try to optimize this.
-        if (!checkSetAcrossDatabases(
-            ((Map<?, ?>) actual).entrySet(), ((Map<?, ?>) expected).entrySet(), path, "Map")) {
-          return false;
-        }
-      } else if (expected instanceof Set) {
-        if (!(actual instanceof Set)) {
-          if (path != null) {
-            throw new AssertionError("At " + path + ": " + actual + " is not a Set.");
-          } else {
-            return false;
-          }
-        }
-
-        if (!checkSetAcrossDatabases((Set<?>) actual, (Set<?>) expected, path, "Set")) {
-          return false;
-        }
-      } else if (expected instanceof Collection) {
-        if (!(actual instanceof Collection)) {
-          if (path != null) {
-            throw new AssertionError("At " + path + ": " + actual + " is not a Collection.");
-          } else {
-            return false;
-          }
-        }
-
-        if (!checkListAcrossDatabases((Collection<?>) actual, (Collection<?>) expected, path)) {
-          return false;
-        }
-
-        // Give Map.Entry special treatment to facilitate the use of Set comparison for verification
-        // of Map.
-      } else if (expected instanceof Map.Entry) {
-        if (!(actual instanceof Map.Entry)) {
-          if (path != null) {
-            throw new AssertionError("At " + path + ": " + actual + " is not a Map.Entry.");
-          } else {
-            return false;
-          }
-        }
-
-        if (!checkObjectAcrossDatabases(
-                ((Map.Entry<?, ?>) actual).getKey(), ((Map.Entry<?, ?>) expected).getKey(), path)
-            || !checkObjectAcrossDatabases(
-                ((Map.Entry<?, ?>) actual).getValue(),
-                ((Map.Entry<?, ?>) expected).getValue(),
-                path)) {
-          return false;
-        }
+    // Check for null first
+    if (actual == null && expected != null) {
+      if (path != null) {
+        throw new AssertionError("At " + path + ": expected " + expected + "got null.");
       } else {
-        if (!actual.equals(expected)) {
-          if (path != null) {
-            throw new AssertionError("At " + path + ": " + actual + " is not equal to " + expected);
-          } else {
-            return false;
-          }
+        return false;
+      }
+    } else if (actual != null && expected == null) {
+      if (path != null) {
+        throw new AssertionError("At " + path + ": expected null, got " + actual);
+      } else {
+        return false;
+      }
+
+      // For immutable objects, we have to recurse since the contained
+      // object could have DoNotCompare fields, too.
+    } else if (expected instanceof ImmutableObject) {
+      // We only verify that actual is an ImmutableObject so we get a good error message instead
+      // of a context-less ClassCastException.
+      if (!(actual instanceof ImmutableObject)) {
+        if (path != null) {
+          throw new AssertionError("At " + path + ": " + actual + " is not an immutable object.");
+        } else {
+          return false;
+        }
+      }
+
+      if (!checkImmutableAcrossDatabases(
+          (ImmutableObject) actual, (ImmutableObject) expected, path)) {
+        return false;
+      }
+    } else if (expected instanceof Map) {
+      if (!(actual instanceof Map)) {
+        if (path != null) {
+          throw new AssertionError("At " + path + ": " + actual + " is not a Map.");
+        } else {
+          return false;
+        }
+      }
+
+      // We could do better performance-wise by assuming that keys can be compared across
+      // databases using .equals() -- then we could just iterate over the entries and verify that
+      // the key has the same value in the other set.  However, there is currently no way for us
+      // to guard against the invalidation of this assumption, and it's less code to simply reuse
+      // the set comparison.  As long as the performance is adequate in the context of a test, it
+      // doesn't seem like a good idea to try to optimize this.
+      if (!checkSetAcrossDatabases(
+          ((Map<?, ?>) actual).entrySet(), ((Map<?, ?>) expected).entrySet(), path, "Map")) {
+        return false;
+      }
+    } else if (expected instanceof Set) {
+      if (!(actual instanceof Set)) {
+        if (path != null) {
+          throw new AssertionError("At " + path + ": " + actual + " is not a Set.");
+        } else {
+          return false;
+        }
+      }
+
+      if (!checkSetAcrossDatabases((Set<?>) actual, (Set<?>) expected, path, "Set")) {
+        return false;
+      }
+    } else if (expected instanceof Collection) {
+      if (!(actual instanceof Collection)) {
+        if (path != null) {
+          throw new AssertionError("At " + path + ": " + actual + " is not a Collection.");
+        } else {
+          return false;
+        }
+      }
+
+      if (!checkListAcrossDatabases((Collection<?>) actual, (Collection<?>) expected, path)) {
+        return false;
+      }
+
+      // Give Map.Entry special treatment to facilitate the use of Set comparison for verification
+      // of Map.
+    } else if (expected instanceof Map.Entry) {
+      if (!(actual instanceof Map.Entry)) {
+        if (path != null) {
+          throw new AssertionError("At " + path + ": " + actual + " is not a Map.Entry.");
+        } else {
+          return false;
+        }
+      }
+
+      if (!checkObjectAcrossDatabases(
+              ((Map.Entry<?, ?>) actual).getKey(), ((Map.Entry<?, ?>) expected).getKey(), path)
+          || !checkObjectAcrossDatabases(
+              ((Map.Entry<?, ?>) actual).getValue(),
+              ((Map.Entry<?, ?>) expected).getValue(),
+              path)) {
+        return false;
+      }
+    } else {
+      if (!actual.equals(expected)) {
+        if (path != null) {
+          throw new AssertionError("At " + path + ": " + actual + " is not equal to " + expected);
+        } else {
+          return false;
         }
       }
     }
--- a/core/src/test/java/google/registry/model/ImmutableObjectSubject.java
+++ b/core/src/test/java/google/registry/model/ImmutableObjectSubject.java
@@ -18,6 +18,7 @@ import static com.google.common.truth.Truth.assertAbout;
 import static com.google.common.truth.Truth.assertThat;
 import static google.registry.testing.truth.TruthUtils.assertNullnessParity;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.truth.Correspondence;
 import com.google.common.truth.Correspondence.BinaryPredicate;
@@ -26,6 +27,10 @@ import com.google.common.truth.SimpleSubjectBuilder;
 import com.google.common.truth.Subject;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -56,6 +61,24 @@ public final class ImmutableObjectSubject extends Subject {
     }
   }
 
+  private static String formatItems(String prefix, Iterator<?> iter) {
+    StringBuilder message = new StringBuilder(prefix);
+    boolean first = true;
+    while (iter.hasNext()) {
+      if (!first) {
+        message.append(", ");
+      }
+      message.append(iter.next());
+      first = false;
+    }
+    return message.toString();
+  }
+
+  /** Null-safe function to get the class name. */
+  private static String getClassName(Object object) {
+    return object != null ? object.getClass().getName() : "null";
+  }
+
   /**
    * Checks that {@code expected} has the same contents as {@code actual} except for fields that are
    * marked with {@link ImmutableObject.DoNotCompare}.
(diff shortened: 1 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @mindhog)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvgcp3-NtQCi5VdOsq:-MRvgcp3-NtQCi5VdOsr:bh20gn5) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L63)):*
> ```Java
>   }
> 
>   private static String formatItems(String prefix, Iterator<?> iter) {
> ```

We can use Joiner.on(", ").join(iter) instead of this method (or to simplify this method)
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvjBHm-9tF25sHptbf:-MRvjBHm-9tF25sHptbg:b-6zwtjy) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L76)):*
> ```Java
>   }
> 
>   /** Null-safe function to get the class name. */
> ```

This is only used in the one place below right? We can probably inline it 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 97 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvlqyM8LALLuquHAkM:-MRvlqyM8LALLuquHAkN:b-xkkcyj) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L97)):*
> ```Java
> 
>   @VisibleForTesting
>   static boolean checkObjectAcrossDatabases(
> ```

In general I don't see the need to mix a return value with a thrown exception in this function, and it makes the code significantly more complex to try to essentially return two different types of values. 

I think we can probably get rid of path then make this either return a direct boolean (pros: simpler, can assert that objects are either equal or nonequal) or throw an exception (pros: can potentially have a more detailed error message in failing tests). I'd lean toward the former I think because I believe usually failures with nonequal objects are going to require object inspection anyway even with a helpful error message. It also allows us to simplify the nesting of the function drastically.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvmKnm6zu-vojSSKfu:-MRvmKnm6zu-vojSSKfv:b-jy4equ) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L98)):*
> ```Java
>   @VisibleForTesting
>   static boolean checkObjectAcrossDatabases(
>       @Nullable Object actual, @Nullable Object expected, @Nullable String path) {
> ```

What's the point of the `path` variable? The tests don't really seem to display how it would be useful in practice, and if we're writing tests using this function aren't we going to generally know what objects are being compared anyway? 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 99 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvjS5l8kzcbadFVoSk:-MRvjS5l8kzcbadFVoSl:b-l904m5) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L99)):*
> ```Java
>   static boolean checkObjectAcrossDatabases(
>       @Nullable Object actual, @Nullable Object expected, @Nullable String path) {
>     if (actual != expected && (actual == null || !actual.equals(expected))) {
> ```

We can just use `if (!Objects.equals(actual, expected))` instead of this line; I think it's much simpler and it does pretty much the exact same code

In addition, there's a lot of nesting in this function and we can remove at least one layer of that by just having 

`if (Objects.equals(actual, expected) return true;` at the beginning (formatted correctly of course) 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvn7Is0AP_xG5-lrfp:-MRvn7Is0AP_xG5-lrfq:b-o7bn4t) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L142)):*
> ```Java
>         }
> 
>         // We could do better performance-wise by assuming that keys can be compared across
> ```

yeah i don't think it's even necessary to clarify any performance concerns -- with the overhead of setting up databases et al, I would be willing to be that this never matters in the slightest 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvnVwwAJJ366y1pyaM:-MRvnVwwAJJ366y1pyaN:bkn8w55) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L210)):*
> ```Java
>   }
> 
>   private static boolean checkSetAcrossDatabases(
> ```

This can probably be simplified somewhat using Stream::anyMatch, just asserting that for each expected elem, actual.stream().anyMatch(...) is true 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 272 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvo18_2JE29YPXXFB_:-MRvo18_2JE29YPXXFBa:brqkj42) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L272)):*
> ```Java
>     return true;
>   }
> ```

Is there no way to ask if two lists are equal with using a custom comparator?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvgcp3-NtQCi5VdOsq:-MRyj3e4BwoYNEXNk_L6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L63)):*

We can use Joiner.on(", ").join(iter) instead of this method (or to simplify this method)
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvjBHm-9tF25sHptbf:-MRyjLU-B5HKM9YxlR_N:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L76)):*

This is only used in the one place below right? We can probably inline it 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 97 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvlqyM8LALLuquHAkM:-MRyjXQa8B9vkDcuu6X6:but5k4m) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L97)):*

In general I don't see the need to mix a return value with a thrown exception in this function, and it makes the code significantly more complex to try to essentially return two different types of values. 

I think we can probably get rid of path then make this either return a direct boolean (pros: simpler, can assert that objects are either equal or nonequal) or throw an exception (pros: can potentially have a more detailed error message in failing tests). I'd lean toward the former I think because I believe usually failures with nonequal objects are going to require object inspection anyway even with a helpful error message. It also allows us to simplify the nesting of the function drastically.
</blockquote></details>

I've added a little bit more of an explanation to the comment string above the method set.  I'm not wild about this approach, but what it comes down to is that we want the location details in one context but not in the other.

An alternative to the return-vs-throw approach would be to always return an object that is essentially both a flag (corresponding to the returned boolean) and an optional error message.  We would then only throw the AssertionError in the top-level, isEqualAcrossDatabases() method.  We'd still use a null path as an indicator that we don't need to construct the message, thus saving the cost of string construction in the case where we don't need it.

I have to say, I do like that approach better than what I've implemented here.  It would remove a lot of complexity from both the code and the tests.  What do you think?
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvmKnm6zu-vojSSKfu:-MRyn0zk9BWP2_0nyDgN:b-aa2u0f) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L98)):*

What's the point of the `path` variable? The tests don't really seem to display how it would be useful in practice, and if we're writing tests using this function aren't we going to generally know what objects are being compared anyway? 
</blockquote></details>

Having dealt with these comparison failures a lot recently, I can say with certainty that trying to extract the specific differences between entities that fail comparison from two object dumps is tedious and time-consuming.  Having a full path pointing to the specific nested field that differs is _much_ easier (I dealt with this yesterday for the RenewFlowTest, it was easily half the work I'd done in the past).

You don't so much need to know the specific object that's failing, what is generally important is the _type_ of object that's failing, and that's part of the path.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 99 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvjS5l8kzcbadFVoSk:-MRyqe2b13Yf9oY_nnlc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L99)):*

We can just use `if (!Objects.equals(actual, expected))` instead of this line; I think it's much simpler and it does pretty much the exact same code

In addition, there's a lot of nesting in this function and we can remove at least one layer of that by just having 

`if (Objects.equals(actual, expected) return true;` at the beginning (formatted correctly of course) 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvn7Is0AP_xG5-lrfp:-MRyqfd84-WaNb7-csYg:b-bevc40) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L142)):*

yeah i don't think it's even necessary to clarify any performance concerns -- with the overhead of setting up databases et al, I would be willing to be that this never matters in the slightest 
</blockquote></details>

It's generally worth spelling out the reasoning for non-obvious decisions in the comments, as a later reader might well ask "why not just do X?"  It's difficult to say how much this comparison will affect performance, but some version of this check is likely to end up happening during the database migration (as part of database verification), so there's a case to be made for efficiency at least at the macroscopic level.

 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvnVwwAJJ366y1pyaM:-MRyudT69F4ydGxWnfc3:brmdjrg) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L210)):*

This can probably be simplified somewhat using Stream::anyMatch, just asserting that for each expected elem, actual.stream().anyMatch(...) is true 
</blockquote></details>

I don't think there's a way to do this efficiently using any of the stream methods.  It's not enough to know that a match exists in the other set, we have to be able to verify that only one match exists in the other set and be able to report on (at least one, and ideally all) of the differences.

We could use Stream.filter(...).collect(...), but I believe an algorithm based on that would be both less efficient and less readable. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 272 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvo18_2JE29YPXXFB_:-MRywEHZ5QcjhzqL8SIO:bpyi5sk) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L272)):*

Is there no way to ask if two lists are equal with using a custom comparator? 
</blockquote></details>

I don't think so?  There's a discussion on this [here](https://stackoverflow.com/questions/30106376/how-can-i-compare-two-array-lists-for-equality-with-a-custom-comparator) that basically suggests what I've implemented.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @mindhog)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 97 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvlqyM8LALLuquHAkM:-MS--R5D10nL1JJxntS0:bdnm68n) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L97)):*

I've added a little bit more of an explanation to the comment string above the method set.  I'm not wild about this approach, but what it comes down to is that we want the location details in one context but not in the other.

An alternative to the return-vs-throw approach would be to always return an object that is essentially both a flag (corresponding to the returned boolean) and an optional error message.  We would then only throw the AssertionError in the top-level, isEqualAcrossDatabases() method.  We'd still use a null path as an indicator that we don't need to construct the message, thus saving the cost of string construction in the case where we don't need it.

I have to say, I do like that approach better than what I've implemented here.  It would remove a lot of complexity from both the code and the tests.  What do you think?
</blockquote></details>

Yeah, this is the problem that Golang's dual-return structure is meant to solve, right? :) 

Yeah, we could return a structure or honestly we could really just return the error message string if one exists. This is actually the same structure that we use in parts of the EPP processing code, I believe (it's been a while but it rings a bell). 

Referencing the last sentence of your second paragraph, I don't see the harm in always constructing the string anyway, if it simplifies the code it's not going to be a hit performance-wise and it'll make things shorter probably. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvmKnm6zu-vojSSKfu:-MS--pfEEfAOPl2ADExT:b-t8we03) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L98)):*

Having dealt with these comparison failures a lot recently, I can say with certainty that trying to extract the specific differences between entities that fail comparison from two object dumps is tedious and time-consuming.  Having a full path pointing to the specific nested field that differs is _much_ easier (I dealt with this yesterday for the RenewFlowTest, it was easily half the work I'd done in the past).

You don't so much need to know the specific object that's failing, what is generally important is the _type_ of object that's failing, and that's part of the path.
</blockquote></details>

Do you have an example off the top of your head of an object that was failing? Was it just something with a lot of nested fields etc? 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvn7Is0AP_xG5-lrfp:-MRzoO1q43jOOk5P2io3:brksy7b) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L142)):*

It's generally worth spelling out the reasoning for non-obvious decisions in the comments, as a later reader might well ask "why not just do X?"  It's difficult to say how much this comparison will affect performance, but some version of this check is likely to end up happening during the database migration (as part of database verification), so there's a case to be made for efficiency at least at the macroscopic level.

 
</blockquote></details>

yeah it just seems like this comment is more focused on arguing against the code's efficiency rather than arguing in favor of the code's correctness or simplicity. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvnVwwAJJ366y1pyaM:-MS-1gyeBz2DvAcWRGRp:b-x5sz7y) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L210)):*

I don't think there's a way to do this efficiently using any of the stream methods.  It's not enough to know that a match exists in the other set, we have to be able to verify that only one match exists in the other set and be able to report on (at least one, and ideally all) of the differences.

We could use Stream.filter(...).collect(...), but I believe an algorithm based on that would be both less efficient and less readable. 
</blockquote></details>

Does this method verify that there's only one match in the corresponding other set? I'm not sure that it does. 

If we don't care about verifying that there's only one match in the corresponding other set, we could always just use Set.stream().anyMatch(...) twice -- one to verify that all elements in `actual` are in `expected` and one to verify that all elements in `expected` are in `actual`. We could check that the sizes are equal as well, which seems like it might be fewer lines of code /  branches than this? 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 272 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvo18_2JE29YPXXFB_:-MS--xkz0iTPAVLoOqZB:b-vhjudd) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L272)):*

I don't think so?  There's a discussion on this [here](https://stackoverflow.com/questions/30106376/how-can-i-compare-two-array-lists-for-equality-with-a-custom-comparator) that basically suggests what I've implemented.
</blockquote></details>

ugh yeah I found that yesterday when making this comment and I was afraid this would be the answer :( 

I think for this method it might just be easier to use indexes rather than iterators -- verify that the sizes are the same then for each index verify that actual[index] is equal (using the method) to expected[index]. We might not get as many details in the error message when the sizes aren't equal but that should hopefully be ok. Have you considered this / seen how it looks and behaves?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 97 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvlqyM8LALLuquHAkM:-MS-8xnZ5wP-75bzt6nw:b-8n6fts) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L97)):*

Yeah, this is the problem that Golang's dual-return structure is meant to solve, right? :) 

Yeah, we could return a structure or honestly we could really just return the error message string if one exists. This is actually the same structure that we use in parts of the EPP processing code, I believe (it's been a while but it rings a bell). 

Referencing the last sentence of your second paragraph, I don't see the harm in always constructing the string anyway, if it simplifies the code it's not going to be a hit performance-wise and it'll make things shorter probably. 
</blockquote></details>

PTAL, I've converted the code to consistently return a result object instead.

I made the path formatting optional.  We generally don't recommend sacrificing code clarity for micro-optimizations, but keeping this switch lets us cut out some pretty hefty code-paths with very little additional complexity.  The ComparisonResult abstraction isolates most of it.

I've also retained the "null path" tests, as some of them do exercise the abbreviated code-paths.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvmKnm6zu-vojSSKfu:-MS-7Yey5i4C4K89SM4T:b-4unbzb) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L98)):*

Do you have an example off the top of your head of an object that was failing? Was it just something with a lot of nested fields etc? 
</blockquote></details>

Here's the exception text for the problem I ran into with TransferData:

java.lang.AssertionError: At google.registry.model.domain.DomainBase.transferData.serverApproveAutorenewPollMessage.kind: class google.registry.model.poll.PollMessage$Autorenew is not equal to class google.registry.model.poll.PollMessage

So yes, three levels of nesting in this case.  The message points directly at the problem (class reconstitution in the VKey isn't dealing correctly with the different representation across DBs).
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvn7Is0AP_xG5-lrfp:-MS2lq__4RlIG6avwaQL:b-3di442) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L142)):*

yeah it just seems like this comment is more focused on arguing against the code's efficiency rather than arguing in favor of the code's correctness or simplicity. 
</blockquote></details>

Can you suggest an alternate wording for it?
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvnVwwAJJ366y1pyaM:-MS-NmOL32c-FhCeINa5:bcdhgi5) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L210)):*

Does this method verify that there's only one match in the corresponding other set? I'm not sure that it does. 

If we don't care about verifying that there's only one match in the corresponding other set, we could always just use Set.stream().anyMatch(...) twice -- one to verify that all elements in `actual` are in `expected` and one to verify that all elements in `expected` are in `actual`. We could check that the sizes are equal as well, which seems like it might be fewer lines of code /  branches than this? 
</blockquote></details>

I thought that it did verify, but after testing it found that it didn't.  It does now, so thanks for that :-)

We do care about this, it would be indicative of a flaw in our data model.  I do think you're right that it would simplify the code to do a double-pass through anyMatch(), though.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 272 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvo18_2JE29YPXXFB_:-MS-rJyR2gA39WKvHQNK:b-r798pi) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L272)):*

ugh yeah I found that yesterday when making this comment and I was afraid this would be the answer :( 

I think for this method it might just be easier to use indexes rather than iterators -- verify that the sizes are the same then for each index verify that actual[index] is equal (using the method) to expected[index]. We might not get as many details in the error message when the sizes aren't equal but that should hopefully be ok. Have you considered this / seen how it looks and behaves? 
</blockquote></details>

I can tell you just from looking at the output that it wouldn't be as useful without the missing/extra items.  You'd get "At Class.field: missing N items."  And then I'd swear and go add some extra code to print the missing/extra items so I could debug :-)

We can't currently use indexes because, strictly speaking, these are Collection's.  The assumption that collections that aren't sets or maps is kind of kludgy but there's no List interface and some of the existing ImmutableObject code makes a similar assumption.  I suppose we could convert them to arrays to use indexes, but that seems kind of ugly in its own right.

It's generally true that we can have less functionality with less code.  But do you find the current code difficult to follow?  It would be nice to be able to do this more succinctly, but if it's a choice between a better error message and less code I'm going to push hard for the former.

gbrodman (changes requested):
ugh, sometimes Reviewable is showing me pre-revision code but fortunately I have Github to double-check :) 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942#-:-MS3haA76PkYSTCoc-lg:bcrdehk)* status: 0 of 3 files reviewed, 11 unresolved discussions (waiting on @mindhog)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 97 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvlqyM8LALLuquHAkM:-MS3l3jNB_PDKg-_-PNG:bsfgoxd) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L97)):*

PTAL, I've converted the code to consistently return a result object instead.

I made the path formatting optional.  We generally don't recommend sacrificing code clarity for micro-optimizations, but keeping this switch lets us cut out some pretty hefty code-paths with very little additional complexity.  The ComparisonResult abstraction isolates most of it.

I've also retained the "null path" tests, as some of them do exercise the abbreviated code-paths.

</blockquote></details>

Yeah, I think this is a ton better -- I hope you agree. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvmKnm6zu-vojSSKfu:-MS3bhqw6Aix5akEXUN3:bo7be2r) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L98)):*

Here's the exception text for the problem I ran into with TransferData:

java.lang.AssertionError: At google.registry.model.domain.DomainBase.transferData.serverApproveAutorenewPollMessage.kind: class google.registry.model.poll.PollMessage$Autorenew is not equal to class google.registry.model.poll.PollMessage

So yes, three levels of nesting in this case.  The message points directly at the problem (class reconstitution in the VKey isn't dealing correctly with the different representation across DBs).
</blockquote></details>

Yeah, I see what you mean now. That makes sense, thanks for the detail and info. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/942#-MS3g9et5z_THgtvTUlD:-MS3g9et5z_THgtvTUlE:btoif8u) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L129)):*
> ```Java
>         }
> 
>         if (!checkImmutableAcrossDatabases(
> ```

We can just return the result of the recursive call here, right? since I think it's just else if statements all the way down then we return success

Same for the map, set, entry, collection `else if` blocks I think
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvn7Is0AP_xG5-lrfp:-MS3ep4P33q0u7KV73zr:b-9djcv3) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L142)):*

Can you suggest an alternate wording for it?
</blockquote></details>

Yeah sorry, should have done that before. Maybe just 

```
This would be simpler if we could assume that keys can be compared across databases using .equals(), however we cannot guarantee key equality so the simplest and most correct way to accomplish this is by reusing the set comparison
```

or something like that? 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 210 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvnVwwAJJ366y1pyaM:-MS3lB9tDo7srtjH1zPZ:byune0d) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L210)):*

I thought that it did verify, but after testing it found that it didn't.  It does now, so thanks for that :-)

We do care about this, it would be indicative of a flaw in our data model.  I do think you're right that it would simplify the code to do a double-pass through anyMatch(), though.
</blockquote></details>

Sweet, that makes sense -- this seems clear now and good. Thanks! 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 272 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvo18_2JE29YPXXFB_:-MS3js0-CO9ceVITH_0b:b-f82tby) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L272)):*

I can tell you just from looking at the output that it wouldn't be as useful without the missing/extra items.  You'd get "At Class.field: missing N items."  And then I'd swear and go add some extra code to print the missing/extra items so I could debug :-)

We can't currently use indexes because, strictly speaking, these are Collection's.  The assumption that collections that aren't sets or maps is kind of kludgy but there's no List interface and some of the existing ImmutableObject code makes a similar assumption.  I suppose we could convert them to arrays to use indexes, but that seems kind of ugly in its own right.

It's generally true that we can have less functionality with less code.  But do you find the current code difficult to follow?  It would be nice to be able to do this more succinctly, but if it's a choice between a better error message and less code I'm going to push hard for the former.
</blockquote></details>

Yeah I think with the conversion to ComparisonResult this becomes a ton clearer -- thanks for the consideration but yeah I think this is good 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 105 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3eRKE4Ll8dPCDD7sW:-MS3eRKE4Ll8dPCDD7sX:b-abm7ip) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L105)):*
> ```Java
> 
>     // Check for null first
>     if (actual == null && expected != null) {
> ```

nit: this line and the else if statement two lines below can each have one clause removed -- if both variables are null we'd hit the Objects.equals() line above
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 186 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3hJkWDUpdbmwA_IyX:-MS3hJkWDUpdbmwA_IyY:b-lxxwyx) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L186)):*
> ```Java
>       }
>     } else {
>       if (!actual.equals(expected)) {
> ```

I believe this `equals` query always be false because of the Objects.equals() check at the beginning of the function
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3i1gyE0ATZhf-0tgu:-MS3i1gyE0ATZhf-0tgv:bb1lf9c) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L215)):*
> ```Java
>           // If the element matches multiple elements in "expected," we have a basic problem with
>           // this kind of set that we'll want to know about.
>           if (found.contains(actualElem)) {
> ```

Maybe this is a bit overly clever, but we could just use the result of `found.add()` here I believe, since `add` returns true iff the elem was not previously present 

Separately, do we want to worry about comparing `actualElem` to the elements already in `found` using the recursive manual comparison (rather than the built-in .equals)? If so we can't use .add, but we'd have to add that comparison here 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 233 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3j52D8ND-cekj_nms:-MS3j52D8ND-cekj_nmt:b-1ki4hs) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L233)):*
> ```Java
>     }
> 
>     // Build a set of all objects in actual that don't have counterparts in expected.
> ```

The premise of some of this is that if `path` is null, we just return a simple failure result without worrying about the message, right? 

It might simplify the if statements to build `missing` and `found` without worrying about `path` at all -- then once we have those built, continue on (if we have path) or return a failure (if we don't have a path). It means we don't short-circuit so it takes slightly longer but it removes some branching 

I guess we could even give the missing / unexpected elements as part of the message then (even without a path) but eh that's probably more confusing than helpful 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 295 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3jzTTDnSF3Qg4tFUc:-MS3jzTUDEPHElXaJoRl:b-64vzbh) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L295)):*
> ```Java
> 
>   /** Recursive helper for isEqualAcrossDatabases. */
>   private static ComparisonResult checkImmutableAcrossDatabases(
> ```

nit: should this be `checkImmutableObjectAcrossDatabases`? 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 305 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3kYbO8A8jdv53AF8I:-MS3kYbO8A8jdv53AF8J:b7cokd4) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L305)):*
> ```Java
>       }
> 
>       // Verify that the field values are the same.  We can use "equals()" as a quick check.
> ```

nit: i think the second sentence of this comment is out of date

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 97 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvlqyM8LALLuquHAkM:-MS3pCXbDDgKy2ZIcn2U:b-1238zl) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L97)):*

Yeah, I think this is a ton better -- I hope you agree. 
</blockquote></details>

Yeah, I didn't really like having that check all over the place either.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvmKnm6zu-vojSSKfu:-MS3pd1UCuwQqQeiAnah:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L98)):*

Yeah, I see what you mean now. That makes sense, thanks for the detail and info. 
</blockquote></details>

Ack.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/942#-MS3g9et5z_THgtvTUlD:-MS3qkERC7JFURyjyUqd:b-gv5xr7) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L129)):*

We can just return the result of the recursive call here, right? since I think it's just else if statements all the way down then we return success

Same for the map, set, entry, collection `else if` blocks I think
</blockquote></details>

Good point!  Done.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/942#-MRvn7Is0AP_xG5-lrfp:-MS3tJwE-SkbQDSdztHQ:b-hg1zze) ([raw file](https://github.com/google/nomulus/blob/4d4ba037cbe3da1676eec67ffb66efb3105756a9/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L142)):*

Yeah sorry, should have done that before. Maybe just 

```
This would be simpler if we could assume that keys can be compared across databases using .equals(), however we cannot guarantee key equality so the simplest and most correct way to accomplish this is by reusing the set comparison
```

or something like that? 
</blockquote></details>

Done (though s/simpler/more efficient/ as simplicity is kind of ambiguous in this context -- hope that doesn't contravene your intent...).
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 105 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3eRKE4Ll8dPCDD7sW:-MS3qbnc23IiGAZ_RI_6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L105)):*

nit: this line and the else if statement two lines below can each have one clause removed -- if both variables are null we'd hit the Objects.equals() line above
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 186 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3hJkWDUpdbmwA_IyX:-MS3uUwH0EJtdo86LGIZ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L186)):*

I believe this `equals` query always be false because of the Objects.equals() check at the beginning of the function
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3i1gyE0ATZhf-0tgu:-MS3wJtN3JLwVcSBIyvg:b-6hzox3) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L215)):*

Maybe this is a bit overly clever, but we could just use the result of `found.add()` here I believe, since `add` returns true iff the elem was not previously present 

Separately, do we want to worry about comparing `actualElem` to the elements already in `found` using the recursive manual comparison (rather than the built-in .equals)? If so we can't use .add, but we'd have to add that comparison here 
</blockquote></details>

found.add() -- Done.

We actually want to use even stricter equality with `found` and `missing` because we're concerned with keeping track of the presence/absence of _the objects themselves_ rather than objects that they might be equal to by either ImmutableObject.equals() or a cross-database concept of equallity.  I had wanted to use object identity, but didn't know how to until I thought to look for an "identity set" just now!  Fixed.  Now my life is complete. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 233 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3j52D8ND-cekj_nms:-MS41hNM7BLmGcBRLnUu:bhs7f5q) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L233)):*

The premise of some of this is that if `path` is null, we just return a simple failure result without worrying about the message, right? 

It might simplify the if statements to build `missing` and `found` without worrying about `path` at all -- then once we have those built, continue on (if we have path) or return a failure (if we don't have a path). It means we don't short-circuit so it takes slightly longer but it removes some branching 

I guess we could even give the missing / unexpected elements as part of the message then (even without a path) but eh that's probably more confusing than helpful 
</blockquote></details>

Yeah, that's the basic idea: the more complicated work isn't necessary unless we're building a message.

We always have to build `found`.  `missing` is initialized but not actually built, and `unexpected` is neither initialized nor built when path is null.  I'd like to retain the short-circuit for `missing`, but I think I can simplify the rest of the branching quite a bit here and still avoid unnecessary construction, PTAL. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 295 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3jzTTDnSF3Qg4tFUc:-MS4Bm_--XCamIO6xY-F:b-ri710w) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L295)):*

nit: should this be `checkImmutableObjectAcrossDatabases`? 
</blockquote></details>

I intentionally wanted to avoid the extra verbosity, especially given that it's a private method, but if it seems overly dissonant let me know and I'll change it.
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 305 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3kYbO8A8jdv53AF8I:-MS4ER-E46QEU0JJFnHO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L305)):*

nit: i think the second sentence of this comment is out of date 
</blockquote></details>

Done.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3i1gyE0ATZhf-0tgu:-MS4u6jWD4LK-IZLkslF:b-whe5yw) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L215)):*

found.add() -- Done.

We actually want to use even stricter equality with `found` and `missing` because we're concerned with keeping track of the presence/absence of _the objects themselves_ rather than objects that they might be equal to by either ImmutableObject.equals() or a cross-database concept of equallity.  I had wanted to use object identity, but didn't know how to until I thought to look for an "identity set" just now!  Fixed.  Now my life is complete. 
</blockquote></details>

(discussing out of band) 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 233 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3j52D8ND-cekj_nms:-MS4LwlzBl-INv1gjpV1:b-ux3msj) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L233)):*

Yeah, that's the basic idea: the more complicated work isn't necessary unless we're building a message.

We always have to build `found`.  `missing` is initialized but not actually built, and `unexpected` is neither initialized nor built when path is null.  I'd like to retain the short-circuit for `missing`, but I think I can simplify the rest of the branching quite a bit here and still avoid unnecessary construction, PTAL. 
</blockquote></details>

Yeah, this seems  better I think, thanks. 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 295 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3jzTTDnSF3Qg4tFUc:-MS4LsCzEjod6UIAX26g:b-kbzrai) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L295)):*

I intentionally wanted to avoid the extra verbosity, especially given that it's a private method, but if it seems overly dissonant let me know and I'll change it.
</blockquote></details>

Yeah, sounds fine 
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 232 at r3](https://reviewable.io/reviews/google/nomulus/942#-MS4Lm-V9SgTGGndITeo:-MS4Lm-V9SgTGGndITep:b-p5xj0r) ([raw file](https://github.com/google/nomulus/blob/3f51ce224b629675a3d2f82b885a8a27a335012c/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L232)):*
> ```Java
>                       }));
> 
>       if (!missing.isEmpty() || (unexpected != null && !unexpected.isEmpty())) {
> ```

I don't think `unexpected` can be null here so that check seems unnecessary

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/942#-MS3i1gyE0ATZhf-0tgu:-MS8k94HDxhM34Xb_cui:bzgkwf8) ([raw file](https://github.com/google/nomulus/blob/e647fe589834e72f29f186b9facf7745b3cd3435/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L215)):*

(discussing out of band) 
</blockquote></details>

out of band we're good to go here -- if the concern is that two elements in set A map to a single element in set B, that will be caught below when processing unexpected elements

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 232 at r3](https://reviewable.io/reviews/google/nomulus/942#-MS4Lm-V9SgTGGndITeo:-MS9E3s1Ac3TojykWUn-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3f51ce224b629675a3d2f82b885a8a27a335012c/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L232)):*

I don't think `unexpected` can be null here so that check seems unnecessary
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/942)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #950, opened 2021-02-02
Refactor PremiumList storage and retrieval for dual-database setup

gbrodman:
Previously, the storage and retrieval code was scattered across various
places haphazardly and there was no good way to set up dual database
access. This reorganizes the code so that retrieval is simpler and it
allows for dual-write and dual-read.

This includes the following changes:

- Move all static / object retrieval code out of PremiumList -- the
class should solely consist of its data and methods on its data and it
shouldn't have to worry about complicated caching or retrieval

- Split all PremiumList retrieval methods into PremiumListDatastoreDao
and PremiumListSqlDao that handle retrieval of the premium list entry
objects from the corresponding databases (since the way the actual data
itself is stored is not the same between the two

- Create a dual-DAO for PremiumList retrieval that branches between
SQL/Datastore depending on which is appropriate -- it will read from
and write to both but only log errors for the secondary DB

- Cache the mapping from name to premium list in the dual-DAO. This is a
common code path regardless of database so we can cache it at a high
level

- Cache the ways to go from premium list -> premium entries in the
Datastore and SQL DAOs. These caches are specific to the corresponding
DB and should thus be stored in the corresponding DAO.

- Moves the database-choosing code from the actions to the lower-level
dual-DAO. This is because we will often wish to access this premium list
data in flows and all accesses should use the proper DB-selecting code

- Removes the DAO distinction between `saveNew` and `update` in the SQL world. This is unnecessary and something that can be handled by the action / business logic rather than the DAO. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/950)


gbrodman:
Note to @CydeWeys as reviewer: much of the loading / retrieval code blocks are the same, just copied and pasted from one area to another. So hopefully there's fewer differences to actually review than it appears like at first glance.

sarahcaseybot (changes requested):
> Reviewed 17 of 31 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 17 of 31 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSiiwO7-9xcnE6wdGlK:-MSiiwO8CfpL4Gp5FxWN:bpqxd50) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L136)):*
> ```Java
>       if (!primaryResult.equals(secondaryResult)) {
>         logger.atWarning().log(
>             "Unequal prices for domain %s.%s from primary DB (%s) and secondary DB (%s).",
> ```

To help with debugging, I think it would be useful if this specifies what database is primary in the log
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 141 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSij7by6CX-qGAwxieX:-MSij7by6CX-qGAwxieY:b-n08s34) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L141)):*
> ```Java
>     } catch (Throwable t) {
>       logger.atSevere().withCause(t).log(
>           "Error loading price of domain %s.%s from Cloud SQL.", label, registry.getTldStr());
> ```

I think the database here could also be Datastore
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 174 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSijTrC8frTSxwljkyA:-MSijTrC8frTSxwljkyB:b-1ok4ec) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L174)):*
> ```Java
>     if (tm().isOfy()) {
>       PremiumListDatastoreDao.delete(premiumList);
>       runAndLogExceptions(() -> PremiumListSqlDao.delete(premiumList), "deletion");
> ```

Check out `DatabaseMigrationUtils.suppressExceptionUnlessInTest`
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 216 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSikjz7-EMt_33x1-9U:-MSikjz7-EMt_33x1-9V:bid9wk5) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L216)):*
> ```Java
>   }
> 
>   private static Optional<PremiumList> loadPremiumListUncached(String premiumListName) {
> ```

Why does this not read from both databases?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 17 of 31 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 174 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSijTrC8frTSxwljkyA:-MSinPcE4d267ifGSpEb:b-puoma3) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L174)):*

Check out `DatabaseMigrationUtils.suppressExceptionUnlessInTest`
</blockquote></details>

yeah good call, can add more tests then

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 17 of 31 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSiiwO7-9xcnE6wdGlK:-MSnqZaf64xEd2QlyzOD:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L136)):*

To help with debugging, I think it would be useful if this specifies what database is primary in the log
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 141 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSij7by6CX-qGAwxieX:-MSnqa4t1f2xUC-iVEIP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L141)):*

I think the database here could also be Datastore
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 216 at r1](https://reviewable.io/reviews/google/nomulus/950#-MSikjz7-EMt_33x1-9U:-MSnqeEY6BlN72u2sFrU:b-4tc7lj) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L216)):*

Why does this not read from both databases?
</blockquote></details>

Yeah, we can. There are a couple downsides. 

1. In order to avoid random exceptions scattered across tests that use premium lists at all (this encompasses a ton of tests) we must initialize SQL (and save the premium list to SQL) every time. This just leads to some annoying boilerplate code in tests, the JPA extension. 

2. The objects aren't directly comparable themselves because we store different fields in the Datastore world versus the SQL world, so all we can really do is check for existence.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 14 of 37 files reviewed, 9 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/OteAccountBuilder.java, line 292 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2lRvd2qNphp_n36is:-MT2lRvd2qNphp_n36it:bofmh4p) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/OteAccountBuilder.java#L292)):*
> ```Java
>       int roidSuffix) {
>     String tldNameAlphaNumerical = tldName.replaceAll("[^a-z0-9]", "");
>     Optional<PremiumList> premiumList = PremiumListDualDao.getLatestRevision(DEFAULT_PREMIUM_LIST);
> ```

It's not clear to me at the callsite now whether the cached or uncached codepath is being called.  This is a regression.  This needs to be calling the uncached code path, by the way.

(Whether the cached or uncached code path is being called is a very important distinction as it has significant impacts on both performance, freshness, and correctness of mutations.)
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2m-6Y8kP_4d1AjDtG:-MT2m-6Z3jqTsU5ZZoAx:bl6ihku) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L144)):*
> ```Java
>     // If we're dealing with a list from SQL, reload from Datastore if necessary
>     if (premiumList.getRevisionKey() == null) {
>       Optional<PremiumList> fromDatastore = getLatestRevisionUncached(premiumList.getName());
> ```

This is wrong.  `getPremiumPrice()` is called in registrar-callable code paths and accordingly it needs high performance.  So it absolutely has to call the cached code path.

Is the problem that we don't have a cache by premium list name, only by revision key?
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 222 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2mnDr2DpDy6qZqU6k:-MT2mnDs6Gf6Qce0kDDH:b9vkrkk) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L222)):*
> ```Java
>             });
> 
>     // TODO(b/79888775): Enqueue the oldPremiumList for deletion after at least
> ```

You can remove this TODO.  We're never going to get around to it at this point as the end plan now is to just abandon all data in Datastore once the migration is complete.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 260 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2oLNL2Cw1PTtuPyNr:-MT2oLNL2Cw1PTtuPyNs:b-2ku6bd) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L260)):*
> ```Java
>   }
> 
>   private static void assertOfyFieldsEqual(
> ```

`assert` isn't exactly the right verb to use here, as you aren't actually `assert`ing anything, nor are you throwing any `AssertionError`s.  `check`?
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 165 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2p-wT4wvQiIxzCrGW:-MT2p-wT4wvQiIxzCrGX:b-5fvk89) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L165)):*
> ```Java
>   public static PremiumList save(String name, List<String> inputData) {
>     PremiumList result;
>     if (tm().isOfy()) {
> ```

So we're letting `tm()` tell us which DB is primary?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 14 of 37 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/OteAccountBuilder.java, line 292 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2lRvd2qNphp_n36is:-MT6hF_g63v2dNv4SLYC:b8utrsv) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/OteAccountBuilder.java#L292)):*

It's not clear to me at the callsite now whether the cached or uncached codepath is being called.  This is a regression.  This needs to be calling the uncached code path, by the way.

(Whether the cached or uncached code path is being called is a very important distinction as it has significant impacts on both performance, freshness, and correctness of mutations.)
</blockquote></details>

Why does this particular callsite need the uncached code path? If it's just the default sandbox list it shouldn't matter whether or not it's cached. 

This PR, in general, removes the uncached code path from the dual-DAO path. Given that the cache is a relatively short-lived cache, this shouldn't affect much. 
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2m-6Y8kP_4d1AjDtG:-MT6hg6k3DQXvcoiMEf9:b-1ko6su) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L144)):*

This is wrong.  `getPremiumPrice()` is called in registrar-callable code paths and accordingly it needs high performance.  So it absolutely has to call the cached code path.

Is the problem that we don't have a cache by premium list name, only by revision key?
</blockquote></details>

Ugh yeah, this is an artifact of using the same object with two almost-completely-different sets of fields. 

Maybe we just need to have a name -> PremiumList cache in both individual DAOs. Very annoying but doable. 
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 222 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2mnDr2DpDy6qZqU6k:-MT6gbye2HsvZIFqPocJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L222)):*

You can remove this TODO.  We're never going to get around to it at this point as the end plan now is to just abandon all data in Datastore once the migration is complete.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 260 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2oLNL2Cw1PTtuPyNr:-MT6gQFD8deJQl9bSwgY:b-p4g1n3) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L260)):*

`assert` isn't exactly the right verb to use here, as you aren't actually `assert`ing anything, nor are you throwing any `AssertionError`s.  `check`?
</blockquote></details>

Yeah, can do `check`. 
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java, line 165 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2p-wT4wvQiIxzCrGW:-MT6_aSS67mAtNaeHzSo:b-wp58ma) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDualDao.java#L165)):*

So we're letting `tm()` tell us which DB is primary?
</blockquote></details>

At the moment, in this PR, yes. This is obviously not ideal, and it should use the database scheduler added by Sarah in another PR (that went in after this PR was started). 

I was hoping to avoid adding even more complexity to this already-huge PR -- I'll put a TODO to use the new scheduler though. 

Just realized I also even had a TODO for this in the class-level javadoc. Cool, will also add at callsites

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 14 of 37 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2m-6Y8kP_4d1AjDtG:-MT6j32mE76H-WAndclC:b-iuqsfb) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L144)):*

Ugh yeah, this is an artifact of using the same object with two almost-completely-different sets of fields. 

Maybe we just need to have a name -> PremiumList cache in both individual DAOs. Very annoying but doable. 
</blockquote></details>

FYI i haven't done this yet, wanted to respond to the other comments though

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 14 of 37 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/OteAccountBuilder.java, line 292 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2lRvd2qNphp_n36is:-MT7gWsxCp3CoHHmtw_c:b-8b37jf) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/OteAccountBuilder.java#L292)):*

Why does this particular callsite need the uncached code path? If it's just the default sandbox list it shouldn't matter whether or not it's cached. 

This PR, in general, removes the uncached code path from the dual-DAO path. Given that the cache is a relatively short-lived cache, this shouldn't affect much. 
</blockquote></details>

My question is about `PremiumListDualDao.getLatestRevision()`, not specifically this callsite.

So `getLatestRevision()` is always cached, and there's no uncached way to get the latest revision?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: 14 of 37 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/OteAccountBuilder.java, line 292 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2lRvd2qNphp_n36is:-MT7hCBY1vytI30MKhoe:b-ibn4j3) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/OteAccountBuilder.java#L292)):*

My question is about `PremiumListDualDao.getLatestRevision()`, not specifically this callsite.

So `getLatestRevision()` is always cached, and there's no uncached way to get the latest revision?
</blockquote></details>

Basically yes, which shouldn't really matter since the cache time is fairly short and it shouldn't ever be invalid anyway 
___
*[core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/950#-MT2m-6Y8kP_4d1AjDtG:-MT7hLa680C3JZWw2M7o:boik682) ([raw file](https://github.com/google/nomulus/blob/f29156a0c18fb19665f30d60444f906e275ec1e2/core/src/main/java/google/registry/model/registry/label/PremiumListDatastoreDao.java#L144)):*

FYI i haven't done this yet, wanted to respond to the other comments though 
</blockquote></details>

OK now this is done. So many caches!

sarahcaseybot (changes requested):
> Reviewed 8 of 31 files at r1, 8 of 13 files at r2, 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/test/java/google/registry/export/ExportPremiumTermsActionTest.java, line 108 at r1](https://reviewable.io/reviews/google/nomulus/950#-MT8xBuI0NOZkpjBvNht:-MT8xBuI0NOZkpjBvNhu:b-6vg9rz) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/test/java/google/registry/export/ExportPremiumTermsActionTest.java#L108)):*
> ```Java
> 
>   @Test
>   void test_exportPremiumTerms_success_emptyPremiumList() throws IOException {
> ```

Why did you delete this test?
___
*[core/src/test/java/google/registry/schema/tld/PremiumListSqlDaoTest.java, line 103 at r3](https://reviewable.io/reviews/google/nomulus/950#-MT940XU6INgmL4rpuVr:-MT940XU6INgmL4rpuVs:brqfgv) ([raw file](https://github.com/google/nomulus/blob/444f50bb1a45d6b1f8b03b4695d8228be981c684/core/src/test/java/google/registry/schema/tld/PremiumListSqlDaoTest.java#L103)):*
> ```Java
> 
>   @Test
>   void update_worksSuccessfully() {
> ```

Should this be renamed to save_worksSuccessfully()?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java, line 63 at r3](https://reviewable.io/reviews/google/nomulus/950#-MTBcdcb8h73S7IehH0s:-MTBcdcb8h73S7IehH0t:bp0qu5l) ([raw file](https://github.com/google/nomulus/blob/444f50bb1a45d6b1f8b03b4695d8228be981c684/core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java#L63)):*
> ```Java
>   @VisibleForTesting
>   public static void setPremiumListCacheForTest(Optional<Duration> expiry) {
>     Duration effectiveExpiry = expiry.orElse(getSingletonCachePersistDuration());
> ```

Why is this a different cache length than is used a few lines above?
___
*[core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java, line 76 at r3](https://reviewable.io/reviews/google/nomulus/950#-MTBcSe79LuUQcPp2h0J:-MTBcSe79LuUQcPp2h0K:b-l7w89t) ([raw file](https://github.com/google/nomulus/blob/444f50bb1a45d6b1f8b03b4695d8228be981c684/core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java#L76)):*
> ```Java
>               @Override
>               public Optional<PremiumList> load(final String name) {
>                 return jpaTm().doTransactionless(() -> getLatestRevisionUncached(name));
> ```

And just to be 100% clear, this isn't loading the contents of the premium list, just its name and a small other amount of metadata, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java, line 63 at r3](https://reviewable.io/reviews/google/nomulus/950#-MTBcdcb8h73S7IehH0s:-MTBfWz5A_mTyqhvhzaH:bo18nri) ([raw file](https://github.com/google/nomulus/blob/444f50bb1a45d6b1f8b03b4695d8228be981c684/core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java#L63)):*

Why is this a different cache length than is used a few lines above?
</blockquote></details>

You're right, it should be the domain label list cache duration.  Done.
___
*[core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java, line 76 at r3](https://reviewable.io/reviews/google/nomulus/950#-MTBcSe79LuUQcPp2h0J:-MTBhUpa96RvOpaZLxnm:bu15yry) ([raw file](https://github.com/google/nomulus/blob/444f50bb1a45d6b1f8b03b4695d8228be981c684/core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java#L76)):*

And just to be 100% clear, this isn't loading the contents of the premium list, just its name and a small other amount of metadata, right?
</blockquote></details>

Yes and no. 

The field `labelsToPrices` is a map (containing essentially the whole list) that is loaded lazily from the premium entry table. If we are calling this from a transaction and specifically, in that transaction, request to access `labelsToPrices` we load that field from the premium entry table. Otherwise, if we access this outside of a transaction or we return it outside of the transaction in which it was loaded, it only contains the currency, bloom filter, revision ID, name, creation time, and update time. 
___
*[core/src/test/java/google/registry/export/ExportPremiumTermsActionTest.java, line 108 at r1](https://reviewable.io/reviews/google/nomulus/950#-MT8xBuI0NOZkpjBvNht:-MTBfogoDkHo4KSbMUmn:b-lq9woo) ([raw file](https://github.com/google/nomulus/blob/7a6587249d6072a1b7b23f7716d202c43fd6d3c7/core/src/test/java/google/registry/export/ExportPremiumTermsActionTest.java#L108)):*

Why did you delete this test?
</blockquote></details>

We don't allow empty lists in the SQL world, as there must be at least one item to set the currency of the list. 
___
*[core/src/test/java/google/registry/schema/tld/PremiumListSqlDaoTest.java, line 103 at r3](https://reviewable.io/reviews/google/nomulus/950#-MT940XU6INgmL4rpuVr:-MTBgY7G-7y5fjm_OOA0:b-4sm0ms) ([raw file](https://github.com/google/nomulus/blob/444f50bb1a45d6b1f8b03b4695d8228be981c684/core/src/test/java/google/registry/schema/tld/PremiumListSqlDaoTest.java#L103)):*

Should this be renamed to save_worksSuccessfully()?
</blockquote></details>

nah, I think the wording here is good to differentiate it from the case above where we're only saving one list

sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

gbrodman (commented):
@cydeweys any further thoughts on this when you get a chance? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950#-:-MTfc7DM4S2rhlX94Nw5:b-kanwcs)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

CydeWeys (approved):
Can you make sure to test this on alpha?  I can see some potential issues here in that if a given premium price now fails to load for whatever reason (i.e. a code issue), that domain will then be available at the standard price.  We've had bugs with this before, where domains were incorrectly available at a standard price due to a complicated interaction between deletion of an old revision and caching.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950#-:-MTl4yzU99-epOCH9KeI:b-9e1jxs)* status: :shipit: complete! all files reviewed, all discussions resolved

gbrodman (commented):
Just tested on alpha. Updates, checks, and clears work as expected in the normal case. I did SQL surgery to remove entries from a list and I got the expected IllegalArgumentException in the logs when I requested one of the deleted domains, but the EPP command still returned as expected with no error in the output. It seems like we're good to go here. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/950#-:-MUBfY5GCIZfb_rFi39K:bb6xk9x)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #964, opened 2021-02-11
Partially convert RDAP ofy calls to tm calls

gbrodman:
This converts the simple retrieval actions but does not touch the more
complicated search actions -- those use some ofy() query searching logic
and will likely end up being significantly more complicated than this
change. Here, though, we are just changing the calls that can be
converted easily to tm() lookups.

To change in future PRs:
- RdapDomainSearchAction
- RdapEntitySearchAction
- RdapNameserverSearchAction
- RdapSearchActionBase

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/964)

### DIFF
--- a/core/src/main/java/google/registry/model/registrar/Registrar.java
+++ b/core/src/main/java/google/registry/model/registrar/Registrar.java
@@ -996,6 +996,13 @@ public class Registrar extends ImmutableObject
     return CACHE_BY_CLIENT_ID.get().values();
   }
 
+  /** Loads all registrar keys using an in-memory cache. */
+  public static ImmutableSet<VKey<Registrar>> loadAllKeysCached() {
+    return CACHE_BY_CLIENT_ID.get().keySet().stream()
+        .map(Registrar::createVKey)
+        .collect(toImmutableSet());
+  }
+
   /** Loads and returns a registrar entity by its client id directly from Datastore. */
   public static Optional<Registrar> loadByClientId(String clientId) {
     checkArgument(!Strings.isNullOrEmpty(clientId), "clientId must be specified");
--- a/core/src/main/java/google/registry/rdap/RdapEntityAction.java
+++ b/core/src/main/java/google/registry/rdap/RdapEntityAction.java
@@ -14,7 +14,8 @@
 
 package google.registry.rdap;
 
-import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 import static google.registry.rdap.RdapUtils.getRegistrarByIanaIdentifier;
 import static google.registry.rdap.RdapUtils.getRegistrarByName;
 import static google.registry.request.Action.Method.GET;
@@ -23,9 +24,9 @@ import static google.registry.request.Action.Method.HEAD;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.primitives.Longs;
 import com.google.re2j.Pattern;
-import com.googlecode.objectify.Key;
 import google.registry.model.contact.ContactResource;
 import google.registry.model.registrar.Registrar;
+import google.registry.persistence.VKey;
 import google.registry.rdap.RdapJsonFormatter.OutputDataType;
 import google.registry.rdap.RdapMetrics.EndpointType;
 import google.registry.rdap.RdapObjectClasses.RdapEntity;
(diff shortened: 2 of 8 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/964)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/RdapEntityAction.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTkwmPk8ynzvA6SZzoV:-MTkwmPk8ynzvA6SZzoW:b-7fs6uj) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/RdapEntityAction.java#L75)):*
> ```Java
>       VKey<ContactResource> contactVKey = VKey.create(ContactResource.class, pathSearchString);
>       Optional<ContactResource> contactResource =
>           transactIfJpaTm(() -> tm().loadByKeyIfPresent(contactVKey));
> ```

What's the thought process behind the several calls to `transactIfJpaTm()` in this PR?
___
*[core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTksMWo6pG0_ynDgOJL:-MTksMWo6pG0_ynDgOJM:b-euk86q) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java#L198)):*
> ```Java
>   public void run() {
>     ImmutableSetMultimap<String, String> ianaToBaseUrls = getRdapBaseUrlsPerIanaId();
>     Registrar.loadAllCached()
> ```

This is a bit different than what the previous line was doing.  I think we do want to add a `Registrar.loadAllKeys()` method to call from here.  It should be very similar to the existing `Registrar.loadAll()` method, just calling `.keys()` for `ofy` (like the line you removed here) and then also the SQL equivalent to load the ids only.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/964)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapEntityAction.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTkwmPk8ynzvA6SZzoV:-MTl7Mc39cMm-kboW62j:b-tudnn8) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/RdapEntityAction.java#L75)):*

What's the thought process behind the several calls to `transactIfJpaTm()` in this PR?
</blockquote></details>

This is fairly common across the codebase because ofy() doesn't necessarily require transactions for reads but SQL does. 
___
*[core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTksMWo6pG0_ynDgOJL:-MTl8MVo5Wsc0gQqYTMQ:b-j6egrx) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java#L198)):*

This is a bit different than what the previous line was doing.  I think we do want to add a `Registrar.loadAllKeys()` method to call from here.  It should be very similar to the existing `Registrar.loadAll()` method, just calling `.keys()` for `ofy` (like the line you removed here) and then also the SQL equivalent to load the ids only.
</blockquote></details>

I see what you mean, and it's actually even simpler than what you are thinking it is. Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/964)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/RdapEntityAction.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTkwmPk8ynzvA6SZzoV:-MTmhm_Z0hUz6lG-f0HJ:b77po8p) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/RdapEntityAction.java#L75)):*

This is fairly common across the codebase because ofy() doesn't necessarily require transactions for reads but SQL does. 
</blockquote></details>

Hrm, I do wonder what the performance implications of all of these additional read transactions in SQL might end up looking like.  As RDAP becomes more populart his might need to use a cached codepath (like WHOIS).

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/964)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapEntityAction.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTkwmPk8ynzvA6SZzoV:-MTrH7-G0JAmuIGSjdM-:b2nawm2) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/RdapEntityAction.java#L75)):*

Hrm, I do wonder what the performance implications of all of these additional read transactions in SQL might end up looking like.  As RDAP becomes more populart his might need to use a cached codepath (like WHOIS).
</blockquote></details>

Agreed, we will likely need to add a caching layer.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/964)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapEntityAction.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/964#-MTkwmPk8ynzvA6SZzoV:-MTrIVod5RqnEfgxmXGP:b-a9bkyx) ([raw file](https://github.com/google/nomulus/blob/36f1202a4a101b1c1c91eee4ce8b8ccb4a03ff97/core/src/main/java/google/registry/rdap/RdapEntityAction.java#L75)):*

Agreed, we will likely need to add a caching layer. 
</blockquote></details>

filed b/180645230

PR #968, opened 2021-02-16
Fix obscure bug when checking restore prices of duplicate domain names

CydeWeys:
There were instances of "java.lang.IllegalArgumentException: Multiple entries
with same key" in the logs, caused by attempting to construct an ImmutableMap
containing duplicate keys. It turns out this was happening in the domain check
flow when the following conditions were all simultaneously met:

1. The older v06 fee extension is used
2. The same domain name is being queried multiple times in a single check
   command (which is valid per the spec but doesn't actually make any sense)
3. Said domain exists
4. The cost of a restore (an uncommon operation) is being checked

When all of those conditions were met, an error was being thrown when the
dupe-containing list of domain names was used as the keys of a new Map. This
fixes that bug by calling .distinct() first.

Give enough registrars enough typewriters ...

BUG=179052195

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/968)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -301,6 +301,7 @@ public final class DomainCheckFlow implements Flow {
           feeCheck.getItems().stream()
               .filter(fc -> fc.getCommandName() == CommandName.RESTORE)
               .map(FeeCheckCommandExtensionItem::getDomainName)
+              .distinct()
               .collect(toImmutableList());
     } else if (feeCheck.getItems().stream()
         .anyMatch(fc -> fc.getCommandName() == CommandName.RESTORE)) {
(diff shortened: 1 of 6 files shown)


jianglai (approved):
I think at some point we should revisit the versions of fee extension that we support. The RFC has been [out](https://tools.ietf.org/html/rfc8748) since last year and v06 was from 2015. It doesn't make sense to support some old obscure draft versions (unless registrars are still actively using them?).

Also, are all the .xml files new? I'd thought that they were branched from existing files and modified. It's a shame that git is not smart enough to figure that out.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/968#-:-MTiGCtF2mdcWbAioKnt:bk483g4)* status: :shipit: complete! all files reviewed, all discussions resolved

CydeWeys (commented):
Yes, it's on my TODO list to look at.  There is a finalized version now with an RFC # and everything that we'd want to support: https://tools.ietf.org/html/rfc8748  We should finish the migration first though, as what we have now is working well enough.

Note that LOTS of registrars are still using v06 though.  If anything it's v11 and v12 that are really draft-like and that probably aren't as widely used.  This is one thing we probably want to get some analytics on at some point.

And no, the xml files are not new, and I copied them through IntelliJ which *said* it was adding the copy actions to git, but I guess that didn't translate over properly somehow.  They looked branched when I was checking them in!

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/968#-:-MTkx2M--00n_xElQeGr:box5yp5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #970, opened 2021-02-18
Reject handshakes with bad TLS protocols and ciphers

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/970)

### DIFF
--- a/networking/build.gradle
+++ b/networking/build.gradle
@@ -43,7 +43,7 @@ dependencies {
 
   annotationProcessor deps['com.google.dagger:dagger-compiler']
   testAnnotationProcessor deps['com.google.dagger:dagger-compiler']
-    compile 'joda-time:joda-time:2.9.2'
+  compile 'joda-time:joda-time:2.9.2'
 }
 
 // Make testing artifacts available to be depended up on by other projects.
--- a/networking/build.gradle
+++ b/networking/build.gradle
@@ -43,6 +43,7 @@ dependencies {
 
   annotationProcessor deps['com.google.dagger:dagger-compiler']
   testAnnotationProcessor deps['com.google.dagger:dagger-compiler']
+    compile 'joda-time:joda-time:2.9.2'
 }
 
 // Make testing artifacts available to be depended up on by other projects.
--- a/networking/build.gradle
+++ b/networking/build.gradle
@@ -28,6 +28,7 @@ dependencies {
   compile deps['org.bouncycastle:bcpkix-jdk15on']
   compile deps['org.bouncycastle:bcprov-jdk15on']
   compile project(':util')
+  compile project(':common')
 
   runtime deps['com.google.flogger:flogger-system-backend']
   runtime deps['io.netty:netty-tcnative-boringssl-static']
@@ -38,6 +39,7 @@ dependencies {
   testCompile deps['org.junit.jupiter:junit-jupiter-params']
   testCompile deps['org.bouncycastle:bcpkix-jdk15on']
   testCompile deps['org.bouncycastle:bcprov-jdk15on']
+  testCompile project(path: ':common', configuration: 'testing')
 
   annotationProcessor deps['com.google.dagger:dagger-compiler']
   testAnnotationProcessor deps['com.google.dagger:dagger-compiler']
--- a/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
+++ b/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
@@ -19,6 +19,7 @@ import static google.registry.util.X509Utils.getCertificateHash;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.flogger.FluentLogger;
+import google.registry.util.Clock;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandler.Sharable;
@@ -42,6 +43,7 @@ import java.security.cert.X509Certificate;
 import java.security.interfaces.RSAPublicKey;
 import java.util.function.Supplier;
 import javax.net.ssl.SSLSession;
+import org.joda.time.DateTime;
 
 /**
  * Adds a server side SSL handler to the channel pipeline.
@@ -72,10 +74,11 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
    *
    * <p>This list includes all of the current TLS1.3 ciphers and a collection of TLS1.2 ciphers with
    * no known security vulnerabilities. Note that OpenSSL uses a separate nomenclature for the
-   * ciphers than the ones used here. More information about these cipher suites and their OpenSSL
-   * names can be found at ciphersuite.info.
+   * ciphers internally but the IANA names listed here will be transparently translated by the
+   * OpenSSL provider (if used), so there is no need to include the OpenSSL name variants here. More
+   * information about these cipher suites and their OpenSSL names can be found at ciphersuite.info.
    */
-  private static final ImmutableList TLS_CIPHERS =
+  private static final ImmutableList ALLOWED_TLS_CIPHERS =
       ImmutableList.of(
           "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
           "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
@@ -99,13 +102,19 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
   private final Supplier<PrivateKey> privateKeySupplier;
   private final Supplier<ImmutableList<X509Certificate>> certificatesSupplier;
   private final ImmutableList<String> supportedSslVersions;
+  // TODO(sarahbot): Remove this variable and its check after enforcement start date has passed.
+  private final ImmutableList<String> oldSupportedSslVersions;
+  private final DateTime enforcementStartTime;
+  private final Clock clock;
 
   public SslServerInitializer(
       boolean requireClientCert,
       boolean validateClientCert,
       SslProvider sslProvider,
       Supplier<PrivateKey> privateKeySupplier,
-      Supplier<ImmutableList<X509Certificate>> certificatesSupplier) {
+      Supplier<ImmutableList<X509Certificate>> certificatesSupplier,
+      DateTime enforcementStartTime,
+      Clock clock) {
     logger.atInfo().log("Server SSL Provider: %s", sslProvider);
     checkArgument(
         requireClientCert || !validateClientCert,
@@ -121,6 +130,12 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
             // JDK support for TLS 1.3 won't be available until 2021-04-20 at the earliest.
             // See: https://java.com/en/jre-jdk-cryptoroadmap.html
             : ImmutableList.of("TLSv1.2");
+    this.oldSupportedSslVersions =
+        sslProvider == SslProvider.OPENSSL
+            ? ImmutableList.of("TLSv1.3", "TLSv1.2", "TLSv1.1", "TLSv1")
+            : ImmutableList.of("TLSv1.2", "TLSv1.1", "TLSv1");
+    this.enforcementStartTime = enforcementStartTime;
+    this.clock = clock;
   }
 
   @Override
@@ -132,8 +147,13 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
             .sslProvider(sslProvider)
             .trustManager(InsecureTrustManagerFactory.INSTANCE)
             .clientAuth(requireClientCert ? ClientAuth.REQUIRE : ClientAuth.NONE)
-            .protocols(supportedSslVersions)
-            .ciphers(TLS_CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
+            .protocols(
+                enforcementStartTime.isBefore(clock.nowUtc())
+                    ? supportedSslVersions
+                    : oldSupportedSslVersions)
+            .ciphers(
+                enforcementStartTime.isBefore(clock.nowUtc()) ? ALLOWED_TLS_CIPHERS : null,
+                SupportedCipherSuiteFilter.INSTANCE)
             .build();
 
     logger.atInfo().log("Available Cipher Suites: %s", sslContext.cipherSuites());
--- a/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
+++ b/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
@@ -29,6 +29,7 @@ import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.ssl.SslProvider;
+import io.netty.handler.ssl.SupportedCipherSuiteFilter;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
 import io.netty.util.AttributeKey;
 import io.netty.util.concurrent.Future;
@@ -102,6 +103,23 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
 
   @Override
   protected void initChannel(C channel) throws Exception {
+    ImmutableList.Builder ciphers =
+        ImmutableList.builder()
+            .add(
+                "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
+                "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+                "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
+                "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+                "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+                "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384");
+    if (sslProvider.equals(SslProvider.OPENSSL)) {
+      ciphers.add(
+          "TLS_AES_128_GCM_SHA256",
+          "TLS_AES_256_GCM_SHA384",
+          "TLS_CHACHA20_POLY1305_SHA256",
+          "TLS_AES_128_CCM_SHA256",
+          "TLS_AES_128_CCM_8_SHA256");
+    }
     SslContext sslContext =
         SslContextBuilder.forServer(
                 privateKeySupplier.get(),
@@ -110,7 +128,9 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
             .trustManager(InsecureTrustManagerFactory.INSTANCE)
             .clientAuth(requireClientCert ? ClientAuth.REQUIRE : ClientAuth.NONE)
             .protocols(supportedSslVersions)
+            .ciphers(ciphers.build(), SupportedCipherSuiteFilter.INSTANCE)
             .build();
+
     logger.atInfo().log("Available Cipher Suites: %s", sslContext.cipherSuites());
     SslHandler sslHandler = sslContext.newHandler(channel.alloc());
     if (requireClientCert) {
--- a/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
+++ b/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
@@ -95,10 +95,10 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
     this.certificatesSupplier = certificatesSupplier;
     this.supportedSslVersions =
         sslProvider == SslProvider.OPENSSL
-            ? ImmutableList.of("TLSv1.3", "TLSv1.2", "TLSv1.1", "TLSv1")
-            // JDK support for TLS 1.3 won't be available until 2020-07-14 at the earliest.
+            ? ImmutableList.of("TLSv1.3", "TLSv1.2")
+            // JDK support for TLS 1.3 won't be available until 2021-04-20 at the earliest.
             // See: https://java.com/en/jre-jdk-cryptoroadmap.html
-            : ImmutableList.of("TLSv1.2", "TLSv1.1", "TLSv1");
+            : ImmutableList.of("TLSv1.2");
   }
 
   @Override
--- a/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
+++ b/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
@@ -67,6 +67,28 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
   public static final AttributeKey<Promise<X509Certificate>> CLIENT_CERTIFICATE_PROMISE_KEY =
       AttributeKey.valueOf("CLIENT_CERTIFICATE_PROMISE_KEY");
 
+  /**
+   * The list of cipher suites that are currently acceptable to create a successful handshake.
+   *
+   * <p>This list includes all of the current TLS1.3 ciphers and a collection of TLS1.2 ciphers with
+   * no known security vulnerabilities. Note that OpenSSL uses a separate nomenclature for the
+   * ciphers than the ones used here. More information about these cipher suites and their OpenSSL
+   * names can be found at ciphersuite.info.
+   */
+  private static final ImmutableList TLS_CIPHERS =
+      ImmutableList.of(
+          "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
+          "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+          "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
+          "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+          "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+          "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+          "TLS_AES_128_GCM_SHA256",
+          "TLS_AES_256_GCM_SHA384",
+          "TLS_CHACHA20_POLY1305_SHA256",
+          "TLS_AES_128_CCM_SHA256",
+          "TLS_AES_128_CCM_8_SHA256");
+
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private final boolean requireClientCert;
   // TODO(jianglai): Always validate client certs (if required).
@@ -103,23 +125,7 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
 
   @Override
   protected void initChannel(C channel) throws Exception {
-    ImmutableList.Builder ciphers =
-        ImmutableList.builder()
-            .add(
-                "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
-                "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
-                "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
-                "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
-                "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
-                "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384");
-    if (sslProvider.equals(SslProvider.OPENSSL)) {
-      ciphers.add(
-          "TLS_AES_128_GCM_SHA256",
-          "TLS_AES_256_GCM_SHA384",
-          "TLS_CHACHA20_POLY1305_SHA256",
-          "TLS_AES_128_CCM_SHA256",
-          "TLS_AES_128_CCM_8_SHA256");
-    }
+    ImmutableList.Builder ciphers = ImmutableList.builder().addAll(TLS_CIPHERS);
     SslContext sslContext =
         SslContextBuilder.forServer(
                 privateKeySupplier.get(),
--- a/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
+++ b/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java
@@ -125,7 +125,6 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
 
   @Override
   protected void initChannel(C channel) throws Exception {
-    ImmutableList.Builder ciphers = ImmutableList.builder().addAll(TLS_CIPHERS);
     SslContext sslContext =
         SslContextBuilder.forServer(
                 privateKeySupplier.get(),
@@ -134,7 +133,7 @@ public class SslServerInitializer<C extends Channel> extends ChannelInitializer<
             .trustManager(InsecureTrustManagerFactory.INSTANCE)
             .clientAuth(requireClientCert ? ClientAuth.REQUIRE : ClientAuth.NONE)
             .protocols(supportedSslVersions)
-            .ciphers(ciphers.build(), SupportedCipherSuiteFilter.INSTANCE)
+            .ciphers(TLS_CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
             .build();
 
     logger.atInfo().log("Available Cipher Suites: %s", sslContext.cipherSuites());
--- a/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java
+++ b/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java
@@ -227,10 +227,52 @@ class SslServerInitializerTest {
 
   @ParameterizedTest
   @MethodSource("provideTestCombinations")
+  void testSuccess_someCiphersNotAccepted(SslProvider sslProvider) throws Exception {
+    SelfSignedCaCertificate serverSsc = SelfSignedCaCertificate.create(SSL_HOST);
+    LocalAddress localAddress = new LocalAddress("SOME_CIPHERS_NOT_ACCEPTED_" + sslProvider);
+
+    nettyExtension.setUpServer(
+        localAddress,
+        new SslServerInitializer<LocalChannel>(
+            true,
+            true,
+            sslProvider,
+            Suppliers.ofInstance(serverSsc.key()),
+            Suppliers.ofInstance(ImmutableList.of(serverSsc.cert())),
+            DateTime.parse("2021-04-01T16:00:00Z"),
+            new FakeClock(DateTime.parse("2021-05-01T16:00:00Z"))));
+    SelfSignedCaCertificate clientSsc =
+        SelfSignedCaCertificate.create(
+            "CLIENT",
+            Date.from(Instant.now().minus(Duration.ofDays(2))),
+            Date.from(Instant.now().plus(Duration.ofDays(1))));
+    nettyExtension.setUpClient(
+        localAddress,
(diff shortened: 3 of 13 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 109 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUHBgHzAE9V56QpAcCY:-MUHBgI-5mPXmYQZier8:bozjogq) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L109)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
>                 "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
>                 "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
>                 "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
>                 "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
>                 "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
>                 "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384");
> ```
</details>

These should `static final` constants defined at the beginning of the class. Same below. We should also include a link to a reference website like ciphersuite.info.
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 117 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUHCSFs3vjWqg19cJHf:-MUHCSFs3vjWqg19cJHg:b-xjxpcs) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L117)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
>           "TLS_AES_128_GCM_SHA256",
>           "TLS_AES_256_GCM_SHA384",
>           "TLS_CHACHA20_POLY1305_SHA256",
>           "TLS_AES_128_CCM_SHA256",
>           "TLS_AES_128_CCM_8_SHA256");
> ```
</details>

Are we sure we should be **adding** these cipher suites to the allowed cipher suites when using OpenSSL? My understand is that OpenSSL uses different a nomenclature for cipher suites so we should use a different set for OpenSSL altogether. For example  `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` is the IANA name used by the JDK SSL engine, where the OpenSSL engine uses `ECDHE-ECDSA-AES128-GCM-SHA256` for the same cipher suite.

It looks to me that the cipher suites listed here are actually for TLS 1.3, which only OpenSSL supports for now. They actually have the same name between IANA and OpenSSL. So I think the right thing to do is to maintain two lists, one for OpenSSL and one for JDK, regardless of TLS versions. Because this is an allowlist it doesn't matter if we allow a TLS 1.3 only cipher suite when TLS 1.2 is negotiated.
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUHEzSd6RBMJPDktcCF:-MUHEzSd6RBMJPDktcCG:b-k2zd7t) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L132)):*
> ```
> SupportedCipherSuiteFilter.INSTANCE
> ```
Actually, since you have the filter here, maybe it is not even necessary to maintain two lists? We can just store all the cipher suites that we allow in both their IANA and their OpenSSL names in a single list (in a way that clearly shows the relationship between the two name variant for the same cipher suite, of course) and have the filter remove the ones not supported by the SSL engine.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJSyKx2HivykM6cSul:-MUJSyKx2HivykM6cSum:b-ae8x1p) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L110)):*
> ```
> ciphers
> ```
cipher?
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJVBXrArQ1rZiTZcdG:-MUJVBXrArQ1rZiTZcdH:b-wmribb) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L120)):*
> ```
> sslContextBuilder.protocols("TLSv1.2").ciphers(Collections.singletonList(ciphers));
> ```
Why resetting the protocol to 1.2 when the cipher is null? I don't think I understand the logic here. Also for v1.2 the cipher names are generally different between different SSL providers so I'm not sure what would happen when a cipher is given that is not recognized by the provider. I think this method should just set the protocol and cipher as is. null cipher/protocol would just be handled as system defaults, which should conform to our standard. If we want to be extra sure that tests don't depend on the system, see my comment below.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 149 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJd_PH1iepXClBNNmg:-MUJd_PH1iepXClBNNmh:b-5kkc3) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L149)):*
> ```
>         sslProvider.equals(SslProvider.OPENSSL) ? "TLSv1.3" : "TLSv1.2",
>         null);
> ```
Since this helper is used by all existing tests that are not concerned with TLS versions or cipher suites, we should pick arguments that are most "normal". This means the version should be 1.2 (1.3 is still rare and not supported by the JDK SSL engine) and the cipher should be one of the common 1.2 cipher, using the name variant appropriate for the provider.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 218 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJguKm3Do7A1czmChM:-MUJguKm3Do7A1czmChN:b-3sooe4) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L218)):*
> ```
>             sslProvider.equals(SslProvider.OPENSSL) ? "TLSv1.3" : "TLSv1.2",
>             "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"));
> ```
As I mentioned above, we should stick with v1.2 and the cipher's name needs to depend on the ssl provider used.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: all files reviewed, 7 unresolved discussions (waiting on @sarahcaseybot)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: all files reviewed, 6 unresolved discussions (waiting on @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 117 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUHCSFs3vjWqg19cJHf:-MUMFhk-5sqGnNA9jch4:bdfmsf9) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L117)):*

After playing around with this a bit, it seems like the difference in nomenclature between openssl and the IANA name used in the code does not cause any issues. In fact, the code actually doesn't work at all if I try to use the openssl name for the cipher suite. 

While playing around with this, I also discovered that it actually seems impossible to connect via  one of the ECDSA ciphers (in this PR and with the current deployed proxy). This is likely why none of our clients use ECDSA ciphers. It is worth investigating more into why ECDSA ciphers does not seem to be working.
</blockquote></details>

OK, I guess it makes sense the API only takes IANA nomenclature and the OpenSSL implementation actually converts it internally. Still, is there a reason to separate the ciphers this way? Why not just include everything in the same list? These are all the ciphers we support, and if the SSL engine doesn't support TLS 1.3, they will be filtered by `SupportedCipherSuiteFilter.INSTANCE`, right?

I'll take look into the ECDSA cipher issue later.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @jianglai)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 109 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUHBgHzAE9V56QpAcCY:-MUPKl-rDFiR7GjAlmg-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L109)):*

> ```
>                 "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
>                 "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
>                 "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
>                 "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
>                 "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
>                 "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384");
> ```
These should `static final` constants defined at the beginning of the class. Same below. We should also include a link to a reference website like ciphersuite.info.
</blockquote></details>

Done.
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUHEzSd6RBMJPDktcCF:-MUPAnJp8HDR8ImvZrYi:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L132)):*

> ```
> SupportedCipherSuiteFilter.INSTANCE
> ```
Actually, since you have the filter here, maybe it is not even necessary to maintain two lists? We can just store all the cipher suites that we allow in both their IANA and their OpenSSL names in a single list (in a way that clearly shows the relationship between the two name variant for the same cipher suite, of course) and have the filter remove the ones not supported by the SSL engine.
</blockquote></details>

Done.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJSyKx2HivykM6cSul:-MULSuj40L1sd7B5PTdW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L110)):*

> ```
> ciphers
> ```
cipher?
</blockquote></details>

Done.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJVBXrArQ1rZiTZcdG:-MUPC9fO9IyJc7tNq03L:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L120)):*

> ```
> sslContextBuilder.protocols("TLSv1.2").ciphers(Collections.singletonList(ciphers));
> ```
Why resetting the protocol to 1.2 when the cipher is null? I don't think I understand the logic here. Also for v1.2 the cipher names are generally different between different SSL providers so I'm not sure what would happen when a cipher is given that is not recognized by the provider. I think this method should just set the protocol and cipher as is. null cipher/protocol would just be handled as system defaults, which should conform to our standard. If we want to be extra sure that tests don't depend on the system, see my comment below.
</blockquote></details>

Done.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 149 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJd_PH1iepXClBNNmg:-MUPCIh498M3Deye47Gk:b-mv5oey) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L149)):*

> ```
>         sslProvider.equals(SslProvider.OPENSSL) ? "TLSv1.3" : "TLSv1.2",
>         null);
> ```
Since this helper is used by all existing tests that are not concerned with TLS versions or cipher suites, we should pick arguments that are most "normal". This means the version should be 1.2 (1.3 is still rare and not supported by the JDK SSL engine) and the cipher should be one of the common 1.2 cipher, using the name variant appropriate for the provider.
</blockquote></details>

changed to nulls since this would just use the system defaults
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 218 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJguKm3Do7A1czmChM:-MUPCS2H8YLOIf1kNRy2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L218)):*

> ```
>             sslProvider.equals(SslProvider.OPENSSL) ? "TLSv1.3" : "TLSv1.2",
>             "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"));
> ```
As I mentioned above, we should stick with v1.2 and the cipher's name needs to depend on the ssl provider used.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 78 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUUpm2N-YU4Lpou4mxq:-MUUpm2OBeTDaM8DGL0L:b-wvwndu) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L78)):*
> ```Java
>    * names can be found at ciphersuite.info.
>    */
>   private static final ImmutableList TLS_CIPHERS =
> ```

The name should be more specific, something like `ALLOWED_TLS_CIPHERS`.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 149 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJd_PH1iepXClBNNmg:-MUUpygOBWCY5DSEcTrB:b-fnmpet) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L149)):*

changed to nulls since this would just use the system defaults
</blockquote></details>

Do we want to use the system defaults?  Might that ever do the wrong thing?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 0 of 2 files reviewed, 8 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 136 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUV4m-sAeyIoqcxZf8G:-MUV4m-sAeyIoqcxZf8H:bau3uqq) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L136)):*
> ```Java
>             .clientAuth(requireClientCert ? ClientAuth.REQUIRE : ClientAuth.NONE)
>             .protocols(supportedSslVersions)
>             .ciphers(TLS_CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
> ```

Question for you @jianglai : Any reason this can't be gated behind an `if (DateTime.now(UTC) > ENFORCEMENT_START_DATE) { ... }` clause so that enforcement starts at a specific time of our choosing, rather than whenever we happen to roll this out?  And when are these requirements enforced exactly?  I'm guessing it's on the start of a new connection, so that all newly attempted connections after our chosen DateTime would start failing?

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 74 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUWGz3Z3G6GCZKRzafl:-MUWGz3Z3G6GCZKRzafm:b5snhde) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L74)):*
> ```
>    * no known security vulnerabilities. Note that OpenSSL uses a separate nomenclature for the
>    * ciphers than the ones used here. More information about these cipher suites and their OpenSSL
> ```
This sentence is somewhat misleading. How about "Note that OpenSSL uses a separate nomenclature for the ciphers internally but the IANA names listed here will be transparently translated by the OpenSSL provider (if used), so there is no need to include the OpenSSL name variants here."
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 136 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUV4m-sAeyIoqcxZf8G:-MUWN7vG-5aBtVrY0PcE:buq5huf) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L136)):*

Question for you @jianglai : Any reason this can't be gated behind an `if (DateTime.now(UTC) > ENFORCEMENT_START_DATE) { ... }` clause so that enforcement starts at a specific time of our choosing, rather than whenever we happen to roll this out?  And when are these requirements enforced exactly?  I'm guessing it's on the start of a new connection, so that all newly attempted connections after our chosen DateTime would start failing?
</blockquote></details>

Sarah and I were actually just discussion this over chat. Doing this via a config is certainly preferable. One of the complication is that the networking subproject doesn't have access to proxy's config. The proposed solution is to set the enforcement deadline in the proxy config and pass that deadline into the initializer's constructor. A clock should probably also be injected for testing. There will be some throwaway work involved but I do think it is worth it.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJVBXrArQ1rZiTZcdG:-MUWQ1zz0e9hbLYtq4CP:b-qp08oo) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L120)):*

Done.
</blockquote></details>

The ifs are not necessary. If the argument is null the method will take the default value.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 149 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJd_PH1iepXClBNNmg:-MUWT5saDtgf1gHZ9EzI:bui7ysj) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L149)):*

Do we want to use the system defaults?  Might that ever do the wrong thing?
</blockquote></details>

I think it does make sense here to use the default. This is server test so we tell the server a set of ciphers and TLS version it accepts, a reasonable client (not too old, like this one here) should pick a cipher and version that work with the server. The only place where we want to specify a version/cipher is to explicitly test that those not on the allowed list are forbidden.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 0 of 13 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 74 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUWGz3Z3G6GCZKRzafl:-MUjnMAv4PROUQGu6wlm:b-896fix) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L74)):*

> ```
>    * no known security vulnerabilities. Note that OpenSSL uses a separate nomenclature for the
>    * ciphers than the ones used here. More information about these cipher suites and their OpenSSL
> ```
This sentence is somewhat misleading. How about "Note that OpenSSL uses a separate nomenclature for the ciphers internally but the IANA names listed here will be transparently translated by the OpenSSL provider (if used), so there is no need to include the OpenSSL name variants here."
</blockquote></details>

Done.
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 78 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUUpm2N-YU4Lpou4mxq:-MUjnS9uCq6VVrKqo4E1:b-896fix) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L78)):*

The name should be more specific, something like `ALLOWED_TLS_CIPHERS`.
</blockquote></details>

Done.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/970#-MUJVBXrArQ1rZiTZcdG:-MUjyXKU5xVi5Z_5KdaW:bi680ya) ([raw file](https://github.com/google/nomulus/blob/f272c40d986e117b94b960613722a7bedde8ade1/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L120)):*

The ifs are not necessary. If the argument is null the method will take the default value.
</blockquote></details>

This works with the cipher, but when I tried with the protocol, it keeps giving errors that no protocol is specified.

jianglai (changes requested):
> Reviewed 13 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 134 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUkjT5K0WZnWC2sBpwj:-MUkjT5K0WZnWC2sBpwk:b-dglzqv) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L134)):*

> ```
>     this.oldSupportedSslVersions =
>         sslProvider == SslProvider.OPENSSL
>             ? ImmutableList.of("TLSv1.3", "TLSv1.2", "TLSv1.1", "TLSv1")
>             : ImmutableList.of("TLSv1.2", "TLSv1.1", "TLSv1");
> ```
</details>

This is not necessary. We currently already only support TLS 1.2 and up. We only need to configure even more stringent restrictions based on the enforcement time.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 123 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUkqDvBCJLSzzDFPwqv:-MUkqDvBCJLSzzDFPwqw:b-609vk1) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L123)):*
> ```
> .ciphers(cipher);
> ```
Let's add another test where the client supports three ciphers, with the first one not on the list and verify that the second cipher gets selected during handshake.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 233 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUknFvy1nTIo26qhPR5:-MUknFvy1nTIo26qhPR6:bo8j7f4) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L233)):*
> ```
> CIPHER_NOT_ACCEPTED_
> ```
Each local address name must be unique, otherwise we risk different tests connecting to each other's socket when running in parallel.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 296 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUknTjF82yk2g6vYnRQ:-MUknTjF82yk2g6vYnRR:b-p6icwu) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L296)):*
> ```
> PROTOCOL_NOT_ACCEPTED_
> ```
Same here.
___
*[proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java, line 147 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUklpp38DPmhxgvefha:-MUklpp38DPmhxgvefhb:b3fomu) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java#L147)):*
> ```
> DateTime.parse(config.tlsEnforcementStartTime)
> ```
We need to think about doing this more carefully. The web whois protocol is accessed via a browser and monitored by ICANN. Even though all modern browsers should support the cipher suites we chose,  I'm less certain about what ICANN's prober support. Maybe we should loop in Crystal on this? @cydeweys thoughts?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 134 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUkjT5K0WZnWC2sBpwj:-MUoZJB_22LLHD2nGC2e:b-vzktdf) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L134)):*

It is not true that we already only support TLS 1.2, there are currently 3 partners still connecting via TLSv1. I think the oldSupportedSslVersion variable is still necessary
</blockquote></details>

My bad. I was looking at the wrong diff.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 10 of 13 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 123 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUkqDvBCJLSzzDFPwqv:-MUoa2uM5hO7bgCWwN37:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L123)):*

> ```
> .ciphers(cipher);
> ```
Let's add another test where the client supports three ciphers, with the first one not on the list and verify that the second cipher gets selected during handshake.
</blockquote></details>

Done.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 233 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUknFvy1nTIo26qhPR5:-MUoXbci9aUsCNmITI9G:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L233)):*

> ```
> CIPHER_NOT_ACCEPTED_
> ```
Each local address name must be unique, otherwise we risk different tests connecting to each other's socket when running in parallel.
</blockquote></details>

Done.
___
*[networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java, line 296 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUknTjF82yk2g6vYnRQ:-MUoXbsp83HkC9sw1z8G:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/networking/src/test/java/google/registry/networking/handler/SslServerInitializerTest.java#L296)):*

> ```
> PROTOCOL_NOT_ACCEPTED_
> ```
Same here.
</blockquote></details>

Done.
___
*[proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java, line 147 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUklpp38DPmhxgvefha:-MUoabNf04tNTbVjLxdR:b-lslxdu) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java#L147)):*

> ```
> DateTime.parse(config.tlsEnforcementStartTime)
> ```
We need to think about doing this more carefully. The web whois protocol is accessed via a browser and monitored by ICANN. Even though all modern browsers should support the cipher suites we chose,  I'm less certain about what ICANN's prober support. Maybe we should loop in Crystal on this? @cydeweys thoughts?
</blockquote></details>

Ok, for now I changed to END_OF_TIME so that enforcement is never started for this module

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 10 of 13 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 136 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUV4m-sAeyIoqcxZf8G:-MUp5htZ23ti5wiFQdRQ:b-ld5vuc) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L136)):*

Sarah and I were actually just discussion this over chat. Doing this via a config is certainly preferable. One of the complication is that the networking subproject doesn't have access to proxy's config. The proposed solution is to set the enforcement deadline in the proxy config and pass that deadline into the initializer's constructor. A clock should probably also be injected for testing. There will be some throwaway work involved but I do think it is worth it.
</blockquote></details>

Given that we'll never need this config option again, and will remove it once this migration is complete, is it worth doing?
___
*[proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java, line 147 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUklpp38DPmhxgvefha:-MUp63Td6OxhsYlmtb_F:b-ui25kk) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java#L147)):*

Ok, for now I changed to END_OF_TIME so that enforcement is never started for this module
</blockquote></details>

Yeah, loop in Crystal.  I don't have strong feelings either way.

jianglai (approved):
> Reviewed 1 of 13 files at r3, 2 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 12 of 13 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[networking/build.gradle, line 46 at r4](https://reviewable.io/reviews/google/nomulus/970#-MUq0TGV8ghgziYgcsWi:-MUq0TGWEPCfex5FVYrl:b-vh2k6w) ([raw file](https://github.com/google/nomulus/blob/9b4620489f55122cbb4ed1f2c52bc6ff4687f65e/networking/build.gradle#L46)):*
> ```
>   
> ```
indentation.
___
*[networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java, line 136 at r2](https://reviewable.io/reviews/google/nomulus/970#-MUV4m-sAeyIoqcxZf8G:-MUq0XiR63t_Zy5psIvl:b-pbdqz3) ([raw file](https://github.com/google/nomulus/blob/194276887c88f2301cca37968525f03cf0998e13/networking/src/main/java/google/registry/networking/handler/SslServerInitializer.java#L136)):*

I think it is worth doing since it is the only way to set a separate start time for Sandbox and Production
</blockquote></details>

Agreed.
___
*[proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java, line 147 at r3](https://reviewable.io/reviews/google/nomulus/970#-MUklpp38DPmhxgvefha:-MUq1ADm-kwTNI9i0uyb:bws32mh) ([raw file](https://github.com/google/nomulus/blob/c971a01798f76382d07fe984c4c95ec9eaa37338/proxy/src/main/java/google/registry/proxy/WebWhoisProtocolsModule.java#L147)):*

Yeah, loop in Crystal.  I don't have strong feelings either way.
</blockquote></details>

I'm fine with `END_OF_TIME` for now. I think this PR is ready for submission. Let's ask Crystal via email. Basically we want to know if this is something that 1) we should do because some old browser may fail to connect to web WHOIS after the change and 2) we should notify ICANN about because their monitoring may fail (hopefully not).

jianglai (commented):
> Reviewed 1 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970)* status: 12 of 13 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[networking/build.gradle, line 46 at r4](https://reviewable.io/reviews/google/nomulus/970#-MUq0TGV8ghgziYgcsWi:-MUtQEg1BRCrq2Wf4BP5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9b4620489f55122cbb4ed1f2c52bc6ff4687f65e/networking/build.gradle#L46)):*

> ```
>   
> ```
indentation.
</blockquote></details>

Done.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/970#-:-MUtpRHKA7aT0XVrFuvu:bnfp4nl)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

PR #972, opened 2021-02-23
Add GetPremiumListCommand

gbrodman:
When testing the premium list refactor, it would have been nice and
convenient to have this. Currently we have no way of inspecting premium
list contents that I'm aware of that doesn't involve manually looking
at the Datastore or SQL entries.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/972)

### DIFF
--- a/core/src/main/java/google/registry/export/ExportPremiumTermsAction.java
+++ b/core/src/main/java/google/registry/export/ExportPremiumTermsAction.java
@@ -31,6 +31,7 @@ import com.google.common.flogger.FluentLogger;
 import com.google.common.net.MediaType;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.registry.Registry;
+import google.registry.model.registry.label.PremiumList.PremiumListEntry;
 import google.registry.model.registry.label.PremiumListDualDao;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
@@ -141,7 +142,7 @@ public class ExportPremiumTermsAction implements Runnable {
         PremiumListDualDao.exists(premiumListName), "Could not load premium list for " + tld);
     SortedSet<String> premiumTerms =
         Streams.stream(PremiumListDualDao.loadAllPremiumListEntries(premiumListName))
-            .map(entry -> Joiner.on(",").join(entry.getLabel(), entry.getValue()))
+            .map(PremiumListEntry::toString)
             .collect(ImmutableSortedSet.toImmutableSortedSet(String::compareTo));
 
     return Joiner.on("\n")
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -209,6 +209,12 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
     }
 
     @Override
+    public String toString() {
+      // Don't include the comment so that we can use this when exporting the premium list
+      return String.format("%s,%s", label, price);
+    }
+
+    @Override
     public Builder asBuilder() {
       return new Builder(clone(this));
     }
--- a/core/src/main/java/google/registry/tools/GetPremiumListCommand.java
+++ b/core/src/main/java/google/registry/tools/GetPremiumListCommand.java
@@ -1,0 +1,49 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.google.common.collect.Streams;
+import google.registry.model.registry.label.PremiumList.PremiumListEntry;
+import google.registry.model.registry.label.PremiumListDualDao;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/** Retrieves and prints one or more premium lists. */
+@Parameters(separators = " =", commandDescription = "Show one or more premium lists")
+public class GetPremiumListCommand implements CommandWithRemoteApi {
+
+  @Parameter(description = "Name(s) of the premium list(s) to retrieve", required = true)
+  private List<String> mainParameters;
+
+  @Override
+  public void run() throws Exception {
+    for (String premiumListName : mainParameters) {
+      if (PremiumListDualDao.exists(premiumListName)) {
+        System.out.printf(
+            "%s:\n%s\n",
+            premiumListName,
+            Streams.stream(PremiumListDualDao.loadAllPremiumListEntries(premiumListName))
+                .sorted(Comparator.comparing(PremiumListEntry::getLabel))
+                .map(PremiumListEntry::toString)
+                .collect(Collectors.joining("\n")));
+      } else {
+        System.out.println(String.format("No list found with name %s.", premiumListName));
+      }
+    }
+  }
+}
--- a/core/src/main/java/google/registry/tools/RegistryTool.java
+++ b/core/src/main/java/google/registry/tools/RegistryTool.java
@@ -76,6 +76,7 @@ public final class RegistryTool {
           .put("get_host", GetHostCommand.class)
           .put("get_keyring_secret", GetKeyringSecretCommand.class)
           .put("get_operation_status", GetOperationStatusCommand.class)
+          .put("get_premium_list", GetPremiumListCommand.class)
           .put("get_registrar", GetRegistrarCommand.class)
           .put("get_resource_by_key", GetResourceByKeyCommand.class)
           .put("get_routing_map", GetRoutingMapCommand.class)
--- a/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java
+++ b/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java
@@ -1,0 +1,71 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools;
+
+import static google.registry.testing.DatabaseHelper.createTld;
+import static org.junit.Assert.assertThrows;
+
+import com.beust.jcommander.ParameterException;
+import google.registry.testing.DualDatabaseTest;
+import google.registry.testing.TestOfyAndSql;
+import org.junit.jupiter.api.BeforeEach;
+
+@DualDatabaseTest
+public class GetPremiumListCommandTest extends CommandTestCase<GetPremiumListCommand> {
+
+  private static final String BASE_LIST_CONTENTS =
+      "tld:\n"
+          + "aluminum,USD 11.00\n"
+          + "brass,USD 20.00\n"
+          + "copper,USD 15.00\n"
+          + "diamond,USD 1000000.00\n"
+          + "gold,USD 24317.00\n"
+          + "iridium,USD 13117.00\n"
+          + "palladium,USD 877.00\n"
+          + "platinum,USD 87741.00\n"
+          + "rhodium,USD 88415.00\n"
+          + "rich,USD 100.00\n"
+          + "richer,USD 1000.00\n"
+          + "silver,USD 588.00\n";
+
+  @BeforeEach
+  void beforeEach() {
+    createTld("tld");
+  }
+
+  @TestOfyAndSql
+  void testSuccess_list() throws Exception {
+    runCommand("tld");
+    assertStdoutIs(BASE_LIST_CONTENTS);
+  }
+
+  @TestOfyAndSql
+  void testSuccess_onlyOneExists() throws Exception {
+    runCommand("tld", "nonexistent");
+    assertStdoutIs(BASE_LIST_CONTENTS + "No list found with name nonexistent.\n");
+  }
+
+  @TestOfyAndSql
+  void testFailure_nonexistent() throws Exception {
+    runCommand("nonexistent", "othernonexistent");
+    assertStdoutIs(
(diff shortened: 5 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/tools/GetPremiumListCommand.java, line 31 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUo7l178CVxj_tyijJ:-MUUo7l178CVxj_tyijK:b-g94uh7) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/main/java/google/registry/tools/GetPremiumListCommand.java#L31)):*
> ```Java
> public class GetPremiumListCommand implements CommandWithRemoteApi {
> 
>   public static final String PATH = "/_dr/admin/getPremiumList";
> ```

What is this about?
___
*[core/src/main/java/google/registry/tools/GetPremiumListCommand.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUoGMp4eDJrsAyNpXg:-MUUoGMp4eDJrsAyNpXh:b-97c4en) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/main/java/google/registry/tools/GetPremiumListCommand.java#L42)):*
> ```Java
>         ImmutableList<String> printedEntities =
>             Streams.stream(PremiumListDualDao.loadAllPremiumListEntries(premiumListName))
>                 .map(entry -> String.format("%s,%s", entry.getLabel(), entry.getValue()))
> ```

Implement this as a `.toString()` method on the `PremiumListEntry` instead.  This seems more widely applicable (and indeed it may already have some existing uses like this that may need to be refactored to use your new `toString()`).
___
*[core/src/main/java/google/registry/tools/GetPremiumListCommand.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUosrz5PUXAyY1cO9m:-MUUosrz5PUXAyY1cO9n:bv0i5s5) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/main/java/google/registry/tools/GetPremiumListCommand.java#L45)):*
> ```Java
>                 .collect(toImmutableList());
>         System.out.println(String.format("%s:", premiumListName));
>         System.out.println(Joiner.on("\n").join(printedEntities));
> ```

Instead of `collect()ing` to an ImmutableList above (which I suspect uses a bit more memory), you can use a concatenating String collector to directly collect to a string.  All of this then ends up looking like:

```
System.out.printf(
    "%s: %s\n",
    premiumListName,
    Streams.stream(PremiumListDualDao.loadAllPremiumListEntries(premiumListName).map(PremiumListEntry::toString).Collectors.joining("\n"));
```
___
*[core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUpVxWBEymvX9-PudF:-MUUpVxWBEymvX9-PudG:b-j097cb) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java#L37)):*
> ```Java
>     runCommand("tld");
>     // can't guarantee ordering of the list entries
>     assertInStdout("tld:\n");
> ```

Use `assertStdoutIs()` here to make a stronger more ironclad assertion, rather than merely asserting about what it might contain.
___
*[core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUpa3S8jRWjinMQvU5:-MUUpa3S8jRWjinMQvU6:b-816h6l) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java#L45)):*
> ```Java
>     runCommand("tld", "nonexistent");
>     // can't guarantee ordering of the list entries
>     assertInStdout("tld:\n");
> ```

Same.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/tools/GetPremiumListCommand.java, line 31 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUo7l178CVxj_tyijJ:-MUUtHfY-0NJtKlowzO9:b-uhcop5) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/main/java/google/registry/tools/GetPremiumListCommand.java#L31)):*

What is this about?
</blockquote></details>

sneaky lil bit of code, ignore it 
___
*[core/src/main/java/google/registry/tools/GetPremiumListCommand.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUoGMp4eDJrsAyNpXg:-MUUuljX8Hf8NCUo9f3D:bhcb2hi) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/main/java/google/registry/tools/GetPremiumListCommand.java#L42)):*

Implement this as a `.toString()` method on the `PremiumListEntry` instead.  This seems more widely applicable (and indeed it may already have some existing uses like this that may need to be refactored to use your new `toString()`).
</blockquote></details>

Done. (and one other usage) 
___
*[core/src/main/java/google/registry/tools/GetPremiumListCommand.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUosrz5PUXAyY1cO9m:-MUUvjOg7mWBmmO_idx4:b-hfluvg) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/main/java/google/registry/tools/GetPremiumListCommand.java#L45)):*

Instead of `collect()ing` to an ImmutableList above (which I suspect uses a bit more memory), you can use a concatenating String collector to directly collect to a string.  All of this then ends up looking like:

```
System.out.printf(
    "%s: %s\n",
    premiumListName,
    Streams.stream(PremiumListDualDao.loadAllPremiumListEntries(premiumListName).map(PremiumListEntry::toString).Collectors.joining("\n"));
```
</blockquote></details>

yeah, we can do something like that (not exactly what you wrote, but very close) 
___
*[core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUpVxWBEymvX9-PudF:-MUUwGmu6ObcA_3SaHSb:b-sshrun) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java#L37)):*

Use `assertStdoutIs()` here to make a stronger more ironclad assertion, rather than merely asserting about what it might contain.
</blockquote></details>

nah, see the comment (which I can make more detailed) -- we can't guarantee the ordering in Postgres unless we specifically sort it, which doesn't really seem to make sense for something like this 
___
*[core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUpa3S8jRWjinMQvU5:-MUUwRhv40QsUEtOJeQB:bsmfnc3) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java#L45)):*

Same.
</blockquote></details>

see other comment

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 213 at r2](https://reviewable.io/reviews/google/nomulus/972#-MUVI8FnElEzTJnTRHzS:-MUVI8FnElEzTJnTRHzT:b-8xw39w) ([raw file](https://github.com/google/nomulus/blob/9b869b24d6fa1851f8bcc6139bab7d7a5e995741/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L213)):*
> ```Java
>     @Override
>     public String toString() {
>       // Don't include the comment so that we can use this when exporting the premium list
> ```

Why is the comment being excluded?  This seems useful, no?  It doesn't make sense that you would save one thing but then get it back with one of that data missing?
___
*[core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUpVxWBEymvX9-PudF:-MUVIRFP1a7T1_1z2HJe:b-z616ox) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java#L37)):*

nah, see the comment (which I can make more detailed) -- we can't guarantee the ordering in Postgres unless we specifically sort it, which doesn't really seem to make sense for something like this 
</blockquote></details>

Disagree, it sounds like specifically sorting it would be super helpful.  We keep our .txt files sorted and the entire point of this PR is you want to be able to easily compare lists ... being sorted makes that a lot easier!

Phrased in a different way, why would randomly shuffled be preferable to a sensible ordering?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 213 at r2](https://reviewable.io/reviews/google/nomulus/972#-MUVI8FnElEzTJnTRHzS:-MUVItAn2i_d1ItGWMV-:b8tx5ly) ([raw file](https://github.com/google/nomulus/blob/9b869b24d6fa1851f8bcc6139bab7d7a5e995741/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L213)):*

Why is the comment being excluded?  This seems useful, no?  It doesn't make sense that you would save one thing but then get it back with one of that data missing?
</blockquote></details>

Hrm, it's because the comment is for internal use only which we don't want to include in the registrar-accessible export generated in `ExportPremiumTermsAction`, huh.

It sounds like the comment should be included here, and then we leave `ExportPremiumTermsAction` alone and continue to only do the label/price there.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 213 at r2](https://reviewable.io/reviews/google/nomulus/972#-MUVI8FnElEzTJnTRHzS:-MUVLIaFBJ5UXe80iYHW:b-vp3pc1) ([raw file](https://github.com/google/nomulus/blob/9b869b24d6fa1851f8bcc6139bab7d7a5e995741/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L213)):*

Hrm, it's because the comment is for internal use only which we don't want to include in the registrar-accessible export generated in `ExportPremiumTermsAction`, huh.

It sounds like the comment should be included here, and then we leave `ExportPremiumTermsAction` alone and continue to only do the label/price there.
</blockquote></details>

Talked offline and we'll keep this as is, since we're getting rid of the `comment` storage in the SQL world 
___
*[core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java, line 37 at r1](https://reviewable.io/reviews/google/nomulus/972#-MUUpVxWBEymvX9-PudF:-MUVMdp65wEIXj1UDGTS:bpf8mte) ([raw file](https://github.com/google/nomulus/blob/d04c8be1162074a9549ba3a96286dcc6effa1ed7/core/src/test/java/google/registry/tools/GetPremiumListCommandTest.java#L37)):*

Disagree, it sounds like specifically sorting it would be super helpful.  We keep our .txt files sorted and the entire point of this PR is you want to be able to easily compare lists ... being sorted makes that a lot easier!

Phrased in a different way, why would randomly shuffled be preferable to a sensible ordering?
</blockquote></details>

Yeah, we can sort it -- the lists should always be sorted anyway

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java, line 200 at r3](https://reviewable.io/reviews/google/nomulus/972#-MUp4hEL2BnGw8LT60ES:-MUp4hEL2BnGw8LT60ET:bnaftb6) ([raw file](https://github.com/google/nomulus/blob/f237292212318dd3027250c40577486be0ac1a4a/core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java#L200)):*
> ```Java
>                     .createQuery(
>                         "FROM PremiumEntry pe WHERE pe.revisionId = :revisionId "
>                             + "ORDER BY pe.domainLabel ASC",
> ```

So this wasn't exactly what I was suggesting, as the ordering isn't always required, but only in the get command and some other stuff.  But seeing as how this isn't in the main code path actually used for checking domain prices, eh.  It's not a big deal.  This particular code path isn't used for all that much anyway.

Note that you're only ordering the Cloud SQL code path, so what exactly is going on in the Datastore code path?  It's taking advantage of the fact that we happen to already sort lists prior to saving them in Datastore, but this isn't necessarily enforced anywhere?  It's a little bit messy.  Leaving the DAO alone, and then doing the sorting in-memory in just the `GetPremiumListCommand` (where it will sort the premium list equally regardless of which data source it's coming from) feels cleaner to me.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java, line 200 at r3](https://reviewable.io/reviews/google/nomulus/972#-MUp4hEL2BnGw8LT60ES:-MUp8uf7032vKY9-yfzG:bnd4agf) ([raw file](https://github.com/google/nomulus/blob/f237292212318dd3027250c40577486be0ac1a4a/core/src/main/java/google/registry/schema/tld/PremiumListSqlDao.java#L200)):*

So this wasn't exactly what I was suggesting, as the ordering isn't always required, but only in the get command and some other stuff.  But seeing as how this isn't in the main code path actually used for checking domain prices, eh.  It's not a big deal.  This particular code path isn't used for all that much anyway.

Note that you're only ordering the Cloud SQL code path, so what exactly is going on in the Datastore code path?  It's taking advantage of the fact that we happen to already sort lists prior to saving them in Datastore, but this isn't necessarily enforced anywhere?  It's a little bit messy.  Leaving the DAO alone, and then doing the sorting in-memory in just the `GetPremiumListCommand` (where it will sort the premium list equally regardless of which data source it's coming from) feels cleaner to me.
</blockquote></details>

Yeah, the lists are already sorted in Datastore. That seems reasonable, done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/972)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jianglai)

PR #975, opened 2021-02-24
Add replay & compare to more flow tests

mindhog:
Add replay and comparison testing to another batch of flow tests, apply fixes
as needed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/975)

### DIFF
--- a/core/src/test/java/google/registry/flows/contact/ContactCheckFlowTest.java
+++ b/core/src/test/java/google/registry/flows/contact/ContactCheckFlowTest.java
@@ -25,11 +25,18 @@ import google.registry.flows.ResourceCheckFlowTestCase;
 import google.registry.flows.exceptions.TooManyResourceChecksException;
 import google.registry.model.contact.ContactResource;
 import google.registry.testing.DualDatabaseTest;
+import google.registry.testing.ReplayExtension;
 import google.registry.testing.TestOfyAndSql;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 /** Unit tests for {@link ContactCheckFlow}. */
 @DualDatabaseTest
 class ContactCheckFlowTest extends ResourceCheckFlowTestCase<ContactCheckFlow, ContactResource> {
+
+  @Order(value = Order.DEFAULT - 2)
+  @RegisterExtension
+  final ReplayExtension replayExtension = ReplayExtension.createWithCompare(clock);
 
   ContactCheckFlowTest() {
     setEppInput("contact_check.xml");
--- a/core/src/test/java/google/registry/flows/contact/ContactDeleteFlowTest.java
+++ b/core/src/test/java/google/registry/flows/contact/ContactDeleteFlowTest.java
@@ -38,12 +38,19 @@ import google.registry.model.eppcommon.StatusValue;
 import google.registry.model.eppcommon.Trid;
 import google.registry.model.reporting.HistoryEntry;
 import google.registry.testing.DualDatabaseTest;
+import google.registry.testing.ReplayExtension;
 import google.registry.testing.TestOfyAndSql;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 /** Unit tests for {@link ContactDeleteFlow}. */
 @DualDatabaseTest
 class ContactDeleteFlowTest extends ResourceFlowTestCase<ContactDeleteFlow, ContactResource> {
+
+  @Order(value = Order.DEFAULT - 2)
+  @RegisterExtension
+  final ReplayExtension replayExtension = ReplayExtension.createWithCompare(clock);
 
   @BeforeEach
   void initFlowTest() {
--- a/core/src/test/java/google/registry/flows/contact/ContactInfoFlowTest.java
+++ b/core/src/test/java/google/registry/flows/contact/ContactInfoFlowTest.java
@@ -39,12 +39,19 @@ import google.registry.model.eppcommon.AuthInfo.PasswordAuth;
 import google.registry.model.eppcommon.PresenceMarker;
 import google.registry.model.eppcommon.StatusValue;
 import google.registry.testing.DualDatabaseTest;
+import google.registry.testing.ReplayExtension;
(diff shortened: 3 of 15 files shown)


gbrodman (changes requested):
> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/975)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java, line 1355 at r1](https://reviewable.io/reviews/google/nomulus/975#-MUP-8iHBh7dO8ehJ556:-MUP-8iHBh7dO8ehJ557:bduxve9) ([raw file](https://github.com/google/nomulus/blob/883d10670e82d1b83b70a0213337e22152d64e8f/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L1355)):*
> ```Java
>   }
> 
>   class SetClockExtension implements BeforeEachCallback {
> ```

Why not just do this in the beforeEach method of this class? I assume there's a reason, since I see you removing the clock-setting statement in another class. 

I don't think the replay extension does any operations involving the clock in its `beforeEach` (other than to just set it as a field in DatabaseHelper)

Also, it seem odd that we define this class multiple times with different clock start dates (in different test classes).

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/975)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java, line 1355 at r1](https://reviewable.io/reviews/google/nomulus/975#-MUP-8iHBh7dO8ehJ556:-MUPHmWW0MUBPZXHACEZ:b-gto8jr) ([raw file](https://github.com/google/nomulus/blob/883d10670e82d1b83b70a0213337e22152d64e8f/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L1355)):*

Why not just do this in the beforeEach method of this class? I assume there's a reason, since I see you removing the clock-setting statement in another class. 

I don't think the replay extension does any operations involving the clock in its `beforeEach` (other than to just set it as a field in DatabaseHelper)

Also, it seem odd that we define this class multiple times with different clock start dates (in different test classes). 
</blockquote></details>

yeah, it comes down to ordering -- this has to be called before the AppEngineExtension, which stores a few basic objects.  Note the "order" annotation above.  Do you think this merits a comment?

We could certainly define this class once and make it parameterizable, but I think we'd have to do so at The FlowTest level because it modifies the "clock" instance variable.  It didn't seem like much of a win to me at the time, given the size of this class and the amount of variation and the lower priortization of DRY in test code, but I wouldn't be averse to doing it that way if you think it preferable.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/975)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java, line 1355 at r1](https://reviewable.io/reviews/google/nomulus/975#-MUP-8iHBh7dO8ehJ556:-MUPMf_w2iK7mnQ2ZPER:b-jhpkd2) ([raw file](https://github.com/google/nomulus/blob/883d10670e82d1b83b70a0213337e22152d64e8f/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L1355)):*

yeah, it comes down to ordering -- this has to be called before the AppEngineExtension, which stores a few basic objects.  Note the "order" annotation above.  Do you think this merits a comment?

We could certainly define this class once and make it parameterizable, but I think we'd have to do so at The FlowTest level because it modifies the "clock" instance variable.  It didn't seem like much of a win to me at the time, given the size of this class and the amount of variation and the lower priortization of DRY in test code, but I wouldn't be averse to doing it that way if you think it preferable.
</blockquote></details>

Yeah, the repetition of code isn't a big deal, I was just curious. Is there a reason why we don't just choose some date that's very far in the past (or even the start of time)?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/975)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java, line 1355 at r1](https://reviewable.io/reviews/google/nomulus/975#-MUP-8iHBh7dO8ehJ556:-MUnop-B4K5-mWvJaV1G:b7iwztb) ([raw file](https://github.com/google/nomulus/blob/883d10670e82d1b83b70a0213337e22152d64e8f/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L1355)):*

Yeah, the repetition of code isn't a big deal, I was just curious. Is there a reason why we don't just choose some date that's very far in the past (or even the start of time)? 
</blockquote></details>

I'm not sure if this is specifically an issue for this test in particular, but generally for these test suites there are date comparisons that depend on the initial date specified in the test and dates derived from this initial date end up in places like the xml testdata files as well as the tests themselves.  In some cases, these values still need to be modified but it's nice to keep those modifications to a minimum and I didn't much care about using any particular start date.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/975)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java, line 1355 at r1](https://reviewable.io/reviews/google/nomulus/975#-MUP-8iHBh7dO8ehJ556:-MUsnRxL64mQomI-LDa0:be7us2v) ([raw file](https://github.com/google/nomulus/blob/883d10670e82d1b83b70a0213337e22152d64e8f/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L1355)):*

I'm not sure if this is specifically an issue for this test in particular, but generally for these test suites there are date comparisons that depend on the initial date specified in the test and dates derived from this initial date end up in places like the xml testdata files as well as the tests themselves.  In some cases, these values still need to be modified but it's nice to keep those modifications to a minimum and I didn't much care about using any particular start date.
</blockquote></details>

Yeah, I see what you mean. Sounds good.

PR #977, opened 2021-02-26
Convert DomainTransferRejectFlow to use tm() methods

gbrodman:
This change includes a few other necessary dependencies to converting
DomainTransferRejectFlowTest to be a dual-database test. Namely:

- The basic "use tm() instead of ofy()" and "branching database
selection on what were previously raw ofy queries"
- Modification of the PollMessage convertVKey methods to do what they
say they do
- Filling the generic pending / response fields in PollMessage based on what type of
poll message it is (this has to be done because SQL is not very good at
storing ambiguous superclasses)
- Setting the generic pending / repsonse fields in PollMessage upon
build
- Filling out the serverApproveEntities field in DomainTransferData with
all necessary poll messages / billing events that should be cancelled on
rejection
- Scattered changes in DatabaseHelper to make sure that we're saving and
loading entities correctly where we weren't before

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/977)

### DIFF
--- a/core/src/main/java/google/registry/flows/ResourceFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/ResourceFlowUtils.java
@@ -18,12 +18,11 @@ import static com.google.common.collect.Sets.intersection;
 import static google.registry.model.EppResourceUtils.getLinkedDomainKeys;
 import static google.registry.model.EppResourceUtils.loadByForeignKey;
 import static google.registry.model.index.ForeignKeyIndex.loadAndGetKey;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
-import com.googlecode.objectify.Key;
 import google.registry.flows.EppException.AuthorizationErrorException;
 import google.registry.flows.EppException.InvalidAuthorizationInformationErrorException;
 import google.registry.flows.EppException.ObjectDoesNotExistException;
@@ -185,16 +184,15 @@ public final class ResourceFlowUtils {
       return;
     }
     // The roid should match one of the contacts.
-    Optional<Key<ContactResource>> foundContact =
+    Optional<VKey<ContactResource>> foundContact =
         domain.getReferencedContacts().stream()
-            .map(VKey::getOfyKey)
-            .filter(key -> key.getName().equals(authRepoId))
+            .filter(key -> key.getOfyKey().getName().equals(authRepoId))
             .findFirst();
     if (!foundContact.isPresent()) {
       throw new BadAuthInfoForResourceException();
     }
     // Check the authInfo against the contact.
-    verifyAuthInfo(authInfo, ofy().load().key(foundContact.get()).now());
+    verifyAuthInfo(authInfo, transactIfJpaTm(() -> tm().loadByKey(foundContact.get())));
   }
 
   /** Check that the given {@link AuthInfo} is valid for the given contact. */
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -25,6 +25,9 @@ import static com.google.common.collect.Iterables.any;
 import static com.google.common.collect.Sets.difference;
 import static com.google.common.collect.Sets.intersection;
 import static com.google.common.collect.Sets.union;
+import static google.registry.model.DatabaseMigrationUtils.getPrimaryDatabase;
+import static google.registry.model.common.DatabaseTransitionSchedule.PrimaryDatabase.DATASTORE;
+import static google.registry.model.common.DatabaseTransitionSchedule.TransitionId.REPLAYED_ENTITIES;
 import static google.registry.model.domain.DomainBase.MAX_REGISTRATION_YEARS;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.model.registry.Registries.findTldForName;
@@ -38,6 +41,7 @@ import static google.registry.model.registry.label.ReservationType.FULLY_BLOCKED
 import static google.registry.model.registry.label.ReservationType.NAME_COLLISION;
 import static google.registry.model.registry.label.ReservationType.RESERVED_FOR_ANCHOR_TENANT;
 import static google.registry.model.registry.label.ReservationType.RESERVED_FOR_SPECIFIC_USE;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.pricing.PricingEngineProxy.isDomainPremium;
 import static google.registry.util.CollectionUtils.nullToEmpty;
@@ -88,6 +92,7 @@ import google.registry.model.domain.DomainCommand.Create;
 import google.registry.model.domain.DomainCommand.CreateOrUpdate;
 import google.registry.model.domain.DomainCommand.InvalidReferencesException;
 import google.registry.model.domain.DomainCommand.Update;
+import google.registry.model.domain.DomainHistory;
 import google.registry.model.domain.ForeignKeyedDesignatedContact;
 import google.registry.model.domain.Period;
 import google.registry.model.domain.fee.BaseFee;
@@ -355,22 +360,23 @@ public class DomainFlowUtils {
 
   static void validateNoDuplicateContacts(Set<DesignatedContact> contacts)
       throws ParameterValuePolicyErrorException {
-    ImmutableMultimap<Type, Key<ContactResource>> contactsByType =
+    ImmutableMultimap<Type, VKey<ContactResource>> contactsByType =
         contacts.stream()
             .collect(
                 toImmutableSetMultimap(
-                    DesignatedContact::getType, contact -> contact.getContactKey().getOfyKey()));
+                    DesignatedContact::getType, contact -> contact.getContactKey()));
 
     // If any contact type has multiple contacts:
     if (contactsByType.asMap().values().stream().anyMatch(v -> v.size() > 1)) {
       // Find the duplicates.
-      Map<Type, Collection<Key<ContactResource>>> dupeKeysMap =
+      Map<Type, Collection<VKey<ContactResource>>> dupeKeysMap =
           Maps.filterEntries(contactsByType.asMap(), e -> e.getValue().size() > 1);
-      ImmutableList<Key<ContactResource>> dupeKeys =
+      ImmutableList<VKey<ContactResource>> dupeKeys =
           dupeKeysMap.values().stream().flatMap(Collection::stream).collect(toImmutableList());
       // Load the duplicates in one batch.
-      Map<Key<ContactResource>, ContactResource> dupeContacts = ofy().load().keys(dupeKeys);
-      ImmutableMultimap.Builder<Type, Key<ContactResource>> typesMap =
+      Map<VKey<? extends ContactResource>, ContactResource> dupeContacts =
+          tm().loadByKeys(dupeKeys);
+      ImmutableMultimap.Builder<Type, VKey<ContactResource>> typesMap =
           new ImmutableMultimap.Builder<>();
       dupeKeysMap.forEach(typesMap::putAll);
       // Create an error message showing the type and contact IDs of the duplicates.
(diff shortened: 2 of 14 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUp6UFQ9UWncAiG8TYS:b6kx2gy) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*
> ```Java
>   private static List<? extends HistoryEntry> findRecentHistoryEntries(
>       DomainBase domainBase, DateTime now, Duration maxSearchPeriod) {
>     // TODO(gbrodman) replace this with the proper DatabaseTransitionSchedule-type check
> ```

Are we ready to make this change now?
___
*[core/src/main/java/google/registry/model/poll/PollMessage.java, line 414 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6uaz6d1YIOllJ8CG:-MUp6uaz6d1YIOllJ8CH:b-mkorgx) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/model/poll/PollMessage.java#L414)):*
> ```Java
>       super.postLoad();
>       if (pendingActionNotificationResponse != null) {
>         if (contactId != null) {
> ```

What's going on with all these changes here in this file?
___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 223 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp76ZHDPlP7V5Bh84r:-MUp76ZHDPlP7V5Bh84s:b-5unp7s) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L223)):*
> ```Java
>   @PostLoad
>   @Override
>   void postLoad() {
> ```

This persists indefinitely, not just until the migration is complete?  Ugh.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUpawiDAOqq8lkW7mOh:btm5ts3) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

Are we ready to make this change now?
</blockquote></details>

It is indeed, after merging that other PR. Done. 
___
*[core/src/main/java/google/registry/model/poll/PollMessage.java, line 414 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6uaz6d1YIOllJ8CG:-MUpe7urBphfRaLXgS1R:bt5ahzp) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/model/poll/PollMessage.java#L414)):*

What's going on with all these changes here in this file?
</blockquote></details>

Basically, PollMessage (more specifically, `OneTime`) has duplicate ways of storing the same set of data, kinda like the PremiumLists have. The  comment at the top of the class, from long ago (<=2016) says "Objectify cannot persist a base class type, so we must have a separate field to hold every possible derived type of ResponseData that we might store."

Basically, in SQL we can store the `pendingActionNotificationResponse` and `transferResponse` each in their own single field, but that's not possible in Objectify. 

The postLoad and onLoad methods basically take the fields that are persisted in one database and fill the other database's fields with the proper values, so that we can always be sure that we have the correctly-filled object. 

Similarly in `setResponseData`, we're just filling the duplicate fields.

Added a couple cmments. 
___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 223 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp76ZHDPlP7V5Bh84r:-MUperuU6A0UxpAv9Q-3:b49jk30) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L223)):*

This persists indefinitely, not just until the migration is complete?  Ugh.
</blockquote></details>

Maybe not; it'll be a LOT easier to refactor this to something nicer once the migration is done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUyq1Fo8wgnwHnIAwTo:bbrwprm) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

It is indeed, after merging that other PR. Done. 
</blockquote></details>

This line is surprisingly unreadable despite how simple it actually is.  If you reverse the order of the equals and statically import some stuff so that it fits on one line, it's a lot more readable.
___
*[core/src/main/java/google/registry/flows/domain/DomainTransferRejectFlow.java, line 104 at r2](https://reviewable.io/reviews/google/nomulus/977#-MUyqQcDEgDs8PtCT-Zs:-MUyqQcDEgDs8PtCT-Zt:b-uewufb) ([raw file](https://github.com/google/nomulus/blob/26608d39ec46d86041d46c93369fd01ae52b0311/core/src/main/java/google/registry/flows/domain/DomainTransferRejectFlow.java#L104)):*
> ```Java
>     DomainBase newDomain =
>         denyPendingTransfer(existingDomain, TransferStatus.CLIENT_REJECTED, now, clientId);
>     tm().putAll(
> ```

Might be worth having a varargs version of `tm().putAll()` so we don't need to do this `ImmutableList` wrapping.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUyxEow--O4f63mpgKm:b-sklo5x) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

This line is surprisingly unreadable despite how simple it actually is.  If you reverse the order of the equals and statically import some stuff so that it fits on one line, it's a lot more readable.
</blockquote></details>

The order of the .equals doesn't matter in terms of space but yeah, it's a lot better with static imports. Done 

fun fact: my old company had a guideline to basically never statically import anything 
___
*[core/src/main/java/google/registry/flows/domain/DomainTransferRejectFlow.java, line 104 at r2](https://reviewable.io/reviews/google/nomulus/977#-MUyqQcDEgDs8PtCT-Zs:-MUyy-JI90tgwdMxOnEo:baf33ic) ([raw file](https://github.com/google/nomulus/blob/26608d39ec46d86041d46c93369fd01ae52b0311/core/src/main/java/google/registry/flows/domain/DomainTransferRejectFlow.java#L104)):*

Might be worth having a varargs version of `tm().putAll()` so we don't need to do this `ImmutableList` wrapping.
</blockquote></details>

Good call, done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUzTo3r1PWBRQI38H-B:bb457n2) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

The order of the .equals doesn't matter in terms of space but yeah, it's a lot better with static imports. Done 

fun fact: my old company had a guideline to basically never statically import anything 
</blockquote></details>

The ordering is part of what makes it less readable.  This makes more sense to me as "get primary database equals X" rather than "X equals get primary data base"

In general you want to structure your conditionals like `x == 1`, not `1 == x` -- subject then value.  Value then subject means you don't actually know the context of what's being checked yet when you see the value, so you need to skip ahead to figure out what's going on, then read backwards to remember what value is being checked.  It's logically backwards, even if the equals operation is technically commutative.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUzf5UMA5-W7x-0vMqT:b-pz3er8) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

The ordering is part of what makes it less readable.  This makes more sense to me as "get primary database equals X" rather than "X equals get primary data base"

In general you want to structure your conditionals like `x == 1`, not `1 == x` -- subject then value.  Value then subject means you don't actually know the context of what's being checked yet when you see the value, so you need to skip ahead to figure out what's going on, then read backwards to remember what value is being checked.  It's logically backwards, even if the equals operation is technically commutative.
</blockquote></details>

yeah, i was biasing toward having the enum constant as the first argument to avoid any possibility of a NPE ever being thrown, but I suppose if the primary database is ever null we'll likely want to know about that immediately and as loud as possible

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MUzpLyv3_i41iMF6GaU:b-qdbphh) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

yeah, i was biasing toward having the enum constant as the first argument to avoid any possibility of a NPE ever being thrown, but I suppose if the primary database is ever null we'll likely want to know about that immediately and as loud as possible 
</blockquote></details>

`getPrimaryDatabase()` cannot return null (you can go look at how it's implemented, and also notice that it's not returning an `Optional` or `@Nullable`).

If it were somehow potentially returning null, then the correct solution would be to fix it so that it's not doing that, not to write your if statements wrong!

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/977)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 1083 at r1](https://reviewable.io/reviews/google/nomulus/977#-MUp6UFQ9UWncAiG8TYR:-MV-GJbk5ipGe6qerDut:b-lm2ogx) ([raw file](https://github.com/google/nomulus/blob/4f5fce5f7214b5492f01966b2c69344a0fe86e31/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1083)):*

`getPrimaryDatabase()` cannot return null (you can go look at how it's implemented, and also notice that it's not returning an `Optional` or `@Nullable`).

If it were somehow potentially returning null, then the correct solution would be to fix it so that it's not doing that, not to write your if statements wrong!
</blockquote></details>

I would disagree that having the enum constant be first is clearly wrong 100% of the time, though after reading more about it (e.g. https://en.wikipedia.org/wiki/Yoda_conditions, go/java-practices/null#tolerant) it does seem like we probably shouldn't be using it here or in most places 

That is, if you expect the first argument to possibly be null, use Objects.equals. Fair enough.

PR #966, opened 2021-02-12
Update a few plugins for Java 11 compatibility

jianglai:
Guice 5.0.0-BETA-1 is now compatible with Java 11. However we don't
directly depend on Guice. Rather Soy depends on Guice. So I added a
direct dependency on Guice 5.0 in the soy configuration in order to frontload
Soy's transitive dependency and pull in the newer version. Similarly Soy
depends on an old version of protobuf that is incompatible with Java 11, 
whereas the protobuf version that we use ourselves are OK. So an explicit
dependency for protobuf was added in the soy configuration as well.

Mockito 3.7.7 is now compatible with Java 11. The complication is that
we need to use the inline version of Mockito, which among other things
also allows mocking for final classes (hooray!). It will eventually
become the default Mockito mock maker but for now it needs to be
manually activated.

Note that the inline version now introduces another warning:

```
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
```

Which I think is WAI due to how the inline mock maker works. Waiting on
the author to confirm.

After these to changes the only illegal reflective access is caused by
App Engine SDK tools, which we will rid ourselves of when we migrate off
of GAE.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/966)

PR #981, opened 2021-03-02
Added "show_upgrade_diffs" script

mindhog:
"show_upgrade_diffs" pulls a git directory and a user branch from nomulus and
compares all of the versions of all dependencies specified in all lockfiles in
the master branch with those of the user branch and prints a nice, terse
little colorized report on the differences.

This is useful for reviewing a dependency upgrade.

### DIFF
--- a/config/show_upgrade_diffs.py
+++ b/config/show_upgrade_diffs.py
@@ -1,0 +1,145 @@
+"""Show the set of dependency diffs introduced by a branch.
+
+Usage:
+    show-upgrade-diffs.py <user> <branch> <directory>
+
+Assumes that there is a <user>/nomulus repository on github with the specified
+branch name.
+"""
+
+import os
+import six
+import subprocess
+import sys
+from typing import cast, Dict, Set, Tuple, Union
+
+
+def run(*args):
+    if subprocess.call(args):
+        raise Abort(f'"{" ".join(args)}" failed')
+
+
+PackageName = Tuple[str, str]
+PackageVersion = Tuple[str, str, str]
+VersionSet = Set[PackageVersion]
+PackageMap = Dict[PackageName, VersionSet]
+
+
+class Abort(Exception):
+    """Raised to abort processing and record an error."""
+
+
+def merge(dest: PackageMap, new: PackageMap) -> None:
+    for key, val in new.items():
+        dest[key] = dest.setdefault(key, set()) | val
+
+
+def parse_lockfile(filename: str) -> PackageMap:
+    result: PackageMap = {}
+    for line in open(filename, 'rb'):
+        if line.startswith(b'#'):
+            continue
+        line = line.rstrip()
+        package = cast(PackageVersion, tuple(line.split(b':')))
+        result.setdefault(package[:-1], set()).add(package[-1])
+    return result
+
+
+def get_all_package_versions(dir: str) -> PackageMap:
+    """Return list of all package versions in the directory."""
+    packages = {}
+    for file in os.listdir(dir):
+        file = os.path.join(dir, file)
+        if file.endswith('.lockfile'):
+            merge(packages, parse_lockfile(file))
+        elif os.path.isdir(file):
+            merge(packages, get_all_package_versions(file))
+    return packages
+
+
+def pr(*args: Union[str, bytes]) -> None:
+    """Print replacement that prints bytes without weird conversions."""
+    for text in args:
+        sys.stdout.buffer.write(six.ensure_binary(text))
+
+
+def format_versions(a: VersionSet, b: VersionSet, missing_esc: bytes) -> bytes:
+    """Returns a formatted string of the elements of "a".
+
+    Returns the elements of "a" as a comma-separated string, colorizes the
+    elements of "a" that are not also in "b" with "missing_esc".
+
+    Args:
+        a: Elements to print.
+        b: Other set, if a printed element is not a member of "b" it is
+            colorized.
+        missing_esc: ANSI terminal sequence to use to colorize elements that
+            are missing from "b".
+    """
+    elems = []
+    for item in a:
+        if item in b:
+            elems.append(item)
+        else:
+            elems.append(missing_esc + item + b'\033[0m')
+    return b', '.join(elems)
+
+
+def main():
+    # Print usage message on explicity help request or bad arguments.
+    want_help = len(sys.argv) > 1 and sys.argv[1] in ('-h', '--help')
+    if len(sys.argv) < 4 or want_help:
+        print(__doc__)
+        sys.exit(0 if want_help else 1)
+
+    user = sys.argv[1]
+    branch = sys.argv[2]
+    dir = sys.argv[3]
+
+    # Either clone or fetch the master branch if it exists.
+    if os.path.exists(dir):
+        pr(f'Reusing directory {dir}\n')
+        os.chdir(dir)
+        run('git', 'fetch', 'git@github.com:google/nomulus', 'master:master')
+        run('git', 'checkout', 'master')
+    else:
+        run('git', 'clone', 'git@github.com:google/nomulus', dir)
+        os.chdir(dir)
+
+    old_packages = get_all_package_versions('.')
+    run('git', 'fetch', f'https://github.com/{user}/nomulus.git',
+        f'{branch}:{branch}')
+    run('git', 'checkout', branch)
+    new_packages = get_all_package_versions('.')
+
+    if new_packages != old_packages:
+        pr('\n\nPackage version change report:\n')
+        pr('package-name: {old versions} -> {new versions}\n')
+        pr('==============================================\n\n')
+        for package, new_versions in new_packages.items():
+            old_versions = old_packages.get(package)
+            if new_versions != old_versions:
+
+                # Print out "package-name: {old versions} -> {new versions} with
+                # pretty colors.
+                formatted_old_versions = (
+                    format_versions(old_versions, new_versions,
+                                    b'\033[40;31;1m'))
+                formatted_new_versions = (
+                    format_versions(new_versions,
+                                    old_versions, b'\033[40;32;1m'))
+                pr(b':'.join(package), ': {', formatted_old_versions, '} -> {',
+                formatted_new_versions, '}\n')
+
+        # Print the list of packages that were removed.
+        for package in old_packages:
+            if package not in new_packages:
+                pr(b':'.join(package), ' removed\n')
+    else:
+        pr('Package versions not updated!\n')
+
+    pr(f'\nRetaining git directory {dir}, to delete: rm -rf {dir}\n')
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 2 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/981)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @mindhog)

___
*[show_upgrade_diffs, line 3 at r1](https://reviewable.io/reviews/google/nomulus/981#-MUsQuWP1fIcDQTmTSPz:-MUsQuWP1fIcDQTmTSQ-:b-fx37pp) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/show_upgrade_diffs#L3)):*
> ```
> $(dirname $0)/config/show_upgrade_diffs.py
> ```
Should this part be in quotes in case the parent directory contains some weird characters?

Also, does this file need a license header as well?
___
*[config/show_upgrade_diffs.py, line 37 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-bJsg8KPk5j4FbIZ5:-MV-bJsg8KPk5j4FbIZ6:b1viou0) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L37)):*
> ```
> Tuple[str, str, str]
> ```
Shouldn't `PackageVersion` just be a `str`?, 

`foo.bar:baz:x.y.z` -> `("foo.bar", "baz")` is `PackageName` and `"x.y.z"` is `PackageVersion`
___
*[config/show_upgrade_diffs.py, line 58 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-bbnR81RWONWjA7_N:-MV-bbnR81RWONWjA7_O:b-99016g) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L58)):*
> ```
> package[-1]
> ```
See here the `result`, which is a `Dict[PackageName, Set[PackageVersion]]` uses a `Tuple[str, str]` as key and `Set[str]` as value, further indicating that `PackageVersion` should be just a `str`.

Alternatively, I think it might be better to record not just the version number, but also in which subproject/configuration the version appears.

For example in your comment in #966, you have:

`b'com.google.code.findbugs:jsr305': {b'3.0.2\n', b'3.0.1\n', b'2.0.3\n'} -> {b'3.0.2\n', b'3.0.1\n'}` 

It tells me 2.0.3 is no longer used, but I have no idea where 2.0.3 was used in the first place. If we record the project and configuration it would be like "core:compileOnly:2.0.3" and it makes it very clear where this version of jsr305 was used in the old branch. But I understand that by doing this the output will probably become to verbose to consume, with each version appearing multiple times in different project/configuration combos.
___
*[config/show_upgrade_diffs.py, line 150 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-f5nUBz1DL3VXLjW3:-MV-f5nUBz1DL3VXLjW4:b61i5xt) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L150)):*

> ```
>         # Print the list of packages that were removed.
>         for package in old_packages:
>             if package not in new_packages:
>                 pr(b':'.join(package), ' removed\n')
> ```
</details>

Since newly added packages are printed as:

`package-name: {} -> {x.y.z, a,b,c, ...}`

Doesn't it make sense for removed package to be printed as:

`package-name: {x.y.z, a.b.c, ...} -> {}`?
___
*[config/show_upgrade_diffs.py, line 155 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-fP8S0gIlBGKXWnBe:-MV-fP8S0gIlBGKXWnBf:bqflq6a) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L155)):*
> ```
>     pr(f'\nRetaining git directory {dir}, to delete: rm -rf {dir}\n')
> ```
Since this tool always compares two checked-in remote branches, doesn't it make more sense to always automatically create a new temp directory, check out the branches, do the comparison, and delete the directory afterward? It feels a bit strange to reuse an existing git directory when the local changes are not used.

I think if we were to compare remote branches, the user should not be asked to provide a dir, and a random temp dir should be used and deleted. Currently the help message doesn't actually tell the user what the `<directory>` argument is for. If we were to ask the user to provide a directory, then the tool should compare the content in that directory with the nomulus master branch.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/981)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[config/show_upgrade_diffs.py, line 150 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-f5nUBz1DL3VXLjW3:-MV9eWroDjGcJImi8zL8:bytqn5p) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L150)):*

Indicating that the package was removed seems more straightforward to me.  Is there something specific that you prefer about the other format?
</blockquote></details>

Mostly for consistency. If I see "xyz removed" I might expect that added packages to be shown as "xyz added". Actually maybe that's what we should do -- keep "removed" and change how added packages are displayed.
___
*[config/show_upgrade_diffs.py, line 155 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-fP8S0gIlBGKXWnBe:-MV9f9664SN0ghvO4PyG:bht092w) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L155)):*

So, I've used this script a number of times at this point (I've had an earlier version that wasn't really ready for prime time for quite a while) and have often found myself pulling multiple commits from the review branch (most recently in one of your own PRs).  This is a lot quicker when I keep a copy of the tree around locally until the review is complete.

It's hard to say what usage model most people will be comfortable with.  If you have a specific pattern that you think would be preferable, we can certainly accommodate it, but I would like to retain the option to specify a directory for longer term use.  Do you still think the default should be a temporary directory? 
</blockquote></details>

I understand that cloning to an existing git directory saves time. Maybe the best way is to make the <directory> variable optional. If it exists, use it, if not, create a temporary one and delete it. What do you think? The help message should be a little more informative also, indicating what this variable does.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/981)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[config/show_upgrade_diffs.py, line 150 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-f5nUBz1DL3VXLjW3:-MVHIN12234c4M9ovXOy:b-5vg2c6) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L150)):*

Mostly for consistency. If I see "xyz removed" I might expect that added packages to be shown as "xyz added". Actually maybe that's what we should do -- keep "removed" and change how added packages are displayed.
</blockquote></details>

We don't typically add packages, the most common operation is to add and remove versions.  But I've changed the output format in the spirit of what you suggested.  Example output:

```
Package version change report:
change package-name: {old versions} -> {new versions}
=====================================================

added foo:added: {2.0, 1.0}
updated foo:updated: {1.5, 1.0} -> {2.0, 1.0}
removed foo:deleted
Retaining git directory <built-in function dir>, to delete: rm -rf <built-in function dir>
```
___
*[config/show_upgrade_diffs.py, line 155 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-fP8S0gIlBGKXWnBe:-MVHi0AgA56HPQSI_lDM:b655341) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L155)):*

I understand that cloning to an existing git directory saves time. Maybe the best way is to make the <directory> variable optional. If it exists, use it, if not, create a temporary one and delete it. What do you think? The help message should be a little more informative also, indicating what this variable does. 
</blockquote></details>

I've added a "-d" option that will preserve (and better document) the existing behavior, the new default behavior is to create a temporary directory which will be cleaned up after the run.  PTAL.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/981)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[config/show_upgrade_diffs.py, line 150 at r1](https://reviewable.io/reviews/google/nomulus/981#-MV-f5nUBz1DL3VXLjW3:-MVK3Toi2ajiu45xGKYx:brha71e) ([raw file](https://github.com/google/nomulus/blob/c0537b6066cc4857f93fe20736b1df0134a73666/config/show_upgrade_diffs.py#L150)):*

We don't typically add packages, the most common operation is to add and remove versions.  But I've changed the output format in the spirit of what you suggested.  Example output:

```
Package version change report:
change package-name: {old versions} -> {new versions}
=====================================================

added foo:added: {2.0, 1.0}
updated foo:updated: {1.5, 1.0} -> {2.0, 1.0}
removed foo:deleted
Retaining git directory <built-in function dir>, to delete: rm -rf <built-in function dir>
```
</blockquote></details>

This looks great. Thanks!
___
*[config/show_upgrade_diffs.py, line 112 at r3](https://reviewable.io/reviews/google/nomulus/981#-MVK0qRw3EEp0pUOVod8:-MVK0qRw3EEp0pUOVod9:bfp0xgj) ([raw file](https://github.com/google/nomulus/blob/a486767e2aaa4d49b702e9c2102d562525698ac4/config/show_upgrade_diffs.py#L112)):*
> ```
>  
> ```
Is there a reason why all full sentences are followed by two spaces here and below?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/981)* status: all files reviewed, 2 unresolved discussions

___
*[config/show_upgrade_diffs.py, line 112 at r3](https://reviewable.io/reviews/google/nomulus/981#-MVK0qRw3EEp0pUOVod8:-MVMJh_sCdliaC5qq78j:blwqnee) ([raw file](https://github.com/google/nomulus/blob/a486767e2aaa4d49b702e9c2102d562525698ac4/config/show_upgrade_diffs.py#L112)):*

> ```
>  
> ```
Is there a reason why all full sentences are followed by two spaces here and below?
</blockquote></details>

Sorry, I missed this -- that's a typesetter's convention that was drilled into a lot of us very early on.  It's actually [become controversial recently](https://www.google.com/search?q=two+spaces+after+period&oq=two+spaces+after+&aqs=chrome.0.0j69i57j0l3j69i64l3.2774j0j7&sourceid=chrome&ie=UTF-8) and, in fact, our markdown formatter will convert those to single spaces.

That said, it's pretty well wired into my fingers at this point and it certainly doesn't harm readability so I haven't felt inclined to consciously change it.

PR #982, opened 2021-03-02
Add SQL queries to RdapDomainSearchAction

gbrodman:
Unfortunately, because ORDER BY uses the locale's sorting functionality,
we end up with some weird sort orders in SQL-land (notably, periods are
ignored / omitted). As a result, a few of the tests have to be separated
out into ofy and SQL versions based on the expected sort order.

In addition, there isn't a way to query @Convert-ed fields in Postgres
via the standard Hibernate / JPA query language, meaning we have to use
a raw Postgres query for that. This mucks with our ability to query
timestamps directly, but we can approximate END_OF_TIME by using its
year.


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 520 at r1](https://reviewable.io/reviews/google/nomulus/982#-MUys2Qd5NHUCEdutkyH:-MUys2Qd5NHUCEdutkyI:b1fuwno) ([raw file](https://github.com/google/nomulus/blob/75a5016384442fa8c56a9a61f9efd3abba76ac7f/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L520)):*
> ```Java
>                                       + "ANY(h.inet_addresses) AND "
>                                       + "h.current_sponsor_registrar_id = :desiredRegistrar AND "
>                                       + "CAST(h.deletion_time AS text) LIKE :endOfTime")
> ```

This is seriously the best option??  Surely in native SQL there's better operations for dealing with timestamps than casting them to strings and then using string operators?  PostgreSQL timestamps don't support the `>` operator?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 520 at r1](https://reviewable.io/reviews/google/nomulus/982#-MUys2Qd5NHUCEdutkyH:-MUz6Y9h4dZyECGU9NJK:b4jeqpz) ([raw file](https://github.com/google/nomulus/blob/75a5016384442fa8c56a9a61f9efd3abba76ac7f/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L520)):*

This is seriously the best option??  Surely in native SQL there's better operations for dealing with timestamps than casting them to strings and then using string operators?  PostgreSQL timestamps don't support the `>` operator?
</blockquote></details>

The primary issue here is that since we're using the native query, `h.deletion_time` in Postgres is a Postgres timestamp, an `@Convert`ed column so we cannot compare it directly to our timestamp objects. 

We could, however, skip this and have the check in the Java code as a stream filter; this is probably preferable.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 520 at r1](https://reviewable.io/reviews/google/nomulus/982#-MUys2Qd5NHUCEdutkyH:-MUz6npSDvDLwkJnRKlM:bgqhmfg) ([raw file](https://github.com/google/nomulus/blob/75a5016384442fa8c56a9a61f9efd3abba76ac7f/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L520)):*

The primary issue here is that since we're using the native query, `h.deletion_time` in Postgres is a Postgres timestamp, an `@Convert`ed column so we cannot compare it directly to our timestamp objects. 

We could, however, skip this and have the check in the Java code as a stream filter; this is probably preferable. 
</blockquote></details>

Ah right, the reason why I did it this way was so that we would only load the `repoId` fields rather than having to load the entire object to check the deletion time. Hmm. i'll play around with it some more to see if there's something better but I'm not sure there is.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 520 at r1](https://reviewable.io/reviews/google/nomulus/982#-MUys2Qd5NHUCEdutkyH:-MUz9NOXEHxMf3DXLTuc:b7gpm8i) ([raw file](https://github.com/google/nomulus/blob/75a5016384442fa8c56a9a61f9efd3abba76ac7f/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L520)):*

Ah right, the reason why I did it this way was so that we would only load the `repoId` fields rather than having to load the entire object to check the deletion time. Hmm. i'll play around with it some more to see if there's something better but I'm not sure there is. 
</blockquote></details>

nvm there's a better way

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 290 at r2](https://reviewable.io/reviews/google/nomulus/982#-MVIATdk02fz8Tnh9IgH:-MVIATdk02fz8Tnh9IgI:b-3lo6zs) ([raw file](https://github.com/google/nomulus/blob/b37f0c5099323d7a865662667587aae2331b93dc/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L290)):*
> ```Java
>                                 DeletedItemHandling.INCLUDE)
>                             .orderBy(
>                                 jpaTm().getEntityManager().getCriteriaBuilder()::asc,
> ```

This really doesn't look good from the callsite.  What's going on here?  Any way to make these better?
___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 596 at r2](https://reviewable.io/reviews/google/nomulus/982#-MVIAeJvBPw97UzX4Z5-:-MVIAeJvBPw97UzX4Z50:b-ezif52) ([raw file](https://github.com/google/nomulus/blob/b37f0c5099323d7a865662667587aae2331b93dc/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L596)):*
> ```Java
>                       queryBuilder =
>                           queryBuilder.where(
>                               jpaTm().getEntityManager().getCriteriaBuilder()::greaterThan,
> ```

Yeah see, this is a LOT of unnecessary verbiage just to say greater than, less than, equal, asc, or desc.

Also this argument order doesn't really make sense.  Shouldn't it be `where("deletionTime", greaterThan, getRequestTime())` ?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 290 at r2](https://reviewable.io/reviews/google/nomulus/982#-MVIATdk02fz8Tnh9IgH:-MVIf431921thQiVEwMR:b-dtuchr) ([raw file](https://github.com/google/nomulus/blob/b37f0c5099323d7a865662667587aae2331b93dc/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L290)):*

This really doesn't look good from the callsite.  What's going on here?  Any way to make these better?
</blockquote></details>

yeah, since there are only really two options (ascending and descending) we can probably just put those methods on CriteriaQueryBuilder 
___
*[core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java, line 596 at r2](https://reviewable.io/reviews/google/nomulus/982#-MVIAeJvBPw97UzX4Z5-:-MVIlQQbAUItWmReVlzp:b-t6qy2r) ([raw file](https://github.com/google/nomulus/blob/b37f0c5099323d7a865662667587aae2331b93dc/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java#L596)):*

Yeah see, this is a LOT of unnecessary verbiage just to say greater than, less than, equal, asc, or desc.

Also this argument order doesn't really make sense.  Shouldn't it be `where("deletionTime", greaterThan, getRequestTime())` ?
</blockquote></details>

I think the switching of argument order does make sense, yeah. And it's a bit cleaner if we refactor out the acquisition of the CriteriaBuilder object into a separate line but I agree that it's unfortunately verbose regardless.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/982#-:-MVMgr1cCu1TqymdaDu1:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #987, opened 2021-03-03
Add SQL queries to RdapNameserverSearchAction

gbrodman:
This has the same issue as the domain-search action where the database
ordering is not consistent between Objectify and SQL -- as a result,
there is one test that we have to duplicate in order to account for the
two sort orders.

In addition, there isn't a way to query @Convert-ed fields in Postgres
via the standard Hibernate / JPA query language, meaning we have to use
a raw Postgres query for that.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/987)

### DIFF
--- a/config/presubmits.py
+++ b/config/presubmits.py
@@ -210,6 +210,7 @@ PRESUBMITS = {
          # CriteriaQueryBuilder is a false positive
          "CriteriaQueryBuilder.java",
          "RdapDomainSearchAction.java",
+         "RdapNameserverSearchAction.java",
          "RdapSearchActionBase.java",
          },
     ):
--- a/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java
+++ b/core/src/main/java/google/registry/rdap/RdapDomainSearchAction.java
@@ -26,6 +26,7 @@ import static google.registry.util.DateTimeUtils.START_OF_TIME;
 
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
@@ -489,40 +490,33 @@ public class RdapDomainSearchAction extends RdapSearchActionBase {
               .map(VKey::from)
               .collect(toImmutableSet());
     } else {
+      // Hibernate does not allow us to query @Converted array fields directly, either
+      // in the CriteriaQuery or the raw text format. However, Postgres does -- so we
+      // use native queries to find hosts where any of the inetAddresses match.
+      StringBuilder queryBuilder =
+          new StringBuilder(
+              "SELECT h.repo_id FROM \"Host\" h WHERE :address = ANY(h.inet_addresses) AND "
+                  + "h.deletion_time = CAST(:endOfTime AS timestamptz)");
+      ImmutableMap.Builder<String, String> parameters =
+          new ImmutableMap.Builder<String, String>()
+              .put("address", InetAddresses.toAddrString(inetAddress))
+              .put("endOfTime", END_OF_TIME.toString());
+      if (desiredRegistrar.isPresent()) {
+        queryBuilder.append(" AND h.current_sponsor_registrar_id = :desiredRegistrar");
+        parameters.put("desiredRegistrar", desiredRegistrar.get());
+      }
       hostKeys =
           jpaTm()
               .transact(
                   () -> {
-                    // Hibernate does not allow us to query @Converted array fields directly, either
-                    // in the CriteriaQuery or the raw text format. However, Postgres does -- so we
-                    // use native queries to find hosts where any of the inetAddresses match.
-                    javax.persistence.Query query;
-                    if (desiredRegistrar.isPresent()) {
-                      query =
-                          jpaTm()
-                              .getEntityManager()
-                              .createNativeQuery(
-                                  "SELECT h.repo_id FROM \"Host\" h WHERE :address = "
-                                      + "ANY(h.inet_addresses) AND "
-                                      + "h.current_sponsor_registrar_id = :desiredRegistrar AND "
-                                      + "h.deletion_time = CAST(:endOfTime AS timestamptz)")
-                              .setParameter("desiredRegistrar", desiredRegistrar.get());
-                    } else {
-                      query =
-                          jpaTm()
-                              .getEntityManager()
-                              .createNativeQuery(
-                                  "SELECT h.repo_id FROM \"Host\" h WHERE :address = "
-                                      + "ANY(h.inet_addresses) AND "
-                                      + "h.deletion_time = CAST(:endOfTime AS timestamptz)");
-                    }
+                    javax.persistence.Query query =
+                        jpaTm()
+                            .getEntityManager()
+                            .createNativeQuery(queryBuilder.toString())
+                            .setMaxResults(maxNameserversInFirstStage);
+                    parameters.build().forEach(query::setParameter);
                     @SuppressWarnings("unchecked")
-                    Stream<String> resultStream =
-                        query
-                            .setParameter("address", InetAddresses.toAddrString(inetAddress))
-                            .setParameter("endOfTime", END_OF_TIME.toString())
-                            .setMaxResults(maxNameserversInFirstStage)
-                            .getResultStream();
+                    Stream<String> resultStream = query.getResultStream();
                     return resultStream
                         .map(repoId -> VKey.create(HostResource.class, repoId))
                         .collect(toImmutableSet());
--- a/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java
+++ b/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java
@@ -15,9 +15,12 @@
 package google.registry.rdap;
 
 import static google.registry.model.EppResourceUtils.loadByForeignKey;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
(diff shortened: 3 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/987)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 289 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5RHC8h2JQMurWbrE:-MVb5RHDEqwJkhCgfouA:b-om3466) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L289)):*
> ```Java
>                     if (getDeletedItemHandling().equals(DeletedItemHandling.EXCLUDE)) {
>                       resultStream =
>                           resultStream.filter(h -> h.getDeletionTime().equals(END_OF_TIME));
> ```

Why are these filters happening in memory after the DB objects are loaded, rather than as part of the SQL query?
___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5bKF7WEVkg42EIhe:-MVb5bKF7WEVkg42EIhf:b-se3uti) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L294)):*
> ```Java
>                       // cursorString here must be the repo ID
>                       resultStream =
>                           resultStream.filter(h -> h.getRepoId().compareTo(cursorString.get()) > 0);
> ```

This seems very wrong to me?  You might potentially be doing a table scan for no reason, only to look at the value of one repo id in memory?  And you might exceed memory limits prior to doing so?

I think all of this needs a rethink -- it needs to happen in the SQL query.  (And were there an occurrences like this in previous PRs that I missed on code review?)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/987)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 289 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5RHC8h2JQMurWbrE:-MVs5UFY989Tyx_3M1SV:blolq0o) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L289)):*

Why are these filters happening in memory after the DB objects are loaded, rather than as part of the SQL query?
</blockquote></details>

ugh yeah, basically I was trying to avoid having to enumerate all of the possible options for the native queries but we might be able to get around this by using a StringBuilder to build up the query as well as a parameter map 
___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5bKF7WEVkg42EIhe:-MVs3fJb7EL4EZMVJV_a:b-fwvsr0) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L294)):*

This seems very wrong to me?  You might potentially be doing a table scan for no reason, only to look at the value of one repo id in memory?  And you might exceed memory limits prior to doing so?

I think all of this needs a rethink -- it needs to happen in the SQL query.  (And were there an occurrences like this in previous PRs that I missed on code review?)
</blockquote></details>

There are no occurrences of this in previous PRs that I'm aware of -- there, there was only one branch or so so it was possible to enumerate the native queries in raw strings without it being super messy.

I'd hope that since we're using a stream we'd avoid OOMing but yeah it's not ideal.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/987)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 289 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5RHC8h2JQMurWbrE:-MW-_1UbE5Bi3JeoWH_R:b-5oo9u1) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L289)):*

ugh yeah, basically I was trying to avoid having to enumerate all of the possible options for the native queries but we might be able to get around this by using a StringBuilder to build up the query as well as a parameter map 
</blockquote></details>

Sounds to me like the wrong trade-off was made.
___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5bKF7WEVkg42EIhe:-MW-ZWBE4hGaCs-lMlLl:bwcj0yt) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L294)):*

There are no occurrences of this in previous PRs that I'm aware of -- there, there was only one branch or so so it was possible to enumerate the native queries in raw strings without it being super messy.

I'd hope that since we're using a stream we'd avoid OOMing but yeah it's not ideal.  
</blockquote></details>

Far from merely not ideal, it would be actively terrible.  It'd be creating an excellent vector for a multiplicative DDOS (remember that RDAP is a public API).  It has to be performant or we can't spin it up at all.  "The code is harder to write this way" isn't a defense.  We can't be writing full table scan SQL queries over millions of records, period, and especially not on a public API!
___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 290 at r2](https://reviewable.io/reviews/google/nomulus/987#-MW-_P4l8Ru60EcSD-6z:-MW-_P4l8Ru60EcSD-7-:b-hwbdvi) ([raw file](https://github.com/google/nomulus/blob/c6ebee847e7f74eaff97f6fa32b9c78ee3a310e9/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L290)):*
> ```Java
>         parameters.put("desiredRegistrar", getDesiredRegistrar().get());
>       }
>       queryBuilder.append(" ORDER BY repo_id ASC");
> ```

No `LIMIT` here?  We're doing an unpaginated query?  This still seems too powerful to be exposed publicly.  Doesn't this still allow a full table scan to be requested, or at least something close enough to one?

You need to be thinking defensively when working with public APIs.  The Datastore queries have a `querySizeLimit` for a reason, but you haven't implemented that at all in these Cloud SQL queries?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/987)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 294 at r1](https://reviewable.io/reviews/google/nomulus/987#-MVb5bKF7WEVkg42EIhe:-MW5bBo13g71yKIBbgsD:b1zf55v) ([raw file](https://github.com/google/nomulus/blob/6a300a21c8e9a7f5eec589f53a30282d7d886d72/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L294)):*

Far from merely not ideal, it would be actively terrible.  It'd be creating an excellent vector for a multiplicative DDOS (remember that RDAP is a public API).  It has to be performant or we can't spin it up at all.  "The code is harder to write this way" isn't a defense.  We can't be writing full table scan SQL queries over millions of records, period, and especially not on a public API!
</blockquote></details>

For sure, and I think I was over-optimizing for simplicity of code and logic here. 

I don't know how soon we'll want to write a caching layer, but we'll probably want it eventually (there's already a bug for it) 
___
*[core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java, line 290 at r2](https://reviewable.io/reviews/google/nomulus/987#-MW-_P4l8Ru60EcSD-6z:-MW5bJvK4XREzwwjBDFU:bu0p1qt) ([raw file](https://github.com/google/nomulus/blob/c6ebee847e7f74eaff97f6fa32b9c78ee3a310e9/core/src/main/java/google/registry/rdap/RdapNameserverSearchAction.java#L290)):*

No `LIMIT` here?  We're doing an unpaginated query?  This still seems too powerful to be exposed publicly.  Doesn't this still allow a full table scan to be requested, or at least something close enough to one?

You need to be thinking defensively when working with public APIs.  The Datastore queries have a `querySizeLimit` for a reason, but you haven't implemented that at all in these Cloud SQL queries?
</blockquote></details>

In this native query no, not yet (done now though). In the CriteriaQueryBuilder queries we do, specifically see https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/rdap/RdapSearchActionBase.java#L208 

In the other situation where we use a native query (RdapDomainSearchAction) we also use a max number of results, and we can do the same here 

Changed this to use a max results param and changed RdapDomainSearchAction to look more like this w/r/t the StringBuilder

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/987#-:-MWP388w6yx2-90NmbOC:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #988, opened 2021-03-04
Refactor Cursor to exist in one class

gbrodman:
This allows us to get rid of the DAO as well as the sanity-checking
methods since we can be reasonably sure that the fields will be the
same. Future PRs will add conversions from ofy() to tm() calls that will
make sure that we get the same proper data in both Datastore and SQL

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/988)


sarahcaseybot (changes requested):
> Reviewed 1 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 1 of 24 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2kXCG26Kk-C_pAI5E:-MV2kXCG26Kk-C_pAI5F:b1ulft5) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/model/common/Cursor.java#L54)):*
> ```
> NonReplicatedEntity
> ```
Using NonReplicatedEntity indicates that you plan to still use Dual-Read/Dual-Write to migrate this entity, but that does not appear to be what's going on. It looks to me like the items are read and written to only one database based on what the TM is. 
___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 156 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2l4rpAU8i3ckxiahJ:-MV2l4rq-NabngbuCfp5:bi9qwyc) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/model/common/Cursor.java#L156)):*
> ```Java
>   UpdateAutoTimestamp lastUpdateTime = UpdateAutoTimestamp.create(null);
> 
>   @OnLoad
> ```

Are OnLoad and PostLoad only called when using Datastore?
___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2lV1g52HQjHnEdxei:-MV2lV1g52HQjHnEdxej:bi4hrcj) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/model/common/Cursor.java#L161)):*
> ```
> Key.valueOf(Splitter.on('_').splitToList(id).get(0))
> ```
Can this be moved to getScope() similar to how it is done for getType()?
___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line 0 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MV2q8eJBkcDY3wG9JRG:b-2p3933) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L0)):*
I don't agree with deleting this file since this is where we were actually writing and reading from both databases. Without the DAO, this current implementation only looks like its using one database at a time. Also, the DAO will make the actual cutover significantly easier if there is only one place that the primary database value needs to be looked up and handled.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 0 of 24 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2kXCG26Kk-C_pAI5E:-MVHduBgDWBmzdmfOtFc:b-pf234p) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/model/common/Cursor.java#L54)):*

> ```
> NonReplicatedEntity
> ```
Using NonReplicatedEntity indicates that you plan to still use Dual-Read/Dual-Write to migrate this entity, but that does not appear to be what's going on. It looks to me like the items are read and written to only one database based on what the TM is. 
</blockquote></details>

Yeah, I think `DatastoreAndSqlEntity` is the way to go 
___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 156 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2l4rpAU8i3ckxiahJ:-MVHdyLU9aU9KfalwZvX:b-u04tlj) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/model/common/Cursor.java#L156)):*

Are OnLoad and PostLoad only called when using Datastore?
</blockquote></details>

OnLoad is called when loading from Datastore, PostLoad is called when loading from SQL 
___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2lV1g52HQjHnEdxei:-MVHetqi2fK-DVQGOrro:b-dqh6u8) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/model/common/Cursor.java#L161)):*

> ```
> Key.valueOf(Splitter.on('_').splitToList(id).get(0))
> ```
Can this be moved to getScope() similar to how it is done for getType()?
</blockquote></details>

Actually, it should be the other way around. `getType` and `getScope` should simply return the simple type and scope variables, and `onLoad` should fill them both out. This is done
___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVHe7k-6cxJBj_CnHqg:b-9tta5l) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*

I don't agree with deleting this file since this is where we were actually writing and reading from both databases. Without the DAO, this current implementation only looks like its using one database at a time. Also, the DAO will make the actual cutover significantly easier if there is only one place that the primary database value needs to be looked up and handled.
</blockquote></details>

Assuming that Cursors are asynchronously populated (which they will be, now that Cursor is a single class that implements DatastoreAndSqlEntity), we do not wish to read and write from both databases at the same time. 

The way to test these is to use the `@DualDatabaseTest` annotation to run the tests in both datastore and sql, as well as to have the replay extension added to tests to verify that they get replayed properly.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVIBNjC7R1tWqMvVgVG:b3c7oli) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*

Assuming that Cursors are asynchronously populated (which they will be, now that Cursor is a single class that implements DatastoreAndSqlEntity), we do not wish to read and write from both databases at the same time. 

The way to test these is to use the `@DualDatabaseTest` annotation to run the tests in both datastore and sql, as well as to have the replay extension added to tests to verify that they get replayed properly. 
</blockquote></details>

So you're saying that `Cursor`s won't be part of the synchronous migration?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVILHBx7jM_xhTylCuy:bn8efna) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*

So you're saying that `Cursor`s won't be part of the synchronous migration?
</blockquote></details>

Is there any reason why they can't be asynchronously propagated like all of the other REPLAYED_ENTITIES ?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVN57iy69sO9JF3UO0_:by9u1ec) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*

Is there any reason why they can't be asynchronously propagated like all of the other REPLAYED_ENTITIES ? 
</blockquote></details>

No, but I thought that the plan had been to do the synchronous migration with these?  Hence all of the dual-read/dual-write code already written that you're deleting here?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 0 of 25 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVRyFoS1QjutH6ChV_u:b-jvy1w6) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*

No, but I thought that the plan had been to do the synchronous migration with these?  Hence all of the dual-read/dual-write code already written that you're deleting here?
</blockquote></details>

Yeah I think the reasoning for that was because we believed that the Datastore Cursor object and the SQL Cursor object were two distinct objects that needed to be saved and compared separately. However, that doesn't appear to need to be the case, and I think it's a simpler flow for this pattern, right?

sarahcaseybot (commented):
> Reviewed 21 of 24 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 23 of 25 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 23 of 25 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 166 at r2](https://reviewable.io/reviews/google/nomulus/988#-MVb3qhg9MnY5HnpBRYG:-MVb3qhg9MnY5HnpBRYH:b-90jdvp) ([raw file](https://github.com/google/nomulus/blob/6fad01c90e46265007faa43f29c75a0d501b49f2/core/src/main/java/google/registry/model/common/Cursor.java#L166)):*
> ```Java
>     // "Generate" the ID based on the scope and type
>     Key<? extends ImmutableObject> scopeKey;
>     if (GLOBAL.equals(scope)) {
> ```

Same comment here (see below).
___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 229 at r2](https://reviewable.io/reviews/google/nomulus/988#-MVb3Ecd-7TQ8hxz3uCG:-MVb3Ecd-7TQ8hxz3uCH:bnbwmua) ([raw file](https://github.com/google/nomulus/blob/6fad01c90e46265007faa43f29c75a0d501b49f2/core/src/main/java/google/registry/model/common/Cursor.java#L229)):*
> ```Java
>     // cross-tld-key) we have to parse the part of the ID
>     Key<?> scopeKey = Key.valueOf(idSplit.get(0));
>     if (getCrossTldKey().equals(scopeKey)) {
> ```

This conditional seems backwards to me.

(Not the first time I've noticed this -- I think you're optimizing for something else over readability, but please prefer readability.  In particular, in this case, `scopeKey` cannot be null, and if it were, it's just going to NPE in the `else` clause anyway.)
___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVb2lksEuiEhkg_Vi7f:bxpohni) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*
> we believed that the Datastore Cursor object and the SQL Cursor object were two distinct objects that needed to be saved and compared separately

They didn't need to be two separate objects, they just happened to be implemented that way.  Remember, this is going way back to the initial dual-read/dual-write pattern, prior to any planning on the async migration.  Whether it's two objects or not is an orthogonal concern to whether it does the sync/async migration.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 23 of 25 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 166 at r2](https://reviewable.io/reviews/google/nomulus/988#-MVb3qhg9MnY5HnpBRYG:-MVbytcsEMlK8bkapjGp:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6fad01c90e46265007faa43f29c75a0d501b49f2/core/src/main/java/google/registry/model/common/Cursor.java#L166)):*

Same comment here (see below).
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 229 at r2](https://reviewable.io/reviews/google/nomulus/988#-MVb3Ecd-7TQ8hxz3uCG:-MVbyrGz2_KDrMeBMAnG:bq3wo43) ([raw file](https://github.com/google/nomulus/blob/6fad01c90e46265007faa43f29c75a0d501b49f2/core/src/main/java/google/registry/model/common/Cursor.java#L229)):*

This conditional seems backwards to me.

(Not the first time I've noticed this -- I think you're optimizing for something else over readability, but please prefer readability.  In particular, in this case, `scopeKey` cannot be null, and if it were, it's just going to NPE in the `else` clause anyway.)
</blockquote></details>

Yeah, I think I wrote this before that conversation we had about this topic. Fixed. 
___
*[core/src/main/java/google/registry/schema/cursor/CursorDao.java, line  at r1](https://reviewable.io/reviews/google/nomulus/988#-MV2q8eJBkcDY3wG9JRF:-MVbQ0bm24xr8rJ8MNi7:b26ja2z) ([raw file](https://github.com/google/nomulus/blob/d12139a9b63e1e3ab4033e023212113fa996efd9/core/src/main/java/google/registry/schema/cursor/CursorDao.java#L)):*

> we believed that the Datastore Cursor object and the SQL Cursor object were two distinct objects that needed to be saved and compared separately

They didn't need to be two separate objects, they just happened to be implemented that way.  Remember, this is going way back to the initial dual-read/dual-write pattern, prior to any planning on the async migration.  Whether it's two objects or not is an orthogonal concern to whether it does the sync/async migration.
</blockquote></details>

Right, so basically as it stands, there is no place that we can say "give me a cursor, I don't care what database we're in". So, we need to either do a fairly heavy refactor and creation of a CursorDualDatabaseDao, or we can make them the same object, use `tm()` methods, and forgo the dual-read-dual-write-and-compare process entirely. 

Since keeping DRDW requires manual management and code changes later (and using `tm()` does not), it seems like moving to the simplified format now is ideal given that we have to make some amount of large changes regardless in order to get a functional mechanism to have some sort of API that gives a cursor regardless of DB

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/988)* status: 19 of 25 files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 166 at r2](https://reviewable.io/reviews/google/nomulus/988#-MVb3qhg9MnY5HnpBRYG:-MVc9nf8ECtoTlvYbuPC:b6bb1wu) ([raw file](https://github.com/google/nomulus/blob/6fad01c90e46265007faa43f29c75a0d501b49f2/core/src/main/java/google/registry/model/common/Cursor.java#L166)):*

Done.
</blockquote></details>

You could use a ternary here if you wanted.

```
Key<? extends ImmutableObject> scopeKey = (scope.equals(GLOBAL)) ? getCrossTldKey() | Key.create(getCrossTldKey(), Registry.class, scope);
```

sarahcaseybot (approved):

PR #989, opened 2021-03-04
Add a "ReplaySpecializer" to fix certain replays

mindhog:
Due to the fact that a given entity in either database type can map to
multiple entities in the other database, there are certain replication
scenarios that don't quite work.  Current known examples include:

- propagation of cascading deletes from datastore to SQL
- creation of datastore indexed entities for SQL entities (where indexes are a
  first-class concept)

This change introduces a ReplaySpecializer class, which allows us to declare
static method hooks at the entity class level that define any special
operations that need to be performed before or after replaying a mutation for
any given entity type.

Currently, "before SQL delete" is the only supported hook.  A change to
DomainContent demonstrating how this facility can be used to fix problems in
cascading delete propagation will be sent as a subsequent PR.

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -37,6 +37,7 @@ import google.registry.request.auth.Auth;
 import google.registry.schema.replay.DatastoreEntity;
 import google.registry.schema.replay.DatastoreOnlyEntity;
 import google.registry.schema.replay.NonReplicatedEntity;
+import google.registry.schema.replay.ReplaySpecializer;
 import google.registry.schema.replay.SqlReplayCheckpoint;
 import google.registry.util.RequestStatusChecker;
 import java.io.IOException;
@@ -179,6 +180,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
     if (!NonReplicatedEntity.class.isAssignableFrom(entityClass)
         && !DatastoreOnlyEntity.class.isAssignableFrom(entityClass)
         && entityClass.getAnnotation(javax.persistence.Entity.class) != null) {
+      ReplaySpecializer.beforeSqlDelete(entityVKey);
       jpaTm().delete(entityVKey);
     }
   }
--- a/core/src/main/java/google/registry/model/ofy/ReplayQueue.java
+++ b/core/src/main/java/google/registry/model/ofy/ReplayQueue.java
@@ -24,6 +24,7 @@ import google.registry.config.RegistryEnvironment;
 import google.registry.model.UpdateAutoTimestamp;
 import google.registry.persistence.VKey;
 import google.registry.schema.replay.DatastoreEntity;
+import google.registry.schema.replay.ReplaySpecializer;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -100,7 +101,9 @@ public class ReplayQueue {
                     .forEach(
                         entry -> {
                           if (entry.getValue().equals(TransactionInfo.Delete.SENTINEL)) {
-                            jpaTm().delete(VKey.from(entry.getKey()));
+                            VKey<?> vkey = VKey.from(entry.getKey());
+                            ReplaySpecializer.beforeSqlDelete(vkey);
+                            jpaTm().delete(vkey);
                           } else {
                             ((DatastoreEntity) entry.getValue())
                                 .toSqlEntity()
--- a/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java
+++ b/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java
@@ -1,0 +1,47 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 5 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/989)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/989#-MVI1Bae0AKJOdAdt1FQ:-MVI1Bae0AKJOdAdt1FR:b700dwx) ([raw file](https://github.com/google/nomulus/blob/1c871748500417d28de93d586f17b4af03806884/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java#L39)):*
> ```Java
>       // Ignore, this just means that the class doesn't need this hook.
>     } catch (IllegalAccessException e) {
>       logger.atSevere().log(
> ```

It seems like we might wish to re-throw the exception for IllegalAccessException and InvocationTargetException so that if someone messes up the creation of the beforeSqlDelete method in a method it fails loudly and obviously for that reason

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/989)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/989#-MVI1Bae0AKJOdAdt1FQ:-MVI3Bh0CVNaz0gaeqp1:bs7wumx) ([raw file](https://github.com/google/nomulus/blob/1c871748500417d28de93d586f17b4af03806884/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java#L39)):*

It seems like we might wish to re-throw the exception for IllegalAccessException and InvocationTargetException so that if someone messes up the creation of the beforeSqlDelete method in a method it fails loudly and obviously for that reason 
</blockquote></details>

Done.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/989)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[config/presubmits.py, line 93 at r2](https://reviewable.io/reviews/google/nomulus/989#-MVI6GED8xXvmktos5eT:-MVI6GED8xXvmktos5eU:b-hgvp7e) ([raw file](https://github.com/google/nomulus/blob/fcc5bd02da8dbc6d2218e2474f5a8b98b11a523a/config/presubmits.py#L93)):*
> ```Python
>         "Source files must end in a newline.",
> 
> #    # System.(out|err).println should only appear in tools/
> ```

i think this file made it in by mistake 
___
*[core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/989#-MVI1Bae0AKJOdAdt1FQ:-MVI6Ldt4bs8lzKRliUr:b-8leqgt) ([raw file](https://github.com/google/nomulus/blob/1c871748500417d28de93d586f17b4af03806884/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java#L39)):*

Done. 
</blockquote></details>

Can we make the exceptions more detailed? If we're throwing, we also probably don't need the log message, we can just throw the exception with the current log message (or something very similar)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/989)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[config/presubmits.py, line 93 at r2](https://reviewable.io/reviews/google/nomulus/989#-MVI6GED8xXvmktos5eT:-MVIR1X2-sioemSslykh:b-j3owmd) ([raw file](https://github.com/google/nomulus/blob/fcc5bd02da8dbc6d2218e2474f5a8b98b11a523a/config/presubmits.py#L93)):*

i think this file made it in by mistake 
</blockquote></details>

Thanks, I keep meaning to add a commit hook to guard against these local hacks.
___
*[core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/989#-MVI1Bae0AKJOdAdt1FQ:-MVIQj_VALoxO2bnkQuG:bocbzis) ([raw file](https://github.com/google/nomulus/blob/1c871748500417d28de93d586f17b4af03806884/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java#L39)):*

Can we make the exceptions more detailed? If we're throwing, we also probably don't need the log message, we can just throw the exception with the current log message (or something very similar) 
</blockquote></details>

Done.

I honestly don't care about the details of how we report this as long as it shows up in test output and logfiles.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/989)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/989#-MVI1Bae0AKJOdAdt1FQ:-MVIY4An0y7hX5SeF5UH:bou7itp) ([raw file](https://github.com/google/nomulus/blob/1c871748500417d28de93d586f17b4af03806884/core/src/main/java/google/registry/schema/replay/ReplaySpecializer.java#L39)):*

Done.

I honestly don't care about the details of how we report this as long as it shows up in test output and logfiles. 
</blockquote></details>

sweet, lgtm

PR #1001, opened 2021-03-10
Don't use --fork-point to determine merge base

mindhog:
It turns out that the --fork-point option is subtle and error-prone.  Its
intent is not to show the nearest common base commit, but rather the commit
on a branch that the HEAD (in this case) was originally forked off of,
_whether it is currently part of the history of the specified branch or not_
(this can happen if the branch is rewritten).  The option also relies on the
presence of the fork point in the reflog for the branch, which can be
discarded in the course of a "git gc".

It is fairly easy to construct a case where the use of --fork-point causes an
error and outputs nothing.  In fact, I discovered the problem as a result of
this occuring spontaneously on one of my own branches (likely related to a
rebase).  Since the fork-point is empty, we end up diffing against the index
instead of the common commit.

This may have been a factor in some of the unrelated reformatting that we've
seen in past PRs.

Change this to a simple "merge-base origin/master HEAD", which outputs the
commit id of the most recent common base revision.

This change also quotes the forkPoint variable, which likely would have
resulted in an error in this case instead of silently producing the wrong
output.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1001)

### DIFF
--- a/java-format/google-java-format-git-diff.sh
+++ b/java-format/google-java-format-git-diff.sh
@@ -55,7 +55,7 @@ function showNoncompliantFiles() {
 
 function callGoogleJavaFormatDiff() {
   local forkPoint
-  forkPoint=$(git merge-base --fork-point origin/master)
+  forkPoint=$(git merge-base origin/master HEAD)
 
   local callResult
   case "$1" in


weiminyu (commented):
IIRC, "git merge-base origin/master HEAD" can cause problems with local workflow. When you have a long lasting branch and origin/master has been updated, you may see lines that are changed or deleted in origin reformatted in this branch. A rebase would fix that but it is not always convenient.

Maybe we can have two versions of the  task, invoke the current one if local, and the new form if on kokoro.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1001#-:-MVWljiP6lfZK1ck615q:b-awpt1f)* status: 0 of 1 files reviewed, all discussions resolved

mindhog (commented):
I think you may be confusing the results of merge-base with diff.  merge-base is just the most recent point of divergence between the two commits.  It should not change as a result of additions to origin/master.  Also note that this change has nothing to do with kokoro: I found the problem in my local workflow.

For our feature branches, the structure generally looks like this:

```
----A---B--- origin/master
     \
      C---D--- feature-branch
```

When in a git directory checked out from `feature-branch`, "merge-base 
origin/master HEAD" will always return A, regardless of changes to 
`origin/master` (unless, of course, the history of `origin/master` is changed, 
but we never do that) until you rebase or merge `feature-branch`.

Since we're diffing from current directory to the merge-base, we should only 
see the diffs introduced by feature-branch.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1001#-:-MVafaQFB3raY7XziNr0:b-rfgbkq)* status: 0 of 1 files reviewed, all discussions resolved

weiminyu (approved):
I can't recall the problem I was trying to solve with --fork-point. Let's make this change and see.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1001#-:-MVauToV6BtpHQ9rMlZY:b-2mp6mj)* status: 0 of 1 files reviewed, all discussions resolved

PR #1005, opened 2021-03-11
Add TODOs regarding cloud sql database name change

weiminyu:
We should choose a different database name for nomulus data since using
'postgres' is bad practice. See b/181693544 more background.

We have decided to delay the db change to the time when we upgrade
postgresql version. This PR adds TODOs to all occurrences of the jdbcUrl
property, including those in the internal-repo. This property will change
when we upgrade, so the TODOs will be noticed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1005)

### DIFF
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -225,6 +225,10 @@ hibernate:
 
 cloudSql:
   # jdbc url for the Cloud SQL database.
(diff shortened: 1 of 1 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1005)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1008, opened 2021-03-13
Use shared jar to stage BEAM pipeline if possible

weiminyu:
Allow multiple BEAM pipelines with the same classes and dependencies to
share one Uber jar.

Added metadata for BulkDeleteDatastorePipeline.

Updated shell and Cloud Build scripts to stage all pipelines in one
step.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1008)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -751,7 +751,8 @@ project.tasks.create('initSqlPipeline', JavaExec) {
 // nom_build :core:bulkDeleteDatastore --args="--project=domain-registry-crash \
 //     --region=us-central1 --runner=DataflowRunner --kindsToDelete=*"
 createToolTask(
-    'bulkDeleteDatastore', 'google.registry.beam.datastore.BulkDeletePipeline')
+    'bulkDeleteDatastore',
+    'google.registry.beam.datastore.BulkDeleteDatastorePipeline')
 
 project.tasks.create('generateSqlSchema', JavaExec) {
   classpath = sourceSets.nonprod.runtimeClasspath
@@ -782,10 +783,13 @@ generateGoldenImages.finalizedBy(findGoldenImages)
 
 createUberJar('nomulus', 'nomulus', 'google.registry.tools.RegistryTool')
 
+// Build the Uber jar shared by all flex-template based BEAM pipelines.
+// This packages more code and dependency than necessary. However, without
+// restructuring the source tree it is difficult to generate leaner jars.
 createUberJar(
-    'init_sql_pipeline',
-    'init_sql_pipeline',
-    'google.registry.beam.initsql.InitSqlPipeline')
+    'beam_pipeline_common',
+    'beam_pipeline_common',
+    '')
 
 // A jar with classes and resources from main sourceSet, excluding internal
 // data. See comments on configurations.nomulus_test above for details.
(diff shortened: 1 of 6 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1008)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman, @jianglai, and @weiminyu)

___
*[release/stage_beam_pipeline.sh, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1008#-MVv1bATAbd0CyzN1BT3:-MVv1bATAbd0CyzN1BT4:b-unt2i9) ([raw file](https://github.com/google/nomulus/blob/5d86c1c84322b2654c2912922387542270425e69/release/stage_beam_pipeline.sh#L54)):*
> ```Shell
> 
> maven_gcs_prefix="gcs://domain-registry-maven-repository"
> ./gradlew clean :core:"${uberjar_name}" \
> ```

This invocation will run gradlew from the current directory, which I don't think you want because it will fail if someone runs this script from anywhere but the project root.  If you instead do:

```
mydir=$(dirname $0)

$(mydir)/../gradlew <args>
```

Then the script will work from anywhere (as long as it's not removed from the source tree).

Also, consider invoking "nom_build" instead, especially since you're passing properties in and nom_build will fail with an error if those properties get renamed instead of silently ignoring them. (use --mavenUrl and --pluginsUrl instead)   
___
*[release/stage_beam_pipeline.sh, line 63 at r1](https://reviewable.io/reviews/google/nomulus/1008#-MVv4V9xAWbkKZEXTHxC:-MVv4V9xAWbkKZEXTHxD:b-m64pze) ([raw file](https://github.com/google/nomulus/blob/5d86c1c84322b2654c2912922387542270425e69/release/stage_beam_pipeline.sh#L63)):*
> ```Shell
>   # Get main_class' simple name in lower_underscore form
>   pipeline_name=$(
>       echo "${main_class}" | rev | cut -d. -f1 | rev | \
> ```

LOL, nice! TIL there's a "rev" command :-)

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1008)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @gbrodman, @jianglai, and @mindhog)

___
*[release/stage_beam_pipeline.sh, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1008#-MVv1bATAbd0CyzN1BT3:-MVvOPJ_-GT3mvMqbRBT:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5d86c1c84322b2654c2912922387542270425e69/release/stage_beam_pipeline.sh#L54)):*

This invocation will run gradlew from the current directory, which I don't think you want because it will fail if someone runs this script from anywhere but the project root.  If you instead do:

```
mydir=$(dirname $0)

$(mydir)/../gradlew <args>
```

Then the script will work from anywhere (as long as it's not removed from the source tree).

Also, consider invoking "nom_build" instead, especially since you're passing properties in and nom_build will fail with an error if those properties get renamed instead of silently ignoring them. (use --mavenUrl and --pluginsUrl instead)   
</blockquote></details>

Done.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1008)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman and @jianglai)

PR #1014, opened 2021-03-16
Allow java-format to use java from the PATH

mindhog:
When invoking java from the google-java-format-git-diff.sh script, if there is
no JAVA_HOME environment variable, attempt to instead run the java binary that
is on the PATH.

This also adds a few checks to verify that a java binary is available in one
of those locations and that the version discovered is Java 11 (which we know
to be compatible with the google-java-format jar).

Tested:
- unset JAVA_HOME, verified that we get the version on the PATH
- Set JAVA_HOME to an invalid directory, verified that we get an error.
- Changed the "which" command to lookup an nonexistent binary, unset JAVA_HOME
  and verified that we get a "java not found" error.
- Changed the path to point to an old version of java, verified that we get a
  "bad java version" error.
- Verified that the script still runs normally.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1014)

### DIFF
--- a/java-format/google-java-format-git-diff.sh
+++ b/java-format/google-java-format-git-diff.sh
@@ -42,6 +42,28 @@ where:
 SCRIPT_DIR="$(realpath $(dirname $0))"
 JAR_NAME="google-java-format-1.8-all-deps.jar"
 
+# Locate the java binary.
+if [ -n "$JAVA_HOME" ]; then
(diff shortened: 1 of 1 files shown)


jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1014)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1015, opened 2021-03-16
Clear autorenew end time when a domain is restored

CydeWeys:
This allows us to still see in the database which now-deleted domains had
reached expiration, while correctly not re-deleting the domain immediately if
the registrar pays to explicitly restore the domain.

This also resolves some TODOs around data migration for this field on domain so
that it's not null, as said migration has already been completed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1015)


jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1015)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1017, opened 2021-03-16
Remove cert hash checks

sarahcaseybot:
Cert hash was being used as a fallback if there was any issue with the certificate checks. The logs don't show that the hash fall back was needed at all, so I am removing it. I also removed the hardcoded start date for certificate enforcement since the date has passed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1017)

### DIFF
--- a/core/src/main/java/google/registry/flows/TlsCredentials.java
+++ b/core/src/main/java/google/registry/flows/TlsCredentials.java
@@ -26,7 +26,6 @@ import com.google.common.net.InetAddresses;
 import dagger.Module;
 import dagger.Provides;
 import google.registry.config.RegistryConfig.Config;
-import google.registry.config.RegistryEnvironment;
 import google.registry.flows.EppException.AuthenticationErrorException;
 import google.registry.flows.certs.CertificateChecker;
 import google.registry.flows.certs.CertificateChecker.InsecureCertificateException;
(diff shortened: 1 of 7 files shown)


jianglai (changes requested):
> Reviewed 2 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 72 at r1](https://reviewable.io/reviews/google/nomulus/1017#-MW-cngl22of4JEcYXUe:-MW-cngl22of4JEcYXUf:bka1teg) ([raw file](https://github.com/google/nomulus/blob/e50fdb5d47d330287dfcb55a3434705405bb00b1/core/src/main/java/google/registry/flows/TlsCredentials.java#L72)):*
> ```
> private final Optional<String> clientCertificateHash;
> ```
Since we are only relying on the full certificate passed from the proxy, this doesn't seem necessary anymore, right?
___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 167 at r1](https://reviewable.io/reviews/google/nomulus/1017#-MW-XF9A5LaWB-WYxX72:-MW-XF9A5LaWB-WYxX73:bnt4m2j) ([raw file](https://github.com/google/nomulus/blob/e50fdb5d47d330287dfcb55a3434705405bb00b1/core/src/main/java/google/registry/flows/TlsCredentials.java#L167)):*
> ```
>   
> ```
Wrong indentation.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/flows/TlsCredentials.java, line 72 at r1](https://reviewable.io/reviews/google/nomulus/1017#-MW-cngl22of4JEcYXUe:-MW5h_kK7TMUIOX4GuSf:bjzpthy) ([raw file](https://github.com/google/nomulus/blob/e50fdb5d47d330287dfcb55a3434705405bb00b1/core/src/main/java/google/registry/flows/TlsCredentials.java#L72)):*

Done.
</blockquote></details>

Actually, I decided to keep this since hash value in the logs is valuable for debugging

jianglai (changes requested):
This PR itself is fine. But I would like to revisit the discussion we has previously in the design doc on passing the full certificate. Now that it looks like we do want to keep the certificate hash in the header for all requests, I don't think there is a reason to pass the full certificate anymore.

1) Comparing hashes is standard practice. [`/etc/shadow`](https://linuxize.com/post/etc-shadow-file/#:~:text=%2Fetc%2Fshadow%20is%20a%20text,shadow%2C%20and%20has%20640%20permissions%20.) does it, and we do it with [EPP passwords](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/registrar/Registrar.java;drc=87f096ae4067b9be2d0fc4772b52a777aea2b232;l=708) as well. So [the world is not really worry about hash collision](https://crypto.stackexchange.com/questions/47809/why-havent-any-sha-256-collisions-been-found-yet) for SHA256, and we shouldn't either. Therefore we can pass the hash in the header and compare if the full certificate in the database hashes to the same value.

2) Passing only the full certificate before login involves extra logic in both Nomulus and the proxy. The more logic there is, the more potential for bugs. And it is slightly less efficient.

3) As you will see in the comment below, when you pass both the full certificate and the hash, you have a relationship between your arguments that you are neither exploiting in your production code nor honoring in your tests. This introduces redundant information in production, and the tests still passe but they are technically testing the wrong condition.

So I think we should remove the full certificate. Let me know what your thoughts are. Also the actual process would be a bit involved because Nomulus and the proxy are released separately. It would be good if we come up with a plan for a seamless rollout.

> Reviewed 1 of 5 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017#-:-MWCZfjs1AY7ALY4Mu4_:bcdxzw3)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 76 at r2](https://reviewable.io/reviews/google/nomulus/1017#-MWCYCf8BCq23ilsO1sb:-MWCYCf8BCq23ilsO1sc:b-xd6577) ([raw file](https://github.com/google/nomulus/blob/2b8a7ed33ec8fe70df593443e098a2097732fb92/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L76)):*
> ```
> Optional.empty()
> ```
This seems to be wrong. I understand that the certificate hash is no longer used in validation anymore, but in reality it is always set. So here they should be set in test as well.

You can see this is another example why passing both the cert and the hash is a bad idea -- there is a intrinsic dependency between the two and this relationship is not captured in either your code logic or your test. See my main comment about removing the full certificate. 
___
*[core/src/test/java/google/registry/flows/EppLoginTlsTest.java, line 142 at r2](https://reviewable.io/reviews/google/nomulus/1017#-MWCZ5EF5P2BALKHvUUJ:-MWCZ5EG1EXuzPGx6pSq:b-7d4nu8) ([raw file](https://github.com/google/nomulus/blob/2b8a7ed33ec8fe70df593443e098a2097732fb92/core/src/test/java/google/registry/flows/EppLoginTlsTest.java#L142)):*
> ```
> proxyEncoded
> ```
I would rather call it `encodedCertString` like in other tests.
___
*[core/src/test/java/google/registry/flows/session/LoginFlowViaTlsTest.java, line 162 at r2](https://reviewable.io/reviews/google/nomulus/1017#-MWCZWgX-euP-Q-TdP6K:-MWCZWgX-euP-Q-TdP6L:b-1vaff) ([raw file](https://github.com/google/nomulus/blob/2b8a7ed33ec8fe70df593443e098a2097732fb92/core/src/test/java/google/registry/flows/session/LoginFlowViaTlsTest.java#L162)):*
> ```
> missingClientCertificateAndHash
> ```
This should be called `missingclientCertificate`

mindhog:
> To be clear, though, I do remember at one point when we were first mapping this implementation out suggesting that we could just check the certificate in the database, and there being some pushback from the team that it would be better to check the actual certificate sent from the client. I can't remember now though what the exact reasoning was other than potentially deprecating all uses of the certificate hash.

I have only minimal context here, but as I recall the hash had two benefits during the .app launch: 1) it was visible in the proxy logs and therefore allowed us to identify the registrar for a connection in the proxy 2) it allowed us to create a mechanism for denying connections from a specified registrar at the proxy (this was a safeguard against abusive registrars, we've never had to use it).  If we're removing the hash, we should continue to have some way to deal with at least 1), and ideally both use-cases.

jianglai (commented):
@sarahcaseybot , your plan look good overall. The key point is to stop processing the full certificate in nomulus before we stop sending it from the proxy. Though I don't think we need to change `ValidateCertificate` to `ValidateCertificateHash` everywhere. Remember all that we are changing is to not have the proxy send the full certificate during login. So all that needs to be changed is in `TlsCredentials` where we compare the full cert in the header with the full cert in datastore. Instead we compute the hash from datastore and compare it to the hash in the header. Everywhere else, if `validateCertificate` is called right now, they should continue to function, because they should not depend on the full cert from the proxy in the first place.

I would suggest that you modify this PR to implement step 2 in your plan. It should be pretty straight forward based on what you have currently.

@mindhog I think we are covered here, what is being discussed here is to not have the proxy send *both* the cert and its hash, as it is duplicate information. Regardless which value we decide to keep in the end, they are both uniquely identifying information that can be used to tie back to a particular registrar.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017#-:-MWWoQOQ9X5k60-2L5Gx:bf1ztjy)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

mindhog:
> @mindhog I think we are covered here, what is being discussed here is to not have the proxy send _both_ the cert and its hash, as it is duplicate information. Regardless which value we decide to keep in the end, they are both uniquely identifying information that can be used to tie back to a particular registrar.

That's fine, but do keep the ergonomics in mind.  As an example, if we're trying to correlate proxy logs with server logs, it's a lot easier if the same identifier is present in both entries as opposed to having to map cert hash on one side with e.g. a registrar id on the other.  And we certainly wouldn't want to have to read or copy/paste around a 4K RSA key for identification.

CydeWeys (commented):
Wait, why would we stop passing the full certificate on login attempt?  Wasn't the entire reason that we started passing it that we needed it?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017#-:-MWZmHg84xC3M_e86SuS:b-wpctkf)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

jianglai (commented):
@mindhog the registrar object always contains both the certificate and the hash and they are [updated in tandem](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/registrar/Registrar.java;drc=3bf697c43c922110fb9ab55424c14a2fc7b23a2c;l=828). For logs the most relevant ones are the login request/response logs recorded by `TlsCredentials`, and we can control which one (or both) to display. So I don't think there will be an ergonomic problem either way we go.

@sarahcaseybot I'm a little unsure about what you proposed. The `clientCertificate`/`clientCertificateHash` fields are injected in the constructor from the HTTP request headers, so we definitely need one of them as fields because that's where we get that information from the proxy. If the proxy is only passing the hash, then it needs to be injected. My concern was not really that the hash logged in Nomulus does not match the certificate, as that should not happen as long as the callers of `TlsCredentials` pass the correct arguments. But still that's an additional assumption that we should rather not make. It's simpler to stick to one or the other, but not both.

The main benefit of having access to the full certificate on the nomulus side is we can do additional checks like validity period and pending expiration, and that's taken care of by mandating all the registrars to upload a certificate. The proxy is not involved in this process. 

So in summary, if we decide to include some identifying information in all proxy request headers for logging purpose, we should just use that one for authentication as well because either is equally secure. It can be the cert itself, or the hash. But we don't need both of them present. I prefer the hash because it is shorter and the length is fixed. Do note that we get limited value out of logging in proxy in production because most requests are *not* logged due to PII concerns. It's only when exceptions are thrown when the request/response is logged.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017#-:-MWatqyzCXjqG_rWlMNZ:b-v4yvwe)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

CydeWeys (commented):
We can punt on this for now and remove login via full client cert in Nomulus at some point in the future (so rather than using cert hash as the fallback as we are now, only use that, as it used to be).  I don't think we necessarily need to remove sending the full cert at client login (i.e. we don't need to do another proxy change/release) simply because having that information logged could be useful for debugging when things go wrong even if we're not using it for authentication.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017#-:-MWeNVrp4Ihxpf2CusCp:bla98jv)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

jianglai (commented):
I don't think there is an urgency to make the change either. But I do rather have it done sooner than later, if for nothing but tidiness.

What information would the full cert give us that the hash cannot? Can you give an example?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1017#-:-MWft_sgEukhPTY3Ux5y:b5hx21x)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

PR #1024, opened 2021-03-19
Pass --java-binary to _all_ formatter invocations

mindhog:
When implementing a flag to pass in the java binary to
google-java-format-diff.py, I missed the location in showNoncompliantFiles
which gets run before a check.

This change also refactors the core logic of the script so that
google-java-format-diff.py is only called from one place and (in all but one
case) only one time.

Tested:
Ran check format and show, with and without diffs present in the tree.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1024)

### DIFF
--- a/java-format/google-java-format-git-diff.sh
+++ b/java-format/google-java-format-git-diff.sh
@@ -64,15 +64,29 @@ if ! "$JAVA_BIN" -version 2>&1 | grep 'version "11\.' >/dev/null; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1024)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1026, opened 2021-03-19
Convert DomainCreateFlow to use generic tm() methods

gbrodman:
Various necessary changes included as part of this:

- Make ForeignKeyIndex completely generic. Previously, only the load()
method that took a DateTime as input could use SQL, and the cached flow
was particular to Objectify Keys. Now, the cached flow and the
non-cached flow can use the same (ish) piece of code to load / create
the relevant index objects before filtering or modifying them as
necessary.
- EntityChanges should use VKeys
- FlowUtils should persist entity changes using tm(), however not all
object types are storable in SQL.
- Filling out PollMessage fields with the proper object type when
loading from SQL
- Changing a few tm() calls to ofyTm() calls when using objectify. This
is because creating a read-only transaction in SQL is quite a footgun at
the moment, because it makes the entire transaction you're in (if you
were already in one) a read-only transaction.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1026)

### DIFF
--- a/core/src/main/java/google/registry/flows/FlowUtils.java
+++ b/core/src/main/java/google/registry/flows/FlowUtils.java
@@ -15,7 +15,7 @@
 package google.registry.flows;
 
 import static com.google.common.base.Preconditions.checkState;
-import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.xml.ValidationMode.LENIENT;
 import static google.registry.xml.ValidationMode.STRICT;
 import static java.nio.charset.StandardCharsets.UTF_8;
@@ -51,8 +51,8 @@ public final class FlowUtils {
 
   /** Persists the saves and deletes in an {@link EntityChanges} to Datastore. */
   public static void persistEntityChanges(EntityChanges entityChanges) {
-    ofy().save().entities(entityChanges.getSaves());
-    ofy().delete().keys(entityChanges.getDeletes());
+    tm().putAll(entityChanges.getSaves());
+    tm().delete(entityChanges.getDeletes());
   }
 
   /**
--- a/core/src/main/java/google/registry/flows/custom/EntityChanges.java
+++ b/core/src/main/java/google/registry/flows/custom/EntityChanges.java
@@ -16,8 +16,8 @@ package google.registry.flows.custom;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableSet;
-import com.googlecode.objectify.Key;
 import google.registry.model.ImmutableObject;
+import google.registry.persistence.VKey;
 
 /** A wrapper class that encapsulates Datastore entities to both save and delete. */
 @AutoValue
@@ -25,7 +25,7 @@ public abstract class EntityChanges {
 
   public abstract ImmutableSet<ImmutableObject> getSaves();
 
-  public abstract ImmutableSet<Key<ImmutableObject>> getDeletes();
+  public abstract ImmutableSet<VKey<ImmutableObject>> getDeletes();
 
   public static Builder newBuilder() {
     // Default both entities to save and entities to delete to empty sets, so that the build()
@@ -48,11 +48,11 @@ public abstract class EntityChanges {
       return this;
     }
 
-    public abstract Builder setDeletes(ImmutableSet<Key<ImmutableObject>> entitiesToDelete);
+    public abstract Builder setDeletes(ImmutableSet<VKey<ImmutableObject>> entitiesToDelete);
 
-    public abstract ImmutableSet.Builder<Key<ImmutableObject>> deletesBuilder();
+    public abstract ImmutableSet.Builder<VKey<ImmutableObject>> deletesBuilder();
 
-    public Builder addDelete(Key<ImmutableObject> entityToDelete) {
+    public Builder addDelete(VKey<ImmutableObject> entityToDelete) {
       deletesBuilder().add(entityToDelete);
       return this;
     }
--- a/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
+++ b/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
@@ -16,17 +16,20 @@ package google.registry.model.index;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.config.RegistryConfig.getEppResourceCachingDuration;
 import static google.registry.config.RegistryConfig.getEppResourceMaxCachedEntries;
 import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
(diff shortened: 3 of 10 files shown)


CydeWeys (changes requested):
It sounds like we need to solve the footgun.  The main reason we're doing separate read-only transactions in Datastore is for performance/Datastore limitation reasons (IIRC; it should avoid enlisting a given entity group inside the main mutating transaction), whereas I don't think it makes any difference in Cloud SQL land.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026#-:-MWZo_G36aOCp1l7T3qI:b6084ao)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/FlowUtils.java, line 65 at r1](https://reviewable.io/reviews/google/nomulus/1026#-MWZoynS5WUGxNhj_hq_:-MWZoynTAMHWQScolCD0:b-kj5lq6) ([raw file](https://github.com/google/nomulus/blob/ce5735013e8c5e6a3e93a1b5c5a169449562d2db/core/src/main/java/google/registry/flows/FlowUtils.java#L65)):*
> ```Java
>                   .filter(
>                       obj ->
>                           obj.getClass().isAnnotationPresent(Entity.class)
> ```

This is a hacky way of solving this at this one particular callsite.  We need a much better, more robust, centralized way for not writing entities that don't exist in SQL to SQL.

gbrodman (commented):
yeah, we may need to implement some sort of `transactNew` functionality for SQL then in order to have the read-only transactions :/ 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026#-:-MW_RMIPCLCFuktYEgr8:b-rodfd5)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/FlowUtils.java, line 65 at r1](https://reviewable.io/reviews/google/nomulus/1026#-MWZoynS5WUGxNhj_hq_:-MW_L5lv94gggc35oJnv:b-ll3pct) ([raw file](https://github.com/google/nomulus/blob/ce5735013e8c5e6a3e93a1b5c5a169449562d2db/core/src/main/java/google/registry/flows/FlowUtils.java#L65)):*

This is a hacky way of solving this at this one particular callsite.  We need a much better, more robust, centralized way for not writing entities that don't exist in SQL to SQL.
</blockquote></details>

hm, yeah it seems like there already is a mechanism for that so this isn't necessary. Wonder why I was hitting issues with this before then, but we should be all good here.

gbrodman (commented):
Added a comment and removed the footgun -- we just will have `transactNew` and `transactNewReadOnly` use standard transactions, as they basically were before. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026#-:-MWa_yZMAz9kDmdp5TLq:b-jxanov)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 198 at r2](https://reviewable.io/reviews/google/nomulus/1026#-MWydXWD-VMBeZQdn4Dz:-MWydXWD-VMBeZQdn4E-:b98erl8) ([raw file](https://github.com/google/nomulus/blob/8b782f300d7feac565d9a79afbcf35194d155678/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L198)):*
> ```Java
>     return loadIndexesFromStore(clazz, foreignKeys).entrySet().stream()
>         .filter(e -> now.isBefore(e.getValue().getDeletionTime()))
>         .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
> ```

Use `entriesToImmutableMap()` (statically imported from `CollectionUtils`)
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 239 at r2](https://reviewable.io/reviews/google/nomulus/1026#-MWydBh_AE3pZp3R0-y8:-MWydBh_AE3pZp3R0-y9:b98erl8) ([raw file](https://github.com/google/nomulus/blob/8b782f300d7feac565d9a79afbcf35194d155678/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L239)):*
> ```Java
>                           .max(Comparator.comparing(ForeignKeyIndex::getDeletionTime))
>                           .get()))
>           .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
> ```

Use `entriesToImmutableMap()` (statically imported from `CollectionUtils`)

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026#-:-MWyddld4efIoBbtFOnb:bnfp4nl)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 198 at r2](https://reviewable.io/reviews/google/nomulus/1026#-MWydXWD-VMBeZQdn4Dz:-MWyeFRX16hvX556tUOb:be9f77) ([raw file](https://github.com/google/nomulus/blob/8b782f300d7feac565d9a79afbcf35194d155678/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L198)):*

Use `entriesToImmutableMap()` (statically imported from `CollectionUtils`)
</blockquote></details>

oh nice, didn't know that that existed (it wasn't here before) 
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 239 at r2](https://reviewable.io/reviews/google/nomulus/1026#-MWydBh_AE3pZp3R0-y8:-MWyeHiQ-Xm1vuudNRV9:be9f77) ([raw file](https://github.com/google/nomulus/blob/8b782f300d7feac565d9a79afbcf35194d155678/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L239)):*

Use `entriesToImmutableMap()` (statically imported from `CollectionUtils`)
</blockquote></details>

oh nice, didn't know that that existed (it wasn't here before)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1026)* status: 0 of 10 files reviewed, 3 unresolved discussions

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 213 at r3](https://reviewable.io/reviews/google/nomulus/1026#-MYB1DS_8kB7IUM1c_VX:-MYB1DSa3J5RAoq_q31r:bqc7bi1) ([raw file](https://github.com/google/nomulus/blob/132ed555e27e2ac5f834494079adeb91ed67bd57/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L213)):*
> ```Java
>     if (tm().isOfy()) {
>       return ImmutableMap.copyOf(
>           tm().doTransactionless(() -> ofy().load().type(mapToFkiClass(clazz)).ids(foreignKeys)));
> ```

In hindsight, this should not have been transaction-less.  This was a bad bug.  What it's replacing is necessarily transactional, otherwise you get the potential to allow EPP resources with duplicate unique IDs (contacts with the same contact ID, domains and hosts with the same name!!).

PR #1028, opened 2021-03-20
Add release info to Nomulus config files

weiminyu:
Add the actual release tag and beam staging project id to the config
file. This allows the Nomulus server to find the right version of the
BEAM pipelines to launch.


weiminyu (commented):
Verified in alpha.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1028#-:-MWCFOF_29gStycK1Da9:bvftwfs)* status: 0 of 5 files reviewed, all discussions resolved

jianglai (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1028)* status: all files reviewed, 3 unresolved discussions (waiting on @mindhog and @weiminyu)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 661 at r1](https://reviewable.io/reviews/google/nomulus/1028#-MWCQfe77oKJP1iftycy:-MWCQfe8AfCetpNrH3rS:b-2aha1d) ([raw file](https://github.com/google/nomulus/blob/0e716a36a9c5411b0da2ffd2b8e0ab3fe75dd0e6/core/src/main/java/google/registry/config/RegistryConfig.java#L661)):*
> ```
> config.release.beamStagingProjectId
> ```
Is the current practice to store the pipelines in a staging project or the runtime project?
___
*[core/src/test/java/google/registry/config/RegistryConfigTest.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/1028#-MWCRush1Pi_s5k9G_Uc:-MWCRusiA_cQGBR1g9Jd:b-1b6par) ([raw file](https://github.com/google/nomulus/blob/0e716a36a9c5411b0da2ffd2b8e0ab3fe75dd0e6/core/src/test/java/google/registry/config/RegistryConfigTest.java#L45)):*
> ```
> .isEqualTo("gs://registry-project-id-deploy/live/beam"
> ```
This makes the test depend on the value of the default config (i. e. the "live" tag and the null staging project ID). Is this a good pattern? I think we should rather separate code logic from config values.

What's more, is it really necessary to test this? There are a lot more provider methods in the config module but we didn't seem to test any of them (other than the reserved terms export disclaimer).
___
*[release/cloudbuild-release.yaml, line 170 at r1](https://reviewable.io/reviews/google/nomulus/1028#-MWCTcyg2WMsePlCsEtU:-MWCTcyg2WMsePlCsEtV:b1214qe) ([raw file](https://github.com/google/nomulus/blob/0e716a36a9c5411b0da2ffd2b8e0ab3fe75dd0e6/release/cloudbuild-release.yaml#L170)):*
> ```
> config_file="./core/src/main/java/google/registry/config/files/default-config.yaml"
> ```
Personally I find modifying the config values itself based on release tags a bit strange. In other places in this script we are changing values used to deploy the binary (e. g. the various cloudbuild-*.yaml scripts), but not the binary itself. This also makes it impossible to individually release/rollback pipelines without a corresponding change in nomulus. We currently have this ability. Not sure if we want to maintain this flexibility, or if we want to strictly tie pipelines with nomulus release.

I think a better approach is to always read from gs://...-deploy/live/beam/ in nomulus, and in the release process populate both gs://...-deploy/<TAG>/beam/ and gs://...-deploy/live/beam/. This is actually automatically done in https://github.com/google/nomulus/blob/master/release/cloudbuild-sync.yaml.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1028)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jianglai and @mindhog)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 661 at r1](https://reviewable.io/reviews/google/nomulus/1028#-MWCQfe77oKJP1iftycy:-MWPP_5f6vUlYB8Kc-iZ:b-zenoe4) ([raw file](https://github.com/google/nomulus/blob/0e716a36a9c5411b0da2ffd2b8e0ab3fe75dd0e6/core/src/main/java/google/registry/config/RegistryConfig.java#L661)):*

> ```
> config.release.beamStagingProjectId
> ```
Is the current practice to store the pipelines in a staging project or the runtime project?
</blockquote></details>

'Staging' is a dataflow term that refers to the action of uploading execution-ready artifacts to GCS.

In our case, when artifacts can be shared across projects, e.g., flex-templates between sandbox and production, it makes sense to stage them in  domain-registry-dev. If not, e.g., flex-templates in alpha and maybe crash and all classical templates deployed by the tool, it makes sense to stage them in the project that runs the pipelines.
___
*[core/src/test/java/google/registry/config/RegistryConfigTest.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/1028#-MWCRush1Pi_s5k9G_Uc:-MWPQrUmDuhPmfVAXX58:b38ef3l) ([raw file](https://github.com/google/nomulus/blob/0e716a36a9c5411b0da2ffd2b8e0ab3fe75dd0e6/core/src/test/java/google/registry/config/RegistryConfigTest.java#L45)):*

> ```
> .isEqualTo("gs://registry-project-id-deploy/live/beam"
> ```
This makes the test depend on the value of the default config (i. e. the "live" tag and the null staging project ID). Is this a good pattern? I think we should rather separate code logic from config values.

What's more, is it really necessary to test this? There are a lot more provider methods in the config module but we didn't seem to test any of them (other than the reserved terms export disclaimer).
</blockquote></details>

I think this parameter needs testing since its provider does not merely echo the content of the config file.
The main goal is to verify that the projectId is injected and the project-independent pattern is correct.

This is moot after the change in configurations. 
___
*[release/cloudbuild-release.yaml, line 170 at r1](https://reviewable.io/reviews/google/nomulus/1028#-MWCTcyg2WMsePlCsEtU:-MWPcyp60NXtrdehwNm6:bykkc2s) ([raw file](https://github.com/google/nomulus/blob/0e716a36a9c5411b0da2ffd2b8e0ab3fe75dd0e6/release/cloudbuild-release.yaml#L170)):*

> ```
> config_file="./core/src/main/java/google/registry/config/files/default-config.yaml"
> ```
Personally I find modifying the config values itself based on release tags a bit strange. In other places in this script we are changing values used to deploy the binary (e. g. the various cloudbuild-*.yaml scripts), but not the binary itself. This also makes it impossible to individually release/rollback pipelines without a corresponding change in nomulus. We currently have this ability. Not sure if we want to maintain this flexibility, or if we want to strictly tie pipelines with nomulus release.

I think a better approach is to always read from gs://...-deploy/live/beam/ in nomulus, and in the release process populate both gs://...-deploy/<TAG>/beam/ and gs://...-deploy/live/beam/. This is actually automatically done in https://github.com/google/nomulus/blob/master/release/cloudbuild-sync.yaml.
</blockquote></details>

You are right. I followed the original proposal in go/dr-beam-deploy without thinking much about what has changed. Using the live folder makes more sense now.

mindhog (approved):
> Reviewed 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1028)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @jianglai)

PR #1030, opened 2021-03-22
Disallow admin triggering of internal endpoints

mindhog:
Stop simply relying on the presence of the X-AppEngine-QueueName as an
indicator that an endpoint has been triggered internally, as this allows
admins to trigger a remote execution vulnerability.

We now supplement this check by ensuring that there is no authenticated user.
Since only an admin user can set these headers, this means that the header
must have been set by an internal request.

Tested:
  In addition to the new unit test, verified on Crash that:
  - Internal requests are still getting authenticated via the internal auth
    mechanism.
  - Admin requests with the X-AppEngine-QueueName header are rejected as
    "unauthorized."

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1030)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1030)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1033, opened 2021-03-23
Embed a ZonedDateTime as the UpdateAutoTimestamp in SQL

gbrodman:
This means we can get rid of the converter and more importantly, means
that reading the object from SQL does not affect the last-read time (the
test added to UpdateAutoTimestampTest failed prior to the production
code change).

For now we keep both time fields in UpdateAutoTimestamp however
post-migration, we can remove the joda-time field if we wish.

Note: I'm not sure why now is the time that we started getting
LazyInitializationExceptions in the LegacyHistoryObject and
ReplayExtension tests but we can solve that by just examining /
initializing the object within the transaction.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1033)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033)* status: 0 of 18 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/BackupGroupRoot.java, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZhYFq1Zd8TFIzPQdK:-MWZhYFq1Zd8TFIzPQdL:b-cumru6) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/core/src/main/java/google/registry/model/BackupGroupRoot.java#L46)):*
> ```Java
>   @Access(AccessType.FIELD)
>   @AttributeOverride(name = "lastUpdateTime", column = @Column(name = "updateTimestamp"))
>   @DoNotCompare
> ```

What's the logic with all the `@DoNotCompare`s in this PR?
___
*[core/src/main/java/google/registry/model/UpdateAutoTimestamp.java, line 57 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZhbMJ1ufN6zOEfuZi:-MWZhbMJ1ufN6zOEfuZj:b-wnggrw) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/core/src/main/java/google/registry/model/UpdateAutoTimestamp.java#L57)):*
> ```Java
>   // because Hibernate does not allow it to be used on @Embeddable classes, see
>   // https://hibernate.atlassian.net/browse/HHH-13235. This is a workaround.
>   @PrePersist
> ```

Is the issue that Mike raised about this potentially not being called recursively on embedded entities addressed?
___
*[core/src/main/java/google/registry/model/UpdateAutoTimestamp.java, line 60 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZggBp-npVWkUGGzRr:-MWZg_pR9E4aJHwLulhm:b-yhaspy) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/core/src/main/java/google/registry/model/UpdateAutoTimestamp.java#L60)):*
> ```Java
>   @PreUpdate
>   void setTimestamp() {
>     if (autoUpdateEnabled() || lastUpdateTime == null) {
> ```

When is auto update not enabled exactly?
___
*[db/src/main/resources/sql/flyway/V90__update_timestamp.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZh4lm-BOoa3mIx-jL:-MWZh4lm-BOoa3mIx-jM:b-dbs3ei) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/db/src/main/resources/sql/flyway/V90__update_timestamp.sql#L15)):*
> ```TSQL
> 
> ALTER TABLE "Registrar" ALTER COLUMN "last_update_time" SET NOT NULL;
> ```

And we have zero null rows in the DB for these tables that this will cause to break?
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 834 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZhKXI6vbo91Zb2_dK:-MWZhKXI6vbo91Zb2_dL:b9svy89) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/db/src/main/resources/sql/schema/nomulus.golden.sql#L834)):*
> ```TSQL
> CREATE TABLE public."RegistryLock" (
>     revision_id bigint NOT NULL,
>     lock_completion_timestamp timestamp with time zone,
> ```

There's now four fields that end with `_timestamp` and one that ends with `_time` ... would it make sense to rename the other 4 for consistency while you're at it?  I agree that `_timestamp` is longer than `_time` without providing particularly any extra value.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033)* status: 0 of 22 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/BackupGroupRoot.java, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZhYFq1Zd8TFIzPQdK:-MWZwM4g7Ik-mZXymTeK:b-e68ud1) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/core/src/main/java/google/registry/model/BackupGroupRoot.java#L46)):*

What's the logic with all the `@DoNotCompare`s in this PR?
</blockquote></details>

Yeah, so actually originally I thought they were necessary (to tell the replay tests not to worry about this field) but after revisiting the replay tests we should not need them, and the only reason why they were being added is that JPA actually nulls out non-persisted fields before the PrePersist methods are called -- fixing the `create` method in UAT means we shouldn't need this, since the auto update is disabled during replay. 
___
*[core/src/main/java/google/registry/model/UpdateAutoTimestamp.java, line 57 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZhbMJ1ufN6zOEfuZi:-MWZwnHK7JpnLWY1qZcY:bmmk0x) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/core/src/main/java/google/registry/model/UpdateAutoTimestamp.java#L57)):*

Is the issue that Mike raised about this potentially not being called recursively on embedded entities addressed?
</blockquote></details>

It is called, I added a little test case in EntityCallbacksListener to verify
___
*[core/src/main/java/google/registry/model/UpdateAutoTimestamp.java, line 60 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZggBp-npVWkUGGzRr:-MWZwyM53aa94IwfmO2W:b2vdcdm) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/core/src/main/java/google/registry/model/UpdateAutoTimestamp.java#L60)):*

When is auto update not enabled exactly?
</blockquote></details>

It has to do with the replay tests / replay extension. You can see the similar deleted logic in the now-deleted UpdateAutoTimestampConverter.  Basically, when replaying items to SQL post-test we don't want to change the timestamp because otherwise the comparison would fail (since time likely advanced since the test was run). This way we avoid changing it during the course of the replay.
___
*[db/src/main/resources/sql/flyway/V90__update_timestamp.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZh4lm-BOoa3mIx-jL:-MWZx4_fDssdoImCr8GT:b-7jkk3s) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/db/src/main/resources/sql/flyway/V90__update_timestamp.sql#L15)):*

And we have zero null rows in the DB for these tables that this will cause to break?
</blockquote></details>

Yeah, last_update_time should have been not-null to begin with, it's never null in the table. 
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 834 at r1](https://reviewable.io/reviews/google/nomulus/1033#-MWZhKXI6vbo91Zb2_dK:-MWZxIO185Y3OMB2mSuB:b-gq2jl3) ([raw file](https://github.com/google/nomulus/blob/616e245851276dba452241b5015bdcfb650fb409/db/src/main/resources/sql/schema/nomulus.golden.sql#L834)):*

There's now four fields that end with `_timestamp` and one that ends with `_time` ... would it make sense to rename the other 4 for consistency while you're at it?  I agree that `_timestamp` is longer than `_time` without providing particularly any extra value.
</blockquote></details>

sure, makes sense -- done

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[db/src/main/resources/sql/flyway/V90__update_timestamp.sql, line 16 at r2](https://reviewable.io/reviews/google/nomulus/1033#-MWewsvHAKneeTKw29PM:-MWewsvHAKneeTKw29PN:btfd47x) ([raw file](https://github.com/google/nomulus/blob/bc188c0640449259fec349110fd79c32b6ea1769/db/src/main/resources/sql/flyway/V90__update_timestamp.sql#L16)):*
> ```TSQL
> 
> ALTER TABLE "Registrar" ALTER COLUMN "last_update_time" SET NOT NULL;
> ALTER TABLE "RegistryLock" RENAME "last_update_timestamp" TO "last_update_time";
> ```

It's worth syncing with Weimin first just to ensure that only having these field renames in the Flyway file are the correct way to do it.

Because it seems to me that if any of the old instances are still running while this schema update runs, then those instances may start failing, and if any new instances start running prior to this schema update being run, then they'll fail as well.  Does the way this is structured not require a 100% downtime of all current instances, run the schema update, then bring up new instances?  I don't see any flexibility anywhere that will handle either/or.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[db/src/main/resources/sql/flyway/V90__update_timestamp.sql, line 16 at r2](https://reviewable.io/reviews/google/nomulus/1033#-MWewsvHAKneeTKw29PM:-MWf3C2N2xS-kGSYRR3l:b-nqyo2i) ([raw file](https://github.com/google/nomulus/blob/bc188c0640449259fec349110fd79c32b6ea1769/db/src/main/resources/sql/flyway/V90__update_timestamp.sql#L16)):*

It's worth syncing with Weimin first just to ensure that only having these field renames in the Flyway file are the correct way to do it.

Because it seems to me that if any of the old instances are still running while this schema update runs, then those instances may start failing, and if any new instances start running prior to this schema update being run, then they'll fail as well.  Does the way this is structured not require a 100% downtime of all current instances, run the schema update, then bring up new instances?  I don't see any flexibility anywhere that will handle either/or.
</blockquote></details>

It all depends on how big the time window with mismatched binary and schema is, and how many requests come in during the window. We've done this in the past without problems because the tables were rarely used. This time we have to make a call. The window is typically 15 - 20 minutes or so. 

Sooner or later we will enable the compatibility check which will forbid these kind of changes. We may want to start practicing a safe change  and write it up as a playbook entry. This is a good case since it presumably requires backfill of the new columns.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033)* status: 0 of 28 files reviewed, all discussions resolved (waiting on @mindhog)

___
*[db/src/main/resources/sql/flyway/V90__update_timestamp.sql, line 16 at r2](https://reviewable.io/reviews/google/nomulus/1033#-MWewsvHAKneeTKw29PM:-MWyeWxL4kN7kR3ixYLF:buwox36) ([raw file](https://github.com/google/nomulus/blob/bc188c0640449259fec349110fd79c32b6ea1769/db/src/main/resources/sql/flyway/V90__update_timestamp.sql#L16)):*

It all depends on how big the time window with mismatched binary and schema is, and how many requests come in during the window. We've done this in the past without problems because the tables were rarely used. This time we have to make a call. The window is typically 15 - 20 minutes or so. 

Sooner or later we will enable the compatibility check which will forbid these kind of changes. We may want to start practicing a safe change  and write it up as a playbook entry. This is a good case since it presumably requires backfill of the new columns.

</blockquote></details>

I'm OK with doing the unsafe way for now out of expediency since registry lock isn't used that frequently (yet?), but yes, we will want a use case for a safer migration than this at some point.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033#-:-MWyelgl6hIyo5vjxTBr:bnfp4nl)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @mindhog and @weiminyu)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1033)* status: 0 of 28 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V90__update_timestamp.sql, line 16 at r2](https://reviewable.io/reviews/google/nomulus/1033#-MWewsvHAKneeTKw29PM:-MWyfprsE5LsWOAvpFrB:b7adliu) ([raw file](https://github.com/google/nomulus/blob/bc188c0640449259fec349110fd79c32b6ea1769/db/src/main/resources/sql/flyway/V90__update_timestamp.sql#L16)):*

I'm OK with doing the unsafe way for now out of expediency since registry lock isn't used that frequently (yet?), but yes, we will want a use case for a safer migration than this at some point.
</blockquote></details>

Sounds good to me; I think we probably want to test that on a change that is separated from any other functionality changes, maybe a change that is *just* a rename of a column or something (I'm sure we have columns that we'd like to rename)

PR #1050, opened 2021-04-01
Convert poll-message-related classes to use SQL as well

gbrodman:
Two relatively complex parts. The first is that we needed a small
refactor on the AckPollMessagesCommand because we could theoretically be
acking more poll messages than the Datastore transaction size boundary.
This means that the normal flow of "gather the poll messages from the DB
into one collection, then act on it" needs to be changed to a more
functional flow.

The second is that acking the poll message (deleting it in most cases)
reduces the number of remaining poll messages in SQL but not in
Datastore, since in Datastore the deletion does not take effect until
after the transaction is over.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1050)

### DIFF
--- a/core/src/main/java/google/registry/flows/poll/PollAckFlow.java
+++ b/core/src/main/java/google/registry/flows/poll/PollAckFlow.java
@@ -16,15 +16,13 @@ package google.registry.flows.poll;
 
 import static google.registry.flows.FlowUtils.validateClientIsLoggedIn;
 import static google.registry.flows.poll.PollFlowUtils.ackPollMessage;
-import static google.registry.flows.poll.PollFlowUtils.getPollMessagesQuery;
+import static google.registry.flows.poll.PollFlowUtils.getPollMessageCount;
 import static google.registry.model.eppoutput.Result.Code.SUCCESS_WITH_NO_MESSAGES;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.model.poll.PollMessageExternalKeyConverter.makePollMessageExternalId;
 import static google.registry.model.poll.PollMessageExternalKeyConverter.parsePollMessageExternalId;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.DateTimeUtils.isBeforeOrAt;
 
-import com.googlecode.objectify.Key;
 import google.registry.flows.EppException;
 import google.registry.flows.EppException.AuthorizationErrorException;
 import google.registry.flows.EppException.ObjectDoesNotExistException;
@@ -39,6 +37,8 @@ import google.registry.model.poll.MessageQueueInfo;
 import google.registry.model.poll.PollMessage;
 import google.registry.model.poll.PollMessageExternalKeyConverter;
 import google.registry.model.poll.PollMessageExternalKeyConverter.PollMessageExternalKeyParseException;
+import google.registry.persistence.VKey;
+import java.util.Optional;
 import javax.inject.Inject;
 import org.joda.time.DateTime;
 
@@ -71,7 +71,7 @@ public class PollAckFlow implements TransactionalFlow {
       throw new MissingMessageIdException();
     }
 
-    Key<PollMessage> pollMessageKey;
+    VKey<PollMessage> pollMessageKey;
     // Try parsing the messageId, and throw an exception if it's invalid.
     try {
       pollMessageKey = parsePollMessageExternalId(messageId);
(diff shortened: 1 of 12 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java, line 171 at r1](https://reviewable.io/reviews/google/nomulus/1050#-MXIgM-v6oTveU1w960p:-MXIgM-w68UIui_mVMuy:b-qgk13i) ([raw file](https://github.com/google/nomulus/blob/6a351624dd15df19248229fb6de51fecb92e1fb8/core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java#L171)):*
> ```Java
>         .isTrue();
>     assertThat(one.getSqlKey()).isEqualTo(two.getSqlKey());
>     assertThat(one.getOfyKey()).isEqualTo(two.getOfyKey());
> ```

Does .equals() not work here?  In general it should, as VKey derives from ImmutableObject

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java, line 171 at r1](https://reviewable.io/reviews/google/nomulus/1050#-MXIgM-v6oTveU1w960p:-MXIlacwCThQkL56UMRs:bo8sojx) ([raw file](https://github.com/google/nomulus/blob/6a351624dd15df19248229fb6de51fecb92e1fb8/core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java#L171)):*

Does .equals() not work here?  In general it should, as VKey derives from ImmutableObject
</blockquote></details>

I added a comment explaining this -- the kind / class of the VKey may not be precisely the same, since when reconstituting the key from the poll message string we only know that it is the generic PollMessage superclass, which is not technically (as far as Java knows) equal to the PollMessage.OneTime (or recurring, whatever) class. 

The DB treats them as the same though so it's fine.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java, line 171 at r1](https://reviewable.io/reviews/google/nomulus/1050#-MXIgM-v6oTveU1w960p:-MXIr3gc4Z5ih3EO6dIc:b-7axlqk) ([raw file](https://github.com/google/nomulus/blob/6a351624dd15df19248229fb6de51fecb92e1fb8/core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java#L171)):*

I added a comment explaining this -- the kind / class of the VKey may not be precisely the same, since when reconstituting the key from the poll message string we only know that it is the generic PollMessage superclass, which is not technically (as far as Java knows) equal to the PollMessage.OneTime (or recurring, whatever) class. 

The DB treats them as the same though so it's fine. 
</blockquote></details>

That's actually a problem, as in general we rely on the ability to compare VKeys.  Also, I thought we'd fixed that for PollMessage.

We can go with this for now, I'll make a note to investigate.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java, line 171 at r1](https://reviewable.io/reviews/google/nomulus/1050#-MXIgM-v6oTveU1w960p:-MXIv7Av2aqJWVqNl9pM:bt1spf7) ([raw file](https://github.com/google/nomulus/blob/6a351624dd15df19248229fb6de51fecb92e1fb8/core/src/test/java/google/registry/model/poll/PollMessageExternalKeyConverterTest.java#L171)):*

That's actually a problem, as in general we rely on the ability to compare VKeys.  Also, I thought we'd fixed that for PollMessage.

We can go with this for now, I'll make a note to investigate.  
</blockquote></details>

More generally, we can for-sure count VKey kinds as equal if one is a subclass of another and if the superclass has the annotation `@Inheritance(strategy = InheritanceType.SINGLE_TABLE)` on it. 

I'm not sure about other situations, e.g. DomainHistory VKey vs HistoryEntry VKey.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/tools/AckPollMessagesCommand.java, line 129 at r2](https://reviewable.io/reviews/google/nomulus/1050#-MXIw7ttE97GlQ1qZH2j:-MXIw7ttE97GlQ1qZH2k:b-8ka3k6) ([raw file](https://github.com/google/nomulus/blob/ec0220f33c9889d3e989dbe9e8dcb74f1709b00e/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java#L129)):*
> ```Java
>                     jpaTm()
>                         .query(SQL_POLL_MESSAGE_QUERY_BY_MESSAGE, PollMessage.class)
>                         .setParameter("msg", String.format("%%%s%%", message));
> ```

`"%" + message + "%"` is honestly less confusing than this.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050#-:-MXIwhDoBH2qeGkSMdCy:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1050)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/tools/AckPollMessagesCommand.java, line 129 at r2](https://reviewable.io/reviews/google/nomulus/1050#-MXIw7ttE97GlQ1qZH2j:-MXJ4Ht0435-stR7jW_k:b9pql5t) ([raw file](https://github.com/google/nomulus/blob/ec0220f33c9889d3e989dbe9e8dcb74f1709b00e/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java#L129)):*

`"%" + message + "%"` is honestly less confusing than this.
</blockquote></details>

Fair point, done.

decodedfile (commented):
Duplicate of #

decodedfile (commented):
Duplicate of #

PR #1053, opened 2021-04-02
Convert RefreshDnsOnHostRenameAction to tm

mindhog:
This is not quite complete because it also requires the conversion of a
map-reduce which is in scope for an entirely different work.  Tests of the
map-reduce functionality are excluded from the SQL run.

This also requires the following additional fixes:

-  Convert Lock to tm, as doing so was necessary to get this action to work.
   As Lock is being targeted as DatastoreOnly, we convert all calls in it to
   use ofyTm()
-  Fix a bug in DualDatabaseTest (the check for an AppEngineExtension field is
   wrong, and captures fields of type Object as AppEngineExtension's)
-  Introduce another VKey.from() method that creates a VKey from a stringified
   Ofy Key.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1053)

### DIFF
--- a/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
+++ b/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
@@ -25,7 +25,7 @@ import static google.registry.batch.AsyncTaskMetrics.OperationType.DNS_REFRESH;
 import static google.registry.mapreduce.inputs.EppResourceInputs.createEntityInput;
 import static google.registry.model.EppResourceUtils.isActive;
 import static google.registry.model.EppResourceUtils.isDeleted;
-import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.DateTimeUtils.latestOf;
 import static java.util.concurrent.TimeUnit.DAYS;
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -44,7 +44,6 @@ import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.flogger.FluentLogger;
-import com.googlecode.objectify.Key;
 import google.registry.batch.AsyncTaskMetrics.OperationResult;
 import google.registry.dns.DnsQueue;
 import google.registry.mapreduce.MapreduceRunner;
@@ -123,7 +122,7 @@ public class RefreshDnsOnHostRenameAction implements Runnable {
     }
 
     ImmutableList.Builder<DnsRefreshRequest> requestsBuilder = new ImmutableList.Builder<>();
-    ImmutableList.Builder<Key<HostResource>> hostKeys = new ImmutableList.Builder<>();
+    ImmutableList.Builder<VKey<HostResource>> hostKeys = new ImmutableList.Builder<>();
     final List<DnsRefreshRequest> requestsToDelete = new ArrayList<>();
 
     for (TaskHandle task : tasks) {
@@ -204,10 +203,10 @@ public class RefreshDnsOnHostRenameAction implements Runnable {
         emit(true, true);
         return;
       }
-      Key<HostResource> referencingHostKey = null;
+      VKey<HostResource> referencingHostKey = null;
       for (DnsRefreshRequest request : refreshRequests) {
         if (isActive(domain, request.lastUpdateTime())
-            && domain.getNameservers().contains(VKey.from(request.hostKey()))) {
+            && domain.getNameservers().contains(request.hostKey())) {
           referencingHostKey = request.hostKey();
           break;
         }
@@ -293,7 +292,8 @@ public class RefreshDnsOnHostRenameAction implements Runnable {
 
     private static final long serialVersionUID = 1772812852271288622L;
 
-    abstract Key<HostResource> hostKey();
+    abstract VKey<HostResource> hostKey();
+
     abstract DateTime lastUpdateTime();
     abstract DateTime requestedTime();
     abstract boolean isRefreshNeeded();
@@ -301,7 +301,8 @@ public class RefreshDnsOnHostRenameAction implements Runnable {
 
     @AutoValue.Builder
     abstract static class Builder {
-      abstract Builder setHostKey(Key<HostResource> hostKey);
+      abstract Builder setHostKey(VKey<HostResource> hostKey);
+
       abstract Builder setLastUpdateTime(DateTime lastUpdateTime);
       abstract Builder setRequestedTime(DateTime requestedTime);
       abstract Builder setIsRefreshNeeded(boolean isRefreshNeeded);
@@ -314,10 +315,11 @@ public class RefreshDnsOnHostRenameAction implements Runnable {
      */
     static DnsRefreshRequest createFromTask(TaskHandle task, DateTime now) throws Exception {
       ImmutableMap<String, String> params = ImmutableMap.copyOf(task.extractParams());
-      Key<HostResource> hostKey =
-          Key.create(checkNotNull(params.get(PARAM_HOST_KEY), "Host to refresh not specified"));
+      VKey<HostResource> hostKey =
+          VKey.from(checkNotNull(params.get(PARAM_HOST_KEY), "Host to refresh not specified"));
       HostResource host =
-          checkNotNull(ofy().load().key(hostKey).now(), "Host to refresh doesn't exist");
+          tm().transact(() -> tm().loadByKeyIfPresent(hostKey))
+              .orElseThrow(() -> new NullPointerException("Host to refresh doesn't exist"));
       boolean isHostDeleted =
           isDeleted(host, latestOf(now, host.getUpdateTimestamp().getTimestamp()));
       if (isHostDeleted) {
--- a/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
+++ b/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java
@@ -63,6 +63,7 @@ import google.registry.util.SystemClock;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.logging.Level;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1053)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java, line 322 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJyCU51hSCENGgbGsZ:-MXJyCU6-vy7w4bRqQKY:bvrx38) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java#L322)):*
> ```Java
>       HostResource host =
>           tm().transact(() -> tm().loadByKeyIfPresent(hostKey))
>               .orElseThrow(() -> new NullPointerException("Host to refresh doesn't exist"));
> ```

might an IllegalArgumentException or NoSuchElementException be better? Not sure what the logic is for having an NPE here 
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 193 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJwimaDZr-tVeILmSB:-MXJwimaDZr-tVeILmSC:blzvwt6) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/model/server/Lock.java#L193)):*
> ```Java
>     // access to resources like GCS that can't be transactionally rolled back. Therefore, the lock
>     // must be definitively acquired before it is used, even when called inside another transaction.
>     AcquireResult acquireResult =
> ```

I think we may want to rethink how we deal with locks. Is there any reason for keeping the SQL locks and Datastore locks as separate Java objects? We may have done that at the start when we didn't really know as much about what we were doing, but like with Cursors it may make sense to combine them into one object if we don't have a seamless LockDualDatabaseDao (which we don't). 

If we have them in the same class, then we can just use tm() methods to acquire them (like what we did with the Cursor refactor in https://github.com/google/nomulus/pull/988
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 233 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJxmHxEgkSfPwYa0r2:-MXJxmHxEgkSfPwYa0r3:b-pk0xj0) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/persistence/VKey.java#L233)):*
> ```Java
>    */
>   @Nullable
>   public static <T> VKey<T> from(String ofyKeyRepr) {
> ```

maybe rename to `fromWebsafeKey` like the `Key` method? this might be a little vague
___
*[core/src/test/java/google/registry/testing/DualDatabaseTestInvocationContextProvider.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJwUEP3HcedVQMGW_8:-MXJwUEP3HcedVQMGW_9:b-dgh8nt) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/test/java/google/registry/testing/DualDatabaseTestInvocationContextProvider.java#L126)):*
> ```Java
>       fieldBuilder.addAll(
>           Stream.of(clazz.getDeclaredFields())
>               .filter(field -> AppEngineExtension.class.isAssignableFrom(field.getType()))
> ```

Wow, I'm surprised that this didn't fail much before

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1053)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java, line 322 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJyCU51hSCENGgbGsZ:-MXWgMH67SaRmfXsQpVW:bm8yr59) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java#L322)):*

might an IllegalArgumentException or NoSuchElementException be better? Not sure what the logic is for having an NPE here 
</blockquote></details>

The original call was checkNotNull(), which (somewhat suprisingly to me) throws a NullPointerException, so my intent was to preserve the original interface.
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 193 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJwimaDZr-tVeILmSB:-MXWgkVG6TxiR9sB0niO:b-qumojj) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/model/server/Lock.java#L193)):*

I think we may want to rethink how we deal with locks. Is there any reason for keeping the SQL locks and Datastore locks as separate Java objects? We may have done that at the start when we didn't really know as much about what we were doing, but like with Cursors it may make sense to combine them into one object if we don't have a seamless LockDualDatabaseDao (which we don't). 

If we have them in the same class, then we can just use tm() methods to acquire them (like what we did with the Cursor refactor in https://github.com/google/nomulus/pull/988
</blockquote></details>

Yeah, I agree.   I didn't really want to get into that in the context of this PR, my intent here was to make the datastore locks (i.e. the ones that we're actually using) work correctly, since they were broken in SQL mode.

I've added this as a discussion topic for today.
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 233 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJxmHxEgkSfPwYa0r2:-MXWkNLfBVNrneM6LQkX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/persistence/VKey.java#L233)):*

maybe rename to `fromWebsafeKey` like the `Key` method? this might be a little vague
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/DualDatabaseTestInvocationContextProvider.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJwUEP3HcedVQMGW_8:-MXWkecoC3DZOMRQ7WBt:b-krscs5) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/test/java/google/registry/testing/DualDatabaseTestInvocationContextProvider.java#L126)):*

Wow, I'm surprised that this didn't fail much before
</blockquote></details>

I guess there aren't many fields of type "Object" in our test classes.

gbrodman (changes requested):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1053)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java, line 322 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJyCU51hSCENGgbGsZ:-MXXlVHG3eV1TDN1qH1i:bqvy1vr) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java#L322)):*

The original call was checkNotNull(), which (somewhat suprisingly to me) throws a NullPointerException, so my intent was to preserve the original interface.
</blockquote></details>

weird, can we change that while we're at it? 
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 193 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJwimaDZr-tVeILmSB:-MXXlZ3PDYdxuGNEAsPn:b-xe86fw) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/model/server/Lock.java#L193)):*

Yeah, I agree.   I didn't really want to get into that in the context of this PR, my intent here was to make the datastore locks (i.e. the ones that we're actually using) work correctly, since they were broken in SQL mode.

I've added this as a discussion topic for today.
</blockquote></details>

Yeah the more I think about it the more i think we should just do what we did for Cursors.  We will discuss today.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1053)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java, line 322 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJyCU51hSCENGgbGsZ:-MXY2wZOCtsqVXkROMAR:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java#L322)):*

weird, can we change that while we're at it? 
</blockquote></details>

Done.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1053)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 193 at r1](https://reviewable.io/reviews/google/nomulus/1053#-MXJwimaDZr-tVeILmSB:-MXbfJpU-ZEJyFHd908w:b-gu0gkf) ([raw file](https://github.com/google/nomulus/blob/f161eff7a2b3a3fd582ec77d82e1574ecd1af170/core/src/main/java/google/registry/model/server/Lock.java#L193)):*

Yeah the more I think about it the more i think we should just do what we did for Cursors.  We will discuss today. 
</blockquote></details>

As discussed yesterday: I'll remove the other lock class and make this code portable.  However, I'd like to do so in the next PR.

gbrodman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1053#-:-MXbo7s51a4nWSfrnT9S:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1054, opened 2021-04-02
Add a CompareReservedListCommand

sarahcaseybot:
This command loads the list of all ReservedLists in Datastore and Cloud SQL and returns the name of any list that is either missing from one of the DBs or that has differing content. The output of this command does not indicate what the actual diff between the lists are, so GetReservedListCommand should be used to actually view the contents of the offending lists. 

I already ran this comparison test on Sandbox and Production and fixed all the issues it found. (One list I missed in Sandbox when resaving, and 7 lists that were only in Cloud SQL in production.)

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1054)

### DIFF
--- a/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
+++ b/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
@@ -49,19 +49,14 @@ final class CompareReservedListsCommand implements CommandWithRemoteApi {
                         .map(ReservedList::getName)
                         .collect(toImmutableSet()));
 
-    ImmutableSet<String> notInCloudSql =
-        Sets.difference(datastoreLists, cloudSqlLists).immutableCopy();
-    ImmutableSet<String> notInDatastore =
-        Sets.difference(cloudSqlLists, datastoreLists).immutableCopy();
-
     int listsWithDiffs = 0;
 
-    for (String listName : notInCloudSql) {
+    for (String listName : Sets.difference(datastoreLists, cloudSqlLists).immutableCopy()) {
       listsWithDiffs++;
       System.out.printf(
           "ReservedList '%s' is present in Datastore, but not in Cloud SQL.%n", listName);
     }
-    for (String listName : notInDatastore) {
+    for (String listName : Sets.difference(cloudSqlLists, datastoreLists).immutableCopy()) {
       listsWithDiffs++;
       System.out.printf(
           "ReservedList '%s' is present in Cloud SQL, but not in Datastore.%n", listName);
--- a/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
+++ b/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
@@ -51,12 +51,12 @@ final class CompareReservedListsCommand implements CommandWithRemoteApi {
 
     int listsWithDiffs = 0;
 
-    for (String listName : Sets.difference(datastoreLists, cloudSqlLists).immutableCopy()) {
+    for (String listName : Sets.difference(datastoreLists, cloudSqlLists)) {
       listsWithDiffs++;
       System.out.printf(
           "ReservedList '%s' is present in Datastore, but not in Cloud SQL.%n", listName);
     }
-    for (String listName : Sets.difference(cloudSqlLists, datastoreLists).immutableCopy()) {
+    for (String listName : Sets.difference(cloudSqlLists, datastoreLists)) {
       listsWithDiffs++;
       System.out.printf(
           "ReservedList '%s' is present in Cloud SQL, but not in Datastore.%n", listName);
--- a/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
+++ b/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
@@ -61,7 +61,7 @@ final class CompareReservedListsCommand implements CommandWithRemoteApi {
       if (!sqlList.isPresent()) {
         listsWithDiffs++;
         System.out.printf(
-            "ReservedList with name %s is present in Datastore, but not in Cloud SQL%n",
+            "ReservedList '%s' is present in Datastore, but not in Cloud SQL.%n",
             datastoreList.getName());
       } else {
         ImmutableMap<String, ReservedListEntry> namesInSql =
@@ -79,7 +79,7 @@ final class CompareReservedListsCommand implements CommandWithRemoteApi {
         if (!namesInDatastore.equals(namesInSql)) {
           listsWithDiffs++;
           System.out.printf(
-              "ReservedList with name %s has different entries in each database%n",
+              "ReservedList '%s' has different entries in each database.%n",
               datastoreList.getName());
         }
       }
@@ -91,11 +91,11 @@ final class CompareReservedListsCommand implements CommandWithRemoteApi {
       if (!datastoreList.isPresent()) {
         listsWithDiffs++;
         System.out.printf(
-            "ReservedList with name %s is present in Cloud SQL, but not in Datastore%n",
+            "ReservedList '%s' is present in Cloud SQL, but not in Datastore.%n",
             sqlList.getName());
       }
     }
 
-    System.out.printf("Found %s unequal list(s).%n", listsWithDiffs);
+    System.out.printf("Found %d unequal list(s).%n", listsWithDiffs);
   }
 }
--- a/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
+++ b/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
@@ -1,0 +1,112 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools;
+
+import static com.google.common.collect.ImmutableSortedSet.toImmutableSortedSet;
+import static google.registry.model.common.EntityGroupRoot.getCrossTldKey;
+import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+
+import com.beust.jcommander.Parameters;
+import com.google.appengine.repackaged.com.google.common.collect.Streams;
+import com.google.common.collect.ImmutableSet;
+import google.registry.model.registry.label.ReservedList;
+import google.registry.model.registry.label.ReservedList.ReservedListEntry;
+import google.registry.model.registry.label.ReservedListDatastoreDao;
+import google.registry.model.registry.label.ReservedListSqlDao;
+import java.util.Comparator;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/** Command to compare all ReservedLists in Datastore to all ReservedLists in Cloud SQL. */
+@Parameters(
+    separators = " =",
+    commandDescription = "Compare all the ReservedLists in Datastore to those in Cloud SQL.")
+final class CompareReservedListsCommand implements CommandWithRemoteApi {
+
+  @Override
+  public void run() {
+    ImmutableSet<ReservedList> datastoreLists =
+        ImmutableSet.copyOf(
+            ofy().load().type(ReservedList.class).ancestor(getCrossTldKey()).list());
+
+    ImmutableSet<ReservedList> cloudSqlLists =
+        jpaTm()
+            .transact(
+                () ->
+                    jpaTm().loadAllOf(ReservedList.class).stream()
+                        .map(ReservedList::getName)
+                        .map(ReservedListSqlDao::getLatestRevision)
+                        .filter(Optional::isPresent)
+                        .map(Optional::get)
+                        .collect(
+                            toImmutableSortedSet(Comparator.comparing(ReservedList::getName))));
+
+    int listsWithDiffs = 0;
+
+    for (ReservedList datastoreList : datastoreLists) {
+      Optional<ReservedList> sqlList =
+          ReservedListSqlDao.getLatestRevision(datastoreList.getName());
+      if (!sqlList.isPresent()) {
+        listsWithDiffs++;
+        System.out.printf(
+            "ReservedList with name %s is present in Datastore, but not in Cloud SQL%n",
+            datastoreList.getName());
+      } else {
+        String datastoreListString =
+            Streams.stream(
+                    ReservedListDatastoreDao.getLatestRevision(datastoreList.getName())
+                        .get()
+                        .getReservedListEntries()
+                        .values())
+                .sorted(Comparator.comparing(ReservedListEntry::getLabel))
+                .map(ReservedListEntry::toString)
+                .collect(Collectors.joining("\n"));
+
+        String sqlListString =
+            Streams.stream(
+                    ReservedListSqlDao.getLatestRevision(datastoreList.getName())
+                        .get()
+                        .getReservedListEntries()
+                        .values())
+                .sorted(Comparator.comparing(ReservedListEntry::getLabel))
+                .map(ReservedListEntry::toString)
+                .collect(Collectors.joining("\n"));
+
+        // This will only print out the name of the unequal list. GetReservedListCommand should be
+        // used to determine what the actual differences are.
+        if (!datastoreListString.equals(sqlListString)) {
+          listsWithDiffs++;
+          System.out.printf(
+              "ReservedList with name %s has different entries in each database%n",
+              datastoreList.getName());
+        }
+      }
+    }
+
+    for (ReservedList sqlList : cloudSqlLists) {
+      Optional<ReservedList> datastoreList =
+          ReservedListDatastoreDao.getLatestRevision(sqlList.getName());
+      if (!datastoreList.isPresent()) {
+        listsWithDiffs++;
+        System.out.printf(
+            "ReservedList with name %s is present in Cloud SQL, but not in Datastore%n",
+            sqlList.getName());
+      }
+    }
+
+    System.out.printf("Found %s unequal lists.%n", listsWithDiffs);
+  }
+}
--- a/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
+++ b/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java
@@ -20,7 +20,7 @@ import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 
 import com.beust.jcommander.Parameters;
-import com.google.appengine.repackaged.com.google.common.collect.Streams;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import google.registry.model.registry.label.ReservedList;
 import google.registry.model.registry.label.ReservedList.ReservedListEntry;
@@ -28,7 +28,6 @@ import google.registry.model.registry.label.ReservedListDatastoreDao;
 import google.registry.model.registry.label.ReservedListSqlDao;
 import java.util.Comparator;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
 /** Command to compare all ReservedLists in Datastore to all ReservedLists in Cloud SQL. */
 @Parameters(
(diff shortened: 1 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJucab1rmT-G-MGFY6:-MXJucab1rmT-G-MGFY7:b9cu3rd) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L68)):*
> ```Java
>             datastoreList.getName());
>       } else {
>         String datastoreListString =
> ```

Why are you comparing a sort of toString value for each list rather than just comparing the entries themselves? 

It's likely cleaner to just do something like 

```
ImmutableSet<String> namesInSql = <load the names from SQL>
ImmutableSet<String> namesInDatastore = <load the names from ds>
Set<String> notInDatastore = Sets.difference(namesInsSql, namesInDatastore)
Set<String> notInSql = Sets.difference(namesInDatastore, namesInSql)

<do things with the differences if they're not empty>

for name in Sets.intersection(namesInSql, namesInDatastore):
  load the list from both dbs, compare the entry maps
```
___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJu3z2EOFBPk2h9qFe:-MXJu3z2EOFBPk2h9qFf:blgdb7q) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L110)):*
> ```Java
>     }
> 
>     System.out.printf("Found %s unequal lists.%n", listsWithDiffs);
> ```

nit: `list(s)`
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 33 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJtnK68Mh9JkEr0-Fl:-MXJtnK68Mh9JkEr0-Fm:b-g2qp8z) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L33)):*
> ```Java
> public class CompareReservedListCommandTest extends CommandTestCase<CompareReservedListsCommand> {
> 
>   private ImmutableMap<String, ReservedListEntry> reservations;
> ```

these two don't need to be separate instance variables

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJucab1rmT-G-MGFY6:-MXZAakB3rPLQ5qs9sin:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L68)):*

Why are you comparing a sort of toString value for each list rather than just comparing the entries themselves? 

It's likely cleaner to just do something like 

```
ImmutableSet<String> namesInSql = <load the names from SQL>
ImmutableSet<String> namesInDatastore = <load the names from ds>
Set<String> notInDatastore = Sets.difference(namesInsSql, namesInDatastore)
Set<String> notInSql = Sets.difference(namesInDatastore, namesInSql)

<do things with the differences if they're not empty>

for name in Sets.intersection(namesInSql, namesInDatastore):
  load the list from both dbs, compare the entry maps
```
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 33 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJtnK68Mh9JkEr0-Fl:-MXZAqyT6JY1smg-YMEi:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L33)):*

these two don't need to be separate instance variables
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: 1 of 3 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 64 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcygEp7YVroXpTZYEj:-MXcygEp7YVroXpTZYEk:bxxhku1) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L64)):*
> ```Java
>         listsWithDiffs++;
>         System.out.printf(
>             "ReservedList with name %s is present in Datastore, but not in Cloud SQL%n",
> ```

Missing period here and elsewhere.
___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 82 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcyjxL27eBOCf6QtPV:-MXcyjxL27eBOCf6QtPW:b-id9yox) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L82)):*
> ```Java
>           listsWithDiffs++;
>           System.out.printf(
>               "ReservedList with name %s has different entries in each database%n",
> ```

Why isn't this printing out the differences?  Is that not useful?  If the differences persist even after re-running the dual-write for some reason (like dual write is bugged), you'd definitely want to know what the differences actually are as your next step.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 43 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcxxLk6MNq3atDh_5q:-MXcxxLk6MNq3atDh_5r:b-twjt59) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L43)):*
> ```Java
>             .setLastUpdateTime(fakeClock.nowUtc())
>             .setShouldPublish(false)
>             .setReservedListMap(
> ```

There's an easier way to do this; see https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/DatabaseHelper.java;l=350
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 71 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcy2s45SYhBWA-RXVA:-MXcy2s45SYhBWA-RXVB:b-nbniwp) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L71)):*
> ```Java
>   void test_success() throws Exception {
>     runCommand();
>     String stdout = getStdoutAsString();
> ```

Can be inlined (elsewhere too).  `assertThat(getStdoutAsString()).....` is perfectly reasonable.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcyK0H0YNWqmVFDdo1:-MXcyK0H0YNWqmVFDdo2:bkspzau) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L77)):*
> ```Java
>   @Test
>   void test_listMissingFromCloudSql() throws Exception {
>     jpaTm().transact(() -> jpaTm().delete(reservedList));
> ```

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/DatabaseHelper.java;drc=eabf056f9b2347f863dcc0334334d6219906e6e9;l=1241 makes these simpler.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 102 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcySZU7QwWFQX0Ds2O:-MXcySZU7QwWFQX0Ds2P:bdwo88c) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L102)):*
> ```Java
>         ImmutableMap.of(
>             "food",
>             ReservedListEntry.create("food", ReservationType.RESERVED_FOR_SPECIFIC_USE, null));
> ```

Another good use case for using `persistReservedList(String listName, String... lines)`

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: 1 of 3 files reviewed, 10 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcz6JR2Io32d-bHPKD:-MXcz6JR2Io32d-bHPKE:b-pyjuo5) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L94)):*
> ```Java
>         listsWithDiffs++;
>         System.out.printf(
>             "ReservedList with name %s is present in Cloud SQL, but not in Datastore%n",
> ```

It's a minor nit but `ReservedList '%s'` reads cleaner to me than `ReservedList with name %s`.  The latter is just a little bit clunky and overly long.
___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 99 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcz3hv6COigSnmZmEH:-MXcz3hv6COigSnmZmEI:bb38a0m) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L99)):*
> ```Java
>     }
> 
>     System.out.printf("Found %s unequal list(s).%n", listsWithDiffs);
> ```

Should be `%d`

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: 1 of 3 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 64 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcygEp7YVroXpTZYEj:-MXd7oG65gvqkoTIrNxH:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L64)):*

Missing period here and elsewhere.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 99 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcz3hv6COigSnmZmEH:-MXd9w7WE1_ItD-JSfaX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L99)):*

Should be `%d`
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 71 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcy2s45SYhBWA-RXVA:-MXdF73T7TLgErwhz_1P:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L71)):*

Can be inlined (elsewhere too).  `assertThat(getStdoutAsString()).....` is perfectly reasonable.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 77 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcyK0H0YNWqmVFDdo1:-MXdF7qJBQVdGYT0bNlE:b1ki99g) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L77)):*

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/DatabaseHelper.java;drc=eabf056f9b2347f863dcc0334334d6219906e6e9;l=1241 makes these simpler.
</blockquote></details>

This won't work since that will only delete from Datastore right now
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 102 at r2](https://reviewable.io/reviews/google/nomulus/1054#-MXcySZU7QwWFQX0Ds2O:-MXdG5zM3XvrnWNUO5Xs:b10tu2r) ([raw file](https://github.com/google/nomulus/blob/9a0325c01520dd103d1a6bac8f1b435012b9296f/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L102)):*

Another good use case for using `persistReservedList(String listName, String... lines)`
</blockquote></details>

persistReservedList would not work here because it would persist the differing list to both DBs and we only want it in datastore for this test

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJucab1rmT-G-MGFY6:-MXdlUVa7A4Cj9kT0bmj:b-7adykc) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L68)):*

Done.
</blockquote></details>

I don't think this is quite done -- you're still doing  some of the "do the databases contain lists with the same names" logic inside the datastoreList for-loop. 

If you compare the name sets initially, you can later loop over only the intersection of the lists and thus guarantee that the list you're checking exists in both DBs.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 60 at r3](https://reviewable.io/reviews/google/nomulus/1054#-MXdlsnoCjSHk3pqr3Gv:-MXdlsnoCjSHk3pqr3Gw:b-lkjbuj) ([raw file](https://github.com/google/nomulus/blob/da3f9a071b94d6c76e3b6fedf80718bad7e6098d/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L60)):*
> ```Java
>   @Test
>   void test_listMissingFromDatastore() throws Exception {
>     ofyTm()
> ```

unlike the comment on the previous test, I do believe we should be able to use DatabaseHelper::deleteResource here? 
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 93 at r3](https://reviewable.io/reviews/google/nomulus/1054#-MXdm-8FAetl9z_1Nwjs:-MXdm-8FAetl9z_1Nwjt:by2pmip) ([raw file](https://github.com/google/nomulus/blob/da3f9a071b94d6c76e3b6fedf80718bad7e6098d/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L93)):*
> ```Java
>   @Test
>   void test_listsDifferAndMissing() throws Exception {
>     ofyTm()
> ```

ditto comment re: using DH:deleteResource?

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 60 at r3](https://reviewable.io/reviews/google/nomulus/1054#-MXdlsnoCjSHk3pqr3Gv:-MXmAdio4pVxVe1Z2k-6:b-i4t7xg) ([raw file](https://github.com/google/nomulus/blob/da3f9a071b94d6c76e3b6fedf80718bad7e6098d/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L60)):*

unlike the comment on the previous test, I do believe we should be able to use DatabaseHelper::deleteResource here? 
</blockquote></details>

Yes, it would only delete from Datastore now, but after the asynchronous migration, it will only delete from Cloud SQL which is the wrong behavior for this test. The Datastore code for ReservedList will likely already be deleted by the time asynchronous migration happens, but in case it is not, I think this test should specify the correct transaction manager to use.
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 93 at r3](https://reviewable.io/reviews/google/nomulus/1054#-MXdm-8FAetl9z_1Nwjs:-MXmBA1n11u43A8df05i:b-33b6k6) ([raw file](https://github.com/google/nomulus/blob/da3f9a071b94d6c76e3b6fedf80718bad7e6098d/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L93)):*

ditto comment re: using DH:deleteResource?
</blockquote></details>

See comment above

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJucab1rmT-G-MGFY6:-MXmeG_d8iUCMbIHSy64:b-uwi2fs) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L68)):*

I'm not quite sure I understand what you're saying here. The if statement checks that the list with the name in Datastore also exists in cloud SQL and only checks the content of the list if it is.
</blockquote></details>

The point is that in order to reduce indentation and reduce the complexity of each loop, the "do the stores contain the same lists?" and "do the lists that are in both DBs match each other?" questions should be answered in different sections, i.e. the pseudocode snippet above 

that way, when we're loading the contents themselves we don't need to worry about whether or not the list is in both DBs since we already checked that separately 
___
*[core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java, line 60 at r3](https://reviewable.io/reviews/google/nomulus/1054#-MXdlsnoCjSHk3pqr3Gv:-MXmea7I87XX3fgWPPsD:b-mtzqx) ([raw file](https://github.com/google/nomulus/blob/da3f9a071b94d6c76e3b6fedf80718bad7e6098d/core/src/test/java/google/registry/tools/CompareReservedListCommandTest.java#L60)):*

Yes, it would only delete from Datastore now, but after the asynchronous migration, it will only delete from Cloud SQL which is the wrong behavior for this test. The Datastore code for ReservedList will likely already be deleted by the time asynchronous migration happens, but in case it is not, I think this test should specify the correct transaction manager to use.
</blockquote></details>

fair enough

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1054#-MXJucab1rmT-G-MGFY6:-MXn1XsS6TAygvRMMBfM:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bb7a78226ab2a62987a9ecad6b465057b26240a8/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L68)):*

The point is that in order to reduce indentation and reduce the complexity of each loop, the "do the stores contain the same lists?" and "do the lists that are in both DBs match each other?" questions should be answered in different sections, i.e. the pseudocode snippet above 

that way, when we're loading the contents themselves we don't need to worry about whether or not the list is in both DBs since we already checked that separately 
</blockquote></details>

Done.

gbrodman (approved):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054#-:-MXswBt48rcCgN09Tj9d:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/CompareReservedListsCommand.java, line 59 at r4](https://reviewable.io/reviews/google/nomulus/1054#-MXsx29J3jy19rxknQsN:-MXsx29J3jy19rxknQsO:bqvfu4k) ([raw file](https://github.com/google/nomulus/blob/aa8a68e3ed727b5af6552e5fb807775ec6d41f4b/core/src/main/java/google/registry/tools/CompareReservedListsCommand.java#L59)):*
> ```Java
>     int listsWithDiffs = 0;
> 
>     for (String listName : notInCloudSql) {
> ```

nit, which is my bad since it was my suggestion to begin with: you can inline the Sets.difference() call in this if statement since I don't think we need it anywhere else (and same for the other similar one)

gbrodman (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1054)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #1060, opened 2021-04-06
Partially convert EppResourceUtils to SQL

gbrodman:
Some of the rest will depend on b/184578521.

The primary conversion in this PR is the change in
NameserverLookupByIpCommand as that is the only place where the removed
EppResourceUtils method was called. We also convert to DualDatabaseTest
the tests of the callers of NLBIC. and use a CriteriaQueryBuilder in the
foreign key index SQL lookup (allowing us to avoid the String.format
call).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1060)

### DIFF
--- a/core/src/main/java/google/registry/model/EppResourceUtils.java
+++ b/core/src/main/java/google/registry/model/EppResourceUtils.java
@@ -195,30 +195,6 @@ public final class EppResourceUtils {
   }
 
   /**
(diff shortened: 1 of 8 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1060)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1062, opened 2021-04-07
Add a beforeSqlSave callback to ReplaySpecializer

gbrodman:
When in the Datastore-primary and SQL-secondary stage, we will want to
save the EppResource-at-this-point-in-time field in the *History
objects so that later on we can examine the *History objects to see what
the resource looked like at that point in time.

Without this PR, the full object at that point in time would be lost
during the asynchronous replay since Datastore doesn't know about it.

In addition, we modify the HistoryEntry weight / priority so that
additions to it come after the additions to the resource off of which it
is based.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1062)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1062)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[db/src/main/resources/sql/flyway/V91__defer_fkeys.sql, line 48 at r1](https://reviewable.io/reviews/google/nomulus/1062#-MXn1_9VBCitjL0eSFXE:-MXn1_9VBCitjL0eSFXF:b-rzxnb3) ([raw file](https://github.com/google/nomulus/blob/890db2f9b42d847c64c4ef371423f1a988f472c8/db/src/main/resources/sql/flyway/V91__defer_fkeys.sql#L48)):*
> ```TSQL
>    FOREIGN KEY (contact_repo_id, contact_history_revision_id)
>    REFERENCES "ContactHistory"(contact_repo_id, history_revision_id)
>    DEFERRABLE INITIALLY DEFERRED;
> ```

Can you also remove the "INITIALLY DEFERRED" status from the fields where we added it based on the existing order?  It should be pretty easy to identify them, as they should be present in flyway files.  Basically, if the deferred status is on a foreign key that is in the same direction as the new order, you'll want to restore it.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1062)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[db/src/main/resources/sql/flyway/V91__defer_fkeys.sql, line 48 at r1](https://reviewable.io/reviews/google/nomulus/1062#-MXn1_9VBCitjL0eSFXE:-MXn7Rk025_UQVPfTvP5:bxsxda6) ([raw file](https://github.com/google/nomulus/blob/890db2f9b42d847c64c4ef371423f1a988f472c8/db/src/main/resources/sql/flyway/V91__defer_fkeys.sql#L48)):*

Can you also remove the "INITIALLY DEFERRED" status from the fields where we added it based on the existing order?  It should be pretty easy to identify them, as they should be present in flyway files.  Basically, if the deferred status is on a foreign key that is in the same direction as the new order, you'll want to restore it.
</blockquote></details>

nah I think we want to keep fk_domain_history_domain_repo_id as deferrable because we don't guarantee the order of saves in many of the flows, including any that use the `EntityChanges` class, in which the objects to save are kept in a Set (so order is of course not guaranteed).

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1062)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[db/src/main/resources/sql/flyway/V91__defer_fkeys.sql, line 48 at r1](https://reviewable.io/reviews/google/nomulus/1062#-MXn1_9VBCitjL0eSFXE:-MXqSOXfDYbE72ZKPCv2:b-ltym96) ([raw file](https://github.com/google/nomulus/blob/890db2f9b42d847c64c4ef371423f1a988f472c8/db/src/main/resources/sql/flyway/V91__defer_fkeys.sql#L48)):*

nah I think we want to keep fk_domain_history_domain_repo_id as deferrable because we don't guarantee the order of saves in many of the flows, including any that use the `EntityChanges` class, in which the objects to save are kept in a Set (so order is of course not guaranteed). 
</blockquote></details>

We've so far been guaranteeing the order of saves under the theory that it's preferable to fail foreign key constraints early when possible.  If we're going to abandon that practice (and I have to say, I think we should) then it should be discussed more broadly than this PR.

If we do abandon it, then all foreign keys become initially deferred and I think we can remove the prioritization code entirely.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1062)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[db/src/main/resources/sql/flyway/V91__defer_fkeys.sql, line 48 at r1](https://reviewable.io/reviews/google/nomulus/1062#-MXn1_9VBCitjL0eSFXE:-MXsvbHb0cFpaLT_ON12:b7f2mo1) ([raw file](https://github.com/google/nomulus/blob/890db2f9b42d847c64c4ef371423f1a988f472c8/db/src/main/resources/sql/flyway/V91__defer_fkeys.sql#L48)):*

We've so far been guaranteeing the order of saves under the theory that it's preferable to fail foreign key constraints early when possible.  If we're going to abandon that practice (and I have to say, I think we should) then it should be discussed more broadly than this PR.

If we do abandon it, then all foreign keys become initially deferred and I think we can remove the prioritization code entirely.

</blockquote></details>

I don't think we'd want to remove it entirely for the same reason that this PR exists: in this case, we want the DomainBase object to be saved before the DomainHistory object so that we can fill the latter with the proper value of the former. 

Either way yes I agree we should make them all deferrable. How about we discuss that later and have this PR go through separately, not blocking on that?

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1062)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V91__defer_fkeys.sql, line 48 at r1](https://reviewable.io/reviews/google/nomulus/1062#-MXn1_9VBCitjL0eSFXE:-MY53w15EiulZ-Oa_y41:b-fnsc6b) ([raw file](https://github.com/google/nomulus/blob/890db2f9b42d847c64c4ef371423f1a988f472c8/db/src/main/resources/sql/flyway/V91__defer_fkeys.sql#L48)):*

I don't think we'd want to remove it entirely for the same reason that this PR exists: in this case, we want the DomainBase object to be saved before the DomainHistory object so that we can fill the latter with the proper value of the former. 

Either way yes I agree we should make them all deferrable. How about we discuss that later and have this PR go through separately, not blocking on that? 
</blockquote></details>

That's actually a very different intent.  The prioritization code was written specifically to allow us to retain non-deferred foreign keys.  This is the first other use-case for prioritization, and, IMO, kind of a scary way to solve the problem of attaching the latest EPP resource to its history record -- though I can't think of a better way to do it.  But ultimately, if this is why we're keeping the prioritization code I think we just end up with two priorities: History records and everything else.

To your second point, I'm ok with unblocking this PR and continuing the discussion out-of-band (mainly just want to make sure @weiminyu is on-board, at this point).

PR #1066, opened 2021-04-08
Try again to fix a flaky test

weiminyu:
Fix DeleteExpiredDomainsActionTest.test_deletesThreeDomainsInOneRun

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1066)

### DIFF
--- a/core/src/test/java/google/registry/batch/DeleteExpiredDomainsActionTest.java
+++ b/core/src/test/java/google/registry/batch/DeleteExpiredDomainsActionTest.java
@@ -16,6 +16,7 @@ package google.registry.batch;
 
 import static com.google.common.truth.Truth.assertThat;
 import static google.registry.model.eppcommon.StatusValue.PENDING_DELETE;
+import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.model.reporting.HistoryEntry.Type.DOMAIN_CREATE;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.testing.DatabaseHelper.createTld;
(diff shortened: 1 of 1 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1066)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[core/src/test/java/google/registry/batch/DeleteExpiredDomainsActionTest.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/1066#-MXm2e_q3sviLpDD3hGN:-MXm2e_r7wQQ5ugC6RuT:bl98vqp) ([raw file](https://github.com/google/nomulus/blob/e8eaa2605de607f2ebe14154c2c4b9177f14df03/core/src/test/java/google/registry/batch/DeleteExpiredDomainsActionTest.java#L132)):*
> ```Java
>     // action.run() executes an ancestor-less query which is subject to eventual consistency (it
>     // uses an index that is updated asynchronously). For a deterministic test outcome, we busy
>     // wait here until all domains above are returned by the query.
> ```

Good explanation, but I'm surprised that this is an issue in the datastore fake (I can see where it would be in the real datastore) unless they're attempting to emulate eventual consistency or doing so simply by virtue of their implementation.

I guess we'll see.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1066)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/test/java/google/registry/batch/DeleteExpiredDomainsActionTest.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/1066#-MXm2e_q3sviLpDD3hGN:-MXm9Mdk62K84XKIjWZk:b-b5kdpz) ([raw file](https://github.com/google/nomulus/blob/e8eaa2605de607f2ebe14154c2c4b9177f14df03/core/src/test/java/google/registry/batch/DeleteExpiredDomainsActionTest.java#L132)):*

Good explanation, but I'm surprised that this is an issue in the datastore fake (I can see where it would be in the real datastore) unless they're attempting to emulate eventual consistency or doing so simply by virtue of their implementation.

I guess we'll see.
</blockquote></details>

The internal spanner testdb behaves this way so I won't be surprised to find it the case here too.

PR #1068, opened 2021-04-09
Convert TmchCrl and ServerSecret to cleaner tm() impls

gbrodman:
When I implemented this originally I knew a lot less than I know now
about how we'll be storing and retrieving these singletons from SQL. The
optimal way here is to use the single SINGLETON_ID as the primary key,
that way we always know how to create the key that we can use in the
tm() retrieval.

This allows us to use generic tm() methods and to remove the handcrafted
SQL queries.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1068)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1068)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/server/ServerSecret.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1068#-MY5RMdtDhZf9XCIcjZE:-MY5RMdtDhZf9XCIcjZF:b28xf94) ([raw file](https://github.com/google/nomulus/blob/6afdea50fe084472dfcffd08c934fbd2c318502f/core/src/main/java/google/registry/model/server/ServerSecret.java#L76)):*
> ```Java
>     return transactIfJpaTm(
>         () -> {
>           // Attempt a quick load if we're in ofy first to short-circuit sans transaction
> ```

I have to say, I think I would have restructured this and just wrapped the whole thing in an`isOfy()` -- if nothing else it will be a lot easier to remove.

That said, if you don't want to do that, can we at least restructure the second check so that it doesn't happen in JPA?  Maybe something like:

    if (!secret.isPresent()) {
      secret = tm().ifOfy() ? tm().loadByKeyIfPresent(vkey) :  
    }
    if (!secret.isPresent()) {
      ... store new one ...
    }
___
*[core/src/main/java/google/registry/model/server/ServerSecret.java, line 87 at r1](https://reviewable.io/reviews/google/nomulus/1068#-MY5UAZ31W11IMCyDcjd:-MY5UAZ45ikeZNl-8sjj:btfrfdd) ([raw file](https://github.com/google/nomulus/blob/6afdea50fe084472dfcffd08c934fbd2c318502f/core/src/main/java/google/registry/model/server/ServerSecret.java#L87)):*
> ```Java
>                             if (!secondSecret.isPresent()) {
>                               secondSecret = Optional.of(create(UUID.randomUUID()));
>                               tm().putWithoutBackup(secondSecret.get());
> ```

Do we have an `insertWithoutBackup()`?  I think that's what we'd want to do here.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1068)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/server/ServerSecret.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1068#-MY5RMdtDhZf9XCIcjZE:-MY5pq5EAULHCAvXE0Jl:boalu99) ([raw file](https://github.com/google/nomulus/blob/6afdea50fe084472dfcffd08c934fbd2c318502f/core/src/main/java/google/registry/model/server/ServerSecret.java#L76)):*

I have to say, I think I would have restructured this and just wrapped the whole thing in an`isOfy()` -- if nothing else it will be a lot easier to remove.

That said, if you don't want to do that, can we at least restructure the second check so that it doesn't happen in JPA?  Maybe something like:

    if (!secret.isPresent()) {
      secret = tm().ifOfy() ? tm().loadByKeyIfPresent(vkey) :  
    }
    if (!secret.isPresent()) {
      ... store new one ...
    }
</blockquote></details>

yeah i think using the isOfy check allows us to reduce indentation / functional complexity (as well as making it clear what to remove whenever we're in SQL-only-land) 
___
*[core/src/main/java/google/registry/model/server/ServerSecret.java, line 87 at r1](https://reviewable.io/reviews/google/nomulus/1068#-MY5UAZ31W11IMCyDcjd:-MY5q6COA7dNEnvTrvAk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6afdea50fe084472dfcffd08c934fbd2c318502f/core/src/main/java/google/registry/model/server/ServerSecret.java#L87)):*

Do we have an `insertWithoutBackup()`?  I think that's what we'd want to do here.
</blockquote></details>

Done.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1068)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

PR #1069, opened 2021-04-09
Implement query abstraction

mindhog:
Implement a query abstraction layer ("QueryComposer") that allows us to
construct fluent-style queries that work across both Objectify and JPA.

As a demonstration of the concept, convert Spec11EmailUtils and its test to
use the new API.

Limitations:
-  The primary limitations of this system are imposed by datastore, for
   example all queryable fields must be indexed, orderBy must coincide with
   the order of any inequality queries, inequality filters are limited to one
   property...
-  JPA queries are limited to a set of where clauses (all of which must match)
   and an "order by" clause.  Joins, functions, complex where logic and
   multi-table queries are simply not allowed.
-  Descending sort order is currently unsupported (this is simple enough to
   add).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1069)

### DIFF
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -379,7 +379,7 @@ public class DatastoreTransactionManager implements TransactionManager {
 
     Query<T> buildQuery() {
       Query<T> result = ofy().load().type(entityClass);
-      for (WhereCondition pred : predicates) {
+      for (WhereClause pred : predicates) {
         result = result.filter(pred.fieldName + pred.comparator.getDatastoreString(), pred.value);
       }
 
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -39,6 +39,7 @@ import google.registry.model.reporting.HistoryEntry;
 import google.registry.persistence.VKey;
 import google.registry.persistence.transaction.QueryComposer;
 import google.registry.persistence.transaction.TransactionManager;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -46,6 +47,7 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import javax.annotation.Nullable;
 import javax.persistence.NoResultException;
+import javax.persistence.NonUniqueResultException;
 import org.joda.time.DateTime;
 
 /** Datastore implementation of {@link TransactionManager}. */
@@ -391,19 +393,26 @@ public class DatastoreTransactionManager implements TransactionManager {
     }
 
     @Override
-    public T first() {
-      T result = buildQuery().first().now();
-      if (result == null) {
+    public Optional<T> first() {
+      return Optional.ofNullable(buildQuery().first().now());
+    }
+
+    @Override
+    public T getSingleResult() {
+      List<T> results = buildQuery().limit(2).list();
+      if (results.size() == 0) {
         // The exception text here is the same as what we get for JPA queries.
         throw new NoResultException("No entity found for query");
+      } else if (results.size() > 1) {
+        throw new NonUniqueResultException("More than one result found for getSingleResult query");
       }
-      return result;
+      return results.get(0);
     }
 
     @Override
     public Stream<T> stream() {
       // TODO: there should be a better way to do this.
-      return buildQuery().list().stream();
+      return Streams.stream(buildQuery());
     }
   }
 }
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -37,12 +37,15 @@ import google.registry.model.domain.DomainHistory;
 import google.registry.model.host.HostHistory;
 import google.registry.model.reporting.HistoryEntry;
 import google.registry.persistence.VKey;
+import google.registry.persistence.transaction.QueryComposer;
 import google.registry.persistence.transaction.TransactionManager;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import javax.annotation.Nullable;
+import javax.persistence.NoResultException;
 import org.joda.time.DateTime;
 
 /** Datastore implementation of {@link TransactionManager}. */
@@ -303,6 +306,11 @@ public class DatastoreTransactionManager implements TransactionManager {
   }
 
   @Override
+  public <T> QueryComposer<T> createQueryComposer(Class<T> entity) {
+    return new QueryComposerImpl(entity);
+  }
+
+  @Override
   public void clearSessionCache() {
     getOfy().clearSessionCache();
   }
@@ -362,5 +370,40 @@ public class DatastoreTransactionManager implements TransactionManager {
       return (T) ((HistoryEntry) obj).toChildHistoryEntity();
     }
     return obj;
+  }
+
+  private static class QueryComposerImpl<T> extends QueryComposer<T> {
+    QueryComposerImpl(Class<T> entity) {
+      super(entity);
+    }
+
+    Query<T> buildQuery() {
+      Query<T> result = ofy().load().type(entity);
+      for (WhereCondition pred : predicates) {
+        result = result.filter(pred.fieldName + pred.comparator.getDatastoreString(), pred.value);
+      }
+
+      if (orderBy != null) {
+        result = result.order(orderBy);
+      }
+
+      return result;
+    }
+
+    @Override
+    public T first() {
+      T result = buildQuery().first().now();
+      if (result == null) {
+        // The exception text here is the same as what we get for JPA queries.
+        throw new NoResultException("No entity found for query");
+      }
+      return result;
+    }
+
+    @Override
+    public Stream<T> stream() {
+      // TODO: there should be a better way to do this.
+      return buildQuery().list().stream();
+    }
   }
 }
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -307,7 +307,7 @@ public class DatastoreTransactionManager implements TransactionManager {
 
   @Override
   public <T> QueryComposer<T> createQueryComposer(Class<T> entity) {
-    return new QueryComposerImpl(entity);
+    return new DatastoreQueryComposerImpl(entity);
   }
 
   @Override
@@ -372,13 +372,13 @@ public class DatastoreTransactionManager implements TransactionManager {
     return obj;
   }
 
-  private static class QueryComposerImpl<T> extends QueryComposer<T> {
-    QueryComposerImpl(Class<T> entity) {
-      super(entity);
+  private static class DatastoreQueryComposerImpl<T> extends QueryComposer<T> {
+    DatastoreQueryComposerImpl(Class<T> entityClass) {
+      super(entityClass);
     }
 
     Query<T> buildQuery() {
-      Query<T> result = ofy().load().type(entity);
+      Query<T> result = ofy().load().type(entityClass);
       for (WhereCondition pred : predicates) {
         result = result.filter(pred.fieldName + pred.comparator.getDatastoreString(), pred.value);
       }
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -411,7 +411,6 @@ public class DatastoreTransactionManager implements TransactionManager {
 
     @Override
     public Stream<T> stream() {
-      // TODO: there should be a better way to do this.
       return Streams.stream(buildQuery());
     }
   }
--- a/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java
+++ b/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java
@@ -18,6 +18,7 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 
 import com.google.common.collect.ImmutableList;
 import java.util.Collection;
+import javax.persistence.EntityManager;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Expression;
@@ -94,7 +95,12 @@ public class CriteriaQueryBuilder<T> {
 
   /** Creates a query builder that will SELECT from the given class. */
   public static <T> CriteriaQueryBuilder<T> create(Class<T> clazz) {
-    CriteriaQuery<T> query = jpaTm().getEntityManager().getCriteriaBuilder().createQuery(clazz);
+    return create(jpaTm().getEntityManager(), clazz);
+  }
+
+  /** Creates a query builder for the given entity manager. */
+  public static <T> CriteriaQueryBuilder<T> create(EntityManager em, Class<T> clazz) {
+    CriteriaQuery<T> query = em.getCriteriaBuilder().createQuery(clazz);
     Root<T> root = query.from(clazz);
     query = query.select(root);
     return new CriteriaQueryBuilder<>(query, root);
--- a/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java
+++ b/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java
@@ -36,7 +36,7 @@ import javax.persistence.criteria.Root;
 public class CriteriaQueryBuilder<T> {
 
   /** Functional interface that defines the 'where' operator, e.g. {@link CriteriaBuilder#equal}. */
-  public interface WhereClause<U> {
+  public interface WhereOperator<U> {
     Predicate predicate(Expression<U> expression, U object);
   }
 
@@ -51,7 +51,8 @@ public class CriteriaQueryBuilder<T> {
   }
 
   /** Adds a WHERE clause to the query, given the specified operation, field, and value. */
-  public <V> CriteriaQueryBuilder<T> where(String fieldName, WhereClause<V> whereClause, V value) {
+  public <V> CriteriaQueryBuilder<T> where(
+      String fieldName, WhereOperator<V> whereClause, V value) {
     Expression<V> expression = root.get(fieldName);
     return where(whereClause.predicate(expression, value));
   }
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -43,6 +43,7 @@ import google.registry.util.Clock;
 import google.registry.util.Retrier;
 import google.registry.util.SystemSleeper;
 import java.lang.reflect.Field;
+import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
@@ -712,7 +713,13 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     }
 
     @Override
-    public T first() {
+    public Optional<T> first() {
+      List<T> results = buildQuery().setMaxResults(1).getResultList();
+      return results.size() > 0 ? Optional.of(results.get(0)) : Optional.empty();
+    }
+
+    @Override
+    public T getSingleResult() {
       return buildQuery().getSingleResult();
     }
 
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -47,6 +47,7 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
@@ -54,6 +55,7 @@ import javax.persistence.EntityTransaction;
 import javax.persistence.PersistenceException;
 import javax.persistence.Query;
 import javax.persistence.TypedQuery;
+import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.metamodel.EntityType;
 import javax.persistence.metamodel.SingularAttribute;
 import org.joda.time.DateTime;
@@ -531,6 +533,11 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
   }
 
   @Override
+  public <T> QueryComposer<T> createQueryComposer(Class<T> entity) {
+    return new QueryComposerImpl<T>(entity, getEntityManager());
+  }
+
+  @Override
   public void clearSessionCache() {
     // This is an intended no-op method as there is no session cache in Postgresql.
   }
@@ -679,6 +686,42 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
           entityManager.persist(persistedTxn.toEntity());
         }
       }
+    }
+  }
+
+  private static class QueryComposerImpl<T> extends QueryComposer<T> {
+
+    EntityManager em;
+
+    QueryComposerImpl(Class<T> entity, EntityManager em) {
+      super(entity);
+      this.em = em;
+    }
+
+    private TypedQuery<T> buildQuery() {
+      CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
+      CriteriaQueryBuilder<T> builder = CriteriaQueryBuilder.create(em, entity);
+
+      for (WhereCondition pred : predicates) {
+        pred.addToCriteriaQueryBuilder(builder, criteriaBuilder);
+      }
+
+      if (orderBy != null) {
+        // TODO: Convert object field name to hibernate column names.
+        builder.orderByAsc(orderBy);
+      }
+
+      return em.createQuery(builder.build());
+    }
+
+    @Override
+    public T first() {
+      return buildQuery().getSingleResult();
+    }
+
+    @Override
+    public Stream<T> stream() {
+      return buildQuery().getResultList().stream();
     }
   }
 }
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -55,7 +55,6 @@ import javax.persistence.EntityTransaction;
 import javax.persistence.PersistenceException;
 import javax.persistence.Query;
 import javax.persistence.TypedQuery;
-import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.metamodel.EntityType;
 import javax.persistence.metamodel.SingularAttribute;
 import org.joda.time.DateTime;
@@ -699,11 +698,10 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     }
 
     private TypedQuery<T> buildQuery() {
-      CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
       CriteriaQueryBuilder<T> queryBuilder = CriteriaQueryBuilder.create(em, entityClass);
 
-      for (WhereCondition<?> pred : predicates) {
-        pred.addToCriteriaQueryBuilder(queryBuilder, criteriaBuilder);
+      for (WhereClause<?> pred : predicates) {
+        pred.addToCriteriaQueryBuilder(queryBuilder);
       }
 
       if (orderBy != null) {
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -534,7 +534,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
 
   @Override
   public <T> QueryComposer<T> createQueryComposer(Class<T> entity) {
-    return new QueryComposerImpl<T>(entity, getEntityManager());
+    return new JpaQueryComposerImpl<T>(entity, getEntityManager());
   }
 
   @Override
@@ -689,29 +689,28 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     }
   }
 
-  private static class QueryComposerImpl<T> extends QueryComposer<T> {
+  private static class JpaQueryComposerImpl<T> extends QueryComposer<T> {
 
     EntityManager em;
 
-    QueryComposerImpl(Class<T> entity, EntityManager em) {
-      super(entity);
+    JpaQueryComposerImpl(Class<T> entityClass, EntityManager em) {
+      super(entityClass);
       this.em = em;
     }
 
     private TypedQuery<T> buildQuery() {
       CriteriaBuilder criteriaBuilder = em.getCriteriaBuilder();
-      CriteriaQueryBuilder<T> builder = CriteriaQueryBuilder.create(em, entity);
+      CriteriaQueryBuilder<T> queryBuilder = CriteriaQueryBuilder.create(em, entityClass);
 
-      for (WhereCondition pred : predicates) {
-        pred.addToCriteriaQueryBuilder(builder, criteriaBuilder);
+      for (WhereCondition<?> pred : predicates) {
+        pred.addToCriteriaQueryBuilder(queryBuilder, criteriaBuilder);
       }
 
       if (orderBy != null) {
-        // TODO: Convert object field name to hibernate column names.
-        builder.orderByAsc(orderBy);
+        queryBuilder.orderByAsc(orderBy);
       }
 
-      return em.createQuery(builder.build());
+      return em.createQuery(queryBuilder.build());
     }
 
     @Override
@@ -721,7 +720,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
 
     @Override
     public Stream<T> stream() {
-      return buildQuery().getResultList().stream();
+      return buildQuery().getResultStream();
     }
   }
 }
--- a/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
+++ b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
@@ -20,6 +20,7 @@ import com.google.common.base.Function;
 import google.registry.persistence.transaction.CriteriaQueryBuilder.WhereOperator;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.persistence.criteria.CriteriaBuilder;
@@ -29,8 +30,12 @@ import javax.persistence.criteria.CriteriaBuilder;
  *
  * <p>Example usage:
  *
- * <p>tm().createQueryComposer(EntityType.class) .where("fieldName", Comparator.EQ, "value"
- * .orderBy("fieldName") .stream()
+ * <pre>
+ *    tm().createQueryComposer(EntityType.class)
+ *        .where("fieldName", Comparator.EQ, "value")
+ *        .orderBy("fieldName")
+ *        .stream()
+ * </pre>
  */
 public abstract class QueryComposer<T> {
 
@@ -69,12 +74,16 @@ public abstract class QueryComposer<T> {
     return this;
   }
 
+  /** Returns the first result of the query or an empty optional if there is none. */
+  public abstract Optional<T> first();
+
   /**
-   * Returns the first result of the query.
+   * Returns the one and only result of a query.
    *
-   * <p>Throws javax.persistence.NoResultException if not found.
+   * <p>Throws a {@link NonUniqueResultException} if there is more than one result, throws {@link
+   * NoResultException} if no results are found.
    */
-  public abstract T first();
+  public abstract T getSingleResult();
 
   /** Returns the results of the query as a stream. */
   public abstract Stream<T> stream();
@@ -108,11 +117,37 @@ public abstract class QueryComposer<T> {
     return criteriaBuilder::greaterThan;
   }
 
+  /**
+   * Enum used to specify comparison operations, e.g. {@code where("fieldName", Comparator.NE,
+   * "someval")'}.
+   *
+   * <p>These contain values that specify the comparison behavior for both objectify and criteria
+   * queries. For objectify, we provide a string to be appended to the field name in a {@code
+   * filter()} expression. For criteria queries we provide a function that knows how to obtain a
+   * {@link WhereOperator} from a {@link CriteriaBuilder}.
+   *
+   * <p>Note that the objectify strings for comparators other than equality are preceded by a space
+   * because {@code filter()} expects the fieldname to be separated from the operator by a space.
+   */
   public enum Comparator {
+    /**
+     * Return only records whose field is equal to the value.
+     *
+     * <p>Note that the datastore string for this is empty, which is consistent with the way {@code
+     * filter()} works (it uses an unadorned field name to check for equality).
+     */
     EQ("", QueryComposer::equal),
+
+    /** Return only records whose field is less than the value. */
     LT(" <", QueryComposer::lessThan),
+
+    /** Return only records whose field is less than or equal to the value. */
     LTE(" <=", QueryComposer::lessThanOrEqualTo),
+
+    /** Return only records whose field is greater than or equal to the value. */
     GTE(" >=", QueryComposer::greaterThanOrEqualTo),
+
+    /** Return only records whose field is greater than the value. */
     GT(" >", QueryComposer::greaterThan);
 
     private final String datastoreString;
--- a/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
+++ b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
@@ -1,0 +1,166 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.persistence.transaction;
+
+import google.registry.persistence.transaction.CriteriaQueryBuilder.WhereClause;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+import javax.annotation.Nullable;
+import javax.persistence.criteria.CriteriaBuilder;
+
+/**
+ * Creates queries that can be used both for objectify and JPA.
+ *
+ * <p>Example usage:
+ *
+ * <p>tm().createQueryComposer(EntityType.class) .where("fieldName", Comparator.EQ, "value"
+ * .orderBy("fieldName") .stream()
+ */
+public abstract class QueryComposer<T> {
+
+  // The class whose entities we're querying.  Note that this limits us to single table queries in
+  // SQL.  In datastore, there's really no other kind of query.
+  protected Class<T> entity;
+
+  // Field to order by, if any.  Null if we don't care about order.
+  @Nullable protected String orderBy;
+
+  protected List<WhereCondition<?>> predicates = new ArrayList<WhereCondition<?>>();
+
+  protected QueryComposer(Class<T> entity) {
+    this.entity = entity;
+  }
+
+  /**
+   * Introduce a "where" clause to the query.
+   *
+   * <p>Causes the query to return only results where the field and value have the relationship
+   * specified by the comparator. For example, "field EQ value", "field GT value" etc.
+   */
+  public <U extends Comparable<? super U>> QueryComposer<T> where(
+      String fieldName, Comparator comparator, U value) {
+    predicates.add(new WhereCondition(fieldName, comparator, value));
+    return this;
+  }
+
+  /** Order the query results by the value of the specified field. */
+  public QueryComposer<T> orderBy(String fieldName) {
+    orderBy = fieldName;
+    return this;
+  }
+
+  /**
+   * Returns the first result of the query.
+   *
+   * <p>Throws javax.persistence.NoResultException if not found.
+   */
+  public abstract T first();
+
+  /** Returns the results of the query as a stream. */
+  public abstract Stream<T> stream();
+
+  // We have to wrap the CriteriaQueryBuilder predicate factories in our own functions because at
+  // the point where we pass them to the Comparator constructor, the compiler can't determine which
+  // of the overloads to use since there is no "value" object for context.
+  // The only place where this context exists is in the call to the "where" method, which is why we
+  // wrap the entire "where" method instead of just the original predicate factory.
+
+  @com.google.errorprone.annotations.Immutable
+  private interface ConditionAppender {
+    <U extends Comparable<? super U>> void add(
+        CriteriaQueryBuilder queryBuilder,
+        CriteriaBuilder criteriaBuilder,
+        String fieldName,
+        U value);
+  }
+
+  private static <U extends Comparable<? super U>> void addEqualCond(
+      CriteriaQueryBuilder queryBuilder,
+      CriteriaBuilder criteriaBuilder,
+      String fieldName,
+      U value) {
+    queryBuilder.where(fieldName, criteriaBuilder::equal, value);
+  }
+
+  private static <U extends Comparable<? super U>> void addLessThanCond(
+      CriteriaQueryBuilder queryBuilder,
+      CriteriaBuilder criteriaBuilder,
+      String fieldName,
+      U value) {
+    queryBuilder.where(fieldName, (WhereClause<U>) criteriaBuilder::lessThan, value);
+  }
+
+  private static <U extends Comparable<? super U>> void addLessThanOrEqualToCond(
+      CriteriaQueryBuilder queryBuilder,
+      CriteriaBuilder criteriaBuilder,
+      String fieldName,
+      U value) {
+    queryBuilder.where(fieldName, (WhereClause<U>) criteriaBuilder::lessThanOrEqualTo, value);
+  }
+
+  private static <U extends Comparable<? super U>> void addGreaterThanOrEqualToCond(
+      CriteriaQueryBuilder queryBuilder,
+      CriteriaBuilder criteriaBuilder,
+      String fieldName,
+      U value) {
+    queryBuilder.where(fieldName, (WhereClause<U>) criteriaBuilder::greaterThanOrEqualTo, value);
+  }
+
+  private static <U extends Comparable<? super U>> void addGreaterThanCond(
+      CriteriaQueryBuilder queryBuilder,
+      CriteriaBuilder criteriaBuilder,
+      String fieldName,
+      U value) {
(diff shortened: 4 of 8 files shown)


gbrodman (changes requested):
> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 7 of 8 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 375 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstPz2-3vBMh2eB4y8:-MXstPz3EFoXvF0d8ote:b-ftg00z) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L375)):*
> ```Java
>   }
> 
>   private static class QueryComposerImpl<T> extends QueryComposer<T> {
> ```

instead of having two classes named QueryComposerImpl could we name this one DatastoreQueryComposer or DatastoreQueryComposerImpl? (same comment for SQL)
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 376 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstOYy2jcenSfdlFHc:-MXstOYy2jcenSfdlFHd:b-55iadh) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L376)):*
> ```Java
> 
>   private static class QueryComposerImpl<T> extends QueryComposer<T> {
>     QueryComposerImpl(Class<T> entity) {
> ```

nit: entityClass
___
*[core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsuPvU-Jb9x4-AXCvQ:-MXsuPvU-Jb9x4-AXCvR:b8vywiy) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java#L102)):*
> ```Java
> 
>   /** Creates a query builder for the given entity manager. */
>   public static <T> CriteriaQueryBuilder<T> create(EntityManager em, Class<T> clazz) {
> ```

yeah this might be getting into the compile-time issues you're discussing elsewhere, but given that we can retain the reference to the CriteriaBuilder in this class it'd be great if we could just have 

```
public static <U extends Comparable<? super U>> WhereClause<U> equal(CriteriaBuilder criteriaBuilder) {
  return criteriabuilder::equal;
}
```

then elsewhere we could just use 

`CriteriaQueryBuilder.create(someClass).where("myFieldName", CriteriaQueryBuilder::equal, "someValue");`
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 710 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsoeKh8r2sQhqHR4yg:-MXsoeKh8r2sQhqHR4yh:bi3by6m) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L710)):*
> ```Java
> 
>       if (orderBy != null) {
>         // TODO: Convert object field name to hibernate column names.
> ```

doesn't hibernate do this TODO for us?
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 724 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsPc3D0LuO2GRfsvqO:-MXsPc3D0LuO2GRfsvqP:b-ug13gk) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L724)):*
> ```Java
>     @Override
>     public Stream<T> stream() {
>       return buildQuery().getResultList().stream();
> ```

should be able to use getResultStream, no? 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsNfzg3cq9Ml4iy55Q:-MXsNfzg3cq9Ml4iy55R:bfa0bnh) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L39)):*
> ```Java
> 
>   // Field to order by, if any.  Null if we don't care about order.
>   @Nullable protected String orderBy;
> ```

I'm thinking that we may wish to create a simple `OrderBy` class that encapsulates the string field-to-order-by and an enum either ascending or descending? 

and then we'd just have an `Optional<OrderBy>`? 

technically one can order by multiple fields in sql but i don't think we care about that for now
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 41 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsOOya1nlKamUFUphL:-MXsOOya1nlKamUFUphM:b-1zbp08) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L41)):*
> ```Java
>   @Nullable protected String orderBy;
> 
>   protected List<WhereCondition<?>> predicates = new ArrayList<WhereCondition<?>>();
> ```

I think it'd be better to use an ImmutableList.Builder here akin to CriteriaQueryBuilder, with the logic being that this is an object that is filled up over time and used precisely once 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsNCXp5QsfZkIdpvp9:-MXsNCXp5QsfZkIdpvpA:bsbnefb) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L79)):*
> ```Java
>   // of the overloads to use since there is no "value" object for context.
>   // The only place where this context exists is in the call to the "where" method, which is why we
>   // wrap the entire "where" method instead of just the original predicate factory.
> ```

Is this comment referring to, for instance, why we have this snippet in JpaTransactionManagerImpl:

```
      for (WhereCondition pred : predicates) {
        pred.addToCriteriaQueryBuilder(builder, criteriaBuilder);
      }
```

as opposed to just  calling either CriteriaQueryBuilder::where (sql) or Query::where (datastore) initially?

It seems unfortunate / bad that each of these method references would require the CriteriaQueryBuilder, the CriteriaBuilder, the field name, and the value, when I'd hope they could just wrap the CriteriaBuilder method e.g.

```
public static <U extends Comparable<? super U>> WhereClause<U> equal(CriteriaBuilder criteriaBuilder) {
  return criteriabuilder::equal;
}
```
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 103 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsoqP-EkAv1v44uDPD:-MXsoqP-EkAv1v44uDPE:b-eo8w8c) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L103)):*
> ```Java
>       String fieldName,
>       U value) {
>     queryBuilder.where(fieldName, (WhereClause<U>) criteriaBuilder::lessThan, value);
> ```

Why are these casts necessary (this and the following methods) when there's no cast in `addEqualCond`? 
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 277 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstnzm6u9hQUqO1jo2:-MXstnzm6u9hQUqO1jo3:b5uj7ev) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L277)):*
> ```Java
> 
>   /**
>    * Returns a QueryComposer which can be used to construct portable queries for the given entity.
> ```

What is the meaning of "portable" here? We mean to say that it essentially returns something that can be treated as a builder right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 5 of 8 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 375 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstPz2-3vBMh2eB4y8:-MY5a9FZ9M8LlYHwD99Z:bjb9rm6) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L375)):*

instead of having two classes named QueryComposerImpl could we name this one DatastoreQueryComposer or DatastoreQueryComposerImpl? (same comment for SQL)
</blockquote></details>

My own preference is to rely on qualifiers to disambiguate, especially for non-public names, but I don't feel strongly about it.  Done.  
___
*[core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsuPvU-Jb9x4-AXCvQ:-MY5eI3N5zYg-HQIWQAa:bd4thjc) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java#L102)):*

yeah this might be getting into the compile-time issues you're discussing elsewhere, but given that we can retain the reference to the CriteriaBuilder in this class it'd be great if we could just have 

```
public static <U extends Comparable<? super U>> WhereClause<U> equal(CriteriaBuilder criteriaBuilder) {
  return criteriabuilder::equal;
}
```

then elsewhere we could just use 

`CriteriaQueryBuilder.create(someClass).where("myFieldName", CriteriaQueryBuilder::equal, "someValue");`
</blockquote></details>

That's funny, I thought I tried exactly that and failed.  Perhaps I had the constraint wrong initially?  In any case, it works now.  Huzzah! 

The only thing is we can't really expose the static method as the comparator: we need something that we can use with both datastore and SQL.  Comparator's job is to bundle them.

I would have also liked to get rid of addToCriteriaQueryBuilder() and just done the where clause in JpaQueryComposerImpl, but once again I'm finding myself thrashing on how to set up the generic constraints (complications are trying to express complicated bounds for an instance variable and the fact that enums cannot be generic).  I'm inclined to declare this "good enough" for now unless you can think of a good way to improve it?  
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 710 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsoeKh8r2sQhqHR4yg:-MY5fwfKCK6sBXswWG8n:b-b2omyb) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L710)):*

doesn't hibernate do this TODO for us?
</blockquote></details>

It does :-).  I forgot to remove the TODO after discovering that.  Done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 724 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsPc3D0LuO2GRfsvqO:-MY5lMuq8lYirlPlYczo:b-896fix) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L724)):*

should be able to use getResultStream, no? 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsNfzg3cq9Ml4iy55Q:-MY5lRCZ-qLQm1DBbmsl:b-lw0201) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L39)):*

I'm thinking that we may wish to create a simple `OrderBy` class that encapsulates the string field-to-order-by and an enum either ascending or descending? 

and then we'd just have an `Optional<OrderBy>`? 

technically one can order by multiple fields in sql but i don't think we care about that for now
</blockquote></details>

Yes, probably?  Although I'd prefer to hold off on that until I actually encounter a case that required "descending".

I'd rather not do Optional here, most authorities on the web seem to indicate Optional is intended for return values (frankly I'm not wild about it there, either... but ok).  It's explicitly deprecated for parameters, haven't found a position for instance variables.  
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 41 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsOOya1nlKamUFUphL:-MY6D_NU8cEcwUuzXt0Z:bmc6zpl) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L41)):*

I think it'd be better to use an ImmutableList.Builder here akin to CriteriaQueryBuilder, with the logic being that this is an object that is filled up over time and used precisely once 
</blockquote></details>

Actually I would argue the contrary for the same reason :-)

As I understand it, the builder pattern is just so we can construct immutable objects, and we use immutable objects because they are threadsafe and not subject to unexpected mutation when passed into other contexts.  What is the case for making something contained within this class immutable?

BTW, I considered making `QueryComposer` immutable, but once again, seeing as how this is something that is typically used immediately within small local contexts, I wanted to avoid the additional verbosity at the call-site.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsNCXp5QsfZkIdpvp9:-MY6Jbm8BrCIvp8u_hPi:b-gcxmb) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L79)):*

Is this comment referring to, for instance, why we have this snippet in JpaTransactionManagerImpl:

```
      for (WhereCondition pred : predicates) {
        pred.addToCriteriaQueryBuilder(builder, criteriaBuilder);
      }
```

as opposed to just  calling either CriteriaQueryBuilder::where (sql) or Query::where (datastore) initially?

It seems unfortunate / bad that each of these method references would require the CriteriaQueryBuilder, the CriteriaBuilder, the field name, and the value, when I'd hope they could just wrap the CriteriaBuilder method e.g.

```
public static <U extends Comparable<? super U>> WhereClause<U> equal(CriteriaBuilder criteriaBuilder) {
  return criteriabuilder::equal;
}
```
</blockquote></details>

Yeah, it's a little better now.  We do have to pass in the CriteriaBuilder and CriteriaQueryBuilder (see comments above).  I've updated the comment, as well.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 103 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsoqP-EkAv1v44uDPD:-MY6M6oC1PspUCNPvDyA:b9d6gkh) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L103)):*

Why are these casts necessary (this and the following methods) when there's no cast in `addEqualCond`? 
</blockquote></details>

The code is gone now, but they were added in an earlier point in trying to get the various typing concerns to line up and I just forgot to try removing them.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 277 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstnzm6u9hQUqO1jo2:-MY6MWYn6QkNqR9iYZGB:bfw29p7) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L277)):*

What is the meaning of "portable" here? We mean to say that it essentially returns something that can be treated as a builder right? 
</blockquote></details>

No, it's just badly written.  Updated with better wording.

gbrodman (changes requested):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: all files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsuPvU-Jb9x4-AXCvQ:-MY7JSY0-Q6ia1WIOY0o:b-mwxnsv) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java#L102)):*

That's funny, I thought I tried exactly that and failed.  Perhaps I had the constraint wrong initially?  In any case, it works now.  Huzzah! 

The only thing is we can't really expose the static method as the comparator: we need something that we can use with both datastore and SQL.  Comparator's job is to bundle them.

I would have also liked to get rid of addToCriteriaQueryBuilder() and just done the where clause in JpaQueryComposerImpl, but once again I'm finding myself thrashing on how to set up the generic constraints (complications are trying to express complicated bounds for an instance variable and the fact that enums cannot be generic).  I'm inclined to declare this "good enough" for now unless you can think of a good way to improve it?  
</blockquote></details>

This is definitely better. I guess my only suggestion (not sure if you've already tried this) is to move the `CriteriaBuilder` storage from `JpaQueryComposerImpl` to `CriteriaQueryBuilder`. Then CQB would take in the `comparisonFactory` object in the `where` method. That way we wouldn't need to pass around the `CriteriaBuilder` so much, I think. 

This also brings to mind the question of what we'll do when we move off Datastore, and I feel like we should hopefully be able to merge a ton of this logic with CriteriaQueryBuilder -- it kinda seems like when we're only working with SQL, this will be an unfortunately duplicate layer of abstraction. 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsNfzg3cq9Ml4iy55Q:-MY7IfguDMgfJnAqlRzn:bq4ri55) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L39)):*

Yes, probably?  Although I'd prefer to hold off on that until I actually encounter a case that required "descending".

I'd rather not do Optional here, most authorities on the web seem to indicate Optional is intended for return values (frankly I'm not wild about it there, either... but ok).  It's explicitly deprecated for parameters, haven't found a position for instance variables.  
</blockquote></details>

Either way, we can punt on OrderBy, probably should add a TODO here -- either implementing both directions of orderBy or specifying somehow that this is order-by-ascending. We might need DESC eventually for such operations as "select the newest premium list by name" which is currently done in a written-out SQL query but could very easily use this. 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 41 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsOOya1nlKamUFUphL:-MY7Ll7tDVwBggCIe1U4:bpxp5bm) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L41)):*

Actually I would argue the contrary for the same reason :-)

As I understand it, the builder pattern is just so we can construct immutable objects, and we use immutable objects because they are threadsafe and not subject to unexpected mutation when passed into other contexts.  What is the case for making something contained within this class immutable?

BTW, I considered making `QueryComposer` immutable, but once again, seeing as how this is something that is typically used immediately within small local contexts, I wanted to avoid the additional verbosity at the call-site.
</blockquote></details>

yeah there's also the logic of just using immutable collections / builders as much as possible because they're never "bad" to use and it's good to not have to think about such things 

either way, this pattern doesn't really fit into the normal usage so i'm fine with it as is 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 85 at r3](https://reviewable.io/reviews/google/nomulus/1069#-MY7I42FDk_M3pUm_71U:-MY7I42FDk_M3pUm_71V:b-31m6x8) ([raw file](https://github.com/google/nomulus/blob/e49ef6deb004851f4edd1b3013b8798485c02b97/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L85)):*
> ```Java
>   }
> 
>   public static <U extends Comparable<? super U>> WhereClause<U> lessThan(
> ```

Looking at these now, especially after you've added `WhereCondition`, it might make sense to rename the previously-existing `WhereClause` to something like `WhereOperator`? and to rename `WhereCondition` to `WhereClause`? I say that because I believe "clause" usually includes both the subject, verb, and object, and the previously-existing `WhereClause` is basically just the verb
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 113 at r3](https://reviewable.io/reviews/google/nomulus/1069#-MY7HxQi86_L4gDWcEIl:-MY7HxQi86_L4gDWcEIm:b-m5iqfp) ([raw file](https://github.com/google/nomulus/blob/e49ef6deb004851f4edd1b3013b8798485c02b97/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L113)):*
> ```Java
> 
>     private final String datastoreString;
>     final Function<CriteriaBuilder, WhereClause<?>> comparisonFactory;
> ```

very nitty, this can be private
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 113 at r3](https://reviewable.io/reviews/google/nomulus/1069#-MY7ITBm7VSPa7Namgnf:-MY7ITBm7VSPa7Namgng:bwzhsw7) ([raw file](https://github.com/google/nomulus/blob/e49ef6deb004851f4edd1b3013b8798485c02b97/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L113)):*
> ```Java
> 
>     private final String datastoreString;
>     final Function<CriteriaBuilder, WhereClause<?>> comparisonFactory;
> ```

I guess if we do the rename, this might be `operatorFactory`? 
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 277 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstnzm6u9hQUqO1jo2:-MY7HaTk5n34ztVj3D42:b-xuwjyk) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L277)):*

No, it's just badly written.  Updated with better wording.
</blockquote></details>

hm i'm not sure that change got added into git

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 3 of 8 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsuPvU-Jb9x4-AXCvQ:-MYASIwY4JxphVs_W8oh:bbzvevz) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java#L102)):*

This is definitely better. I guess my only suggestion (not sure if you've already tried this) is to move the `CriteriaBuilder` storage from `JpaQueryComposerImpl` to `CriteriaQueryBuilder`. Then CQB would take in the `comparisonFactory` object in the `where` method. That way we wouldn't need to pass around the `CriteriaBuilder` so much, I think. 

This also brings to mind the question of what we'll do when we move off Datastore, and I feel like we should hopefully be able to merge a ton of this logic with CriteriaQueryBuilder -- it kinda seems like when we're only working with SQL, this will be an unfortunately duplicate layer of abstraction. 
</blockquote></details>

It seems like CQB already accesses CriteriaBuilder through jpaTm(), and I guess it makes more sense for QueryComposer to just do the same, so done.

Re moving off datastore: I agree that we probably want to merge the two at some point.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXsNfzg3cq9Ml4iy55Q:-MYAVqP_0W3DC4HH1JQc:b235a77) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L39)):*

Either way, we can punt on OrderBy, probably should add a TODO here -- either implementing both directions of orderBy or specifying somehow that this is order-by-ascending. We might need DESC eventually for such operations as "select the newest premium list by name" which is currently done in a written-out SQL query but could very easily use this. 
</blockquote></details>

Added TODO (I didn't open a bug for it, I'm assuming we'll run into this at some point and just add it then).
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 85 at r3](https://reviewable.io/reviews/google/nomulus/1069#-MY7I42FDk_M3pUm_71U:-MYAjJPH1L5mfvcTiBWI:bohqzsr) ([raw file](https://github.com/google/nomulus/blob/e49ef6deb004851f4edd1b3013b8798485c02b97/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L85)):*

Looking at these now, especially after you've added `WhereCondition`, it might make sense to rename the previously-existing `WhereClause` to something like `WhereOperator`? and to rename `WhereCondition` to `WhereClause`? I say that because I believe "clause" usually includes both the subject, verb, and object, and the previously-existing `WhereClause` is basically just the verb
</blockquote></details>

Agreed, done.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 113 at r3](https://reviewable.io/reviews/google/nomulus/1069#-MY7HxQi86_L4gDWcEIl:-MYAkVcO2Pd81odmFxmF:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e49ef6deb004851f4edd1b3013b8798485c02b97/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L113)):*

very nitty, this can be private
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 113 at r3](https://reviewable.io/reviews/google/nomulus/1069#-MY7ITBm7VSPa7Namgnf:-MYAkWiK5jNEdEh22QBD:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e49ef6deb004851f4edd1b3013b8798485c02b97/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L113)):*

I guess if we do the rename, this might be `operatorFactory`? 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 277 at r1](https://reviewable.io/reviews/google/nomulus/1069#-MXstnzm6u9hQUqO1jo2:-MYAQhKM-m_zNkEF7z-y:b-44b29h) ([raw file](https://github.com/google/nomulus/blob/659470b24bf810ede8695308eb23f6bc6f600424/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L277)):*

hm i'm not sure that change got added into git
</blockquote></details>

Hah, sorry -- forgot to "save", so it never even made it out of my editor :-)  Done.

gbrodman (approved):
not sure if ben wants to review this, but lgtm 

> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069#-:-MYCnxee9nHR63gkkkpo:bihkbdo)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 32 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGN70o6Mcgiq2tmQch:-MYGN70o6Mcgiq2tmQci:b-fdtoyf) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L32)):*
> ```Java
>  * <p>Example usage:
>  *
>  * <p>tm().createQueryComposer(EntityType.class) .where("fieldName", Comparator.EQ, "value"
> ```

Format this correctly, e.g. https://reflectoring.io/howto-format-code-snippets-in-javadoc/

Also I think it's flat out wrong as-is; it looks like there's an extra space, a missing end parenthesis, etc.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: all files reviewed, 6 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 380 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNYCSAkylgU3jDH5X:-MYGNYCSAkylgU3jDH5Y:b-9fhlr7) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L380)):*
> ```Java
>     }
> 
>     Query<T> buildQuery() {
> ```

How are you handling all the limitations that Datastore imposes on us, namely, that you can't filter by multiple fields unless you've explicitly created a composite index for it, that you can't order by two separate fields at all, that indexes are ascending/descending-dependent, etc.?  Are you just letting Objectify fail for those and not handling it here?
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 406 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGP4zz8y7ZFjrIE8D_:-MYGP4zz8y7ZFjrIE8Da:b7f4fzx) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L406)):*
> ```Java
>     public Stream<T> stream() {
>       // TODO: there should be a better way to do this.
>       return buildQuery().list().stream();
> ```

The `Query` type that `buildQuery()` returns is itself ultimately an `Iterable` as its base interface, so I think you just want `return Streams.stream(buildQuery());`
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 708 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNRSP7nx1CCL4SDL5:-MYGN0zS3-AQiRNBz-uy:bfnrzwn) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L708)):*
> ```Java
> 
>       if (orderBy != null) {
>         queryBuilder.orderByAsc(orderBy);
> ```

What about ordering in descending order?  That's used for various things.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 77 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNznS8pUwXkNmwoN7:-MYGNznS8pUwXkNmwoN8:b6jxcz6) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L77)):*
> ```Java
>    * <p>Throws javax.persistence.NoResultException if not found.
>    */
>   public abstract T first();
> ```

I think we also want a version of this that returns the _only_ result of the query, and throws an exception if there are either 0 results or more than one.  That's a very common pattern, more common than first.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 80 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGO9zcBS6PN9tEyuyW:-MYGO9zcBS6PN9tEyuyX:b-6g0etk) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L80)):*
> ```Java
> 
>   /** Returns the results of the query as a stream. */
>   public abstract Stream<T> stream();
> ```

There's no facility to return results in some other manner, like an `ImmutableList`?  You don't always want a `Stream`.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: all files reviewed, 12 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 716 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGQ4Ui3EJiBZvjQDMp:-MYGQ4Ui3EJiBZvjQDMq:b4h8zhv) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L716)):*
> ```Java
>     @Override
>     public T first() {
>       return buildQuery().getSingleResult();
> ```

I think this may be surprisingly non-performant if it's running the entire query first rather than running the query with `LIMIT 1`.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 35 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPsQi7G85iUhXXF7t:-MYGPsQi7G85iUhXXF7u:bbdr91c) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L35)):*
> ```Java
>  * .orderBy("fieldName") .stream()
>  */
> public abstract class QueryComposer<T> {
> ```

My biggest issue with this approach overall (and why I gave up after getting part way through it) is that it's not very flexible.  It doesn't account for common use cases like `JOIN`, `LIMIT`, multiple `ORDER BY`s, `IN` or `IS NULL`/`IS NOT NULL`, etc.  By the time you implement all of that you've ended up re-implementing most of the underlying `CriteriaQueryBuilder` API.  It's a big lift.  And if we don't end up implementing all that, then we end up with a situation where you can start off with a simple query using this API, but the second you need to make it slightly more complicated (which happens all the time as codebases evolve), you need to completely rewrite it using either a SQL query or `CriteriaQueryBuilder`, potentially introducing new bugs in the process.  It would make me wary to ever use this `QueryComposer` API at all knowing that I might easily have to throw that work away and drop down to an underlying implementation if things get slightly more complicated.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 111 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPd9v-yOgJZFVFbmP:-MYGPd9v-yOgJZFVFbmQ:bmhlpx3) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L111)):*
> ```Java
>   }
> 
>   public enum Comparator {
> ```

No `IS` or `IN`?
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 111 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPoRTAz-mgoZa4vKH:-MYGPoRTAz-mgoZa4vKI:b1ierpc) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L111)):*
> ```Java
>   }
> 
>   public enum Comparator {
> ```

Needs Javadoc.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 112 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPmEJ2aKesgkzoNpV:-MYGPmEJ2aKesgkzoNpW:b-b21tso) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L112)):*
> ```Java
> 
>   public enum Comparator {
>     EQ("", QueryComposer::equal),
> ```

This is not what I was expecting for `EQ`.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 113 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPipD7UXKR_2cdmUt:-MYGPipD7UXKR_2cdmUu:b-i39tym) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L113)):*
> ```Java
>   public enum Comparator {
>     EQ("", QueryComposer::equal),
>     LT(" <", QueryComposer::lessThan),
> ```

It's a little strange that these have a leading space but not a trailing one?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 3 of 8 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 380 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNYCSAkylgU3jDH5X:-MYGQ2V69GxbDJIPjd61:b-dlkktc) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L380)):*

How are you handling all the limitations that Datastore imposes on us, namely, that you can't filter by multiple fields unless you've explicitly created a composite index for it, that you can't order by two separate fields at all, that indexes are ascending/descending-dependent, etc.?  Are you just letting Objectify fail for those and not handling it here?
</blockquote></details>

That's right.  The level of protection that we offer is no different from what ofy provides.

There may be a case for adding a little more intelligence here.  For example, if we detected two "where" clauses that are incompatible in the ofy-verse, we could conceivably demote one of them to a stream filter and this would allow us to utilize the full power of a SQL query when running in JPA mode.  But as a first pass, it seemed prudent to try not to over-engineer it.  It seems likely that if a query-then-filter works well enough in ofy, then it will work well enough in SQL.
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 406 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGP4zz8y7ZFjrIE8D_:-MYGSahK3pZ9loTQvSs4:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L406)):*

The `Query` type that `buildQuery()` returns is itself ultimately an `Iterable` as its base interface, so I think you just want `return Streams.stream(buildQuery());`
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 708 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNRSP7nx1CCL4SDL5:-MYGScPNAWSK-ufEdLcq:ba81o0m) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L708)):*

What about ordering in descending order?  That's used for various things.
</blockquote></details>

There's a TODO in the base class for that.  It's fairly easy to implement, I just wanted to wait until we ran into a real-world case.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 716 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGQ4Ui3EJiBZvjQDMp:-MYGVCMQ9vFex2waC9Cr:b-yqy02n) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L716)):*

I think this may be surprisingly non-performant if it's running the entire query first rather than running the query with `LIMIT 1`.
</blockquote></details>

`getSingleResult()` should be equivalent to `LIMIT 1` (`setMaxResults()` certainly should be, and they are both methods on the TypedQuery class).  However, in looking into this there's another potential problem: if there is more than one matching result getSingleResult() throws an exception, which doesn't mesh with the semantics of "first()".  Fixed.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 32 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGN70o6Mcgiq2tmQch:-MYGXN9u84FpdZW8k4rX:blrus2w) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L32)):*

Format this correctly, e.g. https://reflectoring.io/howto-format-code-snippets-in-javadoc/

Also I think it's flat out wrong as-is; it looks like there's an extra space, a missing end parenthesis, etc.
</blockquote></details>

Looks like the java formatter had its way with my example.  Converting to a "pre" block seems to suppress it.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 35 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPsQi7G85iUhXXF7t:-MYGXgcL16yRi4iSNy0G:bvzxphk) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L35)):*

My biggest issue with this approach overall (and why I gave up after getting part way through it) is that it's not very flexible.  It doesn't account for common use cases like `JOIN`, `LIMIT`, multiple `ORDER BY`s, `IN` or `IS NULL`/`IS NOT NULL`, etc.  By the time you implement all of that you've ended up re-implementing most of the underlying `CriteriaQueryBuilder` API.  It's a big lift.  And if we don't end up implementing all that, then we end up with a situation where you can start off with a simple query using this API, but the second you need to make it slightly more complicated (which happens all the time as codebases evolve), you need to completely rewrite it using either a SQL query or `CriteriaQueryBuilder`, potentially introducing new bugs in the process.  It would make me wary to ever use this `QueryComposer` API at all knowing that I might easily have to throw that work away and drop down to an underlying implementation if things get slightly more complicated.
</blockquote></details>

Understood.  And there's certainly likely to be cases where we'll need to just use SQL strings.  But a) it seems like we already have a lot of cases where this approach is adequate b) adding ofy/jpa branches into the code is risky (I've reviewed only one or two cases of this and found one [admittedly minor] discrepancy) and c) using a non-string query builder is a solid protection against query injection vulnerabilities, which is something that was called out in our last security review.

As far as where we want to go from here, we can make that decision going forwards.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 77 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNznS8pUwXkNmwoN7:-MYGvIPzEfjdtObv-t6S:b-fmsga3) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L77)):*

I think we also want a version of this that returns the _only_ result of the query, and throws an exception if there are either 0 results or more than one.  That's a very common pattern, more common than first.
</blockquote></details>

Turns out that's actually what `first()` did.  Added `getSingleResult()` which mirrors the semantics of the underlying call.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 80 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGO9zcBS6PN9tEyuyW:-MYGcOjsEInC7f2-wJcw:b2gj7k1) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L80)):*

There's no facility to return results in some other manner, like an `ImmutableList`?  You don't always want a `Stream`.
</blockquote></details>

That's going to be in the next PR :-)
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 111 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPd9v-yOgJZFVFbmP:-MYGcWQF9sPSU1Sra5Mz:bdys1ic) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L111)):*

No `IS` or `IN`?
</blockquote></details>

Once again, trying more for a "minimum viable product" than a full implementation.  We can extend as needed.  Binary operators are easy, unary operators will need a bit more work.

We're also currently limited by what datastore supports, though not sure whether that's relevant to IS and IN.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 111 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPoRTAz-mgoZa4vKH:-MYKhUixERYUHe0OpMbM:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L111)):*

Needs Javadoc.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 112 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPmEJ2aKesgkzoNpV:-MYKc-0d1OdOPNvsjjQF:b8na0qb) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L112)):*

This is not what I was expecting for `EQ`.
</blockquote></details>

That's a peculiarity of ofy's `filter()` method, unadorned fields indicate a check for equality.  Expanded on this in the javadoc. 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 113 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGPipD7UXKR_2cdmUt:-MYGdqvv71Uq2am2TD5m:b-qf3om5) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L113)):*

It's a little strange that these have a leading space but not a trailing one?
</blockquote></details>

Again, filter() expects <field-name> <space> <operator> for non-equality comparisons (it occurred to me that this might be more a convention than a restriction, but I've tried it... it only works with the spaces).

Also added info on this in the javadoc.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 3 of 8 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 380 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNYCSAkylgU3jDH5X:-MYL52Ew5Xl-0aDCWAYA:b-vfsi40) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L380)):*

That's right.  The level of protection that we offer is no different from what ofy provides.

There may be a case for adding a little more intelligence here.  For example, if we detected two "where" clauses that are incompatible in the ofy-verse, we could conceivably demote one of them to a stream filter and this would allow us to utilize the full power of a SQL query when running in JPA mode.  But as a first pass, it seemed prudent to try not to over-engineer it.  It seems likely that if a query-then-filter works well enough in ofy, then it will work well enough in SQL.
</blockquote></details>

I'd be opposed to silently turning a multi-field query into an in-memory table scan.  That's a giant footgun waiting to happen, and it wouldn't be clear at all from callsites what's happening (e.g. for the purpose of reviews).
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 406 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGP4zz8y7ZFjrIE8D_:-MYL5J-x7ChfsK1t1JFZ:bnxembk) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L406)):*

Done.
</blockquote></details>

You can remove the TODO unless you think there's still more improvement to be made here.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 708 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNRSP7nx1CCL4SDL5:-MYL5NLf-2TDotU6QDAW:b87nyjz) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L708)):*

There's a TODO in the base class for that.  It's fairly easy to implement, I just wanted to wait until we ran into a real-world case.
</blockquote></details>

OK, it's gonna be soon I'm guessing.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 406 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGP4zz8y7ZFjrIE8D_:-MYLDER45gOfEpC4lUaC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L406)):*

You can remove the TODO unless you think there's still more improvement to be made here.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 708 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNRSP7nx1CCL4SDL5:-MYLDFriEaD3ftvpRqsn:b-5lezgp) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L708)):*

OK, it's gonna be soon I'm guessing.
</blockquote></details>

Probably, yeah.  I'll have to if there are any instances in the codebase.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 708 at r4](https://reviewable.io/reviews/google/nomulus/1069#-MYGNRSP7nx1CCL4SDL5:-MYLdArV1o01QiEVtfnn:b-x58qwy) ([raw file](https://github.com/google/nomulus/blob/d495dd6a038cf6fc1d85b60eed11b2093bb4c33d/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L708)):*

Probably, yeah.  I'll have to if there are any instances in the codebase.
</blockquote></details>

yeah there's a couple, e.g. ListDomainsAction

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1069)* status: 3 of 8 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1070, opened 2021-04-10
Make Numulus compile on macOS

jianglai:
BSD sed behaves differently than Linux sed. By adding a "-e" flag the
comand works in both systems.

See: https://unix.stackexchange.com/questions/101059/sed-behaves-different-on-freebsd-and-on-linux

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1070)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -423,7 +423,7 @@ task jaxbToJava {
       }
     }
     execInBash(
-        'find . -name *.java -exec sed -i /\\*\\ \\<p\\>\\$/d {} +',
+        'find . -name *.java -exec sed -i -e /\\*\\ \\<p\\>\\$/d {} +',
         generatedDir)
   }
 }
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -423,7 +423,7 @@ task jaxbToJava {
       }
     }
     execInBash(
-        'find . -name *.java -exec sed -i -e /\\*\\ \\<p\\>\\$/d {} +',
+        "find . -name *.java -exec sed -i -e '/" + /\* <p>$/ + "/d' {} +",
         generatedDir)
   }
 }


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1070)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/build.gradle, line 426 at r1](https://reviewable.io/reviews/google/nomulus/1070#-MY4vFrh55q4Dn4fEIgQ:-MY4vFri4lq1whIgCKLu:b-7wstm9) ([raw file](https://github.com/google/nomulus/blob/5955879f84103198c03fb01d840166f4ecf6a4e3/core/build.gradle#L426)):*
> ```Gradle
>     }
>     execInBash(
>         'find . -name *.java -exec sed -i -e /\\*\\ \\<p\\>\\$/d {} +',
> ```

I know this is outside of the scope of this PR, but this regex is fairly incomprehensible and essentially does slash escaping at three different levels.  I think we can improve upon it as:

    "find . -name *.java -exec sed -i -e '/" + /\* <p>$/ + "/d' {} +"

This is a bit deceptive, as the slashes around the regex are actually string delimiters rather than regex delimiters (these are in the enclosing double-quoted strings), but at least in this case the the regex is _just_ the regex.

I'll leave it to your discretion as to whether you want to go there now.

jianglai:
> > Reviewed 1 of 1 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1070)_ status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)
> 
> _[core/build.gradle, line 426 at r1](https://reviewable.io/reviews/google/nomulus/1070#-MY4vFrh55q4Dn4fEIgQ:-MY4vFri4lq1whIgCKLu:b-7wstm9) ([raw file](https://github.com/google/nomulus/blob/5955879f84103198c03fb01d840166f4ecf6a4e3/core/build.gradle#L426)):_
> 
> > ```groovy-gradle
> >     }
> >     execInBash(
> >         'find . -name *.java -exec sed -i -e /\\*\\ \\<p\\>\\$/d {} +',
> > ```
> 
> I know this is outside of the scope of this PR, but this regex is fairly incomprehensible and essentially does slash escaping at three different levels. I think we can improve upon it as:
> 
> ```
> "find . -name *.java -exec sed -i -e '/" + /\* <p>$/ + "/d' {} +"
> ```
> 
> This is a bit deceptive, as the slashes around the regex are actually string delimiters rather than regex delimiters (these are in the enclosing double-quoted strings), but at least in this case the the regex is _just_ the regex.
> 
> I'll leave it to your discretion as to whether you want to go there now.

I'm curious, how does it actually work, with slashes being string delimiters?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1070)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)

___
*[core/build.gradle, line 426 at r1](https://reviewable.io/reviews/google/nomulus/1070#-MY4vFrh55q4Dn4fEIgQ:-MY5HOmt9k5Klkw8IIfR:btgjbl5) ([raw file](https://github.com/google/nomulus/blob/5955879f84103198c03fb01d840166f4ecf6a4e3/core/build.gradle#L426)):*

I'm curious, how does it actually work, with slashes being string delimiters? I tried and it works. I just don't understand why...
</blockquote></details>

It's a Groovy thing.  Slashes are string delimiters for special string literals where there is no escape character and the backslash is always literal.  It's the equivalent of a raw string in python: `r'\* <p>$'`

In the outer, double quoted strings (which are actually apparently _not_ strings, BTW, though that's not important here) we then wrap the regex in slashes for the sed expression and single quotes for bash (backslashes are also not escape characters for single-quoted bash strings).  So the expression above ends up sending the following command to bash:

    find . -name *.java -exec sed -i -e '/\* <p>$/d' {} +

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1070)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)
> 
> _[core/build.gradle, line 426 at r1](https://reviewable.io/reviews/google/nomulus/1070#-MY4vFrh55q4Dn4fEIgQ:-MY5HOmt9k5Klkw8IIfR:btgjbl5) ([raw file](https://github.com/google/nomulus/blob/5955879f84103198c03fb01d840166f4ecf6a4e3/core/build.gradle#L426)):_
> 
> It's a Groovy thing. Slashes are string delimiters for special string literals where there is no escape character and the backslash is always literal. It's the equivalent of a raw string in python: `r'\* <p>$'`
> 
> In the outer, double quoted strings (which are actually apparently _not_ strings, BTW, though that's not important here) we then wrap the regex in slashes for the sed expression and single quotes for bash (backslashes are also not escape characters for single-quoted bash strings). So the expression above ends up sending the following command to bash:
> 
> ```
> find . -name *.java -exec sed -i -e '/\* <p>$/d' {} +
> ```

Aha, the part that I missed is “it’s a Groovy thing”. I was intrigued by how you can quote strings with slashes. Sorry that you have to write the whole thing down. Should have asked the question more directly. 

https://www.baeldung.com/groovy-strings#slashy-string

PR #1075, opened 2021-04-13
Enforce consistency in non-cached FKI loads

gbrodman:
For the cached code path, we do not require consistency but we do
require the ability to load / operate on large numbers of entities (so,
we must do so without a Datastore transaction). For the non-cached code
path, we require consistency but do not care about large numbers of
entities, so we must remain in the transaction that we're already in.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1075)

### DIFF
--- a/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
+++ b/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
@@ -196,7 +196,7 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
    */
   public static <E extends EppResource> ImmutableMap<String, ForeignKeyIndex<E>> load(
       Class<E> clazz, Collection<String> foreignKeys, final DateTime now) {
-    return loadIndexesFromStore(clazz, foreignKeys).entrySet().stream()
+    return loadIndexesFromStore(clazz, foreignKeys, true).entrySet().stream()
         .filter(e -> now.isBefore(e.getValue().getDeletionTime()))
         .collect(entriesToImmutableMap());
   }
@@ -206,13 +206,22 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
    * keys, regardless of whether or not they have been soft-deleted.
    *
    * <p>Used by both the cached (w/o deletion check) and the non-cached (with deletion check) calls.
+   *
+   * <p>Note that in the cached case, we wish to run this outside of any transaction because we may
+   * be loading many entities, going over the Datastore limit on the number of enrolled entity
+   * groups per transaction (25). If we require consistency, however, we must use a transaction.
+   *
+   * @param inTransaction whether or not to use an Objectify transaction
    */
   private static <E extends EppResource>
       ImmutableMap<String, ForeignKeyIndex<E>> loadIndexesFromStore(
-          Class<E> clazz, Collection<String> foreignKeys) {
+          Class<E> clazz, Collection<String> foreignKeys, boolean inTransaction) {
     if (tm().isOfy()) {
+      Class<ForeignKeyIndex<E>> fkiClass = mapToFkiClass(clazz);
       return ImmutableMap.copyOf(
-          tm().doTransactionless(() -> ofy().load().type(mapToFkiClass(clazz)).ids(foreignKeys)));
+          inTransaction
+              ? ofy().load().type(fkiClass).ids(foreignKeys)
+              : tm().doTransactionless(() -> ofy().load().type(fkiClass).ids(foreignKeys)));
     } else {
       String property = RESOURCE_CLASS_TO_FKI_PROPERTY.get(clazz);
       ImmutableList<ForeignKeyIndex<E>> indexes =
@@ -249,7 +258,8 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
           return Optional.ofNullable(
               loadIndexesFromStore(
                       RESOURCE_CLASS_TO_FKI_CLASS.inverse().get(key.getKind()),
-                      ImmutableSet.of(foreignKey))
+                      ImmutableSet.of(foreignKey),
+                      false)
                   .get(foreignKey));
         }
 
@@ -265,7 +275,7 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
               Streams.stream(keys).map(v -> v.getSqlKey().toString()).collect(toImmutableSet());
           ImmutableSet<VKey<ForeignKeyIndex<?>>> typedKeys = ImmutableSet.copyOf(keys);
           ImmutableMap<String, ? extends ForeignKeyIndex<? extends EppResource>> existingFkis =
-              loadIndexesFromStore(resourceClass, foreignKeys);
+              loadIndexesFromStore(resourceClass, foreignKeys, false);
           // ofy() omits keys that don't have values in Datastore, so re-add them in
           // here with Optional.empty() values.
           return Maps.asMap(


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1075)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 212 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBDl3y6d7HL7kwmymM:-MYBDl3y6d7HL7kwmymN:b5u54es) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L212)):*
> ```Java
>   private static <E extends EppResource>
>       ImmutableMap<String, ForeignKeyIndex<E>> loadIndexesFromStore(
>           Class<E> clazz, Collection<String> foreignKeys, boolean forceConsistency) {
> ```

`forceConsistency` doesn't mean much to me as "consistency" as such is not a concept we use generally throughout our codebase by that name.  Just name it what it is: `isTransactional`.

Also, the Javadoc should be explaining what this parameter is, not just an inline comment.
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 214 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBECZJC0v2MLBtTUkK:-MYBECZJC0v2MLBtTUkL:b-iv98q) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L214)):*
> ```Java
>           Class<E> clazz, Collection<String> foreignKeys, boolean forceConsistency) {
>     if (tm().isOfy()) {
>       // If loading from the cache, we may need to load more entities than the Datastore transaction
> ```

Be more specific.  "Datastore transaction size limit" (which is 10 MB) is not what we're actually hitting here -- what we're hitting here is the Datastore limit on number of enrolled entity groups in a transaction, which is 25.
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 219 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBEbi77ylxQnc_N8Ko:-MYBEbi77ylxQnc_N8Kp:bdvpv2i) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L219)):*
> ```Java
>       return ImmutableMap.copyOf(
>           forceConsistency
>               ? ofy().load().type(mapToFkiClass(clazz)).ids(foreignKeys)
> ```

Why is this an `ofy()` call and not `ofyTm()`?  Does it matter?

mindhog:
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 219 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBEbi77ylxQnc_N8Ko:-MYBFiSo9u2dEnNrXVSu:b-p10o5c) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L219)):*

Why is this an `ofy()` call and not `ofyTm()`?  Does it matter?
</blockquote></details>

ofyTm() doeesn't currently support arbitrary ofy queries (IMO, we should make it do so just so even if it's the very same API, just so we can rule out stuff that has already been converted)

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1075)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1075)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 212 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBDl3y6d7HL7kwmymM:-MYBRROWDYE64YGXmUMk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L212)):*

`forceConsistency` doesn't mean much to me as "consistency" as such is not a concept we use generally throughout our codebase by that name.  Just name it what it is: `isTransactional`.

Also, the Javadoc should be explaining what this parameter is, not just an inline comment.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 214 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBECZJC0v2MLBtTUkK:-MYBRWUT90m6zi9Zdr0q:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L214)):*

Be more specific.  "Datastore transaction size limit" (which is 10 MB) is not what we're actually hitting here -- what we're hitting here is the Datastore limit on number of enrolled entity groups in a transaction, which is 25.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 219 at r1](https://reviewable.io/reviews/google/nomulus/1075#-MYBEbi77ylxQnc_N8Ko:-MYBRYoW1HFpQmRjCEzM:bayrnc6) ([raw file](https://github.com/google/nomulus/blob/c4d522e1ce6d06a52848c371da2337e6a8a828a9/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L219)):*

ofyTm() doeesn't currently support arbitrary ofy queries (IMO, we should make it do so just so even if it's the very same API, just so we can rule out stuff that has already been converted) 
</blockquote></details>

we actually could use ofyTm here but it would require converting the collection of Strings to VKeys then converting the resulting map's keys from VKeys back into Strings.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1075)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 228 at r2](https://reviewable.io/reviews/google/nomulus/1075#-MYBSR8mDlmva4Km9myH:-MYBSR8nBu54fnfECx-2:b-hn16vg) ([raw file](https://github.com/google/nomulus/blob/5467f95f7f4c5781d523c1b71cc93579deb3ae93/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L228)):*
> ```Java
>       String property = RESOURCE_CLASS_TO_FKI_PROPERTY.get(clazz);
>       ImmutableList<ForeignKeyIndex<E>> indexes =
>           tm().transact(
> ```

It occurs to me that this explicit call to `transact()` shouldn't be necessary, right?  We're always in a wrapping transaction?  This could be replaced by an `assertInTransaction()` call.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1075)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 228 at r2](https://reviewable.io/reviews/google/nomulus/1075#-MYBSR8mDlmva4Km9myH:-MYBdl9y2H3vNdMMB5ld:b-kk81bq) ([raw file](https://github.com/google/nomulus/blob/5467f95f7f4c5781d523c1b71cc93579deb3ae93/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L228)):*

It occurs to me that this explicit call to `transact()` shouldn't be necessary, right?  We're always in a wrapping transaction?  This could be replaced by an `assertInTransaction()` call.
</blockquote></details>

I'm not sure. Either way, we can check that later and not have it block this commit (the transact call is a no-op if we're already in one)

PR #1090, opened 2021-04-16
Begin saving the EppResource parent in *History objects

gbrodman:
We use DomainCreateFlow as an example here of how this will work. There
were a few changes necessary:

- various changes around GracePeriod / GracePeriodHistory so that we can
actually store them without throwing NPEs
- Creating one injectable *History.Builder field and using in place of
the HistoryEntry.Builder injected field in DomainCreateFlow
- Saving the EppResource as the parent in the *History.Builder setParent
calls
- Converting to/from HistoryEntry/*History classes in
DatastoreTransactionManager. Basically, we'll want to return the
*History subclasses (and similar in the ofy portions of HistoryEntryDao)
- Converting a few HistoryEntry.Builder usages to DomainHistory.Builder
usages. Eventually we should convert all of them.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1090)

### DIFF
--- a/core/src/main/java/google/registry/flows/FlowModule.java
+++ b/core/src/main/java/google/registry/flows/FlowModule.java
@@ -20,6 +20,7 @@ import com.google.common.base.Strings;
 import dagger.Module;
 import dagger.Provides;
 import google.registry.flows.picker.FlowPicker;
+import google.registry.model.domain.DomainHistory;
 import google.registry.model.domain.metadata.MetadataExtension;
 import google.registry.model.eppcommon.AuthInfo;
 import google.registry.model.eppcommon.Trid;
@@ -237,6 +238,12 @@ public class FlowModule {
           .setRequestedByRegistrar(metadataExtension.get().getRequestedByRegistrar());
     }
     return historyBuilder;
+  }
+
+  @Provides
+  static DomainHistory.Builder provideDomainHistoryBuilder(
+      HistoryEntry.Builder historyEntryBuilder) {
+    return new DomainHistory.Builder().copyFrom(historyEntryBuilder.build());
   }
 
   /**
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -80,6 +80,7 @@ import google.registry.model.billing.BillingEvent.Recurring;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.domain.DomainCommand;
 import google.registry.model.domain.DomainCommand.Create;
+import google.registry.model.domain.DomainHistory;
 import google.registry.model.domain.GracePeriod;
 import google.registry.model.domain.Period;
 import google.registry.model.domain.fee.FeeCreateCommandExtension;
@@ -206,7 +207,7 @@ public class DomainCreateFlow implements TransactionalFlow {
   @Inject @ClientId String clientId;
   @Inject @TargetId String targetId;
   @Inject @Superuser boolean isSuperuser;
-  @Inject HistoryEntry.Builder historyBuilder;
+  @Inject DomainHistory.Builder historyBuilder;
   @Inject EppResponse.Builder responseBuilder;
   @Inject AllocationTokenFlowUtils allocationTokenFlowUtils;
   @Inject DomainCreateFlowCustomLogic flowCustomLogic;
@@ -303,8 +304,8 @@ public class DomainCreateFlow implements TransactionalFlow {
         validateSecDnsExtension(eppInput.getSingleExtension(SecDnsCreateExtension.class));
     String repoId = createDomainRepoId(ObjectifyService.allocateId(), registry.getTldStr());
     DateTime registrationExpirationTime = leapSafeAddYears(now, years);
-    HistoryEntry historyEntry = buildHistoryEntry(
-        repoId, registry, now, period, registry.getAddGracePeriodLength());
+    DomainHistory domainHistory =
+        buildHistoryEntry(repoId, registry, now, period, registry.getAddGracePeriodLength());
     // Bill for the create.
     BillingEvent.OneTime createBillingEvent =
         createOneTimeBillingEvent(
@@ -314,20 +315,16 @@ public class DomainCreateFlow implements TransactionalFlow {
             isReserved(domainName, isSunriseCreate),
             years,
             feesAndCredits,
-            historyEntry,
+            domainHistory,
             allocationToken,
             now);
     // Create a new autorenew billing event and poll message starting at the expiration time.
     BillingEvent.Recurring autorenewBillingEvent =
-        createAutorenewBillingEvent(historyEntry, registrationExpirationTime);
+        createAutorenewBillingEvent(domainHistory, registrationExpirationTime);
     PollMessage.Autorenew autorenewPollMessage =
-        createAutorenewPollMessage(historyEntry, registrationExpirationTime);
+        createAutorenewPollMessage(domainHistory, registrationExpirationTime);
     ImmutableSet.Builder<ImmutableObject> entitiesToSave = new ImmutableSet.Builder<>();
-    entitiesToSave.add(
-        historyEntry,
-        createBillingEvent,
-        autorenewBillingEvent,
-        autorenewPollMessage);
+    entitiesToSave.add(createBillingEvent, autorenewBillingEvent, autorenewPollMessage);
     // Bill for EAP cost, if any.
     if (!feesAndCredits.getEapCost().isZero()) {
       entitiesToSave.add(createEapBillingEvent(feesAndCredits, createBillingEvent));
@@ -337,7 +334,7 @@ public class DomainCreateFlow implements TransactionalFlow {
     if (getReservationTypes(domainName).contains(NAME_COLLISION)) {
       statuses.add(SERVER_HOLD);
       entitiesToSave.add(
-          createNameCollisionOneTimePollMessage(targetId, historyEntry, clientId, now));
+          createNameCollisionOneTimePollMessage(targetId, domainHistory, clientId, now));
     }
 
     DomainBase newDomain =
@@ -365,13 +362,13 @@ public class DomainCreateFlow implements TransactionalFlow {
             .build();
     entitiesToSave.add(
         newDomain,
+        domainHistory.asBuilder().setDomainContent(newDomain).build(),
         ForeignKeyIndex.create(newDomain, newDomain.getDeletionTime()),
         EppResourceIndex.create(Key.create(newDomain)));
     if (allocationToken.isPresent()
         && TokenType.SINGLE_USE.equals(allocationToken.get().getTokenType())) {
       entitiesToSave.add(
-          allocationTokenFlowUtils.redeemToken(
-              allocationToken.get(), HistoryEntry.createVKey(Key.create(historyEntry))));
+          allocationTokenFlowUtils.redeemToken(allocationToken.get(), domainHistory.createVKey()));
     }
     enqueueTasks(newDomain, hasSignedMarks, hasClaimsNotice);
 
@@ -379,7 +376,7 @@ public class DomainCreateFlow implements TransactionalFlow {
         flowCustomLogic.beforeSave(
             DomainCreateFlowCustomLogic.BeforeSaveParameters.newBuilder()
                 .setNewDomain(newDomain)
-                .setHistoryEntry(historyEntry)
+                .setHistoryEntry(domainHistory)
                 .setEntityChanges(
                     EntityChanges.newBuilder().setSaves(entitiesToSave.build()).build())
                 .setYears(years)
@@ -483,7 +480,7 @@ public class DomainCreateFlow implements TransactionalFlow {
             : null);
   }
 
-  private HistoryEntry buildHistoryEntry(
+  private DomainHistory buildHistoryEntry(
       String repoId, Registry registry, DateTime now, Period period, Duration addGracePeriod) {
     // We ignore prober transactions
     if (registry.getTldType() == TldType.REAL) {
@@ -511,7 +508,7 @@ public class DomainCreateFlow implements TransactionalFlow {
       boolean isReserved,
       int years,
       FeesAndCredits feesAndCredits,
-      HistoryEntry historyEntry,
+      DomainHistory domainHistory,
       Optional<AllocationToken> allocationToken,
       DateTime now) {
     ImmutableSet.Builder<Flag> flagsBuilder = new ImmutableSet.Builder<>();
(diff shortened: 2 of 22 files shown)


mindhog (changes requested):
> Reviewed 22 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjLnjuDMcXGLcSq1Pv:-MYjLnjuDMcXGLcSq1Pw:b-nsb498) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/domain/DomainHistory.java#L132)):*
> ```Java
>         updatable = false)
>   })
>   Set<DomainDsDataHistory> dsDataHistories = new HashSet<>();
> ```

What's  up with the HashSets?  Is this to appease Hibernate?  Please add a comment if so.
___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjMPuUCKhA4OMwT0Wt:-MYjMPuUCKhA4OMwT0Wu:b-hewwi7) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/domain/GracePeriod.java#L198)):*
> ```Java
>     @Access(AccessType.PROPERTY)
>     @Column(nullable = false)
>     public Long getGracePeriodId() {
> ```

Why do we need Long for a non-nullable id? 
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 140 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjNK6Q23pj4WC_9kpA:-MYjNK6Q23pj4WC_9kpB:bqrrj7d) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L140)):*
> ```Java
>   public void putAll(Object... entities) {
>     syncIfTransactionless(
>         getOfy().save().entities(toHistoryEntriesIfNecessary(ImmutableList.copyOf(entities))));
> ```

This transformation seems like it belongs at the SqlEntity/DatastoreEntity level (although I'm not sure which of them or whether it is necessarily appropriate to either).

In any case, we shouldn't have any entity specific code in TransactionManager -- please rename this to something like "translateToDatastoreEntities"
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 359 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjPAhW8TlB8cb61_Bv:-MYjPAhW8TlB8cb61_Bw:b-2bt8re) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L359)):*
> ```Java
>   private static Object toHistoryEntryIfNecessary(@Nullable Object obj) {
>     if (obj instanceof HistoryEntry) {
>       return ((HistoryEntry) obj).asHistoryEntry();
> ```

Yeah, we should thread this transformation through DatastoreEntity -- or even a separate interface like "NotQuiteADatastoreEntity," or something. It's just a more conventional OOP pattern.  I realize this is the only case where we need this, but having specific entity names in this otherwise very general class is just ugly.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090)* status: all files reviewed, 8 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java, line 379 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjehkPCtIBYbC_Z759:-MYjehkPCtIBYbC_Z75A:b-kwd3ys) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L379)):*
> ```Java
>             DomainCreateFlowCustomLogic.BeforeSaveParameters.newBuilder()
>                 .setNewDomain(newDomain)
>                 .setHistoryEntry(domainHistory)
> ```

I guess we'll want to rename `setHistoryEntry()` and other related methods at some point during the migration (maybe just during the post-migration cleanup phase).
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 199 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjf6Gd0NWggPDViGyD:-MYjf6Gd0NWggPDViGyE:bibj1di) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L199)):*
> ```Java
> 
>     @Override
>     public Builder setParent(@Nullable EppResource eppResource) {
> ```

This should be typed as`ContactResource`/`ContactBase`.  It wouldn't be valid to pass in a domain or host here, yet this typing allows it.  I see you have a `checkArgument` below but this should be enforceable through the type system.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 200 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjeujz3pLrYOQtfTWL:-MYjeujz3pLrYOQtfTWM:b-q5tbt9) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L200)):*
> ```Java
>     @Override
>     public Builder setParent(@Nullable EppResource eppResource) {
>       if (eppResource == null) {
> ```

Shouldn't this be a `checkArgument()` call?  There's a DB invariant that history entries always be associated with an EPP resource.  Is it because we're a little loose with this invariant in some tests?
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 348 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjfiIWBsk3SUn62rm3:-MYjfiIWBsk3SUn62rm4:b-vxlxrr) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/domain/DomainHistory.java#L348)):*
> ```Java
> 
>     @Override
>     public Builder setParent(@Nullable EppResource eppResource) {
> ```

The two previous comments apply here and to hosts also.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java, line 379 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjehkPCtIBYbC_Z759:-MYk7xGT1T1xzMEFthWz:b-o56glo) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L379)):*

I guess we'll want to rename `setHistoryEntry()` and other related methods at some point during the migration (maybe just during the post-migration cleanup phase).
</blockquote></details>

yeah for sure
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 199 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjf6Gd0NWggPDViGyD:-MYk3Og_1ZOCd9zKcGr1:b45crrp) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L199)):*

This should be typed as`ContactResource`/`ContactBase`.  It wouldn't be valid to pass in a domain or host here, yet this typing allows it.  I see you have a `checkArgument` below but this should be enforceable through the type system.
</blockquote></details>

The benefit that I was going for with this is that this allows us to override the superclass (`HistoryEntry.Builder`) method `setParent`, whereas the compiler gets mad if you have an override with a more specific parameter type (e.g. superclass has an EppResource param and the subclass has a DomainContent param) 


Eh, it's probably not a problem to have the methods have different names, and just call the superclass's `setParent` in the `setDomainContent` method of DomainHistory's builder (and similarly for the others) 
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 200 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjeujz3pLrYOQtfTWL:-MYk4nO41mj-gdl_qD0E:b-5o891v) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L200)):*

Shouldn't this be a `checkArgument()` call?  There's a DB invariant that history entries always be associated with an EPP resource.  Is it because we're a little loose with this invariant in some tests?
</blockquote></details>

yeah it's because we're loose with the invariant due to the fact that old migrated-over entities won't be populated 
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjLnjuDMcXGLcSq1Pv:-MYk9m-JApfcb52nhTy6:b-77m83x) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/domain/DomainHistory.java#L132)):*

What's  up with the HashSets?  Is this to appease Hibernate?  Please add a comment if so.
</blockquote></details>

Yep. Basically for these collections Hibernate will often start with the blank placeholder then fill them out once you actually load and inspect the object. You can't fill out an ImmutableSet.

Comment added.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 348 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjfiIWBsk3SUn62rm3:-MYk5gEVBURO0yDpLTAw:bh4p0wp) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/domain/DomainHistory.java#L348)):*

The two previous comments apply here and to hosts also.
</blockquote></details>

yep same 
___
*[core/src/main/java/google/registry/model/domain/GracePeriod.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjMPuUCKhA4OMwT0Wt:-MYk6BB2A5r6GNuY46nP:bi26mg3) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/domain/GracePeriod.java#L198)):*

Why do we need Long for a non-nullable id? 
</blockquote></details>

When loading embedded entities from the database, Hibernate will often (?) have a blank, non-filled out entity used as a placeholder 

tbh i think we can just revert everything to using `long`s here (some were using `Long` before) and then we should be all set 
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 140 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjNK6Q23pj4WC_9kpA:-MYkARrN-3PSPHb9WdAZ:b-zd62fp) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L140)):*

This transformation seems like it belongs at the SqlEntity/DatastoreEntity level (although I'm not sure which of them or whether it is necessarily appropriate to either).

In any case, we shouldn't have any entity specific code in TransactionManager -- please rename this to something like "translateToDatastoreEntities"
</blockquote></details>

yeah we can do something like that, it means that there will be some unnecessary method calls but that's not the worst thing in the world 
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 359 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjPAhW8TlB8cb61_Bv:-MYkDEQ30nM2emv5b0lv:bqrnrq4) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L359)):*

Yeah, we should thread this transformation through DatastoreEntity -- or even a separate interface like "NotQuiteADatastoreEntity," or something. It's just a more conventional OOP pattern.  I realize this is the only case where we need this, but having specific entity names in this otherwise very general class is just ugly.
</blockquote></details>

yeah, i think we should be able to do that through fairly generic conversion methods here 

agreed that it looks ugly even if it is only applicable to this one case

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090)* status: 6 of 27 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 199 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjf6Gd0NWggPDViGyD:-MYl4Clc1AoRLFFxQaiP:bcat91i) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L199)):*

The benefit that I was going for with this is that this allows us to override the superclass (`HistoryEntry.Builder`) method `setParent`, whereas the compiler gets mad if you have an override with a more specific parameter type (e.g. superclass has an EppResource param and the subclass has a DomainContent param) 


Eh, it's probably not a problem to have the methods have different names, and just call the superclass's `setParent` in the `setDomainContent` method of DomainHistory's builder (and similarly for the others) 
</blockquote></details>

It would also be possible to do this using the type system.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 200 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjeujz3pLrYOQtfTWL:-MYl47YaBrGSwkgXDiif:b4krftj) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L200)):*

yeah it's because we're loose with the invariant due to the fact that old migrated-over entities won't be populated 
</blockquote></details>

Right.  Might want to document that.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090#-:-MYl4Rzx4JWMkFc2DZpW:bnfp4nl)* status: 6 of 27 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @mindhog)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090)* status: 6 of 27 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 200 at r1](https://reviewable.io/reviews/google/nomulus/1090#-MYjeujz3pLrYOQtfTWL:-MYl5kCc298Jp9qzmx8X:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d3e894bdba8f21fbcbf22413264134e6efb228f5/core/src/main/java/google/registry/model/contact/ContactHistory.java#L200)):*

Right.  Might want to document that.
</blockquote></details>

Done.

mindhog (approved):
> Reviewed 21 of 21 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1090)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @mindhog)

PR #1091, opened 2021-04-16
Convert GenerateLordnCommand to tm

mindhog:
This makes use of QueryComposer and adds a `list()` method to it.

Since there was no test for GenerateLordnCommand, this also implements one.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1091)

### DIFF
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -444,5 +444,10 @@ public class DatastoreTransactionManager implements TransactionManager {
     public long count() {
       return buildQuery().count();
     }
+
+    @Override
+    public List<T> list() {
+      return buildQuery().list();
+    }
   }
 }
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -737,5 +737,10 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
       CriteriaQueryBuilder<Long> queryBuilder = CriteriaQueryBuilder.createCount(em, entityClass);
       return addCriteria(queryBuilder).getSingleResult();
     }
+
+    @Override
+    public List<T> list() {
+      return buildQuery().getResultList();
+    }
   }
 }
--- a/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
+++ b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
@@ -91,6 +91,9 @@ public abstract class QueryComposer<T> {
   /** Returns the number of results of the query. */
   public abstract long count();
 
+  /** Returns the results of the query as a list. */
+  public abstract List<T> list();
+
   // We have to wrap the CriteriaQueryBuilder predicate factories in our own functions because at
   // the point where we pass them to the Comparator constructor, the compiler can't determine which
   // of the overloads to use since there is no "value" object for context.
--- a/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
@@ -14,7 +14,9 @@
 
 package google.registry.tools;
 
-import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.persistence.transaction.QueryComposer.Comparator;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.joda.time.DateTimeZone.UTC;
 
@@ -53,12 +55,20 @@ final class GenerateLordnCommand implements CommandWithRemoteApi {
       required = true)
   private Path sunriseOutputPath;
 
+  public GenerateLordnCommand() {} // CommandTestCase wants a constructor.
+
   @Override
   public void run() throws IOException {
     DateTime now = DateTime.now(UTC);
     ImmutableList.Builder<String> claimsCsv = new ImmutableList.Builder<>();
     ImmutableList.Builder<String> sunriseCsv = new ImmutableList.Builder<>();
-    for (DomainBase domain : ofy().load().type(DomainBase.class).filter("tld", tld)) {
+    for (DomainBase domain :
+        transactIfJpaTm(
+            () ->
+                tm().createQueryComposer(DomainBase.class)
+                    .where("tld", Comparator.EQ, tld)
+                    .orderBy("repoId")
+                    .list())) {
       String status = " ";
       if (domain.getLaunchNotice() == null && domain.getSmdId() != null) {
         sunriseCsv.add(LordnTaskUtils.getCsvLineForSunriseDomain(domain, domain.getCreationTime()));
--- a/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
@@ -18,7 +18,6 @@ import static google.registry.persistence.transaction.QueryComposer.Comparator;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.joda.time.DateTimeZone.UTC;
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
@@ -26,9 +25,11 @@ import com.google.common.collect.ImmutableList;
 import google.registry.model.domain.DomainBase;
 import google.registry.tmch.LordnTaskUtils;
 import google.registry.tools.params.PathParameter;
+import google.registry.util.Clock;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import javax.inject.Inject;
 import org.joda.time.DateTime;
 
 /** Command to generate a LORDN CSV file for an entire TLD. */
@@ -55,11 +56,11 @@ final class GenerateLordnCommand implements CommandWithRemoteApi {
       required = true)
   private Path sunriseOutputPath;
 
-  public GenerateLordnCommand() {} // CommandTestCase wants a constructor.
+  @Inject Clock clock;
 
   @Override
   public void run() throws IOException {
-    DateTime now = DateTime.now(UTC);
+    DateTime now = clock.nowUtc();
     ImmutableList.Builder<String> claimsCsv = new ImmutableList.Builder<>();
     ImmutableList.Builder<String> sunriseCsv = new ImmutableList.Builder<>();
     for (DomainBase domain :
--- a/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
@@ -65,11 +65,12 @@ final class GenerateLordnCommand implements CommandWithRemoteApi {
     ImmutableList.Builder<String> sunriseCsv = new ImmutableList.Builder<>();
     transactIfJpaTm(
         () ->
-              tm().createQueryComposer(DomainBase.class)
-                  .where("tld", Comparator.EQ, tld)
-                  .orderBy("repoId")
-                  .stream()
-                  .forEach(domain -> processDomain(claimsCsv, sunriseCsv, domain)));
+            tm()
+                .createQueryComposer(DomainBase.class)
+                .where("tld", Comparator.EQ, tld)
+                .orderBy("repoId")
+                .stream()
+                .forEach(domain -> processDomain(claimsCsv, sunriseCsv, domain)));
     ImmutableList<String> claimsRows = claimsCsv.build();
     ImmutableList<String> claimsAll =
         new ImmutableList.Builder<String>()
@@ -88,9 +89,10 @@ final class GenerateLordnCommand implements CommandWithRemoteApi {
     Files.write(sunriseOutputPath, sunriseAll, UTF_8);
   }
 
-  private static void processDomain(ImmutableList.Builder<String> claimsCsv,
-                             ImmutableList.Builder<String> sunriseCsv,
-                             DomainBase domain) {
+  private static void processDomain(
+      ImmutableList.Builder<String> claimsCsv,
+      ImmutableList.Builder<String> sunriseCsv,
+      DomainBase domain) {
     String status = " ";
     if (domain.getLaunchNotice() == null && domain.getSmdId() != null) {
       sunriseCsv.add(LordnTaskUtils.getCsvLineForSunriseDomain(domain, domain.getCreationTime()));
--- a/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateLordnCommand.java
@@ -63,23 +63,13 @@ final class GenerateLordnCommand implements CommandWithRemoteApi {
     DateTime now = clock.nowUtc();
     ImmutableList.Builder<String> claimsCsv = new ImmutableList.Builder<>();
     ImmutableList.Builder<String> sunriseCsv = new ImmutableList.Builder<>();
-    for (DomainBase domain :
-        transactIfJpaTm(
-            () ->
-                tm().createQueryComposer(DomainBase.class)
-                    .where("tld", Comparator.EQ, tld)
-                    .orderBy("repoId")
-                    .list())) {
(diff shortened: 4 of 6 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekEBp3hzPAvFrC98y:-MYekEBp3hzPAvFrC98z:b-6fmne) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L92)):*
> ```Java
> 
>   /** Returns the results of the query as a list. */
>   public abstract List<T> list();
> ```

This should be returning an `ImmutableList`, right?  Otherwise it's giving the impression that mutating the returned list might actually do something?
___
*[core/src/main/java/google/registry/tools/GenerateLordnCommand.java, line 58 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekjGrAvdAuweS7Eic:-MYekjGrAvdAuweS7Eid:b-fem0zu) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/tools/GenerateLordnCommand.java#L58)):*
> ```Java
>   private Path sunriseOutputPath;
> 
>   public GenerateLordnCommand() {} // CommandTestCase wants a constructor.
> ```

I don't see any other tool classes that have this so I don't see why it's necessary.
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekq5IBI1ql6-Z-HPg:-MYekq5IBI1ql6-Z-HPh:biwvyop) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L39)):*
> ```Java
> public class GenerateLordnCommandTest extends CommandTestCase<GenerateLordnCommand> {
> 
>   public GenerateLordnCommandTest() {}
> ```

Why is this necessary?
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekv8DAQcNLkNx5X1A:-MYekv8DAQcNLkNx5X1B:bqqxm0u) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L42)):*
> ```Java
> 
>   @TestOfyAndSql
>   public void testExample() throws Exception {
> ```

Tests don't need to be public; package-private is fine.
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 53 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYel5yO0oLdPE7Wpq7w:-MYel5yO0oLdPE7Wpq7x:b-z1t75j) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L53)):*
> ```Java
>             .build());
>     DateTime now = DateTime.parse("2021-04-16T10:04:00.000Z");
>     try (MockedStatic<DateTime> dateTime = mockStatic(DateTime.class);
> ```

This is an atypical way to handle timing in our tests.  Prefer the use of an `@Inject`ed `Clock` in the command, then just set its value here.  That's how we handle this in dozens of other cases.
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYelb3bAfTgPEdRDcTq:-MYelb3bAfTgPEdRDcTr:btchppr) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L54)):*
> ```Java
>     DateTime now = DateTime.parse("2021-04-16T10:04:00.000Z");
>     try (MockedStatic<DateTime> dateTime = mockStatic(DateTime.class);
>         MockedStatic<Files> files = mockStatic(Files.class)) {
> ```

Use `@TempDir` for this instead of static mocking.  That's how we handle this in other tests and it's overall much more capable.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekEBp3hzPAvFrC98y:-MYffi_vBz3Iv8kZmdNC:b-d47lie) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L92)):*

This should be returning an `ImmutableList`, right?  Otherwise it's giving the impression that mutating the returned list might actually do something?
</blockquote></details>

It's an abstract function so that's really up to the implementation, and JPA's getResultList() returns List (possibly also immutable, but "protected" at runtime in the same way that ImmutableList is).  I'd hate to incur the O(n) copy just to get a nominal indicator of immutability. 
___
*[core/src/main/java/google/registry/tools/GenerateLordnCommand.java, line 58 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekjGrAvdAuweS7Eic:-MYfjRKu5yxneQ2TgU9f:b77uzl5) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/tools/GenerateLordnCommand.java#L58)):*

I don't see any other tool classes that have this so I don't see why it's necessary.
</blockquote></details>

I added this and the constructor below [in order to fix an error from the code that attempts to instantiate the command](https://github.com/google/nomulus/blob/master/core/src/test/java/google/registry/tools/CommandTestCase.java#L252).  I'm not sure why it was failing then, but removing these doesn't seem to break it now so... done.
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 39 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekq5IBI1ql6-Z-HPg:-MYfnAwI7uXmFGL3XKiH:b-4e2l8u) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L39)):*

Why is this necessary?
</blockquote></details>

Removed (see above).
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekv8DAQcNLkNx5X1A:-MYfnYIf7rEozhoC6Z56:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L42)):*

Tests don't need to be public; package-private is fine.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 53 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYel5yO0oLdPE7Wpq7w:-MYfos1X2fthAlOsz6fT:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L53)):*

This is an atypical way to handle timing in our tests.  Prefer the use of an `@Inject`ed `Clock` in the command, then just set its value here.  That's how we handle this in dozens of other cases.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYelb3bAfTgPEdRDcTq:-MYfwNNM1twWtGKXrPtV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/test/java/google/registry/tools/GenerateLordnCommandTest.java#L54)):*

Use `@TempDir` for this instead of static mocking.  That's how we handle this in other tests and it's overall much more capable.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekEBp3hzPAvFrC98y:-MYjTXAl2BbfO8LWFIns:b-3byfro) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L92)):*

It's an abstract function so that's really up to the implementation, and JPA's getResultList() returns List (possibly also immutable, but "protected" at runtime in the same way that ImmutableList is).  I'd hate to incur the O(n) copy just to get a nominal indicator of immutability. 
</blockquote></details>

It's not an O(n) copy; see here: https://github.com/google/guava/wiki/ImmutableCollectionsExplained#copyof-is-smarter-than-you-think

Also see:

http://go/java-practices/choosing-collections#immutable

https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableCollection.html

"For field types and method return types, you should generally use the immutable type (such as ImmutableList) instead of the general collection interface type (such as List). This communicates to your callers all of the semantic guarantees listed above, which is almost always very useful information."

JPA doesn't use Guava so it doesn't return `ImmutableList`, but we do use Guava so we should use it here.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/tools/GenerateLordnCommand.java, line 58 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekjGrAvdAuweS7Eic:-MYjUgskCZIFYAYQ1paq:b-o96vqa) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/tools/GenerateLordnCommand.java#L58)):*

I added this and the constructor below [in order to fix an error from the code that attempts to instantiate the command](https://github.com/google/nomulus/blob/master/core/src/test/java/google/registry/tools/CommandTestCase.java#L252).  I'm not sure why it was failing then, but removing these doesn't seem to break it now so... done.
</blockquote></details>

I think there's some Java shenanigans going on.  Java by default makes a zero arg constructor, but if you add your own constructor with args (which you presumably did here) then that default zero arg constructor, which apparently we're relying on, no longer exists.  So if you add one constructor to a `Command` then you need to add both.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 92 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekEBp3hzPAvFrC98y:-MYjrt7T43u1k24uCAnm:b-nqzjyb) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L92)):*

It's not an O(n) copy; see here: https://github.com/google/guava/wiki/ImmutableCollectionsExplained#copyof-is-smarter-than-you-think

Also see:

http://go/java-practices/choosing-collections#immutable

https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableCollection.html

"For field types and method return types, you should generally use the immutable type (such as ImmutableList) instead of the general collection interface type (such as List). This communicates to your callers all of the semantic guarantees listed above, which is almost always very useful information."

JPA doesn't use Guava so it doesn't return `ImmutableList`, but we do use Guava so we should use it here.
</blockquote></details>

Actually, I'm pretty sure it is an O(n) copy in this case.  `copyOf()` doesn't _guarantee_ a constant time implementation, it just tries to provide one when possible (and doesn't care to document those cases).  However, from the code it appears that [`copyOf()` only does this optimization for `ImmutableCollection`s](https://github.com/google/guava/blob/6e054cea7bc54dc0bea90a9fc27a21224cfcdee1/guava/src/com/google/common/collect/ImmutableList.java#L256). In all other cases, it calls `toArray()` which makes a copy of the list.

As it turns out, getResultList() returns an ArrayList.  So we would be doing an additional copy to get that ImmutableList return type.  Certainly the performance here is dominated by the database call itself, but I'd still count that unnecessary copy as a pretty good reason to just forego the ImmutableList, in this case.
___
*[core/src/main/java/google/registry/tools/GenerateLordnCommand.java, line 58 at r1](https://reviewable.io/reviews/google/nomulus/1091#-MYekjGrAvdAuweS7Eic:-MYjwCiACPkIi9v-ATKE:b-3x0g9r) ([raw file](https://github.com/google/nomulus/blob/b1e550b048b6d102847108d4891e3148393d1f91/core/src/main/java/google/registry/tools/GenerateLordnCommand.java#L58)):*

I think there's some Java shenanigans going on.  Java by default makes a zero arg constructor, but if you add your own constructor with args (which you presumably did here) then that default zero arg constructor, which apparently we're relying on, no longer exists.  So if you add one constructor to a `Command` then you need to add both.
</blockquote></details>

I don't think I did (add another constructor).  I think something else must have been tripping things up, but I'm not sure what.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/tools/GenerateLordnCommand.java, line 66 at r2](https://reviewable.io/reviews/google/nomulus/1091#-MYpVF5P2p7shZjXbrX8:-MYpVF5P2p7shZjXbrX9:b-ttsjs2) ([raw file](https://github.com/google/nomulus/blob/0f331d68b89b7a6de0e8d1ff2e1ba74d02613ac5/core/src/main/java/google/registry/tools/GenerateLordnCommand.java#L66)):*
> ```Java
>     ImmutableList.Builder<String> claimsCsv = new ImmutableList.Builder<>();
>     ImmutableList.Builder<String> sunriseCsv = new ImmutableList.Builder<>();
>     for (DomainBase domain :
> ```

This might be a very large list (all domains in a TLD) so it may be better to use the stream API instead of the list API

gbrodman (commented):
> Reviewed 3 of 5 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/tools/GenerateLordnCommand.java, line 66 at r2](https://reviewable.io/reviews/google/nomulus/1091#-MYpVF5P2p7shZjXbrX8:-MYpYlSt7R4Yju9XpvL1:b-mjs94q) ([raw file](https://github.com/google/nomulus/blob/0f331d68b89b7a6de0e8d1ff2e1ba74d02613ac5/core/src/main/java/google/registry/tools/GenerateLordnCommand.java#L66)):*

This might be a very large list (all domains in a TLD) so it may be better to use the stream API instead of the list API 
</blockquote></details>

We just finished talking about a case where second-guessing the original semantics caused breakage, so I have to say I'm a bit reluctant...  But yeah, you're right.  We should stream this.

Note that this kind of breaks the association between this and the rest of the PR, say the word and I'll split the change out.

gbrodman (approved):
> Reviewed 5 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1091)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1096, opened 2021-04-19
Send an immediate poll message for superuser domain deletes

CydeWeys:
This poll message is in addition to the normal poll message that is sent when
the domain's deletion is effective (typically 35 days later). It's needed
because, in the event of a superuser deletion, the owning registrar won't
otherwise necessarily know it's happening.

Note that, in the case of a --immediate superuser deletion, the normal poll
message is already being sent immediately, so this additional poll message is
not necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1096)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java
@@ -208,11 +208,19 @@ public final class DomainDeleteFlow implements TransactionalFlow {
     // Enqueue the deletion poll message if the delete is asynchronous or if requested by a
     // superuser (i.e. the registrar didn't request this delete and thus should be notified even if
     // it is synchronous).
-    if (!durationUntilDelete.equals(Duration.ZERO) || isSuperuser) {
+    if (durationUntilDelete.isLongerThan(Duration.ZERO) || isSuperuser) {
       PollMessage.OneTime deletePollMessage =
           createDeletePollMessage(existingDomain, historyEntry, deletionTime);
       entitiesToSave.add(deletePollMessage);
       builder.setDeletePollMessage(deletePollMessage.createVKey());
+    }
+
+    // Send a second poll message immediately if the domain is being deleted asynchronously by a
+    // registrar other than the sponsoring registrar (which will necessarily be a superuser).
+    if (durationUntilDelete.isLongerThan(Duration.ZERO)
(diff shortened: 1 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java, line 222 at r1](https://reviewable.io/reviews/google/nomulus/1096#-MYg_TKCERzxZaYjcG3T:-MYg_TKCERzxZaYjcG3U:b-4cyueo) ([raw file](https://github.com/google/nomulus/blob/05e75b13647c0d6a0a1bf1d6d791cd3a30d12028/core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java#L222)):*
> ```Java
>     // registrar other than the sponsoring registrar (which will necessarily be a superuser).
>     if (durationUntilDelete.isLongerThan(Duration.ZERO)
>         && !clientId.equals(existingDomain.getPersistedCurrentSponsorClientId())) {
> ```

Why not use the existing `isSuperuser` check instead of checking the registrar ID?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1096)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java, line 222 at r1](https://reviewable.io/reviews/google/nomulus/1096#-MYg_TKCERzxZaYjcG3T:-MYjSufY4KGjTipes1a_:beug262) ([raw file](https://github.com/google/nomulus/blob/05e75b13647c0d6a0a1bf1d6d791cd3a30d12028/core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java#L222)):*

Why not use the existing `isSuperuser` check instead of checking the registrar ID? 
</blockquote></details>

Because `isSuperuser` isn't the same thing.  We also run superuser flows on our own self-allocated registry domains, and in those cases, we don't need to send ourselves a poll message for an action we just did.  The poll message is only for the specific case of when the action is being performed by someone other than the owning registrar.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1096)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1097, opened 2021-04-20
Set default value of 1 for new not-null columns

gbrodman:
Use 1 since it's the constant singleton ID 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1097)


CydeWeys (approved):
It occurs to me that, since these are singletons, the ID column isn't even strictly necessary, rather, just `SELECT * LIMIT 1` would do the trick, but I get that this helps ensure compatibility with Datastore for now.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1097#-:-MYkijTa9G_ljiNk78Ui:b-74znbj)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1099, opened 2021-04-21
Use CommandWithRemoteApi in SetDatabaseTransitionScheduleCommand

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1099)

### DIFF
--- a/core/src/main/java/google/registry/tools/ConfirmingCommand.java
+++ b/core/src/main/java/google/registry/tools/ConfirmingCommand.java
@@ -20,7 +20,7 @@ import com.beust.jcommander.Parameter;
 import com.google.common.base.Strings;
 
 /** A {@link Command} that implements a confirmation step before executing. */
-public abstract class ConfirmingCommand implements Command {
+public abstract class ConfirmingCommand implements CommandWithRemoteApi {
 
   @Parameter(
       names = {"-f", "--force"},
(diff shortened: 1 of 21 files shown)


gbrodman (changes requested):
I think we should clarify two things in the commit message

1. the reasoning for this change (basically that it removes a footgun from things that would theoretically otherwise need a remote API where that need isn't revealed until we try to run the command after deployment 
2. that there are only two commands that exist that extend ConfirmingCommand that don't already implement CommandWithRemoteApi -- ImportDatastoreCommand (used when restoring from backup, not really relevant) and CreateCdnsTld command. The only real downside here is that we'd instantiate the connection perhaps unnecessarily for CreateCdnsTld command but I don't think we're particularly concerned about speed there. 


Also adding Ben for more context -- reason number 1 is the reason for this change, basically that it's possible to get something that works to change things in tests but not in production after deployment 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1099#-:-MYufi0FDtsnhUTaizcs:b-a7qj5q)* status: 0 of 2 files reviewed, all discussions resolved

gbrodman:
welp, disregard all that

gbrodman (approved):

gbrodman:
yeah seems reasonable, I'm good with it either way

gbrodman (commented):
actually let's change the PR message please

gbrodman (approved):

PR #1073, opened 2021-04-13
Migrate Spec11 pipeline to flex template

jianglai:
Unfortunately this PR has turned out to be much bigger than I initially
conceived. However this is no good way to separate it out because the
changes are intertwined. This PR includes 3 main changes:

1. Change the spec11 pipline to use Dataflow Flex Template.
2. Retire the use of the old JPA layer that relies on credential saved
   in KMS.
3. Some extensive refactoring to streamline the logic and improve test
   isolation.

TESTED=deployed to alpha and verified that the pipeline is launched and finished successfully.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1073)

PR #1100, opened 2021-04-21
Migrate the billing pipeline to flex template

jianglai:
This is similar to the migration of the spec11 pipeline in #1073. Also removed
a few Dagger providers that are no longer needed.

TESTED=tested the dataflow job on alpha.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1100)


jianglai (commented):
> Reviewed 24 of 26 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1100)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (changes requested):
> Reviewed 14 of 26 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1100)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/beam/BeamActionTestBase.java, line 50 at r2](https://reviewable.io/reviews/google/nomulus/1100#-MYrDAfn-120DuVz-A6L:-MYrDAfn-120DuVz-A6M:bvqdz4n) ([raw file](https://github.com/google/nomulus/blob/7c4a984958a135e44566d1614fe784d9a8ac66f8/core/src/test/java/google/registry/beam/BeamActionTestBase.java#L50)):*
> ```Java
>   @BeforeEach
>   void beforeEach() throws Exception {
>     when(dataflow.projects()).thenReturn(projects);
> ```

So lenient() is not needed anymore? Do you know why? At least one failure case in WipeOutDatastoreActionTest does not touch the mocks.

jianglai (commented):
> Reviewed 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1100)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[core/src/test/java/google/registry/beam/BeamActionTestBase.java, line 50 at r2](https://reviewable.io/reviews/google/nomulus/1100#-MYrDAfn-120DuVz-A6L:-MYrgLDk3mFKZi0pI5wx:bpwyp47) ([raw file](https://github.com/google/nomulus/blob/7c4a984958a135e44566d1614fe784d9a8ac66f8/core/src/test/java/google/registry/beam/BeamActionTestBase.java#L50)):*

So lenient() is not needed anymore? Do you know why? At least one failure case in WipeOutDatastoreActionTest does not touch the mocks. 
</blockquote></details>

It's unclear to me as well. The annotation is inherited so it should apply to child classes as well. But seeing that strict stub is better anyway, I think we can leave it like this until it becomes a problem?
___
*[core/src/test/java/google/registry/beam/invoicing/InvoicingPipelineTest.java, line 210 at r3](https://reviewable.io/reviews/google/nomulus/1100#-MYriJ590GrJoJ6L_BkO:-MYriJ590GrJoJ6L_BkP:b9ldbxv) ([raw file](https://github.com/google/nomulus/blob/74136a6b91a03cbb1cd465cc80799fb9be6e4e1f/core/src/test/java/google/registry/beam/invoicing/InvoicingPipelineTest.java#L210)):*
> ```
>       assertThat(detailReport.subList(1, detailReport.size()))
>           .containsExactlyElementsIn(entry.getValue());
> ```
This test is flaky and I cannot figure out why. The actual report would sometimes contain fewer lines then expected. I think there might be a race somewhere but it's not obvious to me where. The pipeline runs till it finishes so presumably IO should be all done by the time the assertion is made. If I print out `detailReport` just before the assertion it usually will succeed.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1100)* status: 25 of 26 files reviewed, all discussions resolved (waiting on @jianglai and @weiminyu)

PR #1101, opened 2021-04-21
Convert more tests to use @DualDatabaseTest and SQL in general

gbrodman:
Nothing super crazy here other than persisting the entity changes in
DomainDeleteFlow at the end of the flow rather than almost at the end.
This means that when we return the results we give the results as they
were originally present, rather than the subsequently-changed values.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1101)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainDeleteFlow.java
@@ -262,7 +262,6 @@ public final class DomainDeleteFlow implements TransactionalFlow {
             .setHistoryEntry(historyEntry)
             .setEntityChanges(EntityChanges.newBuilder().setSaves(entitiesToSave.build()).build())
             .build());
-    persistEntityChanges(entityChanges);
     BeforeResponseReturnData responseData =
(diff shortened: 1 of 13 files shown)


CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1101)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/flows/EppLifecycleDomainTest.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/1101#-MYvKSeYBbsf84-GM0Ux:-MYvKSeZAp4ybsPWaRw2:bxauonx) ([raw file](https://github.com/google/nomulus/blob/aa2894776476c471024c8882db9a4e19f22e32fd/core/src/test/java/google/registry/flows/EppLifecycleDomainTest.java#L70)):*
> ```Java
>       AppEngineExtension.builder()
>           .withDatastoreAndCloudSql()
>           .withClock(clock)
> ```

At some point we might want to devolve this extension into the base `EppTestCase` class, alongside its friend `clock`.

PR #1105, opened 2021-04-22
Fix Spec11 domain check

mindhog:
We should be checking to see if there are _any_ active domains for a given
reported domain, not to see if _the_ domain for the name is active.

The last change caused an exception for domains with soft-deleted past domains
of the same name.  The original code only checked the first domain returned
from the query, which may have been soft-deleted.  This version checks all
domain records to see if any are active.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1105)

### DIFF
--- a/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java
+++ b/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java
@@ -141,8 +141,9 @@ public class Spec11EmailUtils {
                                   "fullyQualifiedDomainName",
                                   Comparator.EQ,
                                   threatMatch.fullyQualifiedDomainName())
-                              .getSingleResult()
-                              .shouldPublishToDns())
+                              .stream()
+                              .filter(domain -> domain.shouldPublishToDns())
+                              .count() > 0)
                   .collect(toImmutableList());
             });
     return RegistrarThreatMatches.create(registrarThreatMatches.clientId(), filteredMatches);
--- a/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java
+++ b/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java
@@ -136,14 +136,14 @@ public class Spec11EmailUtils {
               return registrarThreatMatches.threatMatches().stream()
                   .filter(
                       threatMatch ->
-                          tm().createQueryComposer(DomainBase.class)
+                          tm()
+                              .createQueryComposer(DomainBase.class)
                               .where(
                                   "fullyQualifiedDomainName",
                                   Comparator.EQ,
                                   threatMatch.fullyQualifiedDomainName())
                               .stream()
-                              .filter(domain -> domain.shouldPublishToDns())
-                              .count() > 0)
+                              .anyMatch(DomainBase::shouldPublishToDns))
                   .collect(toImmutableList());
             });
     return RegistrarThreatMatches.create(registrarThreatMatches.clientId(), filteredMatches);
--- a/core/src/test/java/google/registry/reporting/spec11/Spec11EmailUtilsTest.java
+++ b/core/src/test/java/google/registry/reporting/spec11/Spec11EmailUtilsTest.java
@@ -280,7 +280,6 @@ class Spec11EmailUtilsTest {
         Optional.empty());
   }
 
-
   @TestOfyAndSql
   void testOneFailure_sendsAlert() throws Exception {
     // If there is one failure, we should still send the other message and then an alert email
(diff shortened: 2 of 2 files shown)


mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1105)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java, line 146 at r1](https://reviewable.io/reviews/google/nomulus/1105#-MYuoPxn0tPbJ1ttsSAl:-MYurDVy02_S22_tI2pd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6dc50a19782578f87bea0bbd1b2cf3c4987d2069/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java#L146)):*

> ```
>                               .filter(domain -> domain.shouldPublishToDns())
>                               .count() > 0)
> ```
how about anyMatch(DomainBase::shouldPublishDns)
</blockquote></details>

Done.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1105)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @CydeWeys and @weiminyu)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1105)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mindhog and @weiminyu)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java, line 146 at r2](https://reviewable.io/reviews/google/nomulus/1105#-MYvIrXL0vOoqZoI7X6h:-MYvIrXL0vOoqZoI7X6i:bm424k) ([raw file](https://github.com/google/nomulus/blob/b2833139f9bb7a2249306118d58f1112056de2a8/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java#L146)):*
> ```Java
>                                   threatMatch.fullyQualifiedDomainName())
>                               .stream()
>                               .anyMatch(DomainBase::shouldPublishToDns))
> ```

Looking at the definition of `shouldPublishToDns()`, I don't think that's what you want here?  It seems to me that you want `EppResourceUtils.isActive()` ?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1105)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java, line 146 at r2](https://reviewable.io/reviews/google/nomulus/1105#-MYvIrXL0vOoqZoI7X6h:-MYyVCMB878dyd8-M0vE:b-klg34j) ([raw file](https://github.com/google/nomulus/blob/b2833139f9bb7a2249306118d58f1112056de2a8/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java#L146)):*

Looking at the definition of `shouldPublishToDns()`, I don't think that's what you want here?  It seems to me that you want `EppResourceUtils.isActive()` ?
</blockquote></details>

I was curious about that initially, too.  I'm not entirely clear on the original intent, but it seems like shouldPublishToDns() returns true if the domain is not inactive, on client or server hold or in pending delete.

So I guess the question is: do we want to publish abuse reports for domains that are already deleted or on hold as well as for domains that are simply active?

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1105)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java, line 146 at r2](https://reviewable.io/reviews/google/nomulus/1105#-MYvIrXL0vOoqZoI7X6h:-MYzw-d220y_jehGdVWA:b-72mrj) ([raw file](https://github.com/google/nomulus/blob/b2833139f9bb7a2249306118d58f1112056de2a8/core/src/main/java/google/registry/reporting/spec11/Spec11EmailUtils.java#L146)):*

I was curious about that initially, too.  I'm not entirely clear on the original intent, but it seems like shouldPublishToDns() returns true if the domain is not inactive, on client or server hold or in pending delete.

So I guess the question is: do we want to publish abuse reports for domains that are already deleted or on hold as well as for domains that are simply active?
</blockquote></details>

Never mind, just realized that's how it already was anyway -- let's leave it as is for this (unrelated) fix.

PR #1106, opened 2021-04-23
Populate the domain in DomainHistory objects created in Domain flows

gbrodman:
Unfortunately, much of the time there's a bit of a circular dependency
in the object creation, e.g. the Domain object stores references to the
billing events which store references to the history object which
contains the Domain object. As a result, we sort of need to create the
history object twice.

In addition, we add a utility copyFrom method in HistoryEntry.Builder to
avoid unnecessary ID allocations.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1106)

### DIFF
--- a/core/src/main/java/google/registry/flows/FlowModule.java
+++ b/core/src/main/java/google/registry/flows/FlowModule.java
@@ -243,7 +243,7 @@ public class FlowModule {
   @Provides
   static DomainHistory.Builder provideDomainHistoryBuilder(
       HistoryEntry.Builder historyEntryBuilder) {
-    return new DomainHistory.Builder().copyFrom(historyEntryBuilder.build());
+    return new DomainHistory.Builder().copyFrom(historyEntryBuilder);
   }
 
   /**
--- a/core/src/main/java/google/registry/flows/FlowUtils.java
+++ b/core/src/main/java/google/registry/flows/FlowUtils.java
@@ -22,15 +22,19 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.common.base.Throwables;
 import com.google.common.flogger.FluentLogger;
+import com.googlecode.objectify.Key;
 import google.registry.flows.EppException.CommandUseErrorException;
 import google.registry.flows.EppException.ParameterValueRangeErrorException;
 import google.registry.flows.EppException.SyntaxErrorException;
 import google.registry.flows.EppException.UnimplementedProtocolVersionException;
 import google.registry.flows.custom.EntityChanges;
+import google.registry.model.EppResource;
 import google.registry.model.eppcommon.EppXmlTransformer;
 import google.registry.model.eppinput.EppInput.WrongProtocolVersionException;
 import google.registry.model.eppoutput.EppOutput;
 import google.registry.model.host.InetAddressAdapter.IpVersionMismatchException;
+import google.registry.model.ofy.ObjectifyService;
+import google.registry.model.reporting.HistoryEntry;
 import google.registry.model.translators.CurrencyUnitAdapter.UnknownCurrencyException;
 import google.registry.xml.XmlException;
 import java.util.List;
@@ -97,6 +101,11 @@ public final class FlowUtils {
         throw new RuntimeException(e2); // Failing to marshal at all is not recoverable.
       }
     }
+  }
+
+  public static <H extends HistoryEntry> Key<H> createHistoryKey(
+      EppResource parent, Class<H> clazz) {
+    return Key.create(Key.create(parent), clazz, ObjectifyService.allocateId());
   }
 
   /** Registrar is not logged in. */
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -97,7 +97,6 @@ import google.registry.model.eppinput.EppInput;
 import google.registry.model.eppinput.ResourceCommand;
 import google.registry.model.eppoutput.CreateData.DomainCreateData;
 import google.registry.model.eppoutput.EppResponse;
-import google.registry.model.host.HostResource;
 import google.registry.model.index.EppResourceIndex;
 import google.registry.model.index.ForeignKeyIndex;
 import google.registry.model.ofy.ObjectifyService;
@@ -107,11 +106,11 @@ import google.registry.model.poll.PollMessage.Autorenew;
 import google.registry.model.registry.Registry;
 import google.registry.model.registry.Registry.TldState;
 import google.registry.model.registry.Registry.TldType;
+import google.registry.model.registry.label.ReservationType;
 import google.registry.model.reporting.DomainTransactionRecord;
 import google.registry.model.reporting.DomainTransactionRecord.TransactionReportField;
 import google.registry.model.reporting.HistoryEntry;
 import google.registry.model.reporting.IcannReportingTypes.ActivityReportField;
-import google.registry.persistence.VKey;
 import google.registry.tmch.LordnTaskUtils;
 import java.util.Optional;
 import javax.inject.Inject;
@@ -302,10 +301,14 @@ public class DomainCreateFlow implements TransactionalFlow {
     validateFeeChallenge(targetId, now, feeCreate, feesAndCredits);
     Optional<SecDnsCreateExtension> secDnsCreate =
         validateSecDnsExtension(eppInput.getSingleExtension(SecDnsCreateExtension.class));
-    String repoId = createDomainRepoId(ObjectifyService.allocateId(), registry.getTldStr());
     DateTime registrationExpirationTime = leapSafeAddYears(now, years);
-    DomainHistory domainHistory =
-        buildHistoryEntry(repoId, registry, now, period, registry.getAddGracePeriodLength());
+    String repoId = createDomainRepoId(ObjectifyService.allocateId(), registry.getTldStr());
+    Key<DomainHistory> domainHistoryKey =
+        Key.create(
+            Key.create(DomainBase.class, repoId),
+            DomainHistory.class,
+            ObjectifyService.allocateId());
+    historyBuilder.setId(domainHistoryKey.getId());
     // Bill for the create.
     BillingEvent.OneTime createBillingEvent =
         createOneTimeBillingEvent(
@@ -315,14 +318,14 @@ public class DomainCreateFlow implements TransactionalFlow {
             isReserved(domainName, isSunriseCreate),
             years,
             feesAndCredits,
-            domainHistory,
+            domainHistoryKey,
             allocationToken,
             now);
     // Create a new autorenew billing event and poll message starting at the expiration time.
     BillingEvent.Recurring autorenewBillingEvent =
-        createAutorenewBillingEvent(domainHistory, registrationExpirationTime);
+        createAutorenewBillingEvent(domainHistoryKey, registrationExpirationTime);
     PollMessage.Autorenew autorenewPollMessage =
-        createAutorenewPollMessage(domainHistory, registrationExpirationTime);
+        createAutorenewPollMessage(domainHistoryKey, registrationExpirationTime);
     ImmutableSet.Builder<ImmutableObject> entitiesToSave = new ImmutableSet.Builder<>();
     entitiesToSave.add(createBillingEvent, autorenewBillingEvent, autorenewPollMessage);
     // Bill for EAP cost, if any.
@@ -330,14 +333,12 @@ public class DomainCreateFlow implements TransactionalFlow {
       entitiesToSave.add(createEapBillingEvent(feesAndCredits, createBillingEvent));
     }
 
-    ImmutableSet.Builder<StatusValue> statuses = new ImmutableSet.Builder<>();
-    if (getReservationTypes(domainName).contains(NAME_COLLISION)) {
-      statuses.add(SERVER_HOLD);
-      entitiesToSave.add(
-          createNameCollisionOneTimePollMessage(targetId, domainHistory, clientId, now));
-    }
-
-    DomainBase newDomain =
+    ImmutableSet<ReservationType> reservationTypes = getReservationTypes(domainName);
+    ImmutableSet<StatusValue> statuses =
+        reservationTypes.contains(NAME_COLLISION)
+            ? ImmutableSet.of(SERVER_HOLD)
+            : ImmutableSet.of();
+    DomainBase domain =
         new DomainBase.Builder()
             .setCreationClientId(clientId)
             .setPersistedCurrentSponsorClientId(clientId)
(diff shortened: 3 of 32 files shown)


mindhog (changes requested):
Optional: have you considered using a `Builder<DomainHistory>` instead of the `DomainHistory` object itself for the initial history object?  It's a fairly minor point, but it seems like this could save a few extra copies, as well as the associated `toBuilder()...build()` dance.

> Reviewed 26 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106#-:-MZKDmnd6Cs9VH3SxiAH:bl45nj6)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r1](https://reviewable.io/reviews/google/nomulus/1106#-MZKB4-Y-8Gd4eEFHVoU:-MZKB4-Z7pFaMBrB5gCC:b-u4xxvw) ([raw file](https://github.com/google/nomulus/blob/59defe8aa2b09373532929026063ad4c80f1bc60/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*
> ```Java
>       postLoad();
>     }
>     hashCode = null; // reset the hash code since we may have changed the entities
> ```

Do we have evidence that the hash code has been created at this time?  `restoreOfyKeys()` should only be called during object construction, prior to the creation of the hash code.
___
*[core/src/test/java/google/registry/flows/ResourceFlowTestCase.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/1106#-MZKCvjdBVIpRZeYbJOQ:-MZKCvjdBVIpRZeYbJOR:br40uud) ([raw file](https://github.com/google/nomulus/blob/59defe8aa2b09373532929026063ad4c80f1bc60/core/src/test/java/google/registry/flows/ResourceFlowTestCase.java#L198)):*
> ```Java
>         assertAboutImmutableObjects()
>             .that(contactHistory.getContactBase().get())
>             .isEqualExceptFields(resource);
> ```

This can simply be isEqual(), right?  Same below.

gbrodman (commented):
No, we cannot unfortunately. The reason is fundamental to the circular dependency, that we must know the DomainHistory's `id` value when we create the billing events et al. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106#-:-MZKYkuj7vzflJFSJvCs:bn1grea)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r1](https://reviewable.io/reviews/google/nomulus/1106#-MZKB4-Y-8Gd4eEFHVoU:-MZKU0lf7DE-RKnYuC22:b-sy3b70) ([raw file](https://github.com/google/nomulus/blob/59defe8aa2b09373532929026063ad4c80f1bc60/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

Do we have evidence that the hash code has been created at this time?  `restoreOfyKeys()` should only be called during object construction, prior to the creation of the hash code.
</blockquote></details>

Yes we do, it was causing test failures because while the fields were equal in two TransferData objects stored inside their respective two DomainContent objects, the hash codes were not. (so it's an issue with nested objects) 
___
*[core/src/test/java/google/registry/flows/ResourceFlowTestCase.java, line 198 at r1](https://reviewable.io/reviews/google/nomulus/1106#-MZKCvjdBVIpRZeYbJOQ:-MZKUzWn4TqxZhs5qq4Q:bdmbjis) ([raw file](https://github.com/google/nomulus/blob/59defe8aa2b09373532929026063ad4c80f1bc60/core/src/test/java/google/registry/flows/ResourceFlowTestCase.java#L198)):*

This can simply be isEqual(), right?  Same below.
</blockquote></details>

yep, I was thinking that we might need additional fields removed from consideration for host or contact objects but it appears not to be the case (i've worked through the other flows locally)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

mindhog (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106#-:-MZNE0wM1hpirifT5fds:bb6ywmx)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

CydeWeys (changes requested):
> The reason is fundamental to the circular dependency, that we must know the DomainHistory's id value when we create the billing events et al.

That's what `allocateId` is for.  We had this circular dependency problem in Datastore too, and we solved it there.  See e.g.: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java;l=305;drc=95fdd36c7751915efca672dff5a1fb7d3f12c9ce;bpv=1;bpt=1

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106#-:-MZTIUbD8PMzn_Qq7FUI:b-pka41h)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZTJ0p6A34WuAmDTmwa:b-uviizs) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*
> ```Java
>       postLoad();
>     }
>     hashCode = null; // reset the hash code since we may have changed the entities
> ```

This seems hackish.  Why null it out instead of recalculating it?  Isn't there a risk of two entities both having a null hash code accidentally being considered equal?  Why is this line of code even here?

gbrodman (commented):
Done, didn't know about manually allocating the ID. It required some refactoring overall but it's fairly straightforward 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106#-:-MZZS44i1V-OyUvit6kl:bpga77d)* status: 18 of 31 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZZQt4c2YfDuVD_UHYq:b-e2djla) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

This seems hackish.  Why null it out instead of recalculating it?  Isn't there a risk of two entities both having a null hash code accidentally being considered equal?  Why is this line of code even here?
</blockquote></details>

If you look in `ImmutableBuilder`'s `equals` method, it compares the hash codes using the hash code getter, which recalculates the hash codes if they're null. The point in nulling it out here is that there's no need to recompute it immediately, maybe it won't ever be necessary -- so just say "this should be recalculated if we ever need it" by nulling it. 

The line of code should be explained by the comment -- the hash code is computed at some point before this assuming immutability, and we are mutating the object. Thus, that previously-calculated hash code is no longer value.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: 18 of 31 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZmEgzP5-LlEA5f7ZFB:b-cavzh1) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

If you look in `ImmutableBuilder`'s `equals` method, it compares the hash codes using the hash code getter, which recalculates the hash codes if they're null. The point in nulling it out here is that there's no need to recompute it immediately, maybe it won't ever be necessary -- so just say "this should be recalculated if we ever need it" by nulling it. 

The line of code should be explained by the comment -- the hash code is computed at some point before this assuming immutability, and we are mutating the object. Thus, that previously-calculated hash code is no longer value. 
</blockquote></details>

As I consider this further, I'm a bit more concerned about it.  Can we find out where the hash code was previously calculated?  I've seen Hibernate do this in the past when adding objects to internal collections, but I didn't think that happened until after the post-loads for an object were called.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: 18 of 31 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZmgPIA5E7_UHQbeZ15:buu0x90) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

As I consider this further, I'm a bit more concerned about it.  Can we find out where the hash code was previously calculated?  I've seen Hibernate do this in the past when adding objects to internal collections, but I didn't think that happened until after the post-loads for an object were called. 
</blockquote></details>

The first instance that I see of the hash code being calculated in an arbitrarily-selected failing test is https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/domain/DomainContent.java#L800 

But where that's called is somewhat irrelevant, since it could theoretically be called from pretty much anywhere.

mindhog (changes requested):
> Reviewed 13 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java, line 311 at r3](https://reviewable.io/reviews/google/nomulus/1106#-MZmsZDbCKHZ0rqN-TsJ:-MZmsZDc49EQYjNULuyO:b-ry8czt) ([raw file](https://github.com/google/nomulus/blob/7dec1301ddab69ed02e91b45b401d972f833381f/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L311)):*
> ```Java
>             DomainHistory.class,
>             ObjectifyService.allocateId());
>     historyBuilder.setId(domainHistoryKey.getId());
> ```

The ofy keys, they make me sad :-(
___
*[core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java, line 489 at r3](https://reviewable.io/reviews/google/nomulus/1106#-MZmlZiN8cbiT6El-ou7:-MZmlZiN8cbiT6El-ou8:b-2e3m8f) ([raw file](https://github.com/google/nomulus/blob/7dec1301ddab69ed02e91b45b401d972f833381f/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L489)):*
> ```Java
> 
>   private DomainHistory buildDomainHistory(
>       DomainBase newDomain,
> ```

Nit: is this always a new domain?
___
*[core/src/main/java/google/registry/flows/domain/DomainTransferCancelFlow.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1106#-MZmq9-A6dl5P9j3Nwy4:-MZmq9-A6dl5P9j3Nwy5:b-uxo23b) ([raw file](https://github.com/google/nomulus/blob/7dec1301ddab69ed02e91b45b401d972f833381f/core/src/main/java/google/registry/flows/domain/DomainTransferCancelFlow.java#L102)):*
> ```Java
> 
>     Key<DomainHistory> domainHistoryKey =
>         Key.create(Key.create(existingDomain), DomainHistory.class, ObjectifyService.allocateId());
> ```

It looks like you're doing this key creation in a number of places.  Can we put in a static method in DomainHistory, maybe?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java, line 311 at r3](https://reviewable.io/reviews/google/nomulus/1106#-MZmsZDbCKHZ0rqN-TsJ:-MZnqsbl1T7ZzR-UJH17:b-c6q0y0) ([raw file](https://github.com/google/nomulus/blob/7dec1301ddab69ed02e91b45b401d972f833381f/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L311)):*

The ofy keys, they make me sad :-(
</blockquote></details>

yeah :/ 

they're the best way here to preserve that parent relationship as long as it's needed in Datastore I think
___
*[core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java, line 489 at r3](https://reviewable.io/reviews/google/nomulus/1106#-MZmlZiN8cbiT6El-ou7:-MZnqs7I5D5l-YfsAgcf:bqyot8r) ([raw file](https://github.com/google/nomulus/blob/7dec1301ddab69ed02e91b45b401d972f833381f/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L489)):*

Nit: is this always a new domain?
</blockquote></details>

it is indeed
___
*[core/src/main/java/google/registry/flows/domain/DomainTransferCancelFlow.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1106#-MZmq9-A6dl5P9j3Nwy4:-MZns4_E0jdann2623BN:bpy3c5t) ([raw file](https://github.com/google/nomulus/blob/7dec1301ddab69ed02e91b45b401d972f833381f/core/src/main/java/google/registry/flows/domain/DomainTransferCancelFlow.java#L102)):*

It looks like you're doing this key creation in a number of places.  Can we put in a static method in DomainHistory, maybe?
</blockquote></details>

Yes, I think it makes the most sense in FlowUtils (we should be able to reuse it for the contact and host flows as well). Done, thanks.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: 23 of 32 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZoDE5b0Kn9-aKJtPA7:b1b5z85) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

The first instance that I see of the hash code being calculated in an arbitrarily-selected failing test is https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/domain/DomainContent.java#L800 

But where that's called is somewhat irrelevant, since it could theoretically be called from pretty much anywhere. 
</blockquote></details>

It seems safest to me to recalculate the hash code here rather than nulling it out.
___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 275 at r4](https://reviewable.io/reviews/google/nomulus/1106#-MZoDN7E2NXl9qBr8fJk:-MZoDN7E2NXl9qBr8fJl:b-k2pczf) ([raw file](https://github.com/google/nomulus/blob/d4ce83197ddb5a8b56f60022e2078ced4277b1c1/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L275)):*
> ```Java
>     }
>     serverApproveEntities = forceEmptyToNull(serverApproveEntitiesBuilder.build());
>     hashCode = null; // reset the hash code since we may have changed the entities
> ```

Here too.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: 23 of 32 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZpV2QcDB91vTm0KzwT:b-xx4ho8) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

It seems safest to me to recalculate the hash code here rather than nulling it out.
</blockquote></details>

Why is that? The field is protected and any time it needs to be accessed, it is recomputed lazily. Why do the extra computation here when it might not be necessary? 

Especially since the code that recalculates the hash code isn't particularly neat, it's `hashCode = Arrays.hashCode(getSignificantFields().values().toArray());`
___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 275 at r4](https://reviewable.io/reviews/google/nomulus/1106#-MZoDN7E2NXl9qBr8fJk:-MZpVO8p2oEiD7MKUSD9:byz4vxa) ([raw file](https://github.com/google/nomulus/blob/d4ce83197ddb5a8b56f60022e2078ced4277b1c1/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L275)):*

Here too.
</blockquote></details>

see other

mindhog (approved):
> Reviewed 9 of 9 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/transfer/DomainTransferData.java, line 167 at r2](https://reviewable.io/reviews/google/nomulus/1106#-MZTJ0p6A34WuAmDTmw_:-MZrdZ44A8YL0oHCSDLF:b-67vmtz) ([raw file](https://github.com/google/nomulus/blob/6000b7044acf920c18b8376ce24cb0c0030f3011/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L167)):*

Why is that? The field is protected and any time it needs to be accessed, it is recomputed lazily. Why do the extra computation here when it might not be necessary? 

Especially since the code that recalculates the hash code isn't particularly neat, it's `hashCode = Arrays.hashCode(getSignificantFields().values().toArray());`
</blockquote></details>

yeah, it's actually better that we not recalculate it here -- keep in mind that the hash value shouldn't be calculated prior to full initialization of the object otherwise it's likely just wrong.  This function is called as part of the initialization.

That's why I'm concerned about the early initialization.  This is very relevant to issues discussed in #1116.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1106)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

PR #1115, opened 2021-04-28
Remove unnecessary MockitoExtension from Spec11PipelineTest

CydeWeys:
This is kind of a shot in the dark here, but this is one of the obvious
differences between this test class (which frequently experiences flakes) and
the other pipeline test classes which do not.

I also explicitly set the runner in the Spec11PipelineOptions, which seems
possibly related to the NPEs we're seeing in the flakes when it fails to get the
ThreadFactory. This is the only test that explicitly sets these options, so it's
not clear if the usual default to CrashingRunner works if the options are
overridden.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1115)

### DIFF
--- a/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
+++ b/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
@@ -16,6 +16,7 @@ package google.registry.beam.spec11;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth.assertWithMessage;
 import static google.registry.model.ImmutableObjectSubject.immutableObjectCorrespondence;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 
@@ -34,6 +35,7 @@ import google.registry.testing.DatastoreEntityExtension;
 import google.registry.testing.FakeClock;
 import google.registry.util.ResourceUtils;
 import java.io.File;
+import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import org.apache.beam.sdk.coders.KvCoder;
@@ -48,12 +50,8 @@ import org.json.JSONObject;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.api.io.TempDir;
-import org.mockito.junit.jupiter.MockitoExtension;
-import org.mockito.junit.jupiter.MockitoSettings;
-import org.mockito.quality.Strictness;
 
 /**
  * Unit tests for {@link Spec11Pipeline}.
@@ -62,9 +60,8 @@ import org.mockito.quality.Strictness;
  * Therefore we cannot fully test the pipeline but only test the two separate sink IO functions,
  * assuming that date is sourcede correctly the {@code BigQueryIO}.
  */
-@ExtendWith(MockitoExtension.class)
-@MockitoSettings(strictness = Strictness.STRICT_STUBS)
 class Spec11PipelineTest {
+
   private static final String DATE = "2020-01-27";
   private static final String SAFE_BROWSING_API_KEY = "api-key";
   private static final String REPORTING_BUCKET_URL = "reporting_bucket";
@@ -123,6 +120,11 @@ class Spec11PipelineTest {
                     KvCoder.of(
                         SerializableCoder.of(Subdomain.class),
                         SerializableCoder.of(ThreatMatch.class))));
+    assertWithMessage(
+            "Beam pipelines don't run in an App Engine environment, and thus the tests shouldn't be"
+                + " mocking one either")
+        .that(isAppEngine())
+        .isFalse();
   }
 
   @Test
@@ -191,6 +193,25 @@ class Spec11PipelineTest {
             Correspondence.from(
                 new ThreatMatchJsonPredicate(), "has fields with unordered threatTypes equal to"))
         .containsExactlyElementsIn(expectedFileContents.subList(1, expectedFileContents.size()));
+  }
+
+  // Adapted from Guava's MoreExecutors (where it is a private method)
+  private static boolean isAppEngine() {
+    if (System.getProperty("com.google.appengine.runtime.environment") == null) {
+      return false;
+    } else {
+      try {
+        return Class.forName("com.google.apphosting.api.ApiProxy")
+                .getMethod("getCurrentEnvironment")
+                .invoke(null)
+            != null;
+      } catch (ClassNotFoundException
+          | InvocationTargetException
+          | IllegalAccessException
+          | NoSuchMethodException e) {
+        return false;
+      }
+    }
   }
 
   /** Returns the text contents of a file under the beamBucket/results directory. */


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (changes requested):
I don't think setting the runner would matter. We are not using the test pipeline to construct the pipeline. Other parameters of the option is extracted when calling `saveToGcs` and `saveToSql`. It's better to remove it so to not cause confusion.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115#-:-MZOpYqC6ivKQLUViGxw:b-pwo0sr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1115#-MZP-dGG9GlSmhqBfFAN:-MZP-dGG9GlSmhqBfFAO:b-urrw6w) ([raw file](https://github.com/google/nomulus/blob/dd6e5a17c925740a01a5e4d56fac9a1762f607cd/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L94)):*
> ```
> transient
> ```
I talked from memory and made a mistake regarding the 'transient' keyword.

This class does not pass inner classes/lambdas to the pipeline so 'transient' is not needed here.

The likely problems are probably in the Spec11Pipeline class, where lambdas are used three times, once in saveToSql and twice in saveToGcs.

Since Spec11Pipeline saves the 'pipeline' parameter, in tests the TestPipelineExtension is saved there and gets involved in serialization.
We should try turn those three lambdas into static classes.

CydeWeys (commented):
When the options are specified (and this is the only test class where they are) it doesn't appear to default to `CrashingRunner`:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/beam/TestPipelineExtension.java;l=447-455;drc=d99278e7238acd472ded1a73d5b213c1dd27b70e?q=testpipelineextension&ss=nomulus%2Fnomulus

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115#-:-MZPM-05APQuEFI-AUuu:b-pk7xqx)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1115#-MZP-dGG9GlSmhqBfFAN:-MZQGZYo7XwOPVinfQGF:b33yn2w) ([raw file](https://github.com/google/nomulus/blob/dd6e5a17c925740a01a5e4d56fac9a1762f607cd/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L94)):*

> ```
> transient
> ```
I talked from memory and made a mistake regarding the 'transient' keyword.

This class does not pass inner classes/lambdas to the pipeline so 'transient' is not needed here.

The likely problems are probably in the Spec11Pipeline class, where lambdas are used three times, once in saveToSql and twice in saveToGcs.

Since Spec11Pipeline saves the 'pipeline' parameter, in tests the TestPipelineExtension is saved there and gets involved in serialization.
We should try turn those three lambdas into static classes.




</blockquote></details>

This seems like an intrinsic design flaw of the `TestPipelineExtnesion`, or our use of pipeline, is it not? We are encouraged to use lambas that conform to the `SerializableFunction` interface in Beam for readability. If the test pipeline cannot be safely deserialized when lambda is used multiple times due to multiple static initialization, it looks to me that for testing's sake we should not save the `Pipeline` object as a field in `Spec11Pipeline`, or any other pipelines that we write.

Also, `InvoicingPipeline` also calls multiple lambdas, and we haven't encountered any flakiness.

The NPE error about ThreadFactory in failing tests still looks not much related to the possible static initialization issues here, truth be told.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1115#-MZP-dGG9GlSmhqBfFAN:-MZT7Tko1QFqUAHyHwBa:bwi6l0t) ([raw file](https://github.com/google/nomulus/blob/dd6e5a17c925740a01a5e4d56fac9a1762f607cd/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L94)):*

This seems like an intrinsic design flaw of the `TestPipelineExtnesion`, or our use of pipeline, is it not? We are encouraged to use lambas that conform to the `SerializableFunction` interface in Beam for readability. If the test pipeline cannot be safely deserialized when lambda is used multiple times due to multiple static initialization, it looks to me that for testing's sake we should not save the `Pipeline` object as a field in `Spec11Pipeline`, or any other pipelines that we write.

Also, `InvoicingPipeline` also calls multiple lambdas, and we haven't encountered any flakiness.

The NPE error about ThreadFactory in failing tests still looks not much related to the possible static initialization issues here, truth be told.
</blockquote></details>

We should stop saving the pipeline object in our *pipeline classes. Maybe pass it in in the run() method instead.

Looking at the exception site, it seems that the only thing that can cause NPE there is that the code believe it is on AppEngine, which means the property com.google.appengine.runtime.environment is set. Maybe we should add an assert in the test class that this property is not set, and also print it in the kokoro runner.

jianglai (commented):
`CrashingRunner` actually [throws](https://beam.apache.org/releases/javadoc/2.28.0/org/apache/beam/sdk/testing/CrashingRunner.html) when the pipeline runs, which is not what we want.

Here this `Spec11PipelineOptions` object is not actually passed to the `Pipeline` object, so setting the runner has no effect anyway. Note that there's the `Spec11Pipeline` object, and there's the `Pipeline` object. The latter is passed as a parameter to the former during construction (but we should change that as Weimin pointed out) and that's the one that actually consumes the `Spec11PipelineOptions`. But here we are not actually passing an `Pipeline` to `Spec11Pipeline`, but rather constructing a `TestPipelineExtension`, which is then used to test the static functions on `Spec11Pipeline`. We only construct a `Spec11PipelineOptions` because we need some of the parameters (but not all) in the static function.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115#-:-MZTHt1M-3GURo4XZj7p:b-rknn2a)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1115#-MZP-dGG9GlSmhqBfFAN:-MZTJaaM2Sw5wOqQQDvX:b-glr0qx) ([raw file](https://github.com/google/nomulus/blob/dd6e5a17c925740a01a5e4d56fac9a1762f607cd/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L94)):*

We should stop saving the pipeline object in our *pipeline classes. Maybe pass it in in the run() method instead.

Looking at the exception site, it seems that the only thing that can cause NPE there is that the code believe it is on AppEngine, which means the property com.google.appengine.runtime.environment is set. Maybe we should add an assert in the test class that this property is not set, and also print it in the kokoro runner.
</blockquote></details>

I think calling it in run() is the right call (no pun intended). I'll do it in a separate PR. It's strange to think that the environment could **sometimes** be mistaken as app engine. But yeah, adding a check here could help us know for sure. Ben, do you want to add that in this PR?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1115#-MZP-dGG9GlSmhqBfFAN:-MZUEuc46ms-Q0qqyUuX:b69gn7b) ([raw file](https://github.com/google/nomulus/blob/dd6e5a17c925740a01a5e4d56fac9a1762f607cd/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L94)):*

I think calling it in run() is the right call (no pun intended). I'll do it in a separate PR. It's strange to think that the environment could **sometimes** be mistaken as app engine. But yeah, adding a check here could help us know for sure. Ben, do you want to add that in this PR?
</blockquote></details>

I'm struggling to find where the App Engine check is -- do you remember where?  I saw it yesterday.

CydeWeys (commented):
PTAL

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115#-:-MZUMCITEpas5MOcJ9lF:bb71pro)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1115#-MZP-dGG9GlSmhqBfFAN:-MZUMBh485oYrsuQPA9O:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd6e5a17c925740a01a5e4d56fac9a1762f607cd/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L94)):*

I'm struggling to find where the App Engine check is -- do you remember where?  I saw it yesterday.
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1115)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1116, opened 2021-04-28
Always detach entities during load

mindhog:
The mutations on non-transient fields that we do in some of the PostLoad
methods have been causing the objects to be marked as "dirty", and hibernate
has been quietly persisting them during transaction commit.

By detaching the entities on load, we avoid any possibility of this, which
works in our case because we treat all of our model objects as immutable
during normal use.

There is another mixed blessing to this: lazy loading won't work on these
objects once they are detached from a session, meaning that all fields must be
initialized up front.  This is unfortunate in that we don't always need those
lazy-loaded fields and there is a performance cost to loading them, but it is
also useful in that objects will now be complete when used outside of the
transaction that loaded them (prior to this, an attempt to access a
lazy-loaded field after its transaction closed would have caused an error at
runtime).

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -36,7 +36,9 @@ import javax.persistence.Index;
 import javax.persistence.JoinColumn;
 import javax.persistence.JoinTable;
 import javax.persistence.OneToMany;
+import javax.persistence.PostLoad;
 import javax.persistence.Table;
+import org.hibernate.Hibernate;
 import org.joda.time.DateTime;
 
 /**
@@ -74,6 +76,8 @@ public class DomainBase extends DomainContent
     return super.getRepoId();
   }
 
+  // It seems like this should be FetchType.EAGER, but for some reason when we do that we get a lazy
+  // load error during the load of a domain.
   @ElementCollection
   @JoinTable(
       name = "DomainHost",
@@ -133,6 +137,19 @@ public class DomainBase extends DomainContent
   @SuppressWarnings("UnusedMethod")
   private Set<DelegationSignerData> getInternalDelegationSignerData() {
     return dsData;
+  }
+
+  /**
+   * Post-load method to eager load the collections.
+   *
+   * <p>We've named this "postLoad2" because there's already a "postLoad" in DomainContent and our
+   * code to invoke all post-loads doesn't seem to call both if they have the same name.
+   */
+  @PostLoad
+  @SuppressWarnings("UnusedMethod")
+  private void postLoad2() {
+    Hibernate.initialize(getInternalDelegationSignerData());
+    Hibernate.initialize(getInternalGracePeriods());
   }
 
   @Override
--- a/core/src/main/java/google/registry/model/domain/DomainHistory.java
+++ b/core/src/main/java/google/registry/model/domain/DomainHistory.java
@@ -54,6 +54,7 @@ import javax.persistence.JoinTable;
 import javax.persistence.OneToMany;
 import javax.persistence.PostLoad;
 import javax.persistence.Table;
+import org.hibernate.Hibernate;
 
 /**
  * A persisted history entry representing an EPP modification to a domain.
@@ -256,6 +257,12 @@ public class DomainHistory extends HistoryEntry implements SqlEntity {
         }
       }
     }
+
+    // Eager fetch doesn't seem to work for this, so we do it explicitly.
+    Hibernate.initialize(getDomainTransactionRecords());
+    Hibernate.initialize(getNsHosts());
+    Hibernate.initialize(dsDataHistories);
+    Hibernate.initialize(gracePeriodHistories);
   }
 
   // In Datastore, save as a HistoryEntry object regardless of this object's type
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -285,8 +285,9 @@ public class DatastoreTransactionManager implements TransactionManager {
   }
 
   @Override
-  public void delete(Object entity) {
+  public <T> T delete(T entity) {
     syncIfTransactionless(getOfy().delete().entity(entity));
+    return entity;
   }
 
   @Override
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -55,6 +55,7 @@ import javax.persistence.PostLoad;
 import javax.persistence.PrePersist;
 import javax.persistence.Table;
 import javax.persistence.Transient;
+import org.hibernate.Hibernate;
 import org.hibernate.LazyInitializationException;
 import org.joda.money.CurrencyUnit;
 import org.joda.money.Money;
@@ -319,5 +320,6 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   @PostLoad
   void postLoad() {
     creationTime = lastUpdateTime;
+    Hibernate.initialize(labelsToPrices);
   }
 }
--- a/core/src/main/java/google/registry/model/registry/label/ReservedList.java
+++ b/core/src/main/java/google/registry/model/registry/label/ReservedList.java
@@ -52,7 +52,9 @@ import javax.persistence.Embeddable;
 import javax.persistence.Index;
 import javax.persistence.JoinColumn;
 import javax.persistence.MapKeyColumn;
+import javax.persistence.PostLoad;
 import javax.persistence.Table;
+import org.hibernate.Hibernate;
 import org.joda.time.DateTime;
 
 /**
@@ -81,6 +83,11 @@ public final class ReservedList
 
   @Column(nullable = false)
   boolean shouldPublish = true;
+
+  @PostLoad
+  void postLoad() {
+    Hibernate.initialize(reservedListMap);
+  }
 
   /**
    * A reserved list entry entity, persisted to Datastore, that represents a single label and its
--- a/core/src/main/java/google/registry/model/tmch/ClaimsListShard.java
+++ b/core/src/main/java/google/registry/model/tmch/ClaimsListShard.java
@@ -58,8 +58,10 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.JoinColumn;
 import javax.persistence.MapKeyColumn;
+import javax.persistence.PostLoad;
 import javax.persistence.Table;
 import javax.persistence.Transient;
+import org.hibernate.Hibernate;
 import org.joda.time.DateTime;
 
 /**
@@ -136,6 +138,12 @@ public class ClaimsListShard extends ImmutableObject implements NonReplicatedEnt
   @Ignore @Transient boolean isShard = false;
 
   private static final Retrier LOADER_RETRIER = new Retrier(new SystemSleeper(), 2);
+
+  @PostLoad
+  @SuppressWarnings("UnusedMethod")
+  private void postLoad() {
+    Hibernate.initialize(labelsToKeys);
+  }
 
   private static Optional<ClaimsListShard> loadClaimsListShard() {
     // Find the most recent revision.
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -51,6 +51,7 @@ import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import javax.annotation.Nullable;
 import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.EntityTransaction;
@@ -392,7 +393,8 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
   public <T> Optional<T> loadByKeyIfPresent(VKey<T> key) {
     checkArgumentNotNull(key, "key must be specified");
     assertInTransaction();
-    return Optional.ofNullable(getEntityManager().find(key.getKind(), key.getSqlKey()));
+    return Optional.ofNullable(getEntityManager().find(key.getKind(), key.getSqlKey()))
+        .map(this::detach);
   }
 
   @Override
@@ -406,7 +408,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
         .map(
             key ->
                 new SimpleEntry<VKey<? extends T>, T>(
-                    key, getEntityManager().find(key.getKind(), key.getSqlKey())))
+                    key, detach(getEntityManager().find(key.getKind(), key.getSqlKey()))))
         .filter(entry -> entry.getValue() != null)
         .collect(toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
   }
@@ -428,7 +430,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     if (result == null) {
       throw new NoSuchElementException(key.toString());
     }
-    return result;
+    return detach(result);
   }
 
   @Override
@@ -468,12 +470,12 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
   public <T> ImmutableList<T> loadAllOf(Class<T> clazz) {
     checkArgumentNotNull(clazz, "clazz must be specified");
     assertInTransaction();
-    return ImmutableList.copyOf(
-        getEntityManager()
-            .createQuery(
-                String.format("SELECT entity FROM %s entity", getEntityType(clazz).getName()),
-                clazz)
-            .getResultList());
+    return getEntityManager()
+        .createQuery(
+            String.format("SELECT entity FROM %s entity", getEntityType(clazz).getName()), clazz)
+        .getResultStream()
+        .map(this::detach)
+        .collect(toImmutableList());
   }
 
   private int internalDelete(VKey<?> key) {
@@ -505,18 +507,19 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
   }
 
   @Override
-  public void delete(Object entity) {
+  public <T> T delete(T entity) {
     checkArgumentNotNull(entity, "entity must be specified");
     if (isEntityOfIgnoredClass(entity)) {
-      return;
+      return entity;
     }
     assertInTransaction();
     entity = toChildHistoryEntryIfPossible(entity);
-    Object managedEntity = entity;
+    T managedEntity = entity;
     if (!getEntityManager().contains(entity)) {
       managedEntity = getEntityManager().merge(entity);
     }
     getEntityManager().remove(managedEntity);
+    return managedEntity;
   }
 
   @Override
@@ -536,7 +539,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
 
   @Override
   public <T> QueryComposer<T> createQueryComposer(Class<T> entity) {
-    return new JpaQueryComposerImpl<T>(entity, getEntityManager());
+    return new JpaQueryComposerImpl<T>(entity);
   }
 
   @Override
@@ -639,6 +642,15 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     }
   }
 
+  /** Detach the entity, suitable for use in Optional.map(). */
+  @Nullable
+  private <T> T detach(@Nullable T entity) {
+    if (entity != null) {
+      getEntityManager().detach(entity);
+    }
+    return entity;
+  }
+
   private static class TransactionInfo {
     EntityManager entityManager;
     boolean inTransaction = false;
@@ -691,17 +703,15 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     }
   }
 
-  private static class JpaQueryComposerImpl<T> extends QueryComposer<T> {
+  private class JpaQueryComposerImpl<T> extends QueryComposer<T> {
 
-    EntityManager em;
-
-    JpaQueryComposerImpl(Class<T> entityClass, EntityManager em) {
+    JpaQueryComposerImpl(Class<T> entityClass) {
       super(entityClass);
-      this.em = em;
     }
 
     private TypedQuery<T> buildQuery() {
-      CriteriaQueryBuilder<T> queryBuilder = CriteriaQueryBuilder.create(em, entityClass);
+      CriteriaQueryBuilder<T> queryBuilder =
+          CriteriaQueryBuilder.create(getEntityManager(), entityClass);
       return addCriteria(queryBuilder);
     }
 
@@ -714,28 +724,29 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
         queryBuilder.orderByAsc(orderBy);
       }
 
-      return em.createQuery(queryBuilder.build());
+      return getEntityManager().createQuery(queryBuilder.build());
     }
 
     @Override
     public Optional<T> first() {
       List<T> results = buildQuery().setMaxResults(1).getResultList();
-      return results.size() > 0 ? Optional.of(results.get(0)) : Optional.empty();
+      return results.size() > 0 ? Optional.of(detach(results.get(0))) : Optional.empty();
     }
 
     @Override
     public T getSingleResult() {
-      return buildQuery().getSingleResult();
+      return detach(buildQuery().getSingleResult());
     }
 
     @Override
     public Stream<T> stream() {
-      return buildQuery().getResultStream();
+      return buildQuery().getResultStream().map(JpaTransactionManagerImpl.this::detach);
     }
 
     @Override
     public long count() {
-      CriteriaQueryBuilder<Long> queryBuilder = CriteriaQueryBuilder.createCount(em, entityClass);
+      CriteriaQueryBuilder<Long> queryBuilder =
+          CriteriaQueryBuilder.createCount(getEntityManager(), entityClass);
       return addCriteria(queryBuilder).getSingleResult();
     }
   }
--- a/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java
+++ b/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java
@@ -252,8 +252,13 @@ public interface TransactionManager {
   /** Deletes the set of entities by their key id. */
   void delete(Iterable<? extends VKey<?>> keys);
 
-  /** Deletes the given entity from the database. */
-  void delete(Object entity);
+  /**
+   * Deletes the given entity from the database.
+   *
+   * <p>This returns the deleted entity, which may not necessarily be the same as the original
+   * entity passed in.
+   */
+  <T> T delete(T entity);
 
   /**
    * Deletes the entity by its id without writing commit logs if the underlying database is
--- a/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java
+++ b/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java
@@ -74,7 +74,7 @@ class EntityCallbacksListenerTest {
             .transact(
                 () -> {
                   TestEntity removed = jpaTm().loadByKey(VKey.createSql(TestEntity.class, "id"));
-                  jpaTm().delete(removed);
+                  removed = jpaTm().delete(removed);
                   return removed;
                 });
     checkAll(testRemove, 0, 0, 1, 1);
@@ -97,6 +97,19 @@ class EntityCallbacksListenerTest {
   @Test
   void verifyHasMethodAnnotatedWithEmbedded_work() {
     assertThat(hasMethodAnnotatedWithEmbedded(ViolationEntity.class)).isTrue();
+  }
+
+  @Test
+  void verifyCallbacksNotCalledOnCommit() {
+    TestEntity testEntity = new TestEntity();
+    jpaTm().transact(() -> jpaTm().insert(testEntity));
+
+    TestEntity testLoad =
+        jpaTm().transact(() -> jpaTm().loadByKey(VKey.createSql(TestEntity.class, "id")));
+    assertThat(testLoad.entityPreUpdate).isEqualTo(0);
+
+    testLoad = jpaTm().transact(() -> jpaTm().loadByKey(VKey.createSql(TestEntity.class, "id")));
+    assertThat(testLoad.foo).isEqualTo(1); // since we didn't save it, should only be 1
   }
 
   private static boolean hasMethodAnnotatedWithEmbedded(Class<?> entityType) {
@@ -155,12 +168,19 @@ class EntityCallbacksListenerTest {
     int foo = 0;
 
     @Transient int entityPostLoad = 0;
+    @Transient int entityPreUpdate = 0;
 
     @Embedded EntityEmbedded entityEmbedded = new EntityEmbedded();
 
     @PostLoad
     void entityPostLoad() {
       entityPostLoad++;
+      foo++;
+    }
+
+    @PreUpdate
+    void entityPreUpdate() {
+      entityPreUpdate++;
     }
   }
 
--- a/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java
+++ b/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java
@@ -14,8 +14,10 @@
 
 package google.registry.persistence.transaction;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.truth.Truth.assertThat;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.testing.DatabaseHelper.assertDetached;
 import static google.registry.testing.TestDataHelper.fileClassPath;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
@@ -25,9 +27,11 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import google.registry.model.ImmutableObject;
 import google.registry.persistence.VKey;
 import google.registry.persistence.transaction.JpaTestRules.JpaUnitTestExtension;
+import google.registry.testing.DatabaseHelper;
 import google.registry.testing.FakeClock;
 import java.io.Serializable;
 import java.math.BigInteger;
@@ -384,7 +388,7 @@ class JpaTransactionManagerImplTest {
   void load_succeeds() {
     assertThat(jpaTm().transact(() -> jpaTm().exists(theEntity))).isFalse();
     jpaTm().transact(() -> jpaTm().insert(theEntity));
-    TestEntity persisted = jpaTm().transact(() -> jpaTm().loadByKey(theEntityKey));
+    TestEntity persisted = jpaTm().transact(() -> assertDetached(jpaTm().loadByKey(theEntityKey)));
     assertThat(persisted.name).isEqualTo("theEntity");
     assertThat(persisted.data).isEqualTo("foo");
   }
@@ -398,10 +402,19 @@ class JpaTransactionManagerImplTest {
   }
 
   @Test
+  void loadByEntity_succeeds() {
+    jpaTm().transact(() -> jpaTm().insert(theEntity));
+    TestEntity persisted = jpaTm().transact(() -> assertDetached(jpaTm().loadByEntity(theEntity)));
+    assertThat(persisted.name).isEqualTo("theEntity");
+    assertThat(persisted.data).isEqualTo("foo");
+  }
+
+  @Test
   void maybeLoad_succeeds() {
     assertThat(jpaTm().transact(() -> jpaTm().exists(theEntity))).isFalse();
     jpaTm().transact(() -> jpaTm().insert(theEntity));
-    TestEntity persisted = jpaTm().transact(() -> jpaTm().loadByKeyIfPresent(theEntityKey).get());
+    TestEntity persisted =
+        jpaTm().transact(() -> assertDetached(jpaTm().loadByKeyIfPresent(theEntityKey).get()));
     assertThat(persisted.name).isEqualTo("theEntity");
     assertThat(persisted.data).isEqualTo("foo");
   }
@@ -417,17 +430,81 @@ class JpaTransactionManagerImplTest {
   void loadCompoundIdEntity_succeeds() {
     assertThat(jpaTm().transact(() -> jpaTm().exists(compoundIdEntity))).isFalse();
     jpaTm().transact(() -> jpaTm().insert(compoundIdEntity));
-    TestCompoundIdEntity persisted = jpaTm().transact(() -> jpaTm().loadByKey(compoundIdEntityKey));
+    TestCompoundIdEntity persisted =
+        jpaTm().transact(() -> assertDetached(jpaTm().loadByKey(compoundIdEntityKey)));
     assertThat(persisted.name).isEqualTo("compoundIdEntity");
     assertThat(persisted.age).isEqualTo(10);
     assertThat(persisted.data).isEqualTo("foo");
   }
 
   @Test
+  void loadByKeysIfPresent() {
+    jpaTm().transact(() -> jpaTm().insert(theEntity));
+    jpaTm()
+        .transact(
+            () -> {
+              ImmutableMap<VKey<? extends TestEntity>, TestEntity> results =
+                  jpaTm()
+                      .loadByKeysIfPresent(
+                          ImmutableList.of(
+                              theEntityKey, VKey.createSql(TestEntity.class, "does-not-exist")));
+
+              assertThat(results).isEqualTo(ImmutableMap.of(theEntityKey, theEntity));
+              assertDetached(results.get(theEntityKey));
+            });
+  }
+
+  @Test
+  void loadByKeys_succeeds() {
+    jpaTm().transact(() -> jpaTm().insert(theEntity));
+    jpaTm()
+        .transact(
+            () -> {
+              ImmutableMap<VKey<? extends TestEntity>, TestEntity> results =
+                  jpaTm().loadByKeysIfPresent(ImmutableList.of(theEntityKey));
+              assertThat(results).isEqualTo(ImmutableMap.of(theEntityKey, theEntity));
+              assertDetached(results.get(theEntityKey));
+            });
+  }
+
+  @Test
+  void loadByEntitiesIfPresent_succeeds() {
+    jpaTm().transact(() -> jpaTm().insert(theEntity));
+    jpaTm()
+        .transact(
+            () -> {
+              ImmutableList<TestEntity> results =
+                  jpaTm()
+                      .loadByEntitiesIfPresent(
+                          ImmutableList.of(theEntity, new TestEntity("does-not-exist", "bar")));
+              assertThat(results).isEqualTo(ImmutableList.of(theEntity));
+              assertDetached(results.get(0));
+            });
+  }
+
+  @Test
+  void loadByEntities_succeeds() {
+    jpaTm().transact(() -> jpaTm().insert(theEntity));
+    jpaTm()
+        .transact(
+            () -> {
+              ImmutableList<TestEntity> results =
+                  jpaTm().loadByEntities(ImmutableList.of(theEntity));
+              assertThat(results).isEqualTo(ImmutableList.of(theEntity));
+              assertDetached(results.get(0));
+            });
+  }
+
+  @Test
   void loadAll_succeeds() {
     jpaTm().transact(() -> jpaTm().insertAll(moreEntities));
     ImmutableList<TestEntity> persisted =
-        jpaTm().transact(() -> jpaTm().loadAllOf(TestEntity.class));
+        jpaTm()
+            .transact(
+                () ->
+                    jpaTm().loadAllOf(TestEntity.class).stream()
+                        .map(DatabaseHelper::assertDetached)
+                        .collect(toImmutableList()));
     assertThat(persisted).containsExactlyElementsIn(moreEntities);
   }
 
--- a/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java
+++ b/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java
@@ -27,6 +27,7 @@ import com.googlecode.objectify.annotation.Id;
 import com.googlecode.objectify.annotation.Index;
 import google.registry.model.ImmutableObject;
 import google.registry.testing.AppEngineExtension;
+import google.registry.testing.DatabaseHelper;
 import google.registry.testing.DualDatabaseTest;
 import google.registry.testing.FakeClock;
 import google.registry.testing.TestOfyAndSql;
@@ -73,16 +74,9 @@ public class QueryComposerTest {
             transactIfJpaTm(
                 () ->
                     tm().createQueryComposer(TestEntity.class)
-                        .where("name", Comparator.EQ, "bravo")
-                        .first()
-                        .get()))
-        .isEqualTo(bravo);
-    assertThat(
-            transactIfJpaTm(
-                () ->
-                    tm().createQueryComposer(TestEntity.class)
                         .where("name", Comparator.GT, "bravo")
                         .first()
+                        .map(DatabaseHelper::assertDetached)
                         .get()))
         .isEqualTo(charlie);
     assertThat(
@@ -91,6 +85,7 @@ public class QueryComposerTest {
                     tm().createQueryComposer(TestEntity.class)
                         .where("name", Comparator.GTE, "charlie")
                         .first()
+                        .map(DatabaseHelper::assertDetached)
                         .get()))
         .isEqualTo(charlie);
     assertThat(
@@ -99,6 +94,7 @@ public class QueryComposerTest {
                     tm().createQueryComposer(TestEntity.class)
                         .where("name", Comparator.LT, "bravo")
                         .first()
+                        .map(DatabaseHelper::assertDetached)
                         .get()))
         .isEqualTo(alpha);
     assertThat(
@@ -107,6 +103,7 @@ public class QueryComposerTest {
                     tm().createQueryComposer(TestEntity.class)
                         .where("name", Comparator.LTE, "alpha")
                         .first()
+                        .map(DatabaseHelper::assertDetached)
                         .get()))
         .isEqualTo(alpha);
   }
@@ -127,9 +124,10 @@ public class QueryComposerTest {
     assertThat(
             transactIfJpaTm(
                 () ->
-                    tm().createQueryComposer(TestEntity.class)
-                        .where("name", Comparator.EQ, "alpha")
-                        .getSingleResult()))
+                    DatabaseHelper.assertDetached(
+                        tm().createQueryComposer(TestEntity.class)
+                            .where("name", Comparator.EQ, "alpha")
+                            .getSingleResult())))
         .isEqualTo(alpha);
   }
 
@@ -164,17 +162,9 @@ public class QueryComposerTest {
                 () ->
                     tm()
                         .createQueryComposer(TestEntity.class)
-                        .where("name", Comparator.EQ, "alpha")
-                        .stream()
-                        .collect(toImmutableList())))
-        .isEqualTo(ImmutableList.of(alpha));
-    assertThat(
-            transactIfJpaTm(
-                () ->
-                    tm()
-                        .createQueryComposer(TestEntity.class)
                         .where("name", Comparator.GT, "alpha")
                         .stream()
+                        .map(DatabaseHelper::assertDetached)
                         .collect(toImmutableList())))
         .isEqualTo(ImmutableList.of(bravo, charlie));
     assertThat(
@@ -184,6 +174,7 @@ public class QueryComposerTest {
                         .createQueryComposer(TestEntity.class)
                         .where("name", Comparator.GTE, "bravo")
                         .stream()
+                        .map(DatabaseHelper::assertDetached)
                         .collect(toImmutableList())))
         .isEqualTo(ImmutableList.of(bravo, charlie));
     assertThat(
@@ -193,6 +184,7 @@ public class QueryComposerTest {
                         .createQueryComposer(TestEntity.class)
                         .where("name", Comparator.LT, "charlie")
                         .stream()
+                        .map(DatabaseHelper::assertDetached)
                         .collect(toImmutableList())))
         .isEqualTo(ImmutableList.of(alpha, bravo));
     assertThat(
@@ -202,6 +194,7 @@ public class QueryComposerTest {
                         .createQueryComposer(TestEntity.class)
                         .where("name", Comparator.LTE, "bravo")
                         .stream()
+                        .map(DatabaseHelper::assertDetached)
                         .collect(toImmutableList())))
         .isEqualTo(ImmutableList.of(alpha, bravo));
   }
@@ -214,6 +207,7 @@ public class QueryComposerTest {
                     tm().createQueryComposer(TestEntity.class)
                         .where("val", Comparator.EQ, 2)
                         .first()
+                        .map(DatabaseHelper::assertDetached)
                         .get()))
         .isEqualTo(bravo);
   }
@@ -228,6 +222,7 @@ public class QueryComposerTest {
                         .where("val", Comparator.GT, 1)
                         .orderBy("val")
                         .stream()
+                        .map(DatabaseHelper::assertDetached)
                         .collect(toImmutableList())))
         .isEqualTo(ImmutableList.of(bravo, alpha));
   }
--- a/core/src/test/java/google/registry/persistence/transaction/TransactionManagerTest.java
+++ b/core/src/test/java/google/registry/persistence/transaction/TransactionManagerTest.java
@@ -378,6 +378,17 @@ public class TransactionManagerTest {
         () -> tm().transact(() -> tm().loadAllOf(TestEntity.class)));
   }
 
+  @TestOfyAndSql
+  void mutatedObjectNotPersisted() {
+    tm().transact(() -> tm().insert(theEntity));
+    tm().transact(
+            () -> {
+              TestEntity e = tm().loadByKey(theEntity.key());
+              e.data = "some other data!";
+            });
+    assertThat(tm().transact(() -> tm().loadByKey(theEntity.key())).data).isEqualTo("foo");
+  }
+
   private static void assertEntityExists(TestEntity entity) {
     assertThat(tm().transact(() -> tm().exists(entity))).isTrue();
   }
--- a/core/src/test/java/google/registry/testing/DatabaseHelper.java
+++ b/core/src/test/java/google/registry/testing/DatabaseHelper.java
@@ -117,7 +117,6 @@ import google.registry.model.transfer.TransferData;
 import google.registry.model.transfer.TransferStatus;
 import google.registry.persistence.VKey;
 import google.registry.tmch.LordnTaskUtils;
-import java.lang.reflect.Field;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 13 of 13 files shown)


gbrodman (changes requested):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOm7rz8WBE2TQpYyfr:-MZOm7s-0OIYuFm_EfZD:b-ud4uxn) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainBase.java#L79)):*
> ```Java
>   }
> 
>   // It seems like this should be FetchType.EAGER, but for some reason when we do that we get a lazy
> ```

any way of debugging why this might be the case?
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOd0iQ4gwa6fYcFGAX:-MZOd0iQ4gwa6fYcFGAY:b-5mnw1k) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainBase.java#L150)):*
> ```Java
>   @PostLoad
>   @SuppressWarnings("UnusedMethod")
>   private void postLoad2() {
> ```

we can just also call this postLoad and call super.postLoad() as the first line; I believe that's what we do elsewhere 
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOmJtl-Sfvf49NC___:-MZOmJtl-Sfvf49NC__a:ba3ydry) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*
> ```Java
> 
>     // Eager fetch doesn't seem to work for this, so we do it explicitly.
>     Hibernate.initialize(getDomainTransactionRecords());
> ```

should be able to refer to the field directly rather than the getter?  (same for nsHosts)
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 647 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOviBW5O8Mg9iYQG9v:-MZOviBW5O8Mg9iYQG9w:b-b0a3k6) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L647)):*
> ```Java
>   /** Detach the entity, suitable for use in Optional.map(). */
>   @Nullable
>   private <T> T detach(@Nullable T entity) {
> ```

It's very annoying that Optional doesn't have a `peek` method akin to that of the Stream 
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOwCK18SKJjaIUg9Op:-MZOwCK18SKJjaIUg9Oq:bxh8e96) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L112)):*
> ```Java
> 
>     testLoad = jpaTm().transact(() -> jpaTm().loadByKey(VKey.createSql(TestEntity.class, "id")));
>     assertThat(testLoad.foo).isEqualTo(1); // since we didn't save it, should only be 1
> ```

why use `foo` instead of the more aptly-named `postLoad`?

CydeWeys (changes requested):
Overall I think this needs a rethink, as always loading some of these really expensive fields on domains doesn't seem workable.  We need the ability to only load them when necessary.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116#-:-MZTIDXoAnPcTB2SuUM3:b-49ep90)* status: all files reviewed, 8 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOmJtl-Sfvf49NC___:-MZTH4e-1fruAXXxhtX3:bbsvl05) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

should be able to refer to the field directly rather than the getter?  (same for nsHosts)
</blockquote></details>

Agreed that this makes more sense, as getters might have logic in them that could cause issues here.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-MZTHeXA7okyRwLX1yvR:b-5quv3c) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*
> ```Java
> 
>     // Eager fetch doesn't seem to work for this, so we do it explicitly.
>     Hibernate.initialize(getDomainTransactionRecords());
> ```

There's a few fields here (`domainTransactionRecords`, `dsDataHistories`, `gracePeriodHistories`) that we definitely do not want to load every time along with the domain, as we don't need them in the vast majority of situations and they also have the potential to grow in size without bounds.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 263 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHw-YAqsMMjdd8zv0:-MZTHw-YAqsMMjdd8zv1:b-7a524y) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L263)):*
> ```Java
>     // Eager fetch doesn't seem to work for this, so we do it explicitly.
>     Hibernate.initialize(getDomainTransactionRecords());
>     Hibernate.initialize(getNsHosts());
> ```

This can get pretty expensive considering we have a lot of use cases that load the domain for some simple operation (e.g. checking its status) but doesn't need all the nameservers.  We really need the option to only learn this stuff when necessary.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 264 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHV9WDuz7OaA3I_mj:-MZTHV9WDuz7OaA3I_mk:bw4rbsj) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L264)):*
> ```Java
>     Hibernate.initialize(getDomainTransactionRecords());
>     Hibernate.initialize(getNsHosts());
>     Hibernate.initialize(dsDataHistories);
> ```

All of these are necessary because eager loading isn't working for some reason?  I think we should try harder to get eager loading working, as this seems errorprone.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 647 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHK1a2_N0L7mnXeu5:-MZTHK1a2_N0L7mnXeu6:b-qkm8q6) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L647)):*
> ```Java
>   /** Detach the entity, suitable for use in Optional.map(). */
>   @Nullable
>   private <T> T detach(@Nullable T entity) {
> ```

Why not return an `Optional`?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOm7rz8WBE2TQpYyfr:-MZTdWzDE1xnJK2VJxlb:b-iw30n4) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainBase.java#L79)):*

any way of debugging why this might be the case?
</blockquote></details>

Sure, we have the hibernate source code :-)

I used FetchType.EAGER repeatedly for a number of these fields, and in every case hibernate still attempted to lazy-load them after the initial load.  I'm having a hard time finding it now, but someone on SO asserted that EAGER can be used on at most two fields.  I so far haven't found any reference to any such limitation in any of the documentation I've read, but clearly something weird is going on here.

I don't think we want to invest too much in this.  I think EAGER should be viewed with distrust.  I would recommend verifying its behavior in every case where it's used. 
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOd0iQ4gwa6fYcFGAX:-MZTj5yFB6QZi08sq0SO:b-o0761l) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainBase.java#L150)):*

we can just also call this postLoad and call super.postLoad() as the first line; I believe that's what we do elsewhere 
</blockquote></details>

That works too, done.  Honestly, I'm not sure which is better.  The unfortunate thing is that by using private methods for these, we don't discover that there's a problem until it we somehow discover that the base class method isn't being called, so we should probably either 1) fix the annotation processor so it does this correctly or 2) require that these methods be derived-class-visible from within the annotation processor so the compiler will notify us if we're not doing an override correctly.   
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOmJtl-Sfvf49NC___:-MZTpgp72NDVy7PcstOT:b-cox49g) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

Agreed that this makes more sense, as getters might have logic in them that could cause issues here.
</blockquote></details>

I wasn't clear on whether that would work since the annotation exists on the getters/setters in the derived classes in these cases instead of the field itself.  However, it seems to work so... done.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-MZTrTBUCPa7dclKKJ8O:b-aso4aa) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

There's a few fields here (`domainTransactionRecords`, `dsDataHistories`, `gracePeriodHistories`) that we definitely do not want to load every time along with the domain, as we don't need them in the vast majority of situations and they also have the potential to grow in size without bounds.
</blockquote></details>

Yeah, we need to make a judgement call on this.  AFAICT, there are two ways we can solve this problem with detaching:

1) Always detach at load.  In this case, lazy load doesn't exist and we have to pre-load all fields.
2) Detach loaded objects prior to commit.  This works for preventing unexpected saves, but doesn't solve another problem that we currently have, in that entities that leak out of the transaction are incomplete and lazy load may throw an exception somewhere down the line if they are accessed.

I agree that loading these fields is less performant than the load before commit approach. But if we don't have a principled way of dealing with the lazy loads and ensuring that they can not occur, we're running the risk of introducing runtime errors in the production environment.  I chose 1) out of a preference for safety over performance.  Also, 1) happens to be easier to implement :-)

What do you guys think? If we're going to do 2) I think we should at least have some way to ensure that  lazily populated objects can't escape their transactions. 
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 263 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHw-YAqsMMjdd8zv0:-MZTve3m3c_-drMf0E-O:b-wx09wi) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L263)):*

This can get pretty expensive considering we have a lot of use cases that load the domain for some simple operation (e.g. checking its status) but doesn't need all the nameservers.  We really need the option to only learn this stuff when necessary.
</blockquote></details>

yeah, it's not optimal.  Though it's not quite as bad as it seems: remember that we're only loading nameserver VKeys, not the nameserver records themselves.  We really don't load more than we would for datastore, the difference is that in this case we're loading across multiple tables.

If the performance hit is tolerable for our current load, we might want to consider just putting up with it until after the migration.  It will be easier to change at that point.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 264 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHV9WDuz7OaA3I_mj:-MZTuEjxBHEnw-hAaWJe:b-mw4iso) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L264)):*

All of these are necessary because eager loading isn't working for some reason?  I think we should try harder to get eager loading working, as this seems errorprone.
</blockquote></details>

Yes, see my response above in DomainBase.java: it's frustrating that eager loading isn't working as expected, if someone wants to investigate I'd be happy to see a better solution, but honestly I'm a bit worn out from fighting with Hibernate at this point.

Also, I have to say that this actually seems to me to be the less errorprone approach: it directly calls the function responsible for forcing field load as opposed to relying on annotations and magic happening (and quietly behaving differently when it doesn't...)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 11 of 14 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-MZXWXkE1tTeRDfhglLQ:b6gbwle) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

Yeah, we need to make a judgement call on this.  AFAICT, there are two ways we can solve this problem with detaching:

1) Always detach at load.  In this case, lazy load doesn't exist and we have to pre-load all fields.
2) Detach loaded objects prior to commit.  This works for preventing unexpected saves, but doesn't solve another problem that we currently have, in that entities that leak out of the transaction are incomplete and lazy load may throw an exception somewhere down the line if they are accessed.

I agree that loading these fields is less performant than the load before commit approach. But if we don't have a principled way of dealing with the lazy loads and ensuring that they can not occur, we're running the risk of introducing runtime errors in the production environment.  I chose 1) out of a preference for safety over performance.  Also, 1) happens to be easier to implement :-)

What do you guys think? If we're going to do 2) I think we should at least have some way to ensure that  lazily populated objects can't escape their transactions. 
</blockquote></details>

Actually, as I think about this further I don't see how we can safely lazy-load at all.  Doing so would invalidate the hash code.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 11 of 14 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-MZZQi9BAbqxM77zWpSe:bljjlxi) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

Actually, as I think about this further I don't see how we can safely lazy-load at all.  Doing so would invalidate the hash code.  
</blockquote></details>

This is an unacceptable hit to performance, though.  We're talking about loading an unbounded number of additional rows from other tables as part of loading a single domain.  That has an unbounded negative impact on our performance.  We have to be able to only load the expensive stuff when we actually need it.  We have to go with option 2, if those are the only 2 options.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 13 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, @mindhog, and @override)

___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOm7rz8WBE2TQpYyfr:-MZm76U77_0zDAEz80HP:b44e998) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainBase.java#L79)):*

Yeah makes sense, we should be LAZYing as much as possible anyway, which is part of what makes the change in general problematic (Ben's comments). 
</blockquote></details>

I'm not so sure that "always lazy-load" is optimal from a performance perspective.  Past benchmarking suggests that fewer operations against a larger set of tables may actually be preferable.

But in any case, lazy loads are going to require at least some extra effort for immutable objects, see my comments below. 
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOd0iQ4gwa6fYcFGAX:-MZmDvBiCyHFjmja0Ivr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainBase.java#L150)):*

IIRC, Checkstyle or the Java compiler will warn / give errors if you have a method that's overriding a superclass's method without an @Override notation. We could potentially have some sort of presubmit rule that checks to make sure that if we have 

```
@Override
(private )? void postLoad() {
```

that we immediately call `super.postLoad()`? Either way, yeah I think it's best to use the same name. I think we use package-protected postLoad methods elsewhere anyway. 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-MZm8k2J6hBvRDIJ7Z7w:b18x4l5) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

Option 2 doesn't seem to me to be a big problem. For one, I think we've seen that relatively infrequently. For two, I think it's not the worst thing in the world to continue to be in a transaction as you use the object you loaded (this will likely perform better once we have proper read-only transactions) 
</blockquote></details>

I can accept  that we'll need to make some changes to deal with performance, but at this point I don't see how we can do anything other than option 1 while still preserving immutability.  Lazy loading mutates the state of an object after the object has been exposed to user code.

If we need to defer loading of this data, it needs to not be part of the object's immutable state.  We can do this by marking any such fields with the Immutable.Insignificant annotation.  Although if we're dealing with potentially unbounded collections here, I have to wonder if we shouldn't be processing the result set without storing it (effectively streaming).

Would either of these options work?
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 263 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHw-YAqsMMjdd8zv0:-MZmCKS_6WS1Wppnn9Q4:bfolrjf) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L263)):*

How would we potentially fix this in a post-migration world? Maybe storing the Set<VKey<>> in an array field like we do for some sets of strings? 

Either way, in the general case, this seems like it could potentially cause problems down the road very easily if we always initialize all fields and detach all entities. 
</blockquote></details>

Honestly, for nsHosts we don't need to.  This was the first complication we dealt with and we benchmarked it and the performance of an eager load was found to be adequate.

In general, post-migration we may want to entirely rethink the way we do model objects.  We've been running into this problem since day one: JPA's view of proxy objects is not compatible with immutability.  We should choose one or the other.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 12 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @override)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 647 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOviBW5O8Mg9iYQG9v:-MZrsgvQEPee6Gkd8Vo8:bo6apac) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L647)):*

It's very annoying that Optional doesn't have a `peek` method akin to that of the Stream 
</blockquote></details>

Indeed.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 647 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHK1a2_N0L7mnXeu5:-MZrsitG8pNdA2ZzfJcM:b-lnu6mi) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L647)):*

Why not return an `Optional`?
</blockquote></details>

Because all of the places that we use this are internal within the class and either null-friendly or non-nullable.  That is to say, we'd just end up adding "orElse(null)" or "get()" at the call-sites. 
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 112 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZOwCK18SKJjaIUg9Op:-MZruW3Z2Mq1b2d42aPo:ba7wid6) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L112)):*

why use `foo` instead of the more aptly-named `postLoad`? 
</blockquote></details>

You mean the `entityPostLoad` variable?  Well, it's a bit different because `foo` is non-transient, so modifying it would normally trigger the update logic.  But as I look at this again, we should probably verify that 1) the load callback is called 2) the pre-update callback isn't and 3) the non-transient value has been modified exactly once.  Also, it occurs to me that `nonTransientField` is probably a better name for this than `foo`.

mindhog (commented):
PTAL.  Some notes:

- The latest commit fixes the problem with reloading (and as such, detaching) objects that have been inserted/updated by removing the offending (and unnecessary) test code and throwing an exception when the condition is discovered.
- As per discussion, we're continuing to force eager load for the fields Ben brought up, as these are fields in the DomainHistory object and we don't think they're likely to affect performance.
- I haven't attempted to resolve why the "Eager" annotation isn't working here: If we really want to understand why it is happening, I'll take it up out of band but I have a feeling that the result will be that I burn a day or two on it and discover that Hibernate is WAI.  I suspect that Hibernate is treating the "Eager" semantics as an eager fetch from the database but is still only populating the datastructure on demand, which makes sense from a performance perspective and suggests that we should both annotate fields as "Eager" and run `Hibernate.initialize()` on the collection :-/

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116#-:-M_MqCVmCAL7Ibe5ypVZ:b9o7x4)* status: 7 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @override)

mindhog (commented):
Actually, hold off on reviewing.  I'm adding some tests and changing the exception type.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116#-:-M_QdyDL2_fipbQPimrR:b-ldt3v6)* status: 7 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @override)

mindhog (commented):
Glad I added the tests, my check was broken for merge.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116#-:-M_RQBPQ1lzY3YKOjMVP:b-67ls5u)* status: 6 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @override)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 6 of 14 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 433 at r3](https://reviewable.io/reviews/google/nomulus/1116#-M_W6VcqDxn03BHQ1sxr:-M_W6VcqDxn03BHQ1sxs:b-wjam7r) ([raw file](https://github.com/google/nomulus/blob/e9386016b17b1e8850e8ffc282aa1256db7ea311/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L433)):*
> ```Java
>       throw new NoSuchElementException(key.toString());
>     }
>     return detach(result);
> ```

I thought our conclusion was that we should only detach the loaded objects on commit (instead of immediately)? That way we can have the lazy loading / population of fields work as you'd expect them to.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 6 of 14 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 433 at r3](https://reviewable.io/reviews/google/nomulus/1116#-M_W6VcqDxn03BHQ1sxr:-M_W8FX71idfmBHaFQvQ:b-5wpbfc) ([raw file](https://github.com/google/nomulus/blob/e9386016b17b1e8850e8ffc282aa1256db7ea311/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L433)):*

I thought our conclusion was that we should only detach the loaded objects on commit (instead of immediately)? That way we can have the lazy loading / population of fields work as you'd expect them to. 
</blockquote></details>

That wasn't the understanding that I walked away with.  Lazy load is incompatible with immutability, we simply can't do lazy load.  The work-around I've described is to load things that we'd like to load lazily through secondary queries.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 6 of 14 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 433 at r3](https://reviewable.io/reviews/google/nomulus/1116#-M_W6VcqDxn03BHQ1sxr:-M_WAeLx0pxm7klmxMRc:b9tgm7z) ([raw file](https://github.com/google/nomulus/blob/e9386016b17b1e8850e8ffc282aa1256db7ea311/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L433)):*

That wasn't the understanding that I walked away with.  Lazy load is incompatible with immutability, we simply can't do lazy load.  The work-around I've described is to load things that we'd like to load lazily through secondary queries.
</blockquote></details>

Right sorry, my brain is a bit fried this morning sorry. 

Do we have a tracking bug / list of the fields that will need to be converted to a secondary query? I'm thinking of fields that might be potentially unbounded in length, or ones that we'll need infrequently (among those converted in this PR, I mean)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 6 of 14 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 433 at r3](https://reviewable.io/reviews/google/nomulus/1116#-M_W6VcqDxn03BHQ1sxr:-M_Wr3zA45qxA3LEe_fB:bu5ywu4) ([raw file](https://github.com/google/nomulus/blob/e9386016b17b1e8850e8ffc282aa1256db7ea311/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L433)):*

Right sorry, my brain is a bit fried this morning sorry. 

Do we have a tracking bug / list of the fields that will need to be converted to a secondary query? I'm thinking of fields that might be potentially unbounded in length, or ones that we'll need infrequently (among those converted in this PR, I mean) 
</blockquote></details>

Actually, at this point we have not identified any such cases.  None of our tests have revealed any cases of this - it's possible there are objects that we are only loading in queries (which are not yet detaching) but in general we load by keys in tests.  The only case that comes to mind is the one @rachelguan encountered (reserved list entries from the reserved list, IIRC?) which I suppose we're dealing with through the DAO.

Within this PR, we discussed doing that for some of the history fields, but it turned out we didn't care since these were actually bounded and in an object that is not loaded frequently.

gbrodman (changes requested):
> Reviewed 6 of 8 files at r3, 8 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: all files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 555 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gA8BD-z7jC7_o3jt3:-M_gA8BEE5EPaKE2nXNn:bkcdudc) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L555)):*
> ```Java
>     Recurring recurring = tm().loadByKey(domain.getAutorenewBillingEvent());
>     tm().put(recurring.asBuilder().setRecurrenceEndTime(newEndTime).build());
>     return recurring;
> ```

Shouldn't we be returning the new, updated-end-time object? 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 433 at r3](https://reviewable.io/reviews/google/nomulus/1116#-M_W6VcqDxn03BHQ1sxr:-M_gAV8H3KVMyO44-ruL:bjcx6qo) ([raw file](https://github.com/google/nomulus/blob/e9386016b17b1e8850e8ffc282aa1256db7ea311/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L433)):*

Actually, at this point we have not identified any such cases.  None of our tests have revealed any cases of this - it's possible there are objects that we are only loading in queries (which are not yet detaching) but in general we load by keys in tests.  The only case that comes to mind is the one @rachelguan encountered (reserved list entries from the reserved list, IIRC?) which I suppose we're dealing with through the DAO.

Within this PR, we discussed doing that for some of the history fields, but it turned out we didn't care since these were actually bounded and in an object that is not loaded frequently.
</blockquote></details>

Fair enough -- we can handle this problem when / if it comes up then.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 474 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gAals6fMfh30Zx-B_:-M_gAals6fMfh30Zx-Ba:bdj5lfx) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L474)):*
> ```Java
>     return getEntityManager()
>         .createQuery(
>             String.format("SELECT entity FROM %s entity", getEntityType(clazz).getName()), clazz)
> ```

we don't have to change this here, but actually we could just use the statement 

`.createQuery(String.format("FROM %s", getEntityType(clazz).getName()), clazz)`

i.e. the SELECT entity part is unnecessary 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 652 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gCsav5t5cYKgHADrD:-M_gCsav5t5cYKgHADrE:be9gdzu) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L652)):*
> ```Java
> 
>       // If the entity was previously persisted or merged, we have to throw an exception.
>       if (transactionInfo.get().objectsToSave.contains(entity)) {
> ```

nit: I think this is the only place where objectsToSave is actually used; should we make this "question" a  method on TransactionInfo?

this could be instead something like 

`if (transactionInfo().get().willPersist(entity)) {`
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 671 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gD7Vt8tTLekR7xiJh:-M_gD7Vt8tTLekR7xiJi:b-5cn0gc) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L671)):*
> ```Java
>     // The set of entity objects that have been either persisted (via insert()) or merged (via
>     // put()/update()).  If the entity manager returns these as a result of a find() or query
>     // operation, we must not detach them but instead must clone them -- detaching removes them from
> ```

Is this comment up to date? Seems like we detach them regardless in the load methods. 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 689 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gBuSn40IL280-plWI:-M_gBuSn40IL280-plWJ:b-dcirbt) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L689)):*
> ```Java
>       transactionTime = null;
>       contentsBuilder = null;
>       objectsToSave = Collections.newSetFromMap(new IdentityHashMap<Object, Boolean>());
> ```

gross that there's no simple creator :( 
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 258 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gC50H9gX32A8xErz7:-M_gC50H9gX32A8xErz8:b-oa46ft) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L258)):*
> ```Java
>    * Deletes the given entity from the database.
>    *
>    * <p>This returns the deleted entity, which may not necessarily be the same as the original
> ```

This comment is a bit confusing, from the reader's perspective, why would the deleted entity not be the one passed in? 
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 176 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gBG8PDtHllN_vCPsF:-M_gBG8PDtHllN_vCPsG:b-mjpb6s) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L176)):*
> ```Java
> 
>     @Transient int entityPostLoad = 0;
>     @Transient int entityPreUpdate = 0;
> ```

We should probably have a test verifying that this actually gets incremented on updates, right? 
___
*[core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java, line 480 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gCFsZClEI20XxVGu3:-M_gCFsZClEI20XxVGu4:btfetap) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java#L480)):*
> ```Java
>                       .loadByEntitiesIfPresent(
>                           ImmutableList.of(theEntity, new TestEntity("does-not-exist", "bar")));
>               assertThat(results).isEqualTo(ImmutableList.of(theEntity));
> ```

nit: `containsExactly`, same w/r/t the other list checks in this class
___
*[core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java, line 543 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gCStO-U3O8pJ4bM2I:-M_gCStO-U3O8pJ4bM2J:bcmptyc) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java#L543)):*
> ```Java
> 
>   @Test
>   void loadAfterInsert_fails() {
> ```

For these three test methods let's also test the exception message thrown since it can be inlined fairly easily, `assertThat(assertThrows(...)).hasMessageThat().isEqualTo(...)`

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 5 of 16 files reviewed, 13 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 555 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gA8BD-z7jC7_o3jt3:-M_gJyQn-QaW4Ub_G2p_:bqzrhhj) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L555)):*

Shouldn't we be returning the new, updated-end-time object? 
</blockquote></details>

It may actually need to return what's still in the DB to be compatible with the rest of it.  But the comment then is still wrong.
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 147 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gLEauEd5VHKyF45Iy:-M_gLEauEd5VHKyF45Iz:bqsvzol) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/model/domain/DomainBase.java#L147)):*
> ```Java
>   protected void postLoad() {
>     super.postLoad();
>     // TODO(b/188044616): Determine why Eager loading doesn't work here.
> ```

Too bad this is still unresolved.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 5 of 16 files reviewed, 14 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-M_gLaBl6UHI3KuJzB4H:b-r163o7) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

I can accept  that we'll need to make some changes to deal with performance, but at this point I don't see how we can do anything other than option 1 while still preserving immutability.  Lazy loading mutates the state of an object after the object has been exposed to user code.

If we need to defer loading of this data, it needs to not be part of the object's immutable state.  We can do this by marking any such fields with the Immutable.Insignificant annotation.  Although if we're dealing with potentially unbounded collections here, I have to wonder if we shouldn't be processing the result set without storing it (effectively streaming).

Would either of these options work?
</blockquote></details>

Hadn't we decided we were going to use a getter to handle these so these wouldn't be loaded every time a domain is?
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 324 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gMmRg1idBKWWjK7Wq:-M_gMmRg1idBKWWjK7Wr:bahsf40) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L324)):*
> ```Java
>     creationTime = lastUpdateTime;
>     // TODO(b/188044616): Determine why Eager loading doesn't work here.
>     Hibernate.initialize(labelsToPrices);
> ```

This kind of destroys the entire way that premium lists work (see `PremiumListSqlDao.getPremiumPrice()`).  The intent of the way premium lists is architected is that a lightweight `PremiumList` entity is persisted, and is only loaded occasionally to verify what the latest revision is.  Beyond that, a premium price check occurs by checking for the existence of a row in the `PremiumEntry` table by composite primary key of `revisionId` and `label`.

But this change does an end-run around all of that, and loads up all of an entire revision from the `PremiumEntry` table into memory, without even using it!  Note, premium lists can be large (we've contemplated hundreds of thousands of entries), and aren't guaranteed to fit into memory.  You definitely don't want to have to load all of that just to check the revision or whateve.r

So, like the pricey fields on `Domain` that we discussed, this also should be a getter, and shouldn't be loaded alongside the `PremiumList` entity itself.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 5 of 16 files reviewed, 16 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gNtrQ6Lo3RfojPboZ:-M_gNtrQ6Lo3RfojPbo_:b-lt1lx5) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L90)):*
> ```Java
>   void postLoad() {
>     // TODO(b/188044616): Determine why Eager loading doesn't work here.
>     Hibernate.initialize(reservedListMap);
> ```

Same comment as from premium lists; this isn't how the domain label lists work.
___
*[core/src/main/java/google/registry/model/tmch/ClaimsListShard.java, line 146 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gO6yZCSHB4WFeka8g:-M_gO6yZCSHB4WFeka8h:b-rvd547) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/tmch/ClaimsListShard.java#L146)):*
> ```Java
>   private void postLoad() {
>     // TODO(b/188044616): Determine why Eager loading doesn't work here.
>     Hibernate.initialize(labelsToKeys);
> ```

Again, same.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 258 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gC50H9gX32A8xErz7:-M_gP5Ql-ZmSBBEnXMA3:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L258)):*

This comment is a bit confusing, from the reader's perspective, why would the deleted entity not be the one passed in? 
</blockquote></details>

Ditto.
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 77 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gPB4M9zkIm0eC-TsK:-M_gPB4M9zkIm0eC-TsL:baj9gaf) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L77)):*
> ```Java
>                 () -> {
>                   TestEntity removed = jpaTm().loadByKey(VKey.createSql(TestEntity.class, "id"));
>                   removed = jpaTm().delete(removed);
> ```

Why not just `return jpaTm().delete(removed);` ?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: 5 of 16 files reviewed, 15 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @rachelguan)

___
*[core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java, line 555 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gA8BD-z7jC7_o3jt3:-M_gFsAjE0AUvlP5VvUI:b-2dt474) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L555)):*

Shouldn't we be returning the new, updated-end-time object? 
</blockquote></details>

Yeah, we should -- though it doesn't much matter in practice as all we use it for is the cost.  But done.
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 147 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gLEauEd5VHKyF45Iy:-M_gQmod2p12dIAC0WCg:b-h9x6ad) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/model/domain/DomainBase.java#L147)):*

Too bad this is still unresolved.
</blockquote></details>

Yeah, I've added my comments on the bug.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 262 at r1](https://reviewable.io/reviews/google/nomulus/1116#-MZTHeXA7okyRwLX1yvQ:-M_gR1ke63hxfbDCjVQ8:b-t8lhd2) ([raw file](https://github.com/google/nomulus/blob/6d968dbbd1ec5028cfdf4995af54090727998b01/core/src/main/java/google/registry/model/domain/DomainHistory.java#L262)):*

Hadn't we decided we were going to use a getter to handle these so these wouldn't be loaded every time a domain is?
</blockquote></details>

No, we decided that we didn't need to do this since this is a _DomainHistory_ object and you pointed out that we don't load these often :-)
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 324 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gMmRg1idBKWWjK7Wq:-M_gRLDNDiC2UpOsrg9I:b6b3ab) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L324)):*

This kind of destroys the entire way that premium lists work (see `PremiumListSqlDao.getPremiumPrice()`).  The intent of the way premium lists is architected is that a lightweight `PremiumList` entity is persisted, and is only loaded occasionally to verify what the latest revision is.  Beyond that, a premium price check occurs by checking for the existence of a row in the `PremiumEntry` table by composite primary key of `revisionId` and `label`.

But this change does an end-run around all of that, and loads up all of an entire revision from the `PremiumEntry` table into memory, without even using it!  Note, premium lists can be large (we've contemplated hundreds of thousands of entries), and aren't guaranteed to fit into memory.  You definitely don't want to have to load all of that just to check the revision or whateve.r

So, like the pricey fields on `Domain` that we discussed, this also should be a getter, and shouldn't be loaded alongside the `PremiumList` entity itself.
</blockquote></details>

Yes, I had noticed this in relation to some work @rachelguan was doing (she ran into the lazy load exception).  I'll fix these Monday morning.
___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gNtrQ6Lo3RfojPboZ:-M_gS3uS2ZRQp5mmWMB9:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L90)):*

Same comment as from premium lists; this isn't how the domain label lists work.
</blockquote></details>

Ack.
___
*[core/src/main/java/google/registry/model/tmch/ClaimsListShard.java, line 146 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gO6yZCSHB4WFeka8g:-M_gS5p9AGrlXDEQL9qv:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/tmch/ClaimsListShard.java#L146)):*

Again, same.
</blockquote></details>

Ack.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 474 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gAals6fMfh30Zx-B_:-M_gLkkQ-ZYYqPm0Mk_Y:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L474)):*

we don't have to change this here, but actually we could just use the statement 

`.createQuery(String.format("FROM %s", getEntityType(clazz).getName()), clazz)`

i.e. the SELECT entity part is unnecessary 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 652 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gCsav5t5cYKgHADrD:-M_gMJngB-myEMWNQHAq:b-iq1nxd) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L652)):*

nit: I think this is the only place where objectsToSave is actually used; should we make this "question" a  method on TransactionInfo?

this could be instead something like 

`if (transactionInfo().get().willPersist(entity)) {`
</blockquote></details>

Yeah, I dithered on that point.  But done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 671 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gD7Vt8tTLekR7xiJh:-M_gMlaR9DMsfZavmaul:b-ubb87g) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L671)):*

Is this comment up to date? Seems like we detach them regardless in the load methods. 
</blockquote></details>

Good catch, thanks.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 258 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gC50H9gX32A8xErz7:-M_gNlXKADfYGGtsBWDQ:b-ue1d0j) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L258)):*

This comment is a bit confusing, from the reader's perspective, why would the deleted entity not be the one passed in? 
</blockquote></details>

Expanded on the comment.
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 176 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gBG8PDtHllN_vCPsF:-M_gOMHpEVqxLs4NYe2u:b-r3xooi) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L176)):*

We should probably have a test verifying that this actually gets incremented on updates, right? 
</blockquote></details>

We're verifying that pre-update gets called elsewhere, this variable was just added to make sure that it _doesn't_ get called. :-)
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 77 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gPB4M9zkIm0eC-TsK:-M_gSDrB1PilxQblRZ4O:bzdo2ee) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L77)):*

Why not just `return jpaTm().delete(removed);` ?
</blockquote></details>

I believe I was printing it out at one point and forgot to reconsolidate: done.
___
*[core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java, line 480 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gCFsZClEI20XxVGu3:-M_gP-yVDJIu7RUb9KQr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java#L480)):*

nit: `containsExactly`, same w/r/t the other list checks in this class
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java, line 543 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gCStO-U3O8pJ4bM2I:-M_gQZ-X2HG45y01ViGO:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java#L543)):*

For these three test methods let's also test the exception message thrown since it can be inlined fairly easily, `assertThat(assertThrows(...)).hasMessageThat().isEqualTo(...)`
</blockquote></details>

Done

gbrodman (changes requested):
> Reviewed 10 of 11 files at r5, 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 324 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gMmRg1idBKWWjK7Wq:-MaETvV48t1kVByFduF-:b-7z908s) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L324)):*

Yes, I had noticed this in relation to some work @rachelguan was doing (she ran into the lazy load exception).  I'll fix these Monday morning.
</blockquote></details>

For now can we remove this, then? I know you're working on it in the other PR(s) 
___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gNtrQ6Lo3RfojPboZ:-MaETza-BldStEudVV8B:b-jty4vg) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L90)):*

Ack.
</blockquote></details>

ditto other comment
___
*[core/src/main/java/google/registry/model/tmch/ClaimsListShard.java, line 146 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gO6yZCSHB4WFeka8g:-MaEU-cc1aQuFujm1uGo:b-jty4vg) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/tmch/ClaimsListShard.java#L146)):*

Ack.
</blockquote></details>

ditto other comment
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 790 at r6](https://reviewable.io/reviews/google/nomulus/1116#-MaEUEFpDxEy4ygSSPft:-MaEUEFpDxEy4ygSSPfu:b-pydn2v) ([raw file](https://github.com/google/nomulus/blob/6bb1884dacd73926303277cfccc07d233387b49b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L790)):*
> ```Java
>       }
>       TypedQuery<T> query = buildQuery();
>       if (query instanceof org.hibernate.query.Query) {
> ```

What's the reasoning behind this branching? 
___
*[core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java, line 176 at r4](https://reviewable.io/reviews/google/nomulus/1116#-M_gBG8PDtHllN_vCPsF:-MaEUmFM-3_j_-6ionrE:b-p79ll5) ([raw file](https://github.com/google/nomulus/blob/8161405ac42c9a8ee23d91da01960a8aac196a8b/core/src/test/java/google/registry/persistence/EntityCallbacksListenerTest.java#L176)):*

We're verifying that pre-update gets called elsewhere, this variable was just added to make sure that it _doesn't_ get called. :-)
</blockquote></details>

Right, I see that (it's a bit indirect in the other code)

mindhog (commented):
With the latest commit, this change will likely fail tests until #1177 is merged.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116#-:-MaU-m-_-T_ytWaYFnHB:b8et6k0)* status: 10 of 17 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, @rachelguan, and @weiminyu)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 324 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gMmRg1idBKWWjK7Wq:-MaTVO7r7n311UA4Lr7M:b-896fix) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L324)):*

For now can we remove this, then? I know you're working on it in the other PR(s) 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/registry/label/ReservedList.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gNtrQ6Lo3RfojPboZ:-MaTVR8P7KM9B9dHiWw6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/registry/label/ReservedList.java#L90)):*

ditto other comment
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/tmch/ClaimsListShard.java, line 146 at r5](https://reviewable.io/reviews/google/nomulus/1116#-M_gO6yZCSHB4WFeka8g:-MaTVS3QDiWCESS3Onfa:b-896fix) ([raw file](https://github.com/google/nomulus/blob/eb52cf584e1a13d549e9526078166b4920fc4166/core/src/main/java/google/registry/model/tmch/ClaimsListShard.java#L146)):*

ditto other comment
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 790 at r6](https://reviewable.io/reviews/google/nomulus/1116#-MaEUEFpDxEy4ygSSPft:-MaTWq1x921JMuwi0IQr:b-t1tjo6) ([raw file](https://github.com/google/nomulus/blob/6bb1884dacd73926303277cfccc07d233387b49b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L790)):*

What's the reasoning behind this branching? 
</blockquote></details>

@weiminyu added this in #1155, presumably because JPA's' `TypedQuery` doesn't support `setFetchSize()`.

You're seeing it here because of the rebase.

gbrodman (approved):
> Reviewed 7 of 7 files at r7.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 790 at r6](https://reviewable.io/reviews/google/nomulus/1116#-MaEUEFpDxEy4ygSSPft:-MaUQxrU7RGU3g2QFoJO:b-nu55of) ([raw file](https://github.com/google/nomulus/blob/6bb1884dacd73926303277cfccc07d233387b49b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L790)):*

@weiminyu added this in #1155, presumably because JPA's' `TypedQuery` doesn't support `setFetchSize()`.

You're seeing it here because of the rebase. 
</blockquote></details>

ooooh super weird, all right then

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1116)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1119, opened 2021-04-30
Remove Pipeline as a field in pipeline classes

jianglai:
In tests we use a TestPipelineExtension which does some static
initialization that should not be repeated the same JVM. In our
XXXPipeline classes we save the pipeline as a field and usually write lambdas
that are pass to the pipeline. Because lambdas are effectively anonymous inner
classes they are bound to their enclosing instances. When they get serialized
during pipeline execution, their enclosing classes also do. This might result
in undefined behavior when multiple lambdas in the same XXXPipeline are used
on the same JVM (such as in tests) where the static initialization may be done
multiple times if different class loaders are used. This is very
unlikely to happen but as a best practice we still remove them as
fields.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1119)

### DIFF
--- a/core/src/main/java/google/registry/beam/datastore/BulkDeleteDatastorePipeline.java
+++ b/core/src/main/java/google/registry/beam/datastore/BulkDeleteDatastorePipeline.java
@@ -87,20 +87,18 @@ public class BulkDeleteDatastorePipeline {
 
   private final BulkDeletePipelineOptions options;
 
-  private final Pipeline pipeline;
-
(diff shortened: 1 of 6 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1119)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1126, opened 2021-05-04
Combine the two Lock classes into one class

gbrodman:
This allows us to remove the DAO and to just treat locks the same as we
would treat any other object -- generically grabbing them from the
transaction manager.

We do not need to be concerned about the changeover between Datastore
and SQL because we assume that any such changeover will require
sufficient downtime that any currently-valid acquired locks will expire
during the downtime. Otherwise, we could get into a situation where an
action has acquired a particular lock in Datastore but not SQL.


mindhog (approved):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1126)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1126)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 144 at r1](https://reviewable.io/reviews/google/nomulus/1126#-M_6gXhD4klr2DAWCNlP:-M_6gXhD4klr2DAWCNlQ:b-e4hg4k) ([raw file](https://github.com/google/nomulus/blob/bd8b5dc7225224e3b7dbd440b72372ce2473ceba/core/src/main/java/google/registry/model/server/Lock.java#L144)):*
> ```Java
>   }
> 
>   private static String makeLockId(String resourceName, String tld) {
> ```

Something seems wrong here.  Either the `tld` parameter needs to be `@Nullable` or it needs to be renamed if it always has a value (with global being a possible value).

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1126)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 207 at r1](https://reviewable.io/reviews/google/nomulus/1126#-M_6h0Dc9vMR9G6PJ5Cv:-M_6h0Dd0rH1lsiXQ7V0:b4vz397) ([raw file](https://github.com/google/nomulus/blob/bd8b5dc7225224e3b7dbd440b72372ce2473ceba/core/src/main/java/google/registry/model/server/Lock.java#L207)):*
> ```Java
>       RequestStatusChecker requestStatusChecker,
>       boolean checkThreadRunning) {
>     String namespace = tld != null ? tld : GLOBAL;
> ```

Wrap the conditional clause in parentheses for clarity.
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 207 at r1](https://reviewable.io/reviews/google/nomulus/1126#-M_6hXlu0WNnmoZG6YVa:-M_6hXlu0WNnmoZG6YVb:b1cqjt6) ([raw file](https://github.com/google/nomulus/blob/bd8b5dc7225224e3b7dbd440b72372ce2473ceba/core/src/main/java/google/registry/model/server/Lock.java#L207)):*
> ```Java
>       RequestStatusChecker requestStatusChecker,
>       boolean checkThreadRunning) {
>     String namespace = tld != null ? tld : GLOBAL;
> ```

The word we've used for this in the past has been `scope`.
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 301 at r1](https://reviewable.io/reviews/google/nomulus/1126#-M_6hjZ2A4dqMwFgx_3J:-M_6hjZ2A4dqMwFgx_3K:buob0vh) ([raw file](https://github.com/google/nomulus/blob/bd8b5dc7225224e3b7dbd440b72372ce2473ceba/core/src/main/java/google/registry/model/server/Lock.java#L301)):*
> ```Java
>     String resourceName;
> 
>     String tld;
> ```

Is this nullable?

mindhog (approved):
> Reviewed 16 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1126)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 207 at r1](https://reviewable.io/reviews/google/nomulus/1126#-M_6h0Dc9vMR9G6PJ5Cv:-M_LdtNn3ptMvIMOCIJL:bmog0si) ([raw file](https://github.com/google/nomulus/blob/bd8b5dc7225224e3b7dbd440b72372ce2473ceba/core/src/main/java/google/registry/model/server/Lock.java#L207)):*

Done.
</blockquote></details>

I think the style checker will flag this with a warning.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1126)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[db/src/main/resources/sql/flyway/V94__rename_lock_scope.sql, line 15 at r2](https://reviewable.io/reviews/google/nomulus/1126#-M_N5cVz2OdBxKB044Gj:-M_N5cVz2OdBxKB044Gk:b-cbhml5) ([raw file](https://github.com/google/nomulus/blob/f8d087e3b556ae68efde4ad8215f732cedf8090e/db/src/main/resources/sql/flyway/V94__rename_lock_scope.sql#L15)):*
> ```TSQL
> 
> ALTER TABLE IF EXISTS "Lock" RENAME tld TO scope;
> ```

This would all be a lot cleaner if, instead of using the magic sentinel value `GLOBAL`, we had a nullable `tld` field, with the null one representing a lock that is not scoped to a TLD.  That would make better sense.  Unfortunately, I believe that we tried that already, with the resultant discovery that Hibernate does not allow a nullable column to be using as part of a compound primary key.  Oh well.  This alternate solution isn't great but I suppose we don't have a choice.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1126)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[db/src/main/resources/sql/flyway/V94__rename_lock_scope.sql, line 15 at r2](https://reviewable.io/reviews/google/nomulus/1126#-M_N5cVz2OdBxKB044Gj:-M_NwXNT7CfHtjbLOM2_:b6gm5c3) ([raw file](https://github.com/google/nomulus/blob/f8d087e3b556ae68efde4ad8215f732cedf8090e/db/src/main/resources/sql/flyway/V94__rename_lock_scope.sql#L15)):*

This would all be a lot cleaner if, instead of using the magic sentinel value `GLOBAL`, we had a nullable `tld` field, with the null one representing a lock that is not scoped to a TLD.  That would make better sense.  Unfortunately, I believe that we tried that already, with the resultant discovery that Hibernate does not allow a nullable column to be using as part of a compound primary key.  Oh well.  This alternate solution isn't great but I suppose we don't have a choice.
</blockquote></details>

Yep, unfortunately (my search history is littered with `composite primary key null value`). I added a comment in the lock class mentioning this

PR #1130, opened 2021-05-05
Re-enable tests in RC build

jianglai:
There has been a case where the CI was broken on Friday and no one
noticied or fixed it and a RC build was built with broken tests.
The tests were disabled due to unknown test failures that have since
been fixed.

Also update the machine type used by GCB to be more powerful. This is
necessary for the tests to past because N1_HIGHCPU_8 is RAM constraint
and the tests crashes. I updated all jobs to use the new type which
hopefully will make the build faster as well.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1130)

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -1,4 +1,5 @@
 python/
+node_modules/
 **/build/
 **/out/
 .*/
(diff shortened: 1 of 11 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1130)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1131, opened 2021-05-05
Update xmlhttprequest-ssl version

jianglai:
We are using a version with known vulnerability per https://github.com/google/nomulus/security/dependabot/package-lock.json/xmlhttprequest-ssl/open

The `package-lock.json` file is autogenerated. Not sure if it is correct given that it still has v1.5.5 in it...

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1131)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -404,6 +404,12 @@
           "requires": {
             "ms": "2.0.0"
           }
+        },
+        "xmlhttprequest-ssl": {
+          "version": "1.5.5",
+          "resolved": "https://registry.npmjs.org/xmlhttprequest-ssl/-/xmlhttprequest-ssl-1.5.5.tgz",
+          "integrity": "sha1-wodrBhaKrcQOV9l+gRkayPQ5iz4=",
+          "dev": true
         }
       }
     },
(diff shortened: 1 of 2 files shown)


gbrodman:
Yeah I don't think I trust that this actually updated the xmlhttprequest-ssl version. The dependency analysis shows that it comes from Karma, so we probably should try updating the Karma version

gbrodman:
seems like the most recent version of Karma 6.3.2 doesn't have this dependency (tested locally), running the tests to see if everything still passes

jianglai:
> seems like the most recent version of Karma 6.3.2 doesn't have this dependency (tested locally), running the tests to see if everything still passes

Do you know how npm handles diamond dependencies? Does it pick the higher version or what? Also is there an easier way to see the dependency graph other than going through the `package-lock.json` file manually and searching for the string?

gbrodman:
By diamond dependencies do you mean that there is some package A which depends on B and C, and B and C each depend on different versions of D? 

I don't know how it deals with those, though your guess seems like a good one. 

For viewing dependencies, here I just viewed the package-lock.json file but `npm ls` is useful for more complex situations

PR #1133, opened 2021-05-05
Fix flaky Spec11PipelineTest

jianglai:
It appears when the test fails the call to isAppEngine() returns true. It
is still not clear where that comes from, but it is likely due to the
DatastoreEntityExtension trying to setup an App Engine environment for
Ofy. All Ofy needs is a project ID so we explicitly clear the system
proporty for ApiProxy in the hope that isAppEngine() will now always
return false.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1133)

### DIFF
--- a/core/src/test/java/google/registry/beam/TestPipelineExtension.java
+++ b/core/src/test/java/google/registry/beam/TestPipelineExtension.java
@@ -20,6 +20,7 @@
 package google.registry.beam;
 
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.truth.Truth.assertWithMessage;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 
@@ -27,6 +28,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
 import com.google.common.collect.Maps;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -295,6 +297,16 @@ public class TestPipelineExtension extends Pipeline
 
       enableAbandonedNodeEnforcement(true);
     }
+
+    // Clear this property so that when default Guava ThreadFactory is created
+    // it will not think that it is in App Engine and return an unusable
+    // ThreadFactory.
+    System.clearProperty("com.google.appengine.runtime.environment");
+    assertWithMessage(
+            "Beam pipelines don't run in an App Engine environment, and thus"
+                + " the tests shouldn't be mocking one either.")
+        .that(isAppEngine())
+        .isFalse();
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1133)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/1133#-M_1rZ4dBZXelpEIsMsX:-M_1rZ4dBZXelpEIsMsY:bq9cs3e) ([raw file](https://github.com/google/nomulus/blob/b2287a81033de4a986e82fbeaa54ef490c482034/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L45)):*
> ```Java
>     // Clear this property so that when default Guava ThreadFactory is created it will not think
>     // that it is in App Engine and return an unusable ThreadFactory.
>     System.clearProperty("com.google.appengine.runtime.environment");
> ```

If this line fixes the problem, do we need all the changes in the other classes?

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1133)_ status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)
> 
> _[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 45 at r1](https://reviewable.io/reviews/google/nomulus/1133#-M_1rZ4dBZXelpEIsMsX:-M_1rZ4dBZXelpEIsMsY:bq9cs3e) ([raw file](https://github.com/google/nomulus/blob/b2287a81033de4a986e82fbeaa54ef490c482034/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L45)):_
> 
> > ```java
> >     // Clear this property so that when default Guava ThreadFactory is created it will not think
> >     // that it is in App Engine and return an unusable ThreadFactory.
> >     System.clearProperty("com.google.appengine.runtime.environment");
> > ```
> 
> If this line fixes the problem, do we need all the changes in the other classes?

I think it is still a good idea to put the assertion in any calls that use the test pipeline because it is incompatible with app engine thread factory, and we'd want to know about it if there are other settings that result in the same problem.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1133#-:-M_6jgSt9hKp9uFjCm3W:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (approved):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1133)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1136, opened 2021-05-06
Remove deduplication TODO and test in RegistryJpaIo read connector

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1136)

### DIFF
--- a/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java
+++ b/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java
@@ -26,7 +26,6 @@ import google.registry.persistence.transaction.JpaTransactionManager;
 import google.registry.persistence.transaction.QueryComposer;
 import google.registry.persistence.transaction.TransactionManagerFactory;
 import java.io.Serializable;
-import java.util.HashSet;
 import java.util.Objects;
 import java.util.concurrent.ThreadLocalRandom;
 import javax.persistence.criteria.CriteriaQuery;
@@ -94,11 +93,7 @@ public final class RegistryJpaIO {
    * PCollection<String> contactIds =
    *     pipeline
    *         .apply(RegistryJpaIO.read(
-   *             (JpaTransactionManager tm) -> tm.createQueryComposer...,
-   *             contact -> KV.of(contact.getRepoId(), contact.getContactId()))
-   *         .withCoder(KvCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of())))
-   *         .apply(Deduplicate.keyedValues())
-   *         .apply(Values.create());
+   *             (JpaTransactionManager tm) -> tm.createQueryComposer..., ...);
    * }</pre>
    */
   @AutoValue
@@ -180,21 +175,12 @@ public final class RegistryJpaIO {
       @ProcessElement
       public void processElement(OutputReceiver<T> outputReceiver) {
         // TODO(b/187210388): JpaTransactionManager should support non-transactional query.
-        // TODO(weiminyu): add deduplication
-        HashSet<T> outputs = new HashSet<>();
         jpaTm()
             .transactNoRetry(
-                () -> {
-                  querySupplier.apply(jpaTm()).stream()
-                      .map(resultMapper::apply)
-                      .forEach(
-                          output -> {
-                            if (!outputs.contains(output)) {
-                              outputs.add(output);
-                              outputReceiver.output(output);
-                            }
-                          });
-                });
+                () ->
+                    querySupplier.apply(jpaTm()).stream()
+                        .map(resultMapper::apply)
+                        .forEach(outputReceiver::output));
         // TODO(weiminyu): improve performance by reshuffle.
       }
     }
--- a/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java
+++ b/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java
@@ -26,6 +26,7 @@ import google.registry.persistence.transaction.JpaTransactionManager;
 import google.registry.persistence.transaction.QueryComposer;
 import google.registry.persistence.transaction.TransactionManagerFactory;
 import java.io.Serializable;
+import java.util.HashSet;
 import java.util.Objects;
 import java.util.concurrent.ThreadLocalRandom;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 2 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1136)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/common/RegistryJpaIO.java, line 192 at r1](https://reviewable.io/reviews/google/nomulus/1136#-M_6KZhwDGXDCGJAkO1v:-M_6KZhwDGXDCGJAkO1w:bsuei5y) ([raw file](https://github.com/google/nomulus/blob/2bdffda8580c2a3b38a7070fa0e951e5db35b3d4/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L192)):*
<details><summary><i alt="summary">Quoted 14 lines of code…</i></summary>

> ```
>         HashSet<T> outputs = new HashSet<>();
>         jpaTm()
>             .transactNoRetry(
>                 () -> {
>                   querySupplier.apply(jpaTm()).stream()
>                       .map(resultMapper::apply)
>                       .forEach(
>                           output -> {
>                             if (!outputs.contains(output)) {
>                               outputs.add(output);
>                               outputReceiver.output(output);
>                             }
>                           });
>                 });
> ```
</details>

The duplications we care about do not come from the query.

A BEAM pipeline can be viewed as a collection of nodes connected by uni-directional edges. Each node represents some transformation to be applied to the input  to produce the output. If the pipeline encounters a transient failure when applying the function of a node to an input, it may retry the function with that input.

This QueryRunner class which extends DoFn, is a node in the pipeline. It only has one input, the (Void) null created in the previous stage.
If an exception is thrown in the middle of the query execution, the pipeline may retry it, which means the entire query will be executed again, duplicating those entities already processed during the failed first run. That is, the duplicates are detectable in the output PCollection, not in the QueryRunner instance.

To deduplicate the query result, we should move the code in the test into the Read class's expand() method: first apply a transform that map an entity to KV<someKey, entity>, then apply the builtin Dedepulicate transform, and finally apply the Values transform and retrieve the original values.

As a generic utility, the user should decide whether deduplication is needed. They also should provide the key extractor (e.g., the contact::getRepoId method),  and the key coder if necessary.  

I would add two methods to the Read class:
Optional<SerializableFunction<T, Object>> dedupeKeyExtractor();
Optional<Coder<?>> keyCoder();

The Read.Builder class would validate that: 
1. keyCoder().isPresent requires that dedupeKeyExtractor be present.
2. keyCoder.get().isDeterministic == true

If keyCoder.isPresent, append the deduplicate transforms to the current return value in expand.

weiminyu (changes requested):
Actually I 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1136#-:-M_C7u3d2yMElcW2SjXa:bvnw5zr)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/common/RegistryJpaIO.java, line 183 at r1](https://reviewable.io/reviews/google/nomulus/1136#-M_C87q49aZleIHD0u8j:-M_C87q49aZleIHD0u8k:bfo9ap6) ([raw file](https://github.com/google/nomulus/blob/2bdffda8580c2a3b38a7070fa0e951e5db35b3d4/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L183)):*
> ```
> // TODO(weiminyu): add deduplication
> ```
I didn't think it through when I said this. There are no duplicates due to retries of queries. If a transient failure happens, the partial results related to this query will be discarded.

Let's remove this todo line and the *_dedupe method in the test class.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1136)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/beam/common/RegistryJpaReadTest.java, line 81 at r1](https://reviewable.io/reviews/google/nomulus/1136#-M_C9_uD-m5VbiUtqa1y:-M_C9_uD-m5VbiUtqa1z:b-yvmqnv) ([raw file](https://github.com/google/nomulus/blob/2bdffda8580c2a3b38a7070fa0e951e5db35b3d4/core/src/test/java/google/registry/beam/common/RegistryJpaReadTest.java#L81)):*
> ```
> // builder.add(DatabaseHelper.newContactResource("contact_1"));
> ```
Please remove.
___
*[core/src/test/java/google/registry/beam/common/RegistryJpaReadTest.java, line 87 at r1](https://reviewable.io/reviews/google/nomulus/1136#-M_C9lIU4i2Xpdgwj7Nt:-M_C9lIU4i2Xpdgwj7Nu:b-qn0aaj) ([raw file](https://github.com/google/nomulus/blob/2bdffda8580c2a3b38a7070fa0e951e5db35b3d4/core/src/test/java/google/registry/beam/common/RegistryJpaReadTest.java#L87)):*
> ```
> _noDupe
> ```
Please remove the suffix.
___
*[core/src/test/java/google/registry/beam/common/RegistryJpaReadTest.java, line 102 at r1](https://reviewable.io/reviews/google/nomulus/1136#-M_C9uHp5ZS4fcg9PBSm:-M_C9uHq3sSt4WZc5QPo:b-trmbsa) ([raw file](https://github.com/google/nomulus/blob/2bdffda8580c2a3b38a7070fa0e951e5db35b3d4/core/src/test/java/google/registry/beam/common/RegistryJpaReadTest.java#L102)):*
<details><summary><i alt="summary">Quoted 11 lines of code…</i></summary>

> ```
>   @Test
>   void nonTransactionalQuery_withDuplicates() {
>     jpaTm().transact(() -> jpaTm().put(DatabaseHelper.newContactResource("contact_1")));
>     Read<ContactResource, String> read =
>         RegistryJpaIO.read(
>             (JpaTransactionManager jpaTm) -> jpaTm.createQueryComposer(ContactResource.class),
>             ContactBase::getContactId);
>     PCollection<String> repoIds = testPipeline.apply(read);
>     PAssert.that(repoIds).containsInAnyOrder("contact_0", "contact_1", "contact_2");
>     testPipeline.run();
>   }
> ```
</details>

Please remove the entire method.

weiminyu (commented):
Could you take a look at PR 1155 instead? It has these changes along with other fixes.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1136#-:-M_WvWlC2plJz1usrv1g:ba6jfcv)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @weiminyu)

PR #1139, opened 2021-05-07
Update GCL dependency to avoid security alert

gbrodman:
This required a few changes in addition to the dependency update.

- a few transitive / required dependency updates as well
- updating soyutils_usegoog.js and adding checks.js because they're
necessary as part of the Soy compilation process
- Using a trustedResourceUri in the buildSrc Soy compilation instead of
a string
- changing the arguments to the Soy-to-Java compiler to comply with the
new version
- Moving all Soy UI files to be in the registrar directory. This was
not the case before due to previous thinking that we'd have separate
admin and registrar consoles -- this is no longer the case so it's no
longer necessary. This necessitated various refactorings and reference
changes.
  - The new soy-to-javascript compiler requires this, as it removes the
  "deps" param that we were previously using to say "use the general UI
  utils as dependencies for the registrar-console files".
- Creating a SQL environment and loading test data in the test server
main method -- previously, the local test server did not work.
- Fix some JS code that was referencing now-deleted library functions

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1139)


jianglai (approved):
This is a lot of wrangling, thanks for being a good citizen :D

FYI: I think in general we don't get that much value from the local test server (as you have found out that it was not working due to the lack of SQL environment set up and no one noticed). Maybe we should about whether to decommission it, just like the karma tests. Worth bringing it up in the next meeting?

> Reviewed 103 of 103 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1139#-:-M_l1w2j6IK9YoYgittq:b-vpf0tu)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[build.gradle, line 463 at r1](https://reviewable.io/reviews/google/nomulus/1139#-M_cSFDz2n7uyWmM9dcU:-M_cSEJ516WRPb1qC8K1:b-ok3tsg) ([raw file](https://github.com/google/nomulus/blob/fe5ec77e6982f59155c9aeb3b69ce451861d5e76/build.gradle#L463)):*
> ```
> doclint:accessibility,html,syntax', true)
> ```
You are excluding the `reference` type, which I think is valuable. If this results in javadoc failure we should fix the invalid reference in the javadoc instead of disabling the linter.

https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html

gbrodman (commented):
I think it's actually worth keeping around for sure. I found it to be very useful when I was doing any sort of changes to the front end e.g. registry lock or other random fixes 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1139#-:-M_udXp86b_fvkW9ysE5:b-locs1a)* status: 102 of 103 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[build.gradle, line 463 at r1](https://reviewable.io/reviews/google/nomulus/1139#-M_cSFDz2n7uyWmM9dcU:-M_ueGLU1A4iIOvlAodZ:bq948w1) ([raw file](https://github.com/google/nomulus/blob/fe5ec77e6982f59155c9aeb3b69ce451861d5e76/build.gradle#L463)):*

> ```
> doclint:accessibility,html,syntax', true)
> ```
You are excluding the `reference` type, which I think is valuable. If this results in javadoc failure we should fix the invalid reference in the javadoc instead of disabling the linter.

https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javadoc.html
</blockquote></details>

ugh unfortunately the generated SoyInfo Java files fail the "reference" javadoc check which is why I changed this in the first place 

We can get around this by excluding those Java files while altering the classpath to include those *.class files (since other files depend on them, we can't exclude them entirely)

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1139)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1150, opened 2021-05-12
Switch from using raw HistoryEntries to typed subclasses thereof

jianglai:
HistoryEntry is used to record all histories (contact, domain, host) in
Datastore. In SQL it is now split into three subclasses (and thus
tables): ContactHistory, DomainHistory and HostHistory. Its builder is
genericized as a result which led to a lot of compiler warnings for the
use of a raw HistoryEntry in the existing code base.

This PR cleans things up by replacing all the explicit use of
raw HistoryEntry with the corresponding subclass and also adds some
guardrails to prevent the use of raw HistoryEntry accidentally.

Note that because DomainHistory includes nsHosts and gracePeriodHistory,
both of which are assigned a roid from ofy when built, the assigned roids for
resources after history entries are built are incremented compared to
when only HistoryEntrys are built (before this PR) in
RdapDomainSearchActionTest.

Also added a convenient tm().updateAll() varargs method.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1150)

### DIFF
--- a/core/src/main/java/google/registry/batch/DeleteContactsAndHostsAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteContactsAndHostsAction.java
@@ -408,7 +408,9 @@ public class DeleteContactsAndHostsAction implements Runnable {
       } else {
         resourceToSave = resource.asBuilder().removeStatusValue(PENDING_DELETE).build();
       }
-      auditedOfy().save().<ImmutableObject>entities(resourceToSave, historyEntry, pollMessage);
+      auditedOfy()
+          .save()
+          .<ImmutableObject>entities(resourceToSave, historyEntry.asHistoryEntry(), pollMessage);
       return DeletionResult.create(
           deleteAllowed ? Type.DELETED : Type.NOT_DELETED, pollMessageText);
     }
--- a/core/src/main/java/google/registry/batch/DeleteContactsAndHostsAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteContactsAndHostsAction.java
@@ -370,11 +370,10 @@ public class DeleteContactsAndHostsAction implements Runnable {
                       : "it was transferred prior to deletion");
 
       HistoryEntry historyEntry =
-          new HistoryEntry.Builder()
+          HistoryEntry.createBuilderForResource(resource)
               .setClientId(deletionRequest.requestingClientId())
               .setModificationTime(now)
               .setType(getHistoryEntryType(resource, deleteAllowed))
-              .setParent(deletionRequest.key())
               .build();
 
       PollMessage.OneTime pollMessage =
--- a/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
@@ -44,11 +44,11 @@ import google.registry.mapreduce.MapreduceRunner;
 import google.registry.mapreduce.inputs.EppResourceInputs;
 import google.registry.model.EppResourceUtils;
 import google.registry.model.domain.DomainBase;
+import google.registry.model.domain.DomainHistory;
 import google.registry.model.index.EppResourceIndex;
 import google.registry.model.index.ForeignKeyIndex;
 import google.registry.model.registry.Registry;
 import google.registry.model.registry.Registry.TldType;
-import google.registry.model.reporting.HistoryEntry;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
 import google.registry.request.Response;
@@ -253,9 +253,9 @@ public class DeleteProberDataAction implements Runnable {
                         .setDeletionTime(tm().getTransactionTime())
                         .setStatusValues(null)
                         .build();
-                HistoryEntry historyEntry =
-                    new HistoryEntry.Builder()
-                        .setParent(domain)
+                DomainHistory historyEntry =
+                    new DomainHistory.Builder()
+                        .setDomain(domain)
                         .setType(DOMAIN_DELETE)
                         .setModificationTime(tm().getTransactionTime())
                         .setBySuperuser(true)
@@ -267,7 +267,7 @@ public class DeleteProberDataAction implements Runnable {
                 // poll messages, or auto-renews because these will all be hard-deleted the next
                 // time the
                 // mapreduce runs anyway.
-                auditedOfy().save().entities(deletedDomain, historyEntry);
+                tm().updateAll(ImmutableList.of(deletedDomain, historyEntry));
                 updateForeignKeyIndexDeletionTime(deletedDomain);
                 dnsQueue.addDomainRefreshTask(deletedDomain.getDomainName());
               });
--- a/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
@@ -263,11 +263,9 @@ public class DeleteProberDataAction implements Runnable {
                         .setClientId(registryAdminClientId)
                         .build();
                 // Note that we don't bother handling grace periods, billing events, pending
-                // transfers,
-                // poll messages, or auto-renews because these will all be hard-deleted the next
-                // time the
-                // mapreduce runs anyway.
-                tm().updateAll(ImmutableList.of(deletedDomain, historyEntry));
+                // transfers, poll messages, or auto-renews because these will all be hard-deleted
+                // the next time the mapreduce runs anyway.
+                tm().putAll(deletedDomain, historyEntry);
                 updateForeignKeyIndexDeletionTime(deletedDomain);
                 dnsQueue.addDomainRefreshTask(deletedDomain.getDomainName());
               });
--- a/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
+++ b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
@@ -194,8 +194,8 @@ public class ExpandRecurringBillingEventsAction implements Runnable {
                                 .setBySuperuser(false)
                                 .setClientId(recurring.getClientId())
                                 .setModificationTime(tm().getTransactionTime())
-                                // TODO: modify this to use setDomain instead when converting
-                                // this action to be SQL-aware.
+                                // TODO (jianglai): modify this to use setDomain instead when
+                                //  converting this action to be SQL-aware.
                                 .setDomainRepoId(domainKey.getName())
                                 .setPeriod(Period.create(1, YEARS))
                                 .setReason(
--- a/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
+++ b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
@@ -49,6 +49,7 @@ import google.registry.model.billing.BillingEvent.OneTime;
 import google.registry.model.billing.BillingEvent.Recurring;
 import google.registry.model.common.Cursor;
 import google.registry.model.domain.DomainBase;
+import google.registry.model.domain.DomainHistory;
 import google.registry.model.domain.Period;
 import google.registry.model.registry.Registry;
 import google.registry.model.reporting.DomainTransactionRecord;
@@ -188,12 +189,14 @@ public class ExpandRecurringBillingEventsAction implements Runnable {
                       // an event persisted.
                       for (DateTime billingTime : difference(billingTimes, existingBillingTimes)) {
                         // Construct a new HistoryEntry that parents over the OneTime
-                        HistoryEntry historyEntry =
-                            new HistoryEntry.Builder()
+                        DomainHistory historyEntry =
+                            new DomainHistory.Builder()
                                 .setBySuperuser(false)
                                 .setClientId(recurring.getClientId())
                                 .setModificationTime(tm().getTransactionTime())
-                                .setParent(domainKey)
+                                // TODO: modify this to use setDomain instead when converting
+                                // this action to be SQL-aware.
+                                .setDomainRepoId(domainKey.getName())
                                 .setPeriod(Period.create(1, YEARS))
                                 .setReason(
                                     "Domain autorenewal by ExpandRecurringBillingEventsAction")
--- a/core/src/main/java/google/registry/flows/FlowModule.java
+++ b/core/src/main/java/google/registry/flows/FlowModule.java
@@ -213,13 +213,14 @@ public class FlowModule {
     return Strings.nullToEmpty(((Poll) eppInput.getCommandWrapper().getCommand()).getMessageId());
   }
 
-  private static HistoryEntry.Builder<? extends HistoryEntry, ?> makeHistoryEntryBuilder(
-      HistoryEntry.Builder<? extends HistoryEntry, ?> builder,
-      Trid trid,
-      byte[] inputXmlBytes,
-      boolean isSuperuser,
-      String clientId,
-      EppInput eppInput) {
+  private static <B extends HistoryEntry.Builder<? extends HistoryEntry, ?>>
+      B makeHistoryEntryBuilder(
+          B builder,
+          Trid trid,
+          byte[] inputXmlBytes,
+          boolean isSuperuser,
+          String clientId,
+          EppInput eppInput) {
     builder
         .setTrid(trid)
         .setXmlBytes(inputXmlBytes)
--- a/core/src/main/java/google/registry/flows/FlowModule.java
+++ b/core/src/main/java/google/registry/flows/FlowModule.java
@@ -249,9 +249,8 @@ public class FlowModule {
       @Superuser boolean isSuperuser,
       @ClientId String clientId,
       EppInput eppInput) {
-    return (ContactHistory.Builder)
-        makeHistoryEntryBuilder(
-            new ContactHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
+    return makeHistoryEntryBuilder(
+        new ContactHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
   }
 
   /**
@@ -267,9 +266,8 @@ public class FlowModule {
       @Superuser boolean isSuperuser,
       @ClientId String clientId,
       EppInput eppInput) {
-    return (HostHistory.Builder)
-        makeHistoryEntryBuilder(
-            new HostHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
+    return makeHistoryEntryBuilder(
+        new HostHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
   }
 
   /**
@@ -285,9 +283,8 @@ public class FlowModule {
       @Superuser boolean isSuperuser,
       @ClientId String clientId,
       EppInput eppInput) {
-    return (DomainHistory.Builder)
-        makeHistoryEntryBuilder(
-            new DomainHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
+    return makeHistoryEntryBuilder(
+        new DomainHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
   }
 
   /**
--- a/core/src/main/java/google/registry/flows/FlowModule.java
+++ b/core/src/main/java/google/registry/flows/FlowModule.java
@@ -213,50 +213,80 @@ public class FlowModule {
     return Strings.nullToEmpty(((Poll) eppInput.getCommandWrapper().getCommand()).getMessageId());
   }
 
+  private static HistoryEntry.Builder<? extends HistoryEntry, ?> makeHistoryEntryBuilder(
+      HistoryEntry.Builder<? extends HistoryEntry, ?> builder,
+      Trid trid,
+      byte[] inputXmlBytes,
+      boolean isSuperuser,
+      String clientId,
+      EppInput eppInput) {
+    builder
+        .setTrid(trid)
+        .setXmlBytes(inputXmlBytes)
+        .setBySuperuser(isSuperuser)
+        .setClientId(clientId);
+    Optional<MetadataExtension> metadataExtension =
+        eppInput.getSingleExtension(MetadataExtension.class);
+    metadataExtension.ifPresent(
+        extension ->
+            builder
+                .setReason(extension.getReason())
+                .setRequestedByRegistrar(extension.getRequestedByRegistrar()));
+    return builder;
+  }
+
   /**
-   * Provides a partially filled in {@link HistoryEntry} builder.
+   * Provides a partially filled in {@link ContactHistory.Builder}
    *
    * <p>This is not marked with {@link FlowScope} so that each retry gets a fresh one. Otherwise,
    * the fact that the builder is one-use would cause NPEs.
    */
   @Provides
-  static HistoryEntry.Builder provideHistoryEntryBuilder(
+  static ContactHistory.Builder provideContactHistoryBuilder(
       Trid trid,
       @InputXml byte[] inputXmlBytes,
       @Superuser boolean isSuperuser,
       @ClientId String clientId,
       EppInput eppInput) {
-    HistoryEntry.Builder historyBuilder =
-        new HistoryEntry.Builder()
-            .setTrid(trid)
-            .setXmlBytes(inputXmlBytes)
-            .setBySuperuser(isSuperuser)
-            .setClientId(clientId);
-    Optional<MetadataExtension> metadataExtension =
-        eppInput.getSingleExtension(MetadataExtension.class);
-    metadataExtension.ifPresent(
-        extension ->
-            historyBuilder
-                .setReason(extension.getReason())
-                .setRequestedByRegistrar(extension.getRequestedByRegistrar()));
-    return historyBuilder;
+    return (ContactHistory.Builder)
+        makeHistoryEntryBuilder(
+            new ContactHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
   }
 
+  /**
+   * Provides a partially filled in {@link HostHistory.Builder}
+   *
+   * <p>This is not marked with {@link FlowScope} so that each retry gets a fresh one. Otherwise,
+   * the fact that the builder is one-use would cause NPEs.
+   */
   @Provides
-  static ContactHistory.Builder provideContactHistoryBuilder(
-      HistoryEntry.Builder historyEntryBuilder) {
-    return new ContactHistory.Builder().copyFrom(historyEntryBuilder);
+  static HostHistory.Builder provideHostHistoryBuilder(
+      Trid trid,
+      @InputXml byte[] inputXmlBytes,
+      @Superuser boolean isSuperuser,
+      @ClientId String clientId,
+      EppInput eppInput) {
+    return (HostHistory.Builder)
+        makeHistoryEntryBuilder(
+            new HostHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
   }
 
+  /**
+   * Provides a partially filled in {@link DomainHistory.Builder}
+   *
+   * <p>This is not marked with {@link FlowScope} so that each retry gets a fresh one. Otherwise,
+   * the fact that the builder is one-use would cause NPEs.
+   */
   @Provides
   static DomainHistory.Builder provideDomainHistoryBuilder(
-      HistoryEntry.Builder historyEntryBuilder) {
-    return new DomainHistory.Builder().copyFrom(historyEntryBuilder);
-  }
-
-  @Provides
-  static HostHistory.Builder provideHostHistoryBuilder(HistoryEntry.Builder historyEntryBuilder) {
-    return new HostHistory.Builder().copyFrom(historyEntryBuilder);
+      Trid trid,
+      @InputXml byte[] inputXmlBytes,
+      @Superuser boolean isSuperuser,
+      @ClientId String clientId,
+      EppInput eppInput) {
+    return (DomainHistory.Builder)
+        makeHistoryEntryBuilder(
+            new DomainHistory.Builder(), trid, inputXmlBytes, isSuperuser, clientId, eppInput);
   }
 
   /**
--- a/core/src/main/java/google/registry/model/contact/ContactHistory.java
+++ b/core/src/main/java/google/registry/model/contact/ContactHistory.java
@@ -18,6 +18,7 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.annotation.EntitySubclass;
+import com.googlecode.objectify.annotation.Ignore;
 import google.registry.model.ImmutableObject;
 import google.registry.model.contact.ContactHistory.ContactHistoryId;
 import google.registry.model.reporting.HistoryEntry;
@@ -50,6 +51,9 @@ import javax.persistence.PostLoad;
       @javax.persistence.Index(columnList = "historyType"),
       @javax.persistence.Index(columnList = "historyModificationTime")
     })
+// This class is only marked as an entity subclass and registered with Objectify so that when
+// building it its ID can be auto-populated by Objectify. It is converted to its superclass
+// HistoryEntry when persisted to Datastore.
 @EntitySubclass
 @Access(AccessType.FIELD)
 @IdClass(ContactHistoryId.class)
@@ -57,7 +61,7 @@ public class ContactHistory extends HistoryEntry implements SqlEntity {
 
   // Store ContactBase instead of ContactResource so we don't pick up its @Id
   // Nullable for the sake of pre-Registry-3.0 history objects
-  @Nullable ContactBase contactBase;
+  @Nullable @Ignore ContactBase contactBase;
 
   @Id
   @Access(AccessType.PROPERTY)
--- a/core/src/main/java/google/registry/model/contact/ContactHistory.java
+++ b/core/src/main/java/google/registry/model/contact/ContactHistory.java
@@ -18,7 +18,6 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.annotation.EntitySubclass;
-import com.googlecode.objectify.annotation.Ignore;
 import google.registry.model.ImmutableObject;
 import google.registry.model.contact.ContactHistory.ContactHistoryId;
 import google.registry.model.reporting.HistoryEntry;
@@ -63,7 +62,7 @@ public class ContactHistory extends HistoryEntry implements SqlEntity {
 
   // Store ContactBase instead of ContactResource so we don't pick up its @Id
   // Nullable for the sake of pre-Registry-3.0 history objects
-  @Nullable @Ignore ContactBase contactBase;
+  @Nullable ContactBase contactBase;
 
   @Id
   @Access(AccessType.PROPERTY)
--- a/core/src/main/java/google/registry/model/contact/ContactHistory.java
+++ b/core/src/main/java/google/registry/model/contact/ContactHistory.java
@@ -42,6 +42,11 @@ import javax.persistence.PostLoad;
  * <p>In addition to the general history fields (e.g. action time, registrar ID) we also persist a
  * copy of the contact entity at this point in time. We persist a raw {@link ContactBase} so that
  * the foreign-keyed fields in that class can refer to this object.
+ *
+ * <p>This class is only marked as a Datastore entity subclass and registered with Objectify so that
+ * when building it its ID can be auto-populated by Objectify. It is converted to its superclass
+ * {@link HistoryEntry} when persisted to Datastore using {@link
+ * google.registry.persistence.transaction.TransactionManager}.
  */
 @Entity
 @javax.persistence.Table(
@@ -51,9 +56,6 @@ import javax.persistence.PostLoad;
       @javax.persistence.Index(columnList = "historyType"),
       @javax.persistence.Index(columnList = "historyModificationTime")
     })
-// This class is only marked as an entity subclass and registered with Objectify so that when
-// building it its ID can be auto-populated by Objectify. It is converted to its superclass
-// HistoryEntry when persisted to Datastore.
 @EntitySubclass
 @Access(AccessType.FIELD)
 @IdClass(ContactHistoryId.class)
--- a/core/src/main/java/google/registry/model/domain/DomainHistory.java
+++ b/core/src/main/java/google/registry/model/domain/DomainHistory.java
@@ -21,7 +21,6 @@ import static google.registry.util.CollectionUtils.nullToEmptyImmutableCopy;
 import com.google.common.collect.ImmutableSet;
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.annotation.EntitySubclass;
-import com.googlecode.objectify.annotation.Ignore;
 import google.registry.model.ImmutableObject;
 import google.registry.model.domain.DomainHistory.DomainHistoryId;
 import google.registry.model.domain.GracePeriod.GracePeriodHistory;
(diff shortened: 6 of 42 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 197 at r1](https://reviewable.io/reviews/google/nomulus/1150#-M_W0VlC4s7xeh0yOiJj:-M_W0VlC4s7xeh0yOiJk:b-920lyi) ([raw file](https://github.com/google/nomulus/blob/2c1af20bb3a1ada9e53556b738f0f074c4883c2d/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L197)):*
> ```Java
>                                 .setModificationTime(tm().getTransactionTime())
>                                 // TODO: modify this to use setDomainContent instead when converting
>                                 // This action to be SQL-aware.
> ```

Shouldn't be capitalized.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 467 at r1](https://reviewable.io/reviews/google/nomulus/1150#-M_W1-a_21jpdEXUYAfm:-M_W1-aa2sDf4hE0-LSO:b-z481y2) ([raw file](https://github.com/google/nomulus/blob/2c1af20bb3a1ada9e53556b738f0f074c4883c2d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L467)):*
> ```Java
>   public static <E extends EppResource>
>       HistoryEntry.Builder<? extends HistoryEntry, ?> createBuilderForResource(E parent) {
>     if (parent instanceof DomainContent) {
> ```

This seems like the wrong approach.  Why not use generic types?  Or inheritance?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 467 at r1](https://reviewable.io/reviews/google/nomulus/1150#-M_W1-a_21jpdEXUYAfm:-M_W4DS1EfbCqdMHJsdO:b-3sg2w0) ([raw file](https://github.com/google/nomulus/blob/2c1af20bb3a1ada9e53556b738f0f074c4883c2d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L467)):*

This seems like the wrong approach.  Why not use generic types?  Or inheritance?
</blockquote></details>

Not sure what you meant here. We have to return a concrete subtype of `HistoryEntry.Builder` here. There are a lot of places, especially in tests, where we have a `EppResource` and use it to build a `HistoryEntry`. I guess we can instead add a `asHistoryBuilder` method on `DomainContent`, `HostBase` and `ContactBase` to achieve the same effect. What's wrong with this approach though?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: 0 of 35 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 197 at r1](https://reviewable.io/reviews/google/nomulus/1150#-M_W0VlC4s7xeh0yOiJj:-M_WCgyc2jgQv_ie55zo:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/2c1af20bb3a1ada9e53556b738f0f074c4883c2d/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L197)):*

Shouldn't be capitalized.
</blockquote></details>

Done

jianglai:
Ready for review now. PTAL.

CydeWeys (dismissed):
Commit message could be more descriptive (and in the correct tense).  Something like: "Switch from using raw HistoryEntries to typed subclasses thereof".

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150#-:-M_uaHEe28Xzk6-QGqW8:b-cwjw4m)* status: 0 of 42 files reviewed, 12 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/batch/DeleteProberDataAction.java, line 270 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWLJy8q41nElVb_th:-M_uWLJzEB6LZqHKiF5u:bwcmvfh) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L270)):*
> ```Java
>                 // time the
>                 // mapreduce runs anyway.
>                 tm().updateAll(ImmutableList.of(deletedDomain, historyEntry));
> ```

It seems worth writing an overload of `updateAll()` that takes varargs parameters.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 197 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWVl285H0lvcCEDo8:-M_uWVl285H0lvcCEDo9:b-r0na58) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L197)):*
> ```Java
>                                 .setClientId(recurring.getClientId())
>                                 .setModificationTime(tm().getTransactionTime())
>                                 // TODO: modify this to use setDomain instead when converting
> ```

Whose TODO?
___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 224 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWtJHD3FSOU6moR1r:-M_uWtJHD3FSOU6moR1s:b-xrqhy) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/flows/FlowModule.java#L224)):*
> ```Java
>       EppInput eppInput) {
>     builder
>         .setTrid(trid)
> ```

It feels like `modificationTime` is missing from these?  Where is that getting set?
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 54 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uXE53DM7IqdxB5VKP:-M_uXE53DM7IqdxB5VKQ:b7q99f0) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/contact/ContactHistory.java#L54)):*
> ```Java
>       @javax.persistence.Index(columnList = "historyModificationTime")
>     })
> // This class is only marked as an entity subclass and registered with Objectify so that when
> ```

This seems like it should be in the Javadoc, especially the part about how it's converted to its superclass prior to being persisted to Datastore.  Same for the other 2 classes.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 100 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uXZro55fOevXmwkDw:-M_uXZro55fOevXmwkDx:b-v2x82s) ([raw file](https://github.com/google/nomulus/blob/bf1c34cc3b1678d2d64e78cb26ca264f50febabd/core/src/main/java/google/registry/model/domain/DomainHistory.java#L100)):*
> ```Java
>   // we change to use a composite primary key.
>   // TODO(b/166776754): Investigate if we can reuse domainContent.nsHosts for storing host keys.
>   @Ignore
> ```

Why are these `@Ignore`s being removed?  We don't store this data in the `HistoryEntry` entity in Datastore, right?
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 467 at r1](https://reviewable.io/reviews/google/nomulus/1150#-M_W1-a_21jpdEXUYAfm:-M_uYfUPDZK4HfAofNWB:bfjhx7) ([raw file](https://github.com/google/nomulus/blob/2c1af20bb3a1ada9e53556b738f0f074c4883c2d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L467)):*

Not sure what you meant here. We have to return a concrete subtype of `HistoryEntry.Builder` here. There are a lot of places, especially in tests, where we have a `EppResource` and use it to build a `HistoryEntry`. I guess we can instead add a `asHistoryBuilder` method on `DomainContent`, `HostBase` and `ContactBase` to achieve the same effect. What's wrong with this approach though?
</blockquote></details>

Runtime type inspection is not particularly performant and is rarely the correct way of doing things when all of the types involved are already fully under your control anyway.  You already have the type `E` here; why aren't you just using it?  See `TypeUtils` for some useful methods that can help you use it.

Another good approach would be to make `createBuilderForResource()` abstract, and then implement it correctly using inheritance for the three types of EPP resources.  If it can't be made abstract because the `HistoryEntry` class itself cannot be made abstract, then have it throw `UnsupportedOperationException` here and override it on the three subclasses to do the correct thing.


https://www.artima.com/interfacedesign/PreferPoly.html
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 310 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uY35pEdGuASz40TXP:-M_uY35pEdGuASz40TXQ:b-xg140o) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L310)):*
> ```Java
> 
>   // This method only exists to satisfy the requirement that the HistoryEntry is NOT abstract, it
>   // should never be called directly and all three of the subclass of HistoryEntry implements it.
> ```

If it should never be called directly, then why is it `public`?
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 313 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYC_g9XvipEKYdNjh:-M_uYC_g9XvipEKYdNji:b-3rytj5) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L313)):*
> ```Java
>   @Override
>   public Builder<? extends HistoryEntry, ?> asBuilder() {
>     throw new IllegalStateException(
> ```

I think `UnsupportedOperationException` is a better fit here, as we use `IllegalStateException` in a lot of places with `checkState` that mean a bit of a different thing than "You wrote called method X, and calling method X is never valid".
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-M_uYVds-8s_hdcR03D9:b-ys2n3m) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*
> ```Java
>   // transition so some code duplication is OK.
>   public HistoryEntry asHistoryEntry() {
>     HistoryEntry historyEntry = new HistoryEntry();
> ```

Why this change?  And why is this operating directly on fields rather than using a `Builder` pattern?
___
*[core/src/test/resources/google/registry/model/schema.txt, line 266 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_ua8bZ3H6_p5l_Q1m_:-M_ua8bZ3H6_p5l_Q1ma:bcl1uf3) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/test/resources/google/registry/model/schema.txt#L266)):*
> ```vimhelp
>   java.lang.String otherClientId;
>   java.lang.String reason;
>   java.util.Set<google.registry.model.domain.GracePeriod$GracePeriodHistory> gracePeriodHistories;
> ```

AFAIK these three fields are not actually in the Datastore schema (check BigQuery), whereas the rest of them are.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 280 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_ua4uQ5zfY0MptZMIb:-M_ua4uQ5zfY0MptZMIc:b7vf43p) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/test/resources/google/registry/model/schema.txt#L280)):*
> ```vimhelp
>   org.joda.time.DateTime expirationTime;
> }
> class google.registry.model.domain.GracePeriod$GracePeriodHistory {
> ```

I wouldn't expect to see this table in the Datastore schema.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 322 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_ua0Jz7xoHF3Tiscsu:-M_ua0K-89UiAVOa6hyb:b7vf43p) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/test/resources/google/registry/model/schema.txt#L322)):*
> ```vimhelp
>   int keyTag;
> }
> class google.registry.model.domain.secdns.DomainDsDataHistory {
> ```

I wouldn't expect to see this table in the Datastore schema.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: 0 of 42 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/batch/DeleteProberDataAction.java, line 270 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWLJy8q41nElVb_th:-M_ulzoR0HcKB9T9K3se:b-57i4t7) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L270)):*

It seems worth writing an overload of `updateAll()` that takes varargs parameters.
</blockquote></details>

We have this already for `putAll` fwiw

jianglai (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150#-:-M_vOBPrCG3_XZpY_0_T:b-896fix)* status: 0 of 45 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/batch/DeleteProberDataAction.java, line 270 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWLJy8q41nElVb_th:-M_vEnaQCBsM0fP5oCH7:b-25hfrg) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L270)):*

We have this already for `putAll` fwiw 
</blockquote></details>

Added the varargs overload. But it looks like `putAll` is actually correct because we are adding new entities into the database.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 197 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWVl285H0lvcCEDo8:-M_vGDGP0ZqhlnlsfTlS:b-896fix) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L197)):*

Whose TODO?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 224 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uWtJHD3FSOU6moR1r:-M_vGLkR99ntKrLJuoU2:b5mue7r) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/flows/FlowModule.java#L224)):*

It feels like `modificationTime` is missing from these?  Where is that getting set?
</blockquote></details>

This is a helper function used by the Dagger provider that's used to partially populate a history entry builder based on information available in the flow scope. I think the flow itself must set the modification time depending on what it is trying to do.
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 54 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uXE53DM7IqdxB5VKP:-M_vI1pj8oKz6vK7NLX2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/contact/ContactHistory.java#L54)):*

This seems like it should be in the Javadoc, especially the part about how it's converted to its superclass prior to being persisted to Datastore.  Same for the other 2 classes.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/domain/DomainHistory.java, line 100 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uXZro55fOevXmwkDw:-M_vI3JL-hH3uIxE-EFQ:b5o36q1) ([raw file](https://github.com/google/nomulus/blob/bf1c34cc3b1678d2d64e78cb26ca264f50febabd/core/src/main/java/google/registry/model/domain/DomainHistory.java#L100)):*

Why are these `@Ignore`s being removed?  We don't store this data in the `HistoryEntry` entity in Datastore, right?
</blockquote></details>

As the added comment (javadoc) explained, these classes are never actually persisted to Datastore, it's always the parent class (`HistoryEntry`) that is persisted. So adding `@Ignore` here has no value other than to confuse the reader as to why they are needed (in my opinion).

The fact that the `domainCase` field is not marked `@Ignore` just makes things more confusing. If you actually try to persist this subclass itself directly with ofy without converting it to a `HistoryEntry` you'll get an error saying that `DomainContent` is not a registered entity so the class as is cannot be persisted anyway even with the `@Ignore` annotation on the other fields.

So I felt for consistency we should just leave all the fields alone and emphasize in the javadoc that these classes are not meant to be persisted to datastore directly (and if you try you'll fail). This will be a moot point once we move to SQL only.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 467 at r1](https://reviewable.io/reviews/google/nomulus/1150#-M_W1-a_21jpdEXUYAfm:-M_vMM9Q22yoqQaYbrjh:bjsbjnj) ([raw file](https://github.com/google/nomulus/blob/2c1af20bb3a1ada9e53556b738f0f074c4883c2d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L467)):*

Runtime type inspection is not particularly performant and is rarely the correct way of doing things when all of the types involved are already fully under your control anyway.  You already have the type `E` here; why aren't you just using it?  See `TypeUtils` for some useful methods that can help you use it.

Another good approach would be to make `createBuilderForResource()` abstract, and then implement it correctly using inheritance for the three types of EPP resources.  If it can't be made abstract because the `HistoryEntry` class itself cannot be made abstract, then have it throw `UnsupportedOperationException` here and override it on the three subclasses to do the correct thing.


https://www.artima.com/interfacedesign/PreferPoly.html
</blockquote></details>

I would not worry about performance here because this method is mostly used in tests (where it is easier to reuse the same method for different `EppResource` subclasses) and two actions: `DeleteContactsAndHostsAction` and `CreateSyntheticHistoryEntriesAction`, both of which are going away in the SQL-only world.

In fact one should never call this action outside of tests because as you said if you call this function you already know what the `E` type actually is and you can just create the appropriate `XxxHistory.Builder` and set the parent entity yourself.

Note that this is a static helper method. I don't think inheritance works here. Also we are not creating an instance of `E`, but rather a corresponding `XxxHistory` instance of `E`. Can `TypeUtils` actually be useful here? The mapping between `E` and `XxxHistory` has to happen somewhere.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 310 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uY35pEdGuASz40TXP:-M_vJHZx9zB-asvBdvex:bb4wlq6) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L310)):*

If it should never be called directly, then why is it `public`?
</blockquote></details>

The superclass method is `public` so this one has to be `public`.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 313 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYC_g9XvipEKYdNjh:-M_vJYWpCd8OzwEh-s72:b-896fix) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L313)):*

I think `UnsupportedOperationException` is a better fit here, as we use `IllegalStateException` in a lot of places with `checkState` that mean a bit of a different thing than "You wrote called method X, and calling method X is never valid".
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-M_vJfvS8k_owuizJt3-:b-7uv1vb) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

Why this change?  And why is this operating directly on fields rather than using a `Builder` pattern?
</blockquote></details>

Because the `Builder` for `HistoryEntry` is throwing. We cannot rely on the subclass `Builder`either because the entire point of this method is to return a `HistoryEntry` from a subclass in order to persist it to Datastore.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 266 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_ua8bZ3H6_p5l_Q1m_:-M_vKLHM0lZahuIOpwB_:bu520g6) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/test/resources/google/registry/model/schema.txt#L266)):*

AFAIK these three fields are not actually in the Datastore schema (check BigQuery), whereas the rest of them are.
</blockquote></details>

See the comment on why the `@Ignore` annotations are removed (which is why the fields show up here). Is this file used for anything other than a sanity check? It doesn't actually affect the schema in Datastore itself, right?

And again, this schema is not correct anyway because it currently contains `domainContent` (a couple lines above) which does not exist in Datastore. Furthermore, `DomainHistory` is **not** a entity that exists in Datastore (check production datastore) because, again, it is converted to `HistoryEntry` prior to being saved. These subclasses really shouldn't be annotated as `@EntitySubclass` were it not for the fact that they need to be registered with Objectify in order to be assigned an ID when using the builder.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 280 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_ua4uQ5zfY0MptZMIb:-M_vLIbODGTgUjNl5g4G:b-rqxvo7) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/test/resources/google/registry/model/schema.txt#L280)):*

I wouldn't expect to see this table in the Datastore schema.
</blockquote></details>

ditto.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 322 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_ua0Jz7xoHF3Tiscsu:-M_vLJVH6f31Pc4ey9HG:b-rqxvo7) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/test/resources/google/registry/model/schema.txt#L322)):*

I wouldn't expect to see this table in the Datastore schema.
</blockquote></details>

ditto.

gbrodman (changes requested):
> Reviewed 3 of 35 files at r1, 33 of 39 files at r2, 9 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: all files reviewed, 16 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 216 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vYXIh4RCZNxZVfUe8:-M_vYXIh4RCZNxZVfUe9:b-xsi22m) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/flows/FlowModule.java#L216)):*
> ```Java
>   }
> 
>   private static HistoryEntry.Builder<? extends HistoryEntry, ?> makeHistoryEntryBuilder(
> ```

Can we not parameterize the types here so that it returns the right builder type, e.g. ContactHistory.Builder?
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 66 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vZaBH7sci4B0mSuY4:-M_vZaBI-Y3R7K0etKfo:b-v084rf) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/contact/ContactHistory.java#L66)):*
> ```Java
>   // Store ContactBase instead of ContactResource so we don't pick up its @Id
>   // Nullable for the sake of pre-Registry-3.0 history objects
>   @Nullable @Ignore ContactBase contactBase;
> ```

Shouldn't we remove the @Ignore for the same reason you mention in your DomainHistory comment?
___
*[core/src/main/java/google/registry/model/host/HostHistory.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vZgqA0FgubSmkq2HP:-M_vZgqA0FgubSmkq2HQ:b-v084rf) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/host/HostHistory.java#L67)):*
> ```Java
>   // Store HostBase instead of HostResource so we don't pick up its @Id
>   // Nullable for the sake of pre-Registry-3.0 history objects
>   @Nullable @Ignore HostBase hostBase;
> ```

Shouldn't we remove the @Ignore for the same reason you mention in your DomainHistory comment?
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-M_vY2KO7fgxCxq20DQT:b-6idqbr) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

Because the `Builder` for `HistoryEntry` is throwing. We cannot rely on the subclass `Builder`either because the entire point of this method is to return a `HistoryEntry` from a subclass in order to persist it to Datastore.
</blockquote></details>

Is the `Builder` throwing here? I don't see why we couldn't just keep using the `Builder::copyFrom` method 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 223 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vWIUk6ETE8PaGalyX:-M_vWIUk6ETE8PaGalyY:bfn749n) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L223)):*
> ```Java
>   protected Set<DomainTransactionRecord> domainTransactionRecords;
> 
>   protected HistoryEntry() {
> ```

I don't think this constructor should be necessary to explicitly define?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: 41 of 45 files reviewed, 16 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ignore)

___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 216 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vYXIh4RCZNxZVfUe8:-M_vgX8wAqS1Q3275pYx:bnhcova) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/flows/FlowModule.java#L216)):*

Can we not parameterize the types here so that it returns the right builder type, e.g. ContactHistory.Builder?
</blockquote></details>

Hmm, not sure what you meant exactly? This is only a helper method, which takes a `Builder` and sets a few fields on it and return the `Builder`. The actual Dagger provider function creates a (sublcass) `Builder` and pass it to the helper function.

If we don't parametrize the types here aren't we dealing with raw types again? Can you give an example on what you have in mind?
___
*[core/src/main/java/google/registry/model/contact/ContactHistory.java, line 66 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vZaBH7sci4B0mSuY4:-M_vf-J56Ir4MJVuXg3v:bbiml6b) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/contact/ContactHistory.java#L66)):*

Shouldn't we remove the @Ignore for the same reason you mention in your DomainHistory comment?
</blockquote></details>

Good call. Missed this one.
___
*[core/src/main/java/google/registry/model/host/HostHistory.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vZgqA0FgubSmkq2HP:-M_vf5VD42IH8EbfNJod:bc6riph) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/host/HostHistory.java#L67)):*

Shouldn't we remove the @Ignore for the same reason you mention in your DomainHistory comment?
</blockquote></details>

Ditto. Done.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-M_vfOY94xupAqAghiaY:b-gr29wk) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

Is the `Builder` throwing here? I don't see why we couldn't just keep using the `Builder::copyFrom` method 
</blockquote></details>

Hmm, I guess we can get a `Builder<HistoryEntry, Builder>` and use `copyFrom` from it. But then we have to make the `Builder` non-abstract. Given that the ultimate goal is to not only use the `Bulider` from subclasses I think it's not worth it to make it non-abstract, only to make it abstract again after the migration. This function is temporary anyway.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 223 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vWIUk6ETE8PaGalyX:-M_vf7lN5VXmUuaW3diP:bcffypz) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L223)):*

I don't think this constructor should be necessary to explicitly define?
</blockquote></details>

Only did this so that one cannot create a `HistoryEntry`directly by calling the constructor.

gbrodman (changes requested):
> Reviewed 6 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: all files reviewed, 14 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 216 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vYXIh4RCZNxZVfUe8:-Ma-BQGn1nFMCwworc-S:br7o4ef) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/flows/FlowModule.java#L216)):*

Hmm, not sure what you meant exactly? This is only a helper method, which takes a `Builder` and sets a few fields on it and return the `Builder`. The actual Dagger provider function creates a (sublcass) `Builder` and pass it to the helper function.

If we don't parametrize the types here aren't we dealing with raw types again? Can you give an example on what you have in mind?
</blockquote></details>

Couldn't we have the signature 

`private static <B extends HistoryEntry.Builder<? extends HistoryEntry, ?>> B makeHistoryEntryBuilder(B builder, Trid trid....)`? 

Would that work? Java parameterized types are weird 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-Ma-G2BX5b2AKSZuTgca:b6mzejh) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

Hmm, I guess we can get a `Builder<HistoryEntry, Builder>` and use `copyFrom` from it. But then we have to make the `Builder` non-abstract. Given that the ultimate goal is to not only use the `Bulider` from subclasses I think it's not worth it to make it non-abstract, only to make it abstract again after the migration. This function is temporary anyway.
</blockquote></details>

I mostly just mean that it seems like we have to methods that copy all the fields (one in the builder and one here). There must be a way to have the functionality using only one method , right? 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 223 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vWIUk6ETE8PaGalyX:-Ma-G_Kb2gxqrmwIOZlg:blvwvk2) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L223)):*

Only did this so that one cannot create a `HistoryEntry`directly by calling the constructor.
</blockquote></details>

ah, can we please add a comment saying that? I assume it would be private if not for Hibernate shenanigans?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 216 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vYXIh4RCZNxZVfUe8:-Ma10jFH18926Gs8WeBN:bxgvi8n) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/flows/FlowModule.java#L216)):*

Couldn't we have the signature 

`private static <B extends HistoryEntry.Builder<? extends HistoryEntry, ?>> B makeHistoryEntryBuilder(B builder, Trid trid....)`? 

Would that work? Java parameterized types are weird 
</blockquote></details>

I guess we could? Are you just trying to make sure that the returned type is the same as the type supplied as the argument?

Made the change. But honestly I don't see this being that helpful as what the helper function returns is what is provided in the provider function, so as long as the return type is correct, I don't think it is critical that it matches the argument type. But why not.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-Ma12KcLClpbFyzFsSoe:b-xh1ghr) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

I mostly just mean that it seems like we have to methods that copy all the fields (one in the builder and one here). There must be a way to have the functionality using only one method , right? 
</blockquote></details>

Well, we can't use the builder so it's doable that we have a helper function that just sets the values on the instance directly (without calling the setters), and have both `asHistoryEntry` and `copyFrom` call that. But seeing that this function (`asHistoryEntry`) is only temporary it doesn't seem worth doing, as I prefer the current `copyFrom`, which uses the setters.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 223 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vWIUk6ETE8PaGalyX:-Ma11igUATHUTqYpK8TP:byf3ui3) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L223)):*

ah, can we please add a comment saying that? I assume it would be private if not for Hibernate shenanigans? 
</blockquote></details>

I'd think that it'll be package private, i. e. the same as no access modifier.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 216 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vYXIh4RCZNxZVfUe8:-Ma4D0zbAkKEdeEXRw1l:bixpg4v) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/flows/FlowModule.java#L216)):*

I guess we could? Are you just trying to make sure that the returned type is the same as the type supplied as the argument?

Made the change. But honestly I don't see this being that helpful as what the helper function returns is what is provided in the provider function, so as long as the return type is correct, I don't think it is critical that it matches the argument type. But why not.
</blockquote></details>

The primary reason why I suggested that was that it allows us to avoid having to cast the builders in `FlowModule` 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-Ma4DuYT4vQE_YFSRVvq:b1c0plb) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

Well, we can't use the builder so it's doable that we have a helper function that just sets the values on the instance directly (without calling the setters), and have both `asHistoryEntry` and `copyFrom` call that. But seeing that this function (`asHistoryEntry`) is only temporary it doesn't seem worth doing, as I prefer the current `copyFrom`, which uses the setters.
</blockquote></details>

At the very least, we can absolutely have a static method that copies one HistoryEntry to another, using some combination of `this`, `getInstance()` (in the builder), and the new `HistoryEntry` that we create at some point. If we miss a field or need to copy a new field, we don't want to have to do that in multiple places.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/FlowModule.java, line 216 at r3](https://reviewable.io/reviews/google/nomulus/1150#-M_vYXIh4RCZNxZVfUe8:-Ma69PM13vU45xBgtpVX:b-rs906u) ([raw file](https://github.com/google/nomulus/blob/759cca8c2717e07fc04e012f2d6dcdc5d1dee8a6/core/src/main/java/google/registry/flows/FlowModule.java#L216)):*

The primary reason why I suggested that was that it allows us to avoid having to cast the builders in `FlowModule` 
</blockquote></details>

Ah, that makes total sense. Good suggestion!
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 323 at r2](https://reviewable.io/reviews/google/nomulus/1150#-M_uYVds-8s_hdcR03D8:-Ma6Cy-D7d89iaKOR3N7:b-ma1sum) ([raw file](https://github.com/google/nomulus/blob/352aa7973ec509da9b5427455870c0f3f872b876/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L323)):*

At the very least, we can absolutely have a static method that copies one HistoryEntry to another, using some combination of `this`, `getInstance()` (in the builder), and the new `HistoryEntry` that we create at some point. If we miss a field or need to copy a new field, we don't want to have to do that in multiple places. 
</blockquote></details>

Done. Although like I said before we have to set the fields directly, instead of using the setters defined in the builder.

gbrodman (approved):
:lgtm:

> Reviewed 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1150#-:-Ma9ULGI8EGb91y8zR86:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

PR #1151, opened 2021-05-12
Convert most poll message queries to QueryComposer

mindhog:
We can't convert all of the poll ack logic, but we can convert most of it to QueryComposer.  This significantly simplifies the implementation and commonality of the code.

As part of this, introduce an SQL-only "LIKE" operator to query composer.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1151)

### DIFF
--- a/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java
@@ -15,14 +15,12 @@
 package google.registry.flows.poll;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.QueryComposer.Comparator.EQ;
 import static google.registry.persistence.transaction.QueryComposer.Comparator.LTE;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 import static google.registry.util.DateTimeUtils.isBeforeOrAt;
 
-import com.googlecode.objectify.cmd.Query;
 import google.registry.model.poll.PollMessage;
 import google.registry.persistence.transaction.QueryComposer;
 import java.util.Optional;
@@ -78,16 +76,6 @@ public final class PollFlowUtils {
       throw new IllegalArgumentException("Unknown poll message type: " + pollMessage.getClass());
     }
     return includeAckedMessageInCount;
-  }
-
-  /** A Datastore query for poll messages from the given registrar that are not in the future. */
-  public static Query<PollMessage> datastorePollMessageQuery(String registrarId, DateTime now) {
-    return ofy()
-        .load()
-        .type(PollMessage.class)
-        .filter("clientId", registrarId)
-        .filter("eventTime <=", now.toDate())
-        .order("eventTime");
   }
 
   /**
--- a/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java
@@ -16,56 +16,30 @@ package google.registry.flows.poll;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.model.ofy.ObjectifyService.ofy;
-import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.persistence.transaction.QueryComposer.Comparator.EQ;
+import static google.registry.persistence.transaction.QueryComposer.Comparator.LTE;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 import static google.registry.util.DateTimeUtils.isBeforeOrAt;
 
 import com.googlecode.objectify.cmd.Query;
 import google.registry.model.poll.PollMessage;
+import google.registry.persistence.transaction.QueryComposer;
 import java.util.Optional;
 import org.joda.time.DateTime;
 
 /** Static utility functions for poll flows. */
 public final class PollFlowUtils {
 
-  public static final String SQL_POLL_MESSAGE_QUERY =
-      "FROM PollMessage WHERE clientId = :registrarId AND eventTime <= :now ORDER BY eventTime ASC";
-  private static final String SQL_POLL_MESSAGE_COUNT_QUERY =
-      "SELECT COUNT(*) FROM PollMessage WHERE clientId = :registrarId AND eventTime <= :now";
-
   /** Returns the number of poll messages for the given registrar that are not in the future. */
   public static int getPollMessageCount(String registrarId, DateTime now) {
-    if (tm().isOfy()) {
-      return datastorePollMessageQuery(registrarId, now).count();
-    } else {
-      return jpaTm()
-          .transact(
-              () ->
-                  jpaTm()
-                      .query(SQL_POLL_MESSAGE_COUNT_QUERY, Long.class)
-                      .setParameter("registrarId", registrarId)
-                      .setParameter("now", now)
-                      .getSingleResult()
-                      .intValue());
-    }
+    return transactIfJpaTm(() -> createPollMessageQuery(registrarId, now).count()).intValue();
   }
 
   /** Returns the first (by event time) poll message not in the future for this registrar. */
   public static Optional<PollMessage> getFirstPollMessage(String registrarId, DateTime now) {
-    if (tm().isOfy()) {
-      return Optional.ofNullable(datastorePollMessageQuery(registrarId, now).first().now());
-    } else {
-      return jpaTm()
-          .transact(
-              () ->
-                  jpaTm()
-                      .query(SQL_POLL_MESSAGE_QUERY, PollMessage.class)
-                      .setParameter("registrarId", registrarId)
-                      .setParameter("now", now)
-                      .setMaxResults(1)
-                      .getResultStream()
-                      .findFirst());
-    }
+    return transactIfJpaTm(
+        () -> createPollMessageQuery(registrarId, now).orderBy("eventTime").first());
   }
 
   /**
@@ -114,6 +88,17 @@ public final class PollFlowUtils {
         .filter("clientId", registrarId)
         .filter("eventTime <=", now.toDate())
         .order("eventTime");
+  }
+
+  /**
+   * Returns the QueryComposer for poll messages from the given registrar that are not in the
+   * future.
+   */
+  public static QueryComposer<PollMessage> createPollMessageQuery(
+      String registrarId, DateTime now) {
+    return tm().createQueryComposer(PollMessage.class)
+        .where("clientId", EQ, registrarId)
+        .where("eventTime", LTE, now);
   }
 
   private PollFlowUtils() {}
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -414,7 +414,7 @@ public class DatastoreTransactionManager implements TransactionManager {
         String comparatorString = pred.comparator.getDatastoreString();
         if (comparatorString == null) {
           throw new UnsupportedOperationException(
-              "The " + pred.comparator + " operation is not supported on Datastore.");
+              String.format("The %s operation is not supported on Datastore.", pred.comparator));
         }
         result = result.filter(pred.fieldName + comparatorString, pred.value);
       }
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -414,7 +414,7 @@ public class DatastoreTransactionManager implements TransactionManager {
         String comparatorString = pred.comparator.getDatastoreString();
         if (comparatorString == null) {
           throw new UnsupportedOperationException(
-              "The " + pred.comparator + " operation is not supported on datastore.");
+              "The " + pred.comparator + " operation is not supported on Datastore.");
         }
         result = result.filter(pred.fieldName + comparatorString, pred.value);
       }
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -411,7 +411,12 @@ public class DatastoreTransactionManager implements TransactionManager {
       checkOnlyOneInequalityField();
       Query<T> result = auditedOfy().load().type(entityClass);
       for (WhereClause pred : predicates) {
-        result = result.filter(pred.fieldName + pred.comparator.getDatastoreString(), pred.value);
+        String comparatorString = pred.comparator.getDatastoreString();
+        if (comparatorString == null) {
+          throw new UnsupportedOperationException(
+              "The " + pred.comparator + " operation is not supported on datastore.");
+        }
+        result = result.filter(pred.fieldName + comparatorString, pred.value);
       }
 
       if (orderBy != null) {
--- a/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
+++ b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java
@@ -152,6 +152,10 @@ public abstract class QueryComposer<T> {
     return criteriaBuilder::greaterThan;
   }
 
+  public static WhereOperator<String> like(CriteriaBuilder criteriaBuilder) {
+    return criteriaBuilder::like;
+  }
+
   /**
    * Enum used to specify comparison operations, e.g. {@code where("fieldName", Comparator.NE,
    * "someval")'}.
@@ -183,7 +187,14 @@ public abstract class QueryComposer<T> {
     GTE(" >=", QueryComposer::greaterThanOrEqualTo),
 
     /** Return only records whose field is greater than the value. */
-    GT(" >", QueryComposer::greaterThan);
+    GT(" >", QueryComposer::greaterThan),
+
+    /**
+     * Return only records whose field matches the pattern.
+     *
+     * <p>SQL ONLY.
+     */
+    LIKE(null, QueryComposer::like);
 
     private final String datastoreString;
 
--- a/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java
+++ b/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java
@@ -99,7 +99,7 @@ final class AckPollMessagesCommand implements CommandWithRemoteApi {
    */
   private void ackPollMessagesDatastore() {
     QueryKeys<PollMessage> query =
-        ofy()
+        auditedOfy()
             .load()
             .type(PollMessage.class)
             .filter("clientId", clientId)
--- a/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java
+++ b/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java
@@ -16,7 +16,6 @@ package google.registry.tools;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static google.registry.flows.poll.PollFlowUtils.createPollMessageQuery;
-import static google.registry.flows.poll.PollFlowUtils.datastorePollMessageQuery;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.model.poll.PollMessageExternalKeyConverter.makePollMessageExternalId;
 import static google.registry.persistence.transaction.QueryComposer.Comparator.LIKE;
@@ -99,7 +98,13 @@ final class AckPollMessagesCommand implements CommandWithRemoteApi {
    * the Datastore size limits.
    */
   private void ackPollMessagesDatastore() {
-    QueryKeys<PollMessage> query = datastorePollMessageQuery(clientId, clock.nowUtc()).keys();
+    QueryKeys<PollMessage> query = ofy()
+        .load()
+        .type(PollMessage.class)
+        .filter("clientId", clientId)
+        .filter("eventTime <=", clock.nowUtc())
+        .order("eventTime")
+        .keys();
     for (List<Key<PollMessage>> keys : Iterables.partition(query, BATCH_SIZE)) {
       tm().transact(
               () ->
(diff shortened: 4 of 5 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/flows/poll/PollFlowUtils.java, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_WNKnz91Zx8OQ5r3pe:-M_WNKnz91Zx8OQ5r3pf:b-haosd) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java#L84)):*
> ```Java
> 
>   /** A Datastore query for poll messages from the given registrar that are not in the future. */
>   public static Query<PollMessage> datastorePollMessageQuery(String registrarId, DateTime now) {
> ```

I think this can be moved into AckPollMessagesCommand if that's the only place it's used 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_WNjg7DY8WIpqw5tSw:-M_WNjg7DY8WIpqw5tSx:b-g8dpcx) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L126)):*
> ```Java
>   }
> 
>   public static WhereOperator<String> like(CriteriaBuilder criteriaBuilder) {
> ```

can we have a test please? 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 168 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_WMNrQ-ap1_gIO2jPp:-M_WMNrQ-ap1_gIO2jPq:b-lthbv3) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L168)):*
> ```Java
>      * <p>SQL ONLY.
>      */
>     LIKE(null, QueryComposer::like);
> ```

can we have a nice error message in DatastoreQueryComposerImpl if this is null?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: all files reviewed, 4 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 168 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_X7fU20pu8oX2mZrzi:-M_X7fU20pu8oX2mZrzj:bte1m4x) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L168)):*
> ```Java
>      * <p>SQL ONLY.
>      */
>     LIKE(null, QueryComposer::like);
> ```

I don't think this is valid in Datastore.  Datastore does not support a generic `LIKE` operation; it can do prefix or suffix queries only.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/poll/PollFlowUtils.java, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_WNKnz91Zx8OQ5r3pe:-M_aWOdU5LkFpDh-7MsT:b-9wln6v) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java#L84)):*

I think this can be moved into AckPollMessagesCommand if that's the only place it's used 
</blockquote></details>

Yep, Done.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_WNjg7DY8WIpqw5tSw:-M_aMGLEBpI3hLnom1x_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L126)):*

can we have a test please? 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 168 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_X7fU20pu8oX2mZrzi:-M_aFmky4tPNM_XFKsm-:b-u75gim) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L168)):*

I don't think this is valid in Datastore.  Datastore does not support a generic `LIKE` operation; it can do prefix or suffix queries only.
</blockquote></details>

Yeah, it's not.  Note the comment.  The latest version checks for a null datastore comparator string and throws an exception.

I have mixed feelings about adding SQL-specific operations to this cross-database API.  But allowing this does make the calling code a lot nicer in this case. 
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 168 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_WMNrQ-ap1_gIO2jPp:-M_aWSEd-VxdbAUlPxmC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L168)):*

can we have a nice error message in DatastoreQueryComposerImpl if this is null? 
</blockquote></details>

Done.

gbrodman (commented):
> Reviewed 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 289 at r2](https://reviewable.io/reviews/google/nomulus/1151#-M_autSx4FdqWogkLG7O:-M_autSx4FdqWogkLG7P:bajmo15) ([raw file](https://github.com/google/nomulus/blob/9c4453caeb1696e485c6788c296d9c76789840b4/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L289)):*
> ```Java
>                         .stream()
>                         .collect(toImmutableList())))
>         .isEqualTo(ImmutableList.of(alpha));
> ```

nit: `containsExactly(alpha)`, same above with `charlie`
___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 309 at r2](https://reviewable.io/reviews/google/nomulus/1151#-M_aunnC-QBVzox1fay3:-M_aunnC-QBVzox1fay4:bxdq12r) ([raw file](https://github.com/google/nomulus/blob/9c4453caeb1696e485c6788c296d9c76789840b4/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L309)):*
> ```Java
>                         .stream()
>                         .collect(toImmutableList())))
>         .isEqualTo(ImmutableList.of());
> ```

nit: `.isEmpty()`, same a few lines above

gbrodman (commented):
> Reviewed 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 289 at r2](https://reviewable.io/reviews/google/nomulus/1151#-M_autSx4FdqWogkLG7O:-M_b30Fp2wexFn-WzO49:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9c4453caeb1696e485c6788c296d9c76789840b4/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L289)):*

nit: `containsExactly(alpha)`, same above with `charlie`
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 309 at r2](https://reviewable.io/reviews/google/nomulus/1151#-M_aunnC-QBVzox1fay3:-M_b320K8vaCAwb7E72o:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9c4453caeb1696e485c6788c296d9c76789840b4/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L309)):*

nit: `.isEmpty()`, same a few lines above
</blockquote></details>

Done.

CydeWeys (changes requested):
This entire PR just seems like churn?  I'm not sure we're going in the right direction with `QueryComposer`.  We need to finish the migration, not rewrite entire swaths of the codebase for the second time as part of the migration.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151#-:-M_fLW9O0hL4_G8qCvw_:b-f49r2e)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/flows/poll/PollFlowUtils.java, line 34 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fKJzY4X4wl2BZeQVX:-M_fKJzZ60xrB-USPnxo:b-pdysmb) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java#L34)):*
> ```Java
>   /** Returns the number of poll messages for the given registrar that are not in the future. */
>   public static int getPollMessageCount(String registrarId, DateTime now) {
>     return transactIfJpaTm(() -> createPollMessageQuery(registrarId, now).count()).intValue();
> ```

Are you sure that this is correctly running a `SELECT COUNT(*)` query?  If it's not, and it's actually loading the full contents of those poll messages, then it's a significant performance degradation.  Keep in mind some registrars have very large numbers of unacked poll messages (potentially 5 figures).
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 417 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fM5Dt40LObLiuuu2F:-M_fM5Dt40LObLiuuu2G:b2pcdn2) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L417)):*
> ```Java
>         if (comparatorString == null) {
>           throw new UnsupportedOperationException(
>               "The " + pred.comparator + " operation is not supported on datastore.");
> ```

Use `String.format`
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 168 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_X7fU20pu8oX2mZrzi:-M_fKrgt2sCxGXSIHHfK:b-3pfmb2) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L168)):*

Yeah, it's not.  Note the comment.  The latest version checks for a null datastore comparator string and throws an exception.

I have mixed feelings about adding SQL-specific operations to this cross-database API.  But allowing this does make the calling code a lot nicer in this case. 
</blockquote></details>

Wasn't the entire point of `QueryComposer` to make operations that can be run on both Datastore and Cloud SQL go through a unified API?
___
*[core/src/main/java/google/registry/tools/AckPollMessagesCommand.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fMTcv3n39mIIuTMDS:-M_fMTcv3n39mIIuTMDT:b-hqveq1) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java#L102)):*
> ```Java
>   private void ackPollMessagesDatastore() {
>     QueryKeys<PollMessage> query =
>         ofy()
> ```

Why did you inline this instead of using your renamed `createPollMessageQuery()` method from the other poll flow?  Now we've got the same query duplicated across two separate locations.  It needs to be the same query for safety's sake, as the EPP poll operation only gives you the topmost poll message at a time, and that's the only one that can be acked.
___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 350 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fLHK4BGMh55sjk5SW:-M_fLHK4BGMh55sjk5SX:bn99hqr) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L350)):*
> ```Java
>     assertThat(thrown)
>         .hasMessageThat()
>         .contains("The LIKE operation is not supported on datastore.");
> ```

It's "Datastore".  Without capitalizing it to make it clear that you're referring to the specific product, it's ambiguous.  Cloud SQL is also a datastore, but it does support the `LIKE` operation.  Can call this "Cloud Datastore" for complete unambiguity.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/poll/PollFlowUtils.java, line 34 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fKJzY4X4wl2BZeQVX:-M_fjaAM6gjqiTVLABvs:btqi6ec) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/flows/poll/PollFlowUtils.java#L34)):*

Are you sure that this is correctly running a `SELECT COUNT(*)` query?  If it's not, and it's actually loading the full contents of those poll messages, then it's a significant performance degradation.  Keep in mind some registrars have very large numbers of unacked poll messages (potentially 5 figures).
</blockquote></details>

Yes.  JpaTransactionManagerImpl.QueryComposer does a CriteriaQuery count and DatastoreTransactionManager does an ofy count(), so unless either is doing something extremely dumb... 
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 417 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fM5Dt40LObLiuuu2F:-M_fVdo-ApZAp8a6xYpD:b-4mzsct) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L417)):*

Use `String.format`
</blockquote></details>

I've done some reading on this, can you explain the case for String.format()?

As I see it, String.format() is more verbose, less performant, and inherently less readable since the values being interpolated are outside of the location in the format string where they are expanded.

The only advantage I can see to format() (outside of cases where you want to use special formatting patterns) is that it's easier to localize, something we're not likely to do (and certainly not for exceptions).
___
*[core/src/main/java/google/registry/persistence/transaction/QueryComposer.java, line 168 at r1](https://reviewable.io/reviews/google/nomulus/1151#-M_X7fU20pu8oX2mZrzi:-M_fY-1b2Ale-VcsxdWJ:b48wo2a) ([raw file](https://github.com/google/nomulus/blob/39c2e220c92e1cf311799757b01846579e57535b/core/src/main/java/google/registry/persistence/transaction/QueryComposer.java#L168)):*

Wasn't the entire point of `QueryComposer` to make operations that can be run on both Datastore and Cloud SQL go through a unified API?
</blockquote></details>

Yes, it was.  But again, it's really useful in this case to be able to take the common query and modify it to take advantage of the SQL feature.

As I said, I have mixed feelings.  But I don't see a reason to take a hard-line on portability, especially given that the alternative would seem to be branching the code as it was.

We could conceivably break the non-portable functionality into alternate classes that could be constructed from the unified interface, but that seems like overkill.
___
*[core/src/main/java/google/registry/tools/AckPollMessagesCommand.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fMTcv3n39mIIuTMDS:-M_fYWTIBLXedzEYpffy:b-8kagta) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java#L102)):*

Why did you inline this instead of using your renamed `createPollMessageQuery()` method from the other poll flow?  Now we've got the same query duplicated across two separate locations.  It needs to be the same query for safety's sake, as the EPP poll operation only gives you the topmost poll message at a time, and that's the only one that can be acked.
</blockquote></details>

Because QueryComposer doesn't currently support keys() (it's complicated to support this for SQL), and datastore apparently has a hard time with large result sets.  I inlined it here because @gbrodman  pointed out that it's no longer called from within PollFlowUtils, and at that point it didn't seem worth having it in a separate method.

As I consider this further, if we were to support keys() for datastore (in the same way that I'm supporting LIKE for SQL) we could simplify this to a single query.  What do you think?
___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 350 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fLHK4BGMh55sjk5SW:-M_fqnwmE-8CQEYKExGG:b-896fix) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L350)):*

It's "Datastore".  Without capitalizing it to make it clear that you're referring to the specific product, it's ambiguous.  Cloud SQL is also a datastore, but it does support the `LIKE` operation.  Can call this "Cloud Datastore" for complete unambiguity.
</blockquote></details>

Done.

mindhog (commented):
That has occurred to me.  We certainly should have done QueryComposer prior to doing a lot of the branching, and I'm not sure whether it's worth it to convert a lot of other things at this point, but this piece of code was a pretty egregious case of additional complexity.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151#-:-M_gBaxV4GlE2KJS-N2a:b7rsg4d)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

gbrodman (commented):
> Reviewed 4 of 4 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/tools/AckPollMessagesCommand.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fMTcv3n39mIIuTMDS:-M_uxn_-3HwfA1T1OCby:byt099m) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/tools/AckPollMessagesCommand.java#L102)):*

Because QueryComposer doesn't currently support keys() (it's complicated to support this for SQL), and datastore apparently has a hard time with large result sets.  I inlined it here because @gbrodman  pointed out that it's no longer called from within PollFlowUtils, and at that point it didn't seem worth having it in a separate method.

As I consider this further, if we were to support keys() for datastore (in the same way that I'm supporting LIKE for SQL) we could simplify this to a single query.  What do you think?
</blockquote></details>

It's probably not worth it to support `keys()` if there's only a few use cases for it.  Thinking about it more, this is not part of the core poll message acking flow, but rather, a one-off command that we haven't used in a long time (and likely won't use again while we're still on Datastore).  So this is good enough.
___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 171 at r4](https://reviewable.io/reviews/google/nomulus/1151#-M_uyvN04gh636Bc6MdY:-M_uyvN04gh636Bc6MdZ:bbrz94h) ([raw file](https://github.com/google/nomulus/blob/a26a92c257c4cd8810a93771f831597b169e03f2/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L171)):*
> ```Java
>                         .stream()
>                         .collect(toImmutableList())))
>         .containsExactly(alpha);
> ```

Good catches on these, someone wasn't aware of all the Truth methods.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 417 at r3](https://reviewable.io/reviews/google/nomulus/1151#-M_fM5Dt40LObLiuuu2F:-Ma-W0x34BStm8dr_nYx:b-2qgp5r) ([raw file](https://github.com/google/nomulus/blob/023613146c8f4d2d633fd34093535ad73f950cbf/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L417)):*

I see three main arguments in favor of `String::format`

1. It removes any opportunity for the result string to have malformed spacing, which is relatively easy to happen if you're concatenating strings (e.g. missing a space). 
2. It means you can type your error message coherently without being interrupted by the value
3. Sometimes you do want complicated string formatting or interpolation, and it's beneficial to use the same structure across the entire codebase 

go/java-logging-best-practices#prefer-formatted-logging
</blockquote></details>

Well, I'm still unconvinced ;-)

The logging best practice is mainly because some of our tooling makes use of those string templates -- AFAICT it's really not relevant to non-logging use-cases.   And point 2 is exactly one of the reasons I've cited for preferring concatenation. Apparently I'm not alone because most modern languages have some sort of expression interpolation feature these days (e.g. f'hello {place}' in python).

But ok, we can take this up offline (or not).  For now, Done.
___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 171 at r4](https://reviewable.io/reviews/google/nomulus/1151#-M_uyvN04gh636Bc6MdY:-Ma-YJxj9mCDwzIY3FEh:b-9fh7fe) ([raw file](https://github.com/google/nomulus/blob/a26a92c257c4cd8810a93771f831597b169e03f2/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L171)):*

Good catches on these, someone wasn't aware of all the Truth methods.
</blockquote></details>

Haha, that someone would be me :-)  I do understand the rationale for "containsExactly()", but the discrepancy between that and isEqualTo() for everything else is likely to keep biting me.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1151)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1152, opened 2021-05-12
Convert more ofy() to auditedOfy() calls

gbrodman:
A couple of these use the QueryComposer interface to avoid branching.

In addition, we enforce the Datastore restriction that there can be at
most 1 field with an inequality query, see https://cloud.google.com/appengine/docs/standard/go111/datastore/query-restrictions#inequality_filters_are_limited_to_at_most_one_property

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1152)

### DIFF
--- a/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
+++ b/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java
@@ -14,6 +14,7 @@
 
 package google.registry.model.ofy;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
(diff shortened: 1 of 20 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1152)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 468 at r1](https://reviewable.io/reviews/google/nomulus/1152#-M_WpJI14CBLXnE43xTz:-M_WpJI28aDXcFhUvcT6:b-hen08m) ([raw file](https://github.com/google/nomulus/blob/9574f4983f9879e1812f8aded8a7aa1cde11df65/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L468)):*
> ```Java
>           numInequalityFields <= 1,
>           "Datastore cannot handle inequality queries on multiple fields, we found %s fields.",
>           numInequalityFields);
> ```

Nice.  Though I recall getting an error from datastore for this.

PR #1158, opened 2021-05-12
Clean up some SqlEntity classes

gbrodman:
This started as having a better check for when to run the
ReplayCommitLogsToSqlAction but that'll require a bit more thought, and
this is a fairly simple PR that can be split out.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1158)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/secdns/DomainDsDataHistory.java
+++ b/core/src/main/java/google/registry/model/domain/secdns/DomainDsDataHistory.java
@@ -16,9 +16,7 @@ package google.registry.model.domain.secdns;
 
 import google.registry.model.domain.DomainHistory;
 import google.registry.model.ofy.ObjectifyService;
-import google.registry.schema.replay.DatastoreEntity;
-import google.registry.schema.replay.SqlEntity;
-import java.util.Optional;
+import google.registry.schema.replay.SqlOnlyEntity;
 import javax.persistence.Access;
 import javax.persistence.AccessType;
 import javax.persistence.Column;
@@ -27,7 +25,7 @@ import javax.persistence.Id;
 
 /** Entity class to represent a historic {@link DelegationSignerData}. */
 @Entity
-public class DomainDsDataHistory extends DomainDsDataBase implements SqlEntity {
+public class DomainDsDataHistory extends DomainDsDataBase implements SqlOnlyEntity {
 
   @Id Long dsDataHistoryRevisionId;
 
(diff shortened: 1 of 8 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1158)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1158)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java, line 35 at r1](https://reviewable.io/reviews/google/nomulus/1158#-M_fJeeD73KwCLO5-6dL:-M_fJeeE9d4iNxFU1cyo:b39t5qs) ([raw file](https://github.com/google/nomulus/blob/6121990d534c52773dec4600693d9ba0873d1cc2/core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java#L35)):*
> ```Java
>   public static DateTime get() {
>     jpaTm().assertInTransaction();
>     return jpaTm().loadAllOf(SqlReplayCheckpoint.class).stream()
> ```

I guess `setMaxResults(1)` isn't relevant here because we know there's only 1?

We might want to add a convenience method to `TransactionManager` for the singleton pattern, as `loadAllOf(...).stream().findFirst()` is definitely sub-optimal compared to `loadSingleton(...)` or similar.

And as a side-benefit, `loadSingleton(...)` could throw if there's more than one for some reason.

Note: Doesn't have to be in this PR.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1158)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java, line 35 at r1](https://reviewable.io/reviews/google/nomulus/1158#-M_fJeeD73KwCLO5-6dL:-M_fTlquE4ALKTBvrYv7:by0jv4z) ([raw file](https://github.com/google/nomulus/blob/6121990d534c52773dec4600693d9ba0873d1cc2/core/src/main/java/google/registry/schema/replay/SqlReplayCheckpoint.java#L35)):*

I guess `setMaxResults(1)` isn't relevant here because we know there's only 1?

We might want to add a convenience method to `TransactionManager` for the singleton pattern, as `loadAllOf(...).stream().findFirst()` is definitely sub-optimal compared to `loadSingleton(...)` or similar.

And as a side-benefit, `loadSingleton(...)` could throw if there's more than one for some reason.

Note: Doesn't have to be in this PR.
</blockquote></details>

for sure, b/188181482

PR #1162, opened 2021-05-14
Add loadOnlyOf method to tm()

gbrodman:
In addition there's a bit of a refator of SqlReplayCheckpoint to make it
more in line with the other singletons. This method is useful for the
singleton classes where we expect at most one entity to exist, e.g.
ServerSecret.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1162)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUILq0pi52-lJ-Zjo:-M_uUILq0pi52-lJ-Zjp:b-yn8czh) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*
> ```Java
>   @Override
>   public <T> Optional<T> loadOnlyOf(Class<T> clazz) {
>     List<T> resultList = getOfy().load().type(clazz).list();
> ```

This is not performant.  There needs to be a `limit()` on here so you don't accidentally do a table scan if the wrong class is passed in.  It could be `limit(2)`.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUILq0pi52-lJ-Zjo:-M_uUoqNCi08hh8vVCNr:b-3fojo5) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

This is not performant.  There needs to be a `limit()` on here so you don't accidentally do a table scan if the wrong class is passed in.  It could be `limit(2)`.
</blockquote></details>

Or a `setMaxResults()` or whatever.

Though actually I don't think you should be changing this to a query in Datastore at all; it should continue to be a load-by-key using the singleton ID.  Load by key has stronger protections around it that this kind of query does not.

Also I don't think this query works, generally, as it's not strongly transactionally consistent; you would need to add the `ancestor()` query to it to do it properly, and usually (but not always) the cross-TLD entity group.

Basically, these Datastore changes need to be backed out.  It has to be load by key.
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 279 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUTii51WYoXl1RDaI:-M_uUTii51WYoXl1RDaJ:b-d4o0ga) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L279)):*
> ```Java
>     checkArgument(
>         resultList.size() <= 1,
>         "Expected at most one entity of type %s, found %s",
> ```

`%s` is the wrong format option for an int (though this will go away with the aforementioned limit).
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 484 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUkd21jQKsytcGk51:-M_uUkd21jQKsytcGk52:b-2u935b) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L484)):*
> ```Java
>     List<T> resultList =
>         getEntityManager()
>             .createQuery(String.format("FROM %s", getEntityType(clazz).getName()), clazz)
> ```

Same comment regarding `limit()`

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/persistence/transaction/TransactionManagerTest.java, line 378 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uVr2GDLfsma6N9mN_:-M_uVr2GDLfsma6N9mNa:b-ygx5y4) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/test/java/google/registry/persistence/transaction/TransactionManagerTest.java#L378)):*
> ```Java
> 
>   @TestOfyAndSql
>   void loadOnly_returnsValue_orEmpty() {
> ```

These tests are named wrong.  The method they're testing is named `loadOnlyOf`, not `loadOnly`.
___
*[db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uVZftDteuv8mENFcy:-M_uVZftDteuv8mENFcz:b3g9vdk) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql#L15)):*
> ```TSQL
> 
> ALTER TABLE IF EXISTS "SqlReplayCheckpoint" RENAME revision_id TO id;
> ```

What is `id` exactly?  That's a very non-descriptive column name.  What kind of data does it contain?  What does it do?  A table that only has a single column in it named `id` seems like it does nothing at all, as `id` is implied to be metadata whereas here it seems like it's doing the actual real work.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 5 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUILq0pi52-lJ-Zjo:-M_umiXP4MCI5gCfS0Ck:b-on8oa7) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

Or a `setMaxResults()` or whatever.

Though actually I don't think you should be changing this to a query in Datastore at all; it should continue to be a load-by-key using the singleton ID.  Load by key has stronger protections around it that this kind of query does not.

Also I don't think this query works, generally, as it's not strongly transactionally consistent; you would need to add the `ancestor()` query to it to do it properly, and usually (but not always) the cross-TLD entity group.

Basically, these Datastore changes need to be backed out.  It has to be load by key.
</blockquote></details>

We cannot use `loadByKey` because we cannot guarantee that all possible types of these objects use the same key structure. 

I was considering doing a limit operation but I figured it'd be better to load everything so we can give an accurate exception message, given that we'd expect that there'd only ever be one entity (and certainly not many). 

Do you know if `count()` is particularly performant or non-performant in Datastore? Assuming that's all right, we can do what I have now which is a `count` followed by the already-existing `loadAllOf` 
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 279 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUTii51WYoXl1RDaI:-M_umKYD-SBNt7CIRgTk:byeh0bs) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L279)):*

`%s` is the wrong format option for an int (though this will go away with the aforementioned limit).
</blockquote></details>

You'd expect that %d would be the correct option, right? However, with the `checkArgument` string interpolation it isn't! It doesn't print correctly if you do that. 

https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html

`Only %s is supported`
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 484 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUkd21jQKsytcGk51:-M_upiw_7ChQ82bcTfdX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L484)):*

Same comment regarding `limit()`
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/persistence/transaction/TransactionManagerTest.java, line 378 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uVr2GDLfsma6N9mN_:-M_ups4p4OetvjFMhJeg:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/test/java/google/registry/persistence/transaction/TransactionManagerTest.java#L378)):*

These tests are named wrong.  The method they're testing is named `loadOnlyOf`, not `loadOnly`.
</blockquote></details>

Done.
___
*[db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uVZftDteuv8mENFcy:-M_uq-SE902TShg93ER9:b-etikpj) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql#L15)):*
https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/common/CrossTldSingleton.java#L34 

> A table that only has a single column in it named id seems like it does nothing at all

This is not the case. https://github.com/google/nomulus/blob/master/db/src/main/resources/sql/schema/nomulus.golden.sql#L1000

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUILq0pi52-lJ-Zjo:-M_vMa3P-SoYZ5-HIM7t:bwebtgz) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

We cannot use `loadByKey` because we cannot guarantee that all possible types of these objects use the same key structure. 

I was considering doing a limit operation but I figured it'd be better to load everything so we can give an accurate exception message, given that we'd expect that there'd only ever be one entity (and certainly not many). 

Do you know if `count()` is particularly performant or non-performant in Datastore? Assuming that's all right, we can do what I have now which is a `count` followed by the already-existing `loadAllOf` 
</blockquote></details>

I'm not seeing an answer to the ancestor/transactional consistency problem.
___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 279 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUTii51WYoXl1RDaI:-M_vMlBP9voJSGIq0ViN:bt3ywpz) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L279)):*

You'd expect that %d would be the correct option, right? However, with the `checkArgument` string interpolation it isn't! It doesn't print correctly if you do that. 

https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html

`Only %s is supported`
</blockquote></details>

wtf ... it's not using a real `String.format()` string?!

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUILq0pi52-lJ-Zjo:-M_vNi495aeM6Sfhjvv9:b-is1vh1) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

I'm not seeing an answer to the ancestor/transactional consistency problem.
</blockquote></details>

See e.g. https://github.com/objectify/objectify-legacy-wiki/blob/v4/Concepts.wiki

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uUILq0pi52-lJ-Zjo:-M_vUfabAU0q-63etShd:bjgup7k) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

See e.g. https://github.com/objectify/objectify-legacy-wiki/blob/v4/Concepts.wiki
</blockquote></details>

We delegate to `loadAllOf` which handles whether or not to use an ancestor query.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r2](https://reviewable.io/reviews/google/nomulus/1162#-M_wMO-cC_wohy_I4t8A:-M_wMO-cC_wohy_I4t8B:blegyj1) ([raw file](https://github.com/google/nomulus/blob/ef5ac3f8228e789dc9e5c73f78f71925769f3d71/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*
> ```Java
>   @Override
>   public <T> Optional<T> loadOnlyOf(Class<T> clazz) {
>     int numElements = getOfy().load().type(clazz).count();
> ```

Unfortunately `count()` is not performant when there are a large # of entities, so you definitely want a `setMaxResults()` on this.  Consider the solution I propose for Cloud SQL below, so that it can be done in a single round trip rather than needing to separately inquire of the DB the count and then getting the element.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 486 at r2](https://reviewable.io/reviews/google/nomulus/1162#-M_wLKQiEpY3LuEdd9Eq:-M_wLKQj-QVQw2eRCr7o:bjrlks2) ([raw file](https://github.com/google/nomulus/blob/ef5ac3f8228e789dc9e5c73f78f71925769f3d71/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L486)):*
> ```Java
>         getEntityManager()
>             .createQuery(String.format("SELECT COUNT(*) FROM %s", tableName), Long.class)
>             .getSingleResult();
> ```

FYI this throws an exception if the query does not return exactly 1 result, so I don't think it's what you want; the error-checking code below won't end up triggering at all.  We also shouldn't need to perform two separate queries (which involves two roundtrips to the DB) just to do this.

I think the simplest, best way to do this is to run a `"FROM %s"` `.limit(2)` query, and then return the only element of that result stream if it has size 1, otherwise throw.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r2](https://reviewable.io/reviews/google/nomulus/1162#-M_wMO-cC_wohy_I4t8A:-M_zy7vHDna6I0PuG-GM:b2nign3) ([raw file](https://github.com/google/nomulus/blob/ef5ac3f8228e789dc9e5c73f78f71925769f3d71/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

Unfortunately `count()` is not performant when there are a large # of entities, so you definitely want a `setMaxResults()` on this.  Consider the solution I propose for Cloud SQL below, so that it can be done in a single round trip rather than needing to separately inquire of the DB the count and then getting the element.
</blockquote></details>

yep
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 486 at r2](https://reviewable.io/reviews/google/nomulus/1162#-M_wLKQiEpY3LuEdd9Eq:-M_zy-hnBR4MaL8sUP0F:b-ldv17e) ([raw file](https://github.com/google/nomulus/blob/ef5ac3f8228e789dc9e5c73f78f71925769f3d71/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L486)):*

FYI this throws an exception if the query does not return exactly 1 result, so I don't think it's what you want; the error-checking code below won't end up triggering at all.  We also shouldn't need to perform two separate queries (which involves two roundtrips to the DB) just to do this.

I think the simplest, best way to do this is to run a `"FROM %s"` `.limit(2)` query, and then return the only element of that result stream if it has size 1, otherwise throw.
</blockquote></details>

Yeah I mean the question that I was essentially asking is "is it worth having two queries, one of which is a `count`, to get the right details in the exception" and it sounds like you think the answer is "no", which is all good.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r2](https://reviewable.io/reviews/google/nomulus/1162#-M_wMO-cC_wohy_I4t8A:-Ma0PbRKAVxzcdIQZAs_:brtg00w) ([raw file](https://github.com/google/nomulus/blob/ef5ac3f8228e789dc9e5c73f78f71925769f3d71/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

yep
</blockquote></details>

Now you're not doing the ancestor query correctly though.
___
*[db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uVZftDteuv8mENFcy:-Ma0Qfvn9Ym0mGLj-RGR:b-nlsgzq) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql#L15)):*

https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/common/CrossTldSingleton.java#L34 

> A table that only has a single column in it named id seems like it does nothing at all

This is not the case. https://github.com/google/nomulus/blob/master/db/src/main/resources/sql/schema/nomulus.golden.sql#L1000
</blockquote></details>

So if `id` always has the exact same one value, then why does it even need to exist at all?  It needs to exist in Datastore because every entity must have a key, but here in Cloud SQL it's not necessary.  Your basic `FROM TABLENAME LIMIT 1` query works fine.

CydeWeys (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/tmch/TmchCrl.java, line 50 at r3](https://reviewable.io/reviews/google/nomulus/1162#-Ma0RAAT4xBa6zxsBqBs:-Ma0RAAT4xBa6zxsBqBt:b-ymg298) ([raw file](https://github.com/google/nomulus/blob/6bc51ea2e26397bb6f35808c193e9979680da594/core/src/main/java/google/registry/model/tmch/TmchCrl.java#L50)):*
> ```Java
>   /** Returns the singleton instance of this entity, without memoization. */
>   public static Optional<TmchCrl> get() {
>     return tm().transact(() -> tm().loadOnlyOf(TmchCrl.class));
> ```

It may be the wrong thing to do here to create such a narrowly scoped transaction.  It might be more correct to assert that we're in a transaction.  What do you think?
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 480 at r3](https://reviewable.io/reviews/google/nomulus/1162#-Ma0ReMv3CbWEGP5FZ6S:-Ma0ReMv3CbWEGP5FZ6T:b-y6r8jh) ([raw file](https://github.com/google/nomulus/blob/6bc51ea2e26397bb6f35808c193e9979680da594/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L480)):*
> ```Java
>   @Override
>   public <T> Optional<T> loadOnlyOf(Class<T> clazz) {
>     checkArgumentNotNull(clazz, "clazz must be specified");
> ```

You're asserting that this argument is not null here but you aren't doing the same above for the Objectify implementation.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 276 at r2](https://reviewable.io/reviews/google/nomulus/1162#-M_wMO-cC_wohy_I4t8A:-Ma4A61IA29NO3sB1DNk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ef5ac3f8228e789dc9e5c73f78f71925769f3d71/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L276)):*

Now you're not doing the ancestor query correctly though.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/tmch/TmchCrl.java, line 50 at r3](https://reviewable.io/reviews/google/nomulus/1162#-Ma0RAAT4xBa6zxsBqBs:-Ma4AWOI6S08gvTYyHYn:bh6dd72) ([raw file](https://github.com/google/nomulus/blob/6bc51ea2e26397bb6f35808c193e9979680da594/core/src/main/java/google/registry/model/tmch/TmchCrl.java#L50)):*

It may be the wrong thing to do here to create such a narrowly scoped transaction.  It might be more correct to assert that we're in a transaction.  What do you think?
</blockquote></details>

Nah I think this is actually fine, because it's basically only checked on certificate validation which, IIRC, doesn't have a transaction open already 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 480 at r3](https://reviewable.io/reviews/google/nomulus/1162#-Ma0ReMv3CbWEGP5FZ6S:-Ma4ALAA9AFlZFNWf4oI:b-66o78j) ([raw file](https://github.com/google/nomulus/blob/6bc51ea2e26397bb6f35808c193e9979680da594/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L480)):*

You're asserting that this argument is not null here but you aren't doing the same above for the Objectify implementation.
</blockquote></details>

Yeah, those are each matching what the methods around them do. But it shouldn't be necessary to check for null; this is nonnull anyway. 
___
*[db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1162#-M_uVZftDteuv8mENFcy:-Ma4BPJN9726iiilU7Hi:bz018hy) ([raw file](https://github.com/google/nomulus/blob/8fe67968c9c71f782f4711765a85e7ca295a1776/db/src/main/resources/sql/flyway/V96__rename_sql_checkpoint_fields.sql#L15)):*

So if `id` always has the exact same one value, then why does it even need to exist at all?  It needs to exist in Datastore because every entity must have a key, but here in Cloud SQL it's not necessary.  Your basic `FROM TABLENAME LIMIT 1` query works fine.
</blockquote></details>

For now, it's because it allows us to use the tm() methods to load these generically by key. We can revisit at a later date if we wish.

mindhog (approved):
> Reviewed 8 of 11 files at r1, 1 of 3 files at r3, 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 254 at r4](https://reviewable.io/reviews/google/nomulus/1162#-Ma5AXB_95pSOvPEaxUY:-Ma5AXB_95pSOvPEaxUZ:b-7sll0b) ([raw file](https://github.com/google/nomulus/blob/989d12567bf9fa2f23d0e78cdffe414312c2738e/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L254)):*
> ```Java
>    * <p>Throws an exception if there is more than one element in the table.
>    */
>   <T> Optional<T> loadOnlyOf(Class<T> clazz);
> ```

Optional: how would you feel about naming this "`loadSingleton()`"?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManager.java, line 254 at r4](https://reviewable.io/reviews/google/nomulus/1162#-Ma5AXB_95pSOvPEaxUY:-Ma5CK4M3qQTRa6Uij6S:b-k76jv7) ([raw file](https://github.com/google/nomulus/blob/989d12567bf9fa2f23d0e78cdffe414312c2738e/core/src/main/java/google/registry/persistence/transaction/TransactionManager.java#L254)):*

Optional: how would you feel about naming this "`loadSingleton()`"?
</blockquote></details>

yeah that makes as much sense, done

mindhog (approved):
> Reviewed 7 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1162)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @mindhog)

PR #1167, opened 2021-05-18
Make PremiumList.labelsToPrices "insignificant"

mindhog:
Add the ImmutableObject.Insignificant annotation to labelsToPrices and also
mark it as Transient.  In order to do lazy-loads on this field, we need to do
so explicitly: doing otherwise breaks the immutability contract and prevents
detaching the object upon load.

Note that this is an expedient solution to this problem, but not the optimal
one.  Ideally, the disassociation between PremiumList and its PremiumEntry's
would be more explicit.  However, breaking labelsToPrices out would at minimum
require reworking the Create/UpdatePremiumList commands, which currently rely
on passing around a self-contained PremiumList object, both from the parser
interfaces and to the database.

If this approach is acceptable, we can apply it to ReservedList and ClaimsList
as well (though it may be easier to break the association in those cases).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1167)

### DIFF
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -177,7 +177,6 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
    * using this anyway as it's inefficient to load all of the PremiumEntry rows if you don't need
    * them. To check prices, use {@link PremiumListDao#getPremiumPrice} instead.
    */
-  @Nullable
   public synchronized ImmutableMap<String, BigDecimal> getLabelsToPrices() {
     if (labelsToPrices == null) {
       labelsToPrices =
@@ -332,7 +331,7 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   }
 
   @PreRemove
-  void preDelete() {
+  void preRemove() {
     jpaTm()
         .query("DELETE FROM PremiumEntry WHERE revision_id = :revisionId")
         .setParameter("revisionId", revisionId)
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -57,7 +57,6 @@ import javax.persistence.PrePersist;
 import javax.persistence.PreRemove;
 import javax.persistence.Table;
 import javax.persistence.Transient;
-import org.hibernate.LazyInitializationException;
 import org.joda.money.CurrencyUnit;
 import org.joda.money.Money;
 
@@ -86,7 +85,7 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   /**
    * Mapping from unqualified domain names to their prices.
    *
-   * <p>This field requires special treatment since we want to lazy load it.  We have to remove it
+   * <p>This field requires special treatment since we want to lazy load it. We have to remove it
    * from the immutability contract so we can modify it after construction and we have to handle the
    * database processing on our own so we can detach it after load.
    */
@@ -173,8 +172,8 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
    * Returns a {@link Map} of domain labels to prices.
    *
    * <p>Note that this is lazily loaded and thus must be called inside a transaction. You generally
-   * should not be using this anyway as it's inefficient to load all of the PremiumEntry rows if
-   * you don't need them. To check prices, use {@link PremiumListDao#getPremiumPrice} instead.
+   * should not be using this anyway as it's inefficient to load all of the PremiumEntry rows if you
+   * don't need them. To check prices, use {@link PremiumListDao#getPremiumPrice} instead.
    */
   public synchronized ImmutableMap<String, BigDecimal> getLabelsToPrices() {
     if (labelsToPrices == null) {
@@ -338,8 +337,8 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   }
 
   /**
-   * Hibernate hook called on the insert of a new PremiumList.  Stores the associated
-   * {@link PremiumEntry}'s.
+   * Hibernate hook called on the insert of a new PremiumList. Stores the associated {@link
+   * PremiumEntry}'s.
    *
    * <p>We need to persist the list entries, but only on the initial insert (not on update) since
    * the entries themselves never get changed, so we only annotate it with {@link PostPersist}, not
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -83,9 +83,13 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   @Column(nullable = false)
   CurrencyUnit currency;
 
-  // This field requires special treatment since we want to lazy load it.  We have to remove it from
-  // the immutability contract so we can modify it after construction and we have to handle the
-  // database processing on our own so we can detach it after load.
+  /**
+   * Mapping from unqualified domain names to their prices.
+   *
+   * <p>This field requires special treatment since we want to lazy load it.  We have to remove it
+   * from the immutability contract so we can modify it after construction and we have to handle the
+   * database processing on our own so we can detach it after load.
+   */
   @Ignore @ImmutableObject.Insignificant @Transient ImmutableMap<String, BigDecimal> labelsToPrices;
 
   @Ignore
@@ -335,8 +339,14 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
         .executeUpdate();
   }
 
-  // We need to persist the list entries, but only on the initial insert (not on update) since the
-  // entries themselves never get changed.
+  /**
+   * Hibernate hook called on the insert of a new PremiumList.  Stores the associated
+   * {@link PremiumEntry}'s.
+   *
+   * <p>We need to persist the list entries, but only on the initial insert (not on update) since
+   * the entries themselves never get changed, so we only annotate it with {@link PostPersist}, not
+   * {@link PostUpdate}.
+   */
   @PostPersist
   void postPersist() {
     // If the price map is loaded, persist it too.
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -54,6 +54,7 @@ import javax.persistence.Index;
 import javax.persistence.PostLoad;
 import javax.persistence.PostPersist;
 import javax.persistence.PrePersist;
+import javax.persistence.PreRemove;
 import javax.persistence.Table;
 import javax.persistence.Transient;
 import org.hibernate.LazyInitializationException;
@@ -324,6 +325,13 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   @PostLoad
   void postLoad() {
     creationTime = lastUpdateTime;
+  }
+
+  @PreRemove
+  void preDelete() {
+    jpaTm().query("DELETE FROM PremiumEntry WHERE revision_id = :revisionId")
+        .setParameter("revisionId", revisionId)
+        .executeUpdate();
   }
 
   // We need to persist the list entries, but only on the initial insert (not on update) since the
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -172,10 +172,9 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
   /**
    * Returns a {@link Map} of domain labels to prices.
    *
-   * <p>Note that this is lazily loaded and thus will throw a {@link LazyInitializationException} if
-   * used outside the transaction in which the given entity was loaded. You generally should not be
-   * using this anyway as it's inefficient to load all of the PremiumEntry rows if you don't need
-   * them. To check prices, use {@link PremiumListDao#getPremiumPrice} instead.
+   * <p>Note that this is lazily loaded and thus must be called inside a transaction. You generally
+   * should not be using this anyway as it's inefficient to load all of the PremiumEntry rows if
+   * you don't need them. To check prices, use {@link PremiumListDao#getPremiumPrice} instead.
    */
   public synchronized ImmutableMap<String, BigDecimal> getLabelsToPrices() {
     if (labelsToPrices == null) {
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -329,7 +329,8 @@ public final class PremiumList extends BaseDomainLabelList<Money, PremiumList.Pr
 
   @PreRemove
   void preDelete() {
-    jpaTm().query("DELETE FROM PremiumEntry WHERE revision_id = :revisionId")
+    jpaTm()
+        .query("DELETE FROM PremiumEntry WHERE revision_id = :revisionId")
         .setParameter("revisionId", revisionId)
         .executeUpdate();
   }
--- a/core/src/main/java/google/registry/model/registry/label/PremiumList.java
+++ b/core/src/main/java/google/registry/model/registry/label/PremiumList.java
@@ -16,9 +16,12 @@ package google.registry.model.registry.label;
 
 import static com.google.common.base.Charsets.US_ASCII;
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1167)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 86 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0VMR0CiUCebb7h8HD:-Ma0VMR0CiUCebb7h8HE:bckalgk) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L86)):*
> ```Java
>   CurrencyUnit currency;
> 
>   // This field requires special treatment since we want to lazy load it.  We have to remove it from
> ```

This should be a Javadoc.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 89 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0WKjO-9-f54ZASe5T:-Ma0WKjO-9-f54ZASe5U:bi2v5px) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L89)):*
> ```Java
>   // the immutability contract so we can modify it after construction and we have to handle the
>   // database processing on our own so we can detach it after load.
>   @Ignore @ImmutableObject.Insignificant @Transient ImmutableMap<String, BigDecimal> labelsToPrices;
> ```

What exactly are the consequences of marking this as `@ImmutableObject.Insignificant`?  It will cause `.equals()` to return true even if these fields differ, for starters.  Does that matter?  Anything else?
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0VzDh0vSOUICe45bG:-Ma0VzDh0vSOUICe45bH:b-2gxvqs) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L338)):*
> ```Java
>   }
> 
>   // We need to persist the list entries, but only on the initial insert (not on update) since the
> ```

Should be a Javadoc.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0W-NM0kq8lukditky:-Ma0W-NM0kq8lukditkz:b-wi5hvb) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L338)):*
> ```Java
>   }
> 
>   // We need to persist the list entries, but only on the initial insert (not on update) since the
> ```

How does this know that it's on the initial insert and not an update?
___
*[core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0Wo2O31UncnvFM92z:-Ma0Wo2O31UncnvFM93-:b-topu4m) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java#L61)):*
> ```Java
>                     .filter(Optional::isPresent)
>                     .map(Optional::get)
>                     .peek(list -> list.getLabelsToPrices())
> ```

I don't think we need to load this at all.  There's no reasonable use case in which you'd want to list the contents of all of the premium lists themselves alongside just listing the names of the fields.  That output would be a complete unparseable mess.  There's a different command that's suitable for retrieving the contents of lists (`GetPremiumListCommand`), but it doesn't make sense to do so here given what the output format of a list objects command looks like.
___
*[core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0XBBmBB0mqtBwDW8c:-Ma0XBBmBB0mqtBwDW8d:b-ohhv9b) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java#L70)):*
> ```Java
>    */
>   @Override
>   public ImmutableMap<String, String> getFieldOverrides(PremiumList list) {
> ```

Given the above, this isn't necessary either.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 865 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0XGwD51ddIyvvNHa4:-Ma0XGwD51ddIyvvNHa5:b-qw5iqv) ([raw file](https://github.com/google/nomulus/blob/21aeedae11162c2c24d86dbe29f19279d085a302/db/src/main/resources/sql/schema/db-schema.sql.generated#L865)):*
> ```
>        references "DomainHistory";
> 
>     alter table if exists "PremiumEntry" 
> ```

The removal of this foreign key seems incorrect, and also absent from a Flyway file?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1167)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 86 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0VMR0CiUCebb7h8HD:-Ma3frM-Bs8TaTXlEsaT:bgxwytf) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L86)):*

This should be a Javadoc.
</blockquote></details>

Done.

I should point out my rationale, as the plain comments here and below were intentional.  In general, I use javadocs (or doc-strings or doc-comments in other languages) to communicate information relevant to the _users_ of an element and plain comments to communicate information relevant to its implementation details (i.e. for the _maintainer_ of the code).  In this case, for example, the comment is relevant to implementation details that a user shouldn't care about.

That said, having the information in a javadoc isn't particularly harmful. It may be confusing to a consumer in some cases, but it doesn't look like protected fields show up in javadoc, anyway.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 89 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0WKjO-9-f54ZASe5T:-Ma3fcZN0PMLa6CjIjiz:b-uqw7p8) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L89)):*

What exactly are the consequences of marking this as `@ImmutableObject.Insignificant`?  It will cause `.equals()` to return true even if these fields differ, for starters.  Does that matter?  Anything else?
</blockquote></details>

`Insignificant` means that the field is ignored by all of the field processing code in ImmutableObject.  It is not compared for equality, used in the hash, stringified, etc.

This matters in cases where client code is relying on the presence of those fields, which can be a tricky question to answer in general.  The one case where this is obviously true here is in the "list" command.  

It is unlikely to be an issue for comparison: AIUI, premium list revisions are intended to be immutable.  We always add new revisions and keep the old ones intact.  Therefore, comparison of the revision id should always be sufficient for the purpose of determining whether two premium lists are identical.  In fact, since they are joined by their entries by revision id/tld, they pretty much have to be.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0VzDh0vSOUICe45bG:-Ma3lr7F3JevqLnxkbcG:b-ukw5pn) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L338)):*

Should be a Javadoc.
</blockquote></details>

Done (see caveats above)
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0W-NM0kq8lukditky:-Ma3lt0o5XXlzcaCTLUo:b-jt8lzl) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L338)):*

How does this know that it's on the initial insert and not an update?
</blockquote></details>

Because it's PostPersist and not PostPersist + PostUpdate.  Added some clarifications to the javadoc.
___
*[core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0Wo2O31UncnvFM92z:-Ma3m6WO0Ydx_isARjBw:b-hw1fgi) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java#L61)):*

I don't think we need to load this at all.  There's no reasonable use case in which you'd want to list the contents of all of the premium lists themselves alongside just listing the names of the fields.  That output would be a complete unparseable mess.  There's a different command that's suitable for retrieving the contents of lists (`GetPremiumListCommand`), but it doesn't make sense to do so here given what the output format of a list objects command looks like.
</blockquote></details>

I agree, but someone thought otherwise as there is a test for this.  Let's chat about this on Monday, I can remove the code and the test if we're sure it's appropriate.
___
*[core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0XBBmBB0mqtBwDW8c:-Ma3mbi674GMNcnrdEhc:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java#L70)):*

Given the above, this isn't necessary either.
</blockquote></details>

Ack.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 865 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0XGwD51ddIyvvNHa4:-Ma3mdVJ-evZzrLWr1wI:b-jgpaaf) ([raw file](https://github.com/google/nomulus/blob/21aeedae11162c2c24d86dbe29f19279d085a302/db/src/main/resources/sql/schema/db-schema.sql.generated#L865)):*

The removal of this foreign key seems incorrect, and also absent from a Flyway file?
</blockquote></details>

Haha, yeah I knew this was going to come up :-)  We actually still want the foreign key constraint.  Hibernate just doesn't know to generate it anymore since we've removed the relationship from its control.

Note that this is the _generated_ schema (which we use for reference purposes) and not the _golden_ schema, which should match the actual database schema generated by the flyway files.

gbrodman (approved):
> Reviewed 3 of 5 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1167)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 86 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0VMR0CiUCebb7h8HD:-Ma4Er0y3QkKXZTSnNHg:bpxskky) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L86)):*

Done.

I should point out my rationale, as the plain comments here and below were intentional.  In general, I use javadocs (or doc-strings or doc-comments in other languages) to communicate information relevant to the _users_ of an element and plain comments to communicate information relevant to its implementation details (i.e. for the _maintainer_ of the code).  In this case, for example, the comment is relevant to implementation details that a user shouldn't care about.

That said, having the information in a javadoc isn't particularly harmful. It may be confusing to a consumer in some cases, but it doesn't look like protected fields show up in javadoc, anyway.
</blockquote></details>

eh I think the users of this may need to know about the odd lazy loading behavior?
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 89 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0WKjO-9-f54ZASe5T:-Ma4I-wx8cOqVQ0OnSnD:b-hmbzh7) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L89)):*

`Insignificant` means that the field is ignored by all of the field processing code in ImmutableObject.  It is not compared for equality, used in the hash, stringified, etc.

This matters in cases where client code is relying on the presence of those fields, which can be a tricky question to answer in general.  The one case where this is obviously true here is in the "list" command.  

It is unlikely to be an issue for comparison: AIUI, premium list revisions are intended to be immutable.  We always add new revisions and keep the old ones intact.  Therefore, comparison of the revision id should always be sufficient for the purpose of determining whether two premium lists are identical.  In fact, since they are joined by their entries by revision id/tld, they pretty much have to be.
</blockquote></details>

yeah sounds like this is more or less a cache rather than a normal field (and hopefully a cache that is infrequently used, since we generally don't need / load the full map)
___
*[core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0Wo2O31UncnvFM92z:-Ma4HGxV8L9jCNfZtXqd:b-ncgwjo) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/tools/server/ListPremiumListsAction.java#L61)):*

I agree, but someone thought otherwise as there is a test for this.  Let's chat about this on Monday, I can remove the code and the test if we're sure it's appropriate.
</blockquote></details>

yeah fwiw I agree with Ben, we likely only need the names and it'd make the code here much simpler but yeah we can discuss later 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 865 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0XGwD51ddIyvvNHa4:-Ma4JP7BCDhhr4N4Jtlt:b-r45m17) ([raw file](https://github.com/google/nomulus/blob/21aeedae11162c2c24d86dbe29f19279d085a302/db/src/main/resources/sql/schema/db-schema.sql.generated#L865)):*

Haha, yeah I knew this was going to come up :-)  We actually still want the foreign key constraint.  Hibernate just doesn't know to generate it anymore since we've removed the relationship from its control.

Note that this is the _generated_ schema (which we use for reference purposes) and not the _golden_ schema, which should match the actual database schema generated by the flyway files.
</blockquote></details>

yeah generally I think that the autogenerated systems only make the relationship explicit if you store the object itself via some sort of mapping, rather than simply storing the raw long / string ID

mindhog (commented):
> Dismissed @CydeWeys from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1167)* status: 4 of 5 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 86 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0VMR0CiUCebb7h8HD:-Ma5DLVZ4ui1ohJGJJnp:b5lhgwh) ([raw file](https://github.com/google/nomulus/blob/9723d3718b7af5c63632f89e0e4abaa1522d8130/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L86)):*

eh I think the users of this may need to know about the odd lazy loading behavior?
</blockquote></details>

There's a comment on the getter that indicates the lazy load -- that's the place the user can/should be looking at.  That said, that comment needed an update, so Done.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 865 at r1](https://reviewable.io/reviews/google/nomulus/1167#-Ma0XGwD51ddIyvvNHa4:-Ma5Ch0A3Swp_lSzDb-v:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/21aeedae11162c2c24d86dbe29f19279d085a302/db/src/main/resources/sql/schema/db-schema.sql.generated#L865)):*

yeah generally I think that the autogenerated systems only make the relationship explicit if you store the object itself via some sort of mapping, rather than simply storing the raw long / string ID 
</blockquote></details>

Ack.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1167)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @mindhog)

PR #1170, opened 2021-05-20
Add an object to store database migration stages

gbrodman:
go/registry-3.0-stage-management for more details

This basically boils down to storing an enum in the database so that we
can tell what stage of the migration we're in.

We use a cross-TLD parent so that we can have strong transactional
consistency on retrieval.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1170)

### DIFF
--- a/core/src/main/java/google/registry/model/EntityClasses.java
+++ b/core/src/main/java/google/registry/model/EntityClasses.java
@@ -17,6 +17,7 @@ package google.registry.model;
 import com.google.common.collect.ImmutableSet;
 import google.registry.model.billing.BillingEvent;
 import google.registry.model.common.Cursor;
+import google.registry.model.common.DatabaseMigrationStateWrapper;
 import google.registry.model.common.DatabaseTransitionSchedule;
 import google.registry.model.common.EntityGroupRoot;
 import google.registry.model.common.GaeUserIdConverter;
@@ -75,6 +76,7 @@ public final class EntityClasses {
           ContactHistory.class,
           ContactResource.class,
           Cursor.class,
+          DatabaseMigrationStateWrapper.class,
           DatabaseTransitionSchedule.class,
           DomainBase.class,
           DomainHistory.class,
--- a/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java
+++ b/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java
@@ -1,0 +1,122 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.model.common;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.persistence.transaction.TransactionManagerFactory.ofyTm;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.googlecode.objectify.annotation.Entity;
+import google.registry.model.annotations.InCrossTld;
+import google.registry.model.common.DatabaseTransitionSchedule.PrimaryDatabase;
+import google.registry.schema.replay.DatastoreOnlyEntity;
+
+/**
+ * A wrapper object representing the current stage of the Registry 3.0 Cloud SQL migration.
+ *
+ * <p>The entity is stored in Datastore throughout the entire migration so as to have a single point
+ * of access (avoiding a two-phase commit problem).
+ */
+@Entity
+@InCrossTld
+public class DatabaseMigrationStateWrapper extends CrossTldSingleton
+    implements DatastoreOnlyEntity {
+
+  /**
+   * The current phase of the migration plus information about which database to use and whether or
+   * not the phase is read-only.
+   */
+  public enum MigrationState {
+    DATASTORE_ONLY(PrimaryDatabase.DATASTORE, false),
+    DATASTORE_PRIMARY(PrimaryDatabase.DATASTORE, false),
+    DATASTORE_PRIMARY_READ_ONLY(PrimaryDatabase.DATASTORE, true),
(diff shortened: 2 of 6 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1170)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/1170#-Ma9tTepAl_07Dq3_u0F:-Ma9tTepAl_07Dq3_u0G:b-kblscd) ([raw file](https://github.com/google/nomulus/blob/7b1ea9f45ce4babcb7746ba96b854571ca6083f2/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java#L70)):*
> ```Java
>   // The valid state transitions. Basically, at state N, state N+1 is valid as well as all previous
>   // states, with one exception: when in SQL_ONLY, we can only move back to SQL_PRIMARY because we
>   // must assume that some writes that occurred in SQL_ONLY will not have propagated to Datastore.
> ```

As I think about it, we have a similar problem going from SQL_PRIMARY to any of the datastore primary states: there could be transactions still in the queue that would need to be applied before the datastore can be safely mutated.

You may want to introduce another state: DATASTORE_PRIMARY_READ_ONLY_ROLLBACK where we would block on a transition to DATASTORE_PRIMARY or DATASTORE_ONLY until after all transactions have been propagated.  Alternately, we can just use DATASTORE_PRIMARY_READ_ONLY for this and just make it part of the rollback procedure for the operator to verify this manually.  There's a case to be made for not making the operational constraints too strong in a case like this where human judgement is the primary driver.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/1170#-Ma9wMXq3nrnB32Q9qbM:-Ma9wMXq3nrnB32Q9qbN:btnge0r) ([raw file](https://github.com/google/nomulus/blob/7b1ea9f45ce4babcb7746ba96b854571ca6083f2/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java#L105)):*
> ```Java
>         currentState,
>         newState);
>     DatabaseMigrationStateWrapper stage = new DatabaseMigrationStateWrapper();
> ```

Remember that DatabaseMigrationStateWrapper is supposed to be an ImmutableObject: much better to pass the migration state into the constructor and make it final.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1170)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java, line 70 at r1](https://reviewable.io/reviews/google/nomulus/1170#-Ma9tTepAl_07Dq3_u0F:-MaA75Gz7mTfAO733Gpf:bwmmghw) ([raw file](https://github.com/google/nomulus/blob/7b1ea9f45ce4babcb7746ba96b854571ca6083f2/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java#L70)):*

As I think about it, we have a similar problem going from SQL_PRIMARY to any of the datastore primary states: there could be transactions still in the queue that would need to be applied before the datastore can be safely mutated.

You may want to introduce another state: DATASTORE_PRIMARY_READ_ONLY_ROLLBACK where we would block on a transition to DATASTORE_PRIMARY or DATASTORE_ONLY until after all transactions have been propagated.  Alternately, we can just use DATASTORE_PRIMARY_READ_ONLY for this and just make it part of the rollback procedure for the operator to verify this manually.  There's a case to be made for not making the operational constraints too strong in a case like this where human judgement is the primary driver.
</blockquote></details>

Yeah, good point. I think the easiest thing to do is just to limit it to one step back, i.e. going to the read-only state. 
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/1170#-Ma9wMXq3nrnB32Q9qbM:-MaA79rj3oQeJrcB556o:bf8c0yy) ([raw file](https://github.com/google/nomulus/blob/7b1ea9f45ce4babcb7746ba96b854571ca6083f2/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java#L105)):*

Remember that DatabaseMigrationStateWrapper is supposed to be an ImmutableObject: much better to pass the migration state into the constructor and make it final.
</blockquote></details>

Unfortunately we cannot do that -- Objectify / Datastore require that there exists a no-arg public constructor so while this is an ImmutableObject, we must fill it out after-the-fact like Objectify does. 

See similar patterns in ServerSecret and TmchCrl

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1170)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/1170#-Ma9wMXq3nrnB32Q9qbM:-MaDhhFo2tgebbl_Megg:bx81r10) ([raw file](https://github.com/google/nomulus/blob/7b1ea9f45ce4babcb7746ba96b854571ca6083f2/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java#L105)):*

Unfortunately we cannot do that -- Objectify / Datastore require that there exists a no-arg public constructor so while this is an ImmutableObject, we must fill it out after-the-fact like Objectify does. 

See similar patterns in ServerSecret and TmchCrl 
</blockquote></details>

Why not just add an additional constructor?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1170)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/1170#-Ma9wMXq3nrnB32Q9qbM:-MaEQkSg7ODkaicTP2zR:b-gxztt3) ([raw file](https://github.com/google/nomulus/blob/7b1ea9f45ce4babcb7746ba96b854571ca6083f2/core/src/main/java/google/registry/model/common/DatabaseMigrationStateWrapper.java#L105)):*

Why not just add an additional constructor?
</blockquote></details>

Fair enough

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1170)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1171, opened 2021-05-20
Remove a couple additional ofy() calls

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1171)

### DIFF
--- a/core/src/main/java/google/registry/model/Buildable.java
+++ b/core/src/main/java/google/registry/model/Buildable.java
@@ -16,7 +16,7 @@ package google.registry.model;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
-import static google.registry.model.ofy.ObjectifyService.ofy;
+import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 
 import google.registry.model.ofy.ObjectifyService;
 import google.registry.util.TypeUtils.TypeInstantiator;
@@ -57,8 +57,14 @@ public interface Buildable {
         // If this object has a Long or long Objectify @Id field that is not set, set it now.
         Field idField = null;
         try {
-          idField = ModelUtils.getAllFields(instance.getClass()).get(
-              ofy().factory().getMetadata(instance.getClass()).getKeyMetadata().getIdFieldName());
+          idField =
+              ModelUtils.getAllFields(instance.getClass())
+                  .get(
+                      auditedOfy()
+                          .factory()
+                          .getMetadata(instance.getClass())
+                          .getKeyMetadata()
+                          .getIdFieldName());
         } catch (Exception e) {
           // Expected if the class is not registered with Objectify.
         }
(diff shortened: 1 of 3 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1171)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 212 at r1](https://reviewable.io/reviews/google/nomulus/1171#-MaT_MTd6Qx9mfa3iM4L:-MaT_MTeCRmpPefzR42Q:b-f0qb3u) ([raw file](https://github.com/google/nomulus/blob/f9d6dfd585940d277778951ec230d55c3c8897a7/core/src/main/java/google/registry/model/common/Cursor.java#L212)):*
> ```Java
>       CursorType cursorType, Key<? extends ImmutableObject> scope) {
>     checkArgument(
>         cursorType.getScopeClass().getSimpleName().equals(scope.getKind()),
> ```

This seems to indicate that `getScopeClass()` can no longer return `EntityGroupRoot`.  Is that correct?  Even when running against datastore?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1171)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 212 at r1](https://reviewable.io/reviews/google/nomulus/1171#-MaT_MTd6Qx9mfa3iM4L:-MaTyWTJ2ztgrcLcSDHJ:brpidcp) ([raw file](https://github.com/google/nomulus/blob/f9d6dfd585940d277778951ec230d55c3c8897a7/core/src/main/java/google/registry/model/common/Cursor.java#L212)):*

This seems to indicate that `getScopeClass()` can no longer return `EntityGroupRoot`.  Is that correct?  Even when running against datastore?
</blockquote></details>

I don't think that's correct. Key::getKind will return either `Registry` or `EntityGroupRoot` (if the key in question is the result of `getCrossTldKey()`. 

And `CursorType` is just an enum; it still returns `EntityGroupRoot` in the standard cases (see the enum above)

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1171)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/Cursor.java, line 212 at r1](https://reviewable.io/reviews/google/nomulus/1171#-MaT_MTd6Qx9mfa3iM4L:-MaU0CNO2zDJJzacUMXQ:bl58dtw) ([raw file](https://github.com/google/nomulus/blob/f9d6dfd585940d277778951ec230d55c3c8897a7/core/src/main/java/google/registry/model/common/Cursor.java#L212)):*

I don't think that's correct. Key::getKind will return either `Registry` or `EntityGroupRoot` (if the key in question is the result of `getCrossTldKey()`. 

And `CursorType` is just an enum; it still returns `EntityGroupRoot` in the standard cases (see the enum above) 
</blockquote></details>

Ah, ok.  I see.

PR #1172, opened 2021-05-21
Add methods to return subtypes of HistoryEntry when querying

jianglai:
This is useful when we expect a specific subtype in the return value so
that we can set the parent resource (e. g. DomainContent for
DomainHistory) on it, or when a specific subtype is needed from the call
site.

This PR also fixes some use of generic return values. It is always better to
return <HistoryEntry> than a wildcard <? extends HistoryEntry>, because for
immutable collections, <? extends HistoryEntry> is no different than
<HistoryEntry> as return value -- you can only get a HistoryEntry from it.
The wildcard return value means that even if you are indeed getting a
<DomainHistory> from the query, the call site has no compile time knowledge of
it and can only assume it is a <HistoryEntry>.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1172)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -317,6 +317,7 @@ dependencies {
   testCompile deps['com.google.monitoring-client:contrib']
   testCompile deps['com.google.truth:truth']
   testCompile deps['com.google.truth.extensions:truth-java8-extension']
+  testCompile deps['org.checkerframework:checker-qual']
   testCompile deps['org.hamcrest:hamcrest']
   testCompile deps['org.hamcrest:hamcrest-core']
   testCompile deps['org.hamcrest:hamcrest-library']
--- a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java
+++ b/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java
@@ -14,6 +14,7 @@
 
 package google.registry.model.reporting;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
@@ -49,7 +50,7 @@ import org.joda.time.DateTime;
 public class HistoryEntryDao {
 
   /** Loads all history objects in the times specified, including all types. */
-  public static ImmutableList<? extends HistoryEntry> loadAllHistoryObjects(
+  public static ImmutableList<HistoryEntry> loadAllHistoryObjects(
       DateTime afterTime, DateTime beforeTime) {
     if (tm().isOfy()) {
       return Streams.stream(
@@ -77,13 +78,22 @@ public class HistoryEntryDao {
   }
 
   /** Loads all history objects corresponding to the given {@link EppResource}. */
-  public static ImmutableList<? extends HistoryEntry> loadHistoryObjectsForResource(
+  public static ImmutableList<HistoryEntry> loadHistoryObjectsForResource(
       VKey<? extends EppResource> parentKey) {
     return loadHistoryObjectsForResource(parentKey, START_OF_TIME, END_OF_TIME);
   }
 
+  /**
+   * Loads all history objects corresponding to the given {@link EppResource} and casted to the
+   * appropriate subclass.
+   */
+  public static <T extends HistoryEntry> ImmutableList<T> loadHistoryObjectsForResource(
+      VKey<? extends EppResource> parentKey, Class<T> subclazz) {
+    return loadHistoryObjectsForResource(parentKey, START_OF_TIME, END_OF_TIME, subclazz);
+  }
+
   /** Loads all history objects in the time period specified for the given {@link EppResource}. */
-  public static ImmutableList<? extends HistoryEntry> loadHistoryObjectsForResource(
+  public static ImmutableList<HistoryEntry> loadHistoryObjectsForResource(
       VKey<? extends EppResource> parentKey, DateTime afterTime, DateTime beforeTime) {
     if (tm().isOfy()) {
       return Streams.stream(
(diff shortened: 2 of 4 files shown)


gbrodman (changes requested):
I think something went wrong and a few words / phrases went missing from the commit message somehow 

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1172#-:-MaFVFCk35u_FrWu8hDW:b-fvvb5y)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java, line 90 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFWOYzBq3o9z4Ej7HC:-MaFWOYzBq3o9z4Ej7HD:b-j291lh) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java#L90)):*
> ```Java
>    * appropriate subclass.
>    */
>   public static <T extends HistoryEntry> ImmutableList<T> loadHistoryObjectsForResource(
> ```

I think this will run in to some problems when running in Datastore; we probably want to have some sort of assertion somewhere to make sure we're only running this in SQL. 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java, line 131 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFUqmdCH2ISiFWAYlL:-MaFUqme5T3l1D6Tgjn9:b3s2gk2) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java#L131)):*
> ```Java
>     Class<? extends HistoryEntry> expectedSubclazz = getHistoryClassFromParent(parentKey.getKind());
>     checkArgument(
>         subclazz == expectedSubclazz,
> ```

`.equals` 
___
*[core/src/test/java/google/registry/testing/DatabaseHelper.java, line 1135 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFUjX1AoBy1zTrO_fx:-MaFUjX1AoBy1zTrO_fy:bxnivum) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1135)):*
> ```Java
>       EppResource resource, final HistoryEntry.Type type, Class<T> subclazz) {
>     return getHistoryEntries(resource, subclazz).stream()
>         .filter(entry -> entry.getType() == type)
> ```

prefer `.equals()` for enums 
___
*[core/src/test/java/google/registry/testing/DatabaseHelper.java, line 1145 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFUc8vACNsZm0_MUNS:-MaFUc8vACNsZm0_MUNT:b-e2dt6g) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1145)):*
> ```Java
>   public static HistoryEntry getOnlyHistoryEntryOfType(
>       EppResource resource, final HistoryEntry.Type type) {
>     List<HistoryEntry> historyEntries = getHistoryEntriesOfType(resource, type);
> ```

Just use `Iterables.getOnlyElement()` and it simplifies these methods

jianglai (commented):
Updated commit message, PTAL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1172#-:-MaGVZ3S91LyU5luN-W-:b-lp1n6l)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java, line 90 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFWOYzBq3o9z4Ej7HC:-MaGTC-Q1d13mT8mEYGR:b-l3d1wv) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java#L90)):*

I think this will run in to some problems when running in Datastore; we probably want to have some sort of assertion somewhere to make sure we're only running this in SQL. 
</blockquote></details>

Why? the `loadHistoryObjectsForResource` method that this one ultimately calls `toChildHistoryEntry` on the `HistoryEntry` returned from Datastore, so the downcast should be safe here.
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java, line 131 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFUqmdCH2ISiFWAYlL:-MaGTvOIAf-KguVWn9D4:bedcd6) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java#L131)):*

`.equals` 
</blockquote></details>

They are actually the same since `equals` on `Class` actually [just checks for identity](https://stackoverflow.com/questions/7322469/java-equals-for-a-class-is-same-as-equals). But sure.
___
*[core/src/test/java/google/registry/testing/DatabaseHelper.java, line 1135 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFUjX1AoBy1zTrO_fx:-MaGUgp740aVX2urDF7S:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1135)):*

prefer `.equals()` for enums 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/DatabaseHelper.java, line 1145 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFUc8vACNsZm0_MUNS:-MaGV--4CEi9Nu1KKtgz:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1145)):*

Just use `Iterables.getOnlyElement()` and it simplifies these methods
</blockquote></details>

Done.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1172)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java, line 90 at r1](https://reviewable.io/reviews/google/nomulus/1172#-MaFWOYzBq3o9z4Ej7HC:-MaU-FnF9M7oVUtquPgZ:bn1d905) ([raw file](https://github.com/google/nomulus/blob/f7d4db730e327de2a29114a38da08f48db392a3a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java#L90)):*

Why? the `loadHistoryObjectsForResource` method that this one ultimately calls `toChildHistoryEntry` on the `HistoryEntry` returned from Datastore, so the downcast should be safe here.
</blockquote></details>

Ah, yeah that makes sense, I didn't see that since it's elsewhere in the file

PR #1175, opened 2021-05-21
Use DB migration state to determine running async replay DS->SQL

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1175)

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -27,7 +27,9 @@ import com.google.appengine.tools.cloudstorage.GcsFileMetadata;
 import com.google.appengine.tools.cloudstorage.GcsService;
 import com.google.common.collect.ImmutableList;
 import com.google.common.flogger.FluentLogger;
-import google.registry.config.RegistryConfig;
+import google.registry.model.common.DatabaseMigrationStateSchedule;
+import google.registry.model.common.DatabaseMigrationStateSchedule.MigrationState;
+import google.registry.model.common.DatabaseMigrationStateSchedule.ReplayDirection;
 import google.registry.model.server.Lock;
 import google.registry.model.translators.VKeyTranslatorFactory;
 import google.registry.persistence.VKey;
@@ -39,6 +41,7 @@ import google.registry.schema.replay.DatastoreOnlyEntity;
 import google.registry.schema.replay.NonReplicatedEntity;
 import google.registry.schema.replay.ReplaySpecializer;
 import google.registry.schema.replay.SqlReplayCheckpoint;
+import google.registry.util.Clock;
 import google.registry.util.RequestStatusChecker;
 import java.io.IOException;
 import java.io.InputStream;
@@ -69,15 +72,19 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
   @Inject Response response;
   @Inject RequestStatusChecker requestStatusChecker;
   @Inject GcsDiffFileLister diffLister;
+  @Inject Clock clock;
 
   @Inject
   ReplayCommitLogsToSqlAction() {}
 
   @Override
   public void run() {
-    if (!RegistryConfig.getCloudSqlReplayCommitLogs()) {
-      String message = "ReplayCommitLogsToSqlAction was called but disabled in the config.";
-      logger.atWarning().log(message);
+    MigrationState state = DatabaseMigrationStateSchedule.getValueAtTime(clock.nowUtc());
+    if (!state.getReplayDirection().equals(ReplayDirection.DATASTORE_TO_SQL)) {
+      String message =
+          String.format(
+              "Skipping ReplayCommitLogsToSqlAction because we are in migration phase %s.", state);
+      logger.atInfo().log(message);
       // App Engine will retry on any non-2xx status code, which we don't want in this case.
       response.setStatus(SC_NO_CONTENT);
       response.setPayload(message);
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1502,22 +1502,6 @@ public final class RegistryConfig {
     CONFIG_SETTINGS.get().cloudSql.replicateTransactions = replicateTransactions;
   }
 
-  /**
-   * Returns whether or not to replay commit logs to the SQL database after export to GCS.
-   *
-   * <p>If true, we will trigger the {@link google.registry.backup.ReplayCommitLogsToSqlAction}
-   * after the {@link google.registry.backup.ExportCommitLogDiffAction} to load the commit logs and
-   * replay them to SQL.
-   */
-  public static boolean getCloudSqlReplayCommitLogs() {
-    return CONFIG_SETTINGS.get().cloudSql.replayCommitLogs;
-  }
-
-  @VisibleForTesting
-  public static void overrideCloudSqlReplayCommitLogs(boolean replayCommitLogs) {
-    CONFIG_SETTINGS.get().cloudSql.replayCommitLogs = replayCommitLogs;
-  }
-
   /** Returns the roid suffix to be used for the roids of all contacts and hosts. */
   public static String getContactAndHostRoidSuffix() {
     return CONFIG_SETTINGS.get().registryPolicy.contactAndHostRoidSuffix;
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -127,7 +127,6 @@ public class RegistryConfigSettings {
     public String username;
     public String instanceConnectionName;
     public boolean replicateTransactions;
-    public boolean replayCommitLogs;
   }
 
   /** Configuration for Apache Beam (Cloud Dataflow). */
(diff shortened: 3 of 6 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 89 at r1](https://reviewable.io/reviews/google/nomulus/1175#-MbD9xPnEpUw-A0fzhk0:-MbD9xPoB7ftvm_sCewo:bj692a7) ([raw file](https://github.com/google/nomulus/blob/687cf815803c3a0462edbbcce3d44365c98abd65/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L89)):*
> ```Java
>       String message =
>           String.format(
>               "ReplayCommitLogsToSqlAction was called in the invalid migration state %s.",
> ```

This  is a little bit of a strong message and logging level.

This is the expected happy path for a decent amount of time, right?  We'll configure the cronjob first, and remove it after, the actual migration is supposed to take place.  So this isn't an invalid state that needs to be warned about, it's expected.  So I'd say log it at `INFO` and tone down the message more along the lines of "Skipping commit log replay to SQL because we're not in that phase of the migration."

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 85 at r1](https://reviewable.io/reviews/google/nomulus/1175#-MbDBSov3KZp_qQ5JhWg:-MbDBSov3KZp_qQ5JhWh:baqbele) ([raw file](https://github.com/google/nomulus/blob/687cf815803c3a0462edbbcce3d44365c98abd65/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L85)):*
> ```Java
>   public void run() {
>     MigrationState currentState =
>         DatabaseMigrationStateSchedule.get().getValueAtTime(clock.nowUtc());
> ```

The way this looks at the callsite is a little bit overly verbose (and this is by far the most common usage pattern, right?).  You might consider making the API looking like the following:

```
// Seems like we definitely want because so many actions declare their own Clocks for testing purposes.
MigrationState currentState = DatabaseMigrationStateSchedule.getStateAt(clock.nowUtc());

// We may also want this (if it ends up being used) for callsites that don't have Clocks; this would simply
// call DateTime.now(DateTimeZone.UTC) internally.
MigrationState currentState = DatabaseMigrationStateSchedule.getCurrentState();
```

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 71 at r1](https://reviewable.io/reviews/google/nomulus/1175#-MbD72y3-rtTNVDnAha1:-MbD72y3-rtTNVDnAha2:b-zag0em) ([raw file](https://github.com/google/nomulus/blob/687cf815803c3a0462edbbcce3d44365c98abd65/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L71)):*
> ```Java
>   private static final Duration LEASE_LENGTH = standardHours(1);
>   private static final ImmutableSet<MigrationState> VALID_MIGRATION_STATES =
>       ImmutableSet.of(MigrationState.DATASTORE_PRIMARY, MigrationState.DATASTORE_PRIMARY_READ_ONLY);
> ```

Optional: You might want to make this part of `MigrationState` itself in the interests of keeping all of the details of migration state semantics in one place.  You could then have a `MigrationState` method like `replicateToSql()` that just returns true or false as appropriate.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: 3 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 71 at r1](https://reviewable.io/reviews/google/nomulus/1175#-MbD72y3-rtTNVDnAha1:-MbDUDRf50iAe72QWxuu:bxeqtnn) ([raw file](https://github.com/google/nomulus/blob/687cf815803c3a0462edbbcce3d44365c98abd65/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L71)):*

Optional: You might want to make this part of `MigrationState` itself in the interests of keeping all of the details of migration state semantics in one place.  You could then have a `MigrationState` method like `replicateToSql()` that just returns true or false as appropriate.
</blockquote></details>

Yeah, seems reasonable. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 85 at r1](https://reviewable.io/reviews/google/nomulus/1175#-MbDBSov3KZp_qQ5JhWg:-MbDCQXU9Fxo_3QVlaQY:bo71rst) ([raw file](https://github.com/google/nomulus/blob/687cf815803c3a0462edbbcce3d44365c98abd65/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L85)):*

The way this looks at the callsite is a little bit overly verbose (and this is by far the most common usage pattern, right?).  You might consider making the API looking like the following:

```
// Seems like we definitely want because so many actions declare their own Clocks for testing purposes.
MigrationState currentState = DatabaseMigrationStateSchedule.getStateAt(clock.nowUtc());

// We may also want this (if it ends up being used) for callsites that don't have Clocks; this would simply
// call DateTime.now(DateTimeZone.UTC) internally.
MigrationState currentState = DatabaseMigrationStateSchedule.getCurrentState();
```
</blockquote></details>

I actually only anticipate this being called at three places 

1. here
2. the equivalent but reverse async replay action
3. in TransactionManagerFactory::tm

I think we can rely on passing in the DateTime into the DatabaseMigrationSchedule. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 89 at r1](https://reviewable.io/reviews/google/nomulus/1175#-MbD9xPnEpUw-A0fzhk0:-MbDPos-5LO1lQf9OAb5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/687cf815803c3a0462edbbcce3d44365c98abd65/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L89)):*

This  is a little bit of a strong message and logging level.

This is the expected happy path for a decent amount of time, right?  We'll configure the cronjob first, and remove it after, the actual migration is supposed to take place.  So this isn't an invalid state that needs to be warned about, it's expected.  So I'd say log it at `INFO` and tone down the message more along the lines of "Skipping commit log replay to SQL because we're not in that phase of the migration."
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 67 at r2](https://reviewable.io/reviews/google/nomulus/1175#-MbDlaDV8w328k2H8hHC:-MbDlaDW4Pnt4uloSj5q:b-94hdlz) ([raw file](https://github.com/google/nomulus/blob/3c33613073e849b1d5074a1b5c698e251edb8009/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L67)):*
> ```Java
>     private final PrimaryDatabase primaryDatabase;
>     private final boolean isReadOnly;
>     private final boolean shouldReplayToSql;
> ```

Rather than modeling this as two booleans, which is easy to get wrong (e.g. `true, true` is invalid), I think it makes more sense to model it as a three-state enum something like `enum MigrationDirection{ NONE, DATASTORE_TO_SQL, SQL_TO_DATASTORE}`.  (Not 100% sure I love all those names but you get the gist.)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 67 at r2](https://reviewable.io/reviews/google/nomulus/1175#-MbDlaDV8w328k2H8hHC:-MbE4UX21UZcgXlM--hL:bgs8yup) ([raw file](https://github.com/google/nomulus/blob/3c33613073e849b1d5074a1b5c698e251edb8009/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L67)):*

Rather than modeling this as two booleans, which is easy to get wrong (e.g. `true, true` is invalid), I think it makes more sense to model it as a three-state enum something like `enum MigrationDirection{ NONE, DATASTORE_TO_SQL, SQL_TO_DATASTORE}`.  (Not 100% sure I love all those names but you get the gist.)
</blockquote></details>

yep, i came to a conclusion like that shortly after pushing this (was gone unfortunately). 

Done.

mindhog (approved):
> Reviewed 1 of 3 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1175)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1181, opened 2021-05-26
Make ExpandRecurringBillingEventAction SQL-aware

jianglai:
There is some complication regarding how the
CancellationMatchingBillingEvent of the generated OneTime can be
reconstructed when loading from SQL. I decided to only address it in
testing as there is no real value to fully reconstruct this VKey in
production where we are either in SQL or Ofy mode, both never in both.
Therefore the VKey in a particular mode only needs to contain the
corresponding key in order to function.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1181)


mindhog (commented):
So, I haven't really gotten into this review yet, but just from your comments here: be very careful about introducing "one-sided" (asymmetric) VKeys.  They can exist, but if they end up in any replicated entities they won't be persisted correctly to the other database.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181#-:-MacR71wEgvD4VwhSHnm:b-2tkq2b)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @gbrodman)

jianglai (commented):
Hmm, I have not considered the effects on async replay. However this PR did not introduce the asymmetric VKey. It merely didn’t fix it… And this would only affect the replay back to datastore when sql is primary. Do you think it is worth it to go through the trouble to introduce a column and resave all relevant entities to properly fix it?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181#-:-MacpeI36QRUKnDJgsAz:b-qkxaba)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @gbrodman)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181)* status: 3 of 5 files reviewed, 10 unresolved discussions (waiting on @gbrodman, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 91 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MachDTDBUTxb7Nt4rgJ:-MadD-SB4NwOKLVM06dx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L91)):*

It's generally recommended to define constants for string literals, but can we not do this for SQL statements that are executed once?  It makes it very hard to follow when the query is defined in one place and parameterized and evaluated in another.

Also, I think you can simplify this query string to `"FROM BillingRecurrence WHERE event_time <= :executeTime AND event_time < recurrence_end_time"`
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1181#-Macm7tZBnCjyqDJDcRQ:-MadD3dTEoQSOCyAokKd:bod4ri3) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L138)):*

Is there a reason for collecting the stream to a list and then converting it back to a stream?  If so, this merits a comment.
</blockquote></details>

lol. My bad. I copied over the existing code which collects the steam into a list and then realized I need to do more with it. Totally forgot that it didn't need to be turned into a list...
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MacmrA11WvaxY9owmfl:-MadDY0-71taxTY_0y4E:b6nx1n7) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L150)):*

> ```
> [%s, %s).
> ```

Mix of brackets and parens here.
</blockquote></details>

I believe this is intentional as the the left range is inclusive (this is again copied over from the original code).
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 260 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MacvA-8ApbiCfHsfuqN:-MadEFJADzCHv1QUKi0C:b-uvkfub) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L260)):*

Can we change this to auditedOfy()?  This way we won't mistake this for code that still needs to be converted.  Or better yet, I think you can just use ofyTm().loadByKeyIfPresent() here.
</blockquote></details>

I think `auditedOfy()` is fine here. The reducer will be deleted after we migrate, so there's no need to migrate it.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1181#-Macy_zsAqfMLl9tRlKL:-MadEPzy6pbQGDg9evyT:b-896fix) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L300)):*

auditedOfy()
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/DatabaseHelper.java, line 798 at r1](https://reviewable.io/reviews/google/nomulus/1181#-Mad2Hoz89EBFGbDsdTk:-MadEiLSCze7OjDiEZAq:b-8j9v97) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/test/java/google/registry/testing/DatabaseHelper.java#L798)):*

Then it's just broken :-/

If we don't restore the key correctly, we won't be able to replicate it asynchronously in at least one direction.  Let's take a closer look at this.  I think we've introduced columns to deal with this case before.
</blockquote></details>

You may be referring to `domainHistoryRevisionId` in `BillingEvent` that (AFAICT) was only used to reconstruct the parent `VKey<DomainHistory>` in the `@PostLoad` listener. So we will need to add another `recurringEventDomainHistoryRevisionId` or something to that effect in `OneTime` because `VKey<Recurring>`'s ancestor path is `Recurring -> DomainHistory -> DomainBase`.

Actually now that I realize that we don't have to do a full resave because `OneTime` is populated in SQL yet. I think adding a column is worth it in this case. It is the price for a smooth migration I figure.

mindhog (changes requested):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181)* status: all files reviewed, 7 unresolved discussions (waiting on @gbrodman, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MacmrA11WvaxY9owmfl:-MbBzJijDrVR74dkyGhI:betw6vj) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L150)):*

I believe this is intentional as the the left range is inclusive (this is again copied over from the original code).
</blockquote></details>

That's a fairly obscure convention, it's probably better to make that more explicit, e.g. `%s <= x < %s` or "(%s (inclusive), %s)".
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 407 at r1](https://reviewable.io/reviews/google/nomulus/1181#-Mad-dCj7wPqze_cCBKA:-MbByxcl-lTn083zZ2Ck:b-d0upn2) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L407)):*

With a symmetric VKey, this should just work.
</blockquote></details>

Looks like you've addressed this.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 129 at r2](https://reviewable.io/reviews/google/nomulus/1181#-MbBy_Zc18UwA8bVA3iX:-MbBy_Zd9UNmMNdhCkfO:b7apch8) ([raw file](https://github.com/google/nomulus/blob/2ef853e79280be196d22328b78aba0bf20eb5a3b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L129)):*
> ```Java
>                       jpaTm()
>                           .query(
>                               "FROM BillingRecurrence r "
> ```

I don't think you need the "r" in this query, right?
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 328 at r2](https://reviewable.io/reviews/google/nomulus/1181#-MbC-Nqq4Dkxko-KRS1B:-MbC-Nqq4Dkxko-KRS1C:bfqmpfe) ([raw file](https://github.com/google/nomulus/blob/2ef853e79280be196d22328b78aba0bf20eb5a3b/core/src/main/java/google/registry/model/billing/BillingEvent.java#L328)):*
> ```Java
>     /**
>      * For {@link Flag#SYNTHETIC} events, a {@link Key} to the {@link Recurring} from which
>      * this @{@link OneTime} was created. This is needed in order to properly match billing events
> ```

I think you want to omit the initial "@" sign (the one before the brackets)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181)* status: 2 of 11 files reviewed, 7 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 142 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MacoEVz7d0ZmE_6qSk2:-MbCx-8-BjisvxCx_o5b:b-x6ktlz) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L142)):*

Do we have any idea how well this will perform against a production-sized dataset?  IIUIC there are limits to the amount of data you can save in a transaction.
</blockquote></details>

With the new detach-on-load logic this part doesn't work anymore because when multiple recurring billing events are returned from the query there's a chance that they would have generated identical one time billing events (this was tested in a test method), which means `expandBillingEvent` would have loaded the same `OneTime` that it just saved when processing a previous recurring event, causing the detachment to fail.

Looking at how App Engine mapreduce worked, I don't think there's any guarantee that all the recurring events (inputs to the map function) are processed within a transaction. So it makes sense to break out this stream so that each `expandBillingEvent` is executed in its own transaction, just like how mapreduce does it.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 150 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MacmrA11WvaxY9owmfl:-MbC8NIB4KYwVGwVtJLr:bh3zahi) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L150)):*

That's a fairly obscure convention, it's probably better to make that more explicit, e.g. `%s <= x < %s` or "(%s (inclusive), %s)".
</blockquote></details>

I don't think it's obscure, per se. It's pretty standard math expression for intervals and more succinct than using inequalities: https://en.wikipedia.org/wiki/Interval_(mathematics)
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 155 at r1](https://reviewable.io/reviews/google/nomulus/1181#-MacnZ2OBDP5xHEQ1JzA:-MbD0AB69_r58NJeVsNA:b-s9qc2h) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L155)):*

Can we just run the entire sequence (load cursor, expand events, save cursor) in a transaction when in SQL mode?  It seems like the bulk of the time here will be spent expanding the billing events, so if we're not breaking that out into multiple transactions, it doesn't seem useful to do so for the cursor processing.
</blockquote></details>

It's moot now. Each recurring events are expanded within their own transactions.
___
*[core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java, line 129 at r2](https://reviewable.io/reviews/google/nomulus/1181#-MbBy_Zc18UwA8bVA3iX:-MbC5K4i5QUZ_HdN9GZQ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2ef853e79280be196d22328b78aba0bf20eb5a3b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L129)):*

I don't think you need the "r" in this query, right?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 328 at r2](https://reviewable.io/reviews/google/nomulus/1181#-MbC-Nqq4Dkxko-KRS1B:-MbC8wxu7XIdCjB93jCI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2ef853e79280be196d22328b78aba0bf20eb5a3b/core/src/main/java/google/registry/model/billing/BillingEvent.java#L328)):*

I think you want to omit the initial "@" sign (the one before the brackets)
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/batch/ExpandRecurringBillingEventsActionTest.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1181#-Mad3AdP2BpX2AJN2LXu:-MbD0N6Q-HRADbYU8-ym:b-vl4csm) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/test/java/google/registry/batch/ExpandRecurringBillingEventsActionTest.java#L76)):*

I think we should add ReplayExtension to this class.  If I'm thinking about this correctly, it might actually work right out of the box  for this PR but will (in its current implementation) fail when replaying from SQL to datastore. 

It might actually be more useful for us to try to simulate the reverse replay by running ReplicateToDatastoreAction here, though I'm not sure if you want to be the first to open that particular can of worms :-)
</blockquote></details>

Tried to add the extension but it resulted in a lot of commit log failures. Let me try to figure out what's wrong but I don't want to hold up this review, so please review it as is for now.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181)* status: 3 of 11 files reviewed, 7 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/test/java/google/registry/batch/ExpandRecurringBillingEventsActionTest.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1181#-Mad3AdP2BpX2AJN2LXu:-MbEeMcd7ynJajPSWJjy:b-5vik46) ([raw file](https://github.com/google/nomulus/blob/57ffa9cd2488f0e1823baf2b341ca84ac0ab288b/core/src/test/java/google/registry/batch/ExpandRecurringBillingEventsActionTest.java#L76)):*

Tried to add the extension but it resulted in a lot of commit log failures. Let me try to figure out what's wrong but I don't want to hold up this review, so please review it as is for now.
</blockquote></details>

Done. Fixed all the timestamp inversion issues.

mindhog (approved):
> Reviewed 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1181)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @mindhog)

PR #1183, opened 2021-05-27
Detach result objects obtained through jpaTm().query()

mindhog:
Detach all entity result objects obtained through `TypedQuery`s returned by the `jpaTm().query()` method.

This is in line with the general practice of always detaching objects on load, which prevents issues with implicit saves and immutability violations.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1183)

### DIFF
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java
@@ -40,8 +40,8 @@ public interface JpaTransactionManager extends TransactionManager {
    * <p>This is a convenience method for the longer <code>
    * jpaTm().getEntityManager().createQuery(...)</code>.
    *
-   * <p>Note that while this method can legally be used for queries that return results, <ul>it
-   * should not be</ul>, as it does not correctly detach entities as must be done for nomulus model
+   * <p>Note that while this method can legally be used for queries that return results, <u>it
+   * should not be</u>, as it does not correctly detach entities as must be done for nomulus model
    * objects.
    */
   Query query(String sqlString);
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java
@@ -35,10 +35,14 @@ public interface JpaTransactionManager extends TransactionManager {
   <T> TypedQuery<T> query(String sqlString, Class<T> resultClass);
 
   /**
-   * Creates a JPA SQL query for the given query string (which does not return results).
+   * Creates a JPA SQL query for the given query string.
    *
    * <p>This is a convenience method for the longer <code>
    * jpaTm().getEntityManager().createQuery(...)</code>.
+   *
+   * <p>Note that while this method can legally be used for queries that return results, <ul>it
+   * should not be</ul>, as it does not correctly detach entities as must be done for nomulus model
+   * objects.
    */
   Query query(String sqlString);
 
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -45,6 +45,7 @@ import google.registry.persistence.VKey;
 import google.registry.util.Clock;
 import google.registry.util.Retrier;
 import google.registry.util.SystemSleeper;
+import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.util.Calendar;
 import java.util.Collections;
@@ -125,9 +126,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
   @Override
   public <T> TypedQuery<T> query(String sqlString, Class<T> resultClass) {
     return new TransformingTypedQuery(
-        getEntityManager().createQuery(sqlString, resultClass),
-        // TODO: replace with "detach"
-        x -> x);
+        getEntityManager().createQuery(sqlString, resultClass), this::detachIfEntity);
   }
 
   @Override
@@ -674,6 +673,31 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
         throw e;
       }
     }
+  }
+
+  @Nullable
+  private <T> T detachIfEntity(@Nullable T object) {
+    if (object != null) {
+
+      // Check if the object is an array, if so we'll want to recurse through the elements.
+      if (object.getClass().isArray()) {
+        for (int i = 0; i < Array.getLength(object); ++i) {
+          detachIfEntity(Array.get(object, i));
+        }
+        return object;
+      }
+
+      // Check to see if it is an entity.
+      try {
+        getEntityManager().getMetamodel().entity(object.getClass());
+      } catch (IllegalArgumentException e) {
+        // The object is not an entity.  Return without detaching.
+        return object;
+      }
+    }
+
+    // At this point, object is either an entity or null.
+    return detach(object);
   }
 
   /** Detach the entity, suitable for use in Optional.map(). */
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -677,26 +677,28 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
 
   @Nullable
   private <T> T detachIfEntity(@Nullable T object) {
-    if (object != null) {
-
-      // Check if the object is an array, if so we'll want to recurse through the elements.
-      if (object.getClass().isArray()) {
-        for (int i = 0; i < Array.getLength(object); ++i) {
-          detachIfEntity(Array.get(object, i));
-        }
-        return object;
-      }
-
-      // Check to see if it is an entity.
-      try {
-        getEntityManager().getMetamodel().entity(object.getClass());
-      } catch (IllegalArgumentException e) {
-        // The object is not an entity.  Return without detaching.
-        return object;
-      }
+    if (object == null) {
+      return null;
     }
 
-    // At this point, object is either an entity or null.
+    // Check if the object is an array, if so we'll want to recurse through the elements.
+    if (object.getClass().isArray()) {
+      for (int i = 0; i < Array.getLength(object); ++i) {
+        detachIfEntity(Array.get(object, i));
+      }
+      return object;
+    }
+
+    // Check to see if it is an entity (queries can return raw column values or counts, so this
+    // could be String, Long, ...).
+    try {
+      getEntityManager().getMetamodel().entity(object.getClass());
+    } catch (IllegalArgumentException e) {
+      // The object is not an entity.  Return without detaching.
+      return object;
+    }
+
+    // At this point, object must be an entity.
     return detach(object);
   }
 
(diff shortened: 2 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1183)* status: all files reviewed, 6 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiFCFC4C3yvQ6VDtHv:-MaiFCFC4C3yvQ6VDtHw:b-aym15w) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java#L44)):*
> ```Java
>    *
>    * <p>Note that while this method can legally be used for queries that return results, <ul>it
>    * should not be</ul>, as it does not correctly detach entities as must be done for nomulus model
> ```

nit, but i don't think `ul` is the right tag here 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 680 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiJgiKDc_lUkekep7Q:-MaiJgiKDc_lUkekep7R:b-178mt7) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L680)):*
> ```Java
>   @Nullable
>   private <T> T detachIfEntity(@Nullable T object) {
>     if (object != null) {
> ```

bit nitty, but we can reduce indentation by just starting this function with 

```
if (object == null) {
  return null;
}
```

then the rest isn't indented so much (thanks TotT that ben linked!)
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 685 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiJ4mN0Vo0mi8DkB4d:-MaiJ4mN0Vo0mi8DkB4e:b-y0u0x5) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L685)):*
> ```Java
>       if (object.getClass().isArray()) {
>         for (int i = 0; i < Array.getLength(object); ++i) {
>           detachIfEntity(Array.get(object, i));
> ```

and they said we'd never use recursion in day-to-day code!
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 693 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiJTCP2wJNei1SCm4Q:-MaiJTCP2wJNei1SCm4R:br17daw) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L693)):*
> ```Java
>       try {
>         getEntityManager().getMetamodel().entity(object.getClass());
>       } catch (IllegalArgumentException e) {
> ```

What can throw this? Or rather, what else could `object` be? 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 802 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiI3h0-Qij9I0nJtI7:-MaiI3h0-Qij9I0nJtI8:b-gr2mqa) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L802)):*
> ```Java
>    */
>   @VisibleForTesting
>   static class TransformingTypedQuery<T> implements TypedQuery<T> {
> ```

`DetachingTypedQuery`? 

Also this may be substantial enough to move into a new class (so we don't end up with a ton of extra classes in JpaTransactionManagerImpl, thoughts? 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 804 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiHurh1tA1xEjGAzgI:-MaiHurh1tA1xEjGAzgJ:bf9a1qg) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L804)):*
> ```Java
>   static class TransformingTypedQuery<T> implements TypedQuery<T> {
> 
>     TypedQuery<T> rep;
> ```

`rep` is vague, I don't know what it's referring to. `delegate`? 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 807 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiHpPyBe7kuWZw3J35:-MaiHpPyBe7kuWZw3J36:b-1kefyx) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L807)):*
> ```Java
>     Function<T, T> detach;
> 
>     public TransformingTypedQuery(TypedQuery<T> rep, Function<T, T> detach) {
> ```

Why do we pass in the function here if it's always just the detacher? 

I realize that it's used differently in the tests, but I'd assume there would have to be a better way (we may not even need that test)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1183)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiFCFC4C3yvQ6VDtHv:-MaiOlWyCK7KuViTqawz:b3r5n7x) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java#L44)):*

nit, but i don't think `ul` is the right tag here 
</blockquote></details>

gah!  Thanks.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 680 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiJgiKDc_lUkekep7Q:-MaiP9VZ-knfmAPe0vTW:b-7rl0jd) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L680)):*

bit nitty, but we can reduce indentation by just starting this function with 

```
if (object == null) {
  return null;
}
```

then the rest isn't indented so much (thanks TotT that ben linked!)
</blockquote></details>

done.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 685 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiJ4mN0Vo0mi8DkB4d:-MaiPBJl56DWpEUnKhrr:b-j6ojyc) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L685)):*

and they said we'd never use recursion in day-to-day code!
</blockquote></details>

This is Google.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 693 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiJTCP2wJNei1SCm4Q:-MaiPaU85FlnHA6a4nsz:brckjbh) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L693)):*

What can throw this? Or rather, what else could `object` be? 
</blockquote></details>

Any of the types that work for column values.  Expanded on the comment.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 802 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiI3h0-Qij9I0nJtI7:-MaiPmEH2S8mItjQE488:b-whvbcw) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L802)):*

`DetachingTypedQuery`? 

Also this may be substantial enough to move into a new class (so we don't end up with a ton of extra classes in JpaTransactionManagerImpl, thoughts? 
</blockquote></details>

Well, it's currently transforming as opposed to necessarily detaching (though the instance variable suggests otherwise).  

I don't feel strongly one way or another about moving this out into the package.  I'm inclined to leave it be until other code needs direct access to it, but happy to break out if you have a preference.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 804 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiHurh1tA1xEjGAzgI:-MaiQ7qdC683etLo_52R:bwwoh48) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L804)):*

`rep` is vague, I don't know what it's referring to. `delegate`? 
</blockquote></details>

"rep" is a convention in some contexts, I guess short for "representation?"  Converted to delegate.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 807 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiHpPyBe7kuWZw3J35:-MaiR2pC1YFWp6JtkRNv:b1d2l8g) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L807)):*

Why do we pass in the function here if it's always just the detacher? 

I realize that it's used differently in the tests, but I'd assume there would have to be a better way (we may not even need that test)
</blockquote></details>

I was a bit squeamish about making this a non-static class, given some of the needs we've seen in the beam pipelines for object serialization.  Injecting also helps with testability (we could probably make more extensive use of that for this class).

But I don't feel strongly about it (unless it breaks stuff), would you like me to change that?

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1183)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 807 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiHpPyBe7kuWZw3J35:-MajwFMZEYCv4CRZ8OHF:b3r9why) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L807)):*

I was a bit squeamish about making this a non-static class, given some of the needs we've seen in the beam pipelines for object serialization.  Injecting also helps with testability (we could probably make more extensive use of that for this class).

But I don't feel strongly about it (unless it breaks stuff), would you like me to change that? 
</blockquote></details>

I'm not sure that keeping this a non-static class has any benefits with regards to serialization -- if we're passing in the parent JpaTransactionManagerImpl object's `detach` method, I think it'll require serializing the parent `JpaTransactionManagerImpl` object, right? I remember seeing something like that a while back

and I think the methods that you added in JPATMIT should check the detaching, right? Specifically `query_detachesResults`

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1183)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 807 at r1](https://reviewable.io/reviews/google/nomulus/1183#-MaiHpPyBe7kuWZw3J35:-Mb718Yp17D2lMe6tPJo:bx68dhp) ([raw file](https://github.com/google/nomulus/blob/04f3e9173ec14c9c2668d62e3460219273d4a5f0/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L807)):*

I'm not sure that keeping this a non-static class has any benefits with regards to serialization -- if we're passing in the parent JpaTransactionManagerImpl object's `detach` method, I think it'll require serializing the parent `JpaTransactionManagerImpl` object, right? I remember seeing something like that a while back

and I think the methods that you added in JPATMIT should check the detaching, right? Specifically `query_detachesResults`
</blockquote></details>

Yes, I think you're right about serializing the detach method.

Converted to non-static, calling detach directly.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1186, opened 2021-05-27
Use a TimedTransitionProperty for the DB migration schedule

gbrodman:
This includes the following changes:
- Convert the single-valued database migration state to a timed
transition property, meaning that we can switch all instances over at
the same time and schedule it in advance
- Use a "cache" (technically an expiring memoized supplier) when
retrieving the database migration state value
- Delete the old DatabaseTransitionSchedule because it is no longer
necessary. We took the idea from that and used it for the new
DatabaseMigrationStateSchedule, though we cannot reuse the entity itself
because the structure is fundamentally different.
- Removed references to the DatabaseTransitionSchedule, mainly in the
getter/setter commands+tests and a few odd references elsewhere.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1186)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Xo397oN4_Ph9Weje:-Mb8Xo397oN4_Ph9Wejf:b-eu84b9) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L61)):*
> ```Java
>     DATASTORE_ONLY(PrimaryDatabase.DATASTORE, false),
>     DATASTORE_PRIMARY(PrimaryDatabase.DATASTORE, false),
>     DATASTORE_PRIMARY_READ_ONLY(PrimaryDatabase.DATASTORE, true),
> ```

Are we sure we need both `READ_ONLY` states?
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 67 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Xx1Q1FPbyjHoLMCG:-Mb8Xx1Q1FPbyjHoLMCH:b3qfegz) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L67)):*
> ```Java
> 
>     private final PrimaryDatabase primaryDatabase;
>     private final boolean readOnly;
> ```

How about `isReadOnly`?
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 99 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Y7mx0v6oMOJYY7k8:-Mb8Y7mx0v6oMOJYY7k9:b-9hjiq8) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L99)):*
> ```Java
> 
>   // Each instance should cache the migration schedule for five minutes before reloading
>   private static final Duration CACHE_INTERVAL = Duration.standardMinutes(5);
> ```

If this is only used in one place, maybe just inline it?  The alternative idea would be to make it configurable, but I don't think that's necessary (also we won't be using this long-term anyway).
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 100 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Y-cV-6hi1xPKzp6C:-Mb8Y-cV-6hi1xPKzp6D:biobel6) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L100)):*
> ```Java
>   // Each instance should cache the migration schedule for five minutes before reloading
>   private static final Duration CACHE_INTERVAL = Duration.standardMinutes(5);
>   // Cache of the current migration schedule. The key is meaningless; this is essentially a memoized
> ```

Missing newline before this.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8YQm35fgZGWC8f52L:-Mb8YQm35fgZGWC8f52M:b-mjz1bt) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L126)):*
> ```Java
>   @VisibleForTesting
>   public static final TimedTransitionProperty<MigrationState, MigrationStateTransition>
>       DATASTORE_ONLY_TRANSITION_MAP =
> ```

How about `DEFAULT_TRANSITION_MAP`?
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 166 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Ymuj9-qDvwItpJCd:-Mb8Ymuj9-qDvwItpJCe:b-k99jvi) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L166)):*
> ```Java
>   }
> 
>   /** Loads the currently-set migration chedule from Datastore. */
> ```

Needs to specify that it uses the default when none exists.  This is more than a basic load operation.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 178 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8YumK3uUhyvNoSgxf:-Mb8YumK3uUhyvNoSgxg:bdtifl) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L178)):*
> ```Java
>   }
> 
>   // A new map can be valid by itself but imply a bad transition at the current time from old->new
> ```

Not immediately clear from reading this what this refers to.  This is in the case when a previous transition in the map has somehow been missed, correct?  Or if the map itself has changed?  Needs more explanation, may as well be a Javadoc.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 193 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8ZKx6CeTMJtPY2JsH:-Mb8ZKx6CeTMJtPY2JsI:b-wzl7qv) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L193)):*
> ```Java
>   // number of steps, as long as the step we're moving back to has the same primary database as the
>   // one we're in. Otherwise, we must move to the corresponding READ_ONLY stage first.
>   private static ImmutableMultimap<MigrationState, MigrationState> createValidStateTransitions() {
> ```

It's probably easiest for readability if this is defined adjacent to where it is called, seeing as how it's only called in one place.

CydeWeys (commented):
+reviewer:@sarahcaseybot 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186#-:-Mb8ZkXdAOOpq4AaJ7J4:b-ar4bh4)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

CydeWeys (commented):
Adding Sarah because this touches on work she's done and the migration she's been shepherding along.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186#-:-Mb8Znrc50oi_5425twH:bqp5qo4)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186)* status: 0 of 23 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Xo397oN4_Ph9Weje:-Mb8sc58-EozcN60FTpX:b9llv8g) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L61)):*

Are we sure we need both `READ_ONLY` states?
</blockquote></details>

Ideally, we don't. However in the situation where we're in `SQL_PRIMARY` and we want to roll back to Datastore being the primary in some way, we might. Assuming that the replay job that asynchronously replays SQL commits to Datastore is active and working, we'd want to give that time to finish running before shifting back to Datastore. 
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 67 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Xx1Q1FPbyjHoLMCG:-Mb8speS-C2ABW2_Szkl:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L67)):*

How about `isReadOnly`?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 99 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Y7mx0v6oMOJYY7k8:-Mb8sqUi7jstHP817NNm:bq60w0p) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L99)):*

If this is only used in one place, maybe just inline it?  The alternative idea would be to make it configurable, but I don't think that's necessary (also we won't be using this long-term anyway).
</blockquote></details>

Sure, I doubt we'll configure it.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 100 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Y-cV-6hi1xPKzp6C:-Mb8ss4R8vuCZ9pNIWRC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L100)):*

Missing newline before this.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8YQm35fgZGWC8f52L:-Mb8t8020EJ_jPC_aWWN:bb74ce2) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L126)):*

How about `DEFAULT_TRANSITION_MAP`?
</blockquote></details>

Sure
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 166 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Ymuj9-qDvwItpJCd:-Mb8tDdF9tspSvbAjhTW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L166)):*

Needs to specify that it uses the default when none exists.  This is more than a basic load operation.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 178 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8YumK3uUhyvNoSgxf:-Mb8tlqR5MAqu3b3HgB1:b-6bm90p) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L178)):*

Not immediately clear from reading this what this refers to.  This is in the case when a previous transition in the map has somehow been missed, correct?  Or if the map itself has changed?  Needs more explanation, may as well be a Javadoc.
</blockquote></details>

Done. 

The transition of old_map(now) -> new_map(now) must also be valid.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 193 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8ZKx6CeTMJtPY2JsH:-Mb8twvIBZcCMRmVIdP0:b-5buo2u) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L193)):*

It's probably easiest for readability if this is defined adjacent to where it is called, seeing as how it's only called in one place.
</blockquote></details>

Done. 

It's less clear how to order things in a class like this relative to a usual, simple, class :)

sarahcaseybot (changes requested):
> Reviewed 13 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186)* status: 13 of 23 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/DatabaseMigrationUtils.java, line 26 at r2](https://reviewable.io/reviews/google/nomulus/1186#-Mb9CDnfBV_RHEZ2hudw:-Mb9CDng0OdsUW2jEbO1:brvzzoe) ([raw file](https://github.com/google/nomulus/blob/66f79972db1f4d896b15ff321304539c5d289e8e/core/src/main/java/google/registry/model/DatabaseMigrationUtils.java#L26)):*
> ```Java
> 
>   /** Throws exceptions only in unit tests, otherwise only logs exceptions. */
>   public static void suppressExceptionUnlessInTest(Runnable work, String message) {
> ```

I don't think we are using this method anywhere anymore, so this whole class could probably be deleted

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186)* status: 13 of 23 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 166 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Ymuj9-qDvwItpJCd:-MbC8UH2Coh5RhJUrNaS:b-oz5qhx) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L166)):*

Done.
</blockquote></details>

s/of/if/
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 121 at r2](https://reviewable.io/reviews/google/nomulus/1186#-MbC7ed34pgL1Vlpm6Tv:-MbC7ed44eS9Xb9w6A_D:b-8wutt4) ([raw file](https://github.com/google/nomulus/blob/66f79972db1f4d896b15ff321304539c5d289e8e/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L121)):*
> ```Java
>       createValidStateTransitions();
> 
>   // The valid state transitions. Generally, one can advance the state one step or move backward any
> ```

As a general comment on this PR, you're using multi-line `//` comments in places that really could be Javadoc comments instead, which provide some benefits in Javadoc-intelligibility and IDE support.  A reasonable rule of thumb would be that, if it looks like it could be a Javadoc comment, it probably should be one.  See also: https://softwareengineering.stackexchange.com/a/267307

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186#-:-MbC8hTN3WhN3lMqActb:bnfp4nl)* status: 13 of 23 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186)* status: 12 of 23 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/DatabaseMigrationUtils.java, line 26 at r2](https://reviewable.io/reviews/google/nomulus/1186#-Mb9CDnfBV_RHEZ2hudw:-MbC9IklDaYof_1WWK9B:b-896fix) ([raw file](https://github.com/google/nomulus/blob/66f79972db1f4d896b15ff321304539c5d289e8e/core/src/main/java/google/registry/model/DatabaseMigrationUtils.java#L26)):*

I don't think we are using this method anywhere anymore, so this whole class could probably be deleted
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 166 at r1](https://reviewable.io/reviews/google/nomulus/1186#-Mb8Ymuj9-qDvwItpJCd:-MbC9P8eBHFfZbA7XQgO:bb7r7cw) ([raw file](https://github.com/google/nomulus/blob/e2f98862fe65862a413001f628d7955720b8eb58/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L166)):*

s/of/if/
</blockquote></details>

womp
___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 121 at r2](https://reviewable.io/reviews/google/nomulus/1186#-MbC7ed34pgL1Vlpm6Tv:-MbC9dPXDno6dEn5UAHH:b-dqya1f) ([raw file](https://github.com/google/nomulus/blob/66f79972db1f4d896b15ff321304539c5d289e8e/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L121)):*

As a general comment on this PR, you're using multi-line `//` comments in places that really could be Javadoc comments instead, which provide some benefits in Javadoc-intelligibility and IDE support.  A reasonable rule of thumb would be that, if it looks like it could be a Javadoc comment, it probably should be one.  See also: https://softwareengineering.stackexchange.com/a/267307
</blockquote></details>

Fair. I generally haven't thought that it mattered for private fields / methods. Changed.

sarahcaseybot (approved):
> Reviewed 8 of 23 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1186)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

PR #1187, opened 2021-06-02
Handle cases where periodYears is NULL in a OneTime

jianglai:
There are cases where periodYears is not set when creating a OneTime
billing event, for example when performing a registry lock (default cost = $0)
or when performing a server status update, such as applying the
serverUpdateProhibited status (default cost = $20). This is not currently
handled correctly in the billing pipeline because the parseFromRecord
method checks for nullness for all fields. Even if it does not validate
the fields, the null periodYears will still cause problem when the
billing event is converted to CSV files.

This PR alters the BigQuery SQL file to convert a NULL to 0 when
creating the BillingEvent in the invoicing pipeline. It also sets the EndDate
in the invoice CSV to an empty string when periodYears is 0. Note that when the
cost is also 0, the billing event is filtered out in the invoice CSV so only
the non-free OneTime with null periodYear will have an impact on the output.
For detailed reports all billing events are included and the zero
periodYears is printed as is.

Setting the EndDate to empty is the correct behavior per
go/manual-integration-csv#end-date.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1187)

### DIFF
--- a/core/src/main/java/google/registry/beam/invoicing/BillingEvent.java
+++ b/core/src/main/java/google/registry/beam/invoicing/BillingEvent.java
@@ -251,7 +251,14 @@ public abstract class BillingEvent implements Serializable {
   InvoiceGroupingKey getInvoiceGroupingKey() {
     return new AutoValue_BillingEvent_InvoiceGroupingKey(
         billingTime().toLocalDate().withDayOfMonth(1).toString(),
-        billingTime().toLocalDate().withDayOfMonth(1).plusYears(years()).minusDays(1).toString(),
+        years() == 0
+            ? ""
+            : billingTime()
+                .toLocalDate()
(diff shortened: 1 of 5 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1187)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

CydeWeys (approved):
Resolved, we're already including them in the detail report CSV and that's not changing.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1187#-:-MbCFv3G1vgYcLwj3VJi:b-1qni6j)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1191, opened 2021-06-03
Use DB migration state to determine running async replay SQL->DS

gbrodman:
The SQL->DS replay likely could use more work (locking, returning the
right codes, things like that) but that's outside the scope of this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1191)


mindhog (approved):
There are likely improvements, but we don't need locking for this: even if Cloud Scheduler allows an overrun (it's documentation says it doesn't, so this would be a bug) the changes are applied transactionally.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1191#-:-MbNWaj2EpXLISlovrSY:bk82i8v)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1194, opened 2021-06-05
Convert EppResourceUtils::loadAtPointInTime to SQL+DS

gbrodman:
This required the following changes:
- The branching / conversion logic itself, where we load the most recent
history object for the resource in question (or just return the resource
itself)
- For simplicity's sake, adding a method in the *History objects that
returns the generic resource -- this means that it can be called when we
don't know or care which subclass it is.
- Populating the domain's dsData and gracePeriods fields from the
DomainHistory fields, and adding factories in the relevant classes to
allow us to do the conversions nicely (the history classes are almost
the same as the regular ones, but not quite).
- Change the tests to use the clocks properly and to allow comparison of
e.g. DomainContent to DomainBase. The objects aren't the same (one is a
superclass of the other) but the fields are.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1194)

### DIFF
--- a/core/src/main/java/google/registry/model/EppResourceUtils.java
+++ b/core/src/main/java/google/registry/model/EppResourceUtils.java
@@ -17,10 +17,10 @@ package google.registry.model;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
-import static google.registry.model.ofy.ObjectifyService.ofy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
+import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static google.registry.util.DateTimeUtils.isAtOrAfter;
 import static google.registry.util.DateTimeUtils.isBeforeOrAt;
 import static google.registry.util.DateTimeUtils.latestOf;
@@ -43,10 +43,13 @@ import google.registry.model.index.ForeignKeyIndex;
 import google.registry.model.ofy.CommitLogManifest;
 import google.registry.model.ofy.CommitLogMutation;
 import google.registry.model.registry.Registry;
+import google.registry.model.reporting.HistoryEntry;
+import google.registry.model.reporting.HistoryEntryDao;
 import google.registry.model.transfer.DomainTransferData;
 import google.registry.model.transfer.TransferData;
 import google.registry.model.transfer.TransferStatus;
 import google.registry.persistence.VKey;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -266,26 +269,43 @@ public final class EppResourceUtils {
    * Rewinds an {@link EppResource} object to a given point in time.
    *
    * <p>This method costs nothing if {@code resource} is already current. Otherwise it needs to
-   * perform a single asynchronous key fetch operation.
+   * perform a single fetch operation.
    *
    * <p><b>Warning:</b> A resource can only be rolled backwards in time, not forwards; therefore
    * {@code resource} should be whatever's currently in Datastore.
    *
-   * <p><b>Warning:</b> Revisions are granular to 24-hour periods. It's recommended that
-   * {@code timestamp} be set to midnight. Otherwise you must take into consideration that under
-   * certain circumstances, a resource might be restored to a revision on the previous day, even if
-   * there were revisions made earlier on the same date as {@code timestamp}; however, a resource
-   * will never be restored to a revision occurring after {@code timestamp}. This behavior is due to
-   * the way {@link google.registry.model.translators.CommitLogRevisionsTranslatorFactory
+   * <p><b>Warning:</b> In Datastore, revisions are granular to 24-hour periods. It's recommended
+   * that {@code timestamp} be set to midnight. If you don't use midnight, you must take into
+   * consideration that under certain circumstances, a resource might be restored to a revision on
+   * the previous day, even if there were revisions made earlier on the same date as {@code
+   * timestamp}; however, a resource will never be restored to a revision occurring after {@code
+   * timestamp}. This behavior is due to the way {@link
+   * google.registry.model.translators.CommitLogRevisionsTranslatorFactory
    * CommitLogRevisionsTranslatorFactory} manages the {@link EppResource#revisions} field. Please
    * note however that the creation and deletion times of a resource are granular to the
    * millisecond.
    *
+   * <p>Example: a resource in Datastore has three revisions A, B, and C
+   *
+   * <ul>
+   *   <li>A: Day 0, 1pm
+   *   <li>B: Day 1, 1pm
+   *   <li>C: Day 1, 3pm
+   * </ul>
+   *
+   * <p>If one requests the resource as of day 1 at 2pm, we will return revision A because as far as
+   * the commit logs are concerned, revision C completely overwrites the existence of revision B.
+   *
+   * <p>When using the SQL backend (post-Registry-3.0-migration) this restriction goes away and
+   * objects can be restored to any revision.
+   *
+   * <p>TODO(b/177567432): Once Datastore is completely removed, remove the Result wrapping.
+   *
    * @return an asynchronous operation returning resource at {@code timestamp} or {@code null} if
    *     resource is deleted or not yet created
    */
-  public static <T extends EppResource>
-      Result<T> loadAtPointInTime(final T resource, final DateTime timestamp) {
+  public static <T extends EppResource> Result<T> loadAtPointInTime(
+      final T resource, final DateTime timestamp) {
     // If we're before the resource creation time, don't try to find a "most recent revision".
     if (timestamp.isBefore(resource.getCreationTime())) {
       return new ResultNow<>(null);
@@ -300,7 +320,8 @@ public final class EppResourceUtils {
             : loadMostRecentRevisionAtTime(resource, timestamp);
     return () -> {
       T loadedResource = loadResult.now();
-      return (loadedResource == null) ? null
+      return (loadedResource == null)
+          ? null
           : (isActive(loadedResource, timestamp)
               ? cloneProjectedAtTime(loadedResource, timestamp)
               : null);
@@ -308,26 +329,43 @@ public final class EppResourceUtils {
   }
 
   /**
-   * Returns an asynchronous result holding the most recent Datastore revision of a given
-   * EppResource before or at the provided timestamp using the EppResource revisions map, falling
-   * back to using the earliest revision or the resource as-is if there are no revisions.
+   * Returns an asynchronous result holding the most recent revision of a given EppResource before
+   * or at the provided timestamp, falling back to using the resource as-is if there are no
+   * revisions.
    *
    * @see #loadAtPointInTime(EppResource, DateTime)
    */
   private static <T extends EppResource> Result<T> loadMostRecentRevisionAtTime(
       final T resource, final DateTime timestamp) {
+    if (tm().isOfy()) {
+      return loadMostRecentRevisionAtTimeDatastore(resource, timestamp);
+    } else {
+      return loadMostRecentRevisionAtTimeSql(resource, timestamp);
+    }
+  }
+
+  /**
+   * Returns an asynchronous result holding the most recent Datastore revision of a given
+   * EppResource before or at the provided timestamp using the EppResource revisions map, falling
+   * back to using the resource as-is if there are no revisions.
+   *
+   * @see #loadAtPointInTime(EppResource, DateTime)
+   */
+  private static <T extends EppResource> Result<T> loadMostRecentRevisionAtTimeDatastore(
+      final T resource, final DateTime timestamp) {
     final Key<T> resourceKey = Key.create(resource);
-    final Key<CommitLogManifest> revision = findMostRecentRevisionAtTime(resource, timestamp);
+    final Key<CommitLogManifest> revision =
+        findMostRecentDatastoreRevisionAtTime(resource, timestamp);
     if (revision == null) {
       logger.atSevere().log("No revision found for %s, falling back to resource.", resourceKey);
       return new ResultNow<>(resource);
     }
     final Result<CommitLogMutation> mutationResult =
-        ofy().load().key(CommitLogMutation.createKey(revision, resourceKey));
+        auditedOfy().load().key(CommitLogMutation.createKey(revision, resourceKey));
     return () -> {
       CommitLogMutation mutation = mutationResult.now();
       if (mutation != null) {
-        return ofy().load().fromEntity(mutation.getEntity());
+        return auditedOfy().load().fromEntity(mutation.getEntity());
       }
       logger.atSevere().log(
           "Couldn't load mutation for revision at %s for %s, falling back to resource."
@@ -337,9 +375,37 @@ public final class EppResourceUtils {
     };
   }
 
+  /**
+   * Returns an asynchronous result holding the most recent SQL revision of a given EppResource
+   * before or at the provided timestamp using *History objects, falling back to using the resource
+   * as-is if there are no revisions.
+   *
+   * @see #loadAtPointInTime(EppResource, DateTime)
+   */
+  @SuppressWarnings("unchecked")
+  private static <T extends EppResource> Result<T> loadMostRecentRevisionAtTimeSql(
+      T resource, DateTime timestamp) {
+    T resourceAtPointInTime =
+        (T)
+            HistoryEntryDao.loadHistoryObjectsForResource(
+                    resource.createVKey(), START_OF_TIME, timestamp)
+                .stream()
+                .max(Comparator.comparing(HistoryEntry::getModificationTime))
+                .flatMap(HistoryEntry::getResourceAtPointInTime)
+                .orElse(null);
+    if (resourceAtPointInTime == null) {
+      logger.atSevere().log(
+          "Couldn't load resource at % for key %s, falling back to resource %s.",
+          timestamp, resource.createVKey(), resource);
+      return new ResultNow<>(resource);
+    }
+    return new ResultNow<>(resourceAtPointInTime);
+  }
+
   @Nullable
-  private static <T extends EppResource> Key<CommitLogManifest>
-      findMostRecentRevisionAtTime(final T resource, final DateTime timestamp) {
+  private static <T extends EppResource>
+      Key<CommitLogManifest> findMostRecentDatastoreRevisionAtTime(
+          final T resource, final DateTime timestamp) {
     final Key<T> resourceKey = Key.create(resource);
     Entry<?, Key<CommitLogManifest>> revision = resource.getRevisions().floorEntry(timestamp);
     if (revision != null) {
--- a/core/src/main/java/google/registry/model/contact/ContactHistory.java
+++ b/core/src/main/java/google/registry/model/contact/ContactHistory.java
@@ -18,6 +18,7 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.annotation.EntitySubclass;
+import google.registry.model.EppResource;
 import google.registry.model.ImmutableObject;
 import google.registry.model.contact.ContactHistory.ContactHistoryId;
 import google.registry.model.reporting.HistoryEntry;
@@ -105,6 +106,11 @@ public class ContactHistory extends HistoryEntry implements SqlEntity {
   @SuppressWarnings("unchecked")
   public VKey<ContactHistory> createVKey() {
     return (VKey<ContactHistory>) createVKey(Key.create(this));
+  }
+
+  @Override
+  public Optional<? extends EppResource> getResourceAtPointInTime() {
+    return getContactBase();
   }
 
   @PostLoad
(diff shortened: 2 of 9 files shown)


jianglai (changes requested):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 293 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbkz8eHB4UhIBdke1yV:-Mbkz8eHB4UhIBdke1yW:b-m60ic1) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L293)):*
> ```
> loadAtPointInTime
> ```
I think it might be better to return the resource `T` directly here as `Result` is a concept that is only meaningful to mapreduce. The return value can be wrapped in the call site if needed.
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 366 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkwDdU2HvkJTnMabHp:-MbkwDdU2HvkJTnMabHq:bcz87mm) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L366)):*
> ```
>    * before or at the provided timestamp using *History objects, falling back to using the earliest
>    * revision or the resource as-is if there are no revisions.
> ```
I don't see this logic implemented? 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkyBZh7KStFO7DKnwu:-MbkyBZh7KStFO7DKnwv:b-45aajv) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L300)):*
> ```
>  return Optional.empty();
> ```
Throw an `UnsupportedOperationException` here (see the`asBuilder()` method below).
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 146 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbl-VQm14eHTQeH1rdj:-Mbl-VQm14eHTQeH1rdk:b-q01idh) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L146)):*
> ```
> assertAboutImmutableObjects()
> ```
If there a reason to use this if the plain old `assertThat()` works?
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 153 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbl1E-T0HTR9FAtkWGN:-Mbl1E-T0HTR9FAtkWGO:b64f7hy) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L153)):*
> ```
>       // Both updates happened on the same day. Since the revisions field has day granularity in
>       // Datastore, the key to the first update should have been overwritten by the second, and its
>       // timestamp rolled forward. So we have to fall back to the last revision before midnight.
> ```
I know this is not your change, but the comment here and the ones in `loadAtPointInTime` about datastore revision being on a 24-h granularity and the suggestion about setting the revision time to midnight still are a bit confusing for me. If a modification happens on 6/9/2021 10:00AM, it it suggesting to set the revision on 6/9/2021 12:00AM (the midnight before), or 6/10/2021 12:00AM (the midnight after)? It looks like it's setting it to the midnight after so that you get the revision from the previous day, but if that is the intended behavior, why does `loadAtPointInTime` say "   It's recommended that {@code timestamp} be set to midnight. **Otherwise** you must take into consideration that under certain circumstances, a resource might be restored to a revision on the previous day". Isn't it the expected behavior?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 5 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 293 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbkz8eHB4UhIBdke1yV:-MblMsTN8yo_d-IoNhEo:b-uv7pi2) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L293)):*

> ```
> loadAtPointInTime
> ```
I think it might be better to return the resource `T` directly here as `Result` is a concept that is only meaningful to mapreduce. The return value can be wrapped in the call site if needed.
</blockquote></details>

That seems like something we may wish to do as part of a separate PR; the goal here is to modify the EppResourceUtils call itself, not to modify the call sites. 

Given that this is still called in the mapreduces in the Datastore world, we would want to wait to change that until after the full migration to SQL.
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 366 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkwDdU2HvkJTnMabHp:-MblNzxc601GfGS3DoQ-:b-rb74gs) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L366)):*

> ```
>    * before or at the provided timestamp using *History objects, falling back to using the earliest
>    * revision or the resource as-is if there are no revisions.
> ```
I don't see this logic implemented? 
</blockquote></details>

You're right, done. 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkyBZh7KStFO7DKnwu:-MblMfXu1XwFzx-elgw2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L300)):*

> ```
>  return Optional.empty();
> ```
Throw an `UnsupportedOperationException` here (see the`asBuilder()` method below).
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 146 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbl-VQm14eHTQeH1rdj:-MblMMtL1TfoLt4LdhrC:bbri1qm) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L146)):*

> ```
> assertAboutImmutableObjects()
> ```
If there a reason to use this if the plain old `assertThat()` works?
</blockquote></details>

No, but it doesn't work. One object may be a subclass of the other so the strict equality check fails 
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 153 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbl1E-T0HTR9FAtkWGN:-MblO6aqBayZjAhrCwQb:bptlybz) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L153)):*

> ```
>       // Both updates happened on the same day. Since the revisions field has day granularity in
>       // Datastore, the key to the first update should have been overwritten by the second, and its
>       // timestamp rolled forward. So we have to fall back to the last revision before midnight.
> ```
I know this is not your change, but the comment here and the ones in `loadAtPointInTime` about datastore revision being on a 24-h granularity and the suggestion about setting the revision time to midnight still are a bit confusing for me. If a modification happens on 6/9/2021 10:00AM, it it suggesting to set the revision on 6/9/2021 12:00AM (the midnight before), or 6/10/2021 12:00AM (the midnight after)? It looks like it's setting it to the midnight after so that you get the revision from the previous day, but if that is the intended behavior, why does `loadAtPointInTime` say "   It's recommended that {@code timestamp} be set to midnight. **Otherwise** you must take into consideration that under certain circumstances, a resource might be restored to a revision on the previous day". Isn't it the expected behavior?
 
</blockquote></details>

Yeah, it's expected (or at least it was when the thing was written). You'd want to use midnight on the next day, in your example. Basically, if you have three revisions of a thing 

A: day 0, 1pm
B: day 1, 1pm
C: day 1, 3pm

what'll happen in Datastore is that the commit log for C overwrites the commit log for B, so if you request "day 1, 2pm" it won't even see B at all so it will fall back to A

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 293 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbkz8eHB4UhIBdke1yV:-MbmSWLAEuD_moDHbYwX:btsu7ju) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L293)):*

That seems like something we may wish to do as part of a separate PR; the goal here is to modify the EppResourceUtils call itself, not to modify the call sites. 

Given that this is still called in the mapreduces in the Datastore world, we would want to wait to change that until after the full migration to SQL.
</blockquote></details>

I'm fine with this being left to a separate PR, but I do think it should be done before the migration as the mapreduces themselves need to be migrated to equivalent logics in SQL (like the ones I'm doing). So If we don't unwrap here, we will need to unwrap in the call sites when `tm().isOfy() == false`, only to get rid of the wrapping entirely after the migration.
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 366 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkwDdU2HvkJTnMabHp:-MbmTART1on9R7tjdSMF:b81x6w4) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L366)):*

You're right, done. 
</blockquote></details>

It still doesn't "fall back to the earliest revision", does it? Come to think of it, does the datastore method implement this logic either? I also don't think it is necessary to do this, does it?
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkyBZh7KStFO7DKnwu:-MbmU34i6XRCJFHcdght:b-t3gny4) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L300)):*

Done.
</blockquote></details>

Can you also call out the fact that this method is implemented in subclasses in the javadoc, like `asBuilder()` does?
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 153 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbl1E-T0HTR9FAtkWGN:-MbmUSw82oJfIHgxUX2w:bx3zcc4) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L153)):*

Yeah, it's expected (or at least it was when the thing was written). You'd want to use midnight on the next day, in your example. Basically, if you have three revisions of a thing 

A: day 0, 1pm
B: day 1, 1pm
C: day 1, 3pm

what'll happen in Datastore is that the commit log for C overwrites the commit log for B, so if you request "day 1, 2pm" it won't even see B at all so it will fall back to A
</blockquote></details>

Right, can you change the comment in `loadAtPointInTime` a bit so that it doesn't use "otherwise"?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 293 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbkz8eHB4UhIBdke1yV:-MbmZqqu9DoQ4uJxcuLY:b-8tg63r) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L293)):*

I'm fine with this being left to a separate PR, but I do think it should be done before the migration as the mapreduces themselves need to be migrated to equivalent logics in SQL (like the ones I'm doing). So If we don't unwrap here, we will need to unwrap in the call sites when `tm().isOfy() == false`, only to get rid of the wrapping entirely after the migration.
</blockquote></details>

Yeah, eventually we'll want to get rid of the wrapping + unwrapping entirely. We should just wait to do that until the Datastore world is completely gone. Added a TODO
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 366 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkwDdU2HvkJTnMabHp:-Mbm_rgpB8vAkB7ZjfoE:b8zvzxv) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L366)):*

It still doesn't "fall back to the earliest revision", does it? Come to think of it, does the datastore method implement this logic either? I also don't think it is necessary to do this, does it?
</blockquote></details>

Yeah, changed the comment in all locations (it was just flat-out wrong before) 
___
*[core/src/main/java/google/registry/model/reporting/HistoryEntry.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1194#-MbkyBZh7KStFO7DKnwu:-MbmbkaV-EI2kX7m1R8Y:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/reporting/HistoryEntry.java#L300)):*

Can you also call out the fact that this method is implemented in subclasses in the javadoc, like `asBuilder()` does?
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 153 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbl1E-T0HTR9FAtkWGN:-Mbmbj0E1NeEBoeq-_g-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L153)):*

Right, can you change the comment in `loadAtPointInTime` a bit so that it doesn't use "otherwise"?
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 293 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbkz8eHB4UhIBdke1yV:-MbnbqxN0w8-YKgg5-kE:b-a3b1do) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L293)):*

Yeah, eventually we'll want to get rid of the wrapping + unwrapping entirely. We should just wait to do that until the Datastore world is completely gone. Added a TODO
</blockquote></details>

Well I think it should be done **before** we use this function in SQL, not after Datastore is gone. We can defer this discussion to a later time.
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 278 at r3](https://reviewable.io/reviews/google/nomulus/1194#-MbnbhtsAjIIz8VD4jAg:-MbnbhtsAjIIz8VD4jAh:bjxoef1) ([raw file](https://github.com/google/nomulus/blob/2a611bcf9a48fdd174b5c3ad049aef38f0ca3517/core/src/main/java/google/registry/model/EppResourceUtils.java#L278)):*
> ```
> If you don't use midnight
> ```
If you do use midnight, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 293 at r1](https://reviewable.io/reviews/google/nomulus/1194#-Mbkz8eHB4UhIBdke1yV:-MbqL_Bx2UJkyLBLENlD:b-ka4bsq) ([raw file](https://github.com/google/nomulus/blob/c6964c619d5adf703dbf5f270828e143dff7822d/core/src/main/java/google/registry/model/EppResourceUtils.java#L293)):*

Well I think it should be done **before** we use this function in SQL, not after Datastore is gone. We can defer this discussion to a later time.
</blockquote></details>

Sure, we can talk about it later. Not super important. 
___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 278 at r3](https://reviewable.io/reviews/google/nomulus/1194#-MbnbhtsAjIIz8VD4jAg:-MbqLRYj1b5Vnlbh3ctn:bi0wxsn) ([raw file](https://github.com/google/nomulus/blob/2a611bcf9a48fdd174b5c3ad049aef38f0ca3517/core/src/main/java/google/registry/model/EppResourceUtils.java#L278)):*

> ```
> If you don't use midnight
> ```
If you do use midnight, right?
</blockquote></details>

No, this is if you do not. If you use midnight, then by definition there are no previous times of the day that you "should" be getting that you aren't.

gbrodman (commented):
(only change in what I just pushed was fixing the Javadoc formatting) 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194#-:-MbqMHoJB0O5eQUQfN0q:b-aelowp)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @jianglai)

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 278 at r3](https://reviewable.io/reviews/google/nomulus/1194#-MbnbhtsAjIIz8VD4jAg:-MbqND9f2JQDzG5ohk_z:b-74yswi) ([raw file](https://github.com/google/nomulus/blob/2a611bcf9a48fdd174b5c3ad049aef38f0ca3517/core/src/main/java/google/registry/model/EppResourceUtils.java#L278)):*

No, this is if you do not. If you use midnight, then by definition there are no previous times of the day that you "should" be getting that you aren't. 
</blockquote></details>

Hmm, so the granularity is with **reading** but not **writing** the revisions? If I I write everything at midnight, of course earlier revisions in the day will be overwritten and there'd be no previous revisions that you should be getting, like you said. But if I write them as is, would I be storing two revisions at the actual timestamp but only getting the revision on the previous day when loading?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 278 at r3](https://reviewable.io/reviews/google/nomulus/1194#-MbnbhtsAjIIz8VD4jAg:-MbqQBxz3t2Yu1uIllmc:b-8kpkgk) ([raw file](https://github.com/google/nomulus/blob/2a611bcf9a48fdd174b5c3ad049aef38f0ca3517/core/src/main/java/google/registry/model/EppResourceUtils.java#L278)):*

Hmm, so the granularity is with **reading** but not **writing** the revisions? If I I write everything at midnight, of course earlier revisions in the day will be overwritten and there'd be no previous revisions that you should be getting, like you said. But if I write them as is, would I be storing two revisions at the actual timestamp but only getting the revision on the previous day when loading?
</blockquote></details>

`timestamp` is referring to reading, not writing (since this is the call that reads things). The point is that if you read at midnight you'll never get a "missed" revision from earlier in that day, since there is no "earlier" than midnight.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 278 at r3](https://reviewable.io/reviews/google/nomulus/1194#-MbnbhtsAjIIz8VD4jAg:-Mbqd0Rg8eZkm4Hl24Cj:by78m2q) ([raw file](https://github.com/google/nomulus/blob/2a611bcf9a48fdd174b5c3ad049aef38f0ca3517/core/src/main/java/google/registry/model/EppResourceUtils.java#L278)):*

`timestamp` is referring to reading, not writing (since this is the call that reads things). The point is that if you read at midnight you'll never get a "missed" revision from earlier in that day, since there is no "earlier" than midnight. 
</blockquote></details>

We don't have to obsess over this. This PR is ready to merge. I guess my confusion is that if you can specify a timestamp (for reading) down to the exact millisecond, why would the revision granularity be only to the days? There must be a process in either reading or writing that discards the time information and only pass the days downstream for processing, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1194)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/main/java/google/registry/model/EppResourceUtils.java, line 278 at r3](https://reviewable.io/reviews/google/nomulus/1194#-MbnbhtsAjIIz8VD4jAg:-MbqjIgo38ukv7xI-t8T:beiddus) ([raw file](https://github.com/google/nomulus/blob/2a611bcf9a48fdd174b5c3ad049aef38f0ca3517/core/src/main/java/google/registry/model/EppResourceUtils.java#L278)):*

We don't have to obsess over this. This PR is ready to merge. I guess my confusion is that if you can specify a timestamp (for reading) down to the exact millisecond, why would the revision granularity be only to the days? There must be a process in either reading or writing that discards the time information and only pass the days downstream for processing, right?
</blockquote></details>

Theoretically yeah, though I think this is often called in an ad-hoc manner? I don't know. Either way, this won't be relevant soon enough.

PR #1195, opened 2021-06-05
Make ExportDomainListsAction SQL-aware

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1195)

### DIFF
--- a/core/src/main/java/google/registry/export/ExportDomainListsAction.java
+++ b/core/src/main/java/google/registry/export/ExportDomainListsAction.java
@@ -135,15 +135,13 @@ public class ExportDomainListsAction implements Runnable {
             logger.atInfo().log(
                 "Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
             exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
-            logger.atInfo().log("domain lists for TLD %s written out to GCS", tld);
             exportToDrive(tld, domainsList, driveConnection);
-            logger.atInfo().log("domain lists for TLD %s written out to Drive", tld);
           });
     }
   }
 
-  protected static void exportToDrive(String tld, String domains, DriveConnection driveConnection) {
-    verifyNotNull(driveConnection, "expecting non-null driveConnection");
+  protected static boolean exportToDrive(String tld, String domains, DriveConnection driveConnection) {
+    verifyNotNull(driveConnection, "Expecting non-null driveConnection");
     try {
       Registry registry = Registry.get(tld);
       if (registry.getDriveFolderId() == null) {
@@ -162,11 +160,13 @@ public class ExportDomainListsAction implements Runnable {
       }
     } catch (Throwable e) {
       logger.atSevere().withCause(e).log(
-          "Error exporting registered domains for TLD %s to Drive", tld);
+          "Error exporting registered domains for TLD %s to Drive, skipping...", tld);
+      return false;
     }
+    return true;
   }
 
-  protected static void exportToGcs(
+  protected static boolean exportToGcs(
       String tld, String domains, String gcsBucket, int gcsBufferSize) {
     GcsFilename filename = new GcsFilename(gcsBucket, tld + ".txt");
     GcsUtils cloudStorage =
@@ -174,10 +174,12 @@ public class ExportDomainListsAction implements Runnable {
     try (OutputStream gcsOutput = cloudStorage.openOutputStream(filename);
         Writer osWriter = new OutputStreamWriter(gcsOutput, UTF_8)) {
       osWriter.write(domains);
-    } catch (IOException e) {
+    } catch (Throwable e) {
       logger.atSevere().withCause(e).log(
-          "Error exporting registered domains for TLD %s to GCS.", tld);
+          "Error exporting registered domains for TLD %s to GCS, skipping...", tld);
+      return false;
     }
+    return true;
   }
 
   static class ExportDomainListsMapper extends Mapper<DomainBase, String, String> {
@@ -237,10 +239,16 @@ public class ExportDomainListsAction implements Runnable {
       ImmutableList<String> domains = ImmutableList.sortedCopyOf(() -> fqdns);
       String domainsList = Joiner.on('\n').join(domains);
       logger.atInfo().log("Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
-      exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
-      getContext().incrementCounter("domain lists written out to GCS");
-      exportToDrive(tld, domainsList, driveConnection);
-      getContext().incrementCounter("domain lists written out to Drive");
+      if (exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize)) {
+        getContext().incrementCounter("domain lists successful written out to GCS");
+      } else {
+        getContext().incrementCounter("domain lists failed to write out to GCS");
+      }
+      if (exportToDrive(tld, domainsList, driveConnection)) {
+        getContext().incrementCounter("domain lists successfully written out to Drive");
+      } else {
+        getContext().incrementCounter("domain lists failed to write out to Drive");
+      }
     }
 
     @VisibleForTesting
--- a/core/src/main/java/google/registry/export/ExportDomainListsAction.java
+++ b/core/src/main/java/google/registry/export/ExportDomainListsAction.java
@@ -57,6 +57,7 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.util.Collection;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 import javax.inject.Inject;
 import org.joda.time.DateTime;
 
@@ -104,26 +105,28 @@ public class ExportDomainListsAction implements Runnable {
       ImmutableListMultimap.Builder<String, String> tldToDomainMapBuilder =
           new ImmutableListMultimap.Builder<>();
       tm().transact(
-              () ->
-                  jpaTm()
-                      .query()
-                      .getEntityManager()
-                      .createNativeQuery(
-                          "SELECT tld, domain_name FROM \"Domain\" "
-                              + "WHERE tld IN (:tlds) "
-                              + "AND creation_time <= :now "
-                              + "AND deletion_time > :now")
-                      .setParameter("tlds", realTlds)
-                      .setParameter("now", clock.nowUtc().toDate())
-                      .getResultStream()
-                      .forEach(
-                          row -> {
-                            @SuppressWarnings("unchecked")
-                            String tld = (String) ((Object[]) row)[0];
-                            @SuppressWarnings("unchecked")
-                            String domain = (String) ((Object[]) row)[1];
-                            tldToDomainMapBuilder.put(tld, domain);
-                          }));
+              () -> {
+                @SuppressWarnings("unchecked")
+                Stream<Object[]> stream =
+                    (Stream<Object[]>)
+                        jpaTm()
+                            .query(
+                                "SELECT tld, fullyQualifiedDomainName FROM Domain "
+                                    + "WHERE tld IN (:tlds) "
+                                    + "AND deletionTime > :now "
+                                    + "AND creationTime <= :now")
+                            .setParameter("tlds", realTlds)
+                            .setParameter("now", clock.nowUtc())
+                            .getResultStream();
+                stream.forEach(
+                    row -> {
+                      @SuppressWarnings("unchecked")
+                      String tld = (String) ((Object[]) row)[0];
+                      @SuppressWarnings("unchecked")
+                      String domain = (String) ((Object[]) row)[1];
+                      tldToDomainMapBuilder.put(tld, domain);
+                    });
+              });
       ImmutableListMultimap<String, String> tldToDomainMap =
           tldToDomainMapBuilder.orderValuesBy(Ordering.natural()).build();
       tldToDomainMap.keySet().stream()
--- a/core/src/main/java/google/registry/export/ExportDomainListsAction.java
+++ b/core/src/main/java/google/registry/export/ExportDomainListsAction.java
@@ -140,7 +140,8 @@ public class ExportDomainListsAction implements Runnable {
     }
   }
 
-  protected static boolean exportToDrive(String tld, String domains, DriveConnection driveConnection) {
+  protected static boolean exportToDrive(
+      String tld, String domains, DriveConnection driveConnection) {
     verifyNotNull(driveConnection, "Expecting non-null driveConnection");
     try {
       Registry registry = Registry.get(tld);
--- a/core/src/main/java/google/registry/export/ExportDomainListsAction.java
+++ b/core/src/main/java/google/registry/export/ExportDomainListsAction.java
@@ -33,9 +33,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Ordering;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.MediaType;
 import google.registry.config.RegistryConfig.Config;
@@ -55,9 +53,8 @@ import java.io.ObjectInputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
-import java.util.Collection;
+import java.util.List;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 import javax.inject.Inject;
 import org.joda.time.DateTime;
 
@@ -102,58 +99,45 @@ public class ExportDomainListsAction implements Runnable {
               ImmutableList.of(createEntityInput(DomainBase.class)))
           .sendLinkToMapreduceConsole(response);
     } else {
-      ImmutableListMultimap.Builder<String, String> tldToDomainMapBuilder =
-          new ImmutableListMultimap.Builder<>();
-      tm().transact(
-              () -> {
-                // Note that the order of deletionTime and creationTime in the query is significant.
-                // When Hibernate substitutes "now" it will first validate that the **first** field
-                // that is to be compared with it (deletionTime) is assignable from the substituted
-                // Java object (click.nowUtc()). Since creationTime is a CreateAutoTimestamp, if it
-                // comes first, we will need to substitute "now" with
-                // CreateAutoTimestamp.creat(clock.nowUtc()). This might look a bit strange as the
-                // Java object type is clearly incompatible between the two fields deletionTime
-                // (DateTime) and creationTime, yet they are compared with the same "now". It is
-                // actually OK because in the end Hibernate converts everything to SQL types (and
-                // Java field names to SQL column names) to run the query. Both CreateAutoTimestamp
-                // and DateTime are persisted as timestamp_z in SQL. It is only the validation that
-                // compares the Java types, and only with the first field that compares with the
-                // substituted value.
-                @SuppressWarnings("unchecked")
-                Stream<Object[]> stream =
-                    (Stream<Object[]>)
-                        jpaTm()
-                            .query(
-                                "SELECT tld, fullyQualifiedDomainName FROM Domain "
-                                    + "WHERE tld IN (:tlds) "
-                                    + "AND deletionTime > :now "
-                                    + "AND creationTime <= :now")
-                            .setParameter("tlds", realTlds)
-                            .setParameter("now", clock.nowUtc())
-                            .getResultStream();
-                stream.forEach(
-                    row -> {
-                      @SuppressWarnings("unchecked")
-                      String tld = (String) ((Object[]) row)[0];
-                      @SuppressWarnings("unchecked")
-                      String domain = (String) ((Object[]) row)[1];
-                      tldToDomainMapBuilder.put(tld, domain);
-                    });
-              });
-      ImmutableListMultimap<String, String> tldToDomainMap =
-          tldToDomainMapBuilder.orderValuesBy(Ordering.natural()).build();
-      tldToDomainMap.keySet().stream()
-          .forEach(
-              tld -> {
-                Collection<String> domains = tldToDomainMap.get(tld);
-                String domainsList = Joiner.on("\n").join(domains);
-                logger.atInfo().log(
-                    "Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
-                exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
-                logger.atInfo().log("domain lists for TLD %s written out to GCS", tld);
-                exportToDrive(tld, domainsList, driveConnection);
-                logger.atInfo().log("domain lists for TLD %s written out to Drive", tld);
-              });
+      realTlds.forEach(
+          tld -> {
+            List<String> domains =
+                tm().transact(
+                        () ->
+                            // Note that the order of deletionTime and creationTime in the query is
+                            // significant. When Hibernate substitutes "now" it will first validate
+                            // that the **first** field that is to be compared with it
+                            // (deletionTime) is assignable from the substituted Java object
+                            // (click.nowUtc()). Since creationTime is a CreateAutoTimestamp, if it
+                            // comes first, we will need to substitute "now" with
+                            // CreateAutoTimestamp.creat(clock.nowUtc()). This might look a bit
+                            // strange as the Java object type is clearly incompatible between the
+                            // two fields deletionTime (DateTime) and creationTime, yet they are
+                            // compared with the same "now". It is actually OK because in the end
+                            // Hibernate converts everything to SQL types (and Java field names to
+                            // SQL column names) to run the query. Both CreateAutoTimestamp and
+                            // DateTime are persisted as timestamp_z in SQL. It is only the
+                            // validation that compares the Java types, and only with the first
+                            // field that compares with the substituted value.
+                            jpaTm()
+                                .query(
+                                    "SELECT fullyQualifiedDomainName FROM Domain "
+                                        + "WHERE tld = :tld "
+                                        + "AND deletionTime > :now "
+                                        + "AND creationTime <= :now "
+                                        + "ORDER by fullyQualifiedDomainName ASC",
+                                    String.class)
+                                .setParameter("tld", tld)
+                                .setParameter("now", clock.nowUtc())
+                                .getResultList());
+            String domainsList = Joiner.on("\n").join(domains);
+            logger.atInfo().log(
+                "Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
+            exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
+            logger.atInfo().log("domain lists for TLD %s written out to GCS", tld);
+            exportToDrive(tld, domainsList, driveConnection);
+            logger.atInfo().log("domain lists for TLD %s written out to Drive", tld);
+          });
     }
   }
 
--- a/core/src/main/java/google/registry/export/ExportDomainListsAction.java
+++ b/core/src/main/java/google/registry/export/ExportDomainListsAction.java
@@ -112,7 +112,7 @@ public class ExportDomainListsAction implements Runnable {
                             // with it (deletionTime) is assignable from the substituted Java object
                             // (click.nowUtc()). Since creationTime is a CreateAutoTimestamp, if it
                             // comes first, we will need to substitute "now" with
-                            // CreateAutoTimestamp.creat(clock.nowUtc()). This might look a bit
+                            // CreateAutoTimestamp.create(clock.nowUtc()). This might look a bit
                             // strange as the Java object type is clearly incompatible between the
                             // two fields deletionTime (DateTime) and creationTime, yet they are
                             // compared with the same "now". It is actually OK because in the end
--- a/core/src/main/java/google/registry/export/ExportDomainListsAction.java
+++ b/core/src/main/java/google/registry/export/ExportDomainListsAction.java
@@ -19,9 +19,10 @@ import static com.google.common.base.Verify.verifyNotNull;
 import static google.registry.mapreduce.inputs.EppResourceInputs.createEntityInput;
 import static google.registry.model.EppResourceUtils.isActive;
 import static google.registry.model.registry.Registries.getTldsOfType;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.Action.Method.POST;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.joda.time.DateTimeZone.UTC;
 
 import com.google.appengine.tools.cloudstorage.GcsFilename;
 import com.google.appengine.tools.cloudstorage.RetryParams;
@@ -32,7 +33,9 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Ordering;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.MediaType;
 import google.registry.config.RegistryConfig.Config;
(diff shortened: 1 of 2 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 104 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbapnrGC2Yq03b3E7Y4:-MbapnrGC2Yq03b3E7Y5:b-jertp9) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L104)):*
> ```Java
>           .sendLinkToMapreduceConsole(response);
>     } else {
>       ImmutableListMultimap.Builder<String, String> tldToDomainsMapBuilder =
> ```

nit: should probably be "tldToDomainMapBuilder" (singular)
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbaqCCa23bAtCYV0aLm:-MbaqCCa23bAtCYV0aLn:b3uztb4) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L110)):*
> ```Java
>                   jpaTm()
>                       .getEntityManager()
>                       .createNativeQuery(
> ```

LoL I just got done replacing all of these :-).  Please use jpaTm().query() instead (it doesn't matter in this case, since there are no model objects that need to be detached in the result, but it will be a lot easier to verify that we don't have to worry about detaching if we just consistently use our wrapper functions).
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbarIz6Di8U_1G2D56b:-MbarIz6Di8U_1G2D56c:b-nlr5ha) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L126)):*
> ```Java
>                             tldToDomainsMapBuilder.put(tld, domain);
>                           }));
>       ImmutableListMultimap<String, String> tldToDomainsMap =
> ```

same nit here.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbaqCCa23bAtCYV0aLm:-Mbbyeju3e5eh5RwqGNS:b-gykoze) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L110)):*

Will it work? This is a native query and would never return an entity anyway. I don't think anything returned by a native query would be registered with the entity manager and therefore needs detaching. `jpaTm().query()` uses JPQL, not native SQL, right?
</blockquote></details>

Yes, it is JPQL.  Is there a reason you can't use JPQL for this?  I believe the only difference is that you can omit the annoying double-quotes from "Domain".

Yes, detaching is not necessary here.  Avoiding the underlying call just makes it easier to verify that we're definitely doing it correctly.  We might also want to remove getEntityManager() just as a form of verification.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbaqCCa23bAtCYV0aLm:-Mbd5J34537FzNLxSMWj:brn1cfj) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L110)):*

Yes, it is JPQL.  Is there a reason you can't use JPQL for this?  I believe the only difference is that you can omit the annoying double-quotes from "Domain".

Yes, detaching is not necessary here.  Avoiding the underlying call just makes it easier to verify that we're definitely doing it correctly.  We might also want to remove getEntityManager() just as a form of verification.
</blockquote></details>

I think removing `getEntityManager()` is probably the better way of doing things. That way we are sure that we are calling our implementation of `createNativeQuery`. Maybe we don't need to expose the the entity manager at all to the user?

The reason why I used native query is because `creationDate` is a wrapper around `DateTime` and I am having trouble comparing it to `now`. Even though it is saved as a `timestamp_z`, with JPQL we are not dealing with columns directly, but Java fields (so 'creationDate` rather than `creation_date`), and I haven't found a way to set `now` to a value that JPQL thinks is comparable with `creationDate`. Do you have any suggestions?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbaqCCa23bAtCYV0aLm:-MbdRkRU3aFcxSPFnrps:b-5i6vtq) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L110)):*

I think removing `getEntityManager()` is probably the better way of doing things. That way we are sure that we are calling our implementation of `createNativeQuery`. Maybe we don't need to expose the the entity manager at all to the user?

The reason why I used native query is because `creationDate` is a wrapper around `DateTime` and I am having trouble comparing it to `now`. Even though it is saved as a `timestamp_z`, with JPQL we are not dealing with columns directly, but Java fields (so 'creationDate` rather than `creation_date`), and I haven't found a way to set `now` to a value that JPQL thinks is comparable with `creationDate`. Do you have any suggestions?
</blockquote></details>

Hmm, actually I got JPQL to work. It looks like I have to convert the `DateTime` to a `CreateAutoTimestamp` when setting the parameters to pass validation. This kinda make sense because JPQL is trying to be smart and is comparing the Java type instead of the SQL type. But on the other hand it is misleading because you can see that the `:now` substitution is used to compare both `creationDate` and `deletionDate`, the latter of which is just a plain old `DateTime`. So the validation appears to be only getting the *first* bound type in the SQL statement when comparing the substituted type. In the end this is just validation, the Java types are converted to SQL types in constructing the actual query to run against the DB, so it works out OK as both `creationDate` and `deletionDate` are persisted as `timestamp_z`.

In fact after I swap the SQL statement to put `deletionDate` before `creationDate`, I don't have to convert the `DateTime` to `CreateAutoTimestamp` at all!

mindhog (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 110 at r1](https://reviewable.io/reviews/google/nomulus/1195#-MbaqCCa23bAtCYV0aLm:-MbfYs2m11IDWDO9k8vC:b-da8fo7) ([raw file](https://github.com/google/nomulus/blob/9ff8fd10013df165064ce3097d0348f6888e5328/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L110)):*

Hmm, actually I got JPQL to work. It looks like I have to convert the `DateTime` to a `CreateAutoTimestamp` when setting the parameters to pass validation. This kinda make sense because JPQL is trying to be smart and is comparing the Java type instead of the SQL type. But on the other hand it is misleading because you can see that the `:now` substitution is used to compare both `creationDate` and `deletionDate`, the latter of which is just a plain old `DateTime`. So the validation appears to be only getting the *first* bound type in the SQL statement when comparing the substituted type. In the end this is just validation, the Java types are converted to SQL types in constructing the actual query to run against the DB, so it works out OK as both `creationDate` and `deletionDate` are persisted as `timestamp_z`.

In fact after I swap the SQL statement to put `deletionDate` before `creationDate`, I don't have to convert the `DateTime` to `CreateAutoTimestamp` at all!
</blockquote></details>

That's seriously weird -- sorry to have lead you down that path!

Can you add a comment explaining that the field order is significant? That's the kind of thing that a maintainer needs to know :-)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

mindhog (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195#-:-Mbg2gXC3Ya5p_FkiCHK:bb6ywmx)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: all files reviewed, 8 unresolved discussions (waiting on @jianglai and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/1195#-Mbghtt45kxKxm_t1vQ-:-Mbghtt45kxKxm_t1vQ0:b-972hxe) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L105)):*
> ```Java
>           .sendLinkToMapreduceConsole(response);
>     } else {
>       ImmutableListMultimap.Builder<String, String> tldToDomainMapBuilder =
> ```

The `run()` function is too long now -- separate the Cloud SQL part out into a separate method.  Might want multiple helpers owing to length + indentation level. 
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 106 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgiuH-9FREJH8DAtN0:-MbgiuH-9FREJH8DAtN1:b-at0h8m) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L106)):*
> ```Java
>     } else {
>       ImmutableListMultimap.Builder<String, String> tldToDomainMapBuilder =
>           new ImmutableListMultimap.Builder<>();
> ```

Worth calling out that, with this approach, you are loading all domain names (and their TLDs) into memory simultaneously, 2x (you're also storing a sorted copy separately).  This should work at our current scale, but it's not completely futureproof.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 123 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgiKrV-0mFLL25L82N:-MbgiKrV-0mFLL25L82O:be32dk4) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L123)):*
> ```Java
>                 // substituted value.
>                 @SuppressWarnings("unchecked")
>                 Stream<Object[]> stream =
> ```

Why is this such a weak type?  You're getting strings, right?
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 134 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgiXgO9SXpV6w6VVE5:-MbgiXgO9SXpV6w6VVE6:bw4uyls) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L134)):*
> ```Java
>                             .setParameter("now", clock.nowUtc())
>                             .getResultStream();
>                 stream.forEach(
> ```

Is this maximally efficient?  Is batching being handled under the covers by Hibernate, and if so, what does that look like?  What is the batch size?
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 137 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgiT_E5ZlHvttb-iOG:-MbgiT_E5ZlHvttb-iOH:bqvocnj) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L137)):*
> ```Java
>                     row -> {
>                       @SuppressWarnings("unchecked")
>                       String tld = (String) ((Object[]) row)[0];
> ```

No way to get the correct String type out to begin with?
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 144 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgjC8o7YR3U6UMOtz0:-MbgjC8o7YR3U6UMOtz1:b-gesc39) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L144)):*
> ```Java
>               });
>       ImmutableListMultimap<String, String> tldToDomainMap =
>           tldToDomainMapBuilder.orderValuesBy(Ordering.natural()).build();
> ```

Would it be more efficient to maintain the list in sorted order to begin with than to sort it all at the end here like this?  It seems like this approach might use roughly 2X the memory.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 152 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgjQ2Y5-MLTmfjN4pF:-MbgjQ2Y5-MLTmfjN4pG:b-cohg7e) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L152)):*
> ```Java
>                 logger.atInfo().log(
>                     "Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
>                 exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
> ```

There seems to be some missing error handling here.  What errors do these `export...()` methods throw?  They're talking to external systems, so they can definitely fail.  As currently written, if the GCS export fails then the Drive export is not even attempted.  That seems wrong.  The two exports should have the chance of succeeding independently, no?

mindhog (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 115 at r5](https://reviewable.io/reviews/google/nomulus/1195#-MbksM7T6ZHJYIvk8-I_:-MbksM7U6EjCTNGTB78v:b-e2g5cu) ([raw file](https://github.com/google/nomulus/blob/14cf4bc5e67042aad171679af55d3d70044ee49d/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L115)):*
> ```Java
>                             // (click.nowUtc()). Since creationTime is a CreateAutoTimestamp, if it
>                             // comes first, we will need to substitute "now" with
>                             // CreateAutoTimestamp.creat(clock.nowUtc()). This might look a bit
> ```

`.create(` -- (missing an "e")

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 106 at r4](https://reviewable.io/reviews/google/nomulus/1195#-MbgiuH-9FREJH8DAtN0:-MblO0iwAUtR_IjjYJAr:bwpi9u2) ([raw file](https://github.com/google/nomulus/blob/1bb398d0c632d7b5032e11abf2293f91e60829b0/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L106)):*

I've thought about this. The alternative is that since we already have the real TLD list we can query one TLD at a time (we don't have to get the real TLD separately if we only work in SQL -- we can just join the domain table with the tld table and filter on TLD type, but it is already available anyway).

This would reduce the memory requirement to hold all domains in one TLD instead of all TLDs. Sort of trading time for space as we have to run SQL multiple times. The reducer is also working at this level of memory requirement. It's a background job so I guess time is not critical so the trade off is probably worth it to guard against OOM down the road if we become too successful :P It also makes the processing downstream much easier as we don't have to keep a map anymore.
</blockquote></details>

I like this new approach.  It's simpler and less memory-intensive, and running separate queries by a single short indexed string column can't be that much more expensive on the DB than running a single query over all.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 129 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblMZxn0-fjICD-rP4b:-MblMZxn0-fjICD-rP4c:brtxgom) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L129)):*
> ```Java
>                                         + "WHERE tld = :tld "
>                                         + "AND deletionTime > :now "
>                                         + "ORDER by fullyQualifiedDomainName ASC",
> ```

Note that this doesn't actually do the right thing, e.g.:

```
// Raw string sorted order, as in your current query:
appz.dev
appmm.dev

// Sorted correctly by label:
appmm.dev
appz.dev
```

(Because period comes in the ASCII table than letters do.)

The simplest way I can think of to sort this correctly is to, in SQL, strip off the TLD part (should be very easy) and then sort just on the domain label.  Then, in Java land, you can append back on the TLD when writing to file.

Alternatively, we could just give up and not do this, as it seems like the Datastore implementation is already incorrectly sorting them, and frankly it's not thaaaat important that the contents of this file be sorted in a truly correct manner.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 10 unresolved discussions (waiting on @CydeWeys, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 138 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOY79-0qBN3p0DvHh:-MblOY79-0qBN3p0DvHi:b-g709iw) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L138)):*
> ```Java
>                 "Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
>             exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
>             logger.atInfo().log("domain lists for TLD %s written out to GCS", tld);
> ```

Missing capitalization and a period, but also, I think it reads better in a move active voice, e.g.:

"Successfully wrote domain lists for TLD %s to GCS."
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 138 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblPWwS9Yn6q2nZBK9e:-MblPWwS9Yn6q2nZBK9f:brjjse9) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L138)):*
> ```Java
>                 "Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
>             exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
>             logger.atInfo().log("domain lists for TLD %s written out to GCS", tld);
> ```

This isn't correct, because the helper function itself can fail, but it's not rethrowing on failure, so this logging statement will run even on failure.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 140 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOoTE6ls5j_Ybfwks:-MblOoTE6ls5j_Ybfwkt:b-21tyrs) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L140)):*
> ```Java
>             logger.atInfo().log("domain lists for TLD %s written out to GCS", tld);
>             exportToDrive(tld, domainsList, driveConnection);
>             logger.atInfo().log("domain lists for TLD %s written out to Drive", tld);
> ```

Ditto on the exact logged message and incorrectly logging success even when a failure has occurred.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 146 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOPLwD__2zmtxhr5_:-MblOPLwD__2zmtxhr5a:b-ctkv96) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L146)):*
> ```Java
> 
>   protected static void exportToDrive(String tld, String domains, DriveConnection driveConnection) {
>     verifyNotNull(driveConnection, "expecting non-null driveConnection");
> ```

Should be "Expecting"
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 161 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblPADECvMxrEc6bMZv:-MblPADECvMxrEc6bMZw:b6vkss7) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L161)):*
> ```Java
>                 domains.getBytes(UTF_8));
>             "Exporting registered domains succeeded for TLD %s, response was: %s", tld, resultMsg);
> ```

Isn't this a double log statement?  It seems to me it's duplicative of the logging statement in `run()` above.  Either log success inside the helper function or at the callsite, but not in both places.  Since you want to log the result message as well, logging within the helper function probably makes the most sense.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 179 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOva05GW_IoLCpYG9:-MblOva05GW_IoLCpYGA:b-p1tqpt) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L179)):*
> ```Java
>       logger.atSevere().withCause(e).log(
>           "Error exporting registered domains for TLD %s to GCS.", tld);
> ```

I know this is a pre-existing error message, but it can be improved upon while we're here making changes. It doesn't say what happens after the error.  The following is better:

"Error exporting registered domains for TLD %s to GCS; skipping."

Same for export to Drive above.

CydeWeys (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @jianglai, and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 241 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblPlW65gKFFsXfDpKJ:-MblPlW65gKFFsXfDpKK:bn9qufm) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L241)):*
> ```Java
>       logger.atInfo().log("Exporting %d domains for TLD %s to GCS and Drive.", domains.size(), tld);
>       exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
>       getContext().incrementCounter("domain lists written out to GCS");
> ```

These two counters are incorrectly incremented even on failures.  To do this correctly you'd want these two export functions to return success/failure booleans, and then increment a counter appropriately, e.g.:

```
boolean successGcs = exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
getContext().incrementCounter(String.format("domain lists %s to GCS", successGcs ? "successfully written" : "that failed to write to"));
```
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 243 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblQEaC1jTQyceGOjQi:-MblQEaC1jTQyceGOjQj:b90rs1z) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L243)):*
> ```Java
>       getContext().incrementCounter("domain lists written out to GCS");
>       exportToDrive(tld, domainsList, driveConnection);
>       getContext().incrementCounter("domain lists written out to Drive");
> ```

Same comments here.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 115 at r5](https://reviewable.io/reviews/google/nomulus/1195#-MbksM7T6ZHJYIvk8-I_:-MbmUoGf-4WH7DZiD7ZV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/14cf4bc5e67042aad171679af55d3d70044ee49d/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L115)):*

`.create(` -- (missing an "e")
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 129 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblMZxn0-fjICD-rP4b:-MbmXHPGDmqaW3t9OuvE:berj8db) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L129)):*

Note that this doesn't actually do the right thing, e.g.:

```
// Raw string sorted order, as in your current query:
appz.dev
appmm.dev

// Sorted correctly by label:
appmm.dev
appz.dev
```

(Because period comes in the ASCII table than letters do.)

The simplest way I can think of to sort this correctly is to, in SQL, strip off the TLD part (should be very easy) and then sort just on the domain label.  Then, in Java land, you can append back on the TLD when writing to file.

Alternatively, we could just give up and not do this, as it seems like the Datastore implementation is already incorrectly sorting them, and frankly it's not thaaaat important that the contents of this file be sorted in a truly correct manner.
</blockquote></details>

I think you mean period comes **before** letters? Your comment is missing a crucial word. Doesn't that work in our favor as when you compare period with a letter it always comes first? so "abc.dev" vs "abcd.dev" will compare "." with "d" and place the former before the latter, which is what we want, right?

Also in your example the order is decided when you compare the fourth character "z" vs. "m", right? It has nothing to do with the dot.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 138 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOY79-0qBN3p0DvHh:-MbmYunnE3sI4OUEuuV9:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L138)):*

Missing capitalization and a period, but also, I think it reads better in a move active voice, e.g.:

"Successfully wrote domain lists for TLD %s to GCS."
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 138 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblPWwS9Yn6q2nZBK9e:-MbmZIvH4NQdm3Qkyp2z:b6zrrkh) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L138)):*

This isn't correct, because the helper function itself can fail, but it's not rethrowing on failure, so this logging statement will run even on failure.
</blockquote></details>

Good catch.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 140 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOoTE6ls5j_Ybfwks:-MbmZKik5VkYb5SOSOfX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L140)):*

Ditto on the exact logged message and incorrectly logging success even when a failure has occurred.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 146 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOPLwD__2zmtxhr5_:-MbmZNzV8QJx2fX5ShK_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L146)):*

Should be "Expecting"
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 161 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblPADECvMxrEc6bMZv:-MbmZXTC8BT7J47Om1Zs:b43iprh) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L161)):*

Isn't this a double log statement?  It seems to me it's duplicative of the logging statement in `run()` above.  Either log success inside the helper function or at the callsite, but not in both places.  Since you want to log the result message as well, logging within the helper function probably makes the most sense.
</blockquote></details>

You are right, I was just blinding converting the `getContext()` calls to logs without realizing that they are redundant.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 179 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblOva05GW_IoLCpYG9:-MbnKwiNCoNc-wF9TUsn:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L179)):*

I know this is a pre-existing error message, but it can be improved upon while we're here making changes. It doesn't say what happens after the error.  The following is better:

"Error exporting registered domains for TLD %s to GCS; skipping."

Same for export to Drive above.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 241 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblPlW65gKFFsXfDpKJ:-MbnTMbr-bpDUFoGwuEz:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L241)):*

These two counters are incorrectly incremented even on failures.  To do this correctly you'd want these two export functions to return success/failure booleans, and then increment a counter appropriately, e.g.:

```
boolean successGcs = exportToGcs(tld, domainsList, gcsBucket, gcsBufferSize);
getContext().incrementCounter(String.format("domain lists %s to GCS", successGcs ? "successfully written" : "that failed to write to"));
```
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/ExportDomainListsAction.java, line 243 at r6](https://reviewable.io/reviews/google/nomulus/1195#-MblQEaC1jTQyceGOjQi:-MbnTMrL8dXklq8pFijG:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7b491d2fa572d3ce6412fbf3387cb3db6249f807/core/src/main/java/google/registry/export/ExportDomainListsAction.java#L243)):*

Same comments here.
</blockquote></details>

Done.

mindhog (approved):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: all files reviewed, 10 unresolved discussions (waiting on @CydeWeys)

jianglai (commented):
> Dismissed @CydeWeys from a discussion.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1195)* status: 1 of 2 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @mindhog)

PR #1199, opened 2021-06-07
Start the DS->SQL replay cron job in non-prod environments

gbrodman:
This should be a no-op since we haven't enabled it but this means that
when we set the schedule, we'll start replaying

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1199)


mindhog (approved):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1199)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1204, opened 2021-06-10
Fix a flaky test

jianglai:
In testSuccess_expandSingleEvent_notIdempotentforDifferentRecurring(),
two Recurring entities are created with the only difference being their IDs. If
we don't order the Recurrings by ID when loading them there is no guarantee
which one is expanded first. In this test the expected OneTime entities are
created with the assumption that the first loaded DomainHistory (parent of a
OneTime) corresponds to the expanding the Recurring with the smaller ID (2L).
Since the DomainHistory entities are loaded in order of IDs, and the IDs are
created monotonically in time in tests, we need to load the Recurrings in
order of their IDs to ensure that the first DomainHistory is the result of
expanding the Recurring with ID of 2L. This should impose minimum performance
penalty as we are ordering by the primary key.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1204)

### DIFF
--- a/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
+++ b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
@@ -127,10 +127,11 @@ public class ExpandRecurringBillingEventsAction implements Runnable {
                       jpaTm()
                           .query(
                               "FROM BillingRecurrence "
-                                  + "WHERE event_time <= :executeTime "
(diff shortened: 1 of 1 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1204)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1206, opened 2021-06-10
Change DomainLists on Registry objects to Strings

sarahcaseybot:
Reserved and premium lists were stored as keys on Registry objects in Datastore. For Cloud SQL, they are stored simply as a string representation of the list name. Since Premium and Reserved lists are no longer saved to Datastore, we can change the datatype of these lists in the Registry object.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1206)

### DIFF
--- a/core/src/main/java/google/registry/export/ExportPremiumTermsAction.java
+++ b/core/src/main/java/google/registry/export/ExportPremiumTermsAction.java
@@ -138,7 +138,7 @@ public class ExportPremiumTermsAction implements Runnable {
 
   private String getFormattedPremiumTerms(Registry registry) {
     checkState(registry.getPremiumList().isPresent(), "%s does not have a premium list", tld);
-    String premiumListName = registry.getPremiumList().get().getName();
+    String premiumListName = registry.getPremiumList().get();
     checkState(
         PremiumListDao.getLatestRevision(premiumListName).isPresent(),
         "Could not load premium list for " + tld);
--- a/core/src/main/java/google/registry/export/ExportUtils.java
+++ b/core/src/main/java/google/registry/export/ExportUtils.java
@@ -14,12 +14,15 @@
 
 package google.registry.export;
 
+import static com.google.common.base.Preconditions.checkState;
+
 import com.google.common.base.Joiner;
-import com.googlecode.objectify.Key;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.registry.Registry;
 import google.registry.model.registry.label.ReservedList;
 import google.registry.model.registry.label.ReservedList.ReservedListEntry;
+import google.registry.model.registry.label.ReservedListDao;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
 import javax.inject.Inject;
@@ -39,10 +42,11 @@ public final class ExportUtils {
   public String exportReservedTerms(Registry registry) {
     StringBuilder termsBuilder = new StringBuilder(reservedTermsExportDisclaimer).append("\n");
     Set<String> reservedTerms = new TreeSet<>();
-    for (Key<ReservedList> key : registry.getReservedLists()) {
-      ReservedList reservedList = ReservedList.load(key).get();
-      if (reservedList.getShouldPublish()) {
-        for (ReservedListEntry entry : reservedList.getReservedListEntries().values()) {
+    for (String reservedListName : registry.getReservedLists()) {
+      Optional<ReservedList> reservedList = ReservedListDao.getLatestRevision(reservedListName);
+      checkState(reservedList.isPresent(), "Reserved list %s does not exist", reservedListName);
+      if (reservedList.get().getShouldPublish()) {
+        for (ReservedListEntry entry : reservedList.get().getReservedListEntries().values()) {
           reservedTerms.add(entry.getLabel());
         }
       }
--- a/core/src/main/java/google/registry/model/pricing/StaticPremiumListPricingEngine.java
+++ b/core/src/main/java/google/registry/model/pricing/StaticPremiumListPricingEngine.java
@@ -41,7 +41,7 @@ public final class StaticPremiumListPricingEngine implements PremiumPricingEngin
     Optional<Money> premiumPrice =
         registry
             .getPremiumList()
-            .flatMap(listKey -> PremiumListDao.getPremiumPrice(listKey.getName(), label));
+            .flatMap(listKey -> PremiumListDao.getPremiumPrice(listKey, label));
     return DomainPrices.create(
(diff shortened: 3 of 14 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1206)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 396 at r1](https://reviewable.io/reviews/google/nomulus/1206#-MbvT-wR478-7om4fZZH:-MbvT-wR478-7om4fZZI:b-jjdz6j) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/main/java/google/registry/model/registry/Registry.java#L396)):*
> ```Java
>   /** Retrieves an ImmutableSet of all ReservedLists associated with this tld. */
>   public ImmutableSet<String> getReservedLists() {
>     return nullToEmptyImmutableCopy(reservedLists);
> ```

Can the reserved_list_names column ever be null?  I presume it's stored as an array, right?  Wouldn't that map to an empty set?
___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 610 at r1](https://reviewable.io/reviews/google/nomulus/1206#-MbvSkUJ7T_qDJUnGWD8:-MbvSkUJ7T_qDJUnGWD9:b-p1vjc9) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/main/java/google/registry/model/registry/Registry.java#L610)):*
> ```Java
> 
>   public Optional<String> getPremiumList() {
>     return Optional.ofNullable(premiumList);
> ```

I don't think we need this to be an optional any more, do we?  the column is not nullable.
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 272 at r1](https://reviewable.io/reviews/google/nomulus/1206#-MbvYCSc4njdIio2YkOD:-MbvYCSc4njdIio2YkOE:bv7n36p) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L272)):*
> ```Java
> 
>   @Override
>   public boolean refersToKey(Registry registry, Key<? extends BaseDomainLabelList<?, ?>> key) {
> ```

Is Key<BaseDomainLabelList still used anymore?  Does it make sense to just replace this with String, too?

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1206)* status: all files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/test/resources/google/registry/model/schema.txt, line 677 at r1](https://reviewable.io/reviews/google/nomulus/1206#-Mbvq_n6EmWnSh6EzrjI:-Mbvq_n6EmWnSh6EzrjJ:b-lmgpq3) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/test/resources/google/registry/model/schema.txt#L677)):*
> ```help
>   java.util.Set<java.lang.String> allowedRegistrantContactIds;
>   java.util.Set<java.lang.String> dnsWriters;
>   java.util.Set<java.lang.String> reservedLists;
> ```

Yeah, sorry, I totally should have noticed this on my last review -- as Ben said, this requires a data change for the registry objects.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1206)* status: all files reviewed, 7 unresolved discussions (waiting on @mindhog and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 392 at r1](https://reviewable.io/reviews/google/nomulus/1206#-Mbw4BLd7wZUArvcRcZR:-MbwVTVH666XctR7tI9B:b-j5mhsr) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/main/java/google/registry/model/registry/Registry.java#L392)):*

The Keys were already only used to get the names of the lists, so I don't see why it's necessary to include a reference to the whole lists in these objects when only their names are used.
</blockquote></details>

That's kind of orthogonal to the point I was making, which is that `Key<ReservedList> foo` tells me a _lot_ more what `foo` is actually used for than `String foo` does.  `String foo` could be just about anything.

Now if we were using orthodox Hibernate, it would make more sense to have it be `ReservedList foo`, with an appropriate Hibernate annotation that makes that possible (e.g. I think for this use case that `@ManyToOne` is correct?), along with lazy loading so that we don't always load the `ReservedList` along with every `Registry` entity.  However, we're not using orthodox Hibernate because we're retaining our concept of immutability, which requires detaching on load (which breaks all lazy loading), so maybe the next best alternative is to do `VKey<ReservedList>`?

The larger point still stands: it would be fairly unusable if most of our schema consisted of bare Strings and longs, which don't immediately make it clear what other table said String or long is actually referring to.  By using the actual class name whose id is being referred to here it makes everything more readable and browseable.

Also, make sure that the correct foreign key constraint is set in the Postgres schema.
___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 396 at r1](https://reviewable.io/reviews/google/nomulus/1206#-MbvT-wR478-7om4fZZH:-Mbw_4tW4dLUff8QjtBO:b-71gw6a) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/main/java/google/registry/model/registry/Registry.java#L396)):*

Can the reserved_list_names column ever be null?  I presume it's stored as an array, right?  Wouldn't that map to an empty set?
</blockquote></details>

It's gonna start off null on a newly created `Registry` object.
___
*[core/src/main/java/google/registry/model/registry/Registry.java, line 610 at r1](https://reviewable.io/reviews/google/nomulus/1206#-MbvSkUJ7T_qDJUnGWD8:-Mbw_F8J65D6ZKUwQPD6:btywzjs) ([raw file](https://github.com/google/nomulus/blob/6287855798f4ae9b6c6ff1d5ae7c88c28cfd07b1/core/src/main/java/google/registry/model/registry/Registry.java#L610)):*

I don't think we need this to be an optional any more, do we?  the column is not nullable.
</blockquote></details>

Same comment regarding it being null on a newly created `Registry`.

And the column definitely should be nullable, because not all TLDs have a premium list set.  .new doesn't, for example.

PR #1209, opened 2021-06-15
Make RegistryJpaIO use CriteriaQuery intead of QueryComposer

jianglai:
QueryComposer could be used when the transaction manager is not
determined (i. e. it supports both ofy and sql), but this also imposes
limits on what you can do with it. For example it does not support IN
operator in the where clause.

Since QueryComposer itself creates a CriteriaQuery for JPA TM it make
sense to have RegistryJpaIO take a CriteriaQuery directly as it only
uses JPA.

Also add some more helper methods to use native queries and typed
queires, and fix some generic type warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1209)

### DIFF
--- a/config/presubmits.py
+++ b/config/presubmits.py
@@ -202,6 +202,8 @@ PRESUBMITS = {
         "java",
         # ActivityReportingQueryBuilder deals with Dremel queries
         {"src/test", "ActivityReportingQueryBuilder.java",
+         # This class contains helper method to make queries in Beam.
+         "RegistryJpaIO.java",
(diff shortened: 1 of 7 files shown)


sarahcaseybot (approved):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1209)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1211, opened 2021-06-15
Add SQL functionality to DeleteLoadTestDataAction

gbrodman:
This isn't directly meant to be run in production so some of the rough
edges (doesn't delete domains, can't delete contacts that are referenced
by an existing domain) are fine. We can handle those in
DeleteProberTestAction when we do the more comprehensive deletions.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1211)

### DIFF
--- a/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java
@@ -15,12 +15,14 @@
 package google.registry.batch;
 
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.config.RegistryEnvironment.PRODUCTION;
 import static google.registry.mapreduce.MapreduceRunner.PARAM_DRY_RUN;
 import static google.registry.mapreduce.inputs.EppResourceInputs.createEntityInput;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.Action.Method.POST;
+import static google.registry.util.DateTimeUtils.END_OF_TIME;
 
 import com.google.appengine.tools.mapreduce.Mapper;
 import com.google.common.collect.ImmutableList;
@@ -28,16 +30,24 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.flogger.FluentLogger;
 import com.googlecode.objectify.Key;
 import google.registry.config.RegistryEnvironment;
+import google.registry.flows.poll.PollFlowUtils;
 import google.registry.mapreduce.MapreduceRunner;
 import google.registry.model.EppResource;
+import google.registry.model.EppResourceUtils;
 import google.registry.model.contact.ContactResource;
+import google.registry.model.domain.DomainBase;
 import google.registry.model.host.HostResource;
 import google.registry.model.index.EppResourceIndex;
 import google.registry.model.index.ForeignKeyIndex;
+import google.registry.model.poll.PollMessage;
+import google.registry.model.reporting.HistoryEntry;
+import google.registry.model.reporting.HistoryEntryDao;
+import google.registry.persistence.VKey;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
+import google.registry.util.Clock;
 import java.util.List;
 import javax.inject.Inject;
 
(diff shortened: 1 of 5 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1211)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java, line 157 at r1](https://reviewable.io/reviews/google/nomulus/1211#-Mcqg9DqArGnQjAzJhJv:-Mcqg9DqArGnQjAzJhJw:bhukb1q) ([raw file](https://github.com/google/nomulus/blob/504a75174724e682f002788ea90bf2695e3b9c0f/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java#L157)):*
> ```
> Integer.MAX_VALUE
> ```
This can be `null`, which does not set the query limit.
___
*[core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java, line 160 at r1](https://reviewable.io/reviews/google/nomulus/1211#-McqgKgn1P92PvE9oYrQ:-McqgKgn1P92PvE9oYrR:bmkkd8f) ([raw file](https://github.com/google/nomulus/blob/504a75174724e682f002788ea90bf2695e3b9c0f/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java#L160)):*
> ```
> ImmutableSet.of()
> ```
Why setting it to empty? Shouldn't we just delete the host in question?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1211)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java, line 157 at r1](https://reviewable.io/reviews/google/nomulus/1211#-Mcqg9DqArGnQjAzJhJv:-Mct3HB0BmKB0OSAb37I:b-896fix) ([raw file](https://github.com/google/nomulus/blob/504a75174724e682f002788ea90bf2695e3b9c0f/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java#L157)):*

> ```
> Integer.MAX_VALUE
> ```
This can be `null`, which does not set the query limit.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java, line 160 at r1](https://reviewable.io/reviews/google/nomulus/1211#-McqgKgn1P92PvE9oYrQ:-Mct3Koi8_xAhvXqECl8:b-kxwq9e) ([raw file](https://github.com/google/nomulus/blob/504a75174724e682f002788ea90bf2695e3b9c0f/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java#L160)):*

> ```
> ImmutableSet.of()
> ```
Why setting it to empty? Shouldn't we just delete the host in question?
</blockquote></details>

Yes but if we delete it without removing it from the domain first, it'll throw a foreign key exception

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1211)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java, line 160 at r1](https://reviewable.io/reviews/google/nomulus/1211#-McqgKgn1P92PvE9oYrQ:-MctEAJtDlqWp47C_AaX:bkvpkbo) ([raw file](https://github.com/google/nomulus/blob/504a75174724e682f002788ea90bf2695e3b9c0f/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java#L160)):*

Yes but if we delete it without removing it from the domain first, it'll throw a foreign key exception 
</blockquote></details>

I meant we should just delete the host from the list of nameservers, instead of setting the list to empty, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1211)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java, line 160 at r1](https://reviewable.io/reviews/google/nomulus/1211#-McqgKgn1P92PvE9oYrQ:-MctN02vEgizWbpcImq2:bwenlh9) ([raw file](https://github.com/google/nomulus/blob/504a75174724e682f002788ea90bf2695e3b9c0f/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java#L160)):*

I meant we should just delete the host from the list of nameservers, instead of setting the list to empty, right?
</blockquote></details>

Ah, done. Yeah it's more complex (and it shouldn't really matter) but still

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1211)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1212, opened 2021-06-18
Remove some unnecessary Ofy key creation

jianglai:
When an Ofy key is created, the code checks if the current thread is an
GAE thread. For this to work in test we can either set up the full GAE
proxy by registering the AppEngineExtension, or the lightweight
DatastoreEntityExtension. The latter only masquerades the thread as an
GAe thread and does not provide the full database emulator, etc.

It is however misleading to think that by registering the
DatastoreEntityExtention all Ofy key creation related issues are now
fixed. For example when testing a Beam pipeline with a DirectRunner,
even though the pipeline runs locally, it is still scheduled by an
executor a different tread than the main testing thread (which the
extension has masqueraded). Therefore any Beam transform that involves
key creation fails in tests.

In an ideal world we would make sure that no Ofy key is created when
only SQL is needed during tests. However this is not practical given the
sheer number of calls to Key.create(). We can, however, pick some low
hanging fruits that make the tests less dependent on Ofy related
extensions when it's not necessary.

In this PR the following changes are made toward this effort:

1. Several dual datastore entities have hardcoded parents of
   getCrossTldKey() which gets called when the object is initialized.
   This is problematic because when testing SQL (for example when
   loading an entity from SQL) we are creating Ofy keys which
   necessitates the use of the extension. Since these @parent fields are
   never actually used in the Java code and are purely for putting these
   entities in the same entity group in datastore, we have
   getCrossTldKey() returns null if the current thread is not detected as an
   GAE thread.

2. All Buildable objects have a build() method which automatically
   assigns a Long id for entities that have an @id Long field, if it's
   not set explicitly. It does so by first using ObjectifyService.auditedOfy()
   to locate the field (if any), and then calling ObjectifyService.allocateId()
   to be assigned an unique id. This presents a problems in testing: Whenever
   the ObjectifyService class is loaded, it calls its static initialization
   block and registers all datastore entities. During this process each of
   these entities are instantiated. When such a class has a Key field that is
   hardcoded (like the classes in (1), but not limited to the @parent field),
   they will fail to instantiate when the extension is not registered in tests.
   This means that if any build() call (even if it is building a SQL only
   entity) exists in a test, the extension must be present. To fix this
   we need to avoid loading the ObjectifyService class:

   2.1 We instead use the ModelUtils.getAllFields() method to find the
     field with @id annotation.

   2.2 We move the allocateId() method to its own class IdService. It's
     implementation will be replaced with a SQL equivalent when we
     finish the migration to SQL and no longer use Ofy to allocate the
     ids.

Also removes a now-unnecessary extension to unset the environmental
variable for App Engine runtime in RegistryReadTest, which is already
handled by the TestPipelineExtension.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1212)


mindhog (changes requested):
> Reviewed 18 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/model/Buildable.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoFF1a3J8hVH7p-sHa:-McoFF1b9RVYVC_uKYIc:btt0xvq) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/Buildable.java#L63)):*
> ```Java
>                 .filter(field -> field.isAnnotationPresent(Id.class))
>                 .findFirst()
>                 .orElse(null);
> ```

Is it possible that an entity can have multiple objectify ids?  I don't think so, but if it is not we might want to mention this in the comment and if so we need to deal with it.
___
*[core/src/main/java/google/registry/model/common/EntityGroupRoot.java, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoFw7Q0kDNPs5BTymp:-McoFw7Q0kDNPs5BTymq:bw2zlzz) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/common/EntityGroupRoot.java#L46)):*
> ```Java
> 
>   /** The root key for cross-tld resources such as registrars. */
>   public static Key<EntityGroupRoot> getCrossTldKey() {
> ```

Tack a `@Nullable` annotation onto the return type?
___
*[core/src/main/java/google/registry/model/ofy/ObjectifyService.java, line 107 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoGz7L4JCCX7Q8pIR-:-McoGz7L4JCCX7Q8pIR0:b397qjt) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/ofy/ObjectifyService.java#L107)):*
> ```Java
>             // In the unit test environment, wrap the Datastore service in a proxy that can be used
>             // to
>             // examine the number of requests sent to Datastore.
> ```

Reflow the comment?
___
*[core/src/main/java/google/registry/model/ofy/ObjectifyService.java, line 113 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoHAEvA2AW_KLLMSXe:-McoHAEvA2AW_KLLMSXf:b-fdqotw) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/ofy/ObjectifyService.java#L113)):*
> ```Java
>                 : service;
>           }
>         });
> ```

This change appears to be format-only, please try to avoid.
___
*[core/src/main/java/google/registry/model/tmch/ClaimsList.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoM-C99OYGEwSnxVoS:-McoIT1IAsyaYWUyOdB2:b-bz79op) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/tmch/ClaimsList.java#L49)):*
> ```Java
> import javax.persistence.GenerationType;
> import javax.persistence.PostPersist;
> import javax.persistence.PostUpdate;
> ```

I assume these new imports are for the `@code` references in the docs.  We should consider doing a cleanup on these.
___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 117 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoNs4KBp2PbFHs193m:-McoNs4KBp2PbFHs193n:b-1iin5) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L117)):*
> ```Java
>   @RegisterExtension
>   @Order(Order.DEFAULT - 1)
>   final transient DatastoreEntityExtension datastore = new DatastoreEntityExtension();
> ```

Do we still need the extension?
___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 32 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoOlfB7NDgmDi3O2MQ:-McoOlfB7NDgmDi3O2MR:b-siesz5) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L32)):*
> ```Java
>  *
>  * <p>When an Ofy key is created, by calling the various Key.create() methods, whether the current
>  * executing thread is an GAE thread is checked, which this extension masquerades. This happens
> ```

s/an/a/

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @id, and @mindhog)

___
*[core/src/main/java/google/registry/model/Buildable.java, line 63 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoFF1a3J8hVH7p-sHa:-McpL66h0aBI4xAqpDAu:b-e50zm3) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/Buildable.java#L63)):*

Is it possible that an entity can have multiple objectify ids?  I don't think so, but if it is not we might want to mention this in the comment and if so we need to deal with it.
</blockquote></details>

I don't think it is possible. Per [Objectify doc](https://github.com/objectify/objectify/wiki/Entities): "Entities must have one field annotated with @Id."

I quickly tested whether you can have two @Id annotated fields, one in a base class and one in a child class. It is not possible to have the child class field shadow the parent class -- this results in a failure when Ofy loads.
___
*[core/src/main/java/google/registry/model/common/EntityGroupRoot.java, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoFw7Q0kDNPs5BTymp:-McpOzmlA3UWGyfs1Y8D:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/common/EntityGroupRoot.java#L46)):*

Tack a `@Nullable` annotation onto the return type?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/ofy/ObjectifyService.java, line 107 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoGz7L4JCCX7Q8pIR-:-McpP3GqBIrlk1a7Uggp:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/ofy/ObjectifyService.java#L107)):*

Reflow the comment?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/ofy/ObjectifyService.java, line 113 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoHAEvA2AW_KLLMSXe:-McpPBlM05Gyex9lIBxT:btq55gm) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/ofy/ObjectifyService.java#L113)):*

This change appears to be format-only, please try to avoid.
</blockquote></details>

Acked. But I think in this instance the reformatting does improves readability.
___
*[core/src/main/java/google/registry/model/tmch/ClaimsList.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoM-C99OYGEwSnxVoS:-McpPJfF6cT9bb4s7Fep:b-kqi8sr) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/main/java/google/registry/model/tmch/ClaimsList.java#L49)):*

I assume these new imports are for the `@code` references in the docs.  We should consider doing a cleanup on these.
</blockquote></details>

Yes. This happens when the imports are automatically changed due to the removal of the `SystemPropertyExtension`. It's for `@link` though, not `@code`.
___
*[core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java, line 117 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoNs4KBp2PbFHs193m:-McpPZrT7Gfi6ujkL1IO:br7jazw) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L117)):*

Do we still need the extension?
</blockquote></details>

Unfortunately yes. As I stated in the PR description the use of `key.create()` is very prevalent in our codebase. In this case I think when you call `createTld()`, `Registry.setPremiumList` is called, which creates a key.
___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 32 at r1](https://reviewable.io/reviews/google/nomulus/1212#-McoOlfB7NDgmDi3O2MQ:-McpTG0QDXrJSDFlX39a:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d41e59766d52a59e95d44424501829a7704b74ed/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L32)):*

s/an/a/
</blockquote></details>

Done.

mindhog (commented):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212)* status: all files reviewed, 5 unresolved discussions (waiting on @jianglai and @mindhog)

___
*[core/src/main/java/google/registry/model/Buildable.java, line 63 at r2](https://reviewable.io/reviews/google/nomulus/1212#-McyVQ2m17mcWq_FiunX:-McyVQ2m17mcWq_FiunY:b93k4jo) ([raw file](https://github.com/google/nomulus/blob/740384bf476659098d69835e1a038956d52cd587/core/src/main/java/google/registry/model/Buildable.java#L63)):*
> ```Java
>             getAllFields(instance.getClass()).values().stream()
>                 .filter(field -> field.isAnnotationPresent(Id.class))
>                 .findFirst()
> ```

Should an error be thrown if more than 1 are returned?  That seems like it shouldn't happen.
___
*[core/src/main/java/google/registry/model/IdService.java, line 24 at r2](https://reviewable.io/reviews/google/nomulus/1212#-McyW0cS9MvAvsZd1s-v:-McyW0cS9MvAvsZd1s-w:b1ierpc) ([raw file](https://github.com/google/nomulus/blob/740384bf476659098d69835e1a038956d52cd587/core/src/main/java/google/registry/model/IdService.java#L24)):*
> ```Java
> import java.util.concurrent.atomic.AtomicLong;
> 
> public final class IdService {
> ```

Needs Javadoc.
___
*[core/src/main/java/google/registry/model/common/EntityGroupRoot.java, line 48 at r2](https://reviewable.io/reviews/google/nomulus/1212#-McyWSSj-_ImuPJKQW0e:-McyWSSj-_ImuPJKQW0f:b-8miluk) ([raw file](https://github.com/google/nomulus/blob/740384bf476659098d69835e1a038956d52cd587/core/src/main/java/google/registry/model/common/EntityGroupRoot.java#L48)):*
> ```Java
>   /** The root key for cross-tld resources such as registrars. */
>   public static @Nullable Key<EntityGroupRoot> getCrossTldKey() {
>     return ApiProxy.getCurrentEnvironment() == null
> ```

This needs an explanation here.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212)* status: 16 of 18 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @id, and @mindhog)

___
*[core/src/main/java/google/registry/model/Buildable.java, line 63 at r2](https://reviewable.io/reviews/google/nomulus/1212#-McyVQ2m17mcWq_FiunX:-Mcyg2Y7-r1mXbvTelDI:b-tz2wuj) ([raw file](https://github.com/google/nomulus/blob/740384bf476659098d69835e1a038956d52cd587/core/src/main/java/google/registry/model/Buildable.java#L63)):*

Should an error be thrown if more than 1 are returned?  That seems like it shouldn't happen.
</blockquote></details>

If there are more than one @Id fields, calling `auditedOfy()` will fail. That would probably happen before this method is called. 
___
*[core/src/main/java/google/registry/model/IdService.java, line 24 at r2](https://reviewable.io/reviews/google/nomulus/1212#-McyW0cS9MvAvsZd1s-v:-Mcygmoa8xJdQIP_lrRI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/740384bf476659098d69835e1a038956d52cd587/core/src/main/java/google/registry/model/IdService.java#L24)):*

Needs Javadoc.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/common/EntityGroupRoot.java, line 48 at r2](https://reviewable.io/reviews/google/nomulus/1212#-McyWSSj-_ImuPJKQW0e:-McyiOH-CrvnN5kCpQY6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/740384bf476659098d69835e1a038956d52cd587/core/src/main/java/google/registry/model/common/EntityGroupRoot.java#L48)):*

This needs an explanation here.
</blockquote></details>

Done.

mindhog (approved):
LGTM

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212#-:-Mcss50VBHSvT9BzVXQP:bb6ywmx)* status: 16 of 18 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @id, and @mindhog)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1212)* status: 16 of 18 files reviewed, 3 unresolved discussions (waiting on @jianglai and @mindhog)

___
*[core/src/main/java/google/registry/model/IdService.java, line 27 at r4](https://reviewable.io/reviews/google/nomulus/1212#-MczRPWY-2q8UaGtMHzE:-MczRPWY-2q8UaGtMHzF:bcxhu5u) ([raw file](https://github.com/google/nomulus/blob/af5723b6e31c8e8817a598cfb57372d71455beac/core/src/main/java/google/registry/model/IdService.java#L27)):*
> ```Java
>  * Allocates a globally unique {@link Long} number to use as an Ofy {@code @Id}.
>  *
>  * <p>In non-test environments the Id is generated by datastore, whereas in tests it's from an
> ```

Datastore

PR #1216, opened 2021-06-23
Support testing SQL -> DS replication in ReplayExt

mindhog:
Support testing of Postgres -> Datastore replication in the ReplayExtension
when running in SQL mode in a DualDatabaseTest.

This is currently only enabled for one test (HostInfoFlowTest) since this form
of replication is likely to be problematic in many cases.

As part of this change:

- Add a thread-local flag so that we don't attempt to do certain data
  transformations when serializing entities for storage in a Transaction
  record. (These typically need to be called in a datastore transaction).
- Replace tm() in datastore translators with ofyTm() (these should only be
  called from within an ofy transaction) and also in the replay system itself.
- Add a transactWithoutBackup() method for use within the replay itself.
- Prevent replication of entities that are not intended to be replicated.
- Make some of the ReplicateToDatastoreAction methods public so we can invoke
  them from ReplayExtension.
- Change the way that the test type is stored in the extension context in a
  DualDatabaseTest so that we can check for it from the ReplayExtension.

### DIFF
--- a/core/src/main/java/google/registry/model/translators/CommitLogRevisionsTranslatorFactory.java
+++ b/core/src/main/java/google/registry/model/translators/CommitLogRevisionsTranslatorFactory.java
@@ -17,13 +17,14 @@ package google.registry.model.translators;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static google.registry.config.RegistryConfig.getCommitLogDatastoreRetention;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
-import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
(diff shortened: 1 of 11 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1216)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

weiminyu (approved):
> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1216)* status: 10 of 12 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1219, opened 2021-06-25
Add a Beam pipeline to generate RDE deposit (part 1)

jianglai:
This is the first part of the RdeStagingAction SQL migration where the
mapper logic is implemented in Beam.

A few helper methods are added to convert the DomainContent, HostBase
and ContactBase to their respective terminal child classes. This is
necessary and possible because the child classes do not have extra
fields and the base classes exist only to be embedded to other entities
(such as the various HistoryEntry entities). The conversion is necessary
because most of our code expects the terminal classes, such as the
RdeMarshaller's various marshallXXX() methods. The alternative would be
to change all the call sites, which seems to be much more disruptive.

Unfortunately there is is no good way to do this conversion than just
creating a builder and setting every fields there is.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1219)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -777,6 +777,10 @@ if (environment in ['alpha', 'crash']) {
           mainClass: 'google.registry.beam.invoicing.InvoicingPipeline',
           metaData: 'google/registry/beam/invoicing_pipeline_metadata.json'
       ],
+      [
+          mainClass: 'google.registry.beam.rde.RdePipeline',
+          metaData: 'google/registry/beam/rde_pipeline_metadata.json'
+      ],
   ]
   project.tasks.create("stage_beam_pipelines") {
     doLast {
--- a/core/src/main/java/google/registry/beam/rde/RdePipeline.java
+++ b/core/src/main/java/google/registry/beam/rde/RdePipeline.java
@@ -132,6 +132,8 @@ public class RdePipeline implements Serializable {
                 "SELECT clientIdentifier FROM Registrar WHERE type NOT IN (:types)",
                 ImmutableMap.of("types", IGNORED_REGISTRAR_TYPES),
                 String.class,
+                // TODO: consider adding coders for entities and pass them directly instead of using
+                // VKeys.
                 id -> VKey.createSql(Registrar.class, id)))
         .apply(
             "Marshal Registrar into DepositFragment",
@@ -155,7 +157,12 @@ public class RdePipeline implements Serializable {
           Pipeline pipeline, Class<T> clazz) {
     return createInputs(pipeline, clazz)
         .apply("Marshal " + clazz.getSimpleName() + " into DepositFragment", mapToFragments(clazz))
-        .apply("Reshuffle to prevent fusion", Reshuffle.of());
+        .setCoder(KvCoder.of(PendingDepositCoder.of(), SerializableCoder.of(DepositFragment.class)))
+        .apply(
+            "Reshuffle KV<PendingDeposit, DepositFragment> of "
+                + clazz.getSimpleName()
+                + " to prevent fusion",
+            Reshuffle.of());
   }
 
   <T extends EppResource> PCollection<VKey<T>> createInputs(Pipeline pipeline, Class<T> clazz) {
@@ -167,6 +174,8 @@ public class RdePipeline implements Serializable {
                 ? ImmutableMap.of("tlds", pendings.keySet())
                 : ImmutableMap.of(),
             String.class,
+            // TODO: consider adding coders for entities and pass them directly instead of using
+            // VKeys.
             x -> VKey.create(clazz, x)));
   }
 
--- a/core/src/main/java/google/registry/beam/rde/RdePipeline.java
+++ b/core/src/main/java/google/registry/beam/rde/RdePipeline.java
@@ -1,4 +1,4 @@
-// Copyright 2018 The Nomulus Authors. All Rights Reserved.
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -58,6 +58,7 @@ import org.apache.beam.sdk.coders.SerializableCoder;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.apache.beam.sdk.transforms.FlatMapElements;
 import org.apache.beam.sdk.transforms.Flatten;
+import org.apache.beam.sdk.transforms.Reshuffle;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionList;
@@ -148,11 +149,13 @@ public class RdePipeline implements Serializable {
                     }));
   }
 
+  @SuppressWarnings("deprecation") // Reshuffle is still recommended by Dataflow.
   <T extends EppResource>
       PCollection<KV<PendingDeposit, DepositFragment>> processNonRegistrarEntities(
           Pipeline pipeline, Class<T> clazz) {
     return createInputs(pipeline, clazz)
-        .apply("Marshal " + clazz.getSimpleName() + " into DepositFragment", mapToFragments(clazz));
+        .apply("Marshal " + clazz.getSimpleName() + " into DepositFragment", mapToFragments(clazz))
+        .apply("Reshuffle to prevent fusion", Reshuffle.of());
   }
 
   <T extends EppResource> PCollection<VKey<T>> createInputs(Pipeline pipeline, Class<T> clazz) {
--- a/core/src/main/java/google/registry/beam/rde/RdePipeline.java
+++ b/core/src/main/java/google/registry/beam/rde/RdePipeline.java
@@ -1,0 +1,248 @@
+// Copyright 2018 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.beam.rde;
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static google.registry.model.EppResourceUtils.loadAtPointInTimeAsync;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSetMultimap;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import com.google.common.io.BaseEncoding;
+import google.registry.beam.common.RegistryJpaIO;
+import google.registry.model.EppResource;
+import google.registry.model.contact.ContactResource;
+import google.registry.model.domain.DomainBase;
+import google.registry.model.host.HostResource;
+import google.registry.model.rde.RdeMode;
+import google.registry.model.registrar.Registrar;
+import google.registry.model.registrar.Registrar.Type;
+import google.registry.persistence.VKey;
+import google.registry.rde.DepositFragment;
+import google.registry.rde.PendingDeposit;
+import google.registry.rde.PendingDeposit.PendingDepositCoder;
+import google.registry.rde.RdeFragmenter;
+import google.registry.rde.RdeMarshaller;
+import google.registry.xml.ValidationMode;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+import javax.persistence.Entity;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.PipelineResult;
+import org.apache.beam.sdk.coders.KvCoder;
+import org.apache.beam.sdk.coders.SerializableCoder;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.FlatMapElements;
+import org.apache.beam.sdk.transforms.Flatten;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.sdk.values.TypeDescriptors;
+import org.joda.time.DateTime;
+
+/**
+ * Definition of a Dataflow Flex template, which generates RDE/BRDA deposits.
+ *
+ * <p>To stage this template locally, run the {@code stage_beam_pipeline.sh} shell script.
+ *
+ * <p>Then, you can run the staged template via the API client library, gCloud or a raw REST call.
+ *
+ * @see <a href="https://cloud.google.com/dataflow/docs/guides/templates/using-flex-templates">Using
+ *     Flex Templates</a>
+ */
+public class RdePipeline implements Serializable {
+
+  private final RdeMarshaller marshaller;
+  private final ImmutableSetMultimap<String, PendingDeposit> pendings;
+
+  // Registrars to be excluded from data escrow. Not including the sandbox-only OTE type so that
+  // if sneaks into production we would get an extra signal.
+  private static final ImmutableSet<Type> IGNORED_REGISTRAR_TYPES =
+      Sets.immutableEnumSet(Registrar.Type.MONITORING, Registrar.Type.TEST);
+
+  private static final String EPP_RESOURCE_QUERY =
+      "SELECT id FROM %entity% "
+          + "WHERE COALESCE(creationClientId, '') NOT LIKE 'prober-%' "
+          + "AND COALESCE(currentSponsorClientId, '') NOT LIKE 'prober-%' "
+          + "AND COALESCE(lastEppUpdateClientId, '') NOT LIKE 'prober-%'";
+
+  public static String createEppResourceQuery(Class<? extends EppResource> clazz) {
+    return EPP_RESOURCE_QUERY.replace("%entity%", clazz.getAnnotation(Entity.class).name())
+        + (clazz.equals(DomainBase.class) ? " AND tld in (:tlds)" : "");
+  }
+
+  RdePipeline(RdePipelineOptions options) throws IOException, ClassNotFoundException {
+    this.marshaller = new RdeMarshaller(ValidationMode.valueOf(options.getValidationMode()));
+    this.pendings = decodePendings(options.getPendings());
+  }
+
+  @VisibleForTesting
+  PipelineResult run(Pipeline pipeline) {
+    createFragments(pipeline);
+    return pipeline.run();
+  }
+
+  PipelineResult run() {
+    return run(Pipeline.create());
+  }
+
+  PCollection<KV<PendingDeposit, DepositFragment>> createFragments(Pipeline pipeline) {
+    PCollection<KV<PendingDeposit, DepositFragment>> fragments =
+        PCollectionList.of(processRegistrars(pipeline))
+            .and(processNonRegistrarEntities(pipeline, DomainBase.class))
+            .and(processNonRegistrarEntities(pipeline, ContactResource.class))
+            .and(processNonRegistrarEntities(pipeline, HostResource.class))
+            .apply(Flatten.pCollections())
+            .setCoder(
+                KvCoder.of(PendingDepositCoder.of(), SerializableCoder.of(DepositFragment.class)));
+    return fragments;
+  }
+
+  PCollection<KV<PendingDeposit, DepositFragment>> processRegistrars(Pipeline pipeline) {
+    return pipeline
+        .apply(
+            "Read all production Registrar entities",
+            RegistryJpaIO.read(
+                "SELECT clientIdentifier FROM Registrar WHERE type NOT IN (:types)",
+                ImmutableMap.of("types", IGNORED_REGISTRAR_TYPES),
+                String.class,
+                id -> VKey.createSql(Registrar.class, id)))
+        .apply(
+            "Marshal Registrar into DepositFragment",
+            FlatMapElements.into(
+                    TypeDescriptors.kvs(
+                        TypeDescriptor.of(PendingDeposit.class),
+                        TypeDescriptor.of(DepositFragment.class)))
+                .via(
+                    (VKey<Registrar> key) -> {
+                      Registrar registrar = jpaTm().transact(() -> jpaTm().loadByKey(key));
+                      DepositFragment fragment = marshaller.marshalRegistrar(registrar);
+                      return pendings.values().stream()
+                          .map(pending -> KV.of(pending, fragment))
+                          .collect(toImmutableSet());
+                    }));
+  }
+
+  <T extends EppResource>
+      PCollection<KV<PendingDeposit, DepositFragment>> processNonRegistrarEntities(
+          Pipeline pipeline, Class<T> clazz) {
+    return createInputs(pipeline, clazz)
+        .apply("Marshal " + clazz.getSimpleName() + " into DepositFragment", mapToFragments(clazz));
+  }
+
+  <T extends EppResource> PCollection<VKey<T>> createInputs(Pipeline pipeline, Class<T> clazz) {
+    return pipeline.apply(
+        "Read all production " + clazz.getSimpleName() + " entities",
+        RegistryJpaIO.read(
+            createEppResourceQuery(clazz),
+            clazz.equals(DomainBase.class)
+                ? ImmutableMap.of("tlds", pendings.keySet())
+                : ImmutableMap.of(),
+            String.class,
+            x -> VKey.create(clazz, x)));
+  }
+
+  <T extends EppResource>
+      FlatMapElements<VKey<T>, KV<PendingDeposit, DepositFragment>> mapToFragments(Class<T> clazz) {
+    return FlatMapElements.into(
+            TypeDescriptors.kvs(
+                TypeDescriptor.of(PendingDeposit.class), TypeDescriptor.of(DepositFragment.class)))
+        .via(
+            (VKey<T> key) -> {
+              T resource = jpaTm().transact(() -> jpaTm().loadByKey(key));
+              // The set of all TLDs to which this resource should be emitted.
+              ImmutableSet<String> tlds =
+                  clazz.equals(DomainBase.class)
+                      ? ImmutableSet.of(((DomainBase) resource).getTld())
+                      : pendings.keySet();
+              // Get the set of all point-in-time watermarks we need, to minimize rewinding.
+              ImmutableSet<DateTime> dates =
+                  tlds.stream()
+                      .map(pendings::get)
+                      .flatMap(ImmutableSet::stream)
+                      .map(PendingDeposit::watermark)
+                      .collect(toImmutableSet());
+              // Launch asynchronous fetches of point-in-time representations of resource.
+              ImmutableMap<DateTime, Supplier<EppResource>> resourceAtTimes =
+                  ImmutableMap.copyOf(
+                      Maps.asMap(dates, input -> loadAtPointInTimeAsync(resource, input)));
+              // Convert resource to an XML fragment for each watermark/mode pair lazily and cache
+              // the result.
+              RdeFragmenter fragmenter = new RdeFragmenter(resourceAtTimes, marshaller);
+              List<KV<PendingDeposit, DepositFragment>> results = new ArrayList<>();
+              for (String tld : tlds) {
+                for (PendingDeposit pending : pendings.get(tld)) {
+                  // Hosts and contacts don't get included in BRDA deposits.
+                  if (pending.mode() == RdeMode.THIN && !clazz.equals(DomainBase.class)) {
+                    continue;
+                  }
+                  Optional<DepositFragment> fragment =
+                      fragmenter.marshal(pending.watermark(), pending.mode());
+                  fragment.ifPresent(
+                      depositFragment -> results.add(KV.of(pending, depositFragment)));
+                }
+              }
+              return results;
+            });
+  }
+
+  /**
+   * Decodes the pipeline option extracted from the URL parameter sent by the pipeline launcher to
+   * the original TLD to pending deposit map.
+   */
+  @SuppressWarnings("unchecked")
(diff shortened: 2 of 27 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 0 of 24 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-Md5A_b_AWOLi80yCKKN:bau7w5a) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*
> ```
> id -> VKey.createSql(Registrar.class, id)
> ```
Nice trick to sidestep the need for a coder for EppResource objects but passing the VKey around.

However, this does require an extra load-by-key query for each individual EppResource object. We may want investigate whether it has an impact on performance.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 0 of 24 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdHtaQb6rHNvOSlvVag:b-rswr2o) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

> ```
> id -> VKey.createSql(Registrar.class, id)
> ```
Nice trick to sidestep the need for a coder for EppResource objects but passing the VKey around.

However, this does require an extra load-by-key query for each individual EppResource object. We may want investigate whether it has an impact on performance. 
</blockquote></details>

I think this is just delaying the load from the read connector to the first transform. Unless there's a significant performance difference between loading all the resources in the initial query vs. loading a resources individually by its primary key?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 0 of 24 files reviewed, 9 unresolved discussions (waiting on @gbrodman, @jianglai, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdHzfAu6cObwG6XgwLa:b-lhklt6) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

I think this is just delaying the load from the read connector to the first transform. Unless there's a significant performance difference between loading all the resources in the initial query vs. loading a resources individually by its primary key?
</blockquote></details>

There really aren't that many real `Registrar` entities anyway, and this is only run once daily, so I can't see how it should affect performance, not unless the registrar is being loaded for each domain (that's not happening here, right?)

But there is another potential wrinkle -- with this approach we can get an inconsistent set of `Registrar` entities, e.g. if several of them happen to be updated right around the time we run RDE, we might randomly get some of the updates but not others.  Does that matter?
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 1 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdHylL_2yBiFhR3kDWa:-MdHylLaAnvSvsFthp47:b-vmzkyv) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L1)):*
> ```Java
> // Copyright 2018 The Nomulus Authors. All Rights Reserved.
> ```

Wrong year.
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 90 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdHyyvxDet1-CqU07ww:-MdHyyvxDet1-CqU07wx:b70vnd4) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L90)):*
> ```Java
>   private static final String EPP_RESOURCE_QUERY =
>       "SELECT id FROM %entity% "
>           + "WHERE COALESCE(creationClientId, '') NOT LIKE 'prober-%' "
> ```

Are all 3 of these clauses necessary?  We already have these?

And is `COALESCE` really necessary here?  Surely `NULL NOT LIKE '...'` doesn't return true?
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 95 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdHzE6i9grRP9Wd5KVH:-MdHzE6i9grRP9Wd5KVI:b-1knum7) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L95)):*
> ```Java
> 
>   public static String createEppResourceQuery(Class<? extends EppResource> clazz) {
>     return EPP_RESOURCE_QUERY.replace("%entity%", clazz.getAnnotation(Entity.class).name())
> ```

This isn't how templated SQL queries work.  Because you're not doing any escaping here, this could, in theory, maybe somehow, leave open a SQL injection attack.

Why not just template the name of the table properly?
___
*[core/src/main/java/google/registry/model/contact/ContactBase.java, line 294 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI01si6DTb8bUr68m9:-MdI01si6DTb8bUr68mA:bkun8he) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/model/contact/ContactBase.java#L294)):*
> ```Java
>   }
> 
>   public ContactResource asContactResource() {
> ```

A Javadoc explaining what this is, why it's necessary, what it's used for, etc., would be helpful.
___
*[core/src/main/java/google/registry/model/contact/ContactBase.java, line 294 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI0RJ62CYHr3fsAyI-:-MdI0RJ62CYHr3fsAyI0:b3gm26p) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/model/contact/ContactBase.java#L294)):*
> ```Java
>   }
> 
>   public ContactResource asContactResource() {
> ```

@gbrodman, you're confirming that there's no better way to do this?  Seeing as how `ContactResource` directly extends `ContactBase`, I'm surprised that this is the best way to do it?  Also might it make more sense for this to be in `ContactResource`, as in, something like`new ContactResource.Builder().setFieldsFromContactBase(contactBase).build()` or `ContactResource.fromContactBase(contactBase)` or `new ContactResource(contactBase)`?  It feels a little strange to see the base class referring to the extending class, rather than vice-versa.

And this approach seems fragile; you can imagine a new field being added to an EPP resource (like the auto-renew end time which I recently added), but then forgetting to add it here, and then it only fails in RDE in production.

Are there tests we can add to guard against this?
___
*[core/src/main/java/google/registry/rde/PendingDeposit.java, line 98 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1h1fCvn6FIe8ooFf:-MdI1h1gCHpR0iHsgZfx:b-olp2y1) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/rde/PendingDeposit.java#L98)):*
> ```Java
>   PendingDeposit() {}
> 
>   public static class PendingDepositCoder extends AtomicCoder<PendingDeposit> {
> ```

Missing Javadoc.
___
*[core/src/main/java/google/registry/rde/RdeFragmenter.java, line 36 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1qCg1_2mMnEBqPcO:-MdI1qCg1_2mMnEBqPcP:b63vk2z) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/rde/RdeFragmenter.java#L36)):*
> ```Java
> /** Loading cache that turns a resource into XML for the various points in time and modes. */
> public class RdeFragmenter {
>   private final Map<WatermarkModePair, Optional<DepositFragment>> cache = new HashMap<>();
> ```

Does thread synchronization matter here?  https://crunchify.com/hashmap-vs-concurrenthashmap-vs-synchronizedmap-how-a-hashmap-can-be-synchronized-in-java/

Also why aren't you using Guava's `LoadingCache`, like we do throughout our codebase?

(Is this just for memoization?)
___
*[core/src/main/java/google/registry/rde/RdeFragmenter.java, line 91 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI2gTOEoqK65t_a1I8:-MdI2gTOEoqK65t_a1I9:bfqdbq5) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/rde/RdeFragmenter.java#L91)):*
> ```Java
>       return result;
>     } else {
>       throw new AssertionError(resource.toString());
> ```

Not a useful error message (admittedly this is pre-existing, but we might as well fix it now).

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 0 of 24 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdHwhLE9B65cD55Lb9E:b-i4wffg) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

There really aren't that many real `Registrar` entities anyway, and this is only run once daily, so I can't see how it should affect performance, not unless the registrar is being loaded for each domain (that's not happening here, right?)

But there is another potential wrinkle -- with this approach we can get an inconsistent set of `Registrar` entities, e.g. if several of them happen to be updated right around the time we run RDE, we might randomly get some of the updates but not others.  Does that matter?
</blockquote></details>

I'm also referring to the same pattern used in createInputs().

Batch query is typically much faster than querying elements individually, especially since we are doing millions of calls here.
But depending on the rest of the application it may or may not worth optimize for it. We should probably leave a todo here as a reminder to look back.

gbrodman (changes requested):
> Reviewed 21 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 21 of 24 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[core/src/main/java/google/registry/model/contact/ContactBase.java, line 294 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI0RJ62CYHr3fsAyI-:-MdI3pVp3dmtEk23f8tU:b-oab7ni) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/model/contact/ContactBase.java#L294)):*

@gbrodman, you're confirming that there's no better way to do this?  Seeing as how `ContactResource` directly extends `ContactBase`, I'm surprised that this is the best way to do it?  Also might it make more sense for this to be in `ContactResource`, as in, something like`new ContactResource.Builder().setFieldsFromContactBase(contactBase).build()` or `ContactResource.fromContactBase(contactBase)` or `new ContactResource(contactBase)`?  It feels a little strange to see the base class referring to the extending class, rather than vice-versa.

And this approach seems fragile; you can imagine a new field being added to an EPP resource (like the auto-renew end time which I recently added), but then forgetting to add it here, and then it only fails in RDE in production.

Are there tests we can add to guard against this?
</blockquote></details>

I agree that this seems fragile, and we should put this in a builder (similar to how we already use copyFrom in HistoryEntry::builder 
___
*[core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md3jfkA8MZpEICNgWat:-Md3jfkA8MZpEICNgWau:b3ox9j9) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java#L68)):*
> ```Java
>    * Adds a WHERE clause to the query specifying that a value must not be in the given collection.
>    */
>   public CriteriaQueryBuilder<T> whereFieldIsNotIn(String fieldName, Collection<?> values) {
> ```

If we don't use this, we should remove it from the PR as unnecessary code.
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 162 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI2j-KDEL58XUotCTj:-MdI2j-L1h3GjdLx0mWx:b-rhud6c) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L162)):*
> ```Java
>       assertAboutImmutableObjects()
>           .that(loadAtPointInTime(latest, timeAtFirstUpdate))
>           .isEqualExceptFields(domainAfterFirstUpdate, "updateTimestamp");
> ```

Wait, why is the update timestamp getting updated? That shouldn't be happening AFAIK 
___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 54 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1B365X71BbP3wNH7:-MdI1B365X71BbP3wNH8:bfkt808) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L54)):*
> ```Java
>    * Whether all threads should be masqueraded as GAE threads.
>    *
>    * <p>This is particularly useful when new threads are spawn during a test. For example when
> ```

very nitty: spawn -> spawned (and same in the next paragraph) 
___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 59 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1RCQ8AaJ6wRe9LfQ:-MdI1RCQ8AaJ6wRe9LfR:bcmth4l) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L59)):*
> ```Java
>    *
>    * <p>Warning: by setting this value to true, any thread spawn by the current JVM will be
>    * masqueraded during the execution of the current test test, including those running other tests
> ```

1. "test" is duplicated 
2. I'm not sure "masqueraded" is the right term here or at least it might require a bit more information, are you in essence saying that every thread spawned by this thread will pretend to be this thread?

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 21 of 24 files reviewed, 13 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 155 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI472Z9iZWD5F43jMA:-MdI472Z9iZWD5F43jMB:bgo3iac) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L155)):*
> ```
> .apply("Marshal " + clazz.getSimpleName() + " into DepositFragment", mapToFragments(clazz));
> ```
mapToFragments could be a high-fanout function. Consider adding a Reshuffle stage after this:

https://cloud.google.com/dataflow/docs/guides/deploying-a-pipeline#preventing-fusion

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 21 of 24 files reviewed, 13 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdIBSkXEPsNgMZb_lk8:b-lll00o) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

I'm also referring to the same pattern used in createInputs().

Batch query is typically much faster than querying elements individually, especially since we are doing millions of calls here.
But depending on the rest of the application it may or may not worth optimize for it. We should probably leave a todo here as a reminder to look back.
</blockquote></details>

@cydeweys As for concurrent updates, the soft updates don't matter since the pipeline will try to find the correct snapshot at the reporting time.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 8 of 27 files reviewed, 13 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdKCZPHEsIECgXS4AUc:b-wbke1d) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

@cydeweys As for concurrent updates, the soft updates don't matter since the pipeline will try to find the correct snapshot at the reporting time. 
</blockquote></details>

@weiminyu loading snapshots only happens for epp resources, not registrars. Since we don't current load registrars at the time of the watermark anyway I don't think it is a big issue. Of course we are talking about internal inconsistency vs general stale data, so maybe we should ask Crystal for clarification?

Regarding using EntityProto as coder (as suggested in the email thread), that class is from the datastore SDK so we will have to fork that code (or live with a dependency on the SDK).
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 1 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdHylL_2yBiFhR3kDWa:-MdJNkEx36sYUJddGqO8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L1)):*

Wrong year.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 90 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdHyyvxDet1-CqU07ww:-MdJNvPME0lOFU3SghEF:b4qmll8) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L90)):*

Are all 3 of these clauses necessary?  We already have these?

And is `COALESCE` really necessary here?  Surely `NULL NOT LIKE '...'` doesn't return true?
</blockquote></details>

I am copying the logic from mapper, which does check for all three conditions. And `NULL NOT LIKE '...'` returns `NULL`, which is not `TRUE, so the entity got dropped... I'm surprised as well, see this:

https://stackoverflow.com/questions/22818070/behaviour-of-not-like-with-null-values

"Comparisons with NULL not using IS NULL or IS NOT NULL return NULL instead of true or false in most/all database systems."
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 95 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdHzE6i9grRP9Wd5KVH:-MdJQKMW7F1bAh35vUTN:b48xkcq) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L95)):*

This isn't how templated SQL queries work.  Because you're not doing any escaping here, this could, in theory, maybe somehow, leave open a SQL injection attack.

Why not just template the name of the table properly?
</blockquote></details>

Because we are replacing the FROM part. When calling `setParameter` (at least with JPQL), Hibernate actually does the field name -> column name translation for you (that's why you can use field name in your HPL/JPQL query) so it needs to know the entity beforehand. If you ask it to set the table itself it would not work. If we were using native query then yes, but I think in that case there's actually *more* injection attack risk because no sanitation is done at all on the injected parameters.

The injection attack risk here should remain theoretical because we also control the table name (set by the `@Entity` annotation).

See here: https://stackoverflow.com/questions/40956013/how-to-replace-table-name-with-value-from-parameter-while-using-spring-data-jpa
___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 155 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI472Z9iZWD5F43jMA:-MdKPRw15y8BwmqbM6DE:b-896fix) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L155)):*

> ```
> .apply("Marshal " + clazz.getSimpleName() + " into DepositFragment", mapToFragments(clazz));
> ```
mapToFragments could be a high-fanout function. Consider adding a Reshuffle stage after this:

https://cloud.google.com/dataflow/docs/guides/deploying-a-pipeline#preventing-fusion
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/rde/PendingDeposit.java, line 98 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1h1fCvn6FIe8ooFf:-MdKhuph0lweBl_xJCU7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/rde/PendingDeposit.java#L98)):*

Missing Javadoc.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/rde/RdeFragmenter.java, line 36 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1qCg1_2mMnEBqPcO:-MdKjMxW0mG92FCpM86_:b-5hdp4g) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/rde/RdeFragmenter.java#L36)):*

Does thread synchronization matter here?  https://crunchify.com/hashmap-vs-concurrenthashmap-vs-synchronizedmap-how-a-hashmap-can-be-synchronized-in-java/

Also why aren't you using Guava's `LoadingCache`, like we do throughout our codebase?

(Is this just for memoization?)
</blockquote></details>

I don't think synchronization matters here as we create a new `RdeFragmenter` for each resource, which is only accessed by a single thread.

This is also pre-existing code moved from the mapper to its own class. It does not look like `LoadingCache` is a good fit here because when a non-domain resource is marshalled to XML we proactively put them the in cache for more keys that needed. i. e. the key is (watermark, mode), but every time we marshall a resource we always put them as value for both (watermark, FULL) and (watermark, THIN), regardless of what `mode` actually is.
___
*[core/src/main/java/google/registry/rde/RdeFragmenter.java, line 91 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI2gTOEoqK65t_a1I8:-MdKjE5Q6hpJOa8gUYFU:b-896fix) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/rde/RdeFragmenter.java#L91)):*

Not a useful error message (admittedly this is pre-existing, but we might as well fix it now).
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/flows/EppPointInTimeTest.java, line 162 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI2j-KDEL58XUotCTj:-MdKPrw07oJ3tT3rIDbb:b4pont5) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/test/java/google/registry/flows/EppPointInTimeTest.java#L162)):*

Wait, why is the update timestamp getting updated? That shouldn't be happening AFAIK 
</blockquote></details>

It is not getting updated. There is no way to set `updateTimestamp` programmatically, it is automatically set when the entity is persisted into the database. So this field in the copy is null. Given that the copy is only used in RDE and `updateTimestamp` is an implementation detail that is not exposed to RDE, I think it is fine to leave it null. Unless you have reasons to believe it'll cause a problem?
___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 54 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1B365X71BbP3wNH7:-MdKQtus5JSfHZR-LWPH:b-896fix) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L54)):*

very nitty: spawn -> spawned (and same in the next paragraph) 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/DatastoreEntityExtension.java, line 59 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI1RCQ8AaJ6wRe9LfQ:-MdKQuWhCfDd7XcsJ8B_:b-eiv0rp) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/test/java/google/registry/testing/DatastoreEntityExtension.java#L59)):*

1. "test" is duplicated 
2. I'm not sure "masqueraded" is the right term here or at least it might require a bit more information, are you in essence saying that every thread spawned by this thread will pretend to be this thread? 
</blockquote></details>

What I meant was that every thread created by the current JVM will have the same thread local value set to the placeholder value, which might be a problem when you use the AppEngineExtension in other threads, which sets up a more complicated environment to mimic the behavior of GAE.

Updated the comment for clarity.
___
*[core/src/main/java/google/registry/model/contact/ContactBase.java, line 294 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI0RJ62CYHr3fsAyI-:-MdKef-K6vqUohxkUzBU:b-yk8g1) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/model/contact/ContactBase.java#L294)):*

I agree that this seems fragile, and we should put this in a builder (similar to how we already use copyFrom in HistoryEntry::builder 
</blockquote></details>

Moved the code to `ContactResource.Builder.copyFrom()`. I don't think there is a good way to downcast a parent class to a child class in general even though they have exactly the same fields. If you guys can come up with something, I'm all ears.

There are added tests (see `ContactResourceTest.testContactBasetoContactResource()`), but it all depends on if the test resource is constructed with all the fields.
___
*[core/src/main/java/google/registry/model/contact/ContactBase.java, line 294 at r2](https://reviewable.io/reviews/google/nomulus/1219#-MdI01si6DTb8bUr68m9:-MdKf_me6cWFPmNrYVsP:b-tjw9j6) ([raw file](https://github.com/google/nomulus/blob/27b0f73c04d11f6436955ee72c0c66bd26a65dc9/core/src/main/java/google/registry/model/contact/ContactBase.java#L294)):*

A Javadoc explaining what this is, why it's necessary, what it's used for, etc., would be helpful.
</blockquote></details>

It's a moot point now.
___
*[core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md3jfkA8MZpEICNgWat:-MdKfhhS3O8q-6I7Q6Xb:bx6c8qq) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/persistence/transaction/CriteriaQueryBuilder.java#L68)):*

If we don't use this, we should remove it from the PR as unnecessary code.
</blockquote></details>

Created #1225 for this. FWIW it was added because originally the plan was to load the entities from SQL using `CriteriaQueryBuilder` but it involves serializing the entities to pass between stages. So it was changed to use JPQL instead and this addition went unused.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 8 of 27 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdMz6PC9jWoU1zSx86z:bvbytw4) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

@weiminyu loading snapshots only happens for epp resources, not registrars. Since we don't current load registrars at the time of the watermark anyway I don't think it is a big issue. Of course we are talking about internal inconsistency vs general stale data, so maybe we should ask Crystal for clarification?

Regarding using EntityProto as coder (as suggested in the email thread), that class is from the datastore SDK so we will have to fork that code (or live with a dependency on the SDK).
</blockquote></details>

I mentioned EntityProto in the email specifically for the Datastore-sourced interim solution. I don't mean to use it for the long term either. But I expect that a serializer would be needed sometime in the future for other use cases anyway, and we can change it here if the performance impact is not trivial.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 8 of 27 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdePipeline.java, line 135 at r1](https://reviewable.io/reviews/google/nomulus/1219#-Md5A_bZ88JA98IlqIr1:-MdN4W1tCuC5eizW_Gcs:b52klib) ([raw file](https://github.com/google/nomulus/blob/bffec2f867ea8ad40dfa6a282ffd7ae187f9af49/core/src/main/java/google/registry/beam/rde/RdePipeline.java#L135)):*

I mentioned EntityProto in the email specifically for the Datastore-sourced interim solution. I don't mean to use it for the long term either. But I expect that a serializer would be needed sometime in the future for other use cases anyway, and we can change it here if the performance impact is not trivial.
</blockquote></details>

I'll leave a TODO.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 8 of 27 files reviewed, 11 unresolved discussions (waiting on @CydeWeys and @gbrodman)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219)* status: 8 of 27 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

gbrodman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1219#-:-MdSGycB51oXGjThNlMt:bnfp4nl)* status: 8 of 27 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1220, opened 2021-06-25
Migrate invoicing pipeline to read from Cloud SQL

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1220)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/1220#-Me2f1lg1WWSqTqBEQIv:-Me2f1lg1WWSqTqBEQIw:b-r930ph) ([raw file](https://github.com/google/nomulus/blob/da3acb31a4bb961d76ea7e88e4217b17c28234ce/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L120)):*
<details><summary><i alt="summary">Quoted 8 lines of code…</i></summary>

> ```
>                 "select b, r from BillingEvent b join Registrar r on b.clientId ="
>                     + " r.clientIdentifier join Domain d on b.domainRepoId = d.repoId join Tld t"
>                     + " on t.tldStrId = d.tld left join BillingCancellation c on b.id ="
>                     + " c.refOneTime.billingId left join BillingCancellation cr on"
>                     + " b.cancellationMatchingBillingEvent = cr.refRecurring.billingId where"
>                     + " r.billingIdentifier != null and r.type = 'REAL' and t.invoicingEnabled ="
>                     + " true and b.billingTime between CAST('%s' AS timestamp) and CAST('%s' AS"
>                     + " timestamp) and c.id = null and cr.id = null",
> ```
</details>

We should make this query text more readable. Consider putting it in a resource file.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/1220#-Me2f1lg1WWSqTqBEQIv:-MeBYyWGAYOtMePbfWA1:b-896fix) ([raw file](https://github.com/google/nomulus/blob/da3acb31a4bb961d76ea7e88e4217b17c28234ce/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L120)):*

> ```
>                 "select b, r from BillingEvent b join Registrar r on b.clientId ="
>                     + " r.clientIdentifier join Domain d on b.domainRepoId = d.repoId join Tld t"
>                     + " on t.tldStrId = d.tld left join BillingCancellation c on b.id ="
>                     + " c.refOneTime.billingId left join BillingCancellation cr on"
>                     + " b.cancellationMatchingBillingEvent = cr.refRecurring.billingId where"
>                     + " r.billingIdentifier != null and r.type = 'REAL' and t.invoicingEnabled ="
>                     + " true and b.billingTime between CAST('%s' AS timestamp) and CAST('%s' AS"
>                     + " timestamp) and c.id = null and cr.id = null",
> ```
We should make this query text more readable. Consider putting it in a resource file.
</blockquote></details>

Done.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 29 at r2](https://reviewable.io/reviews/google/nomulus/1220#-MeQSYfc4USMLKzxunx7:-MeQSYfd8PcdxsIpKlqH:b-a6cxk9) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L29)):*
<details><summary><i alt="summary">Quoted 12 lines of code…</i></summary>

> ```
>  select b, r
>  from BillingEvent b join Registrar r on b.clientId = r.clientIdentifier
>  join Domain d on b.domainRepoId = d.repoId
>  join Tld t on t.tldStrId = d.tld
>  left join BillingCancellation c on b.id = c.refOneTime.billingId
>  left join BillingCancellation cr on b.cancellationMatchingBillingEvent = cr.refRecurring.billingId
>  where r.billingIdentifier != null
>  and r.type = 'REAL'
>  and t.invoicingEnabled = true
>  and b.billingTime between CAST('%FIRST_TIMESTAMP_OF_MONTH%' AS timestamp) and CAST('%LAST_TIMESTAMP_OF_MONTH%' AS timestamp)
>  and c.id = null
>  and cr.id = null
> ```
</details>

Could you verify this with the BQ query script? For one thing, the registry.isInvoicingEnabled filter is not applied.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 29 at r2](https://reviewable.io/reviews/google/nomulus/1220#-MeQSYfc4USMLKzxunx7:-MeVJEliBdGlOzk-Lgwj:b-u663x0) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L29)):*

What do you mean by verify? I tried to add an example for each filter in the unit tests. is there a better way to verify that the query is correct? The invoicingEnabled filtering is done on line 26.
</blockquote></details>

Sorry I missed that line.

I meant that you may want to compare this with the original BQ query script and explain the reasons for the changes. 

For example, 

* the BQ query does not use the domain table. What's the reason to use it now? Is the targetId field in OneTime no longer usable?

* The BQ query applies the filters like Registry.invoicingEnabled and Registrar.isReal early to cut down the size of intermediate result sets, while this query puts a lot more rows in the intermediate results and only filters them at the end. This may have negative impact on performance.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 82 at r2](https://reviewable.io/reviews/google/nomulus/1220#-MeWpzq71VaArjIzpA_7:-MeWpzq71VaArjIzpA_8:b-huh8hh) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L82)):*
> ```
> void setupPipeline(Pipeline pipeline)
> ```
As a read-only (wrt sql) application, we can use less strict transaction isolation level when querying the database, E.g., READ_COMMITTED.

You can explicitly set it here like what InitSqlPipeline is doing (https://github.com/google/nomulus/blob/6ce0211537fff967018ad06cd4d542e1e9aac46a/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java#L139), or override the RegistryPipelineOptions.getIsolationOverride() method in your option file and declare the appropriate default there.

Separately, please also apply the same to the spec11 pipeline.

See b/193565763 for details.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_vcP13IO6UzMB-ds7:-Me_vcP13IO6UzMB-ds8:b-cruyh1) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L94)):*
> ```Java
>     this.jobRegion = jobRegion;
>     this.stagingBucketUrl = stagingBucketUrl;
>     if (tm().isOfy() && database.equals("CLOUD_SQL")) {
> ```

Needs a comment here explaining exactly what's going on.  Took me a bit to understand it.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 19 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_vxu-A7je-rIkoVdG:-Me_vxu-A7je-rIkoVdH:b-ktyltb) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L19)):*
> ```TSQL
> 
>  select b, r
>  from BillingEvent b join Registrar r on b.clientId = r.clientIdentifier
> ```

This file needs to be formatted correctly.  SQL keywords should be all caps (`SELECT`, etc.), all `JOIN`s should be on their own line (whereas this one, and only this one, is merged in with the previous line).
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 24 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_wOjG40-NQM6H6TKD:-Me_wOjG40-NQM6H6TKE:b-96c5bj) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L24)):*
> ```TSQL
>  left join BillingCancellation c on b.id = c.refOneTime.billingId
>  left join BillingCancellation cr on b.cancellationMatchingBillingEvent = cr.refRecurring.billingId
>  where r.billingIdentifier != null
> ```

`IS NOT NULL`
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 26 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_wGTC696cX-MJYaP0:-Me_wGTC696cX-MJYaP1:bu8k8i0) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L26)):*
> ```TSQL
>  where r.billingIdentifier != null
>  and r.type = 'REAL'
>  and t.invoicingEnabled = true
> ```

`IS TRUE`?
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 28 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_wEcZ3jDnvDc5cY6X:-Me_wEcZ3jDnvDc5cY6Y:b-6z7xc1) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L28)):*
> ```TSQL
>  and t.invoicingEnabled = true
>  and b.billingTime between CAST('%FIRST_TIMESTAMP_OF_MONTH%' AS timestamp) and CAST('%LAST_TIMESTAMP_OF_MONTH%' AS timestamp)
>  and c.id = null
> ```

Shouldn't these be `IS NULL`?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 82 at r2](https://reviewable.io/reviews/google/nomulus/1220#-MeWpzq71VaArjIzpA_7:-MebGZBx9LI7S_Ar9dKa:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L82)):*

> ```
> void setupPipeline(Pipeline pipeline)
> ```
As a read-only (wrt sql) application, we can use less strict transaction isolation level when querying the database, E.g., READ_COMMITTED.

You can explicitly set it here like what InitSqlPipeline is doing (https://github.com/google/nomulus/blob/6ce0211537fff967018ad06cd4d542e1e9aac46a/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java#L139), or override the RegistryPipelineOptions.getIsolationOverride() method in your option file and declare the appropriate default there.

Separately, please also apply the same to the spec11 pipeline.

See b/193565763 for details.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_vcP13IO6UzMB-ds7:-MebH9CU4OMSpYtDr2Hd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L94)):*

Needs a comment here explaining exactly what's going on.  Took me a bit to understand it.
</blockquote></details>

Done.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 19 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_vxu-A7je-rIkoVdG:-MebGaLR9kUzOg-QRQue:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L19)):*

This file needs to be formatted correctly.  SQL keywords should be all caps (`SELECT`, etc.), all `JOIN`s should be on their own line (whereas this one, and only this one, is merged in with the previous line).
</blockquote></details>

Done.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 24 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_wOjG40-NQM6H6TKD:-Meb1F914ibOjSvRqY_c:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L24)):*

`IS NOT NULL`
</blockquote></details>

Done.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 26 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_wGTC696cX-MJYaP0:-Meb1QSK0VbE9Avqiamr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L26)):*

`IS TRUE`?
</blockquote></details>

Done.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 28 at r2](https://reviewable.io/reviews/google/nomulus/1220#-Me_wEcZ3jDnvDc5cY6X:-Meb1Qh1BVqc94bHCsnd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L28)):*

Shouldn't these be `IS NULL`?
</blockquote></details>

Done.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 29 at r2](https://reviewable.io/reviews/google/nomulus/1220#-MeQSYfc4USMLKzxunx7:-Meedc8c5Mae-o0ALjuP:bxqqana) ([raw file](https://github.com/google/nomulus/blob/f5d3a29e6b1acfd71f8b1452d127246dabf58bb5/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L29)):*

JPQL also does not allow subqueries in the SELECT and FROM clause, so I don't think it's possible to apply the filters earlier in the query
</blockquote></details>

OK. Could you document this reason in the script? And also note that we should measure query performance and see if it is worth switching to native query.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 17 at r3](https://reviewable.io/reviews/google/nomulus/1220#-Mejw8Ya9ffbzDM_QH4I:-MejvfMX5X2MxoMWbjd5:b-ph91fo) ([raw file](https://github.com/google/nomulus/blob/08e02f207429cc13975a259c63e987b2612c89c0/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L17)):*
> ```TSQL
> -- This query gathers all non-canceled billing events for a given
> -- YEAR_MONTH in yyyy-MM format.
> ```

Could you document here the limitation of JPQL which prevents us from certain optimizations that could reduce the size of the intermediate results? Basically the explanation you made earlier that somehow disappeared from here (but still visible in github).

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 17 at r3](https://reviewable.io/reviews/google/nomulus/1220#-Mejw8Ya9ffbzDM_QH4I:-MekVWxA1JKiYRTgcK2p:b-896fix) ([raw file](https://github.com/google/nomulus/blob/08e02f207429cc13975a259c63e987b2612c89c0/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L17)):*

Could you document here the limitation of JPQL which prevents us from certain optimizations that could reduce the size of the intermediate results? Basically the explanation you made earlier that somehow disappeared from here (but still visible in github).
</blockquote></details>

Done.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 123 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUfDw-zx5t6NeLYEr:-MfIUfDz-0kHrTpTxfrN:bxm7u2q) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L123)):*
> ```Java
>         oneTime.getId(),
>         ZonedDateTime.ofInstant(
>             Instant.ofEpochMilli(oneTime.getBillingTime().getMillis()), ZoneId.of("UTC")),
> ```

Use `DateTimeUtils.toZonedDateTime()` for all such conversions.  And if that hadn't already existed, it would be worth creating it for this; this is exactly the kind of logic that should be in a helper for wide reuse.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUwx4EYlCJjwrGSpw:-MfIUwx4EYlCJjwrGSpx:bjlog07) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*
> ```Java
>             .build();
>     // Remove the comments from the query string
>     StringBuilder query = new StringBuilder();
> ```

Why are you bothering with this?  Comments are comments; PostgreSQL already knows what to do with them (i.e., nothing).  But properly stripping out comments requires a full parse of the language, which isn't being done here, so this is quite errorprone (e.g. imagine a query that contains `--` in it but not at the beginning of the line).
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 1 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIVcyM86vymX825N_f:-MfIVcyM86vymX825N_g:b170zbv) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L1)):*
> ```TSQL
> --Copyright 2021 The Nomulus Authors. All Rights Reserved.
> ```

Missing spaces between the `--` and the text on these license header lines.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 129 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIiUKcA3NJRbQFaNuf:-MfIiUKcA3NJRbQFaNug:b-rl2jxj) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L129)):*
> ```Java
>         registrar.getBillingIdentifier().toString(),
>         registrar.getPoNumber().orElse(""),
>         oneTime.getTargetId().substring(oneTime.getTargetId().lastIndexOf('.') + 1),
> ```

Does this work with multi-part TLDs?  Seems like it would fail?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 123 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUfDw-zx5t6NeLYEr:-MfJk1r8-NoDHsmHcp4Z:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L123)):*

Use `DateTimeUtils.toZonedDateTime()` for all such conversions.  And if that hadn't already existed, it would be worth creating it for this; this is exactly the kind of logic that should be in a helper for wide reuse.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 129 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIiUKcA3NJRbQFaNuf:-MfJphvCE_btDVs_BauV:b-krjkyy) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L129)):*

Does this work with multi-part TLDs?  Seems like it would fail?
</blockquote></details>

changed to indexOf()
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 1 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIVcyM86vymX825N_f:-MfK0BqWC4O4c5F0VwlE:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L1)):*

Missing spaces between the `--` and the text on these license header lines.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 129 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIiUKcA3NJRbQFaNuf:-MfdTvdB4o0DkC6L_Nde:b-yftrdg) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L129)):*

changed to indexOf()
</blockquote></details>

Use `DomainNameUtils.getTldFromDomainName()`.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 1 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIVcyM86vymX825N_f:-MfdV97E-nSE4N8JMryR:b5wp2vw) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L1)):*

Done.
</blockquote></details>

Now you've got incorrect leading spaces which shouldn't be there.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 28 at r5](https://reviewable.io/reviews/google/nomulus/1220#-MfdVCgu9AKkTZQjTv24:-MfdVCgu9AKkTZQjTv25:bs9fybq) ([raw file](https://github.com/google/nomulus/blob/0817afaaa54bd407922bd4663d1e5973176d4f0d/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L28)):*
> ```TSQL
> -- query's performance and consider switching to using a native query.
> 
>  SELECT b, r FROM BillingEvent b
> ```

Seems like there's incorrect leading spaces on all of these lines.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUwx4EYlCJjwrGSpw:-MfdXMhsAinK98SeQgRQ:bwo4jqx) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

Without filtering out the comments, I get this error message when I try to run the tests : ERROR org.hibernate.hql.internal.ast.ErrorTracker - line 1:1: unexpected token: -
antlr.NoViableAltException: unexpected token: -
</blockquote></details>

Can you investigate further why it's failing?  Comments shouldn't cause it to fail, and the way you're parsing out comments here is quite errorprone.  Maybe we're not passing the correct option somewhere along in the call chain?  Consult with @weiminyu too.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUwx4EYlCJjwrGSpw:-Mfda8Mq6_4Ct8oV8j9c:bgs9ud3) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

Can you investigate further why it's failing?  Comments shouldn't cause it to fail, and the way you're parsing out comments here is quite errorprone.  Maybe we're not passing the correct option somewhere along in the call chain?  Consult with @weiminyu too.
</blockquote></details>

Seems that Hibernate doesn't support comments in query text.

https://stackoverflow.com/questions/3196975/how-to-put-a-comment-into-hql-hibernate-query-language/3196988

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUwx4EYlCJjwrGSpw:-MfdnK-w7x6kpp44kr9k:boyca0) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

Seems that Hibernate doesn't support comments in query text.

https://stackoverflow.com/questions/3196975/how-to-put-a-comment-into-hql-hibernate-query-language/3196988
</blockquote></details>

Well that's pretty terrible.

We should centralize the comment stripping in `RegistryJpaIO` then (inside the innermost method that calls a query), and change our handling to remove lines that match a regex like `^\s*--`

Currently it's appending, what, everything in a line before any occurrence of `--`, which could easily do the wrong thing of the occurrence of `--` in the SQL query is inside a text string, e.g. imagine a query having something like `WHERE domainName LIKE 'xn--%'`

I believe that mine is safe, albeit less smart; fortunately I don't think we need it to be any smarter than that.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUwx4EYlCJjwrGSpw:-MfhkUSk4Vt27WOLVozp:b-mfqxsv) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

Well that's pretty terrible.

We should centralize the comment stripping in `RegistryJpaIO` then (inside the innermost method that calls a query), and change our handling to remove lines that match a regex like `^\s*--`

Currently it's appending, what, everything in a line before any occurrence of `--`, which could easily do the wrong thing of the occurrence of `--` in the SQL query is inside a text string, e.g. imagine a query having something like `WHERE domainName LIKE 'xn--%'`

I believe that mine is safe, albeit less smart; fortunately I don't think we need it to be any smarter than that.
</blockquote></details>

`^\s*--` is good enough.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 129 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIiUKcA3NJRbQFaNuf:-MfdrSyd5KjkhnLUizO5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L129)):*

Use `DomainNameUtils.getTldFromDomainName()`.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIUwx4EYlCJjwrGSpw:-Mfi_J1g7SHpndCFDIpG:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

`^\s*--` is good enough.
</blockquote></details>

Done.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 1 at r4](https://reviewable.io/reviews/google/nomulus/1220#-MfIVcyM86vymX825N_f:-Mfds0hp9ZeEX2ODd9EI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a12262c231cff46f8c55cf5247fbf3932d09c0ac/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L1)):*

Now you've got incorrect leading spaces which shouldn't be there.
</blockquote></details>

Done.
___
*[core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql, line 28 at r5](https://reviewable.io/reviews/google/nomulus/1220#-MfdVCgu9AKkTZQjTv24:-Mfds18S70ZjFbP7cAKP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0817afaaa54bd407922bd4663d1e5973176d4f0d/core/src/main/resources/google/registry/beam/invoicing/sql/cloud_sql_billing_events.sql#L28)):*

Seems like there's incorrect leading spaces on all of these lines.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r6](https://reviewable.io/reviews/google/nomulus/1220#-MfnnyGN3yfGE4kjEX4x:-MfnnyGN3yfGE4kjEX4y:b-72kgvv) ([raw file](https://github.com/google/nomulus/blob/0489d2aa73454e92454a69bf0659eb384984c630/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*
> ```Java
>     StringBuilder query = new StringBuilder();
>     for (String line : queryWithComments.split("\n")) {
>       query.append(line.split("^\\s*--", -1)[0]).append("\n");
> ```

I'm struggling to make sense of this; splitting on the regex when really you're trying to see if the line matches it is unintuitive.  Plus, it seems like you're adding blank lines still?

How about this?

```
for line in lines :
  if line doesn't match "^\\s*--" :
    append line to query
```

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r6](https://reviewable.io/reviews/google/nomulus/1220#-MfnnyGN3yfGE4kjEX4x:-MfnoVru8DD1g3m8hSwS:byltv7c) ([raw file](https://github.com/google/nomulus/blob/0489d2aa73454e92454a69bf0659eb384984c630/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

I'm struggling to make sense of this; splitting on the regex when really you're trying to see if the line matches it is unintuitive.  Plus, it seems like you're adding blank lines still?

How about this?

```
for line in lines :
  if line doesn't match "^\\s*--" :
    append line to query
```
</blockquote></details>

Also you should be using a compiled regex rather than passing in a raw string each time.  See `Pattern.compile()`

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r6](https://reviewable.io/reviews/google/nomulus/1220#-MfnnyGN3yfGE4kjEX4x:-Mfo82ns40xN-QAU07VX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0489d2aa73454e92454a69bf0659eb384984c630/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

Also you should be using a compiled regex rather than passing in a raw string each time.  See `Pattern.compile()`
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r6](https://reviewable.io/reviews/google/nomulus/1220#-MfnnyGN3yfGE4kjEX4x:-MfrrGDT4pMlYuXKXbo0:b-s0qwtl) ([raw file](https://github.com/google/nomulus/blob/0489d2aa73454e92454a69bf0659eb384984c630/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

Done.
</blockquote></details>

You're re-compiling the regex here every single time you're using it.  It needs to be compiled once, not for every line.

Add something like this as a static field to the class:

```
private static final Pattern SQL_COMMENT_REGEX = Pattern.compile("^\\s*--");
```

and then use `SQL_COMMENT_REGEX` here.
___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 238 at r7](https://reviewable.io/reviews/google/nomulus/1220#-Mfrrnd6BprrTlAKLZe3:-Mfrrnd6BprrTlAKLZe4:b7fruo9) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L238)):*
> ```Java
>     StringBuilder query = new StringBuilder();
>     for (String line : queryWithComments.split("\n")) {
>       if (!Pattern.compile("^\\s*--").matcher(line).find()) {
> ```

You want `.matches()`, not `.find()`.  For more info see:  https://stackoverflow.com/questions/4450045/difference-between-matches-and-find-in-java-regex

I think `find()` still works here by accident because of the `^` at the beginning of the regex, but it's best to be explicit about intent.
___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 239 at r7](https://reviewable.io/reviews/google/nomulus/1220#-MfrsBCa-zSf1iwA_P3U:-MfrsBCa-zSf1iwA_P3V:b-gcf06s) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L239)):*
> ```Java
>     for (String line : queryWithComments.split("\n")) {
>       if (!Pattern.compile("^\\s*--").matcher(line).find()) {
>         query.append(line.concat("\n"));
> ```

You're actually reducing some of the efficacy of your `StringBuilder` here by performing a String concatenation (the exact thing the `StringBuilder` is intended to avoid).  This is copying the entire line into a new memory location just to add a single newline (remember, Strings are immutable in Java).

A better way to do this is:

```
query.append(line).append("\n");
```

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 239 at r7](https://reviewable.io/reviews/google/nomulus/1220#-MfrsBCa-zSf1iwA_P3U:-MfsB_ES5uxvShVEk6Px:b-kiha10) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L239)):*

You're actually reducing some of the efficacy of your `StringBuilder` here by performing a String concatenation (the exact thing the `StringBuilder` is intended to avoid).  This is copying the entire line into a new memory location just to add a single newline (remember, Strings are immutable in Java).

A better way to do this is:

```
query.append(line).append("\n");
```
</blockquote></details>

You can also do a multi-line replaceAll:

Pattern pattern = Pattern.compile("^\\s*--.*\n", Pattern.MULTILINE);
queryWithComments = pattern.matcher(queryWithComments).replaceAll("");

Note that the pattern ends with \n instead of $. Using $ leaves a newline behind, while the given pattern removes the newline too.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 237 at r6](https://reviewable.io/reviews/google/nomulus/1220#-MfnnyGN3yfGE4kjEX4x:-MfsWUvg6s75wq83bKx8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0489d2aa73454e92454a69bf0659eb384984c630/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L237)):*

You're re-compiling the regex here every single time you're using it.  It needs to be compiled once, not for every line.

Add something like this as a static field to the class:

```
private static final Pattern SQL_COMMENT_REGEX = Pattern.compile("^\\s*--");
```

and then use `SQL_COMMENT_REGEX` here.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 239 at r7](https://reviewable.io/reviews/google/nomulus/1220#-MfrsBCa-zSf1iwA_P3U:-MfsWVXJDbnPhadLsz6B:b-896fix) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L239)):*

You can also do a multi-line replaceAll:

Pattern pattern = Pattern.compile("^\\s*--.*\n", Pattern.MULTILINE);
queryWithComments = pattern.matcher(queryWithComments).replaceAll("");

Note that the pattern ends with \n instead of $. Using $ leaves a newline behind, while the given pattern removes the newline too.
</blockquote></details>

Done.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 238 at r7](https://reviewable.io/reviews/google/nomulus/1220#-Mfrrnd6BprrTlAKLZe3:-MfsWYWMBBlfxiwCzNzJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L238)):*

You want `.matches()`, not `.find()`.  For more info see:  https://stackoverflow.com/questions/4450045/difference-between-matches-and-find-in-java-regex

I think `find()` still works here by accident because of the `^` at the beginning of the regex, but it's best to be explicit about intent.
</blockquote></details>

Done.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 239 at r7](https://reviewable.io/reviews/google/nomulus/1220#-MfrsBCa-zSf1iwA_P3U:-MgMJUSpEy_NVm8cnJkx:bgkr7tf) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L239)):*

Done.
</blockquote></details>

Good call, Weimin.

I'm not seeing tests of `makeCloudSqlQuery()` that verify it's removing comments correctly though?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java, line 239 at r7](https://reviewable.io/reviews/google/nomulus/1220#-MfrsBCa-zSf1iwA_P3U:-MgMiaX-ARRHvoj3RkQf:b-896fix) ([raw file](https://github.com/google/nomulus/blob/aaecd5b3c64505beef9c98c5bcff9f0e616318df/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L239)):*

Good call, Weimin.

I'm not seeing tests of `makeCloudSqlQuery()` that verify it's removing comments correctly though?
</blockquote></details>

Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1220)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/beam/invoicing/InvoicingPipelineTest.java, line 333 at r9](https://reviewable.io/reviews/google/nomulus/1220#-MgMnFzz2TfpukHSVrI5:-MgMnFzz2TfpukHSVrI6:b3tpsmw) ([raw file](https://github.com/google/nomulus/blob/e5a3fbc4b474ba5423d2521d8d193c251681ad52/core/src/test/java/google/registry/beam/invoicing/InvoicingPipelineTest.java#L333)):*
> ```Java
>     assertThat(InvoicingPipeline.makeCloudSqlQuery("2017-10"))
>         .isEqualTo(
>             "\n"
> ```

Add a comment here or something pointing out that it's explicitly testing that the comment lines in the file were stripped out.

PR #1222, opened 2021-06-25
Set HistoryEntry modification time in FlowModule

CydeWeys:
Rather than having to set it individually to now (the current transaction time)
in every transactional flow, just do it once at the beginning when the
HistoryEntry.Builder is first being provided. This is also safer, as just doing
it in one place gives us stronger guarantees that it always corresponds to the
execution time of the flow, rather than leaving the potential open that in one
flow it's unintentionally set to the wrong thing.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1222)

### DIFF
--- a/core/src/main/java/google/registry/flows/FlowModule.java
+++ b/core/src/main/java/google/registry/flows/FlowModule.java
@@ -15,6 +15,7 @@
 package google.registry.flows;
 
 import static com.google.common.base.Preconditions.checkState;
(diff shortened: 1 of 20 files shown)


mindhog (approved):
Apologies, I missed this.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1222#-:-MdMlZ6S9CcQ5H-0q5wd:be91j50)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1224, opened 2021-06-28
Set a 5min time limit on the SQL replay action

gbrodman:
This means we avoid GAE request timeouts and can get progress logs more
quickly (logs weren't showing up on GAE in Sandbox).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1224)

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -64,10 +64,13 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
 
   static final String PATH = "/_dr/task/replayCommitLogsToSql";
 
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final int BLOCK_SIZE =
       1024 * 1024; // Buffer 1mb at a time, for no particular reason.
-  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
   private static final Duration LEASE_LENGTH = standardHours(1);
+  // Stop / pause where we are if we've been replaying for more than five minutes to avoid GAE
+  // request timeouts
+  private static final Duration REPLAY_TIMEOUT_DURATION = Duration.standardMinutes(5);
 
   @Inject GcsService gcsService;
   @Inject Response response;
@@ -113,17 +116,26 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
   }
 
   private void replayFiles() {
+    DateTime replayTimeoutTime = clock.nowUtc().plus(REPLAY_TIMEOUT_DURATION);
     // Start at the first millisecond we haven't seen yet
     DateTime fromTime = jpaTm().transact(() -> SqlReplayCheckpoint.get().plusMillis(1));
     // If there's an inconsistent file set, this will throw IllegalStateException and the job
     // will try later -- this is likely because an export hasn't finished yet.
     ImmutableList<GcsFileMetadata> commitLogFiles =
         diffLister.listDiffFiles(fromTime, /* current time */ null);
+    int processedFiles = 0;
     for (GcsFileMetadata metadata : commitLogFiles) {
       // One transaction per GCS file
       jpaTm().transact(() -> processFile(metadata));
+      processedFiles++;
+      if (clock.nowUtc().isAfter(replayTimeoutTime)) {
+        logger.atInfo().log(
+            "Reached max execution time after replaying %d files, leaving %d files for next run.",
+            processedFiles, commitLogFiles.size() - processedFiles);
+        return;
+      }
     }
-    logger.atInfo().log("Replayed %d commit log files to SQL successfully.", commitLogFiles.size());
+    logger.atInfo().log("Replayed %d commit log files to SQL successfully.", processedFiles);
   }
 
   private void processFile(GcsFileMetadata metadata) {


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1224)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/1224#-MdJ6xHm0jC81QFxN_3K:-MdJ6xHn6-xFzc5bRTmj:b-ionceg) ([raw file](https://github.com/google/nomulus/blob/8c82f881e3c90a16c05154d91f38cb0c510b9e7e/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L129)):*
> ```Java
>     for (GcsFileMetadata metadata : commitLogFiles) {
>       // One transaction per GCS file
>       jpaTm().transact(() -> processFile(metadata));
> ```

I'm curious what the maximum amount of time it might take to process a single commit log file might be.  Seeing as how each commit log file contains all of the transactions from a several minute period (the duration between runs of the export commit log action), this could be a non-trivial amount of time in times of high QPS.  We can run with this for now as it is an improvement, but we might still have to consider how we might somehow do partial replays of files w/ resumes in the future if that becomes necessary, if a single one can still take way too long to run.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 133 at r1](https://reviewable.io/reviews/google/nomulus/1224#-MdJ5vZtAx-MUTXuNJ31:-MdJ5vZtAx-MUTXuNJ32:bv19942) ([raw file](https://github.com/google/nomulus/blob/8c82f881e3c90a16c05154d91f38cb0c510b9e7e/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L133)):*
> ```Java
>       if (clock.nowUtc().isAfter(replayTimeoutTime)) {
>             "Skipping %d files due to replay timeout.", commitLogFiles.size() - processedFiles);
> ```

Include the number of files replayed successfully here too and then change the `break` to a `return`.  These are terminal logging statements so it doesn't really make sense to include both in the skipping case.

Also this message should be a little bit clearer.  It's not a "replay timeout" per se, it's a request timeout.  Something like "Reached maximum execution time after replaying %d files; ending execution with %d files remaining to go on next run." is clearer.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1224)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 129 at r1](https://reviewable.io/reviews/google/nomulus/1224#-MdJ6xHm0jC81QFxN_3K:-MdJBxoJEPKplIDZB-7u:bid3voj) ([raw file](https://github.com/google/nomulus/blob/8c82f881e3c90a16c05154d91f38cb0c510b9e7e/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L129)):*

I'm curious what the maximum amount of time it might take to process a single commit log file might be.  Seeing as how each commit log file contains all of the transactions from a several minute period (the duration between runs of the export commit log action), this could be a non-trivial amount of time in times of high QPS.  We can run with this for now as it is an improvement, but we might still have to consider how we might somehow do partial replays of files w/ resumes in the future if that becomes necessary, if a single one can still take way too long to run.
</blockquote></details>

We should have a better idea once this is deployed on sandbox, hopefully 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 133 at r1](https://reviewable.io/reviews/google/nomulus/1224#-MdJ5vZtAx-MUTXuNJ31:-MdJBtZs5fSn1_O4uHry:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8c82f881e3c90a16c05154d91f38cb0c510b9e7e/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L133)):*

Include the number of files replayed successfully here too and then change the `break` to a `return`.  These are terminal logging statements so it doesn't really make sense to include both in the skipping case.

Also this message should be a little bit clearer.  It's not a "replay timeout" per se, it's a request timeout.  Something like "Reached maximum execution time after replaying %d files; ending execution with %d files remaining to go on next run." is clearer.
</blockquote></details>

Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1224)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 133 at r1](https://reviewable.io/reviews/google/nomulus/1224#-MdJ5vZtAx-MUTXuNJ31:-MdJLJws2anOLXBE6tjw:b-dfnk72) ([raw file](https://github.com/google/nomulus/blob/8c82f881e3c90a16c05154d91f38cb0c510b9e7e/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L133)):*

Done.
</blockquote></details>

You didn't like my semicolon huh :/

Also maybe "Exceeded max execution time ..." is better than "Reached..."

PR #1234, opened 2021-07-02
Add a dry-run option to commit-log-replay action and use it in Sandbox

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1234)

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -14,7 +14,9 @@
 
 package google.registry.backup;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.backup.ExportCommitLogDiffAction.DIFF_FILE_PREFIX;
+import static google.registry.backup.RestoreCommitLogsAction.DRY_RUN_PARAM;
 import static google.registry.model.ofy.EntityWritePriorities.getEntityPriority;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
@@ -25,6 +27,7 @@ import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.Key;
 import com.google.appengine.tools.cloudstorage.GcsFileMetadata;
 import com.google.appengine.tools.cloudstorage.GcsService;
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.flogger.FluentLogger;
 import google.registry.model.common.DatabaseMigrationStateSchedule;
@@ -35,6 +38,7 @@ import google.registry.model.translators.VKeyTranslatorFactory;
 import google.registry.persistence.VKey;
 import google.registry.request.Action;
 import google.registry.request.Action.Method;
+import google.registry.request.Parameter;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import google.registry.schema.replay.DatastoreEntity;
@@ -78,6 +82,11 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
   @Inject GcsDiffFileLister diffLister;
   @Inject Clock clock;
 
+  /** If true, will exit after logging the commit log files that would otherwise be replayed. */
+  @Inject
+  @Parameter(DRY_RUN_PARAM)
+  boolean dryRun;
+
   @Inject
   ReplayCommitLogsToSqlAction() {}
 
@@ -108,11 +117,26 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
     }
     try {
       logger.atInfo().log("Beginning replay of commit logs.");
-      replayFiles();
-      response.setStatus(HttpServletResponse.SC_OK);
-      String message = "ReplayCommitLogsToSqlAction completed successfully.";
-      response.setPayload(message);
-      logger.atInfo().log(message);
+      ImmutableList<GcsFileMetadata> commitLogFiles = getFilesToReplay();
+      if (dryRun) {
+        response.setStatus(HttpServletResponse.SC_OK);
+        ImmutableList<String> filenames =
+            commitLogFiles.stream()
+                .limit(10)
+                .map(file -> file.getFilename().getObjectName())
+                .collect(toImmutableList());
+        String dryRunMessage =
+            "Running in dry-run mode; would have processed %d files. They are (limit 10):\n"
+                + Joiner.on('\n').join(filenames);
+        response.setPayload(dryRunMessage);
+        logger.atInfo().log(dryRunMessage);
+      } else {
+        replayFiles(commitLogFiles);
+        response.setStatus(HttpServletResponse.SC_OK);
+        String message = "ReplayCommitLogsToSqlAction completed successfully.";
+        response.setPayload(message);
+        logger.atInfo().log(message);
+      }
     } catch (Throwable t) {
       String message = "Errored out replaying files.";
       logger.atSevere().withCause(t).log(message);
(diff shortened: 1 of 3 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 85 at r1](https://reviewable.io/reviews/google/nomulus/1234#-Mdcqayk4n0zKjE2CSqe:-Mdcqayk4n0zKjE2CSqf:b-ji9w9q) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L85)):*
> ```Java
>   @Inject
>   @Parameter(DRY_RUN_PARAM)
>   boolean dryRun;
> ```

Add a Javadoc here explaining what setting this does, as it doesn't exactly do what I would naively expect it to do (which is to parse all the transactions but not apply them; this is not even doing that).
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1234#-MdcqrlbBmwbxvFidhJ-:-MdcqrlbBmwbxvFidhJ0:b-g5ly4j) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L124)):*
> ```Java
>         logger.atInfo().log(dryRunMessage);
>         response.setPayload(dryRunMessage);
>         commitLogFiles.forEach(file -> logger.atInfo().log(file.getFilename().getObjectName()));
> ```

This is missing introductory text explaining what this is, plus I think this is gonna be way too long.  Also you probably want to concatenate them into a single logging statement (using commas, newlines, whatever) rather than logging each one completely separately, as that gets expensive.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 85 at r1](https://reviewable.io/reviews/google/nomulus/1234#-Mdcqayk4n0zKjE2CSqe:-MdcwYlY7h5nTTT5jLU_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L85)):*

Add a Javadoc here explaining what setting this does, as it doesn't exactly do what I would naively expect it to do (which is to parse all the transactions but not apply them; this is not even doing that).
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1234#-MdcqrlbBmwbxvFidhJ-:-MdcwZIOAJ6OspuucONY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L124)):*

This is missing introductory text explaining what this is, plus I think this is gonna be way too long.  Also you probably want to concatenate them into a single logging statement (using commas, newlines, whatever) rather than logging each one completely separately, as that gets expensive.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1234#-MdcqrlbBmwbxvFidhJ-:-Mdd3k8pDpqfza5_2OZp:beh9z63) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L124)):*

Done.
</blockquote></details>

What's the max size of `filenames`?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1234#-MdcqrlbBmwbxvFidhJ-:-Mdd66pOAqqMg-M57YB9:bwcqtq3) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L124)):*

What's the max size of `filenames`?
</blockquote></details>

Unbounded, I suppose. I don't really think it'll matter since this is for debugging but we can always limit it to 100

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1234#-MdcqrlbBmwbxvFidhJ-:-Mdd6qYXEAfhIp6niyQT:b2idxba) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L124)):*

Unbounded, I suppose. I don't really think it'll matter since this is for debugging but we can always limit it to 100 
</blockquote></details>

I think it might matter if you blow out the logging, is the issue.  So truncating it to a reasonable size is probably merited (probably a good deal less than 100 even).  Beyond, say, 10, is anything else really even serving any purpose?  At that point you could just look at the files in the GCS explorer.

Oh, and make sure to add `<truncated>` or similar at the end when the listing is truncated for clarity's sake.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1234#-MdcqrlbBmwbxvFidhJ-:-Mdd82IpDX29kD_IE00A:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b1aae4e3bdb719061455a7aa0934b751c07fcfbe/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L124)):*

I think it might matter if you blow out the logging, is the issue.  So truncating it to a reasonable size is probably merited (probably a good deal less than 100 even).  Beyond, say, 10, is anything else really even serving any purpose?  At that point you could just look at the files in the GCS explorer.

Oh, and make sure to add `<truncated>` or similar at the end when the listing is truncated for clarity's sake.
</blockquote></details>

Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1234)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 129 at r3](https://reviewable.io/reviews/google/nomulus/1234#-MddAd5LDX1DrhAufKQk:-MddAd5LDX1DrhAufKQl:b-qvjdch) ([raw file](https://github.com/google/nomulus/blob/c83929b99fbabb710aef94bfc8d5cf40fe6e8ee3/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L129)):*
> ```Java
>         String dryRunMessage =
>             String.format(
>                 "Running in dry-run mode, would have processed %d file(s).", commitLogFiles.size());
> ```

You may as well just combine the two messages, something like:

`"Running in dry-run mode; would have processed %d files.  They are (limit 10):\n..."`

That way you'd see them both in the logs and on the response, the latter of which would help save some time when debugging things by running commands manually from the command line.

PR #1240, opened 2021-07-13
Fix runtime issues with commit-log-to-SQL replay

gbrodman:
- We now use a more intelligent prefix to narrow the listObjects search
space in GCS. Otherwise, we're returning >30k objects which can take
roughly 50 seconds. This results in a listObjects time of 1-3 seconds.

- We now search hour by hour to efficiently make use of the prefixing.
Basically, we keep searching for new files until we hit the current time
or until we hit the overall replay timeout.

- Dry-run only prints out the first hour's worth of files

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1240)

### DIFF
--- a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java
+++ b/core/src/main/java/google/registry/backup/GcsDiffFileLister.java
@@ -15,6 +15,7 @@
 package google.registry.backup;
 
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.backup.BackupUtils.GcsMetadataKeys.LOWER_BOUND_CHECKPOINT;
 import static google.registry.backup.ExportCommitLogDiffAction.DIFF_FILE_PREFIX;
 import static google.registry.util.DateTimeUtils.START_OF_TIME;
@@ -23,13 +24,13 @@ import static google.registry.util.DateTimeUtils.latestOf;
 
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.UncheckedExecutionException;
-import dagger.Lazy;
 import google.registry.backup.BackupModule.Backups;
 import google.registry.gcs.GcsUtils;
 import java.io.IOException;
@@ -39,6 +40,7 @@ import java.util.TreeMap;
 import java.util.concurrent.ScheduledExecutorService;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import javax.inject.Provider;
 import org.joda.time.DateTime;
 
 /** Utility class to list commit logs diff files stored on GCS. */
@@ -51,7 +53,7 @@ class GcsDiffFileLister {
 
   @Inject GcsUtils gcsUtils;
 
-  @Inject @Backups Lazy<ListeningExecutorService> lazyExecutor;
+  @Inject @Backups Provider<ListeningExecutorService> executorProvider;
   @Inject ScheduledExecutorService scheduledExecutorService;
 
   @Inject
@@ -113,22 +115,27 @@ class GcsDiffFileLister {
     // (extracted from the filename) to its asynchronously-loaded metadata, keeping only files with
     // an upper checkpoint time > fromTime.
     TreeMap<DateTime, ListenableFuture<BlobInfo>> upperBoundTimesToBlobInfo = new TreeMap<>();
-    ImmutableList<String> strippedFilenames;
+    String commitLogDiffPrefix = getCommitLogDiffPrefix(fromTime, toTime);
+    ImmutableList<String> filenames;
     try {
-      strippedFilenames = gcsUtils.listFolderObjects(gcsBucket, DIFF_FILE_PREFIX);
+      filenames =
+          gcsUtils.listFolderObjects(gcsBucket, commitLogDiffPrefix).stream()
+              .map(s -> commitLogDiffPrefix + s)
+              .collect(toImmutableList());
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
     DateTime lastUpperBoundTime = START_OF_TIME;
 
     TreeMap<DateTime, BlobInfo> sequence = new TreeMap<>();
+    ListeningExecutorService executor = executorProvider.get();
     try {
-      for (String strippedFilename : strippedFilenames) {
-        final String filename = DIFF_FILE_PREFIX + strippedFilename;
+      for (String filename : filenames) {
+        String strippedFilename = filename.replaceFirst(DIFF_FILE_PREFIX, "");
         DateTime upperBoundTime = DateTime.parse(strippedFilename);
         if (isInRange(upperBoundTime, fromTime, toTime)) {
           upperBoundTimesToBlobInfo.put(
-              upperBoundTime, lazyExecutor.get().submit(() -> getBlobInfo(gcsBucket, filename)));
+              upperBoundTime, executor.submit(() -> getBlobInfo(gcsBucket, filename)));
           lastUpperBoundTime = latestOf(upperBoundTime, lastUpperBoundTime);
         }
       }
@@ -173,7 +180,7 @@ class GcsDiffFileLister {
           "Unable to compute commit diff history, there are either gaps or forks in the history "
               + "file set.  Check log for details.");
     } finally {
-      lazyExecutor.get().shutdown();
+      executor.shutdown();
     }
 
     logger.atInfo().log(
@@ -197,5 +204,37 @@ class GcsDiffFileLister {
 
   private BlobInfo getBlobInfo(String gcsBucket, String filename) {
     return gcsUtils.getBlobInfo(BlobId.of(gcsBucket, filename));
+  }
+
+  /**
+   * Returns a prefix guaranteed to cover all commit log diff files in the given range.
+   *
+   * <p>The listObjects call can be fairly slow if we search over many thousands or tens of
+   * thousands of files, so we restrict the search space. The commit logs have a file format of
+   * "commit_diff_until_2021-05-11T06:48:00.070Z" so we can often filter down as far as the hour.
+   *
+   * <p>Here, we get the longest prefix possible based on which fields (year, month, day, hour) the
+   * times in question have in common.
+   */
+  @VisibleForTesting
+  static String getCommitLogDiffPrefix(DateTime from, @Nullable DateTime to) {
+    StringBuilder result = new StringBuilder(DIFF_FILE_PREFIX);
+    if (to == null || from.getYear() != to.getYear()) {
+      return result.toString();
+    }
+    result.append(from.getYear()).append('-');
+    if (from.getMonthOfYear() != to.getMonthOfYear()) {
+      return result.toString();
+    }
+    result.append(String.format("%02d-", from.getMonthOfYear()));
+    if (from.getDayOfMonth() != to.getDayOfMonth()) {
+      return result.toString();
+    }
+    result.append(String.format("%02dT", from.getDayOfMonth()));
+    if (from.getHourOfDay() != to.getHourOfDay()) {
+      return result.toString();
+    }
+    result.append(String.format("%02d:", from.getHourOfDay()));
+    return result.toString();
   }
 }
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -20,13 +20,15 @@ import static google.registry.backup.RestoreCommitLogsAction.DRY_RUN_PARAM;
 import static google.registry.model.ofy.EntityWritePriorities.getEntityPriority;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.util.DateTimeUtils.isAtOrAfter;
+import static google.registry.util.DateTimeUtils.isBeforeOrAt;
 import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
 import static org.joda.time.Duration.standardHours;
 
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.Key;
 import com.google.cloud.storage.BlobInfo;
-import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig.Config;
@@ -120,26 +122,15 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
     }
     try {
       logger.atInfo().log("Beginning replay of commit logs.");
-      ImmutableList<BlobInfo> commitLogFiles = getFilesToReplay();
+      String resultMessage;
       if (dryRun) {
-        response.setStatus(HttpServletResponse.SC_OK);
-        ImmutableList<String> filenames =
-            commitLogFiles.stream()
-                .limit(10)
-                .map(file -> file.getName())
-                .collect(toImmutableList());
-        String dryRunMessage =
-            "Running in dry-run mode; would have processed %d files. They are (limit 10):\n"
-                + Joiner.on('\n').join(filenames);
-        response.setPayload(dryRunMessage);
-        logger.atInfo().log(dryRunMessage);
+        resultMessage = executeDryRun();
       } else {
-        replayFiles(commitLogFiles);
-        response.setStatus(HttpServletResponse.SC_OK);
-        String message = "ReplayCommitLogsToSqlAction completed successfully.";
-        response.setPayload(message);
-        logger.atInfo().log(message);
+        resultMessage = replayFiles();
       }
+      response.setStatus(SC_OK);
+      response.setPayload(resultMessage);
+      logger.atInfo().log(resultMessage);
     } catch (Throwable t) {
       String message = "Errored out replaying files.";
       logger.atSevere().withCause(t).log(message);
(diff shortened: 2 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 186 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tR002m-75DxXd3T7:-Me_tR002m-75DxXd3T8:b75y7v4) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L186)):*
> ```Java
> 
>   /**
>    * Returns a prefix guaranteed to cover all commit log diff files in the given range.
> ```

Needs a little bit more explanation as to what this is doing and how it actually works.  The key insight, that I only realized upon reading the code, is that it depends on whether the from/to times have the year/month/day/hour in common.
___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-Me_tbabDZ0f_H7_2UAo:b-7v4kli) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*
> ```Java
>       return result.toString();
>     }
>     result.append(String.format("%02dT", from.getDayOfMonth()));
> ```

I think there's a bit of a potential performance issue here though, namely, that fully half the time (with your 30 minute look-back window), we're going to be querying the entire day rather than the specific hour.  Is there a way to do better?  Like could we query for files beginning with, say, `2021-07-14T07` *and* `2021-07-14T08` for a from/to time period that overlaps that boundary, rather than just `2021-07-14`?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 154 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_sPcJ8Icu0x7I36in:-Me_sPcJ8Icu0x7I36io:b-5si5ul) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L154)):*
> ```Java
>     DateTime fromTime = jpaTm().transact(() -> SqlReplayCheckpoint.get().plusMillis(1));
>     logger.atInfo().log("Starting replay from: %s.", fromTime);
>     // Limit search space to 30 minutes after the last checkpoint so that the action runs quickly
> ```

What happens if this only returns a tiny number of files that take an inconsequential amount of time to process?  Does it go onto the next 30 minute chunk, and the next, etc., until it hits either the end of all such files or the `replayTimeoutTime`?
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 184 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_siK89NbPBQKR0IEx:-Me_siK89NbPBQKR0IEy:b-khd8fs) ([raw file](https://github.com/google/nomulus/blob/676616a1721363bfc2ad98078b789ef0cda04d0c/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L184)):*
> ```Java
>   private void processFile(BlobInfo metadata) {
>     logger.atInfo().log(
>         "Processing commit log file %s of size %d B.", metadata.getName(), metadata.getSize());
> ```

Why are you removing these logging statements?
___
*[core/src/test/java/google/registry/backup/GcsDiffFileListerTest.java, line 200 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_t-Qn6T-LyD_pFR9V:-Me_t-Qn6T-LyD_pFR9W:bfi6qe8) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/test/java/google/registry/backup/GcsDiffFileListerTest.java#L200)):*
> ```Java
>   void testPrefix_lengthened() {
>     DateTime from = DateTime.parse("2021-05-11T06:48:00.070Z");
>     assertThat(getCommitLogDiffPrefix(from, null)).isEqualTo(DIFF_FILE_PREFIX);
> ```

Don't use `DIFF_FILE_PREFIX` in the tests (unnecessary logic) -- just make assertions on the actual complete filename prefixes.  That makes the tests easier to understand too.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-Me_u26C7qNv-HtMqGT2:bob4acc) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*

I think there's a bit of a potential performance issue here though, namely, that fully half the time (with your 30 minute look-back window), we're going to be querying the entire day rather than the specific hour.  Is there a way to do better?  Like could we query for files beginning with, say, `2021-07-14T07` *and* `2021-07-14T08` for a from/to time period that overlaps that boundary, rather than just `2021-07-14`?
</blockquote></details>

I think a better way to conceptualize this might be to always query for a single hour at a time (the hour of the last replay checkpoint), and then just keep going one more hour at a time if you finish processing the previous hour before hitting the `replayTimeoutTime`.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 186 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tR002m-75DxXd3T7:-Me_zVya8IJWZGeQwxvL:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L186)):*

Needs a little bit more explanation as to what this is doing and how it actually works.  The key insight, that I only realized upon reading the code, is that it depends on whether the from/to times have the year/month/day/hour in common.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-Mea0l2I0yfrCRpOb_BS:b56kgp8) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*

I think a better way to conceptualize this might be to always query for a single hour at a time (the hour of the last replay checkpoint), and then just keep going one more hour at a time if you finish processing the previous hour before hitting the `replayTimeoutTime`.
</blockquote></details>

Unfortunately the latter suggestion doesn't work. We set the SqlReplayCheckpoint based on the timestamp of the last file we find, then in the future we search forward from that time plus one millisecond. This means that we'd always get stuck on the hour boundary, since we wouldn't be finding the files showing up in the next "hour". 

The former suggestion would work for this use case, but it means that we'd have to change around (or duplicate) a lot of the logic for RestoreCommitLogsAction, which takes as input arbitrary times. 

In the end, it's not a big deal to search across an entire day; it still only takes about a second. The big gain here comes from going from 30k+ files to ~500 files. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 154 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_sPcJ8Icu0x7I36in:-Me_z_vV1-bhrmoQbIEm:bweq6uf) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L154)):*

What happens if this only returns a tiny number of files that take an inconsequential amount of time to process?  Does it go onto the next 30 minute chunk, and the next, etc., until it hits either the end of all such files or the `replayTimeoutTime`?
</blockquote></details>

No, it just waits for the next run. This is running every 3 minutes so it's not a huge deal if it doesn't process as much as possible -- it'll just cover the next 30 minutes (or however long) next run
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 184 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_siK89NbPBQKR0IEx:-Me_zij-EUe6MGqExX2A:bcegj7) ([raw file](https://github.com/google/nomulus/blob/676616a1721363bfc2ad98078b789ef0cda04d0c/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L184)):*

Why are you removing these logging statements?
</blockquote></details>

They're quite noisy, and unnecessary based on the other logging statements that exist (this is in practice, when looking at runs on alpha that contained roughly 10 files apiece) 
___
*[core/src/test/java/google/registry/backup/GcsDiffFileListerTest.java, line 200 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_t-Qn6T-LyD_pFR9V:-Mea-IuZ1hzo2NwlUCXh:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/test/java/google/registry/backup/GcsDiffFileListerTest.java#L200)):*

Don't use `DIFF_FILE_PREFIX` in the tests (unnecessary logic) -- just make assertions on the actual complete filename prefixes.  That makes the tests easier to understand too.
</blockquote></details>

Done.

gbrodman (commented):
I guess we may wish to increase the search time from 30 minutes to something larger, just in case there's a commit log gap (there should never be one, but you never know)? I assume we wish to avoid the situation where there's a gap larger than our search space. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240#-:-Mea4PVZ2ijIQc0tg-a0:bmuo1j3)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-MeaN84w9QAvqpvFgzBh:b-j3xzue) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*

Unfortunately the latter suggestion doesn't work. We set the SqlReplayCheckpoint based on the timestamp of the last file we find, then in the future we search forward from that time plus one millisecond. This means that we'd always get stuck on the hour boundary, since we wouldn't be finding the files showing up in the next "hour". 

The former suggestion would work for this use case, but it means that we'd have to change around (or duplicate) a lot of the logic for RestoreCommitLogsAction, which takes as input arbitrary times. 

In the end, it's not a big deal to search across an entire day; it still only takes about a second. The big gain here comes from going from 30k+ files to ~500 files. 
</blockquote></details>

Why would we get stuck on the hour boundary if we query one hour at a time and then keep marching forward until we stop at the present time?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-MeaNN0t9pST2-HRNVii:b-4q7nrx) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*

Why would we get stuck on the hour boundary if we query one hour at a time and then keep marching forward until we stop at the present time?
</blockquote></details>

Finding zero files to process for the current hour just means you quickly query the next hour, up until there are no hours left to query because we've reached the future which hasn't happened yet.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 154 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_sPcJ8Icu0x7I36in:-MeaNgjqBEFoFo4HycLx:bhj32sq) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L154)):*

No, it just waits for the next run. This is running every 3 minutes so it's not a huge deal if it doesn't process as much as possible -- it'll just cover the next 30 minutes (or however long) next run
</blockquote></details>

But why wait for the next run?  Why not just keep going on?  It seems like there's a loop missing.  You're artificially limiting yourself to 10X replay speed for no reason.  That be a real, annoying limitation in testing environments where test data is sparse.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 197 at r2](https://reviewable.io/reviews/google/nomulus/1240#-MeaP_VR-4Zxy69FwtkI:-MeaP_VR-4Zxy69FwtkJ:b-hp82fo) ([raw file](https://github.com/google/nomulus/blob/676616a1721363bfc2ad98078b789ef0cda04d0c/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L197)):*
> ```Java
>       SqlReplayCheckpoint.set(checkpoint);
>       logger.atInfo().log(
>           "Replayed %d transactions from commit log file %s.",
> ```

Can you add the file size to this logging statement then since you're removing the one that contained it?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-Mea_Ioa0TmP2zvbOtOP:bhu281h) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*

Finding zero files to process for the current hour just means you quickly query the next hour, up until there are no hours left to query because we've reached the future which hasn't happened yet.
</blockquote></details>

There's also the issue that, once a month, the current implementation will attempt to read an entire month's worth of files ... if that times out and permanently fails, then we're stuck at that point.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/GcsDiffFileLister.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_tbabDZ0f_H7_2UAn:-MekfHET70Kmho4pq4Nm:bynlz1o) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/GcsDiffFileLister.java#L206)):*

There's also the issue that, once a month, the current implementation will attempt to read an entire month's worth of files ... if that times out and permanently fails, then we're stuck at that point.
</blockquote></details>

Yeah, fair enough.This is done -- we basically search hour by hour until we time out or until we hit the current hour. 

Future runs shouldn't be affected if we search too far in the future since we only update the checkpoint time after processing a file. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 154 at r1](https://reviewable.io/reviews/google/nomulus/1240#-Me_sPcJ8Icu0x7I36in:-MekfPSDBwYcCi50wiSk:b-hzgtp1) ([raw file](https://github.com/google/nomulus/blob/8ffe39c5350c0b522a7ed93e8dc273028237b85a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L154)):*

But why wait for the next run?  Why not just keep going on?  It seems like there's a loop missing.  You're artificially limiting yourself to 10X replay speed for no reason.  That be a real, annoying limitation in testing environments where test data is sparse.
</blockquote></details>

See other comment, this should be done now. 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 197 at r2](https://reviewable.io/reviews/google/nomulus/1240#-MeaP_VR-4Zxy69FwtkI:-MeakoF_-bbFrG2Z49H-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/676616a1721363bfc2ad98078b789ef0cda04d0c/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L197)):*

Can you add the file size to this logging statement then since you're removing the one that contained it?
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 167 at r3](https://reviewable.io/reviews/google/nomulus/1240#-MfDgLFJ0Is5FwN7vBaG:-MfDgLFJ0Is5FwN7vBaH:bpond6u) ([raw file](https://github.com/google/nomulus/blob/6121ed0f1f53d04042aedfb0495ed34a17b887b4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L167)):*
> ```Java
>     int filesProcessed = 0;
>     while (true) {
>       if (!clock.nowUtc().isBefore(replayTimeoutTime)) {
> ```

You can rewrite this to not require a negation, which will make it cognitively simpler to parse.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 171 at r3](https://reviewable.io/reviews/google/nomulus/1240#-MfDgR1GCFUThJTVTZH8:-MfDgR1GCFUThJTVTZH9:bc0uing) ([raw file](https://github.com/google/nomulus/blob/6121ed0f1f53d04042aedfb0495ed34a17b887b4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L171)):*
> ```Java
>             "Reached max execution time after replaying %d file(s).", filesProcessed);
>       }
>       if (!searchStartTime.isBefore(clock.nowUtc())) {
> ```

Ditto.

Also there should be consistency on whether now is on the left or on the right in these two comparisons.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 179 at r3](https://reviewable.io/reviews/google/nomulus/1240#-MfDh6EGB3MfFUZ2ZgPQ:-MfDh6EGB3MfFUZ2ZgPR:b5u1xnt) ([raw file](https://github.com/google/nomulus/blob/6121ed0f1f53d04042aedfb0495ed34a17b887b4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L179)):*
> ```Java
>           searchStartTime.withMinuteOfHour(59).withSecondOfMinute(59).withMillisOfSecond(999);
>       ImmutableList<BlobInfo> fileBatch =
>           diffLister.listDiffFiles(gcsBucket, searchStartTime, searchEndTime);
> ```

This can return an empty list (e.g. if the cursor is already close to the end of the our).  Perhaps logging that the hour has been finished, and we're moving onto the next one, would be helpful in that circumstance?  Right now it looks like that does work, but perhaps only accidentally (intention not clear from reading the code).

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 167 at r3](https://reviewable.io/reviews/google/nomulus/1240#-MfDgLFJ0Is5FwN7vBaG:-MfDmZjd0UQOwJbK3UjT:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6121ed0f1f53d04042aedfb0495ed34a17b887b4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L167)):*

You can rewrite this to not require a negation, which will make it cognitively simpler to parse.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 171 at r3](https://reviewable.io/reviews/google/nomulus/1240#-MfDgR1GCFUThJTVTZH8:-MfDmnZC3QpylViQgIzk:b-ecsndb) ([raw file](https://github.com/google/nomulus/blob/6121ed0f1f53d04042aedfb0495ed34a17b887b4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L171)):*

Ditto.

Also there should be consistency on whether now is on the left or on the right in these two comparisons.
</blockquote></details>

Done.

woo DTUtils
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 179 at r3](https://reviewable.io/reviews/google/nomulus/1240#-MfDh6EGB3MfFUZ2ZgPQ:-MfDn10LBhNnpfkdbLHr:b-fnhu42) ([raw file](https://github.com/google/nomulus/blob/6121ed0f1f53d04042aedfb0495ed34a17b887b4/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L179)):*

This can return an empty list (e.g. if the cursor is already close to the end of the our).  Perhaps logging that the hour has been finished, and we're moving onto the next one, would be helpful in that circumstance?  Right now it looks like that does work, but perhaps only accidentally (intention not clear from reading the code).
</blockquote></details>

It is intentional, but yeah we can log it to be even more clear. The premise is just "continue searching hour by hour until we hit the current time or we run too long".

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 168 at r4](https://reviewable.io/reviews/google/nomulus/1240#-MfDsgQkEtScvdQO1zBZ:-MfDsgQkEtScvdQO1zB_:b-u4195t) ([raw file](https://github.com/google/nomulus/blob/f51917d135261716275c2b3a7ae302e975fa622f/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L168)):*
> ```Java
>     DateTime searchStartTime = jpaTm().transact(() -> SqlReplayCheckpoint.get().plusMillis(1));
>     int filesProcessed = 0;
>     while (true) {
> ```

Adding a few lines of comments here explaining the overall search strategy in plain English might be helpful.  Something like "Starting from the last successfully imported logs, import logs one hour-long batch at a time until we either catch up to the present or have been running too long (in which case we terminate to flush logs and let the next run of the cronjob continue).  Hour-long batches are used because GCS supports a filename prefix search strategy."

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240#-:-MfDtKmV2-2VNrQHrQ0d:bnfp4nl)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1240)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 168 at r4](https://reviewable.io/reviews/google/nomulus/1240#-MfDsgQkEtScvdQO1zBZ:-MfDxaca9Xa7bYGjs0RW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f51917d135261716275c2b3a7ae302e975fa622f/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L168)):*

Adding a few lines of comments here explaining the overall search strategy in plain English might be helpful.  Something like "Starting from the last successfully imported logs, import logs one hour-long batch at a time until we either catch up to the present or have been running too long (in which case we terminate to flush logs and let the next run of the cronjob continue).  Hour-long batches are used because GCS supports a filename prefix search strategy."
</blockquote></details>

Done.

PR #1241, opened 2021-07-14
Create a ReadOnlyTransactionManager for SQL migration

gbrodman:
We don't use it yet, but in the future we will have tm() return a
ReadOnlyTransactionManager wrapping a Datastore/SQL delegate (as
appropriate) when we are in the read-only segments of the migration.
That way, tm() calls will still be read-only but the asynchronous
migration's jpaTm() or ofyTm() calls will still always function.

### DIFF
--- a/core/src/main/java/google/registry/persistence/transaction/ReadOnlyTransactionManager.java
+++ b/core/src/main/java/google/registry/persistence/transaction/ReadOnlyTransactionManager.java
@@ -1,0 +1,268 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.persistence.transaction;
+
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import google.registry.persistence.VKey;
+import java.util.Optional;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+import org.joda.time.DateTime;
+
+/**
+ * A transaction manager that only supports read operations.
+ *
+ * <p>This is used during the Registry 3.0 Datastore-to-SQL migration during the read-only phases so
+ * that we can prevent any new writes from occurring -- the only changes during the read-only phases
+ * should be the asynchronous transaction replay. See {@link
+ * google.registry.model.common.DatabaseMigrationStateSchedule} for more information.
+ */
+public class ReadOnlyTransactionManager implements TransactionManager {
+
+  private final TransactionManager delegate;
+
+  public ReadOnlyTransactionManager(TransactionManager delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public boolean inTransaction() {
+    return delegate.inTransaction();
+  }
+
+  @Override
+  public void assertInTransaction() {
+    delegate.assertInTransaction();
+  }
+
+  @Override
+  public <T> T transact(Supplier<T> work) {
+    return delegate.transact(work);
+  }
+
+  @Override
+  public void transact(Runnable work) {
+    delegate.transact(work);
+  }
+
+  @Override
+  public <T> T transactNew(Supplier<T> work) {
+    return delegate.transactNew(work);
+  }
+
+  @Override
+  public void transactNew(Runnable work) {
+    delegate.transactNew(work);
+  }
+
+  @Override
+  public <R> R transactNewReadOnly(Supplier<R> work) {
+    return delegate.transactNewReadOnly(work);
+  }
+
+  @Override
+  public void transactNewReadOnly(Runnable work) {
+    delegate.transactNewReadOnly(work);
+  }
+
+  @Override
+  public <R> R doTransactionless(Supplier<R> work) {
+    return delegate.doTransactionless(work);
+  }
+
+  @Override
+  public DateTime getTransactionTime() {
+    return delegate.getTransactionTime();
+  }
+
+  @Override
+  public void insert(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void insertAll(ImmutableCollection<?> entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void insertWithoutBackup(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void insertAllWithoutBackup(ImmutableCollection<?> entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void put(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void putAll(Object... entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void putAll(ImmutableCollection<?> entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void putWithoutBackup(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void putAllWithoutBackup(ImmutableCollection<?> entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void update(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void updateAll(ImmutableCollection<?> entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void updateAll(Object... entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void updateWithoutBackup(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void updateAllWithoutBackup(ImmutableCollection<?> entities) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public boolean exists(Object entity) {
+    return delegate.exists(entity);
+  }
+
+  @Override
+  public <T> boolean exists(VKey<T> key) {
+    return delegate.exists(key);
+  }
+
+  @Override
+  public <T> Optional<T> loadByKeyIfPresent(VKey<T> key) {
+    return delegate.loadByKeyIfPresent(key);
+  }
+
+  @Override
+  public <T> ImmutableMap<VKey<? extends T>, T> loadByKeysIfPresent(
+      Iterable<? extends VKey<? extends T>> vKeys) {
+    return delegate.loadByKeysIfPresent(vKeys);
+  }
+
+  @Override
+  public <T> ImmutableList<T> loadByEntitiesIfPresent(Iterable<T> entities) {
+    return delegate.loadByEntitiesIfPresent(entities);
+  }
+
+  @Override
+  public <T> T loadByKey(VKey<T> key) {
+    return delegate.loadByKey(key);
+  }
+
+  @Override
+  public <T> ImmutableMap<VKey<? extends T>, T> loadByKeys(
+      Iterable<? extends VKey<? extends T>> vKeys) {
+    return delegate.loadByKeys(vKeys);
+  }
+
+  @Override
+  public <T> T loadByEntity(T entity) {
+    return delegate.loadByEntity(entity);
+  }
+
+  @Override
+  public <T> ImmutableList<T> loadByEntities(Iterable<T> entities) {
+    return delegate.loadByEntities(entities);
+  }
+
+  @Override
+  public <T> ImmutableList<T> loadAllOf(Class<T> clazz) {
+    return delegate.loadAllOf(clazz);
+  }
+
+  @Override
+  public <T> Stream<T> loadAllOfStream(Class<T> clazz) {
+    return delegate.loadAllOfStream(clazz);
+  }
+
+  @Override
+  public <T> Optional<T> loadSingleton(Class<T> clazz) {
+    return delegate.loadSingleton(clazz);
+  }
+
+  @Override
+  public void delete(VKey<?> key) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void delete(Iterable<? extends VKey<?>> keys) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public <T> T delete(T entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void deleteWithoutBackup(VKey<?> key) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void deleteWithoutBackup(Iterable<? extends VKey<?>> keys) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public void deleteWithoutBackup(Object entity) {
+    throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
+  }
+
+  @Override
+  public <T> QueryComposer<T> createQueryComposer(Class<T> entity) {
+    return delegate.createQueryComposer(entity);
+  }
+
+  @Override
+  public void clearSessionCache() {
+    delegate.clearSessionCache();
+  }
+
+  @Override
+  public boolean isOfy() {
+    return delegate.isOfy();
+  }
+}
--- a/core/src/test/java/google/registry/persistence/transaction/ReadOnlyTransactionManagerTest.java
+++ b/core/src/test/java/google/registry/persistence/transaction/ReadOnlyTransactionManagerTest.java
@@ -1,0 +1,64 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.persistence.transaction;
+
+import static com.google.common.truth.Truth.assertThat;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static org.junit.Assert.assertThrows;
+
+import google.registry.model.registrar.Registrar;
+import google.registry.persistence.VKey;
+import google.registry.testing.AppEngineExtension;
+import google.registry.testing.DualDatabaseTest;
+import google.registry.testing.TestOfyAndSql;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+/** Unit tests for {@link ReadOnlyTransactionManager}. */
+@DualDatabaseTest
+public class ReadOnlyTransactionManagerTest {
+
+  @RegisterExtension
+  final AppEngineExtension appEngine =
+      AppEngineExtension.builder().withDatastoreAndCloudSql().build();
+
+  private TransactionManager readOnlyTm;
+  private VKey<Registrar> registrarKey;
+
+  @BeforeEach
+  void beforeEach() {
+    readOnlyTm = new ReadOnlyTransactionManager(tm());
+    registrarKey = AppEngineExtension.makeRegistrar2().createVKey();
+  }
+
+  @TestOfyAndSql
+  void testRead_succeeds() {
+    assertThat(readOnlyTm.transact(() -> readOnlyTm.loadByKey(registrarKey).getEmailAddress()))
+        .isEqualTo("the.registrar@example.com");
+  }
+
+  @TestOfyAndSql
+  void testWrite_fails() {
+    readOnlyTm.transact(
+        () -> {
+          Registrar registrar = readOnlyTm.loadByKey(registrarKey);
+          Registrar modified = registrar.asBuilder().setEmailAddress("foo@bar.com").build();
+          assertThat(
+                  assertThrows(UnsupportedOperationException.class, () -> readOnlyTm.put(modified)))
+              .hasMessageThat()
+              .isEqualTo("Transaction manager currently in read-only mode");
+        });
+  }
+}


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyTransactionManager.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdriVi6Im6awJOR-jD:-MfdriVi6Im6awJOR-jE:bcjm19w) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/ReadOnlyTransactionManager.java#L94)):*
> ```Java
>   @Override
>   public void insert(Object entity) {
>     throw new UnsupportedOperationException("Transaction manager currently in read-only mode");
> ```

Centralize this error message in one place rather than repeating this string a dozen times in the file.

Might make sense to have a helper function that centralizes the throwing too (so the exception type is centralized).

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-MfdssvZCIlcSuKc9C_K:b-6f795c) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*
> ```Java
>   /** Supplier for the read-only jpaTm so that it is initialized only once on request. */
>   private static Supplier<ReadOnlyTransactionManager> readOnlyJpaTm =
>       Suppliers.memoize(() -> new ReadOnlyTransactionManager(jpaTm.get()));
> ```

What about `JpaTransactionManager`?  There are additional mutating methods on there that need to throw, including most problematically `jpaTm().getEntityManager()`  But I don't see that happening in `ReadOnlyTransactionManager`.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyTransactionManager.java, line 94 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdriVi6Im6awJOR-jD:-Mfdu0EhDzeogDN4jXVv:b-xsro58) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/ReadOnlyTransactionManager.java#L94)):*

Centralize this error message in one place rather than repeating this string a dozen times in the file.

Might make sense to have a helper function that centralizes the throwing too (so the exception type is centralized).
</blockquote></details>

Done (helper to centralize creation, not throwing, since otherwise it wouldn't compile) 
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-MfdtMs502X7SKPCipQ9:bjzz2jk) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*

What about `JpaTransactionManager`?  There are additional mutating methods on there that need to throw, including most problematically `jpaTm().getEntityManager()`  But I don't see that happening in `ReadOnlyTransactionManager`.
</blockquote></details>

I don't believe it is necessary -- we discussed (I believe) applying the read-only nature of the transaction manager to `tm()` calls specifically, allowing explicit `ofyTm()` and `jpaTm()` calls through. 

Every replayed object should be / is using the `tm()` methods

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-MfdwYcl6s8lERzHlV2_:buo3z1o) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*

I don't believe it is necessary -- we discussed (I believe) applying the read-only nature of the transaction manager to `tm()` calls specifically, allowing explicit `ofyTm()` and `jpaTm()` calls through. 

Every replayed object should be / is using the `tm()` methods
</blockquote></details>

I don't understand; why would we want the registry to be only partially read only?  I'm not remembering this discussion.

Why should e.g. something like this be able to succeed when we're in read-only mode?  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/registry/RegistryLockDao.java;l=121;drc=58e413af89c31ffa100d4845127cb93efe25d89f;bpv=1;bpt=1

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-Mfhkk-x6Ll2fddjEw5h:bkjotx3) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*

I don't understand; why would we want the registry to be only partially read only?  I'm not remembering this discussion.

Why should e.g. something like this be able to succeed when we're in read-only mode?  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/registry/RegistryLockDao.java;l=121;drc=58e413af89c31ffa100d4845127cb93efe25d89f;bpv=1;bpt=1
</blockquote></details>

IIRC it was an offhand discussion with Mike, nothing serious. 

Registry locks aren't replayed so the lock operations themselves could succeed without issue, though the transaction-manager operations to save the new version of a domain (if we're actually performing the lock/unlock) would fail. 

I guess it's cleaner to be more thorough about it though. Check out what I have here now.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-Mfj2boLDkKWVAlO33Jw:b66iusz) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*

IIRC it was an offhand discussion with Mike, nothing serious. 

Registry locks aren't replayed so the lock operations themselves could succeed without issue, though the transaction-manager operations to save the new version of a domain (if we're actually performing the lock/unlock) would fail. 

I guess it's cleaner to be more thorough about it though. Check out what I have here now. 
</blockquote></details>

hm I guess we may wish to do the same for the `ofy` calls as well then

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-MfnqWey6Gpp6AqQ-X4m:b-5pf1vy) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*

hm I guess we may wish to do the same for the `ofy` calls as well then 
</blockquote></details>

Yeah, I think we want a read-only mode where all writes to the DB fail.  Otherwise we're not actually read-only, can't make any guarantees about such, and can lose data when switching over.

Why does this need to be implemented as a separate transaction manager using the decorator pattern though, and not simply a `isReadOnly` boolean in both of the existing transaction managers respectively, whose state is checked as the first line in every mutating method?  That seems to me to be a simpler implementation?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 55 at r2](https://reviewable.io/reviews/google/nomulus/1241#-MfdssvY1fpPH5QH9Z5z:-Mft4MFi6qqnyKBFt_Uh:b5zk250) ([raw file](https://github.com/google/nomulus/blob/52cf0dbabb5728147af3c35341eac0f96bb979ba/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L55)):*

Yeah, I think we want a read-only mode where all writes to the DB fail.  Otherwise we're not actually read-only, can't make any guarantees about such, and can lose data when switching over.

Why does this need to be implemented as a separate transaction manager using the decorator pattern though, and not simply a `isReadOnly` boolean in both of the existing transaction managers respectively, whose state is checked as the first line in every mutating method?  That seems to me to be a simpler implementation?
</blockquote></details>

That could work, though we'd still need a delegating implementation of the entity manager + queries (since we don't control that code). I guess adding a boolean check + a couple stack frames wouldn't be a big deal assuming it's cached in the DatabaseMigrationSchedule. 


I'll do that + try to get Ofy working as well -- there's a bit of awkwardness given that the thing we're using to check the schedule uses ofy itself

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 48 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 42 at r7](https://reviewable.io/reviews/google/nomulus/1241#-Mgqazjy0YTW4OGTRTsw:-Mgqazjy0YTW4OGTRTsx:blg4fhb) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L42)):*
> ```Java
>  */
> @Entity
> public class DatabaseMigrationStateSchedule extends CrossTldSingleton implements SqlOnlyEntity {
> ```

Are the changes in this class sneaking in from another PR?

CydeWeys (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 50 files reviewed, 7 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 130 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqbJbNCH16t5DtwX1Z:-MgqbJbNCH16t5DtwX1_:boqwodx) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L130)):*
> ```Java
>     }
>     EntityManager entityManager = transactionInfo.get().entityManager;
>     if (TransactionManagerFactory.isReadOnlyMode()) {
> ```

You can use a ternary here to make the intent more clear.

```
return TransactionManagerFactory.isReadOnlyMode() ? new ReadOnlyEntityManager(entityManager) : entityManager;
```
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 395 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqbY3M9rqw8zNfobWU:-MgqbY3M9rqw8zNfobWV:b-yy0ap2) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L395)):*
> ```Java
>   @Override
>   public void updateAll(ImmutableCollection<?> entities) {
>     TransactionManagerFactory.assertNotReadOnlyMode();
> ```

There's so many of these; you could statically import them and save some visual noise at the callsites.
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyEntityManager.java, line 216 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqbzDd5ZZUh4Fk95fY:-MgqbzDd5ZZUh4Fk95fZ:b-jkj1xs) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyEntityManager.java#L216)):*
> ```Java
>   @Override
>   public StoredProcedureQuery createNamedStoredProcedureQuery(String name) {
>     return delegate.createNamedStoredProcedureQuery(name);
> ```

Stored procedures can easily include mutating operations -- shouldn't all of these be throwing when called?

(Note that I don't believe we make use of any stored procedures anywhere, but if we did, they'd most likely be of the mutating variety.)
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java, line 1 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcAHG5i_TYy9MH7eD:-MgqcAHG5i_TYy9MH7eE:b7gfxl6) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java#L1)):*
> ```Java
> package google.registry.persistence.transaction;
> ```

This file is missing the license header.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 158 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcPlM95GTV4T5hLYT:-MgqcPlM95GTV4T5hLYU:b-xqz0ht) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L158)):*
> ```Java
> 
>   public static IllegalStateException readOnlyModeException() {
>     return new IllegalStateException("Transaction manager currently in read-only mode");
> ```

This exception message might be externally-readable from EPP flows, right?  If so it should be an error message that is suitable for external consumption, something like "Registry is currently in read-only mode".

Also, you should be using a custom Exception class here rather than re-using `IllegalStateException` with a custom message.  The custom exception class should still extend `IllegalStateException`.  But as a general rule of thumb, if you have a new type of exception that has specific semantic meaning in the application, and especially if it's widely used (like this one), then it merits its own exception class.  We already have well over a hundred custom Exception classes within Nomulus, so this one will be in good company.
___
*[core/src/test/java/google/registry/model/common/DatabaseMigrationStateScheduleTest.java, line 142 at r7](https://reviewable.io/reviews/google/nomulus/1241#-Mgqd9saEN8WOpKKIuIt:-Mgqd9saEN8WOpKKIuIu:b-p8xjch) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/test/java/google/registry/model/common/DatabaseMigrationStateScheduleTest.java#L142)):*
> ```Java
>     assertThat(
>             assertThrows(
>                 IllegalStateException.class,
> ```

Yeah, see, tests like these get better and more specific once you have the custom exception class; you can simply assert that the right exception class is thrown, and then not have to bother with checking the error message below.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: 0 of 50 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 42 at r7](https://reviewable.io/reviews/google/nomulus/1241#-Mgqazjy0YTW4OGTRTsw:-MgvOIm06qVPqF629Tih:b-xv8l1c) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L42)):*

Are the changes in this class sneaking in from another PR?
</blockquote></details>

TBQH I had forgotten this PR was still "under review". 

The changes are there but they're not sneaking in -- the read-only PR depends on the DatabaseMigrationStateSchedule-moving-to-SQL PR so I was cherrypicking those changes in to make sure it works 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 130 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqbJbNCH16t5DtwX1Z:-MgvPKv2BbPE2aP2r-mU:b-896fix) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L130)):*

You can use a ternary here to make the intent more clear.

```
return TransactionManagerFactory.isReadOnlyMode() ? new ReadOnlyEntityManager(entityManager) : entityManager;
```
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyEntityManager.java, line 216 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqbzDd5ZZUh4Fk95fY:-MgvPUS54D6l68Lx8PKq:b350h2c) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyEntityManager.java#L216)):*

Stored procedures can easily include mutating operations -- shouldn't all of these be throwing when called?

(Note that I don't believe we make use of any stored procedures anywhere, but if we did, they'd most likely be of the mutating variety.)
</blockquote></details>

Yeah they could be either, I suppose. I know we don't use any so I wasn't particularly concerned about it. We can throw no matter what; it's simpler that way. 
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java, line 1 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcAHG5i_TYy9MH7eD:-MgvPoYqBYFQ1YlHasXj:b-896fix) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java#L1)):*

This file is missing the license header.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 158 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcPlM95GTV4T5hLYT:-MgvRODP2vHSI8tW6QEh:b6zmsy6) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L158)):*

This exception message might be externally-readable from EPP flows, right?  If so it should be an error message that is suitable for external consumption, something like "Registry is currently in read-only mode".

Also, you should be using a custom Exception class here rather than re-using `IllegalStateException` with a custom message.  The custom exception class should still extend `IllegalStateException`.  But as a general rule of thumb, if you have a new type of exception that has specific semantic meaning in the application, and especially if it's widely used (like this one), then it merits its own exception class.  We already have well over a hundred custom Exception classes within Nomulus, so this one will be in good company.
</blockquote></details>

Good call. 
___
*[core/src/test/java/google/registry/model/common/DatabaseMigrationStateScheduleTest.java, line 142 at r7](https://reviewable.io/reviews/google/nomulus/1241#-Mgqd9saEN8WOpKKIuIt:-MgvRMxFDosJRt06Qm74:b-mn6tlq) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/test/java/google/registry/model/common/DatabaseMigrationStateScheduleTest.java#L142)):*

Yeah, see, tests like these get better and more specific once you have the custom exception class; you can simply assert that the right exception class is thrown, and then not have to bother with checking the error message below.
</blockquote></details>

Yeah, this is better.

gbrodman:
Sorry if you were in the middle of reviewing it -- I added methods here to bypass the read-only nature, for use only in replay.

jianglai (changes requested):
> Reviewed 4 of 48 files at r7, 52 of 52 files at r9.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: all files reviewed, 10 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 123 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoSCpZ255QPtBqEapd:-MhoSCpZ255QPtBqEape:b-o28yru) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L123)):*
> ```
> assertNotReadOnlyMode();
> ```
Are all these assertions necessary? I think you can just add them to the ultimate method that is called (like `saveEntity` here and `ofy.save` which already does the same assertion).
___
*[core/src/main/java/google/registry/model/ofy/Ofy.java, line 190 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoRV2m7tn0jXOSr6P5:-MhoRV2m7tn0jXOSr6P6:b-x7nlx8) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/model/ofy/Ofy.java#L190)):*
> ```
> return new AugmentedSaver() {
> ```
You can just call `saveIgnoringReadOnly` from here, right? Same for `DeleteWithoutBackup`.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 332 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoafMr7DQ-OXWZMfN7:-MhoafMr7DQ-OXWZMfN8:b-wpzrym) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L332)):*
> ```
> assertNotReadOnlyMode
> ```
Save comment as in the others, It doesn't look like we need to do the assertion everywhere.
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 681 at r9](https://reviewable.io/reviews/google/nomulus/1241#-Mhoc_h3AAWO8VRmBWag:-Mhoc_h3AAWO8VRmBWah:b-qzgbf3) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L681)):*
> ```
> deleteIgnoringReadOnly
> ```
Can't we just call `internalDelete`?
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java, line 1 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcAHG5i_TYy9MH7eD:-Mhoft5X7X35dmBqOVu6:bov61p2) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java#L1)):*

Done.
</blockquote></details>

This, and `ReadOnlyTypeQuery` and `ReadOnlyTransactionManger` all have a lot of boilerplate code because they use delegates. Is it possible to use inheritance instead of composition? There are probably only one concrete type of `TypedQuery` that we use in the code base and can be extended by this claszs, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: all files reviewed, 10 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 123 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoSCpZ255QPtBqEapd:-Mhtykej-dtSDSTJQE5B:b-9xwthn) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L123)):*

> ```
> assertNotReadOnlyMode();
> ```
Are all these assertions necessary? I think you can just add them to the ultimate method that is called (like `saveEntity` here and `ofy.save` which already does the same assertion).
</blockquote></details>

I think it's better to have them at the entry points to the public API. That way, it's harder to miss things. 
___
*[core/src/main/java/google/registry/model/ofy/Ofy.java, line 190 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoRV2m7tn0jXOSr6P5:-Mhtz1Tr0qiWCX4jfVkI:bb7867m) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/model/ofy/Ofy.java#L190)):*

> ```
> return new AugmentedSaver() {
> ```
You can just call `saveIgnoringReadOnly` from here, right? Same for `DeleteWithoutBackup`.
</blockquote></details>

Yup!
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 332 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoafMr7DQ-OXWZMfN7:-Mhtz2rA0Fh11RcUoEzu:b-ea0e5o) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L332)):*

> ```
> assertNotReadOnlyMode
> ```
Save comment as in the others, It doesn't look like we need to do the assertion everywhere.
</blockquote></details>

I think we should just to be safe; we wouldn't want any calls to sneak through. 
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 681 at r9](https://reviewable.io/reviews/google/nomulus/1241#-Mhoc_h3AAWO8VRmBWag:-MhtzGE15UDH2lzr1qa2:bdqxzmp) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L681)):*

> ```
> deleteIgnoringReadOnly
> ```
Can't we just call `internalDelete`?
</blockquote></details>

They're not quite the same, no. `internalDelete` uses the standard entity manager and query code (which goes through the read-only path) and this doesn't. 
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java, line 1 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcAHG5i_TYy9MH7eD:-MhtzTMj3HuXX1mRcPFL:b-bdet3m) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java#L1)):*

This, and `ReadOnlyTypeQuery` and `ReadOnlyTransactionManger` all have a lot of boilerplate code because they use delegates. Is it possible to use inheritance instead of composition? There are probably only one concrete type of `TypedQuery` that we use in the code base and can be extended by this claszs, right?
</blockquote></details>

`ReadOnlyTransactionManager` is gone now. 

`ReadOnlyTypedQuery`, `ReadOnlyEntityManager`, and `ReadOnlyQuery` must all use composition / delegation rather than inheritance because the non-read-only versions of those objects are interfaces, not classes. Technically we don't "know" the concrete type.

jianglai (changes requested):
> Reviewed 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java, line 123 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoSCpZ255QPtBqEapd:-Mhusjl70VwtPMhqRD7v:bq9xugq) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/model/ofy/DatastoreTransactionManager.java#L123)):*

I think it's better to have them at the entry points to the public API. That way, it's harder to miss things. 
</blockquote></details>

Really? I felt the opposite is true. For example we can imagine that people will in time add more `insert()` overload methods, and by putting the assertion inside the callers the writer will need to remember that. That seems like unnecessary mental burden. It feels more natural to check it in the more "atomic" operation (`put`) here, which we know for sure cannot operate in read only mode.

So as long as we make the insertion in all these atomic operations the callers should not need to worry about it. Plus in some methods you are making double assertions (like when you call `Ofy.save()`, which itselfs does an assertion already).
___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 332 at r9](https://reviewable.io/reviews/google/nomulus/1241#-MhoafMr7DQ-OXWZMfN7:-MhuvV9aBpZ0C6s0i021:bapqbl0) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L332)):*

I think we should just to be safe; we wouldn't want any calls to sneak through. 
</blockquote></details>

ditto above. I think it is actually safer (and cleaner) to make the assertions close to where the actual operation to the database is happening. 
___
*[core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java, line 1 at r7](https://reviewable.io/reviews/google/nomulus/1241#-MgqcAHG5i_TYy9MH7eD:-Mhv0kwi2hjvULkDd0Iw:b-2dn9hb) ([raw file](https://github.com/google/nomulus/blob/04959d8774405d987f41e544c8a59b5c3873e0a9/core/src/main/java/google/registry/persistence/transaction/ReadOnlyQuery.java#L1)):*

`ReadOnlyTransactionManager` is gone now. 

`ReadOnlyTypedQuery`, `ReadOnlyEntityManager`, and `ReadOnlyQuery` must all use composition / delegation rather than inheritance because the non-read-only versions of those objects are interfaces, not classes. Technically we don't "know" the concrete type. 
</blockquote></details>

Yes I understand they are interfaces. What I was suggesting is that in practice there is only one type of concrete class (probably provided by Hibernate if I had to guess) that we are using. But I do see that going this route can probably be more troublesome than its worth.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1241)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java, line 681 at r9](https://reviewable.io/reviews/google/nomulus/1241#-Mhoc_h3AAWO8VRmBWag:-Mhv1Q574btsz-BdGtpH:bqn26h8) ([raw file](https://github.com/google/nomulus/blob/dd9d28af3ffac192361bb7ccea8085704160efb6/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L681)):*

They're not quite the same, no. `internalDelete` uses the standard entity manager and query code (which goes through the read-only path) and this doesn't. 
</blockquote></details>

Ah I do see what you meant. However now that I've give this some more thought I am again doubting the value in adding all these assertions at all. The reason is that these assertions are effectively no-ops -- You have specialized read only classes that throw in mutation operations if in read only mode, and you also updated the transaction manager impl to always wrap queries in a read-only query when in read-only mode. That should be good enough. Your assertions are guarding against coding errors (like forgetting to wrap a query) rather than runtime errors (like malformed input), which I don't think is what assertions are ultimately good at. Look at it this way: you are basically trading one type of coding error (forgetting to wrap) with another type (forgetting to add an assertion) to ensure that your program runs correctly. That doesn't seem like a worthwhile tradeoff. If an error can in theory be caught by a static analyzer, it probably is not a good candidate for assertion. Admittedly it might actually be easier and cleaner to peruse the usage of queries in the transaction manager than to make sure each public API makes an assertion. 

The same goes for `DatastoreTransactionManager`. You need the assertions in `Ofy` because you don't have a "read-only" `Ofy`, but since all `DatastoreTransactionManager` does is to delegate to `Ofy` it should not worry about making the assertion anymore. If you go down the rabbit hole of checking for all the possible ways things can go wrong when calling a function at the call site, you are making things too difficult for yourself.

PR #1246, opened 2021-07-19
Fix flaky test issues caused by lack of ofy init

gbrodman:
These aren't "flakes" in the usual sense but it doesn't seem like you show up if you run the standard testing tasks through Gradle. Rather, they show up if you run the individual tests. 

Using `tm()` calls now has a dependency on Objectify -- JpaTransactionManagerImplTest shouldn't need that so we can modify the assertDetached calls to remove that. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1246)

### DIFF
--- a/core/src/test/java/google/registry/beam/rde/RdePipelineTest.java
+++ b/core/src/test/java/google/registry/beam/rde/RdePipelineTest.java
@@ -23,7 +23,6 @@ import static google.registry.model.common.Cursor.CursorType.RDE_STAGING;
 import static google.registry.model.rde.RdeMode.FULL;
 import static google.registry.model.rde.RdeMode.THIN;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
-import static google.registry.persistence.transaction.TransactionManagerFactory.setTmForTest;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.rde.RdeResourceType.CONTACT;
 import static google.registry.rde.RdeResourceType.DOMAIN;
@@ -50,7 +49,7 @@ import google.registry.model.registrar.Registrar;
 import google.registry.model.registrar.Registrar.State;
 import google.registry.persistence.transaction.JpaTestRules;
 import google.registry.persistence.transaction.JpaTestRules.JpaIntegrationTestExtension;
-import google.registry.persistence.transaction.TransactionManager;
+import google.registry.persistence.transaction.TransactionManagerFactory;
 import google.registry.rde.DepositFragment;
 import google.registry.rde.PendingDeposit;
 import google.registry.rde.RdeResourceType;
@@ -117,12 +116,9 @@ public class RdePipelineTest {
 
   private RdePipeline rdePipeline;
 
-  private TransactionManager originalTm;
-
   @BeforeEach
   void beforeEach() throws Exception {
-    originalTm = tm();
-    setTmForTest(jpaTm());
+    TransactionManagerFactory.setTmForTest(jpaTm());
     loadInitialData();
 
     // Two real registrars have been created by loadInitialData(), named "New Registrar" and "The
@@ -169,7 +165,7 @@ public class RdePipelineTest {
 
   @AfterEach
   void afterEach() {
-    setTmForTest(originalTm);
+    TransactionManagerFactory.removeTmOverrideForTest();
   }
 
   @Test
--- a/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java
+++ b/core/src/test/java/google/registry/persistence/transaction/JpaTransactionManagerImplTest.java
@@ -17,7 +17,7 @@ package google.registry.persistence.transaction;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.truth.Truth.assertThat;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
-import static google.registry.testing.DatabaseHelper.assertDetached;
+import static google.registry.testing.DatabaseHelper.assertDetachedFromEntityManager;
 import static google.registry.testing.TestDataHelper.fileClassPath;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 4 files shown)


sarahcaseybot (dismissed):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 358 at r1](https://reviewable.io/reviews/google/nomulus/1246#-Mf-OgGDABT8tTTxb-yf:-Mf-OgGDABT8tTTxb-yg:b75b23o) ([raw file](https://github.com/google/nomulus/blob/6de2c416c3d8ebfd7cfa7f14162b8762f61d1e63/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L358)):*
> ```Java
> 
>   private static <T> T assertDetachedIfJpa(T entity) {
>     if (!tm().isOfy()) {
> ```

Why is this check needed in both places? Can't you just always call DatabaseHelper.assertDetachedFromEntityManager from here since the tm will be checked there?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 358 at r1](https://reviewable.io/reviews/google/nomulus/1246#-Mf-OgGDABT8tTTxb-yf:-Mf-PVPO48cXgp_oR5qO:bsbts52) ([raw file](https://github.com/google/nomulus/blob/6de2c416c3d8ebfd7cfa7f14162b8762f61d1e63/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L358)):*

Why is this check needed in both places? Can't you just always call DatabaseHelper.assertDetachedFromEntityManager from here since the tm will be checked there?
</blockquote></details>

I think you may be mis-reading the code; the check does not exist in both places (we cannot, in general, guarantee that the check will always work when in JPA mode)

jianglai (changes requested):
Where exactly is the code that ensures that ofy is always initiated when `tm()` is called?

> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246#-:-Mf0Zxg67Q083V8R7zcS:b-aruerh)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[core/src/test/java/google/registry/beam/rde/RdePipelineTest.java, line 169 at r1](https://reviewable.io/reviews/google/nomulus/1246#-Mf0ZP1f921weGELXzbH:-Mf0ZP1g4htlX8rROKPQ:b-8m6rj4) ([raw file](https://github.com/google/nomulus/blob/6de2c416c3d8ebfd7cfa7f14162b8762f61d1e63/core/src/test/java/google/registry/beam/rde/RdePipelineTest.java#L169)):*
> ```
> TransactionManagerFactory.removeTmOverrideForTest();
> ```
Let's be consistent with the call to `setTmForTest`. Either static import both methods, or import the class instead.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/test/java/google/registry/beam/rde/RdePipelineTest.java, line 169 at r1](https://reviewable.io/reviews/google/nomulus/1246#-Mf0ZP1f921weGELXzbH:-Mf0aBh5Av1PlQB1MfvE:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6de2c416c3d8ebfd7cfa7f14162b8762f61d1e63/core/src/test/java/google/registry/beam/rde/RdePipelineTest.java#L169)):*

> ```
> TransactionManagerFactory.removeTmOverrideForTest();
> ```
Let's be consistent with the call to `setTmForTest`. Either static import both methods, or import the class instead.
</blockquote></details>

Done.

gbrodman (commented):
In `TransactionManagerFactory` if you call `tm()` without first setting the override, then we inspect the `DatabaseMigrationSchedule`, which is loaded from Datastore. Does that answer your question? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246#-:-Mf0ajcM83pDWUIDmRFC:bk8dnf1)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

jianglai (approved):
> Reviewed 2 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 358 at r1](https://reviewable.io/reviews/google/nomulus/1246#-Mf-OgGDABT8tTTxb-yf:-Mf3QqYO5vYr4GA-QUoV:b-r2gzme) ([raw file](https://github.com/google/nomulus/blob/6de2c416c3d8ebfd7cfa7f14162b8762f61d1e63/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L358)):*

I think you may be mis-reading the code; the check does not exist in both places (we cannot, in general, guarantee that the check will always work when in JPA mode) 
</blockquote></details>

I think Gus is moving the check here to avoid calling `tm()` in `JpaTransactionManagerImplTest`, which would have tried to initiate ofy and failed, right?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1246)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java, line 358 at r1](https://reviewable.io/reviews/google/nomulus/1246#-Mf-OgGDABT8tTTxb-yf:-Mf3RqaDDKfnlgv22bLd:b2nhtan) ([raw file](https://github.com/google/nomulus/blob/6de2c416c3d8ebfd7cfa7f14162b8762f61d1e63/core/src/test/java/google/registry/persistence/transaction/QueryComposerTest.java#L358)):*

I think Gus is moving the check here to avoid calling `tm()` in `JpaTransactionManagerImplTest`, which would have tried to initiate ofy and failed, right?
</blockquote></details>

Yup

PR #1249, opened 2021-07-20
Write RDE files and advance cursors in Beam pipeline

jianglai:
This PR re-implements most of the logic in the RdeStagingReducer, with
the exception of the last enqueue operations, due to the fact that the
task queue API is not available outside of App Engine SDK. This part
will come in a separate PR.

Another deviation from the reducer is that we forwent the lock -- it is
difficult do it across different beam transforms. Instead we write each
report to a different folder according to its unique beam job name. When
enqueueing the publish tasks we will then pass the folder prefix as a
URL parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1249)


weiminyu (changes requested):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1249)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/1249#-Mf4mCmH7zgtTNHnPMxK:-Mf4mCmH7zgtTNHnPMxL:b-i7afhf) ([raw file](https://github.com/google/nomulus/blob/c5c28969bcd08506066b34db89522eac5c0acddb/core/src/main/java/google/registry/beam/rde/RdeIO.java#L122)):*
> ```
> Iterable<DepositFragment>
> ```
This is the entire file of a TLD right?  There are some references that seem to suggest that size could be a problem although I'm not sure they apply here.

KeyCommitTooLargeException (https://cloud.google.com/dataflow/docs/guides/common-errors#key-commit-too-large-exception)

and below which mentions a 2G limit:

https://stackoverflow.com/questions/55656029/beam-pipeline-does-not-produce-any-output-after-groupbykey-with-windowing-and-i

I guess running a test pipeline that generates more than 2GB per key would give us the answer.
___
*[core/src/main/java/google/registry/beam/rde/RdePipelineOptions.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/1249#-Mf4ZgTMCndtcW-Attb_:-Mf4ZgTN2SGY5aBjJ7zO:b-revcaz) ([raw file](https://github.com/google/nomulus/blob/c5c28969bcd08506066b34db89522eac5c0acddb/core/src/main/java/google/registry/beam/rde/RdePipelineOptions.java#L42)):*
> ```Java
> 
>   void setStagingKey(String value);
> }
> ```

Off topic: you may want to set a default isolation level that is less strict than SERIALIZABLE. See b/193565763 and go/dr-tps-limit

You can override the getIsolationOverride method in this interface and provide a default value using Default.Enum annotation.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1249)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 122 at r1](https://reviewable.io/reviews/google/nomulus/1249#-Mf4mCmH7zgtTNHnPMxK:-MfcZhjS33wJx-7e1aEq:bf7mthx) ([raw file](https://github.com/google/nomulus/blob/c5c28969bcd08506066b34db89522eac5c0acddb/core/src/main/java/google/registry/beam/rde/RdeIO.java#L122)):*

> ```
> Iterable<DepositFragment>
> ```
This is the entire file of a TLD right?  There are some references that seem to suggest that size could be a problem although I'm not sure they apply here.

KeyCommitTooLargeException (https://cloud.google.com/dataflow/docs/guides/common-errors#key-commit-too-large-exception)

and below which mentions a 2G limit:

https://stackoverflow.com/questions/55656029/beam-pipeline-does-not-produce-any-output-after-groupbykey-with-windowing-and-i

I guess running a test pipeline that generates more than 2GB per key would give us the answer.


</blockquote></details>

Both of the links referenced here were discussing streaming mode. I think in our case the beam platform *should* be  smart enough to not load everything in the iterable into memory on the same machine at the same time (even with streaming mode it should be able to do that). Each deposit fragment is just one entity so it should have no problem fitting into the memory. Also I checked the encrypted file deposited on GCS, and the RDE files are around 400MB, which should mean that even if the whole iterable is loaded into memory it should still be fine.

We will certainly run a test with actual data to make it is working properly.
___
*[core/src/main/java/google/registry/beam/rde/RdePipelineOptions.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/1249#-Mf4ZgTMCndtcW-Attb_:-MfegdqpAHDiOgCQcLr-:bbekykq) ([raw file](https://github.com/google/nomulus/blob/c5c28969bcd08506066b34db89522eac5c0acddb/core/src/main/java/google/registry/beam/rde/RdePipelineOptions.java#L42)):*

Off topic: you may want to set a default isolation level that is less strict than SERIALIZABLE. See b/193565763 and go/dr-tps-limit

You can override the getIsolationOverride method in this interface and provide a default value using Default.Enum annotation.
</blockquote></details>

Done. However it'll make all transactions on beam operate at this level. Would this be a problem for write operations (like updating the cursor)?

jianglai:
Ping? PTAL.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1249)* status: 1 of 11 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/beam/rde/RdePipelineOptions.java, line 42 at r1](https://reviewable.io/reviews/google/nomulus/1249#-Mf4ZgTMCndtcW-Attb_:-Mfs7gh-8cuzXOzASmhJ:b-6orj1g) ([raw file](https://github.com/google/nomulus/blob/c5c28969bcd08506066b34db89522eac5c0acddb/core/src/main/java/google/registry/beam/rde/RdePipelineOptions.java#L42)):*

Done. However it'll make all transactions on beam operate at this level. Would this be a problem for write operations (like updating the cursor)?
</blockquote></details>

The difference between SERIALIZABLE isolation and less strict ones is that it throws when T1 commits in the following scenario:

T1: Read A
T2: UPDATE A; COMMIT
T1: COMMIT 

In the RDE case we probably want to ignore the exception anyway since the cursor value written by T2 is almost guaranteed to be the new value T1 is going to commit .

PR #1253, opened 2021-07-21
Use raw EntityManager to load during beforeSqlSave

gbrodman:
If we use the transaction manager methods, JpaTransactionManagerImpl
will attempt to detach the EppResource in question that we're loading --
this fails because that entity has been saved in the same transaction
already. We don't need detaching during these methods (it's just for
resource population) so we can use the raw loads to get around it.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1253)

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -223,7 +223,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
           .toSqlEntity()
           .ifPresent(
               sqlEntity -> {
-                ReplaySpecializer.beforeSqlSave(sqlEntity);
+                sqlEntity.beforeSqlSaveOnReplay();
                 jpaTm().put(sqlEntity);
               });
     } else {
--- a/core/src/main/java/google/registry/model/contact/ContactHistory.java
+++ b/core/src/main/java/google/registry/model/contact/ContactHistory.java
@@ -136,9 +136,11 @@ public class ContactHistory extends HistoryEntry implements SqlEntity {
   }
 
   // Used to fill out the contactBase field during asynchronous replay
-  public static void beforeSqlSave(ContactHistory contactHistory) {
-    contactHistory.contactBase =
-        jpaTm().loadByKey(VKey.createSql(ContactResource.class, contactHistory.getContactRepoId()));
+  @Override
+  public void beforeSqlSaveOnReplay() {
+    if (contactBase == null) {
+      contactBase = jpaTm().getEntityManager().find(ContactResource.class, getContactRepoId());
+    }
   }
 
   /** Class to represent the composite primary key of {@link ContactHistory} entity. */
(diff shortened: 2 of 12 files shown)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1253#-:-MfEzLr-8B9j1i6mRow4:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved

CydeWeys (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1253#-:-MfF4xov3jmDWI4tYSoP:bnfp4nl)* status: 0 of 9 files reviewed, all discussions resolved

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1253)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 227 at r3](https://reviewable.io/reviews/google/nomulus/1253#-MfZ7XRE1eSWL2XyOhnj:-MfZ7XRE1eSWL2XyOhnk:b-3zeltu) ([raw file](https://github.com/google/nomulus/blob/f03b5f6e39ce8afca0ae80b22e76badd27da7484/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L227)):*
> ```Java
>           .ifPresent(
>               sqlEntity -> {
>                 if (sqlEntity instanceof ReplaySpecializedSqlEntity) {
> ```

better way to do this: add `beforeSqlSave()` to `SqlEntity` with a no-op base implementation, override it in the History methods with an actual implementation, then you simply call it normally here without using this inelegant monomorphic invocation.  As a side benefit, `ReplaySpecializedSqlEntity` is no longer necessary at all.

Might want to rename `beforeSqlSave()` too, as that method name isn't exactly accurate -- it's specifically about async replay, right?  So how about something like `beforeSqlReplay()` or similar?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1253)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 227 at r3](https://reviewable.io/reviews/google/nomulus/1253#-MfZ7XRE1eSWL2XyOhnj:-MfZCs5lC3Gaz8nQlNQH:b6zmsy6) ([raw file](https://github.com/google/nomulus/blob/f03b5f6e39ce8afca0ae80b22e76badd27da7484/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L227)):*

better way to do this: add `beforeSqlSave()` to `SqlEntity` with a no-op base implementation, override it in the History methods with an actual implementation, then you simply call it normally here without using this inelegant monomorphic invocation.  As a side benefit, `ReplaySpecializedSqlEntity` is no longer necessary at all.

Might want to rename `beforeSqlSave()` too, as that method name isn't exactly accurate -- it's specifically about async replay, right?  So how about something like `beforeSqlReplay()` or similar?
</blockquote></details>

Good call.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1253#-:-MfZFdrEDnGMYZsnuTxu:bnfp4nl)* status: 0 of 13 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1256, opened 2021-07-22
Remove PremiumList from Datastore schema

sarahcaseybot:
This PR removes the PremiumList table from the Datastore schema since Nomulus is no longer reading or writing PremiumLists to Datastore. Multiple changes were necessary in order to get this to work including:

- Remove the PremiumListRevision object - These are no longer necessary in Cloud SQL
- Remove the PremiumListEntry object - Replaces with the PremiumEntry object for Cloud SQL
- Moved the definition of PremiumEntry objects into the PremiumList class
- Replace the lastUpdateTime with creationTimestamp in BaseDomainLabelList - Since domain label lists are immutable in Cloud SQL, they will not have update times
- Remove PremiumListKeyConverter - This converter is no longer necessary now that we are using list name to reference a list in the Registry object instead of using a Key reference 
- Modify CreateOrUpdatePremiumListCommand to extend ConfirmingCommand instead of MutatingCommand - this is necessary since MutatingCommand relies on the transactionManager which will continue to use Datastore until database cutover for Nomulus occurs in production

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1256)


gbrodman (changes requested):
It seems like this change is doing more than just removing PremiumList from the Datastore schema (or at least it's more complex than just a simple delete), can you elaborate on that + the reasons for it in the commit message please? 

> Reviewed 36 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256#-:-Mfc_zc19mUrrtJYgoFi:b-t6k045)* status: all files reviewed, 8 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 125 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfceXR-EOojZMpyEy5u:-MfceXR-EOojZMpyEy5v:b1lfnt9) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L125)):*
> ```Java
>    */
>   @javax.persistence.Entity(name = "PremiumEntry")
>   public static class PremiumEntry extends DomainLabelEntry<BigDecimal, PremiumList.PremiumEntry>
> ```

We're losing information here; why are we moving from Money to BigDecimal? 

At the very least, the string formatting shouldn't lose the currency (I saw that it changed in at least one of the tests) 
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 361 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgAf39dPctxTuPJJv:-MfcgAf39dPctxTuPJJw:b-2u84k4) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L361)):*
> ```Java
>    */
>   @PostPersist
>   void postPersist() {
> ```

The premise here is that this logic is migrated to the PremiumListDao, right? It seems dangerous to require that the DAO be used to persist this, given that the base structure for the standard types of objects in the codebase is that you can just persist an object directly without having to use a DAO. 

Theoretically, the end goal is that we remove all the DAOs and that all objects are treated the same once we're in SQL, right? 
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 161 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgaWZ8h92OE5Xdw3U:-MfcgaWZ8h92OE5Xdw3V:b-om3qh4) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L161)):*
> ```Java
>     jpaTm().transact(() -> jpaTm().insert(premiumList));
>     premiumListCache.invalidate(premiumList.getName());
>     jpaTm()
> ```

see other comment about handing saving / deletion of the entries in the object itself vs in the DAO 

If this class functions as a cache that's useful too, but it probably shouldn't be doing two things? 
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 271 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgnFhEYfsf9mTZnGR:-MfcgnFhEYfsf9mTZnGS:b-dobvpo) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L271)):*
> ```Java
>             premiumEntry ->
>                 new PremiumEntry.Builder()
>                     .setPrice(premiumEntry.getValue())
> ```

We're losing the currency info here, and I dont' think we need to lose it 
___
*[core/src/main/java/google/registry/tools/CreateOrUpdatePremiumListCommand.java, line 29 at r1](https://reviewable.io/reviews/google/nomulus/1256#-Mfc_r5EAAfNw1FxLJNT:-Mfc_r5EAAfNw1FxLJNU:b-auvlyj) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/tools/CreateOrUpdatePremiumListCommand.java#L29)):*
> ```Java
>  * lists.
>  */
> abstract class CreateOrUpdatePremiumListCommand extends ConfirmingCommand {
> ```

Likely needs to implement CommandWithRemoteApi 
___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfchGIS6pJc658bM-88:-MfchGIS6pJc658bM-89:bhg84xs) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L49)):*
> ```Java
>   void before() {
>     // createTld() overwrites the premium list, so call it first.
>     DatabaseHelper.persistNewRegistrar("TheRegistrar");
> ```

AppEngineExtension should create these automatically; is it not? 
___
*[core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfchYKd9ND11nRQl9RB:-MfchYKd9ND11nRQl9RC:b-bb3e0z) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java#L94)):*
> ```Java
>               assertThat(persistedList.getLabelsToPrices()).containsExactlyEntriesIn(TEST_PRICES);
>               assertThat(persistedList.getCreationTime()).isEqualTo(fakeClock.nowUtc());
>             });
> ```

What's up with this removal? The field `creationTime` still exists right? 
___
*[core/WEB-INF/appengine-generated/local_db.bin, line 0 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcdbCOD3XHLBugG5-I:-MfcdbCPD52L8iQOqHI_:b-y7oee7) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/WEB-INF/appengine-generated/local_db.bin#L0)):*
Accidental add?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 21 of 41 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, @sarahcaseybot, and @weiminyu)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 361 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgAf39dPctxTuPJJv:-Mfd3cypEGR_BNe1-DVQ:b1efo2n) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L361)):*

The premise here is that this logic is migrated to the PremiumListDao, right? It seems dangerous to require that the DAO be used to persist this, given that the base structure for the standard types of objects in the codebase is that you can just persist an object directly without having to use a DAO. 

Theoretically, the end goal is that we remove all the DAOs and that all objects are treated the same once we're in SQL, right? 
</blockquote></details>

I'm not sure what you are suggesting here? Are you recommending that I get rid of the DAO in this PR and move all of its ,methods to this class?
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 271 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgnFhEYfsf9mTZnGR:-Mfd49M8BDn7gMYmwu_E:b-qgt9g7) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L271)):*

We're losing the currency info here, and I dont' think we need to lose it 
</blockquote></details>

It is not needed since the currency is a field of the PremiumList object
___
*[core/src/main/java/google/registry/tools/CreateOrUpdatePremiumListCommand.java, line 29 at r1](https://reviewable.io/reviews/google/nomulus/1256#-Mfc_r5EAAfNw1FxLJNT:-MfdBeL3-QtZSJfnbJbr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/tools/CreateOrUpdatePremiumListCommand.java#L29)):*

Likely needs to implement CommandWithRemoteApi 
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfchGIS6pJc658bM-88:-MfdBqNq1M2eG8tSiYz7:b1ww1an) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L49)):*

AppEngineExtension should create these automatically; is it not? 
</blockquote></details>

It does not. I think it is because I switched the extension to use the .withCloudSql() configuration
___
*[core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java, line 94 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfchYKd9ND11nRQl9RB:-Mfda5589OPfxtKoNm6f:bujr60a) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java#L94)):*

What's up with this removal? The field `creationTime` still exists right? 
</blockquote></details>

The lastUpdateTime is the new creation time field in Cloud SQL, I changed the name of the field and its getter/setters and added these lines back to the tests.
___
*[core/WEB-INF/appengine-generated/local_db.bin, line  at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcdbCOD3XHLBugG5-I:-MfdoShSCI5_ZPy6SyWN:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/WEB-INF/appengine-generated/local_db.bin#L)):*

Accidental add? 
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 21 of 41 files reviewed, 8 unresolved discussions (waiting on @gbrodman, @sarahcaseybot, and @weiminyu)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 361 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgAf39dPctxTuPJJv:-MfnpBX_Aw8mkTAfV2H7:b-5kveoi) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L361)):*
> Theoretically, the end goal is that we remove all the DAOs and that all objects are treated the same once we're in SQL, right?

You sure about that?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 21 of 41 files reviewed, 8 unresolved discussions (waiting on @gbrodman, @sarahcaseybot, and @weiminyu)

___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfchGIS6pJc658bM-88:-MfnpTXREK-3XG1iFC-x:bk26gnn) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L49)):*

It does not. I think it is because I switched the extension to use the .withCloudSql() configuration
</blockquote></details>

We should probably fix that then so that the registrars are created in Cloud SQL when only `.withCloudSql()` is specified.  We're gonna need them, otherwise we're going to have to create them from scratch in every individual test class as you're doing here which is not scalable.

That should probably be a separate precursor PR to this one.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 21 of 41 files reviewed, 8 unresolved discussions (waiting on @gbrodman, @sarahcaseybot, and @weiminyu)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 361 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgAf39dPctxTuPJJv:-MfnpsxDCQ9z4YXqacIs:b-4czk3b) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L361)):*

> Theoretically, the end goal is that we remove all the DAOs and that all objects are treated the same once we're in SQL, right?

You sure about that?
</blockquote></details>

To expand on this, a premium list is a lot more complicated than a single row in a database; I'm not seeing a great way to handle it on its own as if it were.  There's too much logic involved in persisting them that would have to be shoe-horned into the premium list entity classes themselves and various SQL hooks, which is awkward.  A DAO is the more straight-forward way to deal with it.  It's always been this way in Datastore too -- you didn't persist the PremiumList entity directly; you used a DAO method to do so (even though it wasn't named that at the time).

weiminyu (commented):
Agree with Gus above. Would prefer to see a PR that just delete classes.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256#-:-MfsAUnt5wQ9lyoCuqZA:b-6mwmil)* status: 21 of 41 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

gbrodman (changes requested):
> Reviewed 3 of 20 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 24 of 41 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 125 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfceXR-EOojZMpyEy5u:-MfsvXzn-gnF8-r0TUGP:b9obmew) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L125)):*

This is a change that was already done, I just moved the PremiumEntry from a separate file into this one. https://sourcegraph.com/github.com/google/nomulus/-/blob/core/src/main/java/google/registry/schema/tld/PremiumEntry.java?subtree=true

We made this decision to switch from Money to BigDecimal because we decided it was unnecessary to save the currency in each entry since there is only one currency for each list. Instead the currency is saved in the PremiumList object. Also, see go/r3pr/1221
</blockquote></details>

That's fair but we should probably should include the currency in the string representation of the premium lists somewhere right? I'm looking at the change made in the test data of GetPremiumListCommandTest for instance. 
___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 361 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfcgAf39dPctxTuPJJv:-Mfsvu3h96f9fjoJZMmg:b-w59h78) ([raw file](https://github.com/google/nomulus/blob/58e413af89c31ffa100d4845127cb93efe25d89f/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L361)):*

To expand on this, a premium list is a lot more complicated than a single row in a database; I'm not seeing a great way to handle it on its own as if it were.  There's too much logic involved in persisting them that would have to be shoe-horned into the premium list entity classes themselves and various SQL hooks, which is awkward.  A DAO is the more straight-forward way to deal with it.  It's always been this way in Datastore too -- you didn't persist the PremiumList entity directly; you used a DAO method to do so (even though it wasn't named that at the time).
</blockquote></details>

Fair enough; I'm wondering if there should be some guardrails in place so that people don't simply load/save one of them without using the DAO but that's probably a discussion for another time. 
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 168 at r2](https://reviewable.io/reviews/google/nomulus/1256#-MfsvFcA-E35MS5uflw-:-MfsvFcA-E35MS5uflw0:b-r8pj7n) ([raw file](https://github.com/google/nomulus/blob/b5cbb9c8cbdf9406ecaa8b2d94db0bc20903c737/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L168)):*
> ```Java
>                     PremiumListDao.getLatestRevision(premiumList.getName());
>                 premiumList.getLabelsToPrices().entrySet().stream()
>                     .forEach(
> ```

I don't think it matters yet due to the way it's implemented, but rather than calling `.insert` a bunch of times, create all the entries then call a bulk insert at once. It could be significantly faster on the DB level.
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 196 at r2](https://reviewable.io/reviews/google/nomulus/1256#-MfsuwGg64ugZZeL50_e:-MfsuwGg64ugZZeL50_f:bi65ecj) ([raw file](https://github.com/google/nomulus/blob/b5cbb9c8cbdf9406ecaa8b2d94db0bc20903c737/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L196)):*
> ```Java
>                         .getResultList();
> 
>                 entries.forEach(jpaTm()::delete);
> ```

This is inefficient; as long as we're running a raw sql query just run `DELETE FROM PremiumEntry WHERE revisionId = :revisionId` 
___
*[core/src/main/java/google/registry/schema/tld/PremiumListUtils.java, line 38 at r2](https://reviewable.io/reviews/google/nomulus/1256#-MfsujWnAt5RdchidGBt:-MfsujWnAt5RdchidGBu:bu0z2sa) ([raw file](https://github.com/google/nomulus/blob/b5cbb9c8cbdf9406ecaa8b2d94db0bc20903c737/core/src/main/java/google/registry/schema/tld/PremiumListUtils.java#L38)):*
> ```Java
>     ImmutableMap<String, PremiumEntry> prices =
>         new PremiumList.Builder().setName(name).build().parse(inputData);
>     checkArgument(inputData.size() > 0, "Input cannot be empty");
> ```

It seems like we're losing the protections that we had before about there being only one currency; what's the reason for that? 
___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1256#-Mfsu8RsClmJOufREHJW:-Mfsu8Rt2kFowDqNAm19:bo8jtyi) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L54)):*
> ```Java
>     PremiumList pl =
>         persistPremiumList(
>             "tld", "lol,USD 999", "rich,USD 1999", "icann,JPY 100", "johnny-be-goode,USD 20.50");
> ```

shouldn't this be failing if we're trying to create a premium list with mixed currencies? 
___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 111 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfsuRs153fHSOSFIf1v:-MfsuRs153fHSOSFIf1w:blfv3kp) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L111)):*
> ```Java
>             () ->
>                 new PremiumEntry.Builder()
>                     .setPrice(Money.parse("USD 399").getAmount())
> ```

For these PremiumEntry creations, it seems like overkill to have to specify USD every time, right? Like, they should just be "399 whatever-the-list-currency-is", no need to parse a Money

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 8 of 59 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/registry/label/PremiumList.java, line 125 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfceXR-EOojZMpyEy5u:-MgMhlKrEmu7j7wfeW46:b-n62bks) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/main/java/google/registry/model/registry/label/PremiumList.java#L125)):*

That's fair but we should probably should include the currency in the string representation of the premium lists somewhere right? I'm looking at the change made in the test data of GetPremiumListCommandTest for instance. 
</blockquote></details>

added a toString(CurrencyUnit) method to PremiumEntry to be used by the getPremiumListCommand
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 168 at r2](https://reviewable.io/reviews/google/nomulus/1256#-MfsvFcA-E35MS5uflw-:-MgMpGX_7DJWzwRaZfbY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b5cbb9c8cbdf9406ecaa8b2d94db0bc20903c737/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L168)):*

I don't think it matters yet due to the way it's implemented, but rather than calling `.insert` a bunch of times, create all the entries then call a bulk insert at once. It could be significantly faster on the DB level.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/schema/tld/PremiumListDao.java, line 196 at r2](https://reviewable.io/reviews/google/nomulus/1256#-MfsuwGg64ugZZeL50_e:-MgHbg9CAxiykP8PP-L4:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b5cbb9c8cbdf9406ecaa8b2d94db0bc20903c737/core/src/main/java/google/registry/schema/tld/PremiumListDao.java#L196)):*

This is inefficient; as long as we're running a raw sql query just run `DELETE FROM PremiumEntry WHERE revisionId = :revisionId` 
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/schema/tld/PremiumListUtils.java, line 38 at r2](https://reviewable.io/reviews/google/nomulus/1256#-MfsujWnAt5RdchidGBt:-MgC9BOJ7kv5AwgL4Qzq:b-8uktvt) ([raw file](https://github.com/google/nomulus/blob/b5cbb9c8cbdf9406ecaa8b2d94db0bc20903c737/core/src/main/java/google/registry/schema/tld/PremiumListUtils.java#L38)):*

It seems like we're losing the protections that we had before about there being only one currency; what's the reason for that? 
</blockquote></details>

There's only one currency for each list as specified in the CurrencyUnit field of the PremiumList object. I modified the code a bit so that the inputed data may or may not include the currency unit. If it does, the code checks to make sure it is using the correct currency unit for the list. This required some modification to the commands to now explicitly ask for the currency unit when creating a new PremiumList.
___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 49 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfchGIS6pJc658bM-88:-MgC1b9J4GRLr_aoL9iD:b-en7x14) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L49)):*

We should probably fix that then so that the registrars are created in Cloud SQL when only `.withCloudSql()` is specified.  We're gonna need them, otherwise we're going to have to create them from scratch in every individual test class as you're doing here which is not scalable.

That should probably be a separate precursor PR to this one.
</blockquote></details>

done in go/r3pr/1266
___
*[core/src/test/java/google/registry/model/registry/label/PremiumListTest.java, line 111 at r1](https://reviewable.io/reviews/google/nomulus/1256#-MfsuRs153fHSOSFIf1v:-MgC2Yhz9rdZRlRkvaL6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7dfa17ec0abb6915417309688bade81566a3b3d9/core/src/test/java/google/registry/model/registry/label/PremiumListTest.java#L111)):*

For these PremiumEntry creations, it seems like overkill to have to specify USD every time, right? Like, they should just be "399 whatever-the-list-currency-is", no need to parse a Money 
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 11 of 20 files at r2, 32 of 38 files at r3, 8 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/server/UpdatePremiumListAction.java, line 52 at r4](https://reviewable.io/reviews/google/nomulus/1256#-MgR7rndAfBYbtgeP5yC:-MgR7rndAfBYbtgeP5yD:bm57gqw) ([raw file](https://github.com/google/nomulus/blob/d6bcb794ea2852ce2f697cc8fbde60d5eecdd9ab/core/src/main/java/google/registry/tools/server/UpdatePremiumListAction.java#L52)):*
> ```Java
>     Optional<PremiumList> existingList = PremiumListDao.getLatestRevision(name);
>     checkArgument(
>         PremiumListDao.getLatestRevision(name).isPresent(),
> ```

nit, use `existingList` here 
___
*[core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java, line 134 at r4](https://reviewable.io/reviews/google/nomulus/1256#-MgR8bSP6Eb3mUbFc61Z:-MgR8bSP6Eb3mUbFc61_:bz14wdi) ([raw file](https://github.com/google/nomulus/blob/d6bcb794ea2852ce2f697cc8fbde60d5eecdd9ab/core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java#L134)):*
> ```Java
>               assertThat(savedList.getCreationTimestamp()).isEqualTo(fakeClock.nowUtc());
>               assertThat(savedList.getRevisionId()).isGreaterThan(firstRevisionId);
>               assertThat(savedList.getCreationTimestamp()).isEqualTo(fakeClock.nowUtc());
> ```

duplicated check
___
*[core/WEB-INF/appengine-generated/local_db.bin, line 0 at r4](https://reviewable.io/reviews/google/nomulus/1256#-MgR8zQBBoAMFU9BZs5q:-MgR8zQBBoAMFU9BZs5r:b-8h4e7l) ([raw file](https://github.com/google/nomulus/blob/d6bcb794ea2852ce2f697cc8fbde60d5eecdd9ab/core/WEB-INF/appengine-generated/local_db.bin#L0)):*
this got added again? 

if it's generated we can add it to the gitignore file

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: 57 of 60 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java, line 134 at r4](https://reviewable.io/reviews/google/nomulus/1256#-MgR8bSP6Eb3mUbFc61Z:-MgRkS3f0xzLl6qvBmzW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d6bcb794ea2852ce2f697cc8fbde60d5eecdd9ab/core/src/test/java/google/registry/schema/tld/PremiumListDaoTest.java#L134)):*

duplicated check
</blockquote></details>

Done.
___
*[core/WEB-INF/appengine-generated/local_db.bin, line  at r4](https://reviewable.io/reviews/google/nomulus/1256#-MgR8zQBBoAMFU9BZs5q:-MgSADvf0edgspwnt9IR:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d6bcb794ea2852ce2f697cc8fbde60d5eecdd9ab/core/WEB-INF/appengine-generated/local_db.bin#L)):*

this got added again? 

if it's generated we can add it to the gitignore file 
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1256)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #1268, opened 2021-08-04
Use one SQL transaction per Datastore transaction in replay to SQL

gbrodman:
There was a subtle issue that we encountered in sandbox when using one
transaction per file that was difficult to replicate. Basically,
1. Save a domain with dsData
2. Save the domain without dsData
3. Save the domain with the same dsData as step 1
4. Delete literally any object

If one performs steps 2-4 in the same transaction, Hibernate will throw
an exception (cascade re-saving a cascade-deleted object). Note that
step 4 is in fact necessary to reproduce the issue, yay Hibernate.

We will test this and if one transaction per transaction is too slow,
we'll figure out ways to reduce the number of SQL transactions.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1268)

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -189,7 +189,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
             searchStartTime.toString("yyyy-MM-dd HH"));
       }
       for (BlobInfo file : fileBatch) {
-        jpaTm().transact(() -> processFile(file));
+        processFile(file);
         filesProcessed++;
         if (clock.nowUtc().isAfter(replayTimeoutTime)) {
           return String.format(
@@ -205,10 +205,11 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
       // Load and process the Datastore transactions one at a time
       ImmutableList<ImmutableList<VersionedEntity>> allTransactions =
           CommitLogImports.loadEntitiesByTransaction(input);
-      allTransactions.forEach(this::replayTransaction);
+      allTransactions.forEach(
+          transaction -> jpaTm().transact(() -> replayTransaction(transaction)));
       // if we succeeded, set the last-seen time
       DateTime checkpoint = DateTime.parse(metadata.getName().substring(DIFF_FILE_PREFIX.length()));
-      SqlReplayCheckpoint.set(checkpoint);
+      jpaTm().transact(() -> SqlReplayCheckpoint.set(checkpoint));
       logger.atInfo().log(
           "Replayed %d transactions from commit log file %s with size %d B.",
           allTransactions.size(), metadata.getName(), metadata.getSize());
--- a/core/src/test/java/google/registry/backup/ReplayCommitLogsToSqlActionTest.java
+++ b/core/src/test/java/google/registry/backup/ReplayCommitLogsToSqlActionTest.java
@@ -554,6 +554,51 @@ public class ReplayCommitLogsToSqlActionTest {
             });
   }
 
+  @Test
+  void testReplay_deleteAndResaveCascade_withOtherDeletion_noErrors() throws Exception {
+    createTld("tld");
+    DateTime now = fakeClock.nowUtc();
+    jpaTm().transact(() -> SqlReplayCheckpoint.set(now.minusMinutes(1).minusMillis(1)));
+
+    // Save a domain with a particular dsData in SQL as the base
+    ImmutableSet<DelegationSignerData> dsData =
+        ImmutableSet.of(DelegationSignerData.create(1, 2, 3, new byte[] {4, 5, 6}));
+    DomainBase domainWithDsData =
+        newDomainBase("example.tld").asBuilder().setDsData(dsData).build();
+    jpaTm().transact(() -> jpaTm().put(domainWithDsData));
+
+    // Replay a version of that domain without the dsData
+    Key<CommitLogManifest> manifestKeyOne =
+        CommitLogManifest.createKey(getBucketKey(1), now.minusMinutes(3));
+    DomainBase domainWithoutDsData =
+        domainWithDsData.asBuilder().setDsData(ImmutableSet.of()).build();
+    CommitLogMutation domainWithoutDsDataMutation =
+        ofyTm().transact(() -> CommitLogMutation.create(manifestKeyOne, domainWithoutDsData));
+
+    // Create an object (any object) to delete via replay to trigger Hibernate flush events
+    TestObject testObject = TestObject.create("foo", "bar");
+    jpaTm().transact(() -> jpaTm().put(testObject));
+
+    // Replay the original domain, with the original dsData
+    Key<CommitLogManifest> manifestKeyTwo =
+        CommitLogManifest.createKey(getBucketKey(1), now.minusMinutes(2));
+    CommitLogMutation domainWithOriginalDsDataMutation =
+        ofyTm().transact(() -> CommitLogMutation.create(manifestKeyTwo, domainWithDsData));
+
+    // If we try to perform all the events in one transaction (cascade-removal of the dsData,
+    // cascade-adding the dsData back in, and deleting any other random object), Hibernate will
+    // throw an exception
+    saveDiffFile(
+        gcsUtils,
+        createCheckpoint(now.minusMinutes(1)),
+        CommitLogManifest.create(
+            getBucketKey(1), now.minusMinutes(3), ImmutableSet.of(Key.create(testObject))),
+        domainWithoutDsDataMutation,
+        CommitLogManifest.create(getBucketKey(1), now.minusMinutes(2), ImmutableSet.of()),
+        domainWithOriginalDsDataMutation);
+    runAndAssertSuccess(now.minusMinutes(1), 1);
+  }
+
   private void runAndAssertSuccess(DateTime expectedCheckpointTime, int numFiles) {
     action.run();
     assertThat(response.getStatus()).isEqualTo(SC_OK);


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 211 at r1](https://reviewable.io/reviews/google/nomulus/1268#-MgMGCRNAQHqE-iwqWg1:-MgMGCRNAQHqE-iwqWg2:bhnqcxo) ([raw file](https://github.com/google/nomulus/blob/2608a276c12ca2f4a708d2a9500ffec26f302d99/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L211)):*
> ```Java
>       // if we succeeded, set the last-seen time
>       DateTime checkpoint = DateTime.parse(metadata.getName().substring(DIFF_FILE_PREFIX.length()));
>       jpaTm().transact(() -> SqlReplayCheckpoint.set(checkpoint));
> ```

This doesn't work.  This needs to be re-saved inside the transaction below.  Yes, that's going to involve a lot of churn, but there's no other way to do it and be safe.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 211 at r1](https://reviewable.io/reviews/google/nomulus/1268#-MgMGCRNAQHqE-iwqWg1:-MgMaOOnBw1M1Xftdc7m:b9pql5t) ([raw file](https://github.com/google/nomulus/blob/2608a276c12ca2f4a708d2a9500ffec26f302d99/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L211)):*

This doesn't work.  This needs to be re-saved inside the transaction below.  Yes, that's going to involve a lot of churn, but there's no other way to do it and be safe.
</blockquote></details>

Fair point, done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1268#-MgR9TdzB_qnNusaK9AB:-MgR9TdzB_qnNusaK9AC:bk0x5cn) ([raw file](https://github.com/google/nomulus/blob/493b8f1d33c916c92a7bd1be8a84ccfbde55e263/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L215)):*
> ```Java
>                   // if we succeeded, set the last-seen time
>                   DateTime checkpoint =
>                       DateTime.parse(metadata.getName().substring(DIFF_FILE_PREFIX.length()));
> ```

Wait, the checkpoint isn't the specific transaction within the file, it's just the file overall?  How does that work?  Won't this be doing a lot of noops then?

What happens if a file is partially committed, then the task fails, and re-runs 3 minutes later?  How does it successfully resume from the middle of the file?  Won't the earlier transactions that have already been replayed fail?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1268#-MgR9TdzB_qnNusaK9AB:-MgRABgdDAsG--iW7XvV:b-6ltqmo) ([raw file](https://github.com/google/nomulus/blob/493b8f1d33c916c92a7bd1be8a84ccfbde55e263/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L215)):*

Wait, the checkpoint isn't the specific transaction within the file, it's just the file overall?  How does that work?  Won't this be doing a lot of noops then?

What happens if a file is partially committed, then the task fails, and re-runs 3 minutes later?  How does it successfully resume from the middle of the file?  Won't the earlier transactions that have already been replayed fail?
</blockquote></details>

If the checkpoint is on a per-file basis, it wouldn't make sense to keep updating it after every transaction, since most of those are no-ops.  You'd only update it once the entire file has committed and you're moving onto the next one.  And then you also need a way to track progress within the transactions in the file, which doesn't appear to be present.  You're using the export file datetime as the checkpoint, but we also need the transaction times (or indexes) of the transactions within those files, no?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1268#-MgR9TdzB_qnNusaK9AB:-MgRBOcPAzFYDN7yxUHs:bo9cvcc) ([raw file](https://github.com/google/nomulus/blob/493b8f1d33c916c92a7bd1be8a84ccfbde55e263/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L215)):*

If the checkpoint is on a per-file basis, it wouldn't make sense to keep updating it after every transaction, since most of those are no-ops.  You'd only update it once the entire file has committed and you're moving onto the next one.  And then you also need a way to track progress within the transactions in the file, which doesn't appear to be present.  You're using the export file datetime as the checkpoint, but we also need the transaction times (or indexes) of the transactions within those files, no?
</blockquote></details>

That's why the checkpoint was originally per-file, and it probably still could be per-file. We don't anticipate that replaying the same transactions (from the first half of the file before it failed) would cause any issues because the only situation that would cause issues would involve hard-deleted entities that other entities had foreign keys on -- we discussed this if you recall, and basically only poll messages are hard-deleted.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1268#-MgR9TdzB_qnNusaK9AB:-MgRECss9uzxqz3sr1Ry:b-jsdu2m) ([raw file](https://github.com/google/nomulus/blob/493b8f1d33c916c92a7bd1be8a84ccfbde55e263/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L215)):*

That's why the checkpoint was originally per-file, and it probably still could be per-file. We don't anticipate that replaying the same transactions (from the first half of the file before it failed) would cause any issues because the only situation that would cause issues would involve hard-deleted entities that other entities had foreign keys on -- we discussed this if you recall, and basically only poll messages are hard-deleted. 
</blockquote></details>

Ok, so if we're not checkpointing at the file level then, and replay of transactions that have already been successfully committed is OK, then the checkpoint should only be updated at the end of processing the entire file, as it had been previously.

But what about the situation where we have something like "update A with state 1351, update A with state 1240, update A with state 2629", etc., in a single file?  E.g. corresponding to a rapid series of domain update operations?  We're gonna be doing some unnecessary reversions and then re-reversions if we end up replaying most of that file, right?  But the end state is guaranteed to be the same?

Also, separately, I just want to confirm again for safety at what point the eventual consistency from Datastore is converted into strongly ordered transactional consistency with a guaranteed monotonically increasing tx time.  Is that guaranteed true of these commit log export files?  Every single TX within them has a strictly increasing Datastore transaction timestamp?  And there's no situation where there can be transactions in file A that have a later transaction time than any transaction in file A+3?  (Where A+3 is the chronologically next commit log export file.)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1268#-MgR9TdzB_qnNusaK9AB:-MgRVGUA7yHeywSuM2VI:bgkymdu) ([raw file](https://github.com/google/nomulus/blob/493b8f1d33c916c92a7bd1be8a84ccfbde55e263/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L215)):*
> then the checkpoint should only be updated at the end of processing the entire file, as it had been previously.

Done

> We're gonna be doing some unnecessary reversions and then re-reversions if we end up replaying most of that file, right? But the end state is guaranteed to be the same?

Yes to both. Because the objects being replayed are objects and not only diffs, we should be ok. 

> eventual consistency

Take a look at the comment at https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/backup/CommitLogCheckpointStrategy.java#L45-L51

tldr is that any transaction that depends on another transaction will have a later timestamp (mentioned in that file, but not implemented in that file) and that commit logs are bunched together in files such that for each commit log in a file, the file will also contain commit logs that (were written before that commit log && have an earlier timestamp than that commit log) 

I think that should answer your question.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1268)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1273, opened 2021-08-06
Load DatabaseMigrationStateSchedule in a more performant way

CydeWeys:
This performs a direct load-by-key (the most efficient Datastore operation),
rather than attempting to load all entities by type using an ancestor query. The
existing implementation is possibly more error-prone as well, and might be
responsible for the "cross-group transaction need to be explicitly specified"
error we're seeing.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1273)


gbrodman (dismissed):
this is not what we want to do since we're moving this object to SQL

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1273)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1276, opened 2021-08-11
Add tx/s instrumentation to replay action and re-enable it on sandbox

CydeWeys:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1276)

### DIFF
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -58,6 +58,7 @@ import javax.inject.Inject;
 import javax.servlet.http.HttpServletResponse;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
+import org.joda.time.Seconds;
 
 /** Action that replays commit logs to Cloud SQL to keep it up to date. */
 @Action(
@@ -97,7 +98,8 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
 
   @Override
   public void run() {
-    MigrationState state = DatabaseMigrationStateSchedule.getValueAtTime(clock.nowUtc());
+    DateTime startTime = clock.nowUtc();
+    MigrationState state = DatabaseMigrationStateSchedule.getValueAtTime(startTime);
     if (!state.getReplayDirection().equals(ReplayDirection.DATASTORE_TO_SQL)) {
       String message =
           String.format(
@@ -126,7 +128,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
       if (dryRun) {
         resultMessage = executeDryRun();
       } else {
-        resultMessage = replayFiles();
+        resultMessage = replayFiles(startTime);
       }
       response.setStatus(SC_OK);
       response.setPayload(resultMessage);
@@ -160,10 +162,11 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
         fileBatch.stream().limit(10).collect(toImmutableList()));
   }
 
-  private String replayFiles() {
-    DateTime replayTimeoutTime = clock.nowUtc().plus(REPLAY_TIMEOUT_DURATION);
+  private String replayFiles(DateTime startTime) {
+    DateTime replayTimeoutTime = startTime.plus(REPLAY_TIMEOUT_DURATION);
     DateTime searchStartTime = jpaTm().transact(() -> SqlReplayCheckpoint.get().plusMillis(1));
     int filesProcessed = 0;
+    int transactionsProcessed = 0;
     // Starting from one millisecond after the last file we processed, search for and import files
     // one hour at a time until we catch up to the current time or we hit the replay timeout (in
     // which case the next run will pick up from where we leave off).
@@ -171,12 +174,12 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
     // We use hour-long batches because GCS supports filename prefix-based searches.
     while (true) {
       if (isAtOrAfter(clock.nowUtc(), replayTimeoutTime)) {
-        return String.format(
-            "Reached max execution time after replaying %d file(s).", filesProcessed);
+        return createResponseString(
+            "Reached max execution time", startTime, filesProcessed, transactionsProcessed);
       }
       if (isBeforeOrAt(clock.nowUtc(), searchStartTime)) {
-        return String.format(
-            "Caught up to current time after replaying %d file(s).", filesProcessed);
+        return createResponseString(
+            "Caught up to current time", startTime, filesProcessed, transactionsProcessed);
       }
       // Search through the end of the hour
       DateTime searchEndTime =
@@ -189,18 +192,32 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
             searchStartTime.toString("yyyy-MM-dd HH"));
       }
       for (BlobInfo file : fileBatch) {
-        processFile(file);
+        transactionsProcessed += processFile(file);
         filesProcessed++;
         if (clock.nowUtc().isAfter(replayTimeoutTime)) {
-          return String.format(
-              "Reached max execution time after replaying %d file(s).", filesProcessed);
+          return createResponseString(
+              "Reached max execution time", startTime, filesProcessed, transactionsProcessed);
         }
       }
       searchStartTime = searchEndTime.plusMillis(1);
     }
   }
 
-  private void processFile(BlobInfo metadata) {
+  private String createResponseString(
+      String msg, DateTime startTime, int filesProcessed, int transactionsProcessed) {
+    double tps =
+        (double) transactionsProcessed
+            / (double) Seconds.secondsBetween(startTime, clock.nowUtc()).getSeconds();
+    return String.format(
+        "%s after replaying %d file(s) containing %d total transaction(s) (%.2f tx/s).",
+        msg, filesProcessed, transactionsProcessed, tps);
+  }
+
+  /**
+   * Replays the commit logs in the given commit log file and returns the number of transactions
+   * committed.
+   */
+  private int processFile(BlobInfo metadata) {
     try (InputStream input = gcsUtils.openInputStream(metadata.getBlobId())) {
       // Load and process the Datastore transactions one at a time
       ImmutableList<ImmutableList<VersionedEntity>> allTransactions =
@@ -213,6 +230,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
       logger.atInfo().log(
           "Replayed %d transactions from commit log file %s with size %d B.",
           allTransactions.size(), metadata.getName(), metadata.getSize());
+      return allTransactions.size();
     } catch (IOException e) {
       throw new RuntimeException(
           "Errored out while replaying commit log file " + metadata.getName(), e);
--- a/core/src/main/java/google/registry/env/sandbox/default/WEB-INF/cron.xml
+++ b/core/src/main/java/google/registry/env/sandbox/default/WEB-INF/cron.xml
@@ -230,7 +230,7 @@
   </cron>
 
   <cron>
-    <url><![CDATA[/_dr/cron/fanout?queue=replay-commit-logs-to-sql&endpoint=/_dr/task/replayCommitLogsToSql&runInEmpty&dryRun=true]]></url>
+    <url><![CDATA[/_dr/cron/fanout?queue=replay-commit-logs-to-sql&endpoint=/_dr/task/replayCommitLogsToSql&runInEmpty]]></url>
     <description>
       Replays recent commit logs from Datastore to the SQL secondary backend.
     </description>
--- a/core/src/test/java/google/registry/backup/ReplayCommitLogsToSqlActionTest.java
+++ b/core/src/test/java/google/registry/backup/ReplayCommitLogsToSqlActionTest.java
@@ -201,7 +201,7 @@ public class ReplayCommitLogsToSqlActionTest {
         CommitLogMutation.create(manifest2Key, TestObject.create("f")));
     jpaTm().transact(() -> SqlReplayCheckpoint.set(now.minusMinutes(1).minusMillis(1)));
     fakeClock.advanceOneMilli();
-    runAndAssertSuccess(now, 2);
+    runAndAssertSuccess(now, 2, 3);
     assertExpectedIds("previous to keep", "b", "d", "e", "f");
   }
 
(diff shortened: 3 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[-- commits, line 1 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplwl687Q12b9L5KCg:-Mgplwl8DArgzGEVF0XE:bvpttoe) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/-- commits#L1)):*
what is this? 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-Mgplyo53CMtzOZ-w3TU:b-pgv55u) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*
> ```Java
>   @Override
>   public void run() {
>     DateTime startTime = clock.nowUtc();
> ```

I don't see a benefit to starting the timing outside of the actual replay block (`replayFiles`), since we're not using it outside of that.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[-- commits, line 1 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplwl687Q12b9L5KCg:-MgpqgkiB3tH1i-9tXfr:b-8h2l5z) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/-- commits#L1)):*

what is this? 
</blockquote></details>

The commit message?  I dunno, if you look on GitHub, it's not an actual part of the PR.
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-MgpqsIu2ihZTMSqFWis:b-fit5zj) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

I don't see a benefit to starting the timing outside of the actual replay block (`replayFiles`), since we're not using it outside of that. 
</blockquote></details>

We want to measure the execution time of the entire action, not just the execution time of the `replayFiiles()` method.  There is some non-trivial work occurring prior to `replayFiles()` even being called that we need to measure.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[-- commits, line 1 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplwl687Q12b9L5KCg:-MgprO6T76VM0qK8UK4c:b3bjy5m) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/-- commits#L1)):*

The commit message?  I dunno, if you look on GitHub, it's not an actual part of the PR.
</blockquote></details>

Or I should say, it's part of the PR in that it is the commit message manifest or something, but it's not actual file contents of the PR.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[-- commits, line 1 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplwl687Q12b9L5KCg:-MgprUNb-D0B0dnv0fjX:blgalq5) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/-- commits#L1)):*

Or I should say, it's part of the PR in that it is the commit message manifest or something, but it's not actual file contents of the PR.
</blockquote></details>

huh weird 
___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-MgprRLM78iY8ZTvlYbn:b-gg5ibn) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

We want to measure the execution time of the entire action, not just the execution time of the `replayFiiles()` method.  There is some non-trivial work occurring prior to `replayFiles()` even being called that we need to measure.
</blockquote></details>

If the work is trivial, then it seems unnecessary to include it. 

If the work is non-trivial, we wouldn't want to include it in the calculation of number of transactions processed per second, right?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-Mgprccv1QYuyMJnyvZr:b-sydvcf) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

If the work is trivial, then it seems unnecessary to include it. 

If the work is non-trivial, we wouldn't want to include it in the calculation of number of transactions processed per second, right? 
</blockquote></details>

We absolutely do want to include it in the calculation of transactions per second because it affects what the actual transactions per second were.  If the start-up takes 4 minutes for whatever reason, then we're only doing 1 minute of actual work instead of 5 each time this action runs, and that will reduce our transactions to 1/5th of what they'd otherwise be.  We absolutely need to account for this to get any sense of how long it's going to take for things to catch up.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-Mgpt2qFAvLGX1scLGPY:b-a5opl5) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

We absolutely do want to include it in the calculation of transactions per second because it affects what the actual transactions per second were.  If the start-up takes 4 minutes for whatever reason, then we're only doing 1 minute of actual work instead of 5 each time this action runs, and that will reduce our transactions to 1/5th of what they'd otherwise be.  We absolutely need to account for this to get any sense of how long it's going to take for things to catch up.
</blockquote></details>

To phrase it another way, we want to measure wall clock (i.e. real) processing rate.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-MgpsM6_DMWSfzQ4Gqqh:b-k42v09) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

To phrase it another way, we want to measure wall clock (i.e. real) processing rate.
</blockquote></details>

A few comments
1. It is fairly trivial; the total time from the beginning of the action to when the actual replay starts is about one second 
2. If you're including extra one-off work, then transactions per second is not the right name for it. 
3. We're not trying to debug or get information on startup time, but rather only how many transactions are processed and how quickly. One-off startup time is both a known insignificant quantity, and also irrelevant to what we are trying to gain more information about.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-MgpwTz63HWy-1QOd_kJ:b-b8g960) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

A few comments
1. It is fairly trivial; the total time from the beginning of the action to when the actual replay starts is about one second 
2. If you're including extra one-off work, then transactions per second is not the right name for it. 
3. We're not trying to debug or get information on startup time, but rather only how many transactions are processed and how quickly. One-off startup time is both a known insignificant quantity, and also irrelevant to what we are trying to gain more information about. 
</blockquote></details>

What's your preferred name for it then?

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-MgpxIozAxcmH17VY_XW:bhbujmy) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

"Average transactions per second processed over the entire runtime of the action, not some arbitrary subset thereof"?
</blockquote></details>

Why not just move it so that it's actually only transactions-per-second?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-MgqTPCDD4X7ai6GlUxH:b-pauf70) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

Why not just move it so that it's actually only transactions-per-second? 
</blockquote></details>

1.  You're assuming that it's fairly trivial; you don't actually know, and without measuring it, we won't know if it ever degrades.  There are multiple database round-trips involved in the setup here, including a mutating transaction on acquiring the `Lock` object, and the upper bound for how long these kinds of operations can take (and have been observed to take in the past) is quite high, on the order of minutes not seconds.  These preparatory steps are necessary to replay the transactions, and have unbounded potential length, and thus should be included in our metrics.
2. Give me a better name if the name is the sticking point for you.  I'm flexible on the naming; that's not what I care about so much.
3. I _am_ actually trying to measure overall performance of the entire action.  That's what we care about -- overall performance, how long it takes (wall time) to replay a given number of transactions.  That's the only metric we can rely on to make any meaningful extrapolation of how many transactions we can actually replay over a given length of wall time, i.e. how long it will take to catch up given how far we are currently behind, or once we are caught up, what's the maximum transaction rate we can keep up with without falling behind.  Again you are making the assumption that startup time is a known insignificant quantity, which you cannot guarantee, certainly not without even measuring it.
4. It is a common pattern when, measuring how long something takes, to actually measure the entire time that said thing takes, rather than some subset thereof.  E.g. if you ran `time tar -c ___` to benchmark how long it takes to compress a directory using various compression metrics, that will measure the entire execution time of the `tar` program, including any necessary setup like creating the new archive file and writing out its index, along with any necessary teardown like flushing the compressed data to disk.  It wouldn't make sense to exclusively measure the time spent writing out the compressed files within the archive to calculate a "files compressed per second" metric, and ignore the preparatory/teardown time, because the preparatory/teardown time does actually matter and needs to be amortized into the statistics.  Sure, you can calculate other statistics too, but at a minimum you always need how long it took in total -- you're just completely flying blind without that information.
5. I don't understand why you're pushing back so hard against something that seems so seemingly innocuous to me.  if you really want to measure _only_ the processing throughput of the `processFiles()` method for some reason, but ignore all the other time spent within the action, you can feel free to add that on top of this metric after this PR goes in.  I won't argue against more data -- it wouldn't be bad to know all of the following statistics: the total time spent on the action, the amount of time spent finding/loading the commit log files in GCS, and the amount of time spent actually replaying the transactions.  But adding all of those separate data points is a bit more work than merely adding in this one overall time metric which I'm doing here, and I wasn't sure that was all entirely worth it yet, hence in this PR I'm adding the one simple overall metric that at least doesn't exclude any data, mainly, the total amount of time the action takes.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java, line 101 at r1](https://reviewable.io/reviews/google/nomulus/1276#-Mgplyo53CMtzOZ-w3TT:-Mgvdl640Y5cKQYxa_tL:b-x02j73) ([raw file](https://github.com/google/nomulus/blob/bdc1887f0169a44d6efd26115ad13061f92ae691/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java#L101)):*

I don't feel like I'm pushing back particularly hard, at least not any harder than you are or normally do -- business as usual. I just thought we were trying to answer the question of "how do the size / complexity of the transactions affect how long it takes to play them back?" This was based on us already knowing roughly how long the other parts of the action take, by examining the logs. 

This is fine, I think we were just answering different questions. 

You'll want to change the tests as well; I think the messages are causing it to fail. 
</blockquote></details>

Tests are fixed now, PTAL.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1276)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1280, opened 2021-08-12
Add the domain DNS refresh request time field to the DB schema

CydeWeys:
This isn't used yet, but it will eventually be the replacement for the dns-pull
task queue once we get further in the migration.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1280)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1280)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 326 at r1](https://reviewable.io/reviews/google/nomulus/1280#-MgzzGTL2CIs5hahbnF_:-MgzzGTL2CIs5hahbnFa:be04ofh) ([raw file](https://github.com/google/nomulus/blob/68c0d65ad5c1e58180480644f3f796d7b199f3a7/core/src/main/java/google/registry/model/domain/DomainContent.java#L326)):*
> ```
> DateTime dnsRefreshRequestTime
> ```
I don't think this would work with Hibernate. Hibernate will attempt to read this field when loading entities.

There is no way to avoid discrepancy between golden schema and generated schema during a schema change. You may want to leave db-schema.generated as changed in this PR to pass presubmit tests.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1280)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/domain/DomainContent.java, line 326 at r1](https://reviewable.io/reviews/google/nomulus/1280#-MgzzGTL2CIs5hahbnF_:-Mh-Jm9F0LWAYyVKGxgk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/68c0d65ad5c1e58180480644f3f796d7b199f3a7/core/src/main/java/google/registry/model/domain/DomainContent.java#L326)):*

> ```
> DateTime dnsRefreshRequestTime
> ```
I don't think this would work with Hibernate. Hibernate will attempt to read this field when loading entities.

There is no way to avoid discrepancy between golden schema and generated schema during a schema change. You may want to leave db-schema.generated as changed in this PR to pass presubmit tests.
</blockquote></details>

Done.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1280)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1286, opened 2021-08-14
Remove recursive load in DBMSS cache

gbrodman:
This occurs because if we do a standard transaction, the JpaTxnManager
checks to see if we should be doing backups, which involves loading the
migration state schedule (causing the recursion). When starting the
transaction to load the schedule, we should explicitly
transactWithoutBackup so there's no need to check.

This wasn't hit in tests because we previously manually set the
replication to not occur in the JpaTransactionManagerExtension -- we
remove that and related setters.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1286)

### DIFF
--- a/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java
+++ b/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java
@@ -222,7 +222,7 @@ public class DatabaseMigrationStateSchedule extends CrossTldSingleton implements
   @VisibleForTesting
   static TimedTransitionProperty<MigrationState, MigrationStateTransition> getUncached() {
     return jpaTm()
-        .transactNew(
+        .transactWithoutBackup(
             () -> {
               try {
                 return jpaTm()
--- a/core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java
+++ b/core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java
@@ -32,7 +32,6 @@ import google.registry.model.common.DatabaseMigrationStateSchedule.MigrationStat
 import google.registry.model.ofy.CommitLogBucket;
 import google.registry.model.ofy.Ofy;
 import google.registry.persistence.VKey;
-import google.registry.persistence.transaction.JpaTransactionManagerImpl;
 import google.registry.persistence.transaction.TransactionEntity;
 import google.registry.testing.AppEngineExtension;
 import google.registry.testing.DatabaseHelper;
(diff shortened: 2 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1286)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 225 at r1](https://reviewable.io/reviews/google/nomulus/1286#-Mh4hEQT75d1UiFyPEIO:-Mh4hEQT75d1UiFyPEIP:bfn1r1u) ([raw file](https://github.com/google/nomulus/blob/97ac40e83b42b293c96727a8ae1ee806f40f7441/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L225)):*
> ```Java
>   static TimedTransitionProperty<MigrationState, MigrationStateTransition> getUncached() {
>     return jpaTm()
>         .transactWithoutBackup(
> ```

We might want to consider renaming this for JPA, or adding a new version that does what you want, because the intention it's using (no longer writing commit logs in Datastore) doesn't make sense here, and it's not 100% clear why using this one would also not check whether the DB is read-only.  Like I believe that the prober data deletion action is using the `transactWithoutBackup()` method too, but it _should_ respect the read-only mode, whereas here you want something that doesn't.  Seems like a new `jpaTm()`-specific method is warranted.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1286)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java, line 225 at r1](https://reviewable.io/reviews/google/nomulus/1286#-Mh4hEQT75d1UiFyPEIO:-Mh4sYS096T7rJH8ZQv1:bm8cowk) ([raw file](https://github.com/google/nomulus/blob/97ac40e83b42b293c96727a8ae1ee806f40f7441/core/src/main/java/google/registry/model/common/DatabaseMigrationStateSchedule.java#L225)):*

We might want to consider renaming this for JPA, or adding a new version that does what you want, because the intention it's using (no longer writing commit logs in Datastore) doesn't make sense here, and it's not 100% clear why using this one would also not check whether the DB is read-only.  Like I believe that the prober data deletion action is using the `transactWithoutBackup()` method too, but it _should_ respect the read-only mode, whereas here you want something that doesn't.  Seems like a new `jpaTm()`-specific method is warranted.
</blockquote></details>

The read-only mode is completely separate from this; it's unrelated. transactWithoutBackup will respect the read only mode when write actions are requested, as expected. This only means that we won't check the DB migration stage when instantiating this particular transaction. 

The read only mode PR has some other changes that allow us to retrieve or set the schedule without going through the standard pattern, but that's entirely separate from this issue.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1286)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1288, opened 2021-08-17
Remove files that are not longer used for create/update premium list

rachelguan:
Client/servlet model was used to support create or update premium list and the action files were where the logic resides. Both CreatePremiumListCommand and UpdatePremiumListComand recently were refactored to support display entity change in the same way as the CreateReservedListCommand and UpdateReservedListCommand. And now. client/servlet model is no longer applicable to create/update premium list. 
This PR was created to remove all the action files that are no longer used.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1288)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1288)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

PR #1290, opened 2021-08-19
Implement a util class to manage push queues using Cloud Tasks API

jianglai:
Push queues were part of App Engine when they debuted. As a result the
Task Queue API were part of the App Engine SDK and can only be used in
App Engine classic runtime. The new Cloud Tasks API can be used in any
runtime but it only supports push queues. In this PR we implement a util
class (CloudTasksUtils) like TaskQueueUtils to handle enqueuing tasks to
push queues using Cloud Tasks. One action (TldFanoutAction) was
converted to use the new API as a demo. Mass migration of other call sites of
the old API will follow in a separate PR.

TESTED=deployed to alpha and verified that tasks are corrected enqueued
and executed.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1290)


jianglai (commented):
Ready for review. PTAL.

> Reviewed 112 of 112 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290#-:-MhYiPLRDPLY1Wz4TVwZ:b-kkeupw)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

jianglai (commented):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 104 of 112 files at r1, 6 of 9 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: all files reviewed, 5 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/config/CloudTasksUtilsModule.java, line 45 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZE6bPDKk2gfjehabF:-MhZE6bPDKk2gfjehabG:bem1zm5) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java#L45)):*
> ```Java
>       @Config("projectId") String projectId,
>       @Config("locationId") String locationId,
>       Provider<CloudTasksClient> clientProvider,
> ```

Is the reason why we use a provider here so that the caller can maintain the lifecycle + closing of the CloudTasksClient? 
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 14 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZCvNlB2OFdxRuic7f:-MhZCvNlB2OFdxRuic7g:bqcvd1g) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/config/files/default-config.yaml#L14)):*
> ```YAML
>   # they are used without the trailing number in App Engine commands and Google Cloud Console.
>   # See: https://cloud.google.com/appengine/docs/locations
>   locationId: registry-location-id
> ```

Is there a corresponding PR to add the location in our internal repo? 
___
*[core/src/main/java/google/registry/cron/TldFanoutAction.java, line 178 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZV0wtDhES208JnYFd:-MhZV0wuAlXJ2rDxAbIY:b-uvpenc) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/cron/TldFanoutAction.java#L178)):*
> ```Java
>                 .getMillis());
>     return Task.newBuilder(
>             CloudTasksUtils.createPostTask(endpoint, Service.BACKEND.toString(), params))
> ```

Do we ever have any tasks that aren't on the backend service? 
___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 62 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZhZaoEYyrkoo_G_P9:-MhZhZaoEYyrkoo_G_PA:bjnhzco) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L62)):*
> ```Java
> 
> /** Static utility functions for testing task queues. */
> public class CloudTasksHelper {
> ```

This is basically taken from TaskQueueHelper, right? We probably should mention that. 
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZht4U2tMfQR2UtxKz:-MhZht4U2tMfQR2UtxL-:b-f90yg8) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/util/src/main/java/google/registry/util/CloudTasksUtils.java#L105)):*
> ```Java
>   private static Task createTask(
>       String path, HttpMethod method, String service, Multimap<String, String> params) {
>     checkArgument(
> ```

Can we use some sort of URI regularization code to be more lenient about the inputs, so that any input that doesn't conform to "/foo/bar" gets converted to the right format?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/config/CloudTasksUtilsModule.java, line 45 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZE6bPDKk2gfjehabF:-MhaMD6Y6EsTC32kqam1:b-6karsv) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java#L45)):*

Is the reason why we use a provider here so that the caller can maintain the lifecycle + closing of the CloudTasksClient? 
</blockquote></details>

Yes, `CloudTasksClient` implements `Autocloseable`. I'm not 100% sure but I suspect that without closing the client the request may not be guaranteed to be sent down the wire to the client. All the examples use the try-with-resource approach for this client (and probably for all gRPC clients in general). Added some comments to this effect.
___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 14 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZCvNlB2OFdxRuic7f:-MhaNVprDqwD4ShNyckZ:b-rs0mzr) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/config/files/default-config.yaml#L14)):*

Is there a corresponding PR to add the location in our internal repo? 
</blockquote></details>

Yes, sent a Gerrit CR to you. https://team-review.git.corp.google.com/c/domain-registry-eng/nomulus-internal/+/1192771
___
*[core/src/main/java/google/registry/cron/TldFanoutAction.java, line 178 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZV0wtDhES208JnYFd:-MhaNb9gACGcrIUSKSW3:b8dntpd) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/cron/TldFanoutAction.java#L178)):*

Do we ever have any tasks that aren't on the backend service? 
</blockquote></details>

The old way (with Task Queues) is that if you do not specify a service, the service which enqueues a task will be the one that executes it (see comments in CloudTasksUtil). Since this action is on the backend service all tasks it enqueues must already be on the backend service.
___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 62 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZhZaoEYyrkoo_G_P9:-MhaOQ-cCMvYqTx6V6Kf:b-896fix) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L62)):*

This is basically taken from TaskQueueHelper, right? We probably should mention that. 
</blockquote></details>

Done.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZht4U2tMfQR2UtxKz:-MhaOf_VDHuVwl1YOgnE:bc7ck7g) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/util/src/main/java/google/registry/util/CloudTasksUtils.java#L105)):*

Can we use some sort of URI regularization code to be more lenient about the inputs, so that any input that doesn't conform to "/foo/bar" gets converted to the right format? 
</blockquote></details>

I thought about it but it seems unlikely to work. We'd have to employ some heuristics to "regularize" a malformed URI, right? (Unless you are talking about simply adding a leading slash if there is none). That seems too complicated for such a util class. Ideally we'd use the `@Action` annotation from an action and get the path, but we also need to support user-supplied paths from query parameters (like in this action). Ultimately there are so many ways the provided URIs can be wrong I think we should just return an error message instead.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/config/files/default-config.yaml, line 14 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZCvNlB2OFdxRuic7f:-Mhnj4Ng7dbCcuZTznYx:biffcir) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/config/files/default-config.yaml#L14)):*

Yes, sent a Gerrit CR to you. https://team-review.git.corp.google.com/c/domain-registry-eng/nomulus-internal/+/1192771
</blockquote></details>

not sure how I missed that, sorry 
___
*[core/src/main/java/google/registry/cron/TldFanoutAction.java, line 178 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZV0wtDhES208JnYFd:-Mhnj9Y3C9Q1lGQFvTv0:bvkxzfp) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/cron/TldFanoutAction.java#L178)):*

The old way (with Task Queues) is that if you do not specify a service, the service which enqueues a task will be the one that executes it (see comments in CloudTasksUtil). Since this action is on the backend service all tasks it enqueues must already be on the backend service.
</blockquote></details>

Right, my question was if it's necessary to always specify the BACKEND service when calling `createPostTask` or if that can just be assumed 
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZht4U2tMfQR2UtxKz:-MhnjGvzCuVv7FQsUovR:br6lixn) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/util/src/main/java/google/registry/util/CloudTasksUtils.java#L105)):*

I thought about it but it seems unlikely to work. We'd have to employ some heuristics to "regularize" a malformed URI, right? (Unless you are talking about simply adding a leading slash if there is none). That seems too complicated for such a util class. Ideally we'd use the `@Action` annotation from an action and get the path, but we also need to support user-supplied paths from query parameters (like in this action). Ultimately there are so many ways the provided URIs can be wrong I think we should just return an error message instead.
</blockquote></details>

I'm not referring to taking in an input like that, I'm referring to something like a filesystem path object, which would normalize the path automatically, like `Paths.get("/", path).toString()` or something

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/cron/TldFanoutAction.java, line 178 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZV0wtDhES208JnYFd:-MhoBpzgAWTRDX-EOyVV:bqirh1) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/core/src/main/java/google/registry/cron/TldFanoutAction.java#L178)):*

Right, my question was if it's necessary to always specify the BACKEND service when calling `createPostTask` or if that can just be assumed 
</blockquote></details>

I doubt it can be assumed. I think there are cases where we enqueue backend tasks from frontend actions (EPP) or tool actions (Nomulus tool). I haven't done a thorough survey but making it explicit here doesn't seem to hurt as far as readability goes.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZht4U2tMfQR2UtxKz:-MhoG5EC9QTZ3NEmfyaK:b-r0urkr) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/util/src/main/java/google/registry/util/CloudTasksUtils.java#L105)):*

I'm not referring to taking in an input like that, I'm referring to something like a filesystem path object, which would normalize the path automatically, like `Paths.get("/", path).toString()` or something 
</blockquote></details>

Still not convinced that it will actually work. For filesystem paths there's value in using a regularization class like `Path` because different OSes have different delimiter and such. For URIs it's either right or wrong, so trying to regularize the input doesn't make much sense to me.

Also did some search on validation the path, which turned out to be not trivial either. See here: https://stackoverflow.com/questions/1600291/validating-url-in-java. We will also need to construct some dummy prefix to make this an actual URI in order to be used in a validator.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1290)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 105 at r4](https://reviewable.io/reviews/google/nomulus/1290#-MhZht4U2tMfQR2UtxKz:-MhtyYA81em7TariX5Zz:b2ib9nb) ([raw file](https://github.com/google/nomulus/blob/82b178801a43ddcfa2aa90f47508c8a8bae580fb/util/src/main/java/google/registry/util/CloudTasksUtils.java#L105)):*

Still not convinced that it will actually work. For filesystem paths there's value in using a regularization class like `Path` because different OSes have different delimiter and such. For URIs it's either right or wrong, so trying to regularize the input doesn't make much sense to me.

Also did some search on validation the path, which turned out to be not trivial either. See here: https://stackoverflow.com/questions/1600291/validating-url-in-java. We will also need to construct some dummy prefix to make this an actual URI in order to be used in a validator.
</blockquote></details>

Fair enough -- this makes sense then.

PR #1291, opened 2021-08-19
Compare SQL and Datastore objects in SQL->DS replay testing

gbrodman:
Add double-replay to the Host*Flow tests to show how this works. The
only change to the double replay itself is that now we store the
Datastore entity in the TransactionEntity object -- this is because we
use Objectify to serialize the objects into bytes and we need it to know
about the entity in question.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1291)


gbrodman:
hold up on this, might be better to show it in action with some tests

gbrodman:
ok should be good to go now

jianglai (approved):
> Reviewed 1 of 3 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1291)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1299, opened 2021-08-24
Clean up ReplicateToDatastoreAction and tests

gbrodman:
1. applyTransaction should throw an error if it fails; this allows us to
have more information in the caller (and it shouldn't usually happen)
2. Set a response code + payload now, since this is an action that is
called by cron
3. Add a method to the test log subject that allows us to check if a
severe log with a particular Throwable cause was logged (since the cause
isn't contained in the log message itself directly)

### DIFF
--- a/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java
+++ b/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java
@@ -18,6 +18,8 @@ import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.ofyTm;
 import static google.registry.request.Action.Method.GET;
+import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1299)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java, line 71 at r1](https://reviewable.io/reviews/google/nomulus/1299#-Mhy96yW6dL5F4C69I_-:-Mhy96yW6dL5F4C69I_0:bh0sfxx) ([raw file](https://github.com/google/nomulus/blob/84a0ef4ce161f230ec18403a4754cf722900ba27/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java#L71)):*
> ```
> public void run() {
> ```
Can we move this function back to where it was? Otherwise the the diff would not show up.
___
*[core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java, line 118 at r1](https://reviewable.io/reviews/google/nomulus/1299#-MhyEJff6Jqb13EhIZtW:-MhyEJff6Jqb13EhIZtX:b-v98r2y) ([raw file](https://github.com/google/nomulus/blob/84a0ef4ce161f230ec18403a4754cf722900ba27/core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java#L118)):*
> ```
> resetTask();
> ```
I presume you have to do this because `FakeResponse` can only have its payload set once? I am starting to think that this might be a burden rather than a good safeguard. Is it true that the real response can only have its payload set once? If not why are we enforcing this arbitrary restriction on a fake? 
___
*[core/src/test/java/google/registry/testing/LogsSubject.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1299#-MhyBvxVBnHbST22ubl7:-MhyBvxVBnHbST22ubl8:b-2u9vao) ([raw file](https://github.com/google/nomulus/blob/84a0ef4ce161f230ec18403a4754cf722900ba27/core/src/test/java/google/registry/testing/LogsSubject.java#L68)):*
> ```
> public void hasSevereLogWithCause
> ```
I'm not familiar with how `Subject`s work. Where is the logic that filters out only severe logs?

PR #1313, opened 2021-09-09
Update references to RDAP RFCs

gbrodman:
There were minor changes to the RDAP RFCs used -- we don't need to
change anything since we already comply with all of the changes, but we
should refer to the newer RFCs in the code.

https://docs.google.com/document/d/1Y89mJm_5VRp890LhrLajx5-5hmi07N4jAK9ZZS31xLQ/edit#heading=h.uyb2kr1ik95h for a bit more context but it's probably not necessary 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1313)

### DIFF
--- a/core/src/main/java/google/registry/rdap/AbstractJsonableObject.java
+++ b/core/src/main/java/google/registry/rdap/AbstractJsonableObject.java
@@ -339,7 +339,7 @@ abstract class AbstractJsonableObject implements Jsonable {
       return new JsonPrimitive((Boolean) object);
     }
     if (object instanceof DateTime) {
-      // According to RFC7483 section 3, the syntax of dates and times is defined in RFC3339.
+      // According to RFC 9083 section 3, the syntax of dates and times is defined in RFC3339.
       //
       // According to RFC3339, we should use ISO8601, which is what DateTime.toString does!
       return new JsonPrimitive(((DateTime) object).toString());
--- a/core/src/main/java/google/registry/rdap/RdapActionBase.java
+++ b/core/src/main/java/google/registry/rdap/RdapActionBase.java
@@ -50,8 +50,8 @@ import org.joda.time.DateTime;
 /**
  * Base RDAP (new WHOIS) action for all requests.
  *
- * @see <a href="https://tools.ietf.org/html/rfc7482">
- *        RFC 7482: Registration Data Access Protocol (RDAP) Query Format</a>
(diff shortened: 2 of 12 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1313)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rdap/AbstractJsonableObject.java, line 342 at r1](https://reviewable.io/reviews/google/nomulus/1313#-MjVI0ro3JO9rhQGYYQQ:-MjVI0rp3w4nm-Pg85pK:bkmvo0n) ([raw file](https://github.com/google/nomulus/blob/dc1554957fa1c0607b216ce3cc51187db27a707e/core/src/main/java/google/registry/rdap/AbstractJsonableObject.java#L342)):*
> ```Java
>     }
>     if (object instanceof DateTime) {
>       // According to RFC9083 section 3, the syntax of dates and times is defined in RFC3339.
> ```

This PR is inconsistently including/not including the space between "RFC" and the number.  I think you should use the space everywhere, as that's both more readable and also how the RFCs actually refer to themselves.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1313)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/rdap/AbstractJsonableObject.java, line 342 at r1](https://reviewable.io/reviews/google/nomulus/1313#-MjVI0ro3JO9rhQGYYQQ:-MjVISr53_8qEJYcMIfQ:b-u7f09e) ([raw file](https://github.com/google/nomulus/blob/dc1554957fa1c0607b216ce3cc51187db27a707e/core/src/main/java/google/registry/rdap/AbstractJsonableObject.java#L342)):*

This PR is inconsistently including/not including the space between "RFC" and the number.  I think you should use the space everywhere, as that's both more readable and also how the RFCs actually refer to themselves.
</blockquote></details>

Yeah, it was copying the existing code w/r/t spaces or not. Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1313)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1317, opened 2021-09-13
Rename client ID to registrar ID in most places

CydeWeys:
This is a code-only change, that shouldn't require any sort of data
migration. Correspondingly, there are some existing uses of clientId that are
not migrated (e.g. Datastore fields, task queue payloads, URL parameters for
actions that might be hit from task queues, etc.). And it of course doesn't
modify any fields in EPP XML. Note that the Cloud SQL schema fields are
already named using the registar_id pattern.

This also doesn't yet touch on the -c parameters in nomulus tools; that will be
coming later (since that is an external manual touch-point, it will require a
lot more in the way of changes to various meta scripts and documentation).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1317)

PR #1320, opened 2021-09-14
Skip synthetic history entries for resources that don't need them

gbrodman:
The reason for creating synthetic history entries is so that we can
guarantee that each EppResource's most recent *History object contains
that resource at that point in time. If the most recent *History object
in SQL contains that resource already, there is no need to create a
synthetic *History object for that resource.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1320)

### DIFF
--- a/config/presubmits.py
+++ b/config/presubmits.py
@@ -206,15 +206,12 @@ PRESUBMITS = {
          "RegistryJpaIO.java",
          # TODO(b/179158393): Remove everything below, which should be done
          # using Criteria
-         "ForeignKeyIndex.java",
-         "HistoryEntryDao.java",
          "JpaTransactionManager.java",
          "JpaTransactionManagerImpl.java",
          # CriteriaQueryBuilder is a false positive
          "CriteriaQueryBuilder.java",
          "RdapDomainSearchAction.java",
          "RdapNameserverSearchAction.java",
-         "RdapSearchActionBase.java",
          "ReadOnlyCheckingEntityManager.java",
          "RegistryQuery",
          },
--- a/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
+++ b/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
@@ -228,7 +228,7 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
           tm().transact(
                   () ->
                       jpaTm()
-                          .query(
+                          .criteriaQuery(
                               CriteriaQueryBuilder.create(clazz)
                                   .whereFieldIsIn(property, foreignKeys)
                                   .build())
--- a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java
+++ b/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java
@@ -24,6 +24,7 @@ import static google.registry.util.DateTimeUtils.START_OF_TIME;
 
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Streams;
 import google.registry.model.EppResource;
 import google.registry.model.contact.ContactHistory;
@@ -48,6 +49,25 @@ import org.joda.time.DateTime;
  * is currently considered the primary database.
  */
 public class HistoryEntryDao {
+
+  public static ImmutableMap<Class<? extends EppResource>, Class<? extends HistoryEntry>>
+      RESOURCE_TYPES_TO_HISTORY_TYPES =
+          ImmutableMap.of(
+              ContactResource.class,
+              ContactHistory.class,
+              DomainBase.class,
+              DomainHistory.class,
+              HostResource.class,
+              HostHistory.class);
+
+  public static ImmutableMap<Class<? extends HistoryEntry>, String> REPO_ID_FIELD_NAMES =
+      ImmutableMap.of(
+          ContactHistory.class,
+          "contactRepoId",
+          DomainHistory.class,
+          "domainRepoId",
+          HostHistory.class,
+          "hostRepoId");
 
   /** Loads all history objects in the times specified, including all types. */
   public static ImmutableList<HistoryEntry> loadAllHistoryObjects(
@@ -164,7 +184,7 @@ public class HistoryEntryDao {
   private static <T extends HistoryEntry> Stream<T> loadHistoryObjectFromSqlByRegistrars(
       Class<T> historyClass, ImmutableCollection<String> registrarIds) {
     return jpaTm()
-        .query(
+        .criteriaQuery(
             CriteriaQueryBuilder.create(historyClass)
                 .whereFieldIsIn("clientId", registrarIds)
                 .build())
@@ -188,34 +208,32 @@ public class HistoryEntryDao {
 
     return ImmutableList.sortedCopyOf(
         Comparator.comparing(HistoryEntry::getModificationTime),
-        jpaTm().query(criteriaQuery).getResultList());
+        jpaTm().criteriaQuery(criteriaQuery).getResultList());
   }
 
   private static Class<? extends HistoryEntry> getHistoryClassFromParent(
       Class<? extends EppResource> parent) {
-    if (parent.equals(ContactResource.class)) {
-      return ContactHistory.class;
-    } else if (parent.equals(DomainBase.class)) {
-      return DomainHistory.class;
-    } else if (parent.equals(HostResource.class)) {
-      return HostHistory.class;
+    if (!RESOURCE_TYPES_TO_HISTORY_TYPES.containsKey(parent)) {
+      throw new IllegalArgumentException(
+          String.format("Unknown history type for parent %s", parent.getName()));
     }
-    throw new IllegalArgumentException(
-        String.format("Unknown history type for parent %s", parent.getName()));
+    return RESOURCE_TYPES_TO_HISTORY_TYPES.get(parent);
   }
 
   private static String getRepoIdFieldNameFromHistoryClass(
       Class<? extends HistoryEntry> historyClass) {
-    return historyClass.equals(ContactHistory.class)
-        ? "contactRepoId"
-        : historyClass.equals(DomainHistory.class) ? "domainRepoId" : "hostRepoId";
+    if (!REPO_ID_FIELD_NAMES.containsKey(historyClass)) {
+      throw new IllegalArgumentException(
+          String.format("Unknown history type %s", historyClass.getName()));
+    }
+    return REPO_ID_FIELD_NAMES.get(historyClass);
   }
 
   private static <T extends HistoryEntry> List<T> loadAllHistoryObjectsFromSql(
       Class<T> historyClass, DateTime afterTime, DateTime beforeTime) {
     CriteriaBuilder criteriaBuilder = jpaTm().getEntityManager().getCriteriaBuilder();
     return jpaTm()
-        .query(
+        .criteriaQuery(
             CriteriaQueryBuilder.create(historyClass)
                 .where("modificationTime", criteriaBuilder::greaterThanOrEqualTo, afterTime)
                 .where("modificationTime", criteriaBuilder::lessThanOrEqualTo, beforeTime)
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManager.java
@@ -36,7 +36,7 @@ public interface JpaTransactionManager extends TransactionManager {
   <T> TypedQuery<T> query(String sqlString, Class<T> resultClass);
 
   /** Creates a JPA SQU query for the given criteria query. */
-  <T> TypedQuery<T> query(CriteriaQuery<T> criteriaQuery);
+  <T> TypedQuery<T> criteriaQuery(CriteriaQuery<T> criteriaQuery);
 
   /**
    * Creates a JPA SQL query for the given query string.
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -135,7 +135,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
   }
 
   @Override
-  public <T> TypedQuery<T> query(CriteriaQuery<T> criteriaQuery) {
+  public <T> TypedQuery<T> criteriaQuery(CriteriaQuery<T> criteriaQuery) {
     return new DetachingTypedQuery<>(getEntityManager().createQuery(criteriaQuery));
   }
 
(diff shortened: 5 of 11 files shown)


CydeWeys (changes requested):
I'm confused about the logic here.  What I expected was "Give this command a DateTime (after the most recent initial DB import) as input and then it writes out synthetic HistoryEntries for every EPP resource that doesn't have a HistoryEntry more recent than that DateTime".

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-MjduXUD9yzpqudouUSa:blhta1g)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/1320#-MjduMzsEPYde3jsJEZz:-MjduMzsEPYde3jsJE_-:blccdpa) ([raw file](https://github.com/google/nomulus/blob/3a80412481597ac6f50df289b7b3716422c97c99/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L136)):*
> ```Java
>   private static Class<? extends HistoryEntry> getHistoryClassFromParent(
>       Class<? extends EppResource> parent) {
>     if (parent.equals(ContactResource.class)) {
> ```

This should be a Map.

gbrodman (commented):
It isn't necessary to give it a DateTime; this accomplishes the same thing without the manual step. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-MjdwjcW6S1kF5dPnsps:b-fek6xx)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/1320#-MjduMzsEPYde3jsJEZz:-MjdwhtN6SMDkJ32Q3d_:bb74ce2) ([raw file](https://github.com/google/nomulus/blob/3a80412481597ac6f50df289b7b3716422c97c99/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L136)):*

This should be a Map.
</blockquote></details>

Sure

CydeWeys (commented):
So then what is this doing exactly?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-Mje6K3O55eF_cKkKxLb:b-bhrejk)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

gbrodman (commented):
for every resource, if we already have a history object in SQL that contains a representation of that resource at that point in time, we skip it 

alternatively put, for every resource, we save a new HistoryEntry if and only if there does not exist a *History object in SQL that contains that object at that point in time. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-Mje6t8EEHv97qUuLNzq:bd1yzw9)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (commented):
That's not strictly accurate though, right?  It's not if there exists an object in SQL that contains that object at that point in time, it's only if the most _recent_ history object contains that information, right?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-Mje6wMoDvqoYj8QG3Nx:b-6os748)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

gbrodman (commented):
If you want to nitpick semantics yes, however those statements are equivalent. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-Mje7gNo3jK7Xkl4zfjN:b-hzanan)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (changes requested):
They're not equivalent though.  There's a difference between "the most recent X" and "any X".  The former seems like it works here in this case, and fortunately is more performant to calculate, but it's not the same.  The commit log message is still not accurate in that regard and it's inhibiting understanding as to exactly what's going on.

So, is there any situation where the most recent history entry has data, but that data is wrong/unusable for some reason?  Or in all such cases would we be wiping out the existing contents of Cloud SQL and re-importing (as we're currently doing in sandbox and will do in prod soon on account of the incorrect update times)?

Anyway, looking good overall, modulo a few remaining questions.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-MjfD_smDclnTMovTbEv:bysg2tv)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[config/presubmits.py, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfFUI82H4i1-KDCcRA:-MjfFUI96qfP0mwCrSYr:bgfxry) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/config/presubmits.py#L215)):*
> ```Python
>          # CriteriaQueryBuilder is a false positive
>          "CriteriaQueryBuilder.java",
>          "CreateSyntheticHistoryEntriesAction.java",
> ```

Why is this necessary?  Offhand, I don't see you doing anything that would cause this presubmit to fail.
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 126 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfFnO07cZF9cJWSeb7:-MjfFnO07cZF9cJWSeb8:b-25bptn) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L126)):*
> ```Java
>                       .orderByDesc("modificationTime")
>                       .build();
>               return jpaTm().query(criteriaQuery).setMaxResults(1).getResultStream().findFirst();
> ```

What happens if this returns zero results?  Is that even possible?  In theory there's always that initial create, but is it somehow possible for the resource itself to have been async written over but the corresponding create history entry hasn't yet been?
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 143 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfEEwh5cEduaJ-N4i8:-MjfEEwh5cEduaJ-N4i9:b-wnmwp5) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L143)):*
> ```Java
>   private static String getRepoIdFieldNameFromHistoryClass(
>       Class<? extends HistoryEntry> historyClass) {
>     return historyClass.equals(ContactHistory.class)
> ```

This isn't strictly safe -- it's using `hostRepoId` as a default value on an unexpected class rather than throwing an error.  It seems to me like the preferred solution here would be something like `abstract String getRepoIdFieldName();` in `HistoryEntry`, then all implementations are forced to implement this.

gbrodman (commented):
No, the only time we ever write resourceless entries to the History tables is during the initial SQL population. The result is that the questions "are there any history rows with resources?" and "does the last history row have a resource?" will always have the same answer 

Updated commit message

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-MjfKZcWB0UtlK1EFjRK:bntng84)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[config/presubmits.py, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfFUI82H4i1-KDCcRA:-MjfJZXw56Tsd1u6f0n-:b-8593as) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/config/presubmits.py#L215)):*

Why is this necessary?  Offhand, I don't see you doing anything that would cause this presubmit to fail.
</blockquote></details>

it's the call to `jpaTm().query(criteriaQuery)`

because we use a simple regex, we can't differentiate between `query(String)` and `query(CriteriaQuery)`
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 126 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfFnO07cZF9cJWSeb7:-MjfKFcSD96-6hd_qzAi:bpnz4gu) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L126)):*

What happens if this returns zero results?  Is that even possible?  In theory there's always that initial create, but is it somehow possible for the resource itself to have been async written over but the corresponding create history entry hasn't yet been?
</blockquote></details>

It shouldn't be possible, since every transaction that creates an EppResource should also create a HistoryEntry for said resource
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 143 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfEEwh5cEduaJ-N4i8:-MjfJrqT15iGyQiVpIkZ:b-20fqes) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L143)):*

This isn't strictly safe -- it's using `hostRepoId` as a default value on an unexpected class rather than throwing an error.  It seems to me like the preferred solution here would be something like `abstract String getRepoIdFieldName();` in `HistoryEntry`, then all implementations are forced to implement this.
</blockquote></details>

We cannot make `HistoryEntry` abstract until post-migration, but we can use a map.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[config/presubmits.py, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfFUI82H4i1-KDCcRA:-MjjFoJS2JtQxh6XvFKj:bug047i) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/config/presubmits.py#L215)):*

it's the call to `jpaTm().query(criteriaQuery)`

because we use a simple regex, we can't differentiate between `query(String)` and `query(CriteriaQuery)`
</blockquote></details>

It sounds like the presubmit isn't really working out for us so much then if it can't differentiate between these two.  Wonder if it'd be worth renaming the `query(CriteriaQuery)` method so the presubmit can be more targeted?
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 143 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfEEwh5cEduaJ-N4i8:-Mjj-ZrI3kEZmsr_dN52:b6ou0by) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L143)):*

We cannot make `HistoryEntry` abstract until post-migration, but we can use a map. 
</blockquote></details>

It wouldn't need to be abstract -- it could also have a base method that throws an 'Unknown type' exception, and then it's overridden on all three that inherit it.

CydeWeys (approved):
That's good context to have to explain how/why this works.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-MjjGJcMBL2z6IdtmsTT:b-xpl52w)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[config/presubmits.py, line 215 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfFUI82H4i1-KDCcRA:-MjjH5x5EdskMw4u3i2K:b-mh1vhf) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/config/presubmits.py#L215)):*

It sounds like the presubmit isn't really working out for us so much then if it can't differentiate between these two.  Wonder if it'd be worth renaming the `query(CriteriaQuery)` method so the presubmit can be more targeted?
</blockquote></details>

Yes, I think so. Done. 
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 143 at r2](https://reviewable.io/reviews/google/nomulus/1320#-MjfEEwh5cEduaJ-N4i8:-MjjGl2-6mXQoRXxFLks:b-msc265) ([raw file](https://github.com/google/nomulus/blob/384bdd880d09e336532336ccca57959e443121e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L143)):*

It wouldn't need to be abstract -- it could also have a base method that throws an 'Unknown type' exception, and then it's overridden on all three that inherit it.
</blockquote></details>

I think I don't like that given that we don't actually have an instance of the history object here (that is what we are trying to get). So we'd have to do some reflection

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 179 at r4](https://reviewable.io/reviews/google/nomulus/1320#-MjjXz2UAc1XNk1GWolM:-MjjXz2V8OufKswzCVCU:b-s40xj1) ([raw file](https://github.com/google/nomulus/blob/52ea527846039112ed0d066566b022f5ae74ebbf/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L179)):*
> ```Java
>                         .put(
>                             HistoryEntry.createBuilderForResource(eppResource)
>                                 .setClientId(registryAdminRegistrarId)
> ```

You'll need to change this now that https://github.com/google/nomulus/pull/1317 is merged.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 179 at r4](https://reviewable.io/reviews/google/nomulus/1320#-MjjXz2UAc1XNk1GWolM:-MjjdQGg52kzj0IuZgn-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/52ea527846039112ed0d066566b022f5ae74ebbf/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L179)):*

You'll need to change this now that https://github.com/google/nomulus/pull/1317 is merged.
</blockquote></details>

Done.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1320#-:-Mjk-mJb6-DFMmkyEfBF:bnfp4nl)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1323, opened 2021-09-16
Include Cursor in the initial SQL population

gbrodman:
The test for this also required a bit of a fix in the Cursor scope
initialization. If you persist a Key<?> in some object in Datastore, it
persists just the standard data you'd expect, basically the parent, the
kind, and the object's ID/name. Then, when you load it back in from
Datastore it uses the app ID of whatever environment you're loading in
(the Key contains this info even though it isn't included in the
toString() of Key)

If you persist the websafe string format of a Key (which is what we do
for Cursors), it includes the app ID so when you load it, it contains
the old app ID not the new one (if the app ID has changed).

In the pipelines, we use the standard default environment which has a
different app ID from the test environment that's set up by the
DatastoreEntityExtension.

As a result, we should check in Cursor to see if the key is *any*
cross-tld-key, rather than the exact one that exists within this app
environment.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1323)

### DIFF
--- a/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
+++ b/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
@@ -25,6 +25,7 @@ import google.registry.backup.VersionedEntity;
 import google.registry.beam.common.RegistryJpaIO;
 import google.registry.beam.initsql.Transforms.RemoveDomainBaseForeignKeys;
 import google.registry.model.billing.BillingEvent;
+import google.registry.model.common.Cursor;
 import google.registry.model.contact.ContactResource;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.domain.token.AllocationToken;
@@ -62,6 +63,7 @@ import org.joda.time.DateTime;
  * <ol>
  *   <li>{@link Registry}: Assumes that {@code PremiumList} and {@code ReservedList} have been set
  *       up in the SQL database.
+ *   <li>{@link Cursor}: Logically can depend on {@code Registry}, but without foreign key.
  *   <li>{@link Registrar}: Logically depends on {@code Registry}, Foreign key not modeled yet.
  *   <li>{@link ContactResource}: references {@code Registrar}
  *   <li>{@link RegistrarContact}: references {@code Registrar}.
@@ -101,7 +103,11 @@ public class InitSqlPipeline implements Serializable {
    */
   private static final ImmutableList<Class<?>> PHASE_ONE_ORDERED =
       ImmutableList.of(
-          Registry.class, Registrar.class, ContactResource.class, RegistrarContact.class);
+          Registry.class,
+          Cursor.class,
+          Registrar.class,
+          ContactResource.class,
+          RegistrarContact.class);
 
   /**
    * Datastore kinds to be written to the SQL database after the cleansed version of {@link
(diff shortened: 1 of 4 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1323)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1323#-MjkJTvUBPOeiwO1hvSM):*
Why can't we rely on this?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1323)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1323#-MjkJTvUBPOeiwO1hvSM):*

Why can't we rely on this?
</blockquote></details>

I honestly don't know. 

They even have the same String representation -- `Key<?>(EntityGroupRoot("cross-tld"))`

Something weird with `Key`'s `.equals()`?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1323)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1323#-MjkJTvUBPOeiwO1hvSM):*

I honestly don't know. 

They even have the same String representation -- `Key<?>(EntityGroupRoot("cross-tld"))`

Something weird with `Key`'s `.equals()`? 
</blockquote></details>

oh wait, the app ID is different in the keys

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1323)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1323#-MjkJTvUBPOeiwO1hvSM):*

oh wait, the app ID is different in the keys
</blockquote></details>

ok yeah so what's happening here is that if you persist a `Key<?>` in some object in Datastore, it persists just the standard data you'd expect, which basically only includes the parent, the kind, and the ID. Then, when you load it back in from Datastore, it uses the app ID of whatever environment you're loading it in 

If you persist the websafe string (which is what we do for cursors) it includes the app ID 

In the pipelines, we use the default environment which has a different app ID from the test environment

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1323)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @weiminyu)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1323#-MjkJTvUBPOeiwO1hvSM):*

ok yeah so what's happening here is that if you persist a `Key<?>` in some object in Datastore, it persists just the standard data you'd expect, which basically only includes the parent, the kind, and the ID. Then, when you load it back in from Datastore, it uses the app ID of whatever environment you're loading it in 

If you persist the websafe string (which is what we do for cursors) it includes the app ID 

In the pipelines, we use the default environment which has a different app ID from the test environment 
</blockquote></details>

Wow, TIL that the app ID is in the key.  No idea.  May as well update the commit message then.

gbrodman:
Commit message updated

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1323#-:-Mjp3NjpDCTB1vyAhZ2d:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1325, opened 2021-09-17
Clean up tx manager insert() signature and add convenience helper method

CydeWeys:
This is the first of a series of PRs to clean up the type signatures on the
TransactionManager methods (which are way too generic), along with creating some
helper methods for use in tests only that don't require creating transactions
all over the place, thus reducing visual noise at callsites. This first method
is DatabaseHelper.insertInDb(), but there will be plenty of others. Note that
this is only for the Cloud SQL transaction manager -- I'm not bothering to
migrate any Datastore-only code, as that will be going away soon enough.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1325)


sarahcaseybot (approved):
> Reviewed 40 of 40 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1325)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1347, opened 2021-09-29
Canonicalize domain/host names in initial import script

CydeWeys:


CydeWeys:
Where's the best place to add tests for this?

Also I might need to do so using raw `DatastoreService` methods, as you can't currently set a domain/host with a non-canonicalized hostname.

weiminyu (commented):
Looks good to me. 

For testing, there are a few options:

1. Test it as part of RegistryJpaIO.Write, you can use RegistryJpaWriteTest as template.
2. Test it in full pipeline run: use InitSqlPipelineTest, add bad data in setup and assertions in the end.
3. Put your changes in a method and just test it with Datastore, outside of pipelines. DomainBaseUtilTest is an example.  

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1347#-:-MkrCGH54UfSeP5oPb07:b9ddw76)* status: 0 of 1 files reviewed, all discussions resolved

CydeWeys:
PTAL, now with tests.

PR #1349, opened 2021-09-29
Fix LGTM breakage

jianglai:
Set Python version floor at 3.6.0, which is what is available on LGTM
machines.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1349)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -209,17 +209,16 @@ rootProject.ext {
 
 task runPresubmits(type: Exec) {
 
-  // Find a python version greater than 3.7.3 (this is somewhat arbitrary, we
-  // know we'd like at least 3.6, but 3.7.3 is the latest that ships with
-  // Debian so it seems like that should be available anywhere).
-  def MIN_PY_VER = 0x3070300
+  // Find a python version greater than 3.6.0. Some of our infrastructure (like
+  // LGTM) still runs on Ubuntu 18.04, which only has the Python 3.6 branch.
+  def MIN_PY_VER = 0x030600F0
   if (pyver('python') >= MIN_PY_VER) {
     executable 'python'
   } else if (pyver('/usr/bin/python3') >= MIN_PY_VER) {
     executable '/usr/bin/python3'
   } else {
     throw new GradleException("No usable Python version found (build " +
-                              "requires at least python 3.7.3)");
+                              "requires at least python 3.6.0)");
   }
   args('config/presubmits.py')
 }


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1349)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai and @mindhog)

___
*[build.gradle, line 213 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4fyc4HAUxII2bFeo:-Mkn4fyc4HAUxII2bFep:brtimue) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L213)):*
> ```Gradle
> 
>   // Find a python version greater than 3.6.0. Some of our infrastructure (like
>   // LGTM) still runs on Ubuntu 18.04, which only has the Python 3.6 branch.
> ```

Argh.  This sucks.  Is LGTM actually running this task?  And how did it work before if we didn't have python3 installed?
___
*[build.gradle, line 214 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4JLG6YI49sGTZmZt:-Mkn4JLH8Ftvu1BVtPXy:b-441c4l) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L214)):*
> ```Gradle
>   // Find a python version greater than 3.6.0. Some of our infrastructure (like
>   // LGTM) still runs on Ubuntu 18.04, which only has the Python 3.6 branch.
>   def MIN_PY_VER = 0x030600F0
> ```

You should probably replace the "F0" with "00" (unless you know what it means and are sure it's appropriate?  I didn't and wasn't :-)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1349)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[build.gradle, line 213 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4fyc4HAUxII2bFeo:-Mkn8YdB8HONzBMQVG6t:boc73hf) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L213)):*

Argh.  This sucks.  Is LGTM actually running this task?  And how did it work before if we didn't have python3 installed?
</blockquote></details>

I presume it doesn't need Python 3.7.3, and whatever Ubuntu 18.04 comes with is sufficient.
___
*[build.gradle, line 214 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4JLG6YI49sGTZmZt:-Mkn8rJU-v1tv_d0dI6H:b-6chiy4) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L214)):*

You should probably replace the "F0" with "00" (unless you know what it means and are sure it's appropriate?  I didn't and wasn't :-)  
</blockquote></details>

Haha I actually do know. It means it's the final release version (instead of alphas or betas).

https://docs.python.org/3/c-api/apiabiversion.html#apiabiversion

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1349)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)

___
*[build.gradle, line 213 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4fyc4HAUxII2bFeo:-Mkn9HGs409yDC8Q3vcK:bynvpce) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L213)):*

I presume it doesn't need Python 3.7.3, and whatever Ubuntu 18.04 comes with is sufficient.
</blockquote></details>

Not currently, we've deliberately avoided using even python 3.6 features because of this kind of breakage in the past (it's hit us at least 3 times).   But this becomes more of an issue as we move forward.  Is it really reasonable to expect developers to code to a 5 year old version of the language?

But I'm still unclear as to why this is breaking LGTM? Surely they're not doing a gradle build?   Also it seems like [we might be able to specify the python version?](https://help.semmle.com/lgtm-enterprise/user/help/analysis-faqs.html#how-is-the-python-version-identified)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1349)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[build.gradle, line 213 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4fyc4HAUxII2bFeo:-MknGL1qEkLIq6RniH3Z:b-xvw6bk) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L213)):*

Not currently, we've deliberately avoided using even python 3.6 features because of this kind of breakage in the past (it's hit us at least 3 times).   But this becomes more of an issue as we move forward.  Is it really reasonable to expect developers to code to a 5 year old version of the language?

But I'm still unclear as to why this is breaking LGTM? Surely they're not doing a gradle build?   Also it seems like [we might be able to specify the python version?](https://help.semmle.com/lgtm-enterprise/user/help/analysis-faqs.html#how-is-the-python-version-identified)
</blockquote></details>

LGTM is doing a Gradle build. In doing so it can perform compile time analysis on the code base. The python version set up does not work because it only works for the `python` section of analysis, i. e. on Python code (https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html#python-setup). Here we use it during a Java build so the version specified there is not used.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1349)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[build.gradle, line 213 at r2](https://reviewable.io/reviews/google/nomulus/1349#-Mkn4fyc4HAUxII2bFeo:-Mkrvc6nCuAJQST7190Q:b-swdh1) ([raw file](https://github.com/google/nomulus/blob/93a14049b81aed58ed5a2852d29b98a09f660298/build.gradle#L213)):*

LGTM is doing a Gradle build. In doing so it can perform compile time analysis on the code base. The python version set up does not work because it only works for the `python` section of analysis, i. e. on Python code (https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html#python-setup). Here we use it during a Java build so the version specified there is not used.
</blockquote></details>

Maybe we should let this go through before you can find a more permanent solution?

mindhog:
#1352 should fix this if lgtm.com isn't just doing "gradlew build" (just noticed you've approved that, thanks).  If it doesn't, we'll change the build_command to just build the java classes.

jianglai:
> #1352 should fix this if lgtm.com isn't just doing "gradlew build" (just noticed you've approved that, thanks). If it doesn't, we'll change the build_command to just build the java classes.

If you check the lgtm presubmit status you can see it is still broken. Maybe you can add the `build_command` in that PR?

mindhog (commented):
The LGTM failures (both in #1352 and in this PR) appear to be happening when doing a "gradle clean" build on the baseline commit -- I don't think there's any way to fix them in either PR, I think we need to get these committed in order to see the presubmit pass in the next PR.
 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1349#-:-Mkw5WWd6bS3JBzsRQ_s:bgvlp8b)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mindhog)

PR #1352, opened 2021-09-30
Defer python discovery until presubmit task

mindhog:
Our presubmit requires a version of python that is more recent than what
lgtm.com's build environments have installed. LGTM shouldn't be 
running the presubmit, but the check will happen regardless of whether 
the presubmit is run or not, since the check is in the task closure instead 
of in a task action.

Attempt to fix this by moving the check to a "doFirst" (a task action).  
This solves the problem when building locally, unfortunately there's no 
easy way to test it on lgtm.com, since lgtm.com appears to fail on a
"./gradlew clean" build of the baseline commit before even getting to
the PR and a) we don't seem to retain logs from a successful run and b)
I can't find any documentation of what lgtm.com's autobuild does by 
default.

In short, this PR fixes a problem (the build won't run at all on systems
that don't have a recent enough version of python) but there may still
be issues on lgtm.com that will require us to customize the 
build_command used.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1352)

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -4,3 +4,7 @@ extraction:
       packages: "npm"
     index:
       java_version: "11"
+      build_command:
+        # Just build normal classes and test classes, this speeds things up
+        # considerably and avoids problems with the installed python version.
+        - ./gradlew assemble testClasses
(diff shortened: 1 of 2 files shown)


jianglai:
If I had to guess LGTM.com is probably calling "gradle build".

jianglai (approved):

mindhog:
Hey, can you LGTM again?  It doesn't seem to be recognizing it after a rebase.

PR #1354, opened 2021-09-30
Use a more efficient query to find resources in histories

gbrodman:
SELECT EXISTS statements will be faster / use fewer read-locks than statements that try to do sorting or return full resources

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1354)

### DIFF
--- a/config/presubmits.py
+++ b/config/presubmits.py
@@ -213,6 +213,7 @@ PRESUBMITS = {
         {"src/test", "ActivityReportingQueryBuilder.java",
          # This class contains helper method to make queries in Beam.
          "RegistryJpaIO.java",
+         "CreateSyntheticHistoryEntriesAction.java",
          # TODO(b/179158393): Remove everything below, which should be done
          # using Criteria
          "JpaTransactionManager.java",
--- a/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java
+++ b/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java
@@ -21,6 +21,7 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 import static google.registry.persistence.transaction.TransactionManagerFactory.ofyTm;
 
 import com.google.appengine.tools.mapreduce.Mapper;
+import com.google.common.base.CaseFormat;
 import com.google.common.collect.ImmutableList;
 import com.googlecode.objectify.Key;
 import google.registry.config.RegistryConfig.Config;
@@ -29,16 +30,12 @@ import google.registry.mapreduce.inputs.EppResourceInputs;
 import google.registry.model.EppResource;
 import google.registry.model.domain.DomainHistory;
 import google.registry.model.reporting.HistoryEntry;
-import google.registry.persistence.transaction.CriteriaQueryBuilder;
 import google.registry.rde.RdeStagingAction;
 import google.registry.request.Action;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import google.registry.tools.server.GenerateZoneFilesAction;
-import java.util.Optional;
 import javax.inject.Inject;
-import javax.persistence.criteria.CriteriaBuilder;
-import javax.persistence.criteria.CriteriaQuery;
 
 /**
  * A mapreduce that creates synthetic history objects in SQL for all {@link EppResource} objects.
@@ -105,8 +102,9 @@ public class CreateSyntheticHistoryEntriesAction implements Runnable {
         .sendLinkToMapreduceConsole(response);
   }
 
-  // Lifted from HistoryEntryDao
-  private static Optional<? extends HistoryEntry> mostRecentHistoryFromSql(EppResource resource) {
+  // Returns true iff any of the *History objects in SQL contain a representation of this resource
+  // at the point in time that the *History object was created.
+  private static boolean hasHistoryContainingResource(EppResource resource) {
     return jpaTm()
         .transact(
             () -> {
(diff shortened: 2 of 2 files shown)


gbrodman:
testing on QA, this only uses one or two readlocks as opposed to the 30 of the other query

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1354)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/1354#-MkwbGbu98-AIWOdzH2c:-MkwbGbu98-AIWOdzH2d:b-9ym9ad) ([raw file](https://github.com/google/nomulus/blob/b09278d897877b7a43b9955c40aa631b55e96690/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L105)):*
> ```Java
>   }
> 
>   private static boolean resourceAtPointInTimeExists(EppResource resource) {
> ```

I don't understand this method name.  At what point in time are you checking that this resource exists?  What does `AtPointInTime` have to do with anything that's going on here?
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 114 at r1](https://reviewable.io/reviews/google/nomulus/1354#-Mkwa-Dx9ZzluuU2D-nM:-Mkwa-Dx9ZzluuU2D-nN:b-vg139m) ([raw file](https://github.com/google/nomulus/blob/b09278d897877b7a43b9955c40aa631b55e96690/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L114)):*
> ```Java
>               // The field representing repo ID unfortunately varies by history class
>               String repoIdFieldName = getRepoIdFieldNameFromHistoryClass(historyClass);
>               String snakeCaseFieldName =
> ```

Just inline this and make this `repoIdFieldName`.  `snakeCaseFieldName` isn't a good name anyway because it doesn't tell you what it's the field name _of_.  As-is, with two separate variables, it almost feels like this is two separate things, but it's not; it's the same field name and you're only using the snake case version.
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/1354#-MkwbPJJCaULDL_-wKMT:-MkwbPJJCaULDL_-wKMU:bdw4isy) ([raw file](https://github.com/google/nomulus/blob/b09278d897877b7a43b9955c40aa631b55e96690/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L121)):*
> ```Java
>                       .createNativeQuery(
>                           String.format(
>                               "SELECT EXISTS (SELECT 1 FROM \"%s\" WHERE %s = :repoId AND"
> ```

In a sentence or two, can you explain what you're actually doing here?  I think you're (intending to) check if there's a row in the *History table for the EPP resource containing filled-out data, but that's not obvious either from the query or from the method name.  Which field exactly is a proxy for the rest of the data being present?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1354)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 105 at r1](https://reviewable.io/reviews/google/nomulus/1354#-MkwbGbu98-AIWOdzH2c:-MkwoH860KKxCypHeX1z:bhtvpgm) ([raw file](https://github.com/google/nomulus/blob/b09278d897877b7a43b9955c40aa631b55e96690/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L105)):*

I don't understand this method name.  At what point in time are you checking that this resource exists?  What does `AtPointInTime` have to do with anything that's going on here?
</blockquote></details>

yeah it's answering the question "are there any history objects containing the resource at that point in time?"

Updated name
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 114 at r1](https://reviewable.io/reviews/google/nomulus/1354#-Mkwa-Dx9ZzluuU2D-nM:-Mkwoih8DEHxbmwNEJzR:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b09278d897877b7a43b9955c40aa631b55e96690/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L114)):*

Just inline this and make this `repoIdFieldName`.  `snakeCaseFieldName` isn't a good name anyway because it doesn't tell you what it's the field name _of_.  As-is, with two separate variables, it almost feels like this is two separate things, but it's not; it's the same field name and you're only using the snake case version.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 121 at r1](https://reviewable.io/reviews/google/nomulus/1354#-MkwbPJJCaULDL_-wKMT:-Mkwov9z8EtyY69Qf_Df:bxjeqly) ([raw file](https://github.com/google/nomulus/blob/b09278d897877b7a43b9955c40aa631b55e96690/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L121)):*

In a sentence or two, can you explain what you're actually doing here?  I think you're (intending to) check if there's a row in the *History table for the EPP resource containing filled-out data, but that's not obvious either from the query or from the method name.  Which field exactly is a proxy for the rest of the data being present?
</blockquote></details>

Your assumption is correct. Any field that doesn't start with `history_` will only be filled out when the History object contains a representation of the resource at that point in time. We use `creation_time` here because it's as generic as you can get across all three types of EppResources. Added a comment.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1354#-:-MkxMrizEKkwT8MxysEw:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1341, opened 2021-09-24
Add handling for UpdateAutoTimestamp when not in a transaction

CydeWeys:
It's not clear why this is sometimes causing test flakes, but getting better
logging involved should help clear it up.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1341)

PR #1359, opened 2021-10-01
Add non-transactional fallback for create auto timestamp

CydeWeys:
This should never happen in production code (and if it does we'll see errors in
the logs), but it does appear to be happening in tests during the persistence of
initial entities in AppEngineExtension. PR #1341 (already merged) added this
same handling for update auto timestamps.

### DIFF
--- a/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java
+++ b/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java
@@ -13,9 +13,12 @@
 // limitations under the License.
 package google.registry.persistence.converter;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.persistence.transaction.TransactionManagerFactory.ofyTm;
+import static org.joda.time.DateTimeZone.UTC;
 
+import com.google.common.flogger.FluentLogger;
+import com.google.common.flogger.StackSize;
 import google.registry.model.CreateAutoTimestamp;
 import google.registry.util.DateTimeUtils;
 import java.sql.Timestamp;
@@ -31,13 +34,26 @@ import org.joda.time.DateTime;
 public class CreateAutoTimestampConverter
     implements AttributeConverter<CreateAutoTimestamp, Timestamp> {
 
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
   @Override
   @Nullable
   public Timestamp convertToDatabaseColumn(@Nullable CreateAutoTimestamp entity) {
     if (entity == null) {
       return null;
     }
-    DateTime dateTime = firstNonNull(entity.getTimestamp(), jpaTm().getTransactionTime());
+    DateTime dateTime = entity.getTimestamp();
+    if (dateTime == null) {
+      if (jpaTm().inTransaction()) {
+        dateTime = jpaTm().getTransactionTime();
+      } else {
+        logger.atSevere().withStackTrace(StackSize.MEDIUM).log(
+            "Failed to set automatic create timestamp because this wasn't called in a JPA"
+                + " transaction%s.",
+            ofyTm().inTransaction() ? " (but there is an open Ofy transaction)" : "");
+        dateTime = DateTime.now(UTC);
+      }
+    }
     return Timestamp.from(DateTimeUtils.toZonedDateTime(dateTime).toInstant());
   }
 


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwMGte1Rou3zLdml-G:bde6rkz) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*
> ```Java
>                 + " transaction%s.",
>             ofyTm().inTransaction() ? " (but there is an open Ofy transaction)" : "");
>         dateTime = DateTime.now(UTC);
> ```

It seems to me we should really be rethrowing in this case.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwO4F32GKG56occkXe:bg1sdhv) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

It seems to me we should really be rethrowing in this case.
</blockquote></details>

Rethrow what?  There's no exception here.  And the entire point of this PR is to NOT throw an exception here, which is what's causing all the failures.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwOeKKE_qpiyYULyY2:b-gzm9ei) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

Rethrow what?  There's no exception here.  And the entire point of this PR is to NOT throw an exception here, which is what's causing all the failures.
</blockquote></details>

"throwing", then.  I don't think we should mask this error.  If we hit this case, the code will either fail subsequently (due to a similar problem elsewhere) or pass with a potentially incorrect timestamp.  If this were to happen in production (unlikely, but not impossible) we would persist an incorrect timestamp.

That said, I am interested to see what will happen in the tests for this.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwRB0--m_YRWWqfqn4:b-dlmzaj) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

"throwing", then.  I don't think we should mask this error.  If we hit this case, the code will either fail subsequently (due to a similar problem elsewhere) or pass with a potentially incorrect timestamp.  If this were to happen in production (unlikely, but not impossible) we would persist an incorrect timestamp.

That said, I am interested to see what will happen in the tests for this.
</blockquote></details>

This error being thrown is why the tests are failing all the time, and I think it's an artifact of how `AppEngineExtension`'s setup is interacting with persistence callbacks on these entities.  I don't think it will occur in production at all (where we aren't faking the DB), but if it does, it will write an obvious message in the logs.  I'm actually not convinced that the code will fail subsequently.

If you've got a better idea I'm all ears.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1359#-Mkwl5FGAADRSnsomXaY:-Mkwl5FGAADRSnsomXaZ:b8qa3c3) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L46)):*
> ```Java
>     }
>     DateTime dateTime = entity.getTimestamp();
>     if (dateTime == null) {
> ```

Make this logic contingent upon "environment == unittest"
___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwkmAeBT1bZ7nBpVSI:b-fmxned) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

This error being thrown is why the tests are failing all the time, and I think it's an artifact of how `AppEngineExtension`'s setup is interacting with persistence callbacks on these entities.  I don't think it will occur in production at all (where we aren't faking the DB), but if it does, it will write an obvious message in the logs.  I'm actually not convinced that the code will fail subsequently.

If you've got a better idea I'm all ears.
</blockquote></details>

The best thing would be to actually identify the problem.  It's interesting that we're still seeing a test failure (and from this code, no less!).  But we're also only seeing one failure, and it looks like it's successfully hiding the issue in other cases.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwqTrD3BbTQ_iZIt8e:b-1ovmmk) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

The best thing would be to actually identify the problem.  It's interesting that we're still seeing a test failure (and from this code, no less!).  But we're also only seeing one failure, and it looks like it's successfully hiding the issue in other cases.
</blockquote></details>

Actually, scratch that last comment -- it seems to be logging but still failing?  Weird.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1359#-Mkwl5FGAADRSnsomXaY:-MkwuWwhB5wrUAqAb_YF:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L46)):*

Make this logic contingent upon "environment == unittest"
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-MkwuY0K2xWvEWPyJr_8:bt12bpa) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

Actually, scratch that last comment -- it seems to be logging but still failing?  Weird.
</blockquote></details>

What's logging but still failing?

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1359)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1359#-MkwMGtdDG27vuDzLmFT:-Mkx0bJo7Hllu9hwrhHN:b853nad) ([raw file](https://github.com/google/nomulus/blob/fdf8f2236f73c1cf28a1477a76311c5c53789630/core/src/main/java/google/registry/persistence/converter/CreateAutoTimestampConverter.java#L54)):*

What's logging but still failing?
</blockquote></details>

Nevermind, I think I must have somehow pulled the build results for another PR (it was logging about UpdateAutoTimestamp and failing in CreateAutoTimestamp).

CydeWeys:
This was a dead-end.

PR #1360, opened 2021-10-01
Dummy PR for investigating test flakiness.

mindhog:
This is now a dummy PR for debugging test flakiness.  Please ignore for purposes of code-review.

Original comment (for posterity):

This is a work-around for our sudden increase in test flakes.  It basically
just moves tests that have been observed to fail into the "fragile" set,
causing them to be run in isolation.  This trades off test time for less
flakiness.

This also fixes TransactionsReportingQueryBuilderTest, which wasn't working
in isolation because it didn't include the JPA extension.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1360)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -80,6 +80,9 @@ def fragileTestPatterns = [
 
     // This test seems to have some problems.
     "google/registry/reporting/icann/TransactionsReportingQueryBuilderTest.*",
+
+    // Moved here as we see cases of it being the first test that fails.
+    "google/registry/batch/ExpandRecurringBillingEventsActionTest.*",
 ] + dockerIncompatibleTestPatterns
 
 sourceSets {
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -79,8 +79,8 @@ def fragileTestPatterns = [
     "google/registry/flows/host/HostInfoFlowTest.*",
 
     // Tests that have started exhibiting flakiness (IllegalStateException in
-    // JpaTransactionManager.  JPA apparently stops thinking we're in a
-    // transacti
+    // JpaTransactionManager. Somehow our "inTransaction" flag is being
+    // cleared prematurely.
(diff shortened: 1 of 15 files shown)


core/build.gradle
```diff
@@ -77,6 +77,17 @@ def fragileTestPatterns = [
     // Currently changes a global configuration parameter that for some reason
     // results in timestamp inversions for other tests.  TODO(mmuller): fix.
     "google/registry/flows/host/HostInfoFlowTest.*",
+
+    // Tests that have started exhibiting flakiness (IllegalStateException in
+    // JpaTransactionManager.  JPA apparently stops thinking we're in a
+    // transacti
+    "google/registry/tools/UnlockDomainCommandTest.*",
```
CydeWeys, line 84: This seems to just be a randomly selected collection of tests to me, i.e. it's not really going to do anything unless you add MOST tests to fragile (and then the test suite will take forever to run).

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1360)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/build.gradle, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1360#-MkwOAzYQxVGvL5Sh-zM-r1-84:-MkwPw_32dis0JLamGT3:bo4xpdj) ([raw file](https://github.com/google/nomulus/blob/44bcbf1041d492dab4aa0880a8083e67afd2df7d/core/build.gradle#L84)):*

This seems to just be a randomly selected collection of tests to me, i.e. it's not really going to do anything unless you add MOST tests to fragile (and then the test suite will take forever to run).
</blockquote></details>

Not random.  These are tests that I have observed to fail.

This isn't a perfect approach, but if it works at least we can get unblocked.

PR #1375, opened 2021-10-05
Don't use Dagger to provide TransactionManager

CydeWeys:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1375)


CydeWeys:
Closing pull request -- we don't think this actually addresses any real issue.

PR #1378, opened 2021-10-07
Add reason and requestedByRegistrar to domain renew flow 

rachelguan:
When URS triggers the domain renew flow, it needs to indicate reason and whether or not it's requested by registrar. Currently, domain renew flow does not support <reason> and <requestedByRegistrar> and these two are being added here as optional parameters. However, if reason is specified, requestedByRegistrar (corresponding to --registrar_request) becomes required. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1378)


rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1378)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/RenewDomainCommand.java, line 97 at r2](https://reviewable.io/reviews/google/nomulus/1378#-MlkrzFt8uiH3KrSdMnB:-Mll4VaW7gQyud9nMtHr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5dfc9d4d2ed09c3a821608a16034d3d9f424ac2d/core/src/main/java/google/registry/tools/RenewDomainCommand.java#L97)):*

This should be `--registrar_request` and `--reason`
</blockquote></details>

Done.

PR #1383, opened 2021-10-08
Add a beam pipeline to create synthetic history entries in SQL

jianglai:
The logic is mostly lifted from CreateSyntheticHistoryEntriesAction. We
do not need to test for the existence of an embedded EPP resource in the
history entry before create a synthetic one because after
InitSqlPipeline runs it is guaranteed that no embedded resource exists.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1383)

### DIFF
--- a/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java
+++ b/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java
@@ -23,6 +23,8 @@ import google.registry.beam.common.RegistryPipelineOptions;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.config.RegistryConfig.ConfigModule;
 import google.registry.model.EppResource;
+import google.registry.model.UpdateAutoTimestamp;
+import google.registry.model.UpdateAutoTimestamp.DisableAutoUpdateResource;
 import google.registry.model.contact.ContactResource;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.host.HostResource;
@@ -90,16 +92,19 @@ public class CreateSyntheticHistoryEntriesPipeline implements Serializable {
                             .transact(
                                 () -> {
                                   EppResource eppResource = jpaTm().loadByKey(key);
-                                  jpaTm()
-                                      .put(
-                                          HistoryEntry.createBuilderForResource(eppResource)
-                                              .setRegistrarId(registryAdminRegistrarId)
-                                              .setBySuperuser(true)
-                                              .setRequestedByRegistrar(false)
-                                              .setModificationTime(jpaTm().getTransactionTime())
-                                              .setReason(HISTORY_REASON)
-                                              .setType(HistoryEntry.Type.SYNTHETIC)
-                                              .build());
+                                  try (DisableAutoUpdateResource disable =
+                                      UpdateAutoTimestamp.disableAutoUpdate()) {
+                                    jpaTm()
+                                        .put(
+                                            HistoryEntry.createBuilderForResource(eppResource)
+                                                .setRegistrarId(registryAdminRegistrarId)
+                                                .setBySuperuser(true)
+                                                .setRequestedByRegistrar(false)
+                                                .setModificationTime(jpaTm().getTransactionTime())
+                                                .setReason(HISTORY_REASON)
+                                                .setType(HistoryEntry.Type.SYNTHETIC)
+                                                .build());
+                                  }
                                 });
                         return null;
                       }));
--- a/core/src/test/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipelineTest.java
+++ b/core/src/test/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipelineTest.java
@@ -38,6 +38,7 @@ import google.registry.model.reporting.HistoryEntry;
 import google.registry.persistence.transaction.JpaTestExtensions;
 import google.registry.persistence.transaction.JpaTestExtensions.JpaIntegrationTestExtension;
 import google.registry.testing.DatastoreEntityExtension;
+import google.registry.testing.FakeClock;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -46,9 +47,11 @@ import org.junit.jupiter.api.extension.RegisterExtension;
 /** Unit tests for {@link CreateSyntheticHistoryEntriesPipeline}. */
 public class CreateSyntheticHistoryEntriesPipelineTest {
 
+  FakeClock clock = new FakeClock();
+
   @RegisterExtension
   JpaIntegrationTestExtension jpaEextension =
-      new JpaTestExtensions.Builder().buildIntegrationTestExtension();
+      new JpaTestExtensions.Builder().withClock(clock).buildIntegrationTestExtension();
 
   @RegisterExtension
   DatastoreEntityExtension datastoreEntityExtension =
@@ -71,6 +74,7 @@ public class CreateSyntheticHistoryEntriesPipelineTest {
         persistSimpleResource(
             newDomainBase("example.tld").asBuilder().setNameservers(host.createVKey()).build());
     contact = jpaTm().transact(() -> jpaTm().loadByKey(domain.getRegistrant()));
+    clock.advanceOneMilli();
   }
 
   @AfterEach


gbrodman (changes requested):
unit tests?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383#-:-MljqxuI2yF4Iq15DqUO:b250vud)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java, line 44 at r1](https://reviewable.io/reviews/google/nomulus/1383#-MljoIoC38Fp9lTk9vm3:-MljoIoD1fCB8r4gZvcJ:bqcwjhu) ([raw file](https://github.com/google/nomulus/blob/e691494603508f05d6e0a33511f48fafcb63f29b/core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java#L44)):*
> ```Java
>  * <p>The history entries in Datastore does not have the EPP resource embedded in them. Therefore
>  * after {@link google.registry.beam.initsql.InitSqlPipeline} runs, these fields will all be empty.
>  * This pipeline load all EPP resources and for each of them creates a synthetic history entry that
> ```

nit: s/load/loads/g
___
*[core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java, line 88 at r1](https://reviewable.io/reviews/google/nomulus/1383#-MljpPiB2EedmMBp10Kz:-MljpPiB2EedmMBp10L-:b39xfs7) ([raw file](https://github.com/google/nomulus/blob/e691494603508f05d6e0a33511f48fafcb63f29b/core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java#L88)):*
> ```Java
>           .apply(
>               String.format("Save a synthetic HistoryEntry for each %s", clazz),
>               MapElements.into(TypeDescriptor.of(Void.class))
> ```

MapElements seems like overkill, this should instead just be something like 

`.apply(ParDo.of(key -> ....))`

where the function that takes `key` as the input writes to SQL and returns a PDone 
___
*[core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java, line 95 at r1](https://reviewable.io/reviews/google/nomulus/1383#-MljoaBY4JS-cz32_Zp6:-MljoaBY4JS-cz32_Zp7:b-8177dm) ([raw file](https://github.com/google/nomulus/blob/e691494603508f05d6e0a33511f48fafcb63f29b/core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java#L95)):*
> ```Java
>                                 () -> {
>                                   EppResource eppResource = jpaTm().loadByKey(key);
>                                   jpaTm()
> ```

We should check SQL to see if a relevant entity already exists. If it does, we shouldn't save a new entity. That way, this job is idempotent and resilient to failures. 
___
*[core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java, line 117 at r1](https://reviewable.io/reviews/google/nomulus/1383#-Mljqpbo0_ftaGZh-7Ew:-Mljqpbo0_ftaGZh-7Ex:bievtg8) ([raw file](https://github.com/google/nomulus/blob/e691494603508f05d6e0a33511f48fafcb63f29b/core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java#L117)):*
> ```Java
> 
>     @Config("registryAdminClientId")
>     String getRegistryAdminRegistrarId();
> ```

probably not worth defining this separately if we can just use the built-in config component (though it means we'll, for now, have the non-ideal name `registryAdminClientId`

jianglai (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383#-:-Mlu0F-hCVXj3AKlzB9Q:b-896fix)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @weiminyu)

___
*[core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java, line 88 at r1](https://reviewable.io/reviews/google/nomulus/1383#-MljpPiB2EedmMBp10Kz:-MlqGFBI7OrD9eAor98L:bzc6mir) ([raw file](https://github.com/google/nomulus/blob/e691494603508f05d6e0a33511f48fafcb63f29b/core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java#L88)):*

I'm not aware of any performance benefit or penalty; I don't know enough about that. I was just thinking that a mapping, to me, implies that we return some result that that was use later, that the `via` function is just a transform to get something that we use later (rather than here where the `via` function is the primary task). 

Now that I think about it again I don't really have a problem with it though, just kinda musing. 
</blockquote></details>

Conceptually `via` where you'd provide the map function. The operation here is more accurately described as "consume" rather than "map", but Beam doesn't really have that notion as it assumes the terminal transform should be some sort of IO operation.

The most idiomatic pipeline would create these synthetic history entries and then pass them to a terminal IO stage (`RegistryJpaIO.Write`). But we don't have a good way to serialize arbitrary entities yet (we really should work on this @weiminyu as there are more and more use case for this), so we have to lump the "create synthetic entries" step with an ad-hoc JPA put. Using `RegistryJpaIO.Write` also improves performance because it batches the write.
___
*[core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java, line 50 at r2](https://reviewable.io/reviews/google/nomulus/1383#-MlqFd9l4f-F15m4YP3F:-MlqFpvbAqtOYysjJKCq:bb8bsnp) ([raw file](https://github.com/google/nomulus/blob/638f45d4e583c501e17e6f13058b61482e67d001/core/src/main/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipeline.java#L50)):*

i think a typo snuck through here 
</blockquote></details>

It's actually not a typo. `cSHE` is short for `createSyntheticHistoryEntries`. You do not have to spell out the entire task name. You can either just write a prefix of it, or the first letter of each word in the camelCase task name, as long as the result is unambiguous. 
___
*[core/src/test/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipelineTest.java, line 86 at r2](https://reviewable.io/reviews/google/nomulus/1383#-MlqERcR1_CFUBmEajEx:-MlqI31B7PgL5KmZvsVe:b-44zyqg) ([raw file](https://github.com/google/nomulus/blob/638f45d4e583c501e17e6f13058b61482e67d001/core/src/test/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipelineTest.java#L86)):*

Actually we may, in the pipeline, wish to disable the auto timestamp (using the `DisableAutoUpdateResource`) so that we can just compare the entities directly? 

(we don't do this in the mapreduce because we don't save the resource there, but it is disabled in the ReplayCommitLogsToSqlAction) 
</blockquote></details>

Unfortunately this does not work, the reason being that the timestamp is auto updated when the entity is persisted, if it is `null`, regardless of whether auto update is enabled or not:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/UpdateAutoTimestamp.java;l=60;bpv=1?q=updateautotimesta&ss=nomulus%2Fnomulus

So the entity persisted to the database will always have a non-null update timestamp, whereas the cloned entity from the embedded resource will not copy that field over.
___
*[core/src/main/java/google/registry/model/domain/DomainBase.java, line 220 at r2](https://reviewable.io/reviews/google/nomulus/1383#-MlqDzOr4PP8A-d4plyy:-MlqHeF-7Hsswv83R-H_:b5jnckt) ([raw file](https://github.com/google/nomulus/blob/638f45d4e583c501e17e6f13058b61482e67d001/core/src/main/java/google/registry/model/domain/DomainBase.java#L220)):*

nit, i think this was accidental :) 
</blockquote></details>

Oops, sorry.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[core/src/test/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipelineTest.java, line 86 at r2](https://reviewable.io/reviews/google/nomulus/1383#-MlqERcR1_CFUBmEajEx:-MlvMku40ihATByLX4oh:bn8nuc4) ([raw file](https://github.com/google/nomulus/blob/638f45d4e583c501e17e6f13058b61482e67d001/core/src/test/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipelineTest.java#L86)):*

In `beforeEach()` we persisted a bunch of entities (domain, contact, host) and read them back. They all have non-null timestamps regardless of whether we disable auto update. When you run the pipeline the `DomainContent`entities embedded in `DomainHistory` do have the timestamp I think, but the `DomainBase` we get from the `DomainHistory` by calling `DomainHistory.getResourceAtPointInTime()` does not have the timestamp because it calls `DomainBase.Builder.copyFrom()`, which does not copy over the timestamp that the embedded `DomainContent` has.
</blockquote></details>

right, so I think this would work with the same update timestamp if we called e.g. `DomainHistory::getDomainContent` directly right? 

probably not really worth it, just want to make sure. 

Either way we should still have the disable-update-timestamp wrapper in the action

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[core/src/test/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipelineTest.java, line 86 at r2](https://reviewable.io/reviews/google/nomulus/1383#-MlqERcR1_CFUBmEajEx:-Mm3kZHN6ZHtbd8zNlwG:b-mm57d0) ([raw file](https://github.com/google/nomulus/blob/638f45d4e583c501e17e6f13058b61482e67d001/core/src/test/java/google/registry/beam/common/CreateSyntheticHistoryEntriesPipelineTest.java#L86)):*

For posterity / writing, I'd be surprised if it's not updating the update auto timestamp on the resource. It looks like by default we use a fake clock that's never updated (in JpaTestExtensions) so even if it is updated, it'll contain the same value. Let's try updating the fake clock and see if this still works 
</blockquote></details>

You are right. Disabled auto update in the pipeline. Thanks for the good catch!

gbrodman (approved):
> Reviewed 2 of 3 files at r7, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383)* status: all files reviewed, 3 unresolved discussions

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java` line 71 at r9](https://reviewable.io/reviews/google/nomulus/1383#-NSM6SfgBvqewiNRRR6l:-NSM6SfgBvqewiNRRR6m:b89a8op) ([raw file](https://github.com/google/nomulus/blob/d98b0ce40c15ce7ff9756e64de5eee1153fe4bfa/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java#L71)):*
> ```Java
> 
>   private static final ImmutableList<Class<? extends EppResource>> EPP_RESOURCE_CLASSES =
>       ImmutableList.of(DomainBase.class, ContactResource.class, HostResource.class);
> ```

No sense in writing these out for anything but contacts in this go around.
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java` line 74 at r9](https://reviewable.io/reviews/google/nomulus/1383#-NSM69TfCwHtfedu2A7Y:-NSM69TfCwHtfedu2A7Z:bidosjk) ([raw file](https://github.com/google/nomulus/blob/d98b0ce40c15ce7ff9756e64de5eee1153fe4bfa/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java#L74)):*
> ```Java
> 
>   private static final String HISTORY_REASON =
>       "Backfill EppResource history objects after initial backup to SQL";
> ```

Need to update this reason message to be correct for the current use case.
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java` line 103 at r9](https://reviewable.io/reviews/google/nomulus/1383#-NSM6Csg4WgqBees2ZDr:-NSM6Csg4WgqBees2ZDs:bfsdo50) ([raw file](https://github.com/google/nomulus/blob/d98b0ce40c15ce7ff9756e64de5eee1153fe4bfa/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java#L103)):*
> ```Java
>                                                 .setBySuperuser(true)
>                                                 .setRequestedByRegistrar(false)
>                                                 .setModificationTime(jpaTm().getTransactionTime())
> ```

This isn't going to work to fix RDE.  The history entries need to be older than the deposits that need to be regenerated.  I would suggest setting it to the last EPP update time (plus 1 ms) of the Contact.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1383)* status: all files reviewed, 3 unresolved discussions

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java` line 103 at r9](https://reviewable.io/reviews/google/nomulus/1383#-NSM6Csg4WgqBees2ZDr:-NSM713q995Z_YKEeaZP:bo12imu) ([raw file](https://github.com/google/nomulus/blob/d98b0ce40c15ce7ff9756e64de5eee1153fe4bfa/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesPipeline.java#L103)):*

This isn't going to work to fix RDE.  The history entries need to be older than the deposits that need to be regenerated.  I would suggest setting it to the last EPP update time (plus 1 ms) of the Contact.
</blockquote></details>

For this specifically, I was using watermark time, so 2023-04-04.

PR #1384, opened 2021-10-08
Retry datastore transactions during replay

mindhog:
It looks like the datastore failures we've been seeing may very well have been
due to transient failures on the part of the test datastore.  Wrap the replay
transactions in a 5x retrier to remedy.

Note that this problem would not affect the actual replay, nor does this
solution: the actual replay uses a cron job and if a specific transaction
fails it will be retried on the next cycle (though we may consider adding
retry for that, if it's not already present).

### DIFF
--- a/core/src/test/java/google/registry/testing/ReplayExtension.java
+++ b/core/src/test/java/google/registry/testing/ReplayExtension.java
@@ -39,6 +39,8 @@ import google.registry.persistence.transaction.Transaction.Mutation;
 import google.registry.persistence.transaction.Transaction.Update;
 import google.registry.persistence.transaction.TransactionEntity;
 import google.registry.util.RequestStatusChecker;
+import google.registry.util.Retrier;
+import google.registry.util.SystemSleeper;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


core/src/test/java/google/registry/testing/ReplayExtension.java
```diff
@@ -197,6 +199,7 @@ private void replayToSql() {
   }
 
   private void replayToOfy() {
+    Retrier retrier = new Retrier(new SystemSleeper(), 5);
```
CydeWeys, line 202: I think you want this to be an instance field on `ReplayExtension` (like the `InjectExtension`/`Clock`) rather than creating it from scratch for every single test.

CydeWeys (commented):
¯\\\_(ツ)\_/¯

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1384#-:-MlW9zHR3AGXlpyCYZRM:bkglepv)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

PR #1395, opened 2021-10-18
Use READ_COMMITTED serialization level in CreateSyntheticHEA

gbrodman:
I observed an instance in which a couple queries from this action were,
for whatever reason, hanging around as idle for >30 minutes. Assuming
the behavior that we saw before where "an open idle serializable
transaction means all pg read-locks stick around forever" still holds,
that's the reason why the amount of read-locks in use spirals out of
control.

I'm not sure why those queries aren't timing out, but that's a separate
issue.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1395)

### DIFF
--- a/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java
+++ b/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java
@@ -108,6 +108,16 @@ public class CreateSyntheticHistoryEntriesAction implements Runnable {
     return jpaTm()
         .transact(
             () -> {
+              // Use READ COMMITTED isolation level so that any long-living queries don't cause
+              // collection of predicate locks to spiral out of control (as would happen with a
+              // SERIALIZABLE isolation level)
+              //
+              // NB: setting the isolation level inside the transaction only works for Postgres and
+              // will be reverted to the default once the transaction is committed.
+              jpaTm()
+                  .getEntityManager()
+                  .createNativeQuery("SET TRANSACTION ISOLATION LEVEL READ COMMITTED")
+                  .executeUpdate();
               // The class we're searching from is based on which parent type (e.g. Domain) we have
               Class<? extends HistoryEntry> historyClass =
                   getHistoryClassFromParent(resource.getClass());


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1395)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @weiminyu)

___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 120 at r1](https://reviewable.io/reviews/google/nomulus/1395#-MmLO3tuCamSFidhspUM:-MmLO3tv1oQGZOjx61lp:b18x0u0) ([raw file](https://github.com/google/nomulus/blob/b01d9c9b26c988def6bff1d2f07cc2f4a8642ba8/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L120)):*
> ```
> read-locks
> ```
official name is predicate lock
___
*[core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1395#-MmLOEx7EEHBNfni0LJV:-MmLOEx7EEHBNfni0LJW:b-iksks) ([raw file](https://github.com/google/nomulus/blob/b01d9c9b26c988def6bff1d2f07cc2f4a8642ba8/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticHistoryEntriesAction.java#L124)):*
> ```
> createNativeQuery("SET TRANSACTION ISOLATION LEVEL READ COMMITTED")
> ```
SET TRANSACTION must happen before any query is executed. We may want to make this statement forward to make it obvious and future-proof.  

Let's document that this implementation is postgresql -specific. E.g., in mysql SET TRANSACTION must happen before the transaction starts.

Let's also document that the isolation level on the connection automatically reverts to the default when the transaction commits/aborts so no cleanup necessary.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1395)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1398, opened 2021-10-20
Alt entity model for fast JPA bulk query

weiminyu:
Define an alternative entity model that allows fast bulk loading of multi-level entities, DomainBase and DomainHistory,
from Cloud SQL.

Verified in QA with production data.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1398)

### DIFF
--- a/core/src/test/java/google/registry/model/ImmutableObjectSubject.java
+++ b/core/src/test/java/google/registry/model/ImmutableObjectSubject.java
@@ -412,6 +412,7 @@ public final class ImmutableObjectSubject extends Subject {
     // don't use ImmutableMap or a stream->collect model since we can have nulls
     Map<Field, Object> result = new LinkedHashMap<>();
     for (Map.Entry<Field, Object> entry : originalFields.entrySet()) {
+      // TODO(b/203685960): filter by @DoNotCompare instead.
       if (entry.getKey().isAnnotationPresent(ImmutableObject.Insignificant.class)) {
         continue;
       }
@@ -429,6 +430,7 @@ public final class ImmutableObjectSubject extends Subject {
     // don't use ImmutableMap or a stream->collect model since we can have nulls
     Map<Field, Object> result = new LinkedHashMap<>();
     for (Map.Entry<Field, Object> entry : originalFields.entrySet()) {
+      // TODO(b/203685960): filter by @DoNotCompare instead.
       if (!entry.getKey().isAnnotationPresent(annotation)
           && !entry.getKey().isAnnotationPresent(ImmutableObject.Insignificant.class)) {
 


weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1398)* status: 21 of 26 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/bulkquery/DomainBaseLite.java, line 28 at r1](https://reviewable.io/reviews/google/nomulus/1398#-MmSvHn73ixMYPmb-D4f:-MmTiKJ71LapbfNPMsEr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0508dfd7c910c409e4045bb1a09490c0fc2a846c/core/src/main/java/google/registry/model/bulkquery/DomainBaseLite.java#L28)):*

Maybe mention that this works because the associated collections are loaded from DomainBase?
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/bulkquery/TestSetupHelper.java, line 206 at r1](https://reviewable.io/reviews/google/nomulus/1398#-MmTHD310nxJKmc8c9at:-MmTjKI4EsQ7u2vGEiue:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0508dfd7c910c409e4045bb1a09490c0fc2a846c/core/src/test/java/google/registry/model/bulkquery/TestSetupHelper.java#L206)):*

Hah, I don't think I've heard the term "upsert" before :-)
Optional: you may want to consder "put" instead, as that's what we've landed on for TM (I suppose we might have used "upsert" if we had known of it)
</blockquote></details>

Done.
___
*[core/src/test/resources/google/registry/model/schema.txt, line 648 at r1](https://reviewable.io/reviews/google/nomulus/1398#-MmTYrT03HRo535E_J26:-MmTlnpH7RnM277ng1Zx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0508dfd7c910c409e4045bb1a09490c0fc2a846c/core/src/test/resources/google/registry/model/schema.txt#L648)):*

Should these be at-Ignore'd?
</blockquote></details>

Done.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1398)* status: 23 of 26 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/bulkquery/BulkQueryEntities.java, line 63 at r3](https://reviewable.io/reviews/google/nomulus/1398#-MmTr4al0LLFv6ZLb6EG:-MmU-UTDCy3f2Vk_uJ9G:b-1ebn78) ([raw file](https://github.com/google/nomulus/blob/b0b558380bb35af6e50719b0d0c1dceaa4b173d2/core/src/main/java/google/registry/model/bulkquery/BulkQueryEntities.java#L63)):*

Why does this only include the hosts and not some of the other embedded classes like GracePeriod?

edit: oh is it just because this is only a join table between Domain<->Host? Not sure how that would cause this to happen exactly 
</blockquote></details>

GracePeriod, along with the other embedded classes, are already JPA entities. To make them work with in both models, we only need to make the id values public, or in the case of DomainTransactionRecord, expose the columns already in the table (repoId and historyRevisionId) in the Java class.
___
*[core/src/main/java/google/registry/model/bulkquery/DomainHistoryLite.java, line 119 at r3](https://reviewable.io/reviews/google/nomulus/1398#-MmTs7xw5TYklmzeuJdc:-MmTzI4C37JDjl8EH_EJ:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/b0b558380bb35af6e50719b0d0c1dceaa4b173d2/core/src/main/java/google/registry/model/bulkquery/DomainHistoryLite.java#L119)):*

nit is that this can probably be one else if statement
</blockquote></details>

Done
___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 415 at r1](https://reviewable.io/reviews/google/nomulus/1398#-MmTH08pBbegWxcHm13p:-MmTzoLT4gQR7xP9TvOA:bhd1udm) ([raw file](https://github.com/google/nomulus/blob/0508dfd7c910c409e4045bb1a09490c0fc2a846c/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L415)):*

I wondered about this also, and had to refresh myself on the semantics of these.  DoNotCompare is strictly for comparison across databases -- I don't think it's relevant here?
</blockquote></details>

Insignificant is the only annotation considered by ImmutableObject.equals(), and it covers Datastore objects too.

It is a bit risky to change the implementation of equals in this PR.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1398)* status: 25 of 26 files reviewed, 4 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/test/java/google/registry/model/ImmutableObjectSubject.java, line 415 at r1](https://reviewable.io/reviews/google/nomulus/1398#-MmTH08pBbegWxcHm13p:-MmUJnPt83EvMRlzpTAz:brr7xjb) ([raw file](https://github.com/google/nomulus/blob/0508dfd7c910c409e4045bb1a09490c0fc2a846c/core/src/test/java/google/registry/model/ImmutableObjectSubject.java#L415)):*

yeah maybe we should (not necessarily in this PR) change the documentation for Insignificant then. It seems to only discuss immutability / Hibernate 
</blockquote></details>

DoNotCompare is indeed more appropriate, but b/203685960 prevents it from working properly.

Opened that bug to follow up later.

PR #1404, opened 2021-11-02
Fix InitSqlPipeline regarding synthesized history

weiminyu:
There are a few bad domains in Datastore that we hardcoded to ignore
during SQL population. They didn't have history so we didn't try to
filter when writing history.

Recently we created synthesized history for domains, including the bad
domains. Now we need to filter History entries.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1404)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1404)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1406, opened 2021-11-02
Test vkey behaviors when in a task queue 

rachelguan:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1406)

### DIFF
--- a/core/src/test/java/google/registry/persistence/VKeyTest.java
+++ b/core/src/test/java/google/registry/persistence/VKeyTest.java
@@ -13,17 +13,17 @@
 // limitations under the License.
 package google.registry.persistence;
 
+import static com.google.appengine.api.taskqueue.QueueFactory.getQueue;
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.Truth8.assertThat;
 import static google.registry.testing.DatabaseHelper.newDomainBase;
 import static google.registry.testing.DatabaseHelper.persistActiveContact;
+import static google.registry.testing.TaskQueueHelper.assertTasksEnqueued;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
-import com.google.cloud.tasks.v2.Task;
-import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
+import com.google.appengine.api.taskqueue.TaskOptions;
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.LinkedListMultimap;
 import com.googlecode.objectify.Key;
 import com.googlecode.objectify.annotation.Entity;
 import google.registry.model.billing.BillingEvent.OneTime;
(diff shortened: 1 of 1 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1406)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/persistence/VKeyTest.java, line 346 at r1](https://reviewable.io/reviews/google/nomulus/1406#-MnbdwCW321TmKRYS7Oe:-MnbdwCX4jT7ywuUySUT:bvbx5t1) ([raw file](https://github.com/google/nomulus/blob/f1f664cf5ae644236460a9cd29794a6f28e40add/core/src/test/java/google/registry/persistence/VKeyTest.java#L346)):*
> ```Java
>                             .split(tasks.get(0).getAppEngineHttpRequest().getBody().toStringUtf8()))
>                     .get("vkey")))
>         .isEqualTo(vkey);
> ```

Did `TaskMatcher` not work out for this?

This approach is a less satisfying test because it's not using the same request processing code as our system (see [RefreshDnsOnHostRenameAction](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java#L354), for example, which extracts parameters from a TaskHandle).

That said, it's not clear how to do this using the task helper.  Maybe TaskOption is the better approach after all?  We can construct a TaskHandle from one of those.

In the worst case, we should at least use UriParameters.parse() to extract parameters from the body.  See  [CloudTasksHelper](https://github.com/google/nomulus/blob/master/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L238).

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1406)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/persistence/VKeyTest.java, line 346 at r1](https://reviewable.io/reviews/google/nomulus/1406#-MnbdwCW321TmKRYS7Oe:-MngLpQFEjYodMMQkN3z:b-7km50e) ([raw file](https://github.com/google/nomulus/blob/f1f664cf5ae644236460a9cd29794a6f28e40add/core/src/test/java/google/registry/persistence/VKeyTest.java#L346)):*

Did `TaskMatcher` not work out for this?

This approach is a less satisfying test because it's not using the same request processing code as our system (see [RefreshDnsOnHostRenameAction](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/batch/RefreshDnsOnHostRenameAction.java#L354), for example, which extracts parameters from a TaskHandle).

That said, it's not clear how to do this using the task helper.  Maybe TaskOption is the better approach after all?  We can construct a TaskHandle from one of those.

In the worst case, we should at least use UriParameters.parse() to extract parameters from the body.  See  [CloudTasksHelper](https://github.com/google/nomulus/blob/master/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L238).
</blockquote></details>

TaskMatcher is not needed if Cloud Tasks APIs are used. However, using Cloud Tasks APIs currently does not allow us to extract parameters . We would have to explicitly create a method to create a map from getBody() string. 

If we want to use TaskHandle, then we have to use TaskOption, which is something we want to move away from. Because the two vkey methods being tested were written to handle existing ofykey only vkey behaviors, we will use TaskOption objects to test the behavior as well. 

Added the comment and modified the test cases .

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1406)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

PR #1408, opened 2021-11-05
Change TaskQueueUtils to CloudTaskUtils in CommitLogFanoutAction

rachelguan:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1408)

### DIFF
--- a/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java
+++ b/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java
@@ -14,16 +14,20 @@
 
 package google.registry.cron;
 
-import static com.google.appengine.api.taskqueue.QueueFactory.getQueue;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
-import com.google.appengine.api.taskqueue.Queue;
-import com.google.appengine.api.taskqueue.TaskOptions;
+import com.google.cloud.tasks.v2.Task;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.Multimap;
+import com.google.protobuf.Timestamp;
 import google.registry.model.ofy.CommitLogBucket;
 import google.registry.request.Action;
+import google.registry.request.Action.Service;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
-import google.registry.util.TaskQueueUtils;
-import java.time.Duration;
+import google.registry.util.Clock;
+import google.registry.util.CloudTasksUtils;
+import java.time.Instant;
 import java.util.Optional;
 import java.util.Random;
 import javax.inject.Inject;
@@ -40,23 +44,41 @@ public final class CommitLogFanoutAction implements Runnable {
 
   private static final Random random = new Random();
 
-  @Inject TaskQueueUtils taskQueueUtils;
+  @Inject Clock clock;
+  @Inject CloudTasksUtils cloudTasksUtils;
+
   @Inject @Parameter("endpoint") String endpoint;
   @Inject @Parameter("queue") String queue;
   @Inject @Parameter("jitterSeconds") Optional<Integer> jitterSeconds;
   @Inject CommitLogFanoutAction() {}
 
+
+
   @Override
   public void run() {
-    Queue taskQueue = getQueue(queue);
     for (int bucketId : CommitLogBucket.getBucketIds()) {
-      long delay =
-          jitterSeconds.map(i -> random.nextInt((int) Duration.ofSeconds(i).toMillis())).orElse(0);
-      TaskOptions taskOptions =
-          TaskOptions.Builder.withUrl(endpoint)
-              .param(BUCKET_PARAM, Integer.toString(bucketId))
-              .countdownMillis(delay);
-      taskQueueUtils.enqueue(taskQueue, taskOptions);
+      cloudTasksUtils.enqueue(
+          queue, createTask(ImmutableMultimap.of(BUCKET_PARAM, Integer.toString(bucketId))));
     }
+  }
+
+  private Task createTask(Multimap<String, String> params) {
+    Instant scheduleTime =
+        Instant.ofEpochMilli(
+            clock
+                .nowUtc()
+                .plusMillis(
+                    jitterSeconds
+                        .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
+                        .orElse(0))
+                .getMillis());
+    return Task.newBuilder(
+            CloudTasksUtils.createGetTask(endpoint, Service.BACKEND.toString(), params))
+        .setScheduleTime(
+            Timestamp.newBuilder()
+                .setSeconds(scheduleTime.getEpochSecond())
+                .setNanos(scheduleTime.getNano())
+                .build())
+        .build();
   }
 }
--- a/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java
+++ b/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java
@@ -14,12 +14,7 @@
 
 package google.registry.cron;
 
-import static java.util.concurrent.TimeUnit.SECONDS;
-
-import com.google.cloud.tasks.v2.Task;
 import com.google.common.collect.ImmutableMultimap;
-import com.google.common.collect.Multimap;
-import com.google.protobuf.Timestamp;
 import google.registry.model.ofy.CommitLogBucket;
 import google.registry.request.Action;
 import google.registry.request.Action.Service;
@@ -27,9 +22,7 @@ import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
 import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
-import java.time.Instant;
 import java.util.Optional;
-import java.util.Random;
 import javax.inject.Inject;
 
 /** Action for fanning out cron tasks for each commit log bucket. */
@@ -41,8 +34,6 @@ import javax.inject.Inject;
 public final class CommitLogFanoutAction implements Runnable {
 
   public static final String BUCKET_PARAM = "bucket";
-
-  private static final Random random = new Random();
 
   @Inject Clock clock;
   @Inject CloudTasksUtils cloudTasksUtils;
@@ -58,27 +49,13 @@ public final class CommitLogFanoutAction implements Runnable {
   public void run() {
     for (int bucketId : CommitLogBucket.getBucketIds()) {
       cloudTasksUtils.enqueue(
-          queue, createTask(ImmutableMultimap.of(BUCKET_PARAM, Integer.toString(bucketId))));
+          queue,
+          CloudTasksUtils.createGetTask(
+              endpoint,
+              Service.BACKEND.toString(),
+              ImmutableMultimap.of(BUCKET_PARAM, Integer.toString(bucketId)),
+              clock,
+              jitterSeconds));
     }
-  }
-
-  private Task createTask(Multimap<String, String> params) {
-    Instant scheduleTime =
-        Instant.ofEpochMilli(
-            clock
-                .nowUtc()
-                .plusMillis(
-                    jitterSeconds
-                        .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
-                        .orElse(0))
-                .getMillis());
-    return Task.newBuilder(
-            CloudTasksUtils.createGetTask(endpoint, Service.BACKEND.toString(), params))
-        .setScheduleTime(
-            Timestamp.newBuilder()
-                .setSeconds(scheduleTime.getEpochSecond())
-                .setNanos(scheduleTime.getNano())
-                .build())
-        .build();
   }
 }
--- a/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java
+++ b/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java
@@ -50,7 +50,7 @@ public final class CommitLogFanoutAction implements Runnable {
     for (int bucketId : CommitLogBucket.getBucketIds()) {
       cloudTasksUtils.enqueue(
           queue,
-          CloudTasksUtils.createGetTask(
+          CloudTasksUtils.createPostTask(
               endpoint,
               Service.BACKEND.toString(),
               ImmutableMultimap.of(BUCKET_PARAM, Integer.toString(bucketId)),
--- a/core/src/test/java/google/registry/cron/CommitLogFanoutActionTest.java
+++ b/core/src/test/java/google/registry/cron/CommitLogFanoutActionTest.java
@@ -15,14 +15,13 @@
 package google.registry.cron;
 
 import static google.registry.cron.CommitLogFanoutAction.BUCKET_PARAM;
-import static google.registry.testing.TaskQueueHelper.assertTasksEnqueued;
 
 import com.google.common.base.Joiner;
 import google.registry.model.ofy.CommitLogBucket;
 import google.registry.testing.AppEngineExtension;
-import google.registry.testing.TaskQueueHelper.TaskMatcher;
-import google.registry.util.Retrier;
-import google.registry.util.TaskQueueUtils;
+import google.registry.testing.CloudTasksHelper;
+import google.registry.testing.CloudTasksHelper.TaskMatcher;
+import google.registry.testing.FakeClock;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -34,6 +33,7 @@ class CommitLogFanoutActionTest {
 
   private static final String ENDPOINT = "/the/servlet";
   private static final String QUEUE = "the-queue";
+  private final CloudTasksHelper cloudTasksHelper = new CloudTasksHelper();
 
   @RegisterExtension
   final AppEngineExtension appEngineExtension =
@@ -54,15 +54,16 @@ class CommitLogFanoutActionTest {
   @Test
   void testSuccess() {
     CommitLogFanoutAction action = new CommitLogFanoutAction();
-    action.taskQueueUtils = new TaskQueueUtils(new Retrier(null, 1));
+    action.cloudTasksUtils = cloudTasksHelper.getTestCloudTasksUtils();
     action.endpoint = ENDPOINT;
     action.queue = QUEUE;
     action.jitterSeconds = Optional.empty();
+    action.clock = new FakeClock();
     action.run();
     List<TaskMatcher> matchers = new ArrayList<>();
     for (int bucketId : CommitLogBucket.getBucketIds()) {
       matchers.add(new TaskMatcher().url(ENDPOINT).param(BUCKET_PARAM, Integer.toString(bucketId)));
     }
-    assertTasksEnqueued(QUEUE, matchers);
+    cloudTasksHelper.assertTasksEnqueued(QUEUE, matchers);
   }
 }
--- a/util/src/main/java/google/registry/util/CloudTasksUtils.java
+++ b/util/src/main/java/google/registry/util/CloudTasksUtils.java
@@ -94,7 +94,7 @@ public class CloudTasksUtils implements Serializable {
    *     Queue API if no service is specified, the service which enqueues the task will be used to
    *     process the task. Cloud Tasks API does not support this feature so the service will always
    *     needs to be explicitly specified.
-   * @param params A multi-map of URL query parameters. Duplicate keys are saved as is, and it is up
+   * @param params a multi-map of URL query parameters. Duplicate keys are saved as is, and it is up
    *     to the server to process the duplicate keys.
    * @return the enqueued task.
    * @see <a
@@ -135,6 +135,24 @@ public class CloudTasksUtils implements Serializable {
     return Task.newBuilder().setAppEngineHttpRequest(requestBuilder.build()).build();
   }
 
+  /**
+   * Create a {@link Task} to be enqueued.
+   *
+   * @param path the relative URI (staring with a slash and ending without one).
+   * @param method the HTTP method to be used for the request, only GET and POST are supported.
+   * @param service the App Engine service to route the request to. Note that with App Engine Task
+   *     Queue API if no service is specified, the service which enqueues the task will be used to
+   *     process the task. Cloud Tasks API does not support this feature so the service will always
+   *     needs to be explicitly specified.
+   * @param params a multi-map of URL query parameters. Duplicate keys are saved as is, and it is up
+   *     to the server to process the duplicate keys.
+   * @param clock a source of time.
+   * @param jitterSeconds the number of seconds that a task is randomly delayed up to.
+   * @return the enqueued task.
+   * @see <a
+   *     href=ttps://cloud.google.com/appengine/docs/standard/java/taskqueue/push/creating-tasks#target>Specifyinig
+   *     the worker service</a>
+   */
   private static Task createTask(
       String path,
       HttpMethod method,
@@ -142,14 +160,14 @@ public class CloudTasksUtils implements Serializable {
       Multimap<String, String> params,
       Clock clock,
       Optional<Integer> jitterSeconds) {
+    if ((jitterSeconds.isPresent() && jitterSeconds.get() <= 0) || jitterSeconds.isEmpty()) {
+      return createTask(path, method, service, params);
+    }
     Instant scheduleTime =
         Instant.ofEpochMilli(
             clock
                 .nowUtc()
-                .plusMillis(
-                    jitterSeconds
-                        .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
-                        .orElse(0))
+                .plusMillis(random.nextInt((int) SECONDS.toMillis(jitterSeconds.get())))
                 .getMillis());
     return Task.newBuilder(createTask(path, method, service, params))
         .setScheduleTime(
--- a/util/src/main/java/google/registry/util/CloudTasksUtils.java
+++ b/util/src/main/java/google/registry/util/CloudTasksUtils.java
@@ -136,7 +136,7 @@ public class CloudTasksUtils implements Serializable {
   }
 
   /**
-   * Create a {@link Task} to be enqueued.
+   * Create a {@link Task} to be enqueued with a random delay up to {@code jitterSeconds}.
    *
    * @param path the relative URI (staring with a slash and ending without one).
    * @param method the HTTP method to be used for the request, only GET and POST are supported.
@@ -160,7 +160,7 @@ public class CloudTasksUtils implements Serializable {
       Multimap<String, String> params,
       Clock clock,
       Optional<Integer> jitterSeconds) {
-    if ((jitterSeconds.isPresent() && jitterSeconds.get() <= 0) || jitterSeconds.isEmpty()) {
+    if (jitterSeconds.isEmpty() || jitterSeconds.get() <= 0) {
       return createTask(path, method, service, params);
     }
     Instant scheduleTime =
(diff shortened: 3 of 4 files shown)


jianglai (changes requested):
The PR name should be consistent with the others, i. e. "Change TaskQueueUtils to CloudTasksUtils ...".

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408#-:-Mns3B_s0q7MiZKfHhJ9:b-2zjunj)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/cron/CommitLogFanoutAction.java, line 69 at r1](https://reviewable.io/reviews/google/nomulus/1408#-Mns5ZSD5KIzkAl8M7Xm:-Mns5ZSD5KIzkAl8M7Xn:b8n5ubi) ([raw file](https://github.com/google/nomulus/blob/39aaa0f53000a09b9cc0c7cda8c4561578893047/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java#L69)):*
It looks like this pattern is used more than once (in `TldFanoutAction` as well at least). Let's extract it to a common method so it's easier to use:

`CloudTasksUtils.createTask(path, method, service, params, clock, jitterSecond)`.
<details><summary><i alt="summary">Quoted 9 lines of code…</i></summary>

> ```
>     Instant scheduleTime =
>         Instant.ofEpochMilli(
>             clock
>                 .nowUtc()
>                 .plusMillis(
>                     jitterSeconds
>                         .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
>                         .orElse(0))
>                 .getMillis());
> ```
</details>

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/main/java/google/registry/cron/CommitLogFanoutAction.java, line 69 at r1](https://reviewable.io/reviews/google/nomulus/1408#-Mns5ZSD5KIzkAl8M7Xm:-Mo-CMD_8G8myqgg7KUF:b-z7wxp0) ([raw file](https://github.com/google/nomulus/blob/39aaa0f53000a09b9cc0c7cda8c4561578893047/core/src/main/java/google/registry/cron/CommitLogFanoutAction.java#L69)):*

It looks like this pattern is used more than once (in `TldFanoutAction` as well at least). Let's extract it to a common method so it's easier to use:

`CloudTasksUtils.createTask(path, method, service, params, clock, jitterSecond)`.

> ```
>     Instant scheduleTime =
>         Instant.ofEpochMilli(
>             clock
>                 .nowUtc()
>                 .plusMillis(
>                     jitterSeconds
>                         .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
>                         .orElse(0))
>                 .getMillis());
> ```

</blockquote></details>

I noticed that as well.. And it seems like I would have to add some other common methods as well in CloudTasksUtils. For now, I will add a createTask() that takes jitterSecond and clock and I'm going to add a variation that takes different headers in the upcoming PRs.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 138 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5ia0h4g3nuPXge7z6:-Mo5ia0h4g3nuPXge7z7:byuvcn) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L138)):*
> ```
> private static Task createTask
> ```
This warrants a javadoc. What jitterSeconds does is not obvious from the function signature alone.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5j8x4CJeByMBYoqpu:-Mo5j8x4CJeByMBYoqpv:b-iaj62j) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L144)):*
> ```
> Optional<Integer> jitterSeconds
> ```
What's the point of passing an optional here? We should call this method when jitterSeconds is non-zero, right?
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 145 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5iYXwBuSRU1Y5aG_5:-Mo5iYXwBuSRU1Y5aG_6:b-615vt0) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L145)):*
> ```Java
>       Clock clock,
>       Optional<Integer> jitterSeconds) {
> ```

Extra whitespace.
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 91 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5jVAH2UcoiXfyn2PQ:-Mo5jVAH2UcoiXfyn2PR:b-nnvlts) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L91)):*
> ```
>     // epoch time 1636347600 is equivalent to the current clock up to seconds.
>     // no assertion for nano seconds because getNanos() only accesses the underlying
>     // nanosecond-adjustment of the instant.
> ```
I don't quite understand the comment here, the scheduled time is a random number, isn't it? So we can only assert that is is not zero (i. e. it is explicitly set). If you want to be precise you can assert that the scheduled time is within the current time + jitterSeconds.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 145 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5iYXwBuSRU1Y5aG_5:-Mo61SHs427mUgbPlBgS:brpmqc3) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L145)):*

Extra whitespace.
</blockquote></details>

removed.
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 91 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5jVAH2UcoiXfyn2PQ:-Mo61TZW5x5BpW-rXyX7:b-kzvkhu) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L91)):*

> ```
>     // epoch time 1636347600 is equivalent to the current clock up to seconds.
>     // no assertion for nano seconds because getNanos() only accesses the underlying
>     // nanosecond-adjustment of the instant.
> ```
I don't quite understand the comment here, the scheduled time is a random number, isn't it? So we can only assert that is is not zero (i. e. it is explicitly set). If you want to be precise you can assert that the scheduled time is within the current time + jitterSeconds.
</blockquote></details>

removed. originally i was going to see if i could have an assertion for getNanos() as well but ran into an error. Right now im just going to check if the seconds() is not equal to zero.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5j8x4CJeByMBYoqpu:-Mo61myFELifbZvjtEdX:b8mmyby) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L144)):*

> ```
> Optional<Integer> jitterSeconds
> ```
What's the point of passing an optional here? We should call this method when jitterSeconds is non-zero, right?
</blockquote></details>

Based on what I noticed, jitterSeconds is injected and Optional<Integer> is the original type. 

If the non zero check is done outside of this method, the same check will be written in multiple places. Even though it's just a one line check, I think it makes more sense to keep it here to avoid writing the same check in multiple places

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 138 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5ia0h4g3nuPXge7z6:-Mo66P9m5rS81Ljl6vwc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L138)):*

> ```
> private static Task createTask
> ```
This warrants a javadoc. What jitterSeconds does is not obvious from the function signature alone.
</blockquote></details>

Done.

jianglai (changes requested):
> Reviewed 1 of 3 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: all files reviewed, 3 unresolved discussions (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 138 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5ia0h4g3nuPXge7z6:-MoFg5yE1_4Fe2G1mFFU:b2o7wkb) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L138)):*

Done.
</blockquote></details>

What I meant was to write a javadoc for this function. Even though this is a private function, it is still where the jitter seconds is actually used, so we should explain what it does here.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5j8x4CJeByMBYoqpu:-MoFjCZm4fruGYlIZOIQ:b-ccyzbg) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L144)):*

Based on what I noticed, jitterSeconds is injected and Optional<Integer> is the original type. 

If the non zero check is done outside of this method, the same check will be written in multiple places. Even though it's just a one line check, I think it makes more sense to keep it here to avoid writing the same check in multiple places 
</blockquote></details>

I still think it probably makes more sense to check it outside this function, because `CloudTasksUtils` should not care about how the jitter seconds variable is injected, but I get it that this way we make the call sites cleaner.

If we do what to do the check here, you can simply return `createTask(path, method, service, params)` when the optional is empty. There is no need to set the time on the task when no jitter seconds variable is provided.
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 91 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5jVAH2UcoiXfyn2PQ:-MoFk4K-CfhmjCMVoCC1:b-2z5vv) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L91)):*

removed. originally i was going to see if i could have an assertion for getNanos() as well but ran into an error. Right now im just going to check if the seconds() is not equal to zero. 
</blockquote></details>

have you tried checking if seconds + nano seconds falls within current time and current time + jitterSeconds?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 138 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5ia0h4g3nuPXge7z6:-MoGcpyz5b4_2DSrBJcw:bnbhww0) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L138)):*

What I meant was to write a javadoc for this function. Even though this is a private function, it is still where the jitter seconds is actually used, so we should explain what it does here.
</blockquote></details>

I included that for the createPostTask and createGetTask variation since those two are public. 
Included javadoc here as well, which was based off the existing createTask javadoc. 
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 144 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5j8x4CJeByMBYoqpu:-MoGugboBsWTRAHMWEKC:be6nl10) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/main/java/google/registry/util/CloudTasksUtils.java#L144)):*

I still think it probably makes more sense to check it outside this function, because `CloudTasksUtils` should not care about how the jitter seconds variable is injected, but I get it that this way we make the call sites cleaner.

If we do what to do the check here, you can simply return `createTask(path, method, service, params)` when the optional is empty. There is no need to set the time on the task when no jitter seconds variable is provided.
</blockquote></details>

went with the latter approach by calling ```createTask(path, method, service, params)``` when jitterseconds is empty or below or equal to zero. 
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 91 at r2](https://reviewable.io/reviews/google/nomulus/1408#-Mo5jVAH2UcoiXfyn2PQ:-MoH1eYC9wIH7Ad_CVvx:be9f3tr) ([raw file](https://github.com/google/nomulus/blob/697286eb7c11d2975e9f57e0b51e80ce8fc639b1/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L91)):*

have you tried checking if seconds + nano seconds falls within current time and current time + jitterSeconds?
</blockquote></details>

now I changed the way to do time comparison; it's a good idea to check both upper bound and lower bound. Thanks!

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: all files reviewed, 5 unresolved discussions (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 139 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKuuNQ1Vekl918S7hQ:-MoKuuNQ1Vekl918S7hR:b-5560hg) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/main/java/google/registry/util/CloudTasksUtils.java#L139)):*
> ```
> Create a {@link Task} to be enqueued.
> ```
Create a Task to be enqueue with a random delay.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 149 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKv1kGBH0g7ED_ZtNC:-MoKv1kGBH0g7ED_ZtND:blft1p8) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/main/java/google/registry/util/CloudTasksUtils.java#L149)):*
> ```
> @param clock a source of time.
> ```
This is cute :). Though we have not been very strict about requiring every parameter to be explained. You could have skipped this one.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 163 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKuT_E6Hh3OTsFWrUc:-MoKuT_F2U4vLuOeu6G1:b540ib5) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/main/java/google/registry/util/CloudTasksUtils.java#L163)):*
> ```
> (jitterSeconds.isPresent() && jitterSeconds.get() <= 0) || jitterSeconds.isEmpty()
> ```
This can just be `(jitterSeconds.isEmpty() || jitterSeconds.get() <= 0)` because of short circuiting.
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 96 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKwSGf-UhPXmFQ4qt0:-MoKwSGf-UhPXmFQ4qt1:b-4l98w6) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L96)):*
> ```
> plusMillis
> ```
Shouldn't this be `plusSeconds`?
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 98 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKx2zk4FFi8Axw1O3D:-MoKx2zk4FFi8Axw1O3E:bj0r6j9) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L98)):*
> ```
>     assertThat(scheduleTime.getEpochSecond() + scheduleTime.getNano())
>         .isAtLeast(lowerBoundTime.getEpochSecond() + lowerBoundTime.getNano());
> ```
This can be just `assertThat(scheduleTime).isAfter(lowerBoundTime)`, right?

Same below.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 139 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKuuNQ1Vekl918S7hQ:-MoLCnD838I4chbcGe8n:b-nkvj82) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/main/java/google/registry/util/CloudTasksUtils.java#L139)):*

> ```
> Create a {@link Task} to be enqueued.
> ```
Create a Task to be enqueue with a random delay.
</blockquote></details>

Done../gra
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 149 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKv1kGBH0g7ED_ZtNC:-MoLAccv75xBrA_EARr3:b52wly1) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/main/java/google/registry/util/CloudTasksUtils.java#L149)):*

> ```
> @param clock a source of time.
> ```
This is cute :). Though we have not been very strict about requiring every parameter to be explained. You could have skipped this one.
</blockquote></details>

I thought  there's explanation for every single parameter of the existing createTask(), i might as well just add that here for the createTask() variation. This was actually found in our repo :)  
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 163 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKuT_E6Hh3OTsFWrUc:-MoLA_6M9my98zYSj8aA:b-9ci8mb) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/main/java/google/registry/util/CloudTasksUtils.java#L163)):*

> ```
> (jitterSeconds.isPresent() && jitterSeconds.get() <= 0) || jitterSeconds.isEmpty()
> ```
This can just be `(jitterSeconds.isEmpty() || jitterSeconds.get() <= 0)` because of short circuiting.
</blockquote></details>

right. fixed. 
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 98 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKx2zk4FFi8Axw1O3D:-MoLGje_9I9iLNKpX_uQ:b-ccxicb) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L98)):*

> ```
>     assertThat(scheduleTime.getEpochSecond() + scheduleTime.getNano())
>         .isAtLeast(lowerBoundTime.getEpochSecond() + lowerBoundTime.getNano());
> ```
This can be just `assertThat(scheduleTime).isAfter(lowerBoundTime)`, right?

Same below.
</blockquote></details>

there's randomness involved with jitterSeconds.
 ```random.nextInt(bound)``` returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and the specified value (exclusive). 

I think it's safe to say that schedule time is less than the upper bound time, so I don't need to use ```isAtmost```..
However, there's a really low chance that the number ends up being zero and that will break the test case.

jianglai (changes requested):
> Reviewed 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 96 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKwSGf-UhPXmFQ4qt0:-MoOofgj0zlEb57a7zq9:b308ek3) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L96)):*

> ```
> plusMillis
> ```
Shouldn't this be `plusSeconds`?
</blockquote></details>

There's no need to do this, you can just use `plusSeconds(100)`.

https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#plusSeconds-long-
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 98 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKx2zk4FFi8Axw1O3D:-MoOpErC8oDjIQpH1HiN:bd1vs6h) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L98)):*

there's randomness involved with jitterSeconds.
 ```random.nextInt(bound)``` returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and the specified value (exclusive). 

I think it's safe to say that schedule time is less than the upper bound time, so I don't need to use ```isAtmost```..
However, there's a really low chance that the number ends up being zero and that will break the test case.  
</blockquote></details>

Good point about the inclusive lower bound.

However you cannot just add the value of second with the value of nanosecond, can you? You have to convert them to the same unit (nanoseconds).

Plus, you can still use `Instant` to compare the lower bound, just do:

`assertThat(scheduleTime).isBefore(lowerBoundTime).isFalse()`.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 96 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKwSGf-UhPXmFQ4qt0:-MoLICf_C_P5CTzFkzE8:b-gfj8z5) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L96)):*

There's no need to do this, you can just use `plusSeconds(100)`.

https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#plusSeconds-long-
</blockquote></details>

it was a typo, I meant to convert that to milliseconds in the code. 

 I tried to keep it similar to how schedule time was originally written: 
```
    Instant scheduleTime =
        Instant.ofEpochMilli(
            clock
                .nowUtc()
                .plusMillis(
                    jitterSeconds
                        .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
                        .orElse(0))
                .getMillis());
```
And I really like the idea of adding more randomness by converting seconds to milliseconds. 
So for test cases, I tried to create times the same way it is in ```CloudTasksUtils.java'
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 98 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKx2zk4FFi8Axw1O3D:-MoZQp0o5fxbbMVMOXJ4:bjtlibx) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L98)):*

Good point about the inclusive lower bound.

However you cannot just add the value of second with the value of nanosecond, can you? You have to convert them to the same unit (nanoseconds).

Plus, you can still use `Instant` to compare the lower bound, just do:

`assertThat(scheduleTime).isBefore(lowerBoundTime).isFalse()`.
</blockquote></details>

I tried to show explicitly the math but didn't think of converting the values. Comparing Instant object using the APIs it comes with definitely would work better. Thanks for the suggestion!

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 96 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKwSGf-UhPXmFQ4qt0:-MocYeXW-Bp15lgajhSj:b-rjqpr2) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L96)):*

it was a typo, I meant to convert that to milliseconds in the code. 

 I tried to keep it similar to how schedule time was originally written: 
```
    Instant scheduleTime =
        Instant.ofEpochMilli(
            clock
                .nowUtc()
                .plusMillis(
                    jitterSeconds
                        .map(seconds -> random.nextInt((int) SECONDS.toMillis(seconds)))
                        .orElse(0))
                .getMillis());
```
And I really like the idea of adding more randomness by converting seconds to milliseconds. 
So for test cases, I tried to create times the same way it is in ```CloudTasksUtils.java'
</blockquote></details>

It is correct that by using millisecond in generating the random number you get more "randomness", but here you are just creating a fix instant, what is the point of doing the unit conversion here? The instant is the same regardless of what unit you use.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 96 at r5](https://reviewable.io/reviews/google/nomulus/1408#-MoKwSGf-UhPXmFQ4qt0:-Mod_fXD8Vjx18EucWlI:b-oc2jde) ([raw file](https://github.com/google/nomulus/blob/eca93017b0f8bf7868ae70d21e16326652d53b03/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L96)):*

It is correct that by using millisecond in generating the random number you get more "randomness", but here you are just creating a fix instant, what is the point of doing the unit conversion here? The instant is the same regardless of what unit you use.
</blockquote></details>

you got a point and i changed that.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1408)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

PR #1413, opened 2021-11-11
Correct HTTP method in CommitLogCheckPointAction

rachelguan:
There's an ongoing effort to move away from App Engine API to Cloud Task APIs.  The old task APIs in CommitLogCheckPointAction.java and its test case were replaced after PR#1409 : Change TaskQueueUtils to CloudTaskUtils in CommitLogCheckPointAction (https://github.com/google/nomulus/pull/1409/files) 

The HTTP Method used for the task was not explicitly declared and the task is supposed to be created via POST instead of GET.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1413)

PR #1418, opened 2021-11-12
Make TaskMatcher default to POST methods

mindhog:
TaskOptions.Builder.withUrl() defaults to POST methods.  Therefore, it seems
reasonable to verify that task queue methods are using the POST method,
especially given that the method must now be identified explicitly when using
CloudTaskUtils.  This check would have guarded against the bug fixed by #1413.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1418)

### DIFF
--- a/core/src/test/java/google/registry/testing/CloudTasksHelper.java
+++ b/core/src/test/java/google/registry/testing/CloudTasksHelper.java
@@ -188,7 +188,10 @@ public class CloudTasksHelper implements Serializable {
 
     String taskName;
     String service;
-    HttpMethod method = HttpMethod.POST;  // Task queue methods default to "POST".
+    // Task queue methods default to "POST".  This isn't obvious from the code, so we default it to
+    // POST here so that we don't accidentally create an entry with a GET method.  Should we ever
+    // actually want to do a GET, we'll set this explicitly.
+    HttpMethod method = HttpMethod.POST;
     String url;
     Multimap<String, String> headers = ArrayListMultimap.create();
     Multimap<String, String> params = ArrayListMultimap.create();
--- a/core/src/test/java/google/registry/testing/CloudTasksHelper.java
+++ b/core/src/test/java/google/registry/testing/CloudTasksHelper.java
@@ -188,9 +188,11 @@ public class CloudTasksHelper implements Serializable {
 
     String taskName;
     String service;
-    // Task queue methods default to "POST".  This isn't obvious from the code, so we default it to
-    // POST here so that we don't accidentally create an entry with a GET method.  Should we ever
-    // actually want to do a GET, we'll set this explicitly.
+    // App Engine TaskOption methods default to "POST".  This isn't obvious from the code, so we
+    // default it to POST here so that we don't accidentally create an entry with a GET method when
+    // converting to CloudTaskUtils, which requires that the method be specified explicitly.
+    // Should we ever actually want to do a GET, we'll likewise have to set this explicitly for the
+    // tests.
     HttpMethod method = HttpMethod.POST;
     String url;
     Multimap<String, String> headers = ArrayListMultimap.create();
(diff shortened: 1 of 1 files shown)


rachelguan (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1418#-:-MoJirzdBbsGl3p3qHWx:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai:
There's no "default" going forward in `CloudTasksUtils`, so how is making the test default to POST help? When you use `CloudTasksUtils` you will alway have to specify the test method anyway.

I can see that we default to POST because most actions use POST, so it's one few parameter to set on the `Taskmatcher`. But then the comment should reflect that.

mindhog (commented):
It's helpful because the existing default is POST (I can't think of a situation where GET is appropriate) and this is not obvious from looking at our code.  I've updated the comment with more detail.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1418#-:-MoJjkHQ4on_Okyo1eAC:bllz2sg)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1418)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 191 at r2](https://reviewable.io/reviews/google/nomulus/1418#-MoJn_tf62skM78KrPtt:-MoJn_tf62skM78KrPtu:bu0tdve) ([raw file](https://github.com/google/nomulus/blob/e5448c0d6535d0c61ef67df8f450b31f411b285b/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L191)):*
> ```
> Task queue methods default to "POST".
> ```
I don't think this is accurate. As I mentioned with `CloudTasksUtils` there is no default method. The reason we want to set the default here is because most, if not all, actions use POST, so we will enqueue POST tasks anyway.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1418)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 191 at r2](https://reviewable.io/reviews/google/nomulus/1418#-MoJn_tf62skM78KrPtt:-MoJoAsr2-qgo7Q7TiyY:b-arnaba) ([raw file](https://github.com/google/nomulus/blob/e5448c0d6535d0c61ef67df8f450b31f411b285b/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L191)):*

> ```
> Task queue methods default to "POST".
> ```
I don't think this is accurate. As I mentioned with `CloudTasksUtils` there is no default method. The reason we want to set the default here is because most, if not all, actions use POST, so we will enqueue POST tasks anyway.
</blockquote></details>

I verified this.  TaskOptions.withUrl() returns a new TaskOptions object with method POST.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1418)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 191 at r2](https://reviewable.io/reviews/google/nomulus/1418#-MoJn_tf62skM78KrPtt:-MoJuexH-hR0bP_lyjYo:b-a7tigo) ([raw file](https://github.com/google/nomulus/blob/e5448c0d6535d0c61ef67df8f450b31f411b285b/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L191)):*

I verified this.  TaskOptions.withUrl() returns a new TaskOptions object with method POST.
</blockquote></details>

I've further improved the comment.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1418)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1423, opened 2021-11-16
Grandfather in old data for one-time billing event requirement

CydeWeys:
We have data from 2018 and earlier where we didn't consistently set periodYears
for OneTime BillingEvents with certain reasons. This grandfathers in that old
data so that we can successfully move it over to Cloud SQL for now, then we can
later run a query that will backfill it, after which we can then tighten up the
requirement again. Note that the requirement is still being enforced for all
billing events from 2019 onwards.

This also improves the handling of validation, by adding a private field to the
Reason enum rather than creating a throwaway inline ImmmutableSet in the
Builder.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1423)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1423)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1425, opened 2021-11-16
Provide useful error messages on flows run during read-only mode

gbrodman:
We want to keep the read-only-mode-exception as an unchecked exception,
so we introduce a temporary check in the EppController that provides a
specific error message for this situation (rather than letting it fall
through to the generic "command failed" messaging

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1425)

### DIFF
--- a/core/src/main/java/google/registry/flows/EppController.java
+++ b/core/src/main/java/google/registry/flows/EppController.java
@@ -143,9 +143,6 @@ public final class EppController {
   /** Creates a response indicating an EPP failure. */
   @VisibleForTesting
   static EppOutput getErrorResponse(Result result, Trid trid) {
-    return EppOutput.create(new EppResponse.Builder()
-        .setResult(result)
-        .setTrid(trid)
-        .build());
+    return EppOutput.create(new EppResponse.Builder().setResult(result).setTrid(trid).build());
   }
 }
--- a/core/src/main/java/google/registry/flows/EppException.java
+++ b/core/src/main/java/google/registry/flows/EppException.java
@@ -25,10 +25,12 @@ import google.registry.model.eppinput.EppInput.InnerCommand;
 import google.registry.model.eppinput.EppInput.ResourceCommandWrapper;
 import google.registry.model.eppoutput.Result;
 import google.registry.model.eppoutput.Result.Code;
+import google.registry.persistence.transaction.TransactionManagerFactory.ReadOnlyModeException;
 import java.lang.annotation.Documented;
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+import javax.annotation.Nullable;
 
 /** Exception used to propagate all failures containing one or more EPP responses. */
 public abstract class EppException extends Exception {
@@ -37,7 +39,12 @@ public abstract class EppException extends Exception {
 
   /** Create an EppException with a custom message. */
   private EppException(String message) {
-    super(message);
+    this(message, null);
+  }
+
+  /** Create an EppException with a custom message and cause. */
+  private EppException(String message, @Nullable Throwable cause) {
+    super(message, cause);
     Code code = getClass().getAnnotation(EppResultCode.class).value();
     Preconditions.checkState(!code.isSuccess());
     this.result = Result.create(code, message);
@@ -253,6 +260,14 @@ public abstract class EppException extends Exception {
   public static class UnimplementedProtocolVersionException extends EppException {
     public UnimplementedProtocolVersionException() {
       super("Specified protocol version is not implemented");
+    }
+  }
+
+  /** Registry is currently undergoing maintenance and is in read-only mode. */
+  @EppResultCode(Code.COMMAND_FAILED)
+  public static class ReadOnlyModeEppException extends EppException {
+    ReadOnlyModeEppException(ReadOnlyModeException cause) {
+      super("Registry is currently undergoing maintenance and is in read-only mode", cause);
     }
   }
 }
--- a/core/src/main/java/google/registry/flows/FlowRunner.java
+++ b/core/src/main/java/google/registry/flows/FlowRunner.java
@@ -19,6 +19,7 @@ import static google.registry.xml.XmlTransformer.prettyPrint;
 
 import com.google.common.base.Strings;
 import com.google.common.flogger.FluentLogger;
+import google.registry.flows.EppException.ReadOnlyModeEppException;
 import google.registry.flows.FlowModule.DryRun;
 import google.registry.flows.FlowModule.InputXml;
 import google.registry.flows.FlowModule.RegistrarId;
@@ -28,6 +29,7 @@ import google.registry.flows.session.LoginFlow;
 import google.registry.model.eppcommon.Trid;
 import google.registry.model.eppoutput.EppOutput;
 import google.registry.monitoring.whitebox.EppMetric;
+import google.registry.persistence.transaction.TransactionManagerFactory.ReadOnlyModeException;
 import javax.inject.Inject;
 import javax.inject.Provider;
 
@@ -97,6 +99,8 @@ public class FlowRunner {
       return e.output;
     } catch (EppRuntimeException e) {
       throw e.getCause();
+    } catch (ReadOnlyModeException e) {
+      throw new ReadOnlyModeEppException(e);
     }
   }
 
--- a/core/src/main/java/google/registry/flows/contact/ContactCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/contact/ContactCreateFlow.java
@@ -50,6 +50,7 @@ import org.joda.time.DateTime;
 /**
  * An EPP flow that creates a new contact.
  *
+ * @error {@link google.registry.flows.EppException.ReadOnlyModeEppException}
  * @error {@link ResourceAlreadyExistsForThisClientException}
  * @error {@link ResourceCreateContentionException}
  * @error {@link ContactFlowUtils.BadInternationalizedPostalInfoException}
--- a/core/src/main/java/google/registry/flows/contact/ContactDeleteFlow.java
+++ b/core/src/main/java/google/registry/flows/contact/ContactDeleteFlow.java
@@ -60,6 +60,7 @@ import org.joda.time.DateTime;
  * references to the host before the deletion is allowed to proceed. A poll message will be written
  * with the success or failure message when the process is complete.
  *
+ * @error {@link google.registry.flows.EppException.ReadOnlyModeEppException}
  * @error {@link google.registry.flows.ResourceFlowUtils.ResourceDoesNotExistException}
  * @error {@link google.registry.flows.ResourceFlowUtils.ResourceNotOwnedException}
  * @error {@link google.registry.flows.exceptions.ResourceStatusProhibitsOperationException}
--- a/core/src/main/java/google/registry/flows/contact/ContactTransferApproveFlow.java
+++ b/core/src/main/java/google/registry/flows/contact/ContactTransferApproveFlow.java
@@ -54,6 +54,7 @@ import org.joda.time.DateTime;
  * transfer is automatically approved. Within that window, this flow allows the losing client to
  * explicitly approve the transfer request, which then becomes effective immediately.
  *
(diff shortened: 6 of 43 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MooPInZ007idB2Yp3l_:bvaarlm) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*
> ```Java
>       EppException eppEx = (EppException) (e instanceof EppException ? e : e.getCause());
>       return getErrorResponse(eppEx.getResult(), flowComponent.trid());
>     } catch (ReadOnlyModeException e) {
> ```

Why this special-handling here rather than just having an `EppException`?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MooQp3T8GH_KYeTZzEp:b-ifbyy5) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

Why this special-handling here rather than just having an `EppException`?
</blockquote></details>

EppExceptions are checked exceptions, meaning that any place that threw it (e.g. the ReadOnlyCheckingEntityManager) would have to declare that it threw it, which isn't possible since we don't control the EntityManager interface.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MotusR41CSFpge6oqPn:b-tzpgcn) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

EppExceptions are checked exceptions, meaning that any place that threw it (e.g. the ReadOnlyCheckingEntityManager) would have to declare that it threw it, which isn't possible since we don't control the EntityManager interface. 
</blockquote></details>

Can you implement this here as a catch and then rethrow an `EppException`?  This manual handling is a little strange.  I'd prefer if it were handled the same as any other `EppException` in the block above.  See `EppRuntimeException` -- that already has built-in handling and may be exactly what we need here.  That way `ReadOnlyModeException` can correctly be an `EppException`, but throw it wrapped inside an `EppRuntimeException` to get around checking concerns.  Then you won't need any special handling here at all.

The problem with your current implementation is that, by not declaring it as an `EppException` like everything else, you're violating some of the constraints for stuff we have built up around `EppException`s.  Namely, you're bypassing the presubmit that ensures that all exceptions can be thrown from EPP flows are documented properly in `flows.md` (which this read-only exception currently is not).

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-Mp7sYau1Zq7ny-ivj0h:b3278fb) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

Can you implement this here as a catch and then rethrow an `EppException`?  This manual handling is a little strange.  I'd prefer if it were handled the same as any other `EppException` in the block above.  See `EppRuntimeException` -- that already has built-in handling and may be exactly what we need here.  That way `ReadOnlyModeException` can correctly be an `EppException`, but throw it wrapped inside an `EppRuntimeException` to get around checking concerns.  Then you won't need any special handling here at all.

The problem with your current implementation is that, by not declaring it as an `EppException` like everything else, you're violating some of the constraints for stuff we have built up around `EppException`s.  Namely, you're bypassing the presubmit that ensures that all exceptions can be thrown from EPP flows are documented properly in `flows.md` (which this read-only exception currently is not).
</blockquote></details>

not here, this is the place where exceptions get finally caught and converted to reasonable XML outputs. 

`EppRuntimeException` seems like a hack to get around the fact that `tm().transact()` doesn't declare any checked exceptions. 

Should `ReadOnlyModeException` be an EppException after all? It can be thrown from places other than EPP flows.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MpCXCK_C4gmAwl-ikoz:b1q130o) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

not here, this is the place where exceptions get finally caught and converted to reasonable XML outputs. 

`EppRuntimeException` seems like a hack to get around the fact that `tm().transact()` doesn't declare any checked exceptions. 

Should `ReadOnlyModeException` be an EppException after all? It can be thrown from places other than EPP flows.
</blockquote></details>

It should be an `EppException` so that it gets reported correctly through EPP (and documented as such).  If that's not feasible from the `TransactionManager` because it would create a circular package dependency, then add handling to catch and re-throw the correct one inside the flow package.  The most logical place for that would be here:  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/FlowRunner.java;l=82-100?q=flowrunner&ss=nomulus%2Fnomulus  And then, you would remove all of this special-cased handling here and just let it be handled as a normal `EppException`.

But as a general rule, any error that is thrown during an EPP flow and that is visible to the end user _MUST_ (in the RFC sense) be an `EppException`.  We tell registrars about `flows.md` and they use it as documentation when integrating with us.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MpCtkG1D14belAzjPwA:b81s38p) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

It should be an `EppException` so that it gets reported correctly through EPP (and documented as such).  If that's not feasible from the `TransactionManager` because it would create a circular package dependency, then add handling to catch and re-throw the correct one inside the flow package.  The most logical place for that would be here:  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/FlowRunner.java;l=82-100?q=flowrunner&ss=nomulus%2Fnomulus  And then, you would remove all of this special-cased handling here and just let it be handled as a normal `EppException`.

But as a general rule, any error that is thrown during an EPP flow and that is visible to the end user _MUST_ (in the RFC sense) be an `EppException`.  We tell registrars about `flows.md` and they use it as documentation when integrating with us.
</blockquote></details>

It's infeasible to do that from the TransactionManager not because of the circular dependency, but because we don't control the declared checked exceptions on the interfaces that the ReadOnly* classes implement. 

Fun fact: you last paragraph is definitely not satisfied by the current code. An example is that `NotLoggedInException` can be thrown by (pretty much?) any flow but because it's tested in `ResourceFlowTestCase` it's not documented in any of the flow javadocs other than `LogoutFlow`. 

In order to document this exception we'd have to add a test to each flow, as shown here.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 44 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MpEz4QN7cwu-Ks8dAiV:bcgz2zw) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

It's infeasible to do that from the TransactionManager not because of the circular dependency, but because we don't control the declared checked exceptions on the interfaces that the ReadOnly* classes implement. 

Fun fact: you last paragraph is definitely not satisfied by the current code. An example is that `NotLoggedInException` can be thrown by (pretty much?) any flow but because it's tested in `ResourceFlowTestCase` it's not documented in any of the flow javadocs other than `LogoutFlow`. 

In order to document this exception we'd have to add a test to each flow, as shown here.  
</blockquote></details>

It might be worthwhile to separately fix the `NotLoggedInException` issue.
___
*[core/src/main/java/google/registry/flows/FlowRunner.java, line 103 at r2](https://reviewable.io/reviews/google/nomulus/1425#-MpEzMxS3ESnCqlIVTbQ:-MpEzMxS3ESnCqlIVTbR:bmun3iu) ([raw file](https://github.com/google/nomulus/blob/8721a2e5b8d86b18c2d0cf6859b8a431c1cc9c3e/core/src/main/java/google/registry/flows/FlowRunner.java#L103)):*
> ```Java
>       throw e.getCause();
>     } catch (ReadOnlyModeException e) {
>       throw new ReadOnlyModeEppException();
> ```

You're discarding `e` here.  Why not pass it as an inner exception to `ReadOnlyModeEppException`?  Then you can preserve the stack trace, so you can see where it was originally thrown.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 44 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/flows/EppController.java, line 137 at r1](https://reviewable.io/reviews/google/nomulus/1425#-MooPInZ007idB2Yp3lZ:-MpHvpXr475XQbxytxD1:b-olijr) ([raw file](https://github.com/google/nomulus/blob/c8835b16a97e643c28efaa18e1a561914854f290/core/src/main/java/google/registry/flows/EppController.java#L137)):*

It might be worthwhile to separately fix the `NotLoggedInException` issue.
</blockquote></details>

sure, that can be in a separate PR 
___
*[core/src/main/java/google/registry/flows/FlowRunner.java, line 103 at r2](https://reviewable.io/reviews/google/nomulus/1425#-MpEzMxS3ESnCqlIVTbQ:-MpHvSMNCZW2cdX927S3:b-84dktc) ([raw file](https://github.com/google/nomulus/blob/8721a2e5b8d86b18c2d0cf6859b8a431c1cc9c3e/core/src/main/java/google/registry/flows/FlowRunner.java#L103)):*

You're discarding `e` here.  Why not pass it as an inner exception to `ReadOnlyModeEppException`?  Then you can preserve the stack trace, so you can see where it was originally thrown.
</blockquote></details>

It requires rejiggering of EppException but sure

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1425)* status: 0 of 44 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1429, opened 2021-11-18
Remove converter for CreateAutoTimestamp

gbrodman:
We can handle it the same way that we handle UpdateAutoTimestamp, where
we simply populate it in SQL if it doesn't exist. This has the following
benefits:

1. The converter is unnecessary code
2. We get non-null column definitions for free (overridden in
EppResource to allow null creation times so that legacy *History objects
can contain null in that field
3. More importantly, this allows us for proper SQL->DS replay. If the
field is filled out using a converter (as before this PR) then the field
is only actually filled out on transaction commit (rather than when the
write occurs within the transaction). This means that when we serialize
the Transaction object during the transaction (the data that gets
replayed to Datastore), we are crucially missing the creation time.

If the creation time is written on commit, we have to start a new
transaction to write the Transaction object, and it's an absolute
necessity that the record of the transaction be included in the
transaction itself so as to avoid situations where the transaction
succeeds but the record fails.

If the field is filled out in a @PrePersist method, crucially that
occurs on the object write itself (before transaction commit).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1429)


mindhog (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1429#-MotqzRGAQy4IUDjToYH):*
How is it that this gets populated in SQL if it doesn't exist?

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

___
*[-- commits, line 5 at r1](https://reviewable.io/reviews/google/nomulus/1429#-MotqzRGAQy4IUDjToYH):*

How is it that this gets populated in SQL if it doesn't exist?
</blockquote></details>

By populated, I mean that we add it in the PrePersist method if it doesn't exist already (i.e. if this is a new object)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/CreateAutoTimestamp.java, line 74 at r1](https://reviewable.io/reviews/google/nomulus/1429#-Mou0HtG--Kqia0OXEjT:-Mou0HtG--Kqia0OXEjU:b-kyr19u) ([raw file](https://github.com/google/nomulus/blob/ef3b8469d950a204a198943871eb49f9f040f334/core/src/main/java/google/registry/model/CreateAutoTimestamp.java#L74)):*
> ```Java
>     CreateAutoTimestamp instance = new CreateAutoTimestamp();
>     instance.timestamp = timestamp;
>     instance.creationTime = timestamp == null ? null : DateTimeUtils.toZonedDateTime(timestamp);
> ```

Wrap the ternary conditional in parentheses so it's easier to parse.  I believe this is in a style guide somewhere, maybe?
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1429#-Mou0U6Y-yYbxDW6gYCY:-Mou0U6Y-yYbxDW6gYCZ:b-f4ijtc) ([raw file](https://github.com/google/nomulus/blob/ef3b8469d950a204a198943871eb49f9f040f334/db/src/main/resources/sql/schema/nomulus.golden.sql#L46)):*
> ```TSQL
>     allowed_registrar_ids text[],
>     allowed_tlds text[],
>     creation_time timestamp with time zone,
> ```

Why is `NOT NULL` going away here?  Shouldn't these never be `null`?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: 7 of 16 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/CreateAutoTimestamp.java, line 74 at r1](https://reviewable.io/reviews/google/nomulus/1429#-Mou0HtG--Kqia0OXEjT:-Mou2Owl12M6B5pz_8oA:b-i7c33q) ([raw file](https://github.com/google/nomulus/blob/ef3b8469d950a204a198943871eb49f9f040f334/core/src/main/java/google/registry/model/CreateAutoTimestamp.java#L74)):*

Wrap the ternary conditional in parentheses so it's easier to parse.  I believe this is in a style guide somewhere, maybe?
</blockquote></details>

Not that I know of, but sure 
___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 46 at r1](https://reviewable.io/reviews/google/nomulus/1429#-Mou0U6Y-yYbxDW6gYCY:-Mou3Dtb03HJG_0p0Wok:bpww6km) ([raw file](https://github.com/google/nomulus/blob/ef3b8469d950a204a198943871eb49f9f040f334/db/src/main/resources/sql/schema/nomulus.golden.sql#L46)):*

Why is `NOT NULL` going away here?  Shouldn't these never be `null`?
</blockquote></details>

Yes, however this was a limitation of the way that the column definition has moved from the containing class (e.g. AllocationToken) to the CreateAutoTimestamp class itself. 

We can't have it be non-null everywhere because of the legacy *History objects, however you're probably correct in that the default should be non-null and the EppResource class itself should define the attribute overrides that allow it to be null in the History objects.

gbrodman:
Note that both sandbox and production rows have zero "Registrar" entities in SQL where the creation_time is null

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: 7 of 16 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java, line 199 at r4](https://reviewable.io/reviews/google/nomulus/1429#-MpCZaTV6wCzHHB23KTJ:-MpCZaTV6wCzHHB23KTK:b-ijyfk8) ([raw file](https://github.com/google/nomulus/blob/8458309e8f59f9cb40b220d6ec9a2641d6d799a7/core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java#L199)):*
> ```Java
>     // both databases.
>     if (!ReplayExtension.replayTestsEnabled()) {
>       return;
> ```

This is not the correct way to handle this, as it will incorrectly mark the test as having passed when in fact it was skipped.  What you want is something like `assumeTrue()`:  https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/Assumptions.html#assumeTrue-boolean-

Do we use this pattern anywhere else?  We'll have to fix those too.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: 6 of 16 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java, line 199 at r4](https://reviewable.io/reviews/google/nomulus/1429#-MpCZaTV6wCzHHB23KTJ:-MpCnKN9EuV0jBpKO9-E:b-flf2hf) ([raw file](https://github.com/google/nomulus/blob/8458309e8f59f9cb40b220d6ec9a2641d6d799a7/core/src/test/java/google/registry/model/replay/ReplicateToDatastoreActionTest.java#L199)):*

This is not the correct way to handle this, as it will incorrectly mark the test as having passed when in fact it was skipped.  What you want is something like `assumeTrue()`:  https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/Assumptions.html#assumeTrue-boolean-

Do we use this pattern anywhere else?  We'll have to fix those too.
</blockquote></details>

yes, it was already used many times in this test class. done.

mindhog (approved):
> Reviewed 7 of 9 files at r2, 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1429)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1431, opened 2021-11-18
Add method to show readable key info

rachelguan:
Currently, there's only pretty print ofy key info not vkey. If we are replacing ofy keys with vkeys, we need to replace the output statement about ofy keys with vkeys' as well. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1431)

### DIFF
--- a/core/src/main/java/google/registry/persistence/VKey.java
+++ b/core/src/main/java/google/registry/persistence/VKey.java
@@ -311,4 +311,31 @@ public class VKey<T> extends ImmutableObject implements Serializable {
     }
     return key;
   }
+
+  /**
+   * Constructs the readable string representation of a {@link VKey}.
+   *
+   * <p>This readable string representation of a vkey contains its type and its sql key or ofy key,
+   * or both.
+   */
+  @Override
+  public String toString() {
+    if (maybeGetOfyKey().isPresent() && maybeGetSqlKey().isPresent()) {
+      return String.format(
+          "VKey<%s>(%s:%s,%s:%s)",
+          getKind().getSimpleName(),
+          SQL_LOOKUP_KEY,
+          getSqlKey().toString(),
+          OFY_LOOKUP_KEY,
+          getOfyKey().getName());
+    } else if (maybeGetSqlKey().isPresent()) {
+      return String.format(
+          "VKey<%s>(%s:%s)", getKind().getSimpleName(), SQL_LOOKUP_KEY, getSqlKey().toString());
+    } else if (maybeGetOfyKey().isPresent()) {
+      return String.format(
+          "VKey<%s>(%s:%s)", getKind().getSimpleName(), OFY_LOOKUP_KEY, getOfyKey().getName());
+    } else {
+      throw new IllegalStateException("VKey should contain at least one form of key");
+    }
+  }
 }
--- a/core/src/main/java/google/registry/persistence/VKey.java
+++ b/core/src/main/java/google/registry/persistence/VKey.java
@@ -323,19 +323,17 @@ public class VKey<T> extends ImmutableObject implements Serializable {
     if (maybeGetOfyKey().isPresent() && maybeGetSqlKey().isPresent()) {
       return String.format(
           "VKey<%s>(%s:%s,%s:%s)",
-          getKind().getSimpleName(),
-          SQL_LOOKUP_KEY,
-          getSqlKey().toString(),
-          OFY_LOOKUP_KEY,
-          getOfyKey().getName());
+          getKind().getSimpleName(), SQL_LOOKUP_KEY, sqlKey, OFY_LOOKUP_KEY, ofyKeyToString());
     } else if (maybeGetSqlKey().isPresent()) {
-      return String.format(
-          "VKey<%s>(%s:%s)", getKind().getSimpleName(), SQL_LOOKUP_KEY, getSqlKey().toString());
+      return String.format("VKey<%s>(%s:%s)", getKind().getSimpleName(), SQL_LOOKUP_KEY, sqlKey);
     } else if (maybeGetOfyKey().isPresent()) {
-      return String.format(
-          "VKey<%s>(%s:%s)", getKind().getSimpleName(), OFY_LOOKUP_KEY, getOfyKey().getName());
+      return String.format("VKey<%s>(%s:%s)", ofyKey.getKind(), OFY_LOOKUP_KEY, ofyKeyToString());
     } else {
       throw new IllegalStateException("VKey should contain at least one form of key");
     }
+  }
+
+  private String ofyKeyToString() {
+    return ofyKey.getName() == null ? String.valueOf(ofyKey.getId()) : ofyKey.getName();
   }
 }
--- a/core/src/main/java/google/registry/tools/GetResourceByKeyCommand.java
+++ b/core/src/main/java/google/registry/tools/GetResourceByKeyCommand.java
@@ -48,7 +48,7 @@ final class GetResourceByKeyCommand implements CommandWithRemoteApi {
       EppResource resource =
           checkNotNull(
               auditedOfy().load().key(resourceKey.getOfyKey()).now(),
-              "Could not load resource for key: " + resourceKey.getOfyKey());
+              "Could not load resource for key: " + resourceKey);
       System.out.println(expand ? resource.toHydratedString() : resource.toString());
     }
   }
--- a/core/src/test/java/google/registry/persistence/VKeyTest.java
+++ b/core/src/test/java/google/registry/persistence/VKeyTest.java
@@ -418,6 +418,25 @@ class VKeyTest {
     assertThat(VKey.create(vkeyStringFromQueue)).isEqualTo(vkey);
   }
 
+  @Test
+  void testToString_sqlOnlyVKey() {
+    assertThat(VKey.createSql(TestObject.class, "testId").toString())
+        .isEqualTo("VKey<TestObject>(sql:testId)");
+  }
+
+  @Test
+  void testToString_ofyOnlyVKey() {
+    assertThat(VKey.createOfy(TestObject.class, Key.create(TestObject.class, "testId")).toString())
+        .isEqualTo("VKey<TestObject>(ofy:testId)");
+  }
+
+  @Test
+  void testToString_sqlAndOfyVKey() {
+    assertThat(
+            VKey.create(TestObject.class, "foo", Key.create(TestObject.create("ofy"))).toString())
+        .isEqualTo("VKey<TestObject>(sql:foo,ofy:ofy)");
+  }
+
   @Entity
   static class OtherObject {}
 }
--- a/core/src/test/java/google/registry/persistence/VKeyTest.java
+++ b/core/src/test/java/google/registry/persistence/VKeyTest.java
@@ -425,9 +425,16 @@ class VKeyTest {
   }
 
   @Test
-  void testToString_ofyOnlyVKey() {
-    assertThat(VKey.createOfy(TestObject.class, Key.create(TestObject.class, "testId")).toString())
-        .isEqualTo("VKey<TestObject>(ofy:testId)");
+  void testToString_ofyOnlyVKey_withName() {
+    assertThat(
+            VKey.createOfy(TestObject.class, Key.create(TestObject.class, "testName")).toString())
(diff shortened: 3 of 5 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 321 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCeeFO5OKG69O-A3NU:-MpCeeFO5OKG69O-A3NV:b-2bevt3) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L321)):*
> ```Java
>    * or both.
>    */
>   public String getInfo() {
> ```

Can we just make this override `toString()`?  I don't think we're losing anything by overriding `ImmutableObject.toString()`, that's just intended to provide a readable representation -- yours will be more readable.

This will allow us to avoid the `getInfo()` invocations every place we put this into a log message.
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 324 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCfDrR6aWBzOSyn7fe:-MpCfDrR6aWBzOSyn7ff:b-awyz5c) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L324)):*
> ```Java
>     if (maybeGetOfyKey().isPresent() && maybeGetSqlKey().isPresent()) {
>       return String.format(
>           "VKey<%s>(sqlKey:%s,ofyKey:%s)",
> ```

Optional: In the interest of brevity, I would omit the "Key" suffix and just make this "`sql:%s,ofy:%s`".  Below, also.
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-MpCg8u23GUF4EKDpde_:b3lr0tf) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*
> ```Java
>       return String.format(
>           "VKey<%s>(sqlKey:%s,ofyKey:%s)",
>           getKind().getSimpleName(), getSqlKey().toString(), getOfyKey().getName());
> ```

The ofy key doesn't necessarily have a name.  It can be either a name or a (long) id.  I'm also wondering about whether it should include the ofy kind.  I think this should always be the same as the VKey kind, but I'll verify that.  If for some reason that's not always true, we'll want to display it, at least if it's different.
___
*[core/src/test/java/google/registry/tools/GetResourceByKeyCommandTest.java, line 215 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCj9wi1lO_tMdHwGTf:-MpCj9wi1lO_tMdHwGTg:bupeq8) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/test/java/google/registry/tools/GetResourceByKeyCommandTest.java#L215)):*
> ```Java
>     assertThat(thrown)
>         .hasMessageThat()
>         .contains("Could not load resource for key: VKey<DomainBase>(sqlKey:2-TLD,ofyKey:2-TLD)");
> ```

In general, this is going to be a big improvement over what we get from `ImmutableObject`. :-)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @mindhog and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 321 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCeeFO5OKG69O-A3NU:-MpDYNAp2Ws0i1jJnpCt:bv6u3br) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L321)):*

Can we just make this override `toString()`?  I don't think we're losing anything by overriding `ImmutableObject.toString()`, that's just intended to provide a readable representation -- yours will be more readable.

This will allow us to avoid the `getInfo()` invocations every place we put this into a log message.
</blockquote></details>

I didn't use toString() because I was thinking that since there're lots of toString() in our code, it's hard to find this particular method if we want to do code search. But I understand your reasons so i changed it to toString()
___
*[core/src/main/java/google/registry/persistence/VKey.java, line 324 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCfDrR6aWBzOSyn7fe:-MpDZSfo6LzbszRZledH:b-rns92z) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L324)):*

Optional: In the interest of brevity, I would omit the "Key" suffix and just make this "`sql:%s,ofy:%s`".  Below, also.
</blockquote></details>

I actually thought about  changing to that to utilize the lookup keys that are already in the file but i wanted to include keys since we've been calling them and sqlKeys.. done!

mindhog (commented):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @rachelguan)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-MpleZt89Nt3yD9iE4ZW:bsrc2jb) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*

The ofy key doesn't necessarily have a name.  It can be either a name or a (long) id.  I'm also wondering about whether it should include the ofy kind.  I think this should always be the same as the VKey kind, but I'll verify that.  If for some reason that's not always true, we'll want to display it, at least if it's different.
</blockquote></details>

so now for ofykey string, i will be using either name or id, depending on the availability. I'm not sure how to create a vkey with kind that's from ofy kind. It throws an error if I try to create such vkey via VKey.create() since it requires Key and VKey to have the same class.

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @mindhog and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-Mpm4MQC9-P1wGqaE-AL:bqfir5w) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*

so now for ofykey string, i will be using either name or id, depending on the availability. I'm not sure how to create a vkey with kind that's from ofy kind. It throws an error if I try to create such vkey via VKey.create() since it requires Key and VKey to have the same class. 
</blockquote></details>

yes,  the "name or id" approach works.  Though you should probably add a test (I think you can probably even use an entity type with a string key -- just Key.create() the ofy key with a long id instead of a string).

Does your question on creating a vkey relate to this PR?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-MpmFqPa4IE2pPo855WN:b-jo2iqq) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*

yes,  the "name or id" approach works.  Though you should probably add a test (I think you can probably even use an entity type with a string key -- just Key.create() the ofy key with a long id instead of a string).

Does your question on creating a vkey relate to this PR?  
</blockquote></details>

added a toString() test case such that the name and id cases will be indirectly tested. 
and yes it is related to this PR.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @mindhog and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-MpmRlUm1mhMmVendsUh:b-32lqyu) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*

added a toString() test case such that the name and id cases will be indirectly tested. 
and yes it is related to this PR. 
</blockquote></details>

I'm guessing what you meant to ask is how to "create a vkey with kind that's *different* from ofy kind," is that correct?  I might be missing something, but it doesn't look like any of the `VKey.create()` methods currently enforce this (perhaps they should).  Is this to test `toString()` in the case where the two classes differ?
If we're enforcing this as an invariant,  you shouldn't have to do anything special for it in `toString()`.  If we're not, it would be helpful to have the ofy kind visible in the `toString()` output _if it differs from the vkey kind_.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-MpqKqOP3qN8Iw9Urt-m:b-fazn0e) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*

I'm guessing what you meant to ask is how to "create a vkey with kind that's *different* from ofy kind," is that correct?  I might be missing something, but it doesn't look like any of the `VKey.create()` methods currently enforce this (perhaps they should).  Is this to test `toString()` in the case where the two classes differ?
If we're enforcing this as an invariant,  you shouldn't have to do anything special for it in `toString()`.  If we're not, it would be helpful to have the ofy kind visible in the `toString()` output _if it differs from the vkey kind_.
</blockquote></details>

In any case, I think this PR is good as is.  We can discuss the "differing kinds" issue out-of-band, if you like.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1431)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @mindhog)

___
*[core/src/main/java/google/registry/persistence/VKey.java, line 325 at r1](https://reviewable.io/reviews/google/nomulus/1431#-MpCg8u23GUF4EKDpdeZ:-MpqdvQN6KcZrgDcigmU:b-c8n17v) ([raw file](https://github.com/google/nomulus/blob/525590d23e9d28edddd2ad3c6c8b026d427bc12e/core/src/main/java/google/registry/persistence/VKey.java#L325)):*

In any case, I think this PR is good as is.  We can discuss the "differing kinds" issue out-of-band, if you like.
</blockquote></details>

yes that's exactly what i was trying to figure out. If there's a variant (vkey kind different from ofy kind), what format should i use to display and more importantly, how to create such vkey. 

i would like to discuss that outside of this PR.

PR #1432, opened 2021-11-19
Ignore read-only mode in SQL->DS replication process

gbrodman:
We need to be able to save indices and save data about the replication
even when we're in read-only mode.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1432)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1432#-:-MpF-7WHCFqzg2qKezRf:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #1435, opened 2021-11-22
Encode class package in stringify()

rachelguan:
Previously, only ofy or sql keys are encoded via stringify() but class type shows the actual path of the package. Even though the value of class type is websafe, it's not a good practice to expose the package info explicitly in a url.


mindhog (changes requested):
Can you send or chat me a reference to the best practice in question? (that's assuming it's internal, if it's public feel free to post it in response)

There are several reasons I can think of for not having a full java package path in a URL, but the solution here doesn't really fix any of them.  If the concern is safeguarding the code against refactoring (which is the primary concern I have about this) then we really need to use a class identifier that doesn't change when the class name or package does.  The table name defined in the at-Table annotation might work, although we might want to define something special like a static method or variable or annotation for each of the persistent classes so that future editors of the code know about the relationship. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1435#-:-MpCXvot4AphzxlvEUzj:bbuusjy)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/persistence/VKeyTest.java, line 284 at r1](https://reviewable.io/reviews/google/nomulus/1435#-MpCXiTZ5hy3PBgswQz3:-MpCXiTZ5hy3PBgswQz4:b-lxulzi) ([raw file](https://github.com/google/nomulus/blob/a74f4ca5d07b60d2eb3ef4a2d5d3af49ef5269e2/core/src/test/java/google/registry/persistence/VKeyTest.java#L284)):*
> ```Java
>         .contains(
>             "Cannot parse key string: kind:rO0ABXQAJ2dvb2dsZS5yZWdpc3RyeS5tb2RlbC"
>                 + "5kb21haW4uRG9tYWluQmFzZQ@sq:l@ofya:bc");
> ```

You see what I mean about change detection tests? ;-)

rachelguan (commented):
yep. Based on our discussion offline, we decided to introduce a mechanism to use CLASS_REGISTRY outside of VKeyTranslatorFactory, which led to pr#1440. 

Here, the full path is replaced with simple name obtained via getClassName(class) of ClassPathManager and class is created via getClass( class name) of ClassPathManager. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1435#-:-MqILpZo4r0wgH5lkHlM:b-ev02od)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/test/java/google/registry/persistence/VKeyTest.java, line 284 at r1](https://reviewable.io/reviews/google/nomulus/1435#-MpCXiTZ5hy3PBgswQz3:-MqILeu84dNXBzkfHsZQ:b-v4zyz7) ([raw file](https://github.com/google/nomulus/blob/a74f4ca5d07b60d2eb3ef4a2d5d3af49ef5269e2/core/src/test/java/google/registry/persistence/VKeyTest.java#L284)):*

You see what I mean about change detection tests? ;-)
</blockquote></details>

yep but this test case i was trying to test an invalid string starting after kind:[class].

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1435)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/model/common/ClassPathManager.java, line 52 at r2](https://reviewable.io/reviews/google/nomulus/1435#-MqK6kce10YrtYxL4UR0:-MqK6kcf4Tijf8RXmuIh:bihpin8) ([raw file](https://github.com/google/nomulus/blob/606df36f45a7583aa3a5755815e1f7988b70c4ae/core/src/main/java/google/registry/model/common/ClassPathManager.java#L52)):*
> ```Java
>   public static void addTestEntityClass(Class<?> clazz) {
>     CLASS_REGISTRY.put(com.googlecode.objectify.Key.getKind(clazz), clazz);
>     CLASS_NAME_REGISTRY.put(clazz, com.googlecode.objectify.Key.getKind(clazz));
> ```

Can we use clazz.getSimpleName() here instead?  (one less thing we have to change when we remove ofy keys)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1435)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/common/ClassPathManager.java, line 52 at r2](https://reviewable.io/reviews/google/nomulus/1435#-MqK6kce10YrtYxL4UR0:-MqKZF7T8-YYJ-7GbAAS:b-q60oqk) ([raw file](https://github.com/google/nomulus/blob/606df36f45a7583aa3a5755815e1f7988b70c4ae/core/src/main/java/google/registry/model/common/ClassPathManager.java#L52)):*

Can we use clazz.getSimpleName() here instead?  (one less thing we have to change when we remove ofy keys)
</blockquote></details>

i was trying to keep it like how it was originally written but yes i can do that. done.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1435)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

PR #1438, opened 2021-11-29
Write commit logs during SQL->DS replay

gbrodman:
Previously, we had no way to ignore read-only mode while still writing
commit log backups. Now, we added this so we can write commit logs in
the SQL->DS replay.

Note:
- When moving to either of the DATASTORE_PRIMARY stages, one must
manually set the SqlReplayCheckpoint first. We don't write to SQL with
backup in this stage because we already wrote the transaction in
question to Datastore. The fact that we manually set the replay
checkpoint means that we'll ignore the extra commit logs that might
otherwise cause problems if we switched back and forth from
DATASTORE_PRIMARY to SQL_PRIMARY.

- The commit logs written during the SQL_PRIMARY phase will, ideally, be
unused. We write them here only so that in the event of a rollback to
Datastore, we will have them for RDE purposes.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1438)

### DIFF
--- a/core/src/main/java/google/registry/backup/BackupUtils.java
+++ b/core/src/main/java/google/registry/backup/BackupUtils.java
@@ -45,7 +45,7 @@ public class BackupUtils {
    * OutputStream} in delimited protocol buffer format.
    */
   static void serializeEntity(ImmutableObject entity, OutputStream stream) throws IOException {
-    EntityTranslator.convertToPb(auditedOfy().saveIgnoringReadOnly().toEntity(entity))
+    EntityTranslator.convertToPb(auditedOfy().saveIgnoringReadOnlyWithoutBackup().toEntity(entity))
         .writeDelimitedTo(stream);
   }
 
--- a/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java
+++ b/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java
@@ -74,7 +74,7 @@ public final class CommitLogCheckpointAction implements Runnable {
                 return;
               }
               auditedOfy()
-                  .saveIgnoringReadOnly()
+                  .saveIgnoringReadOnlyWithoutBackup()
                   .entities(
                       checkpoint, CommitLogCheckpointRoot.create(checkpoint.getCheckpointTime()));
               // Enqueue a diff task between previous and current checkpoints.
--- a/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
+++ b/core/src/main/java/google/registry/backup/ReplayCommitLogsToSqlAction.java
@@ -264,7 +264,7 @@ public class ReplayCommitLogsToSqlAction implements Runnable {
             .ifPresent(
                 sqlEntity -> {
                   sqlEntity.beforeSqlSaveOnReplay();
-                  jpaTm().putIgnoringReadOnly(sqlEntity);
+                  jpaTm().putIgnoringReadOnlyWithoutBackup(sqlEntity);
                 });
       } else {
         // this should never happen, but we shouldn't fail on it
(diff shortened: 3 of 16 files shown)


mindhog (changes requested):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1438)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 232 at r1](https://reviewable.io/reviews/google/nomulus/1438#-MpmXZYa1uNPC64EgCwm:-MpmXZYa1uNPC64EgCwn:bw827za) ([raw file](https://github.com/google/nomulus/blob/3aae19832450c89e2f997c4d6911465e6764a4eb/core/src/main/java/google/registry/model/EppResource.java#L232)):*
> ```Java
>   protected void saveIndexesToDatastore() {
>     ofyTm().putIgnoringReadOnlyWithoutBackup(ForeignKeyIndex.create(this, getDeletionTime()));
>     ofyTm().putIgnoringReadOnlyWithoutBackup(EppResourceIndex.create(Key.create(this)));
> ```

Don't we want index saves to have backup?  They're in the commit log, right?  Or do they get special treatment during restore?
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 296 at r1](https://reviewable.io/reviews/google/nomulus/1438#-Mpmau668X9waG6nHyRi:-Mpmau668X9waG6nHyRj:bodp8mb) ([raw file](https://github.com/google/nomulus/blob/3aae19832450c89e2f997c4d6911465e6764a4eb/core/src/main/java/google/registry/model/server/Lock.java#L296)):*
> ```Java
>           // Locks are not parented under an EntityGroupRoot (so as to avoid write
>           // contention) and don't need to be backed up.
>           transactionManager.putIgnoringReadOnlyWithoutBackup(newLock);
> ```

I can't think of why we'd need the backup for locks, but it seems odd to change the semantics of this.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1438)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 232 at r1](https://reviewable.io/reviews/google/nomulus/1438#-MpmXZYa1uNPC64EgCwm:-MpmnB6l0Nl3Z_DysOx0:b-drcxm2) ([raw file](https://github.com/google/nomulus/blob/3aae19832450c89e2f997c4d6911465e6764a4eb/core/src/main/java/google/registry/model/EppResource.java#L232)):*

Don't we want index saves to have backup?  They're in the commit log, right?  Or do they get special treatment during restore?
</blockquote></details>

Yeah I think you're right. They should be treated the same as everything else 
___
*[core/src/main/java/google/registry/model/server/Lock.java, line 296 at r1](https://reviewable.io/reviews/google/nomulus/1438#-Mpmau668X9waG6nHyRi:-MpmnL0h6TMg2j2RuYqT:b-xozemt) ([raw file](https://github.com/google/nomulus/blob/3aae19832450c89e2f997c4d6911465e6764a4eb/core/src/main/java/google/registry/model/server/Lock.java#L296)):*

I can't think of why we'd need the backup for locks, but it seems odd to change the semantics of this.
</blockquote></details>

This makes sense because locks have never been backed up / replayed to SQL or Datastore. We're treating them as completely separate (even back in the old Datastore-only days, they weren't backed up).

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1438)* status: 15 of 16 files reviewed, all discussions resolved (waiting on @mindhog)

PR #1440, opened 2021-11-30
Extract CLASS_REGISTRY from VKeyTranslatorFactory

rachelguan:
Currently, CLASS_REGISTRY resides in VKeyTranslatorFactory. CLASS_REGISTRY stores class name and package/class path mapping, but this mapping really could be used for class path look up and add test entity classes, in other classes. 

In this PR, a new class ClassPathManager is created to encapsulate CLASS_REGISTRY and the methods are static. This manager will allow class info being passed around without explicitly showing class path in the URL.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1440)

### DIFF
--- a/core/src/main/java/google/registry/model/common/ClassPathManager.java
+++ b/core/src/main/java/google/registry/model/common/ClassPathManager.java
@@ -1,0 +1,48 @@
+// Copyright 2021 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.model.common;
+
+import static com.google.common.base.Functions.identity;
+import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.model.EntityClasses.ALL_CLASSES;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.googlecode.objectify.annotation.EntitySubclass;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** A helper to manage class name and class path mapping. */
+public class ClassPathManager {
+  /**
+   * Class registry allowing us to restore the original class object from the unqualified class
+   * name, which is all the datastore key gives us. Note that entities annotated
+   * with @EntitySubclass are removed because they share the same kind of the key with their parent
+   * class.
+   */
+  public static final Map<String, Class<?>> CLASS_REGISTRY =
+      ALL_CLASSES.stream()
+          .filter(clazz -> !clazz.isAnnotationPresent(EntitySubclass.class))
+          .collect(Collectors.toMap(com.googlecode.objectify.Key::getKind, identity()));
+
+  @VisibleForTesting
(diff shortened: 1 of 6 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1440)* status: all files reviewed, 3 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/model/common/ClassPathManager.java, line 47 at r1](https://reviewable.io/reviews/google/nomulus/1440#-MpwFe29A_tmPwBL2o0G:-MpwFe2A4jlxpO2QP7xV:b-xl2qxb) ([raw file](https://github.com/google/nomulus/blob/e480866dab61596518e0d4ee47265da3c89e99aa/core/src/main/java/google/registry/model/common/ClassPathManager.java#L47)):*
> ```Java
>     checkArgument(CLASS_REGISTRY.containsKey(className), "Class not found in class registry");
>     return (Class<T>) CLASS_REGISTRY.get(className);
>   }
> ```

Optional: You might want to just add a "getClassName()" method that does the opposite translation at this point (it can be used instead of getSimpleName() in stringify(), allowing us to facilitate name changes directly from this class, and could even just be implemented as getSimpleName() if you like) 
___
*[core/src/test/java/google/registry/model/common/ClassPathManagerTest.java, line 54 at r1](https://reviewable.io/reviews/google/nomulus/1440#-MpwDUX5BQSFNOr5_CML:-MpwDUX5BQSFNOr5_CMM:b-lqlr77) ([raw file](https://github.com/google/nomulus/blob/e480866dab61596518e0d4ee47265da3c89e99aa/core/src/test/java/google/registry/model/common/ClassPathManagerTest.java#L54)):*
> ```Java
>   @Test
>   void getClass_classInClassRegistry_returnsClass() throws ClassNotFoundException {
>     // Below are all classes supported by CLASS_REGISTRY. THis test breaks if someone changes a
> ```

minor typo: "THis" -> "This"
___
*[core/src/test/java/google/registry/model/common/ClassPathManagerTest.java, line 55 at r1](https://reviewable.io/reviews/google/nomulus/1440#-MpwDZk8BHw3E2bIKNmQ:-MpwDZk8BHw3E2bIKNmR:byc6zqt) ([raw file](https://github.com/google/nomulus/blob/e480866dab61596518e0d4ee47265da3c89e99aa/core/src/test/java/google/registry/model/common/ClassPathManagerTest.java#L55)):*
> ```Java
>   void getClass_classInClassRegistry_returnsClass() throws ClassNotFoundException {
>     // Below are all classes supported by CLASS_REGISTRY. THis test breaks if someone changes a
>     // classname without preserving the original name.
> ```

We need to add some information to this about why we're doing this (so no-one thinks the correct fix is to change the name string).

Something like "Class names are used in stringified VKeys, which can be present in task queues, so changing them could break task queue entries that are present during a rollout.  If you want to change the names of any of these classes, you'll need to introduce some mechanism to deal with this."

mindhog (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1440)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/model/common/ClassPathManagerTest.java, line 55 at r1](https://reviewable.io/reviews/google/nomulus/1440#-MpwDZk8BHw3E2bIKNmQ:-MqFpPPzABcbvGgqj4tp:bwbbi42) ([raw file](https://github.com/google/nomulus/blob/e480866dab61596518e0d4ee47265da3c89e99aa/core/src/test/java/google/registry/model/common/ClassPathManagerTest.java#L55)):*

added. 
</blockquote></details>

Looks good, although at this point (with the reverse mapping you've introduced in  `ClassPathManager` you may want to instruct them to special-case the class there.

PR #1446, opened 2021-12-02
Allow command to enqueue poll messages for multiple registrars

CydeWeys:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1446)

### DIFF
--- a/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java
+++ b/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java
@@ -18,14 +18,21 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.model.EppResourceUtils.loadByForeignKey;
 import static google.registry.model.reporting.HistoryEntry.Type.SYNTHETIC;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.util.CollectionUtils.isNullOrEmpty;
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Streams;
+import google.registry.config.RegistryConfig.Config;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.domain.DomainHistory;
 import google.registry.model.poll.PollMessage;
+import google.registry.model.registrar.Registrar;
 import google.registry.model.reporting.HistoryEntry;
+import java.util.List;
 import java.util.Optional;
+import javax.inject.Inject;
 
 /**
  * Tool to enqueue a poll message for a registrar.
@@ -54,14 +61,26 @@ class EnqueuePollMessageCommand extends MutatingCommand {
   String domainName;
 
   @Parameter(
-      names = {"-c", "--client"},
+      names = {"-c", "--clients"},
       description =
-          "Client identifier of the registrar to send the poll message to, if not the owning"
-              + " registrar of the domain")
-  String clientId;
+          "Comma-delimited list of the client identifier(s) of the registrar(s) to send the poll"
+              + " message to, if not the owning registrar of the domain")
+  List<String> clientIds;
+
+  @Parameter(
+      names = {"-a", "--all"},
+      description = "Whether to send the message to all real registrars",
+      arity = 1)
+  boolean sendToAll;
+
+  @Inject
+  @Config("registryAdminClientId")
+  String registryAdminClientId;
 
   @Override
   protected final void init() {
+    checkArgument(
+        !sendToAll || isNullOrEmpty(clientIds), "Cannot specify both --all and --clients");
     tm().transact(
             () -> {
               Optional<DomainBase> domainOpt =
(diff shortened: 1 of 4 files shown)


sarahcaseybot (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1446)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11JEA4W8HOqlDRB5S:-Mq11JEA4W8HOqlDRB5T:b-m7j8ah) ([raw file](https://github.com/google/nomulus/blob/a901b5b2b90fc20350ef789e0ed8ad813137d972/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L77)):*
> ```Java
>   protected final void init() {
>     checkArgument(
>         !sendToAll || isNullOrEmpty(clientIds), "Cannot specify both --all and --clients");
> ```

Should there also be a check for both sendToAll and !isNullOrEmpty(clientIds)?
___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 82 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11_8Q8f5nYbi0BcY8:-Mq11_8Q8f5nYbi0BcY9:b3siqtb) ([raw file](https://github.com/google/nomulus/blob/39b613fe8176945dec7bd49ccbc62bf0c5db3c92/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L82)):*
> ```Java
>                       .setModificationTime(tm().getTransactionTime())
>                       .setRequestedByRegistrar(false)
>                       .setRegistrarId(registrarId)
> ```

This changes the registrarId in the history entry from the id of the registrar receiving the poll message to the id of the owning registrar? Does this potentially lose necessary information? Would it be better to create a new history entry for each poll message sent so that the history entry will have the registrarId that the poll message is sent to?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1446)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11JEA4W8HOqlDRB5S:-MqFeO12AOT-XNDMeyoC:ba24alt) ([raw file](https://github.com/google/nomulus/blob/a901b5b2b90fc20350ef789e0ed8ad813137d972/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L77)):*

Should there also be a check for both sendToAll and !isNullOrEmpty(clientIds)?
</blockquote></details>

Well we definitely aren't enforcing that `sendToAll` has to be true along with something else (otherwise it wouldn't even need to be a parameter), so I'm not sure what you're suggesting.  Can you write your suggestion in code?
___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 82 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11_8Q8f5nYbi0BcY8:-MqFegFi8U_KtSu3KZKl:b8m592d) ([raw file](https://github.com/google/nomulus/blob/39b613fe8176945dec7bd49ccbc62bf0c5db3c92/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L82)):*

This changes the registrarId in the history entry from the id of the registrar receiving the poll message to the id of the owning registrar? Does this potentially lose necessary information? Would it be better to create a new history entry for each poll message sent so that the history entry will have the registrarId that the poll message is sent to? 
</blockquote></details>

I was flailing about a little bit in determining exactly what the registrar ID on the history entry should be.  Now that I think more about it, though, the actual registrar that is initiating this action is the registry pseudo-registrar (i.e. us), so that's what should be set here.  As it currently is it'll be us if we use a nic domain but it'll be someone else if a different domain is used, so I'm going to change this to just always use the registry registrar.

As far as having multiple history entries for multiple poll messages, I considered that, but it just results in lots more data being persisted that frankly isn't that useful.  Ideally we wouldn't want any history entries being persisted here; we're just being limited by the poll message schema.  If we do one per poll message then the history entries of `nic.google` are just gonna end up being a huge mess.  Poll messages are ephemeral by design and we don't really need to maintain a permanent record of them.

However, what is interesting is the idea of including the message that is sent in the reason here, so I'll make that change.  As long as we have the history entry anyway, may as well include the poll message that was sent.

sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1446)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11JEA4W8HOqlDRB5S:-MqGHEyoA9eiGgOSFlRN:boitvpa) ([raw file](https://github.com/google/nomulus/blob/a901b5b2b90fc20350ef789e0ed8ad813137d972/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L77)):*

Well we definitely aren't enforcing that `sendToAll` has to be true along with something else (otherwise it wouldn't even need to be a parameter), so I'm not sure what you're suggesting.  Can you write your suggestion in code?
</blockquote></details>

Nevermind, I forgot how checkArgument works.  I thought it threw the error when the statement is true. This looks good how it is. 
___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 82 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11_8Q8f5nYbi0BcY8:-MqGQIcl8XkKqbjyeFyY:b-7e4mr8) ([raw file](https://github.com/google/nomulus/blob/39b613fe8176945dec7bd49ccbc62bf0c5db3c92/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L82)):*

I was flailing about a little bit in determining exactly what the registrar ID on the history entry should be.  Now that I think more about it, though, the actual registrar that is initiating this action is the registry pseudo-registrar (i.e. us), so that's what should be set here.  As it currently is it'll be us if we use a nic domain but it'll be someone else if a different domain is used, so I'm going to change this to just always use the registry registrar.

As far as having multiple history entries for multiple poll messages, I considered that, but it just results in lots more data being persisted that frankly isn't that useful.  Ideally we wouldn't want any history entries being persisted here; we're just being limited by the poll message schema.  If we do one per poll message then the history entries of `nic.google` are just gonna end up being a huge mess.  Poll messages are ephemeral by design and we don't really need to maintain a permanent record of them.

However, what is interesting is the idea of including the message that is sent in the reason here, so I'll make that change.  As long as we have the history entry anyway, may as well include the poll message that was sent.
</blockquote></details>

Yes, I don't see the changes here yet, but I think using the registry's registrar makes sense.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1446)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java, line 82 at r1](https://reviewable.io/reviews/google/nomulus/1446#-Mq11_8Q8f5nYbi0BcY8:-MqQN9Yn3vM1GvVuzFGd:b-ao1to8) ([raw file](https://github.com/google/nomulus/blob/39b613fe8176945dec7bd49ccbc62bf0c5db3c92/core/src/main/java/google/registry/tools/EnqueuePollMessageCommand.java#L82)):*

Yes, I don't see the changes here yet, but I think using the registry's registrar makes sense. 
</blockquote></details>

OK, PR updated with changes.  PTAL.

sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1446)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1448, opened 2021-12-02
Add support for empty or null params for createTask()

rachelguan:
The existing ```createTask()``` does not contain null or empty check for parameters. Therefore, the method throws an error if an empty map or null value is being passed in as ```params```. 

This can be fixed by 1) adding a guard against null or empty parameters so the existing code doesn't have to be modified. However, there might be existing use cases that would send requests without params, or 2) moving the existing logic under an if block ```if (params !=null || !params.isEmpty())```  

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1448)

### DIFF
--- a/util/src/main/java/google/registry/util/CloudTasksUtils.java
+++ b/util/src/main/java/google/registry/util/CloudTasksUtils.java
@@ -106,30 +106,35 @@ public class CloudTasksUtils implements Serializable {
     checkArgument(
         path != null && !path.isEmpty() && path.charAt(0) == '/',
         "The path must start with a '/'.");
+    checkArgument(
+        method.equals(HttpMethod.GET) || method.equals(HttpMethod.POST),
+        "HTTP method %s is used. Only GET and POST are allowed.",
+        method);
     AppEngineHttpRequest.Builder requestBuilder =
         AppEngineHttpRequest.newBuilder()
             .setHttpMethod(method)
             .setAppEngineRouting(AppEngineRouting.newBuilder().setService(service).build());
-    Escaper escaper = UrlEscapers.urlPathSegmentEscaper();
-    String encodedParams =
-        Joiner.on("&")
-            .join(
-                params.entries().stream()
-                    .map(
-                        entry ->
-                            String.format(
-                                "%s=%s",
-                                escaper.escape(entry.getKey()), escaper.escape(entry.getValue())))
-                    .collect(toImmutableList()));
-    if (method == HttpMethod.GET) {
-      path = String.format("%s?%s", path, encodedParams);
-    } else if (method == HttpMethod.POST) {
-      requestBuilder
-          .putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString())
-          .setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
-    } else {
-      throw new IllegalArgumentException(
-          String.format("HTTP method %s is used. Only GET and POST are allowed.", method));
+
+    if (method == HttpMethod.POST) {
+      requestBuilder.putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString());
+    }
+    if (params != null && !params.isEmpty()) {
+      Escaper escaper = UrlEscapers.urlPathSegmentEscaper();
+      String encodedParams =
+          Joiner.on("&")
+              .join(
+                  params.entries().stream()
+                      .map(
+                          entry ->
+                              String.format(
+                                  "%s=%s",
+                                  escaper.escape(entry.getKey()), escaper.escape(entry.getValue())))
+                      .collect(toImmutableList()));
+      if (method == HttpMethod.GET) {
+        path = String.format("%s?%s", path, encodedParams);
+      } else {
+        requestBuilder.setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
+      }
     }
     requestBuilder.setRelativeUri(path);
     return Task.newBuilder().setAppEngineHttpRequest(requestBuilder.build()).build();
--- a/util/src/main/java/google/registry/util/CloudTasksUtils.java
+++ b/util/src/main/java/google/registry/util/CloudTasksUtils.java
@@ -115,10 +115,7 @@ public class CloudTasksUtils implements Serializable {
             .setHttpMethod(method)
             .setAppEngineRouting(AppEngineRouting.newBuilder().setService(service).build());
 
-    if (method == HttpMethod.POST) {
-      requestBuilder.putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString());
-    }
-    if (params != null && !params.isEmpty()) {
+    if (!CollectionUtils.isNullOrEmpty(params)) {
       Escaper escaper = UrlEscapers.urlPathSegmentEscaper();
       String encodedParams =
           Joiner.on("&")
@@ -133,7 +130,9 @@ public class CloudTasksUtils implements Serializable {
       if (method == HttpMethod.GET) {
         path = String.format("%s?%s", path, encodedParams);
       } else {
-        requestBuilder.setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
+        requestBuilder
+            .putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString())
+            .setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
       }
     }
     requestBuilder.setRelativeUri(path);
--- a/core/src/test/java/google/registry/cron/TldFanoutActionTest.java
+++ b/core/src/test/java/google/registry/cron/TldFanoutActionTest.java
@@ -98,11 +98,7 @@ class TldFanoutActionTest {
   }
 
   private void assertTaskWithoutTld() {
-    cloudTasksHelper.assertTasksEnqueued(
-        QUEUE,
-        new TaskMatcher()
-            .url(ENDPOINT)
-            .header("content-type", "application/x-www-form-urlencoded"));
+    cloudTasksHelper.assertTasksEnqueued(QUEUE, new TaskMatcher().url(ENDPOINT));
   }
 
   @Test
--- a/core/src/test/java/google/registry/testing/CloudTasksHelper.java
+++ b/core/src/test/java/google/registry/testing/CloudTasksHelper.java
@@ -230,7 +230,7 @@ public class CloudTasksHelper implements Serializable {
       String query = null;
       if (method == HttpMethod.GET) {
         query = uri.getQuery();
-      } else if (method == HttpMethod.POST) {
+      } else if (method == HttpMethod.POST && !task.getAppEngineHttpRequest().getBody().isEmpty()) {
         assertThat(
                 headers.containsEntry(
                     Ascii.toLowerCase(HttpHeaders.CONTENT_TYPE), MediaType.FORM_DATA.toString()))
--- a/core/src/test/java/google/registry/testing/CloudTasksHelper.java
+++ b/core/src/test/java/google/registry/testing/CloudTasksHelper.java
@@ -227,22 +227,20 @@ public class CloudTasksHelper implements Serializable {
               });
       headers = headerBuilder.build();
       ImmutableMultimap.Builder<String, String> paramBuilder = new ImmutableMultimap.Builder<>();
-      String query = null;
+      // Note that UriParameters.parse() does not throw an IAE on a bad query string (e.g. one
+      // where parameters are not properly URL-encoded); it always does a best-effort parse.
       if (method == HttpMethod.GET) {
-        query = uri.getQuery();
+        paramBuilder.putAll(UriParameters.parse(uri.getQuery()));
       } else if (method == HttpMethod.POST && !task.getAppEngineHttpRequest().getBody().isEmpty()) {
         assertThat(
                 headers.containsEntry(
                     Ascii.toLowerCase(HttpHeaders.CONTENT_TYPE), MediaType.FORM_DATA.toString()))
             .isTrue();
-        query = task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8);
+        paramBuilder.putAll(
+            UriParameters.parse(
+                task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8)));
       }
-      if (query != null) {
-        // Note that UriParameters.parse() does not throw an IAE on a bad query string (e.g. one
-        // where parameters are not properly URL-encoded); it always does a best-effort parse.
-        paramBuilder.putAll(UriParameters.parse(query));
-        params = paramBuilder.build();
-      }
+      params = paramBuilder.build();
     }
 
     public Map<String, Object> toMap() {
--- a/util/src/main/java/google/registry/util/CollectionUtils.java
+++ b/util/src/main/java/google/registry/util/CollectionUtils.java
@@ -49,6 +49,11 @@ public class CollectionUtils {
     return potentiallyNull == null || potentiallyNull.isEmpty();
   }
 
+  /** Checks if a Multimap is null or empty. */
+  public static boolean isNullOrEmpty(@Nullable Multimap<?, ?> potentiallyNull) {
+    return potentiallyNull == null || potentiallyNull.isEmpty();
+  }
+
   /** Turns a null set into an empty set. JAXB leaves lots of null sets lying around. */
   public static <T> Set<T> nullToEmpty(@Nullable Set<T> potentiallyNull) {
     return firstNonNull(potentiallyNull, ImmutableSet.of());
--- a/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java
+++ b/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java
@@ -25,6 +25,7 @@ import static org.mockito.Mockito.when;
 import com.google.cloud.tasks.v2.HttpMethod;
 import com.google.cloud.tasks.v2.Task;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.LinkedListMultimap;
 import google.registry.testing.FakeClock;
 import google.registry.testing.FakeSleeper;
@@ -77,6 +78,52 @@ public class CloudTasksUtilsTest {
         .isEqualTo("application/x-www-form-urlencoded");
     assertThat(task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8))
         .isEqualTo("key1=val1&key2=val2&key1=val3");
+    assertThat(task.getScheduleTime().getSeconds()).isEqualTo(0);
+  }
+
+  @Test
+  void testSuccess_createGetTasks_withNullParams() {
+    Task task = CloudTasksUtils.createGetTask("/the/path", "myservice", null);
+    assertThat(task.getAppEngineHttpRequest().getHttpMethod()).isEqualTo(HttpMethod.GET);
+    assertThat(task.getAppEngineHttpRequest().getRelativeUri()).isEqualTo("/the/path");
+    assertThat(task.getAppEngineHttpRequest().getAppEngineRouting().getService())
+        .isEqualTo("myservice");
+    assertThat(task.getScheduleTime().getSeconds()).isEqualTo(0);
+  }
+
+  @Test
+  void testSuccess_createPostTasks_withNullParams() {
+    Task task = CloudTasksUtils.createPostTask("/the/path", "myservice", null);
+    assertThat(task.getAppEngineHttpRequest().getHttpMethod()).isEqualTo(HttpMethod.POST);
+    assertThat(task.getAppEngineHttpRequest().getRelativeUri()).isEqualTo("/the/path");
+    assertThat(task.getAppEngineHttpRequest().getAppEngineRouting().getService())
+        .isEqualTo("myservice");
+    assertThat(task.getAppEngineHttpRequest().getHeadersMap().get("Content-Type"))
+        .isEqualTo("application/x-www-form-urlencoded");
+    assertThat(task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8)).isEmpty();
+    assertThat(task.getScheduleTime().getSeconds()).isEqualTo(0);
+  }
+
+  @Test
+  void testSuccess_createGetTasks_withEmptyParams() {
+    Task task = CloudTasksUtils.createGetTask("/the/path", "myservice", ImmutableMultimap.of());
+    assertThat(task.getAppEngineHttpRequest().getHttpMethod()).isEqualTo(HttpMethod.GET);
+    assertThat(task.getAppEngineHttpRequest().getRelativeUri()).isEqualTo("/the/path");
+    assertThat(task.getAppEngineHttpRequest().getAppEngineRouting().getService())
+        .isEqualTo("myservice");
+    assertThat(task.getScheduleTime().getSeconds()).isEqualTo(0);
+  }
+
+  @Test
+  void testSuccess_createPostTasks_withEmptyParams() {
+    Task task = CloudTasksUtils.createPostTask("/the/path", "myservice", ImmutableMultimap.of());
+    assertThat(task.getAppEngineHttpRequest().getHttpMethod()).isEqualTo(HttpMethod.POST);
+    assertThat(task.getAppEngineHttpRequest().getRelativeUri()).isEqualTo("/the/path");
+    assertThat(task.getAppEngineHttpRequest().getAppEngineRouting().getService())
+        .isEqualTo("myservice");
+    assertThat(task.getAppEngineHttpRequest().getHeadersMap().get("Content-Type"))
+        .isEqualTo("application/x-www-form-urlencoded");
+    assertThat(task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8)).isEmpty();
     assertThat(task.getScheduleTime().getSeconds()).isEqualTo(0);
   }
 
--- a/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java
+++ b/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java
@@ -98,8 +98,6 @@ public class CloudTasksUtilsTest {
     assertThat(task.getAppEngineHttpRequest().getRelativeUri()).isEqualTo("/the/path");
     assertThat(task.getAppEngineHttpRequest().getAppEngineRouting().getService())
         .isEqualTo("myservice");
-    assertThat(task.getAppEngineHttpRequest().getHeadersMap().get("Content-Type"))
-        .isEqualTo("application/x-www-form-urlencoded");
     assertThat(task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8)).isEmpty();
     assertThat(task.getScheduleTime().getSeconds()).isEqualTo(0);
   }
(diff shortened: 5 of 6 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 116 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFX3Ga4ndZeAu0TMXF:-MqFX3Ga4ndZeAu0TMXG:b-7tod19) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/main/java/google/registry/util/CloudTasksUtils.java#L116)):*
> ```
> checkNotNull(params, "Params cannot be empty");
> ```
I think it is OK for the params to be empty. We should make sure that the code handles null params correctly.
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 55 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFl2p94Ms4hjS7D-HB:-MqFl2p94Ms4hjS7D-HC:b-oh9btq) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L55)):*
> ```
>     headers.put("X-DomainRegistry-ProjectId", "pid");
>     headers.put("X-DomainRegistry-JobId", "jid");
>     headers.put("X-DomainRegistry-ChainedTaskQueue", "ctq");
> ```
If these are all the things that you plan to put in the headers, why not just put them in the params?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 116 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFX3Ga4ndZeAu0TMXF:-MqFoBXm0g5O7yFo_yUD:be1kb87) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/main/java/google/registry/util/CloudTasksUtils.java#L116)):*

> ```
> checkNotNull(params, "Params cannot be empty");
> ```
I think it is OK for the params to be empty. We should make sure that the code handles null params correctly.
</blockquote></details>

oh ok, because it actually breaks at encodedParam when params is null. 
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 55 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFl2p94Ms4hjS7D-HB:-MqFp6pD9_qhq0tk7cy6:bw4lchd) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L55)):*

> ```
>     headers.put("X-DomainRegistry-ProjectId", "pid");
>     headers.put("X-DomainRegistry-JobId", "jid");
>     headers.put("X-DomainRegistry-ChainedTaskQueue", "ctq");
> ```
If these are all the things that you plan to put in the headers, why not just put them in the params?
</blockquote></details>

i am actually not sure why they were set as headers instead of regular params. 
I don't know if adding them to params would break the existing functionality/ 
https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/export/BigqueryPollJobAction.java;l=170?q=X-DomainRegistry-ProjectId&ss=nomulus%2Fnomulus

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 116 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFX3Ga4ndZeAu0TMXF:-MqGVFlwA_GfjejzxKoB:b-ik2o8v) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/main/java/google/registry/util/CloudTasksUtils.java#L116)):*

oh ok, because it actually breaks at encodedParam when params is null. 
</blockquote></details>

    if (method == HttpMethod.GET) {
      path = String.format("%s?%s", path, encodedParams);
    } else if (method == HttpMethod.POST) {
      requestBuilder
          .putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString())
          .setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
    }
both use cases require params not to be null.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 55 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFl2p94Ms4hjS7D-HB:-MqLv0zIC38sUTG2MbGk:b-xicgh1) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L55)):*

i am actually not sure why they were set as headers instead of regular params. 
I don't know if adding them to params would break the existing functionality/ 
https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/export/BigqueryPollJobAction.java;l=170?q=X-DomainRegistry-ProjectId&ss=nomulus%2Fnomulus
</blockquote></details>

As discussed in standup, since we control both how the request is generated and how it is used, it is up to us where the information is saved, in the header or in query params.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 55 at r1](https://reviewable.io/reviews/google/nomulus/1448#-MqFl2p94Ms4hjS7D-HB:-Mq_ZkbJEW1eG_jaG5VM:b-2bz9d5) ([raw file](https://github.com/google/nomulus/blob/4ad80239af9c297c980390ccec7b2148c73e65f0/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L55)):*

As discussed in standup, since we control both how the request is generated and how it is used, it is up to us where the information is saved, in the header or in query params. 
</blockquote></details>

dropped the idea of adding a new parameter to ```createTask()```. The info is going to be in query params. This PR now is about adding support for null/empty params.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: all files reviewed, 3 unresolved discussions (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-MqbYbnR8Po-ES1lOSPl:b-aobiwl) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*
> ```
> if (params != null) {
> ```
You only need to modify the path or the body when the param is not null, so why not put everything under the `if (params != null)` block?

```
if (params != null && !params.isEmpty()) {
  encodedParams = ...
  if (method == GET) {
  ...  
  } else {
  ...
  }
}
```
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 131 at r2](https://reviewable.io/reviews/google/nomulus/1448#-Mqb_Ao8Cu9fnzS3BgWK:-Mqb_Ao8Cu9fnzS3BgWL:b648bkn) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L131)):*
> ```
> "/" 
> ```
If you do what I suggested above, this would be irrelevant, but shouldn't it be `path : String.format(...)` instead of `"/" : String.format(...)`?
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 132 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYKiG7VLYuFxR0elM:-MqbYKiG7VLYuFxR0elN:b-6jq1xr) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L132)):*
> ```
> else if
> ```
This can just be `else` now, since you ruled out any method that is not a POST or GET.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-MqlzkrKEHal8AHuD_VB:b-xj5x6l) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*

> ```
> if (params != null) {
> ```
You only need to modify the path or the body when the param is not null, so why not put everything under the `if (params != null)` block?

```
if (params != null && !params.isEmpty()) {
  encodedParams = ...
  if (method == GET) {
  ...  
  } else {
  ...
  }
}
```
</blockquote></details>

well if  everything goes under  that check, then it means we are only handling case if param contains data. The cleanest way would be ```checkNotNull()``` at the very beginning for params. 

If GET and SET branches are under not null param block, a request will still be sent out with no param, but the task is still being built. Is that really the output we want to return?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 131 at r2](https://reviewable.io/reviews/google/nomulus/1448#-Mqb_Ao8Cu9fnzS3BgWK:-Mqm8Vnq6N03hBdjm6Hn:b-m7varr) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L131)):*

> ```
> "/" 
> ```
If you do what I suggested above, this would be irrelevant, but shouldn't it be `path : String.format(...)` instead of `"/" : String.format(...)`?
</blockquote></details>

should path stay the same for GET if there's params contains no value or null? 
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 132 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYKiG7VLYuFxR0elM:-Mqm8FCQ4XuIXTClVWKo:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L132)):*

> ```
> else if
> ```
This can just be `else` now, since you ruled out any method that is not a POST or GET.
</blockquote></details>

Done.

jianglai (changes requested):
> Reviewed 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-MqpoxwNEFNF2KbcMiIc:bsr0f1s) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*

well if  everything goes under  that check, then it means we are only handling case if param contains data. The cleanest way would be ```checkNotNull()``` at the very beginning for params. 

If GET and SET branches are under not null param block, a request will still be sent out with no param, but the task is still being built. Is that really the output we want to return? 
</blockquote></details>

Not sure whether we are on the same page here. I am proposing to put the GET and POST branches under the the null param block. You are given a path and potentially a list of query parameters. If the list is not empty then you need to either modify the path in case of a GET, or add a body in case of a POST. If it is null or empty you can just build the request with the given path.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 131 at r2](https://reviewable.io/reviews/google/nomulus/1448#-Mqb_Ao8Cu9fnzS3BgWK:-MqpqmnC23Ap-dZdDw2j:b-kuw3ne) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L131)):*

should path stay the same for GET if there's params contains no value or null? 
</blockquote></details>

Yes, we are only modifying the path if the params are not empty (and the request is a GET). We should never disregard the path that is provided to us.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-MqqBD_d2twCSkwiNGRB:b-17rh4o) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*

Not sure whether we are on the same page here. I am proposing to put the GET and POST branches under the the null param block. You are given a path and potentially a list of query parameters. If the list is not empty then you need to either modify the path in case of a GET, or add a body in case of a POST. If it is null or empty you can just build the request with the given path.
</blockquote></details>

I extracted  ```   if (method == HttpMethod.POST) {
      requestBuilder.putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString());
    } ``` out of the ```if (params != null && !params.isEmpty())``` because it's breaking test cases in ```TldFannoutActionTest```

i believe the cases that failed were the following which had to do with missing headers.  How should I handle these two cases? Should I modify the following test cases or do method check twice . 

```
  @Test
  void testSuccess_methodPostAndServiceBackendAreDefault() {
    run(getParamsMap("runInEmpty", ""));
    cloudTasksHelper.assertTasksEnqueued(
        QUEUE, new TaskMatcher().method(HttpMethod.POST).service("backend"));
  }

  @Test
  void testSuccess_runInEmpty() {
    run(getParamsMap("runInEmpty", ""));
    assertTaskWithoutTld();
  }

```
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 131 at r2](https://reviewable.io/reviews/google/nomulus/1448#-Mqb_Ao8Cu9fnzS3BgWK:-MqqBD831k9QuptLDIFy:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L131)):*

Yes, we are only modifying the path if the params are not empty (and the request is a GET). We should never disregard the path that is provided to us.
</blockquote></details>

Done.

util/src/main/java/google/registry/util/CloudTasksUtils.java
```diff
@@ -106,30 +106,34 @@ private static Task createTask(
     checkArgument(
         path != null && !path.isEmpty() && path.charAt(0) == '/',
         "The path must start with a '/'.");
+    checkArgument(
+        method.equals(HttpMethod.GET) || method.equals(HttpMethod.POST),
+        String.format("HTTP method %s is used. Only GET and POST are allowed.", method));
```
CydeWeys, line 111: You don't need to use `String.format` inside a `checkArgument` call.  There's an overload of `checkArgument()` that already takes a format string on its own.

util/src/main/java/google/registry/util/CloudTasksUtils.java
```diff
@@ -106,30 +106,34 @@ private static Task createTask(
...
-          .putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString())
-          .setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
-    } else {
-      throw new IllegalArgumentException(
-          String.format("HTTP method %s is used. Only GET and POST are allowed.", method));
+
+    if (method == HttpMethod.POST) {
+      requestBuilder.putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString());
+    }
+    if (params != null && !params.isEmpty()) {
```
CydeWeys, line 120: `CollectionUtils.isNullOrEmpty()` should be applicable here (not 100% sure if I got that exactly right but it's used elsewhere in our codebase).

CydeWeys (changes requested):
What's the motivation for this change?  Add a paragraph to the commit description explaining why we want to support null or empty params.  What's the use case there?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 111 at r4](https://reviewable.io/reviews/google/nomulus/1448#-Mpx3CsKTLsekf94cU99-r4-111:-MquH4Nj5wFg7tQ-RG0I:b-896fix) ([raw file](https://github.com/google/nomulus/blob/87266fd11d7159baabc80a43f33faae282099bd7/util/src/main/java/google/registry/util/CloudTasksUtils.java#L111)):*

You don't need to use `String.format` inside a `checkArgument` call.  There's an overload of `checkArgument()` that already takes a format string on its own.
</blockquote></details>

Done.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 120 at r4](https://reviewable.io/reviews/google/nomulus/1448#-Mpx3CsKTLsekf94cU99-r4-120:-MquH54c5_AFOnRUWxgG:b9sd8nt) ([raw file](https://github.com/google/nomulus/blob/87266fd11d7159baabc80a43f33faae282099bd7/util/src/main/java/google/registry/util/CloudTasksUtils.java#L120)):*

`CollectionUtils.isNullOrEmpty()` should be applicable here (not 100% sure if I got that exactly right but it's used elsewhere in our codebase).
</blockquote></details>

```CollectionUtils.isNullOrEmpty()``` takes either a ```Map``` or ```Iterable```   . Because the existing method uses a ```MultiMap```, in order to use the method from ```CollectionUtils```, ```params``` has to be converted using either ```params.asMap()``` or ```params.entries()```, which wouldn't work if ```params``` itself is already null.

**Review by jianglai — CHANGES_REQUESTED** [review]
Given that this is a private method and none of its callers passes a null parameter, the check is not strictly necessary. But it is still good practice, and by moving the nullness and emptyness check before construction of `encodedParams` we bypass some unnecessary code.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448#-:-Mr4CXpJ3gaN2l4eFw_J:bm6mcad)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-Mr4AqJF3--xxTY78Dog:b-55inh2) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*

I extracted  ```   if (method == HttpMethod.POST) {
      requestBuilder.putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString());
    } ``` out of the ```if (params != null && !params.isEmpty())``` because it's breaking test cases in ```TldFannoutActionTest```

i believe the cases that failed were the following which had to do with missing headers.  How should I handle these two cases? Should I modify the following test cases or do method check twice . 

```
  @Test
  void testSuccess_methodPostAndServiceBackendAreDefault() {
    run(getParamsMap("runInEmpty", ""));
    cloudTasksHelper.assertTasksEnqueued(
        QUEUE, new TaskMatcher().method(HttpMethod.POST).service("backend"));
  }

  @Test
  void testSuccess_runInEmpty() {
    run(getParamsMap("runInEmpty", ""));
    assertTaskWithoutTld();
  }

```
</blockquote></details>

The latest commit has the `if (method == POST)` block inside the null & empty conditional block. Is it still breaking the tests? If so, what's the failure message?
___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 104 at r5](https://reviewable.io/reviews/google/nomulus/1448#-Mr4CGNv8ajclRLT8yLP:-Mr4CGNv8ajclRLT8yLQ:b-8gtaeg) ([raw file](https://github.com/google/nomulus/blob/3ad7ecc33f74a9a2091cc4a44b2bf122105ba245/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L104)):*
> ```Java
>         .isEqualTo("application/x-www-form-urlencoded");
>     assertThat(task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8))
>         .isEqualTo("");
> ```

You can use `isEmpty()`. Same below.



_Code quote:_
```Java
.isEqualTo("");
```

rachelguan (commented):
In additional to Lai's comment, there are existing ```TaskOptions``` objects that are created with no parameters. All the public methods for creating tasks in ```CloudTasksUtil``` assume params to always be non empty. For example, in ```IcannReportingStagingAction```, an upload task is created with a path, method, and countdown millis, then it gets added to ```CRON_QUEUE```.             
```TaskOptions uploadTask =
                TaskOptions.Builder.withUrl(IcannReportingUploadAction.PATH)
                    .method(Method.POST)
                    .countdownMillis(Duration.standardMinutes(2).getMillis());
            QueueFactory.getQueue(CRON_QUEUE).add(uploadTask);
```

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448#-:-Mr4IcaK1eaM2CckuQb8:bfsq70m)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-Mr4HS_BD_YBwN7GiMqs:b-ddhnzo) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*
> d and none of its callers passes a null parameter, the check is not strictly necessary. But it is still good practice, and by moving the nullness and emptyness check before construction of encodedParams we bypass some unnecessary code.

the latest commit has ```if(method === POST) ``` twice, one as itself and another inside the null & empty conditional block. This way, it wont break the two existing test cases in ```TldFanoutAction```

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 118 at r2](https://reviewable.io/reviews/google/nomulus/1448#-MqbYbnR8Po-ES1lOSPk:-Mr5JCJxBP7QDxkYBz3l:brm966q) ([raw file](https://github.com/google/nomulus/blob/8bf21543752cdae893960ec29a9d6837eec35740/util/src/main/java/google/registry/util/CloudTasksUtils.java#L118)):*

> d and none of its callers passes a null parameter, the check is not strictly necessary. But it is still good practice, and by moving the nullness and emptyness check before construction of encodedParams we bypass some unnecessary code.

the latest commit has ```if(method === POST) ``` twice, one as itself and another inside the null & empty conditional block. This way, it wont break the two existing test cases in ```TldFanoutAction```
</blockquote></details>

OK I see what you meant. So I gave it some thoughts and I think when there's no param to set (hence no body) in a POST,  we should not set the content type because there is no content. However regardless of if there is a body we should always set the [`Content-Length`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) to be equal to the size of the body. App engine can work without this header because it can determine the content length when the connection is terminated, but other web server might fail to parse a POST without this header when a connection is long lived. For best practice it is always a good idea to explicitly set `Content-Length`.

If there are tests that fail after you make the change, we should modify the tests.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[util/src/test/java/google/registry/util/CloudTasksUtilsTest.java, line 104 at r5](https://reviewable.io/reviews/google/nomulus/1448#-Mr4CGNv8ajclRLT8yLP:-Mr5TBYBAiT39ctkarlY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3ad7ecc33f74a9a2091cc4a44b2bf122105ba245/util/src/test/java/google/registry/util/CloudTasksUtilsTest.java#L104)):*

You can use `isEmpty()`. Same below.
</blockquote></details>

Done.

jianglai (changes requested):
> Reviewed 1 of 1 files at r7, 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @rachelguan)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 240 at r8](https://reviewable.io/reviews/google/nomulus/1448#-Mt8l6Pg1b3AIsP5flgY:-Mt8l6Pg1b3AIsP5flgZ:b-ubfili) ([raw file](https://github.com/google/nomulus/blob/99184a7e07f6ec593c33e3fca05d5f9a1a63c347/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L240)):*
> ```Java
>         query = task.getAppEngineHttpRequest().getBody().toString(StandardCharsets.UTF_8);
>       }
>       if (query != null) {
> ```

You can check `query.isEmpty()`, instead of in the if statement. It's easier for readers to understand why the check is necessary this way.



_Code quote:_
```Java
      if (query != null) {
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1448)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 240 at r8](https://reviewable.io/reviews/google/nomulus/1448#-Mt8l6Pg1b3AIsP5flgY:-Mt9W_ZM9C4a4CCkE-rj:b76h1zf) ([raw file](https://github.com/google/nomulus/blob/99184a7e07f6ec593c33e3fca05d5f9a1a63c347/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L240)):*

You can check `query.isEmpty()`, instead of in the if statement. It's easier for readers to understand why the check is necessary this way.
</blockquote></details>

I actually removed the variable `query` and assigned the string directly to `paramBuilder` since there's no use case for it to be `null`.

PR #1453, opened 2021-12-07
Add more options to the generate escrow deposit command

jianglai:
This adds two new options:

1) An option to run RDE in lenient mode.
2) An option to run RDE with the new Beam pipeline regardless of the datastore setting.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1453)

### DIFF
--- a/core/src/main/java/google/registry/rde/RdeModule.java
+++ b/core/src/main/java/google/registry/rde/RdeModule.java
@@ -45,6 +45,7 @@ public abstract class RdeModule {
   public static final String PARAM_WATERMARK = "watermark";
   public static final String PARAM_WATERMARKS = "watermarks";
   public static final String PARAM_MANUAL = "manual";
+  public static final String PARAM_BEAM = "beam";
   public static final String PARAM_DIRECTORY = "directory";
   public static final String PARAM_MODE = "mode";
   public static final String PARAM_REVISION = "revision";
@@ -70,6 +71,12 @@ public abstract class RdeModule {
   @Parameter(PARAM_MANUAL)
   static boolean provideManual(HttpServletRequest req) {
     return extractBooleanParameter(req, PARAM_MANUAL);
+  }
+
+  @Provides
+  @Parameter(PARAM_BEAM)
+  static boolean provideBeam(HttpServletRequest req) {
+    return extractBooleanParameter(req, PARAM_BEAM);
   }
 
   @Provides
--- a/core/src/main/java/google/registry/rde/RdeStagingAction.java
+++ b/core/src/main/java/google/registry/rde/RdeStagingAction.java
@@ -264,6 +264,11 @@ public final class RdeStagingAction implements Runnable {
   @Inject @Config("beamStagingBucketUrl") String stagingBucketUrl;
   @Inject @Config("rdeBucket") String rdeBucket;
   @Inject @Parameter(RdeModule.PARAM_MANUAL) boolean manual;
+
+  @Inject
+  @Parameter(RdeModule.PARAM_BEAM)
+  boolean beam;
+
   @Inject @Parameter(RdeModule.PARAM_DIRECTORY) Optional<String> directory;
   @Inject @Parameter(RdeModule.PARAM_MODE) ImmutableSet<String> modeStrings;
   @Inject @Parameter(RequestParameters.PARAM_TLDS) ImmutableSet<String> tlds;
@@ -289,7 +294,7 @@ public final class RdeStagingAction implements Runnable {
       logger.atInfo().log("Pending deposit: %s", pending);
     }
     ValidationMode validationMode = lenient ? LENIENT : STRICT;
-    if (tm().isOfy()) {
+    if (tm().isOfy() && !beam) {
       RdeStagingMapper mapper = new RdeStagingMapper(validationMode, pendings);
       RdeStagingReducer reducer = reducerFactory.create(validationMode, gcsUtils);
       mrRunner
@@ -381,6 +386,9 @@ public final class RdeStagingAction implements Runnable {
     }
     if (revision.isPresent()) {
       throw new BadRequestException("Revision parameter not allowed in standard operation");
+    }
+    if (beam) {
+      throw new BadRequestException("Beam parameter not allowed in standard operation");
     }
 
     return ImmutableSetMultimap.copyOf(
--- a/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java
@@ -71,16 +71,14 @@ final class GenerateEscrowDepositCommand implements CommandWithRemoteApi {
       names = {"-l", "--lenient"},
       description =
           "Whether to run RDE in LENIENT mode, which omits validation of the generated "
-              + "XML deposit files.",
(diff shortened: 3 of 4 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1453)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVprxw2-RZ80pG3dLH:-MqVprxw2-RZ80pG3dLI:bgm4a0u) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L75)):*
> ```Java
>           "Whether to run RDE in LENIENT mode, which omits validation of the generated "
>               + "XML deposit files.",
>       arity = 1)
> ```

You don't need to specify an arity parameter for a boolean that defaults to false, and I'd argue that you shouldn't: easier for the user to just pass "-l" than "-l true", for example.

It looks like we have a lot of parameters that are like this, but they're also all long parameters, not single character, as you've defined.  It makes less sense for a single character flag, where you're valuing brevity over readability.

Having said all of this, I'll leave it to you to make the call :-)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1453)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @mindhog)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVprxw2-RZ80pG3dLH:-MqVvRid1RMJpbGBPzwy:bgqwu2z) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L75)):*

You don't need to specify an arity parameter for a boolean that defaults to false, and I'd argue that you shouldn't: easier for the user to just pass "-l" than "-l true", for example.

It looks like we have a lot of parameters that are like this, but they're also all long parameters, not single character, as you've defined.  It makes less sense for a single character flag, where you're valuing brevity over readability.

Having said all of this, I'll leave it to you to make the call :-) 

</blockquote></details>

Agreed that these can just be `boolean`, especially as you're already assigning a default value of `false` to them so they can never be `null` (and if they are, they're gonna NPE anyway when `toString()` is called on them below).
___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 140 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVvx4-41MTLdXGtT2F:-MqVvx404UVyZmYiZ9Jo:b-81bgpl) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L140)):*
> ```Java
>             .param(PARAM_MODE, mode.toString())
>             .param(PARAM_DIRECTORY, outdir)
>             .param(PARAM_LENIENT, lenient.toString())
> ```

Make these fields `boolean` and then switch these to `Boolean.toString(lenient)`.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1453)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVprxw2-RZ80pG3dLH:-MqX7Xua18_qyJgWIjUo:b4e32jx) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L75)):*

Agreed that these can just be `boolean`, especially as you're already assigning a default value of `false` to them so they can never be `null` (and if they are, they're gonna NPE anyway when `toString()` is called on them below).
</blockquote></details>

The reason I added `arity = 1` is so that this flag can be used similarly with other ones like `--beam true --tld dev`. You don't have to stop to think about which one needs an argument and which one does not.

But since you both feel the other way, I'm removing `artiry`.
___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 140 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVvx4-41MTLdXGtT2F:-MqX8Ue86HzIxJXX3C3j:bhddvla) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L140)):*

Make these fields `boolean` and then switch these to `Boolean.toString(lenient)`.
</blockquote></details>

What's wrong with calling `Boolean.toString()` directly?

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1453)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1453)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVprxw2-RZ80pG3dLH:-Mqa5-tT4BmbE5jxsC03:byu8yif) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L75)):*

The reason I added `arity = 1` is so that this flag can be used similarly with other ones like `--beam true --tld dev`. You don't have to stop to think about which one needs an argument and which one does not.

But since you both feel the other way, I'm removing `artiry`.
</blockquote></details>

It's not the arity I have qualms with quite so much as the fact that these should be `boolean` (not `Boolean` as they currently are).  You should only use the boxed version of primitives when it's actually required, and it's not required here.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1453)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1453#-MqVprxw2-RZ80pG3dLH:-MqaMTfm-yImMApCxRTk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/446c1c7ae925e764df499f0366598dec7c613929/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L75)):*

It's not the arity I have qualms with quite so much as the fact that these should be `boolean` (not `Boolean` as they currently are).  You should only use the boxed version of primitives when it's actually required, and it's not required here.
</blockquote></details>

Done.

PR #1457, opened 2021-12-09
Copy into PersistentSets in Domains if applicable

gbrodman:
This is similar to https://github.com/google/nomulus/pull/1456

It is possible that in some cases we could get an exception:

Caused by: org.hibernate.HibernateException: A collection with cascade="all-delete-orphan" was no longer referenced by the owning entity instance: [parent]

The main cause of this, according to research (StackOverflow :P) is that
when Hibernate is calling the setters for these sets of children it's
losing the connection to the previously-managed child entity (which it
needs, in order to know how to delete orphans). Thus, the solution is to
maintain the same instance of the persistent set and just add/remove
to/from it as necessary.

This is complicated by the fact that sometimes the setter is given the
persistent set (the one we want to keep) and sometimes (?) it isn't.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1457)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainContent.java
+++ b/core/src/main/java/google/registry/model/domain/DomainContent.java
@@ -86,6 +86,7 @@ import javax.persistence.Embedded;
 import javax.persistence.MappedSuperclass;
 import javax.persistence.PostLoad;
 import javax.persistence.Transient;
+import org.hibernate.collection.internal.PersistentSet;
 import org.joda.time.DateTime;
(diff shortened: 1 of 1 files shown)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1457#-:-Mq_kUNo8l1kssPn6Yca:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #1465, opened 2021-12-13
Use CloudTaskUtils to enqueue in GenerateEscrowDepositCommand

rachelguan:


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1465)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1465#-Mr5Es9j6iSaebsgVZ4d:-Mr5Es9j6iSaebsgVZ4e:b4h630i) ([raw file](https://github.com/google/nomulus/blob/fc8948d5d9c05c1ddb7745af07a0cd9f0c0fd56c/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L124)):*
> ```Java
>     ImmutableMultimap.Builder<String, String> paramsBuilder =
>         new ImmutableMultimap.Builder<String, String>()
>             .put("Host", hostname)
> ```

As I mentioned in the other PR, when you convert a header to a param, you need to find out where is that header being used and make sure whatever consumes the header is also changed.

You can see here that the original code did not use `TaskQueueUtils`, but rather used `Queue.add()` to directly enqueue a task. That is because the util class only works in App Engine and it will always schedule the action on the same service from which the task is enqueued. When this command executes it is proxied to the tools service but we need to run the action on backend, hence the manual set up of the host header override, which is a [standard HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) used to specify the destination hostname (obviously). This is not something that our code controls.

Luckily the `CloudTasksUtils` allows you to explicitly set the app engine service, so that's all you need to do. You can remove the Host param and everything should still work. You can also remove the comment about "invoking it in a different way".

Lastly for changes like you you should deploy to alpha and test. In your PR description you can add a line like:

`TESTED=deployed to alpha and verified the command enqueues a task that runs`.



_Code quote:_
```Java
.put("Host", hostname)
```
___
*[core/src/test/java/google/registry/tools/GenerateEscrowDepositCommandTest.java, line 203 at r1](https://reviewable.io/reviews/google/nomulus/1465#-Mr5Hul497zejx4cKkzH:-Mr5Hul497zejx4cKkzI:b-23ym4m) ([raw file](https://github.com/google/nomulus/blob/fc8948d5d9c05c1ddb7745af07a0cd9f0c0fd56c/core/src/test/java/google/registry/tools/GenerateEscrowDepositCommandTest.java#L203)):*
> ```Java
>         new TaskMatcher()
>             .url("/_dr/task/rdeStaging")
>             .param("Host", "backend.test.localhost")
> ```

Need to remove all the Host params.



_Code quote:_
```Java
param("Host", "backend.test.localhost")
```

jianglai (commented):
I'm not sure what the status on this PR is. I saw you made some changes but there's no comment. Is it ready for review? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1465#-:-MrXZta-1YwLSgAeACVH:b-30r4c8)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1465)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1465#-Mr5Es9j6iSaebsgVZ4d:-Ms18gH45sV_x7YC-Uy3:b-8pdgc9) ([raw file](https://github.com/google/nomulus/blob/fc8948d5d9c05c1ddb7745af07a0cd9f0c0fd56c/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L124)):*

As I mentioned in the other PR, when you convert a header to a param, you need to find out where is that header being used and make sure whatever consumes the header is also changed.

You can see here that the original code did not use `TaskQueueUtils`, but rather used `Queue.add()` to directly enqueue a task. That is because the util class only works in App Engine and it will always schedule the action on the same service from which the task is enqueued. When this command executes it is proxied to the tools service but we need to run the action on backend, hence the manual set up of the host header override, which is a [standard HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) used to specify the destination hostname (obviously). This is not something that our code controls.

Luckily the `CloudTasksUtils` allows you to explicitly set the app engine service, so that's all you need to do. You can remove the Host param and everything should still work. You can also remove the comment about "invoking it in a different way".

Lastly for changes like you you should deploy to alpha and test. In your PR description you can add a line like:

`TESTED=deployed to alpha and verified the command enqueues a task that runs`.
</blockquote></details>

removed the header. 

I deployed to alpha and tested. The information that went to the rde-report was same as the information that usually goes to rde-report. The difference I spotted was the content length in header. 
___
*[core/src/test/java/google/registry/tools/GenerateEscrowDepositCommandTest.java, line 203 at r1](https://reviewable.io/reviews/google/nomulus/1465#-Mr5Hul497zejx4cKkzH:-Mr5husg9s8NURwGUlug:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fc8948d5d9c05c1ddb7745af07a0cd9f0c0fd56c/core/src/test/java/google/registry/tools/GenerateEscrowDepositCommandTest.java#L203)):*

Need to remove all the Host params.
</blockquote></details>

Done.

jianglai (approved):
:lgtm: Needs to check the RDE staging action can be started by the enqueued task. Otherwise LGTM.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1465#-:-MshFD_ABXuq0PVatYwK:byqbty5)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java, line 124 at r1](https://reviewable.io/reviews/google/nomulus/1465#-Mr5Es9j6iSaebsgVZ4d:-MshEtUgDVeC7CcZGcU9:btl1g5d) ([raw file](https://github.com/google/nomulus/blob/fc8948d5d9c05c1ddb7745af07a0cd9f0c0fd56c/core/src/main/java/google/registry/tools/GenerateEscrowDepositCommand.java#L124)):*

removed the header. 

I deployed to alpha and tested. The information that went to the rde-report was same as the information that usually goes to rde-report. The difference I spotted was the content length in header. 
</blockquote></details>

Did you also check if the enqueued task successfully kicked off an `RdeStagingAction`? That's all that matters in the end. I think it should, but it's worth confirming (and stating in the PR description).

PR #1467, opened 2021-12-13
Use CloudTasksUtils to enqueue in RegistrarSettingsAction

rachelguan:
Currently, a sync registrar sheet task is enqueued in ```RegistrarSettingsAction``` by calling ```enqueueRegistrarSheetSync()``` in ```SyncRegistrarsSheetAction``` . This method is only used once in the codebase and it should be moved to where it's being used, which is ```RegistrarSettingsAction```.  This method is used once and it's a one-liner, the method call is now replaced with ```cloudTasksUtils.enqueue()```. If this is used by others in the same file or it needs to be extended, then it should be taken out as a method again.


jianglai (changes requested):
You should include the stack trace of the failed tests for it to be useful for debugging. Also " The difference between this PR and the previous ones" is a bit confusing because it is not clear which PR "previous one" refers to.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467#-:-MusNpxtBUp4ELkvC5L9:bqkuuoa)* status: all files reviewed, 3 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 613 at r1](https://reviewable.io/reviews/google/nomulus/1467#-MusSOQ51tFbWnCXAvRT:-MusSOQ51tFbWnCXAvRU:bmjlkxb) ([raw file](https://github.com/google/nomulus/blob/f9516009e950c3dc32c6c3a9c01d249e977d20f0/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L613)):*
> ```Java
>                     .getAppEngineHttpRequest()
>                     .toBuilder()
>                     .putHeaders("Host", appEngineServiceUtils.getCurrentVersionHostname("backend")))
> ```

This is not necessary. If you recall we've gone through this discussion before. The reason that the old code does not use `TaskQueueUtils` but use the `Queue` API directly is that it needs to enqueue a task for a different service (which `TaskQueueUtils` does not support). With `CloudTasksUtils` you have to explicitly specify the service anyway.



_Code quote:_
```Java
                    .putHeaders("Host", appEngineServiceUtils.getCurrentVersionHostname("backend")))
```
___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 232 at r1](https://reviewable.io/reviews/google/nomulus/1467#-MusSx642E5uJ83PGLGx:-MusSx65-3GOh3RUaw95:b-ry5dyx) ([raw file](https://github.com/google/nomulus/blob/f9516009e950c3dc32c6c3a9c01d249e977d20f0/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L232)):*
> ```Java
>       // Note that UriParameters.parse() does not throw an IAE on a bad query string (e.g. one
>       // where parameters are not properly URL-encoded); it always does a best-effort parse.
>       if (method == HttpMethod.GET && uri.getQuery() != null) {
> ```

Is this new in this PR? I think it is part of an already merged PR. You probably want to rebase to pick up the latest changes from master.



_Code quote:_
```Java
 && uri.getQuery() != null)
```
___
*[core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java, line 78 at r1](https://reviewable.io/reviews/google/nomulus/1467#-MusTIbF3hE31fEuEC_-:-MusTIbF3hE31fEuEC_0:b-jq7wxy) ([raw file](https://github.com/google/nomulus/blob/f9516009e950c3dc32c6c3a9c01d249e977d20f0/core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java#L78)):*
> ```Java
>             .url(SyncRegistrarsSheetAction.PATH)
>             .method(HttpMethod.GET)
>             .header("Host", "backend.hostname"));
> ```

As mentioned above, you should test for the service here.



_Code quote:_
```Java
.header("Host", "backend.hostname"));
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 613 at r1](https://reviewable.io/reviews/google/nomulus/1467#-MusSOQ51tFbWnCXAvRT:-Muvz4lt9B8_SY7Uap41:b-oc5lqb) ([raw file](https://github.com/google/nomulus/blob/f9516009e950c3dc32c6c3a9c01d249e977d20f0/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L613)):*

This is not necessary. If you recall we've gone through this discussion before. The reason that the old code does not use `TaskQueueUtils` but use the `Queue` API directly is that it needs to enqueue a task for a different service (which `TaskQueueUtils` does not support). With `CloudTasksUtils` you have to explicitly specify the service anyway.
</blockquote></details>

Yes I recall and after trying multiple things, i just reverted it to a direct conversion. Removed that and still couldn't pass the test. 
___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 232 at r1](https://reviewable.io/reviews/google/nomulus/1467#-MusSx642E5uJ83PGLGx:-Muv3UFTDlFfDuOFmw70:b61s57d) ([raw file](https://github.com/google/nomulus/blob/f9516009e950c3dc32c6c3a9c01d249e977d20f0/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L232)):*

Is this new in this PR? I think it is part of an already merged PR. You probably want to rebase to pick up the latest changes from master.
</blockquote></details>

This is new. i did not modify the if statement for ```GET``` method. It fails at this specific case since there's no query in the url, path only. 
___
*[core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java, line 78 at r1](https://reviewable.io/reviews/google/nomulus/1467#-MusTIbF3hE31fEuEC_-:-MuvI80v4TpDs7yD76-8:bh4z3bj) ([raw file](https://github.com/google/nomulus/blob/f9516009e950c3dc32c6c3a9c01d249e977d20f0/core/src/test/java/google/registry/ui/server/registrar/RegistrarSettingsActionTest.java#L78)):*

As mentioned above, you should test for the service here.
</blockquote></details>

yep. I recalled the previous conversation. Before the version you see right here, I've tried the following to see if i could pass the ui console test:
1) put header info in params 
2) not include header info at all 
3) direct conversion 

So, what you saw was just showing the progress. I added a service check here and removed the header check.

jianglai (commented):
OK, so when we say to include the stack trace, it's much easier to include the link to the fusion or result store page of the failed run, than to copy & paste it here. That way the reviewer has much more information to work with. In this case, you did not copy the entire stack trace. There are still some useful lines after:

```
Feb 03, 2022 3:28:20 AM google.registry.request.RequestMetrics record
INFO: Action called for path=/registrar-settings, method=POST, authLevel=USER, success=true. Took: 3.025s.
Feb 03, 2022 3:28:20 AM google.registry.module.ServletBase service
INFO: Finished FrontendServlet request. Latency: 3.027s.
Feb 03, 2022 3:28:20 AM org.mortbay.log.Slf4jLog info
INFO: Stopped SocketConnector@100.83.60.24:56482

value of: getEmailAddress()
expected: test1@example.com
but was : the.registrar@example.com
value of: getEmailAddress()
expected: test1@example.com
but was : the.registrar@example.com
	at google.registry.webdriver.RegistrarConsoleWebTest.testWhoisSettingsEdit(RegistrarConsoleWebTest.java:165)
	at [[Reflective call: 4 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at [[Testing framework: 28 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at [[Testing framework: 9 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at [[Testing framework: 19 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.processAllTestClasses(JUnitPlatformTestClassProcessor.java:99)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor$CollectAllTestClassesExecutor.access$000(JUnitPlatformTestClassProcessor.java:79)
	at org.gradle.api.internal.tasks.testing.junitplatform.JUnitPlatformTestClassProcessor.stop(JUnitPlatformTestClassProcessor.java:75)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:61)
	at [[Reflective call: 4 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at com.sun.proxy.$Proxy5.stop(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.stop(TestWorker.java:133)
	at [[Reflective call: 4 frames collapsed (https://goo.gl/aH3UyP)]].(:0)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:182)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:164)
	at org.gradle.internal.remote.internal.hub.MessageHub$Handler.run(MessageHub.java:414)
	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
	at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:48)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:56)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

So you can see that the test ultimately failed because the assertion failed. This is because in this test (`testWhoisSettingsEdit`), you are simulating a user interacting with the registrar console with webdriver. The change in whois settings triggered the server to enqueque a task to run the `SyncRegistrarsSheetAction`. The server runs in a separate thread, so when it fails you get the error output (like the permission error) but it does not actually make your test fail.

The action `RegistrarSettingsAction` failed to enqueue the task because the test server runs within an emulated App Engine environment, which understands Task Queues (which is an App Engine specific feature and is simulated by the App Engine SDK's test environment). It does not however know what to do with Cloud Tasks, so when the test server enqueues the task it tries to send it to the actual Cloud Tasks endpoint and fails (which you can see in the error message `Permission denied on 'locations/registry-location-id'`, with `registry-location-id` being the placeholder GCP project name that we define in the default config file).

Normally we'd use `CloudTasksHelper` to provide a fake `CloudTasksUtils` to inject into the action that uses it in tests, but here the action is instantiated by Dagger so you have to get creative here if you really want to deal with the tasks gracefully.  However I don't think it is necessary - the test itself failed only because the enqueueing happens within the same transaction in which the updated properties are saved in the database. When it failed, none of the changes persisted, leading to the assertion error, which is checking exactly whether the changes are saved. The test itself is not concerned with the enqueueing is successful, or if the enqueued task gets executed. So I think you can just check if we are in a test environment and skip the enqueueing if we are.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467#-:-MuxbzrFAq073wkOOpb_:b-xrjn8p)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @jianglai)

rachelguan (commented):
I see. Here's the link to stack trace[https://source.cloud.google.com/results/invocations/201af635-4c88-4bb7-a6b5-e336368cdac5/targets/nomulus%2Fcore%2Fbuild%2Ftest-results%2FfragileTest/tests]

I decided to copy and paste a portion of the whole trace since I thought I included the important part (3 attempts, with most of the info from one attempt). Next time I would include a link to the stack trace and include briefly my understanding of it ! 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467#-:-Mv-vGwZAaWJIFByOxD0:b2zzi85)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @jianglai)

rachelguan (commented):
just saw a typo from earlier comment

[link to stack trace](https://source.cloud.google.com/results/invocations/201af635-4c88-4bb7-a6b5-e336368cdac5/targets/nomulus%2Fcore%2Fbuild%2Ftest-results%2FfragileTest/tests)

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467#-:-Mv0UUNy3Asvj5Wfv9g3:b-s67lh)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @jianglai)

jianglai (dismissed):
It seems like you still need to get used to the workflow. Reviewable will save your comments as you type them so you don't have to send them over when the PR itself is not ready. You should have fixed the failures with the two tests and then hit the green "Publish" button. The point I'm trying to make here is that when you "publish", all of the outstanding comments should have been addressed. There is no need to hurriedly reply to a comment just to explain why you didn't include the stack trace, when the comment also suggested a fix to your problem.

Also when I suggested to say "PTAL" when you want the reviewer to take a look, that is more of an exception than a rule. When the review gets a bit out of control and no one is sure who should be working on it, saying "PTAL" clearly signals to the other party that the ball is in their court. However this should normally not happen as long as everyone sticks to the "address all the comments and make all the corresponding code changes before hitting the reply/publish button" principle. And when "PTAL" is used, things should reset and go back to the normal state of one reply/publish per round.

If you still have some questions regarding how to efficiently do code reviews, I'm happy to chat :D

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467#-:-Mv1gec52rB3a0pDX3eF:bfx9wul)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

rachelguan (commented):
I tried to reply to your comment but I accidentally sent with a typo .. and  I wanted to include the link in this conversation. I will keep the idea of rounds in mind. 

I took a look at the existing setup of the console test and I think the quickest way to bypass the enqueueing is to add a global variable in the action file. It remains true at all time unless it's a UI test that requires to skip enqueuing, like the two test cases mentioned above. It should set the value back to true at the end of test case. it does not look pretty but it gets the job done. 

I tried to do it based on the environment but I was not able to find the indicator that shows f enqueueing is triggered by registrar console test. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467#-:-Mv0fPGI6w9hbKoHdd4q:b-q9h5dq)* status: 4 of 8 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 106 at r3](https://reviewable.io/reviews/google/nomulus/1467#-Mv6hSGh7vI6-5Abt-Y-:-Mv6hSGi5P6dB-pzahfb:b-7q0idm) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L106)):*
> ```Java
>    * transaction. It should be false at the beginning of the test case then back to true.
>    */
>   public static Boolean ENABLE_CLOUD_TASKS_UTILS = true;
> ```

What you said about using the environment and resetting the flag back to the default value are both correct.

But know that it's usually a bad practice to have a public flag like this when anyone not remotely related to this action can just go and change it, this is poor encapsulation.

What's more, even if you did clean up, there is still a chance that a concurrent unit test for `RegistrarSettingsAction` will see the altered flag and fail, resulting in flakiness.

So here you need something thread local. I think a better way to do this is to have a `ThreadLocal` flag called something like `isInTestServer` in [`TestServer`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/server/TestServer.java) thst is set to true by the thread that [starts the server](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/server/TestServer.java;l=99), and you can check for this value when you try to enqueue. The flag should also be only accessed by setter and getter so it will not be randomly changed by other objects.

This way the flag is only ever true inside the test server, which is when you want to disable the enqueueing. Unit tests are not affected. As an added bonus you will not need make changes in the tests themselves to flip this flag -- they should just work.



_Code quote:_
```Java
public static Boolean ENABLE_CLOUD_TASKS_UTILS = true;
```

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 619 at r3](https://reviewable.io/reviews/google/nomulus/1467#-MvKBnh-95u9YXG35bal:-MvKBnh-95u9YXG35bam:b-78mnys) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L619)):*
> ```Java
>       cloudTasksUtils.enqueue(
>           SyncRegistrarsSheetAction.QUEUE,
>           CloudTasksUtils.createGetTask(
> ```

This is called from inside a transaction, so there is a risk of spamming the recipient if the transaction retries.



_Code quote:_
```Java
    if (ENABLE_CLOUD_TASKS_UTILS) {
      cloudTasksUtils.enqueue(
          SyncRegistrarsSheetAction.QUEUE,
          CloudTasksUtils.createGetTask(
              SyncRegistrarsSheetAction.PATH, Service.BACKEND.toString(), ImmutableMultimap.of()));
    }
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 106 at r3](https://reviewable.io/reviews/google/nomulus/1467#-Mv6hSGh7vI6-5Abt-Y-:-MvLbI0A2KQcVsdmnoZA:bo6jguz) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L106)):*

What you said about using the environment and resetting the flag back to the default value are both correct.

But know that it's usually a bad practice to have a public flag like this when anyone not remotely related to this action can just go and change it, this is poor encapsulation.

What's more, even if you did clean up, there is still a chance that a concurrent unit test for `RegistrarSettingsAction` will see the altered flag and fail, resulting in flakiness.

So here you need something thread local. I think a better way to do this is to have a `ThreadLocal` flag called something like `isInTestServer` in [`TestServer`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/server/TestServer.java) thst is set to true by the thread that [starts the server](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/server/TestServer.java;l=99), and you can check for this value when you try to enqueue. The flag should also be only accessed by setter and getter so it will not be randomly changed by other objects.

This way the flag is only ever true inside the test server, which is when you want to disable the enqueueing. Unit tests are not affected. As an added bonus you will not need make changes in the tests themselves to flip this flag -- they should just work.
</blockquote></details>

Agreed and replaced with ```ThreadLocal<Boolean>``` variable in ```RegistrarSettingsAction``` and the value is modified in ```TestServer```.  It's indeed better than modifying unit tests for sure!
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 619 at r3](https://reviewable.io/reviews/google/nomulus/1467#-MvKBnh-95u9YXG35bal:-MvLJVUi325r7w30AAK7:b-bkqy7q) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L619)):*

This is called from inside a transaction, so there is a risk of spamming the recipient if the transaction retries.
</blockquote></details>

Yes, I'm thinking to separate the actual email sending from the transaction. Send an email only if the transaction is done successfully, which would be from loading reigstrar info to enqueueing the task. Do you think that would solve the problem?

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 619 at r3](https://reviewable.io/reviews/google/nomulus/1467#-MvKBnh-95u9YXG35bal:-MvMNkgtAmtOXAKOZZvh:b-igi9ll) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L619)):*

Yes, I'm thinking to separate the actual email sending from the transaction. Send an email only if the transaction is done successfully, which would be from loading reigstrar info to enqueueing the task. Do you think that would solve the problem? 
</blockquote></details>

I think that should be fine. The downside is that when the enqueue fails the transaction is not rolled back. We can probably add a cron job to periodically sync the sheet, which is idempotent anyway.
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 109 at r4](https://reviewable.io/reviews/google/nomulus/1467#-MvMNR7c1KMxXY1zNYbp:-MvMNR7c1KMxXY1zNYbq:b-fndk35) ([raw file](https://github.com/google/nomulus/blob/03e58d8e025133250530500b648756f93497d847/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L109)):*
> ```Java
>    * testing.
>    */
>   private static ThreadLocal<Boolean> IS_IN_TEST_DRIVER = ThreadLocal.withInitial(() -> false);
> ```

It's not a constant per se. The custom is to use camelCase here.



_Code quote:_
```Java
IS_IN_TEST_DRIVER
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 619 at r3](https://reviewable.io/reviews/google/nomulus/1467#-MvKBnh-95u9YXG35bal:-MvLjBRj68l1P0MohtHC:b-4txo7u) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L619)):*

I think that should be fine. The downside is that when the enqueue fails the transaction is not rolled back. We can probably add a cron job to periodically sync the sheet, which is idempotent anyway.
</blockquote></details>

I took a closer look at the code and I made some changes to ```update()```. 

Regarding Weimin's comment, I think what we could do is to take out email sending and enqueueing from the transaction. The return type of method ```sendExternalUpdatesIfNecessary```can be boolean instead of void. We could take advantage of that to decide if enqueuing should happen. It returns true if an email is sent successfully, otherwise, it return false. Cases that return false include no recipients, existing contacts is empty, no difference except lastUpdateTime, and exceptions. 

To handle the case of exception with no enqueuing, like Lai said, we can add a cron job to periodically sync the sheet.  
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 109 at r4](https://reviewable.io/reviews/google/nomulus/1467#-MvMNR7c1KMxXY1zNYbp:-MvP6qMTBnKJTg0jevER:by48ej2) ([raw file](https://github.com/google/nomulus/blob/03e58d8e025133250530500b648756f93497d847/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L109)):*

It's not a constant per se. The custom is to use camelCase here.
</blockquote></details>

changed it.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: all files reviewed, 4 unresolved discussions (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 619 at r3](https://reviewable.io/reviews/google/nomulus/1467#-MvKBnh-95u9YXG35bal:-MvRSOtWAMCSdj25mj1j:b-xi2r5j) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L619)):*

I took a closer look at the code and I made some changes to ```update()```. 

Regarding Weimin's comment, I think what we could do is to take out email sending and enqueueing from the transaction. The return type of method ```sendExternalUpdatesIfNecessary```can be boolean instead of void. We could take advantage of that to decide if enqueuing should happen. It returns true if an email is sent successfully, otherwise, it return false. Cases that return false include no recipients, existing contacts is empty, no difference except lastUpdateTime, and exceptions. 

To handle the case of exception with no enqueuing, like Lai said, we can add a cron job to periodically sync the sheet.  
</blockquote></details>

The cron job needs to be added in this PR.
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 237 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRTWPjAiNH5h3Bz4BM:-MvRTWPjAiNH5h3Bz4BN:b-6taoqs) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L237)):*
> ```Java
>     EmailInfo emailInfo = tm().transact(() -> saveUpdates(args, registrarId));
>     // Email the updates
>     if (sendExternalUpdatesIfNecessary(
> ```

Looks like you can just pass `EmailInfo` here and process it inside the function.



_Code quote:_
```Java
sendExternalUpdatesIfNecessary
```
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 242 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRR7Ni3cb1D7ZaJ81H:-MvRR7Ni3cb1D7ZaJ81I:b62d213) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L242)):*
> ```Java
>                 emailInfo.updatedRegistrar(),
>                 emailInfo.updatedContacts())
>             == true
> ```

No need to compare booleans to `true` and `false`. Just use the boolean or prepend it with `!`.



_Code quote:_
```Java
== true
```
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 243 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRRqz58ls8dx4QZY7R:-MvRRqz58ls8dx4QZY7S:bcl599l) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L243)):*
> ```Java
>                 emailInfo.updatedContacts())
>             == true
>         && isInTestDriver.get() == false) {
> ```

The only logic only enqueues the task when the email is sent out successfully, but that's by virtue of them both being in the transaction. It is not important that they both happen or not happen. What is important is that the database change must be committed before the email is sent and before the task is enqueued.

I think you can just do them dependently. One by one.



_Code quote:_
```Java
&& isInTestDriver.get() == false
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 619 at r3](https://reviewable.io/reviews/google/nomulus/1467#-MvKBnh-95u9YXG35bal:-MvVF2d584KOkr2BPdzA:bdht349) ([raw file](https://github.com/google/nomulus/blob/c246835fa93824352964dd593428abfe25e152cf/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L619)):*

The cron job needs to be added in this PR.
</blockquote></details>

I took a look at our codebase and took a closer look at [SynsRegistrarsSheetAction](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/export/sheet/SyncRegistrarsSheetAction.java), where the```enqueueRegistrarSheetSync``` is stored. The task triggers this action, which itself is a CRON job that serves to synchronize the registrars spreadsheet on an hourly basis. So, if even the task requires a retry, this CRON job should already be taking care of it.  
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 237 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRTWPjAiNH5h3Bz4BM:-MvVGABv2OWwetJyydoC:b-7rl0jd) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L237)):*

Looks like you can just pass `EmailInfo` here and process it inside the function.
</blockquote></details>

done.
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 242 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRR7Ni3cb1D7ZaJ81H:-MvV8CSE6sh_xYnat9g4:b-4xdmi6) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L242)):*
> ng in the transaction. It is not important that they both happen or not happen. What is important is that the database change must be committed before the email is sent and before the task is enqueued.
changed the type bak to void so no comparison is going to be done anymore. 
___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 243 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRRqz58ls8dx4QZY7R:-MvV8HNV4g_CsXntQNOt:bs9m42) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L243)):*

The only logic only enqueues the task when the email is sent out successfully, but that's by virtue of them both being in the transaction. It is not important that they both happen or not happen. What is important is that the database change must be committed before the email is sent and before the task is enqueued.

I think you can just do them dependently. One by one.
</blockquote></details>

agreed and removed the condition. 

I moved the enqueuing back into the ```sendExternalUpdatesIfNecessary``` because it should only happen if it there is at least one recipient and there's diff, and both of the checks are done in that method before calling ```sendEmail()```.

jianglai (approved):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1467)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java, line 243 at r5](https://reviewable.io/reviews/google/nomulus/1467#-MvRRqz58ls8dx4QZY7R:-MvZA9dW8Ux-vGQc1IfL:b-4tb6nf) ([raw file](https://github.com/google/nomulus/blob/7347be159a981d87be4a93754aceb5335c3e5413/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L243)):*

agreed and removed the condition. 

I moved the enqueuing back into the ```sendExternalUpdatesIfNecessary``` because it should only happen if it there is at least one recipient and there's diff, and both of the checks are done in that method before calling ```sendEmail()```. 
</blockquote></details>

Strictly speaking we should enqueue an action as long as there's a diff, regardless of whether there's a recipient. But since that is the way the code was structured before, I don't think we necessarily have to change it. Especially given the fact that the sheet will be synced once per hour anyway. The enqueueing just speeds things up a little.

PR #1468, opened 2021-12-13
Change from TaskQueueUtils to CloudTasksUtils in LoadTestAction

rachelguan:
Besides converting the type from `TaskOptions` to `Task`, there're a few other changes had to be made: 
1) Initially, the header info (`X_CSRF_TOKEN`) was put in `params`. Unlike previous cases, a few test cases failed at authentication. The quickest way to fix this was to append header to the task created via `CloudTasksUtils`. 
2)The schedule time doesn't require randomness. The time can be set to task via `setScheduleTime()` using `Instant` and `Timestamp` APIs. 

Currently, there's no test file corresponding to `LoadTestAction`. This can be tested in alpha via `LoadTestCommand`. The flows can be verified logs. We have to ensure that authentication works in `RequestAuthenticator` and the flows are triggered. At the end, `OK` message should show on terminal. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1468)

### DIFF
--- a/core/src/main/java/google/registry/loadtest/LoadTestAction.java
+++ b/core/src/main/java/google/registry/loadtest/LoadTestAction.java
@@ -28,21 +28,22 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Iterators;
 import com.google.common.flogger.FluentLogger;
+import com.google.protobuf.Timestamp;
 import google.registry.config.RegistryEnvironment;
 import google.registry.request.Action;
 import google.registry.request.Action.Service;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
 import google.registry.security.XsrfTokenManager;
-import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Optional;
 import java.util.Random;
 import java.util.function.Function;
 import javax.inject.Inject;
+import org.apache.arrow.util.VisibleForTesting;
 import org.joda.time.DateTime;
 
 /**
@@ -150,7 +151,6 @@ public class LoadTestAction implements Runnable {
   int hostInfosPerSecond;
 
   @Inject CloudTasksUtils cloudTasksUtils;
-  @Inject Clock clock;
 
   private final String xmlContactCreateTmpl;
   private final String xmlContactCreateFail;
@@ -326,27 +326,41 @@ public class LoadTestAction implements Runnable {
     return name.toString();
   }
 
+  @VisibleForTesting
   private List<Task> createTasks(List<String> xmls, DateTime start) {
     ImmutableList.Builder<Task> tasks = new ImmutableList.Builder<>();
     for (int i = 0; i < xmls.size(); i++) {
-      // Space tasks evenly within across a second.
-      int offsetMillis = (int) (1000.0 / xmls.size() * i);
-      CloudTasksUtils.createPostTask(
-          "/_dr/epptool",
-          Service.TOOLS.toString(),
-          ImmutableMultimap.of(
-              X_CSRF_TOKEN,
-              xsrfToken,
-              "clientId",
-              registrarId,
-              "superuser",
-              Boolean.FALSE.toString(),
-              "dryRun",
-              Boolean.FALSE.toString(),
-              "xml",
-              xmls.get(i)),
-          clock,
-          Optional.of((int) (start.getMillis() + offsetMillis)));
+      tasks.add(
+          Task.newBuilder()
+              .setAppEngineHttpRequest(
+                  CloudTasksUtils.createPostTask(
+                          "/_dr/epptool",
+                          Service.TOOLS.toString(),
+                          ImmutableMultimap.of(
+                              "clientId",
+                              registrarId,
+                              "superuser",
+                              Boolean.FALSE.toString(),
+                              "dryRun",
+                              Boolean.FALSE.toString(),
+                              "xml",
+                              xmls.get(i)))
+                      .toBuilder()
+                      // Space tasks evenly within across a second.
+                      .setScheduleTime(
+                          Timestamp.newBuilder()
+                              .setSeconds(
+                                  Instant.ofEpochMilli(
+                                          start
+                                              .plusMillis((int) (1000.0 / xmls.size() * i))
+                                              .getMillis())
+                                      .getEpochSecond())
+                              .build())
+                      .getAppEngineHttpRequest()
+                      .toBuilder()
+                      .putHeaders(X_CSRF_TOKEN, xsrfToken)
+                      .build())
+              .build());
     }
     return tasks.build();
   }
--- a/core/src/main/java/google/registry/loadtest/LoadTestAction.java
+++ b/core/src/main/java/google/registry/loadtest/LoadTestAction.java
@@ -43,7 +43,6 @@ import java.util.List;
 import java.util.Random;
 import java.util.function.Function;
 import javax.inject.Inject;
-import org.apache.arrow.util.VisibleForTesting;
 import org.joda.time.DateTime;
 
 /**
@@ -326,10 +325,12 @@ public class LoadTestAction implements Runnable {
     return name.toString();
   }
 
-  @VisibleForTesting
   private List<Task> createTasks(List<String> xmls, DateTime start) {
     ImmutableList.Builder<Task> tasks = new ImmutableList.Builder<>();
     for (int i = 0; i < xmls.size(); i++) {
+      // Space tasks evenly within across a second.
+      Instant scheduleTime =
+          Instant.ofEpochMilli(start.plusMillis((int) (1000.0 / xmls.size() * i)).getMillis());
       tasks.add(
           Task.newBuilder()
               .setAppEngineHttpRequest(
@@ -346,19 +347,14 @@ public class LoadTestAction implements Runnable {
                               "xml",
                               xmls.get(i)))
                       .toBuilder()
-                      // Space tasks evenly within across a second.
-                      .setScheduleTime(
-                          Timestamp.newBuilder()
-                              .setSeconds(
-                                  Instant.ofEpochMilli(
-                                          start
-                                              .plusMillis((int) (1000.0 / xmls.size() * i))
-                                              .getMillis())
-                                      .getEpochSecond())
-                              .build())
                       .getAppEngineHttpRequest()
                       .toBuilder()
                       .putHeaders(X_CSRF_TOKEN, xsrfToken)
+                      .build())
+              .setScheduleTime(
+                  Timestamp.newBuilder()
+                      .setSeconds(scheduleTime.getEpochSecond())
+                      .setNanos(scheduleTime.getNano())
                       .build())
               .build());
     }
--- a/core/src/main/java/google/registry/loadtest/LoadTestAction.java
+++ b/core/src/main/java/google/registry/loadtest/LoadTestAction.java
@@ -14,7 +14,6 @@
 
 package google.registry.loadtest;
 
-import static com.google.appengine.api.taskqueue.QueueConstants.maxTasksPerAdd;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Lists.partition;
@@ -65,6 +64,7 @@ public class LoadTestAction implements Runnable {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
   private static final int NUM_QUEUES = 10;
+  private static final int MAX_TASKS_PER_LOAD = 100;
   private static final int ARBITRARY_VALID_HOST_LENGTH = 40;
   private static final int MAX_CONTACT_LENGTH = 13;
   private static final int MAX_DOMAIN_LABEL_LENGTH = 63;
@@ -362,7 +362,7 @@ public class LoadTestAction implements Runnable {
   }
 
   private void enqueue(List<Task> tasks) {
-    List<List<Task>> chunks = partition(tasks, maxTasksPerAdd());
+    List<List<Task>> chunks = partition(tasks, MAX_TASKS_PER_LOAD);
     // Farm out tasks to multiple queues to work around queue qps quotas.
     for (int i = 0; i < chunks.size(); i++) {
       cloudTasksUtils.enqueue("load" + (i % NUM_QUEUES), chunks.get(i));
--- a/core/src/main/java/google/registry/loadtest/LoadTestAction.java
+++ b/core/src/main/java/google/registry/loadtest/LoadTestAction.java
@@ -15,7 +15,6 @@
 package google.registry.loadtest;
 
 import static com.google.appengine.api.taskqueue.QueueConstants.maxTasksPerAdd;
-import static com.google.appengine.api.taskqueue.QueueFactory.getQueue;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.Lists.partition;
@@ -24,19 +23,23 @@ import static google.registry.util.ResourceUtils.readResourceUtf8;
 import static java.util.Arrays.asList;
 import static org.joda.time.DateTimeZone.UTC;
 
-import com.google.appengine.api.taskqueue.TaskOptions;
+import com.google.cloud.tasks.v2.Task;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Iterators;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryEnvironment;
 import google.registry.request.Action;
+import google.registry.request.Action.Service;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
 import google.registry.security.XsrfTokenManager;
-import google.registry.util.TaskQueueUtils;
+import google.registry.util.Clock;
+import google.registry.util.CloudTasksUtils;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.Random;
 import java.util.function.Function;
 import javax.inject.Inject;
@@ -146,7 +149,8 @@ public class LoadTestAction implements Runnable {
   @Parameter("hostInfos")
   int hostInfosPerSecond;
 
-  @Inject TaskQueueUtils taskQueueUtils;
+  @Inject CloudTasksUtils cloudTasksUtils;
+  @Inject Clock clock;
 
   private final String xmlContactCreateTmpl;
   private final String xmlContactCreateFail;
@@ -208,7 +212,7 @@ public class LoadTestAction implements Runnable {
     ImmutableList<String> contactNames = contactNamesBuilder.build();
     ImmutableList<String> hostPrefixes = hostPrefixesBuilder.build();
 
-    ImmutableList.Builder<TaskOptions> tasks = new ImmutableList.Builder<>();
+    ImmutableList.Builder<Task> tasks = new ImmutableList.Builder<>();
     for (int offsetSeconds = 0; offsetSeconds < runSeconds; offsetSeconds++) {
       DateTime startSecond = initialStartSecond.plusSeconds(offsetSeconds);
       // The first "failed" creates might actually succeed if the object doesn't already exist, but
@@ -254,7 +258,7 @@ public class LoadTestAction implements Runnable {
                   .collect(toImmutableList()),
               startSecond));
     }
-    ImmutableList<TaskOptions> taskOptions = tasks.build();
+    ImmutableList<Task> taskOptions = tasks.build();
     enqueue(taskOptions);
     logger.atInfo().log("Added %d total load test tasks.", taskOptions.size());
   }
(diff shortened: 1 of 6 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 336 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4EZet-eP0BMMWDRlE:-Mr4EZet-eP0BMMWDRlF:bea02y5) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L336)):*
> ```Java
>       CloudTasksUtils.createPostTask(
>           "/_dr/epptool",
>           Service.BACKEND.toString(),
> ```

You need to check which service this endpoint belongs to. You can see from [`tools_routing.txt`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/resources/google/registry/module/tools/tools_routing.txt;l=10) that you need to use `Service.TOOLS.toString()` instead.



_Code quote:_
```Java
Service.BACKEND.toString()
```
___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4ExApBLz5zp1iBrYl:-Mr4ExApBLz5zp1iBrYm:b2sa0qd) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L338)):*
> ```Java
>           Service.BACKEND.toString(),
>           ImmutableMultimap.of(
>               X_CSRF_TOKEN,
> ```

This is probably not going to work. Remember when you move a pair from the headers to the params you need to make sure that whatever is processing the request is also changed accordingly. In most cases that's just changing the corresponding action to request a param instead of a header, but in this case the header is used in [authentication](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/LegacyAuthenticationMechanism.java;l=65;drc=4e6d14a8aeb7217523184f58b8e9d63aa9d9b4bc;bpv=0;bpt=1).

If you read the comment here, it is possible to use `P_CSRF_TOKEN` to save the token in parameters, but it could break things if you also have actual payload, in addition to payload constructed from params.

Since we know for sure that the request you are constructing here does not have extra payload in its body, you can try using `P_CSRF_TOKEN` here instead. You definitely should deploy to alpha to test make make sure that the enqueued task actually can run. Failing that, we may have to implement the ability to set headers in `CloudTasksUtils`.



_Code quote:_
```Java
              X_CSRF_TOKEN,
              xsrfToken,
```

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4ExApBLz5zp1iBrYl:-Mr4HxZr09JjPFwrRNrt:bz0g0n8) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L338)):*

This is probably not going to work. Remember when you move a pair from the headers to the params you need to make sure that whatever is processing the request is also changed accordingly. In most cases that's just changing the corresponding action to request a param instead of a header, but in this case the header is used in [authentication](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/LegacyAuthenticationMechanism.java;l=65;drc=4e6d14a8aeb7217523184f58b8e9d63aa9d9b4bc;bpv=0;bpt=1).

If you read the comment here, it is possible to use `P_CSRF_TOKEN` to save the token in parameters, but it could break things if you also have actual payload, in addition to payload constructed from params.

Since we know for sure that the request you are constructing here does not have extra payload in its body, you can try using `P_CSRF_TOKEN` here instead. You definitely should deploy to alpha to test make make sure that the enqueued task actually can run. Failing that, we may have to implement the ability to set headers in `CloudTasksUtils`.
</blockquote></details>

Actually we may not have to implement that in `CloudTasksUtils`. Since this is an one-off operation, we could just get a `Task` from `CloudTasksUtils` and add headers to it, right? There's no need to add a helper method if you are going to call it once.

jianglai (commented):
I'm not sure what the status on this PR is. I saw you made some changes but there's no comment. Is it ready for review? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468#-:-MrX_4bGEoceZkytdmk5:b-30r4c8)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 336 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4EZet-eP0BMMWDRlE:-Mr5llGO1XvNFEvPXJuW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L336)):*

You need to check which service this endpoint belongs to. You can see from [`tools_routing.txt`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/resources/google/registry/module/tools/tools_routing.txt;l=10) that you need to use `Service.TOOLS.toString()` instead.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4ExApBLz5zp1iBrYl:-Msfw43G28rrSj9rpzPP:bc93gq0) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L338)):*

Actually we may not have to implement that in `CloudTasksUtils`. Since this is an one-off operation, we could just get a `Task` from `CloudTasksUtils` and add headers to it, right? There's no need to add a helper method if you are going to call it once.
</blockquote></details>

tried both cases in parameter and couldn't pass the console test cases. It works if I put the headers in task. I agree I can be a one time thing if it only gets called once. Some other existing files require headers in task but I'm not sure if they have to be added as headers. If this use case shows up again, we should consider using a helper and modify this part of the code.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4ExApBLz5zp1iBrYl:-MshSTfH50PvcL2c7jnh:b-hznonh) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L338)):*

tried both cases in parameter and couldn't pass the console test cases. It works if I put the headers in task. I agree I can be a one time thing if it only gets called once. Some other existing files require headers in task but I'm not sure if they have to be added as headers. If this use case shows up again, we should consider using a helper and modify this part of the code.  
</blockquote></details>

You should probably do a search and find how many times we used headers when enqueuing tasks and see if they can all be converted to params.
___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 349 at r3](https://reviewable.io/reviews/google/nomulus/1468#-MshLVe4ChXTLc7emq87:-MshLVe5E9X5Okb5XTfn:bu5c01o) ([raw file](https://github.com/google/nomulus/blob/d79890eb3c520360c8e2e7448ef4a968ef770899/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L349)):*
> ```Java
>                           clock,
>                           // Space tasks evenly within across a second.
>                           Optional.of((int) (start.getMillis() + (int) (1000.0 / xmls.size() * i))))
> ```

This will not do what you intended. You should first read what `etaMillis` and `countdownMillis` [do in task queues](https://source.corp.google.com/piper///depot/google3/third_party/java_src/appengine_standard/api/src/main/java/com/google/appengine/api/taskqueue/QueueImpl.java;l=621;bpv=1;bpt=1;rcl=419724084?q=taskoptions&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental).

Unfortunately with cloud tasks, the API does not expose a method for use to specify countdown millis, so in `CloudTasksUitls` the `jitterSeconds` parameter adds a jitter to the current system time and use that timestamp as the schedule time (or ETA in task queue parlance). This means that the conversion that we did before where we replaced calls to `TaskOptions.Builder.withCountdownMillis` with `CloudTasksUtils.createTasks(..., jitterSeconds)` are not exactly correct, but an approximation at best, because the countdown millis should be added to the timestamp when the task is added to the queue ([potentially asynchrously](https://source.corp.google.com/piper///depot/google3/third_party/java_src/appengine_standard/api/src/main/java/com/google/appengine/api/taskqueue/QueueImpl.java;bpv=1;bpt=1;rcl=419724084;l=507?q=taskoptions&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental)), not when the tasks are created. But the two timestamps should be close enough, regardless.

In this case however, since the original code was using `etaMillis`, you can actually convert it exactly to one represented by a `Task` by calling [`setScheduleTime`](https://cs.opensource.google/nomulus/nomulus/+/master:util/src/main/java/google/registry/util/CloudTasksUtils.java;l=173;drc=f87e7eb6e67c64f75a3b865f55e78e5cb52c3a70?q=createtask&ss=nomulus%2Fnomulus), which is equavelent to `etaMillis` on `TaskOptions`.

Also since there's no unit test for this we should test it on alpha as well, at least to make sure that the authentication still works. Just because a not-directly-related unit test (console test) failed and now passed doesn't mean everything is now correct here.



_Code quote:_
```Java
                          Optional.of((int) (start.getMillis() + (int) (1000.0 / xmls.size() * i))))
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 338 at r1](https://reviewable.io/reviews/google/nomulus/1468#-Mr4ExApBLz5zp1iBrYl:-MskmAb81BpJ9G_JleIM:bhepy2x) ([raw file](https://github.com/google/nomulus/blob/b4553768908df2b25a49d68b8ed5abcc6990b734/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L338)):*

You should probably do a search and find how many times we used headers when enqueuing tasks and see if they can all be converted to params.
</blockquote></details>

there's actually a few cases that require that, which was why I wanted to create a PR to add headers.  Some could be solved by changing them as part of parameters. I think it's worth to keep how the information  is originally stored. If we want to put some of them in params, we can do that after.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 349 at r3](https://reviewable.io/reviews/google/nomulus/1468#-MshLVe4ChXTLc7emq87:-Mt4nByS74655eCiF2Fi:b-xzf0ot) ([raw file](https://github.com/google/nomulus/blob/d79890eb3c520360c8e2e7448ef4a968ef770899/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L349)):*

This will not do what you intended. You should first read what `etaMillis` and `countdownMillis` [do in task queues](https://source.corp.google.com/piper///depot/google3/third_party/java_src/appengine_standard/api/src/main/java/com/google/appengine/api/taskqueue/QueueImpl.java;l=621;bpv=1;bpt=1;rcl=419724084?q=taskoptions&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental).

Unfortunately with cloud tasks, the API does not expose a method for use to specify countdown millis, so in `CloudTasksUitls` the `jitterSeconds` parameter adds a jitter to the current system time and use that timestamp as the schedule time (or ETA in task queue parlance). This means that the conversion that we did before where we replaced calls to `TaskOptions.Builder.withCountdownMillis` with `CloudTasksUtils.createTasks(..., jitterSeconds)` are not exactly correct, but an approximation at best, because the countdown millis should be added to the timestamp when the task is added to the queue ([potentially asynchrously](https://source.corp.google.com/piper///depot/google3/third_party/java_src/appengine_standard/api/src/main/java/com/google/appengine/api/taskqueue/QueueImpl.java;bpv=1;bpt=1;rcl=419724084;l=507?q=taskoptions&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental)), not when the tasks are created. But the two timestamps should be close enough, regardless.

In this case however, since the original code was using `etaMillis`, you can actually convert it exactly to one represented by a `Task` by calling [`setScheduleTime`](https://cs.opensource.google/nomulus/nomulus/+/master:util/src/main/java/google/registry/util/CloudTasksUtils.java;l=173;drc=f87e7eb6e67c64f75a3b865f55e78e5cb52c3a70?q=createtask&ss=nomulus%2Fnomulus), which is equavelent to `etaMillis` on `TaskOptions`.

Also since there's no unit test for this we should test it on alpha as well, at least to make sure that the authentication still works. Just because a not-directly-related unit test (console test) failed and now passed doesn't mean everything is now correct here.
</blockquote></details>

fixed the schedule time. will test in alpha.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 349 at r3](https://reviewable.io/reviews/google/nomulus/1468#-MshLVe4ChXTLc7emq87:-MtLgTTDBuzpMt-NMvCH:b-8uzl2x) ([raw file](https://github.com/google/nomulus/blob/d79890eb3c520360c8e2e7448ef4a968ef770899/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L349)):*

fixed the schedule time. will test in alpha.
</blockquote></details>

As discussed at stand up, with cloud tasks it appears that enqueuing all tasks in the action takes too much time, and the command line times out waiting for a response. We should consider sending the http request asynchronously from the command line.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 349 at r3](https://reviewable.io/reviews/google/nomulus/1468#-MshLVe4ChXTLc7emq87:-MunR2YEE_uZou414w8z:b-1izkcg) ([raw file](https://github.com/google/nomulus/blob/d79890eb3c520360c8e2e7448ef4a968ef770899/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L349)):*

As discussed at stand up, with cloud tasks it appears that enqueuing all tasks in the action takes too much time, and the command line times out waiting for a response. We should consider sending the http request asynchronously from the command line.
</blockquote></details>

added a method ```sendAsyncPostRequest``` that calls ```internalAsyncSend``` which uses ```ListenableFuture```to send the request. However, I don't think I'm using it correctly, or I don't see any difference testing in alpha.

rachelguan (commented):
After some discussions that happened yesterday in the chat, I wonder what the next step for this PR to be, since we could 
1) leave it as the way it is (no async or re-architect the action since it works but slower), or 
2) find an alternative like running the test locally or using double injection 

The fact it timed out had to do with the fact that I did not specify ```--run_seconds``` which is "Time to run the load test in seconds.". The default value is 4600 which was assigned arbitrarily. 
```
  // This is a mostly arbitrary value, roughly an hour and a quarter.  It served as a generous
  // timespan for initial backup/restore testing, but has no other special significance.
  private static final int DEFAULT_RUN_SECONDS = 50000;
```

It's very possible that the time out issue actually existed in the old implementation that it was barely able to complete the command I tested, but not other cases that would require more loads of tasks. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468#-:-MuwptMH1MZHzoj5ksQg:b92jmez)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai)

jianglai (commented):
So if you set `DEFAULT_RUN_SECONDS` to a larger number things should work out without all the async stuff? I think we should just do that then.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468#-:-Mv-Me0N5C6o9EUpbwqq:b-wdpmy5)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai)

rachelguan (commented):
i think we should do that for now, at least. Re-architect should be a separate PR. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468#-:-Mv-SE-N7RlHLWO9IzdD:bvjdk45)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai)

jianglai (approved):
> Reviewed 2 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: 6 of 9 files reviewed, all discussions resolved (waiting on @jianglai)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 349 at r3](https://reviewable.io/reviews/google/nomulus/1468#-MshLVe4ChXTLc7emq87:-Mv1kZJl-PAJnc10Eeh-:brtgomt) ([raw file](https://github.com/google/nomulus/blob/d79890eb3c520360c8e2e7448ef4a968ef770899/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L349)):*

added a method ```sendAsyncPostRequest``` that calls ```internalAsyncSend``` which uses ```ListenableFuture```to send the request. However, I don't think I'm using it correctly, or I don't see any difference testing in alpha. 
</blockquote></details>

I think the comment is moot now.

jianglai (approved):
> Reviewed 1 of 3 files at r6, 2 of 4 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan and @sarahcaseybot)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 353 at r10](https://reviewable.io/reviews/google/nomulus/1468#-MvMMz13ACWxhREWiL78:-MvMMz13ACWxhREWiL79:b-xqec0v) ([raw file](https://github.com/google/nomulus/blob/d98e5ba886a4799803137b538ae0103d92134ad0/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L353)):*
> ```Java
>                       .toBuilder()
>                       // instead of adding the X_CSRF_TOKEN to params, this remains as part of
>                       // headers because of the existing setup for authentication  
> ```

Make a reference to the authentication class here.



_Code quote:_
```Java
// headers because of the existing setup for authentication···
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 353 at r10](https://reviewable.io/reviews/google/nomulus/1468#-MvMMz13ACWxhREWiL78:-MvPkapoET-4l5A0WiAS:b-rnrapz) ([raw file](https://github.com/google/nomulus/blob/d98e5ba886a4799803137b538ae0103d92134ad0/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L353)):*

Make a reference to the authentication class here.
</blockquote></details>

done.

jianglai (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468#-:-MvPq1sjAstG8Nzfu0r2:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan and @sarahcaseybot)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 353 at r10](https://reviewable.io/reviews/google/nomulus/1468#-MvMMz13ACWxhREWiL78:-MvPqNWuCRu8GnRND8NH:b-1ah3tz) ([raw file](https://github.com/google/nomulus/blob/d98e5ba886a4799803137b538ae0103d92134ad0/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L353)):*

done. 
</blockquote></details>

I don't think you can have javadoc just anywhere? This probably will not do what you want. You can just have a regular comment but spell out the class name.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai and @sarahcaseybot)

___
*[core/src/main/java/google/registry/loadtest/LoadTestAction.java, line 353 at r10](https://reviewable.io/reviews/google/nomulus/1468#-MvMMz13ACWxhREWiL78:-MvQ3CVX5u1Jcnl4ZO_Y:b2ub9l3) ([raw file](https://github.com/google/nomulus/blob/d98e5ba886a4799803137b538ae0103d92134ad0/core/src/main/java/google/registry/loadtest/LoadTestAction.java#L353)):*

I don't think you can have javadoc just anywhere? This probably will not do what you want. You can just have a regular comment but spell out the class name.
</blockquote></details>

I did that originally but it didn't seem to be clickable. I guess I could just include that as a regular comment. Fixed.

sarahcaseybot (approved):
> Reviewed 1 of 2 files at r1, 2 of 4 files at r9, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1468)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

PR #1470, opened 2021-12-15
Allow usage of a read-only Postgres replica

gbrodman:
This adds the Dagger provider code for both the regular and the BEAM
environments, which are similar but not quite the same.

In addition, this demonstrates usage of the replica DB in the
RdePipeline. I tested this on alpha with a modified version of the
RdePipeline that attempts to write some dummy values to the database and
it failed with the expected message that one cannot write to a replica.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1470)

### DIFF
--- a/core/src/main/java/google/registry/beam/common/RegistryPipelineComponent.java
+++ b/core/src/main/java/google/registry/beam/common/RegistryPipelineComponent.java
@@ -23,6 +23,7 @@ import google.registry.config.RegistryConfig.ConfigModule;
 import google.registry.persistence.PersistenceModule;
 import google.registry.persistence.PersistenceModule.BeamBulkQueryJpaTm;
 import google.registry.persistence.PersistenceModule.BeamJpaTm;
+import google.registry.persistence.PersistenceModule.BeamReadOnlyReplicaJpaTm;
 import google.registry.persistence.PersistenceModule.TransactionIsolationLevel;
 import google.registry.persistence.transaction.JpaTransactionManager;
 import google.registry.privileges.secretmanager.SecretManagerModule;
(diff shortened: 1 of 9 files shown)


mindhog (changes requested):
Sorry I totally missed this!

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1470#-:-MspDZNgD8iYBz3FA3r5:bk7cgx8)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 409 at r1](https://reviewable.io/reviews/google/nomulus/1470#-MspEPl78XDT6knspkH-:-MspEPl78XDT6knspkH0:baw0o0s) ([raw file](https://github.com/google/nomulus/blob/697b3c2a0da5e7111f39e213bba3549763f92e26/core/src/main/java/google/registry/config/RegistryConfig.java#L409)):*
> ```Java
>     public static Optional<String> provideCloudSqlReplicaInstanceConnectionName(
>         RegistryConfigSettings config) {
>       return Optional.ofNullable(config.cloudSql.replicaInstanceConnectionName);
> ```

I presume the intent of making this optional and falling back to the main database is to not introduce this into prod until some later point and then do so with a config change?

gbrodman (commented):
no worries!

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1470#-:-MspRwobBHen0-9WPASB:bbpjafw)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/config/RegistryConfig.java, line 409 at r1](https://reviewable.io/reviews/google/nomulus/1470#-MspEPl78XDT6knspkH-:-MspRySl7m1ojqXABOLi:bubzcgx) ([raw file](https://github.com/google/nomulus/blob/697b3c2a0da5e7111f39e213bba3549763f92e26/core/src/main/java/google/registry/config/RegistryConfig.java#L409)):*

I presume the intent of making this optional and falling back to the main database is to not introduce this into prod until some later point and then do so with a config change?
</blockquote></details>

Yeah, that's one reason. The other reason is in case we (or, more realistically, any other theoretical people that might at some point use this as open-source software) don't have a read-only-replica set up at some point.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1470)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1430, opened 2021-11-18
Replace all existing vkey string to vkey.stringify() 

rachelguan:
This purpose of this PR is to replace all existing vkey strings with new vkey string representation strings created by stringify(). We need to make sure that the vkey strings from task queue (which are stringified) can be used to re-create the vkey successfully. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1430)

PR #1475, opened 2021-12-22
Change resource key type from Key to String for ResaveEntityAction

rachelguan:
Currently, in ```ResaveEntityAction```, resource key is passed in as ```Key<ImmutableObject>``` from ```BatchModule``` . The problem with this is, if the resource key is in the new vkey string representation, this action will fail because a key cannot be created in ```BatchModule``` since the extracted string is not an acceptable websafe string used by ```Key```. 

Instead of creating a ```Key<ImmutableObject>``` object then creating a vkey by calling ```VKey.from(Key<ImmutableObject> object)```, we need to replace with```VKey.create() ``` which can handle both new vkey string representation and old vkey string representation (which is ofy key in string form).  

In this PR, ```throws Exception``` is dropped from the existing ```VKey.create()``` method since there's no unexpected exception other than parsing error, which is already handled. 

This PR is the solution to the bug (b/207363014). This PR should be merged into master before [PR#1430](https://github.com/google/nomulus/pull/1430).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1475)

### DIFF
--- a/core/src/main/java/google/registry/batch/BatchModule.java
+++ b/core/src/main/java/google/registry/batch/BatchModule.java
@@ -33,10 +33,8 @@ import static google.registry.request.RequestParameters.extractSetOfDatetimePara
 
 import com.google.appengine.api.taskqueue.Queue;
 import com.google.common.collect.ImmutableSet;
-import com.googlecode.objectify.Key;
 import dagger.Module;
 import dagger.Provides;
-import google.registry.model.ImmutableObject;
 import google.registry.request.Parameter;
 import java.util.Optional;
 import javax.inject.Named;
@@ -79,9 +77,8 @@ public class BatchModule {
 
   @Provides
   @Parameter(PARAM_RESOURCE_KEY)
-  // TODO(b/207363014): figure out if this needs to be modified for vkey string replacement
-  static Key<ImmutableObject> provideResourceKey(HttpServletRequest req) {
-    return Key.create(extractRequiredParameter(req, PARAM_RESOURCE_KEY));
+  static String provideResourceKey(HttpServletRequest req) {
+    return extractRequiredParameter(req, PARAM_RESOURCE_KEY);
   }
 
   @Provides
--- a/core/src/main/java/google/registry/batch/ResaveEntityAction.java
+++ b/core/src/main/java/google/registry/batch/ResaveEntityAction.java
@@ -22,7 +22,6 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.flogger.FluentLogger;
-import com.googlecode.objectify.Key;
 import google.registry.model.EppResource;
 import google.registry.model.ImmutableObject;
 import google.registry.persistence.VKey;
@@ -50,7 +49,7 @@ public class ResaveEntityAction implements Runnable {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
-  private final Key<ImmutableObject> resourceKey;
+  private final String resourceKey;
   private final DateTime requestedTime;
   private final ImmutableSortedSet<DateTime> resaveTimes;
   private final AsyncTaskEnqueuer asyncTaskEnqueuer;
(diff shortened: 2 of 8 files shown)


mindhog (changes requested):
> Reviewed 7 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1475)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @mindhog and @rachelguan)

___
*[core/src/main/java/google/registry/batch/BatchModule.java, line 80 at r1](https://reviewable.io/reviews/google/nomulus/1475#-MrYPx3c6sKXaLOlgBLg:-MrYPx3c6sKXaLOlgBLh:b-rho7b4) ([raw file](https://github.com/google/nomulus/blob/2a7d28dd583ee0b3a060a64720d89fad145bd253/core/src/main/java/google/registry/batch/BatchModule.java#L80)):*
> ```Java
>   @Provides
>   @Parameter(PARAM_RESOURCE_KEY)
>   static String provideResourceKey(HttpServletRequest req) {
> ```

Why not make this return a VKey?  In addition to providing typesafety it would allow you to remove the conversions in the callsites.
___
*[core/src/test/java/google/registry/persistence/VKeyTest.java, line 162 at r1](https://reviewable.io/reviews/google/nomulus/1475#-MrYQZEU-QFFEJDBYg1A:-MrYQZEU-QFFEJDBYg1B:b9jw8xt) ([raw file](https://github.com/google/nomulus/blob/156344e40813a46209b7b50828ff3e4ede538f52/core/src/test/java/google/registry/persistence/VKeyTest.java#L162)):*
> ```Java
>   /** Test stringify() with vkey created via different ways. */
>   @Test
>   void testStringify_sqlOnlyVKey() throws Exception {
> ```

Out of curiosity, is intellij or something warning you about these?

mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1475)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @mindhog and @rachelguan)

___
*[core/src/main/java/google/registry/module/backend/BackendComponent.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/1475#-MrYS44W7FdZAsPoikY9:-MrYS44X-Gjzy5mWiW2O:bvxdrik) ([raw file](https://github.com/google/nomulus/blob/2a7d28dd583ee0b3a060a64720d89fad145bd253/core/src/main/java/google/registry/module/backend/BackendComponent.java#L59)):*
> ```Java
>       AuthModule.class,
>       BackendRequestComponentModule.class,
>       BatchModule.class,
> ```

I'm curious why we didn't need this before?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1475)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/batch/BatchModule.java, line 80 at r1](https://reviewable.io/reviews/google/nomulus/1475#-MrYPx3c6sKXaLOlgBLg:-MrYUkTT0xnnPQ68M3Zv:b6z0rq4) ([raw file](https://github.com/google/nomulus/blob/2a7d28dd583ee0b3a060a64720d89fad145bd253/core/src/main/java/google/registry/batch/BatchModule.java#L80)):*

Why not make this return a VKey?  In addition to providing typesafety it would allow you to remove the conversions in the callsites.
</blockquote></details>

I was thinking to stay consistent since ```VKey.create(string)``` are used in ```RefreshDnsOnHostRenameAction``` and ```DeleteContactsAndHostsAction``` .  
___
*[core/src/main/java/google/registry/module/backend/BackendComponent.java, line 59 at r1](https://reviewable.io/reviews/google/nomulus/1475#-MrYS44W7FdZAsPoikY9:-MrYTjhZAQZg2zwd4KDT:b-4c3rdu) ([raw file](https://github.com/google/nomulus/blob/2a7d28dd583ee0b3a060a64720d89fad145bd253/core/src/main/java/google/registry/module/backend/BackendComponent.java#L59)):*

I'm curious why we didn't need this before?
</blockquote></details>

no idea .. the binding failed after I changed the return type. 
___
*[core/src/test/java/google/registry/persistence/VKeyTest.java, line 162 at r1](https://reviewable.io/reviews/google/nomulus/1475#-MrYQZEU-QFFEJDBYg1A:-MrYTqRYEZ9zOGOKpD5f:bvo0ov4) ([raw file](https://github.com/google/nomulus/blob/156344e40813a46209b7b50828ff3e4ede538f52/core/src/test/java/google/registry/persistence/VKeyTest.java#L162)):*

Out of curiosity, is intellij or something warning you about these?
</blockquote></details>

@.@ there should be a difference since I removed a lot of ```throws exception```. I did that because I removed ```throws Exception``` of the ```create()``` method in ```VKey.java``` since there is only parsing error which I already took care of. Mostly just different ```VKey.create()``` calls.

mindhog (approved):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1475)* status: all files reviewed, 3 unresolved discussions (waiting on @mindhog)

PR #1478, opened 2021-12-28
Add pending action extension to server update poll messages

CydeWeys:
This is necessary for the poll messages to contain the necessary context
explaining what domain name the relevant statuses were being added/removed
to/from.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1478)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
@@ -43,6 +43,7 @@ import static google.registry.flows.domain.DomainFlowUtils.verifyNotInPendingDel
 import static google.registry.model.reporting.HistoryEntry.Type.DOMAIN_UPDATE;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Ordering;
@@ -77,9 +78,11 @@ import google.registry.model.domain.secdns.SecDnsUpdateExtension;
 import google.registry.model.domain.superuser.DomainUpdateSuperuserExtension;
 import google.registry.model.eppcommon.AuthInfo;
 import google.registry.model.eppcommon.StatusValue;
+import google.registry.model.eppcommon.Trid;
 import google.registry.model.eppinput.EppInput;
 import google.registry.model.eppinput.ResourceCommand;
 import google.registry.model.eppoutput.EppResponse;
+import google.registry.model.poll.PendingActionNotificationResponse.DomainPendingActionNotificationResponse;
 import google.registry.model.poll.PollMessage;
 import google.registry.model.reporting.IcannReportingTypes.ActivityReportField;
 import google.registry.model.tld.Registry;
@@ -149,6 +152,7 @@ public final class DomainUpdateFlow implements TransactionalFlow {
   @Inject @RegistrarId String registrarId;
   @Inject @TargetId String targetId;
   @Inject @Superuser boolean isSuperuser;
+  @Inject Trid trid;
   @Inject DomainHistory.Builder historyBuilder;
   @Inject DnsQueue dnsQueue;
   @Inject EppResponse.Builder responseBuilder;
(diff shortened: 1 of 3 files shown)


CydeWeys:
See http://b/211978679

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1478)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java, line 956 at r2](https://reviewable.io/reviews/google/nomulus/1478#-Ms1Vuef3qTHUTAlIp9d:-Ms1Vuef3qTHUTAlIp9e:b-4mmu00) ([raw file](https://github.com/google/nomulus/blob/d6cb71d19364db3f211c1807c4595273e222d179/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L956)):*
> ```Java
>             .setResponseData(
>                 ImmutableList.of(
>                     DomainPendingActionNotificationResponse.create(
> ```

Technically according to the documentation, this is meant to signify the completion of a pending action on a domain. This isn't really a pending action though right, it's just an update? I'm not sure if that matters.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1478)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java, line 956 at r2](https://reviewable.io/reviews/google/nomulus/1478#-Ms1Vuef3qTHUTAlIp9d:-Ms1ZVf11Cwn3qGlrHMo:b-f57fz7) ([raw file](https://github.com/google/nomulus/blob/d6cb71d19364db3f211c1807c4595273e222d179/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L956)):*

Technically according to the documentation, this is meant to signify the completion of a pending action on a domain. This isn't really a pending action though right, it's just an update? I'm not sure if that matters. 
</blockquote></details>

Yeah, we already crossed that Rubicon in #1400, in which we use the pending extension to send the domain name on an immediate superuser deletion of a domain.

Basically, there were two options: Either use this extension in a way that technically might not be 100% in the spirit of it, but that registrars definitely already support, or (a) create a new extension that registrars would have to add support for, or (b) don't use an extension at all and send the domain name as unstructured data in the poll message.  That latter isn't good because it would then cause registrars to also have to create special case handling for us to parse the String.

So basically doing it this way is the least work for everyone -- it's an extension that registrars already support, and it's the only good way to tell them in a structured format what domain the poll message corresponds to.

gbrodman (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1478)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

___
*[core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java, line 956 at r2](https://reviewable.io/reviews/google/nomulus/1478#-Ms1Vuef3qTHUTAlIp9d:-Ms1kDetCyKAnpRfH6Ow:b1sbh1a) ([raw file](https://github.com/google/nomulus/blob/d6cb71d19364db3f211c1807c4595273e222d179/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L956)):*

Yeah, we already crossed that Rubicon in #1400, in which we use the pending extension to send the domain name on an immediate superuser deletion of a domain.

Basically, there were two options: Either use this extension in a way that technically might not be 100% in the spirit of it, but that registrars definitely already support, or (a) create a new extension that registrars would have to add support for, or (b) don't use an extension at all and send the domain name as unstructured data in the poll message.  That latter isn't good because it would then cause registrars to also have to create special case handling for us to parse the String.

So basically doing it this way is the least work for everyone -- it's an extension that registrars already support, and it's the only good way to tell them in a structured format what domain the poll message corresponds to.

</blockquote></details>

A bit gross, but fair enough.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1478)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

___
*[core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java, line 956 at r2](https://reviewable.io/reviews/google/nomulus/1478#-Ms1Vuef3qTHUTAlIp9d:-Ms1mAgy-Cnjj3T_NFo_:bhrnan3) ([raw file](https://github.com/google/nomulus/blob/d6cb71d19364db3f211c1807c4595273e222d179/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L956)):*

A bit gross, but fair enough. 
</blockquote></details>

Eh, it's not really _that_ gross.  Every single field in the extension is correct and useful for this purpose.  The only thing slightly wrong with it is that it's called a pending action response, but the action that was taken was only "pending" inside of a single atomic transaction.

PR #1479, opened 2021-12-29
Don't throw errors when existing premium list is empty

CydeWeys:
This state is possible to get into when things go wrong and it shouldn't prevent
saving new revisions of the list. Note that it will continue to throw errors if
you attempt to save a new revision that is blank (which is usually a mistake).

See http://b/211774375

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1479)

### DIFF
--- a/core/src/main/java/google/registry/model/tld/label/PremiumListDao.java
+++ b/core/src/main/java/google/registry/model/tld/label/PremiumListDao.java
@@ -14,6 +14,7 @@
 
 package google.registry.model.tld.label;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.config.RegistryConfig.getDomainLabelListCacheDuration;
 import static google.registry.config.RegistryConfig.getSingletonCachePersistDuration;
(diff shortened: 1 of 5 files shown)


CydeWeys (commented):
+reviewer:@gbrodman 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1479#-:-MsBRJYuBxwNbmw2G1xx:b3une8x)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @gbrodman and @sarahcaseybot)

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1479)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[core/src/main/java/google/registry/model/tld/label/PremiumListUtils.java, line 41 at r1](https://reviewable.io/reviews/google/nomulus/1479#-MsBVWqo2ZoyxeAcwgkB:-MsBVWqo2ZoyxeAcwgkC:b6s8v1l) ([raw file](https://github.com/google/nomulus/blob/9d3cbd07fdf12f753d425d6fb342fa5ef6c82341/core/src/main/java/google/registry/model/tld/label/PremiumListUtils.java#L41)):*
> ```Java
>             .build();
>     ImmutableMap<String, PremiumEntry> prices = partialPremiumList.parse(inputData);
>     checkArgument(inputData.size() > 0, "Input cannot be empty");
> ```

Looking at this with fresh eyes it does seem a bit weird that we have this one separate utils class with this one separate parsing method that doesn't enforce the restriction that lists be nonempty. 

I think this will be made moot by my PR where i remove the "parse existing premium list" code from the UpdatePremiumListCommand, and it's not technically wrong to remove this here (since it's enforced in `save`, but maybe we should consider removing this class entirely and moving the code into `PremiumList`?

PR #1481, opened 2021-12-29
Remove static methods in back up actions

rachelguan:
The main goal of this PR was to move the static methods for enqueueing tasks to where the method was called. Since a `CloudTasksUtil` instance needs to be injected at where the task is enqueued, those static methods were replaced with `cloudTasksUtil.enqueue(queueName, task)`. 

These static methods were created in those backup action files. In order to completely remove the dependency of `TaskOptions` and related libraries from those action files, `BigqueryPollJobAction` and related files need to be modified as well. There is a static class BigqueryPollJobEnqueuer` in `BigqueryPollJobAction` that contains two methods of enqueueing tasks. Those two methods were called in multiple files. The easiest way to modify this class was to have the methods return `Task` and they can be enqueued inside the file where the method is called. 

Some tasks require adding headers and payload for `POST` tasks. The purpose of this PR was to replace `TaskQueueUtils` with `CloudTasksUtil`. It's debatable to put those key-values into `params`. For now, `CloudTasksUtil` does not have support for those but it can be done after; payload and headers can be added on top of the tasks created via the helpers in `CloudTasksUtil`. 

Note: the changes made in `CloudTasksUtil` are temporary. A separate [PR](https://github.com/google/nomulus/pull/1448) was made to support null/empty `params` 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1481)

### DIFF
--- a/core/src/main/java/google/registry/export/BackupDatastoreAction.java
+++ b/core/src/main/java/google/registry/export/BackupDatastoreAction.java
@@ -14,9 +14,10 @@
 
 package google.registry.export;
 
-import static google.registry.export.CheckBackupAction.enqueuePollTask;
 import static google.registry.request.Action.Method.POST;
 
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig;
 import google.registry.export.datastore.DatastoreAdmin;
@@ -26,6 +27,9 @@ import google.registry.request.Action;
 import google.registry.request.HttpException.InternalServerErrorException;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
+import google.registry.util.Clock;
+import google.registry.util.CloudTasksUtils;
+import java.util.Optional;
 import javax.inject.Inject;
 
 /**
@@ -59,6 +63,8 @@ public class BackupDatastoreAction implements Runnable {
 
   @Inject DatastoreAdmin datastoreAdmin;
   @Inject Response response;
+  @Inject Clock clock;
+  @Inject CloudTasksUtils cloudTasksUtils;
 
   @Inject
   BackupDatastoreAction() {}
@@ -73,8 +79,22 @@ public class BackupDatastoreAction implements Runnable {
               .execute();
 
       String backupName = backup.getName();
-      // Enqueue a poll task to monitor the backup and load REPORTING-related kinds into bigquery.
-      enqueuePollTask(backupName, AnnotatedEntities.getReportingKinds());
+
+      // Enqueue a poll task to monitor the backup for completion and load REPORTING-related kinds
+      // into bigquery.
+      cloudTasksUtils.enqueue(
+          CheckBackupAction.QUEUE,
+          CloudTasksUtils.createPostTask(
+              CheckBackupAction.PATH,
+              CheckBackupAction.SERVICE,
+              ImmutableMultimap.of(
+                  CheckBackupAction.CHECK_BACKUP_NAME_PARAM,
+                  backupName,
+                  CheckBackupAction.CHECK_BACKUP_KINDS_TO_LOAD_PARAM,
+                  Joiner.on(',').join(AnnotatedEntities.getReportingKinds())),
+              clock,
+              Optional.of((int) CheckBackupAction.POLL_COUNTDOWN.getMillis())));
+
       String message =
           String.format(
               "Datastore backup started with name: %s\nSaving to %s",
--- a/core/src/main/java/google/registry/export/BackupDatastoreAction.java
+++ b/core/src/main/java/google/registry/export/BackupDatastoreAction.java
@@ -24,6 +24,7 @@ import google.registry.export.datastore.DatastoreAdmin;
 import google.registry.export.datastore.Operation;
 import google.registry.model.annotations.DeleteAfterMigration;
 import google.registry.request.Action;
+import google.registry.request.Action.Service;
 import google.registry.request.HttpException.InternalServerErrorException;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
@@ -78,20 +79,18 @@ public class BackupDatastoreAction implements Runnable {
               .execute();
 
       String backupName = backup.getName();
-
-      // Enqueue a poll task to monitor the backup for completion and load REPORTING-related kinds
+      // Enqueue a poll task to monitor the backup for completion and load reporting-related kinds
       // into bigquery.
       cloudTasksUtils.enqueue(
           CheckBackupAction.QUEUE,
-          CloudTasksUtils.createPostTask(
+          cloudTasksUtils.createPostTaskWithDelay(
               CheckBackupAction.PATH,
-              CheckBackupAction.SERVICE,
+              Service.BACKEND.toString(),
               ImmutableMultimap.of(
                   CheckBackupAction.CHECK_BACKUP_NAME_PARAM,
                   backupName,
                   CheckBackupAction.CHECK_BACKUP_KINDS_TO_LOAD_PARAM,
                   Joiner.on(',').join(AnnotatedEntities.getReportingKinds())),
-              clock,
               CheckBackupAction.POLL_COUNTDOWN));
       String message =
           String.format(
--- a/core/src/main/java/google/registry/export/BackupDatastoreAction.java
+++ b/core/src/main/java/google/registry/export/BackupDatastoreAction.java
@@ -16,11 +16,9 @@ package google.registry.export;
 
 import static google.registry.request.Action.Method.POST;
 
-import com.google.cloud.tasks.v2.Task;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
-import com.google.protobuf.Timestamp;
 import google.registry.config.RegistryConfig;
 import google.registry.export.datastore.DatastoreAdmin;
 import google.registry.export.datastore.Operation;
@@ -31,7 +29,6 @@ import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
-import java.time.Instant;
 import javax.inject.Inject;
 
 /**
@@ -84,31 +81,18 @@ public class BackupDatastoreAction implements Runnable {
 
       // Enqueue a poll task to monitor the backup for completion and load REPORTING-related kinds
       // into bigquery.
-      // TODO(rachelguan): replace it with schedule time helper once that PR gets merged
-      Instant scheduleTime =
-          Instant.ofEpochMilli(
-              clock
-                  .nowUtc()
-                  .plusMillis((int) CheckBackupAction.POLL_COUNTDOWN.getMillis())
-                  .getMillis());
       cloudTasksUtils.enqueue(
           CheckBackupAction.QUEUE,
-          Task.newBuilder(
-                  CloudTasksUtils.createPostTask(
-                      CheckBackupAction.PATH,
-                      CheckBackupAction.SERVICE,
-                      ImmutableMultimap.of(
-                          CheckBackupAction.CHECK_BACKUP_NAME_PARAM,
-                          backupName,
-                          CheckBackupAction.CHECK_BACKUP_KINDS_TO_LOAD_PARAM,
-                          Joiner.on(',').join(AnnotatedEntities.getReportingKinds()))))
-              .setScheduleTime(
-                  Timestamp.newBuilder()
-                      .setSeconds(scheduleTime.getEpochSecond())
-                      .setNanos(scheduleTime.getNano())
-                      .build())
-              .build());
-
+          CloudTasksUtils.createPostTask(
+              CheckBackupAction.PATH,
+              CheckBackupAction.SERVICE,
+              ImmutableMultimap.of(
+                  CheckBackupAction.CHECK_BACKUP_NAME_PARAM,
+                  backupName,
+                  CheckBackupAction.CHECK_BACKUP_KINDS_TO_LOAD_PARAM,
+                  Joiner.on(',').join(AnnotatedEntities.getReportingKinds())),
+              clock,
+              CheckBackupAction.POLL_COUNTDOWN));
       String message =
           String.format(
               "Datastore backup started with name: %s\nSaving to %s",
--- a/core/src/main/java/google/registry/export/BackupDatastoreAction.java
+++ b/core/src/main/java/google/registry/export/BackupDatastoreAction.java
@@ -16,9 +16,11 @@ package google.registry.export;
 
 import static google.registry.request.Action.Method.POST;
 
+import com.google.cloud.tasks.v2.Task;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
+import com.google.protobuf.Timestamp;
 import google.registry.config.RegistryConfig;
 import google.registry.export.datastore.DatastoreAdmin;
 import google.registry.export.datastore.Operation;
@@ -29,7 +31,7 @@ import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
-import java.util.Optional;
+import java.time.Instant;
 import javax.inject.Inject;
 
 /**
@@ -82,18 +84,30 @@ public class BackupDatastoreAction implements Runnable {
 
       // Enqueue a poll task to monitor the backup for completion and load REPORTING-related kinds
       // into bigquery.
+      // TODO(rachelguan): replace it with schedule time helper once that PR gets merged
+      Instant scheduleTime =
+          Instant.ofEpochMilli(
+              clock
+                  .nowUtc()
+                  .plusMillis((int) CheckBackupAction.POLL_COUNTDOWN.getMillis())
+                  .getMillis());
       cloudTasksUtils.enqueue(
           CheckBackupAction.QUEUE,
-          CloudTasksUtils.createPostTask(
-              CheckBackupAction.PATH,
-              CheckBackupAction.SERVICE,
-              ImmutableMultimap.of(
-                  CheckBackupAction.CHECK_BACKUP_NAME_PARAM,
-                  backupName,
-                  CheckBackupAction.CHECK_BACKUP_KINDS_TO_LOAD_PARAM,
-                  Joiner.on(',').join(AnnotatedEntities.getReportingKinds())),
-              clock,
-              Optional.of((int) CheckBackupAction.POLL_COUNTDOWN.getMillis())));
+          Task.newBuilder(
+                  CloudTasksUtils.createPostTask(
+                      CheckBackupAction.PATH,
+                      CheckBackupAction.SERVICE,
+                      ImmutableMultimap.of(
+                          CheckBackupAction.CHECK_BACKUP_NAME_PARAM,
+                          backupName,
+                          CheckBackupAction.CHECK_BACKUP_KINDS_TO_LOAD_PARAM,
+                          Joiner.on(',').join(AnnotatedEntities.getReportingKinds()))))
+              .setScheduleTime(
+                  Timestamp.newBuilder()
+                      .setSeconds(scheduleTime.getEpochSecond())
+                      .setNanos(scheduleTime.getNano())
+                      .build())
+              .build());
 
       String message =
           String.format(
--- a/core/src/main/java/google/registry/export/BigqueryPollJobAction.java
+++ b/core/src/main/java/google/registry/export/BigqueryPollJobAction.java
@@ -20,25 +20,18 @@ import com.google.api.services.bigquery.Bigquery;
 import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.JobReference;
 import com.google.cloud.tasks.v2.Task;
-import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
-import com.google.protobuf.ByteString;
 import dagger.Lazy;
 import google.registry.request.Action;
-import google.registry.request.Action.Service;
 import google.registry.request.Header;
 import google.registry.request.HttpException.BadRequestException;
 import google.registry.request.HttpException.NotModifiedException;
 import google.registry.request.Payload;
 import google.registry.request.auth.Auth;
-import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-import java.util.Optional;
 import javax.inject.Inject;
 import org.joda.time.Duration;
 
@@ -72,8 +65,6 @@ public class BigqueryPollJobAction implements Runnable {
   @Inject @Header(CHAINED_TASK_QUEUE_HEADER) Lazy<String> chainedQueueName;
   @Inject @Header(PROJECT_ID_HEADER) String projectId;
   @Inject @Header(JOB_ID_HEADER) String jobId;
-
-  @Inject Clock clock;
 
   @Inject @Payload byte[] payload;
 
@@ -131,50 +122,5 @@ public class BigqueryPollJobAction implements Runnable {
     }
     logger.atInfo().log("Bigquery job succeeded - %s.", jobRefString);
     return true;
-  }
-
-  /** Helper class to enqueue a bigquery poll job. */
-  public static class BigqueryPollJob {
-    /** Enqueue a task to poll for the success or failure of the referenced BigQuery job. */
-    public static Task createGetTask(JobReference jobRef, Clock clock) {
-      return Task.newBuilder()
-          .setAppEngineHttpRequest(
-              CloudTasksUtils.createGetTask(
-                      PATH,
-                      Service.BACKEND.toString(),
-                      ImmutableMultimap.of(),
-                      clock,
-                      Optional.of((int) POLL_COUNTDOWN.getMillis()))
-                  .getAppEngineHttpRequest()
-                  .toBuilder()
-                  .putHeaders(PROJECT_ID_HEADER, jobRef.getProjectId())
-                  .putHeaders(JOB_ID_HEADER, jobRef.getJobId())
-                  .build())
-          .build();
-    }
-
-    /**
-     * Enqueue a task to poll for the success or failure of the referenced BigQuery job and to
-     * launch the provided task in the specified queue if the job succeeds.
-     */
-    public static Task createPostTask(
-        JobReference jobRef, Task chainedTask, String chainedTaskQueue) throws IOException {
-      // Serialize the chainedTask into a byte array to put in the task payload.
-      ByteArrayOutputStream taskBytes = new ByteArrayOutputStream();
-      new ObjectOutputStream(taskBytes).writeObject(chainedTask);
-      return Task.newBuilder()
-          .setAppEngineHttpRequest(
-              CloudTasksUtils.createPostTask(
-                      PATH, Service.BACKEND.toString(), ImmutableMultimap.of())
-                  .getAppEngineHttpRequest()
-                  .toBuilder()
-                  .putHeaders(PROJECT_ID_HEADER, jobRef.getProjectId())
-                  .putHeaders(JOB_ID_HEADER, jobRef.getJobId())
-                  .putHeaders(CHAINED_TASK_QUEUE_HEADER, chainedTaskQueue)
-                  .setBody(ByteString.copyFrom(taskBytes.toByteArray()))
-                  .build())
-          .build();
-    }
-
   }
 }
--- a/core/src/main/java/google/registry/export/BigqueryPollJobAction.java
+++ b/core/src/main/java/google/registry/export/BigqueryPollJobAction.java
@@ -14,31 +14,31 @@
 
 package google.registry.export;
 
-import static com.google.appengine.api.taskqueue.QueueFactory.getQueue;
-import static com.google.appengine.api.taskqueue.TaskOptions.Builder.withUrl;
 import static google.registry.bigquery.BigqueryUtils.toJobReferenceString;
 
 import com.google.api.services.bigquery.Bigquery;
 import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.JobReference;
-import com.google.appengine.api.taskqueue.Queue;
-import com.google.appengine.api.taskqueue.TaskHandle;
-import com.google.appengine.api.taskqueue.TaskOptions;
-import com.google.appengine.api.taskqueue.TaskOptions.Method;
+import com.google.cloud.tasks.v2.Task;
+import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
+import com.google.protobuf.ByteString;
 import dagger.Lazy;
 import google.registry.request.Action;
+import google.registry.request.Action.Service;
 import google.registry.request.Header;
 import google.registry.request.HttpException.BadRequestException;
 import google.registry.request.HttpException.NotModifiedException;
 import google.registry.request.Payload;
 import google.registry.request.auth.Auth;
-import google.registry.util.TaskQueueUtils;
+import google.registry.util.Clock;
+import google.registry.util.CloudTasksUtils;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.util.Optional;
 import javax.inject.Inject;
 import org.joda.time.Duration;
 
@@ -67,11 +67,16 @@ public class BigqueryPollJobAction implements Runnable {
   static final Duration POLL_COUNTDOWN = Duration.standardSeconds(20);
 
   @Inject Bigquery bigquery;
-  @Inject TaskQueueUtils taskQueueUtils;
+  @Inject CloudTasksUtils cloudTasksUtils;
+
   @Inject @Header(CHAINED_TASK_QUEUE_HEADER) Lazy<String> chainedQueueName;
   @Inject @Header(PROJECT_ID_HEADER) String projectId;
   @Inject @Header(JOB_ID_HEADER) String jobId;
+
+  @Inject Clock clock;
+
   @Inject @Payload byte[] payload;
+
   @Inject BigqueryPollJobAction() {}
 
   @Override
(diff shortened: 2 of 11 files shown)


jianglai (commented):
> that contains two methods of enqueueing tasks. Those two methods were called in multiple files. The easiest way to modify this class was to have the methods return
I believe you have an extra backquote before "that contains two methods..." that messed up the rendering.

Also the PR to deal with null/empty params is merged. Need to rebase this PR.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481#-:-MtLnmWH0oUM4Dwds3BB:b-cobyj)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @jianglai)

rachelguan (commented):
fixed that. PTAL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481#-:-Mur92ya5DZYFiHSqJYx:b-xyjj8r)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @jianglai)

rachelguan (commented):
This PR has delayed schedule time as well. it would be great if you could take a look at everything else. Once the schedule time PR goes in, I would modify the part that has to do with schedule time. thanks!

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481#-:-Mur975s-htPsbjTZJ6F:b-7qyhls)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (commented):
IIRC, none of the  task enqueue calls in these Actions happen inside transactions. Please verify this is in need the case.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481#-:-MvKCYLK119jXtNfhBWN:b-mb653p)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @jianglai)

rachelguan (commented):
I agree that none of them is actually transactional 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481#-:-MvKtOqA3RT3QJPsRNxU:brso5ds)* status: 0 of 11 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @jianglai and @rachelguan)

___
*[core/src/main/java/google/registry/export/BackupDatastoreAction.java, line 87 at r3](https://reviewable.io/reviews/google/nomulus/1481#-MvMcZ5VANKoQymVz4pe:-MvMcZ5VANKoQymVz4pf:b-y22vxh) ([raw file](https://github.com/google/nomulus/blob/530c12f76a120c83b006b196687f2ebdd8a8509c/core/src/main/java/google/registry/export/BackupDatastoreAction.java#L87)):*
> ```Java
>       // Enqueue a poll task to monitor the backup for completion and load REPORTING-related kinds
>       // into bigquery.
>       // TODO(rachelguan): replace it with schedule time helper once that PR gets merged
> ```

You can do this now.



_Code quote:_
```Java
 // TODO(rachelguan): replace it with schedule time helper once that PR gets merged
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/main/java/google/registry/export/BackupDatastoreAction.java, line 87 at r3](https://reviewable.io/reviews/google/nomulus/1481#-MvMcZ5VANKoQymVz4pe:-MvQ8HdC4pZesLtrf7cO:b3tsjhm) ([raw file](https://github.com/google/nomulus/blob/530c12f76a120c83b006b196687f2ebdd8a8509c/core/src/main/java/google/registry/export/BackupDatastoreAction.java#L87)):*

You can do this now.
</blockquote></details>

removed. that TODO. One of the tasks in this PR was not created via ```CloudTasksUtil``` because of the headers. I want to avoid changing the structure of the task since these backup actions are gonna be deleted after migration.

jianglai (changes requested):
This PR needs to be deployed to alpha and tested.

> Reviewed 5 of 11 files at r1, 1 of 1 files at r2, 1 of 3 files at r3, 3 of 4 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481#-:-MtLniKTC1I82ScmWlk3:bwj9qvf)* status: all files reviewed, 7 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/export/BackupDatastoreAction.java, line 87 at r3](https://reviewable.io/reviews/google/nomulus/1481#-MvMcZ5VANKoQymVz4pe:-MvZIvlW1Mjxx_1heNrD:b-8gjwhn) ([raw file](https://github.com/google/nomulus/blob/530c12f76a120c83b006b196687f2ebdd8a8509c/core/src/main/java/google/registry/export/BackupDatastoreAction.java#L87)):*

removed. that TODO. One of the tasks in this PR was not created via ```CloudTasksUtil``` because of the headers. I want to avoid changing the structure of the task since these backup actions are gonna be deleted after migration. 
</blockquote></details>

Well, as it stands now you have to use headers because the chained tasks are serialized and stored as payload, with preclude the use of other params. But I think you can just make it another param, like `payload=[serialized url safe blob]` and you can convert the headers to params as well.

But it's your call. I think it probably is neater if we convert these headers to params, but as you said it's throwaway code anyway.
___
*[core/src/main/java/google/registry/export/BackupDatastoreAction.java, line 82 at r4](https://reviewable.io/reviews/google/nomulus/1481#-MvRVK-REeddaovWp8Ue:-MvRVK-REeddaovWp8Uf:b-x4ypgb) ([raw file](https://github.com/google/nomulus/blob/d318ddc6cdb80584456d1f6a21df8ec11a55b181/core/src/main/java/google/registry/export/BackupDatastoreAction.java#L82)):*
> ```Java
>       String backupName = backup.getName();
> 
>       // Enqueue a poll task to monitor the backup for completion and load REPORTING-related kinds
> ```

No idea why this was capitalized. I don't think they need to be.



_Code quote:_
```Java
REPORTING
```
___
*[core/src/main/java/google/registry/export/CheckBackupAction.java, line 62 at r4](https://reviewable.io/reviews/google/nomulus/1481#-MvRZ4UU6NuP0bum1ei7:-MvRZ4UU6NuP0bum1ei8:b-c59hrr) ([raw file](https://github.com/google/nomulus/blob/d318ddc6cdb80584456d1f6a21df8ec11a55b181/core/src/main/java/google/registry/export/CheckBackupAction.java#L62)):*
> ```Java
>     automaticallyPrintOk = true,
>     auth = Auth.AUTH_INTERNAL_OR_ADMIN)
> ```

extra line
___
*[core/src/main/java/google/registry/export/CheckBackupAction.java, line 74 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZH6d3ARD7OJyMilka:-MvZH6d3ARD7OJyMilkb:b-hqgqfs) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/CheckBackupAction.java#L74)):*
> ```Java
>   static final String QUEUE = "export-snapshot-poll"; // See queue.xml.
> 
>   static final String SERVICE = Service.BACKEND.toString();
> ```

Not necessary, you can just call `Service.BACKEND` directly when you enqueue. Ideally we'd read this value directly from the class annotation, but that's out of scope for this PR.



_Code quote:_
```Java
static final String SERVICE = Service.BACKEND.toString();
```
___
*[core/src/main/java/google/registry/export/UpdateSnapshotViewAction.java, line 60 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZL8mdBTSTEZi6oM-R:-MvZL8mdBTSTEZi6oM-S:bs2dogg) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UpdateSnapshotViewAction.java#L60)):*
> ```Java
>   static final String PATH = "/_dr/task/updateSnapshotView"; // See web.xml.
> 
>   static final String SERVICE = Service.BACKEND.toString();
> ```

Delete



_Code quote:_
```Java
static final String SERVICE = Service.BACKEND.toString();
```
___
*[core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java, line 80 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZHYPW6xd1x2EbsDOB:-MvZHYPW6xd1x2EbsDOC:bg5u30) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java#L80)):*
> ```Java
>   static final String PATH = "/_dr/task/uploadDatastoreBackup"; // See web.xml.
> 
>   static final String SERVICE = Service.BACKEND.toString();
> ```

Same here. Delete.



_Code quote:_
```Java
static final String SERVICE = Service.BACKEND.toString();
```
___
*[core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java, line 165 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZPJx5Eh1MmetmuNlx:-MvZPJx5Eh1MmetmuNly:bmln1vd) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java#L165)):*
> ```Java
>               clock
>                   .nowUtc()
>                   .plusMillis((int) BigqueryPollJobAction.POLL_COUNTDOWN.getMillis())
> ```

Use `DateTime.plus`.



_Code quote:_
```Java
plusMillis
```
___
*[core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1481#-Mv_WxhI827pGFOKH03F:-Mv_WxhI827pGFOKH03G:b-zh0sax) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java#L90)):*
> ```Java
> 
>   @Test
>   void testSuccess_enqueuePollTask_withChainedTask() throws Exception {
> ```

I failed to see what this test does, it is not using any of the code in the action. If you want to test that the chained task gets enquequed correctly, it needs to run as part of the action.

Put it in another word, you basically duplicated the serialization code here, and if you were to change the serialization code in the action in any way, the change will not affect the test at all. The test is completely self-contained and has no dependency on the production code that it is supposed to test.

Also, doesn't `testSuccess_chainedPayloadAndJobSucceeded_enqueuesChainedTask` test exactly what you want to verify?



_Code quote:_
```Java
testSuccess_enqueuePollTask_withChainedTask
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 4 of 11 files reviewed, 7 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/export/BackupDatastoreAction.java, line 82 at r4](https://reviewable.io/reviews/google/nomulus/1481#-MvRVK-REeddaovWp8Ue:-Mv_ca4_6Ad9t6MeoVdv:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d318ddc6cdb80584456d1f6a21df8ec11a55b181/core/src/main/java/google/registry/export/BackupDatastoreAction.java#L82)):*

No idea why this was capitalized. I don't think they need to be.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/CheckBackupAction.java, line 62 at r4](https://reviewable.io/reviews/google/nomulus/1481#-MvRZ4UU6NuP0bum1ei7:-Mv_clkC1aMHfSKgum2L:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d318ddc6cdb80584456d1f6a21df8ec11a55b181/core/src/main/java/google/registry/export/CheckBackupAction.java#L62)):*

extra line
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/CheckBackupAction.java, line 74 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZH6d3ARD7OJyMilka:-Mv_dDiGEzsRRxUDVqGp:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/CheckBackupAction.java#L74)):*

Not necessary, you can just call `Service.BACKEND` directly when you enqueue. Ideally we'd read this value directly from the class annotation, but that's out of scope for this PR.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/UpdateSnapshotViewAction.java, line 60 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZL8mdBTSTEZi6oM-R:-Mv_dF4Z8V68p_oO76kV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UpdateSnapshotViewAction.java#L60)):*

Delete
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java, line 80 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZHYPW6xd1x2EbsDOB:-Mv_dJngEtgC7tUKczqt:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java#L80)):*

Same here. Delete.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java, line 165 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZPJx5Eh1MmetmuNlx:-Mv_dUwIDiS2bCfWlFWi:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java#L165)):*

Use `DateTime.plus`.
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1481#-Mv_WxhI827pGFOKH03F:-Mv_phh6BereqjKvAWb3:blt36m6) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java#L90)):*
>  the code in the action. If you want to test that the chained task gets enquequed correctly, it needs to run as part of the action.
The attempt was to rewrite the same test case. The original intention was to test the enqueuing behavior and see if the chained task could be read and written properly. Originally, the enqueueing was a method call instead. Thank you for spotting that..I now trigger enqueuing by calling ```action.run()``` to check if the original task is enqueued and chain task could be read.

jianglai (changes requested):
> Reviewed 3 of 5 files at r6, 6 of 7 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java, line 165 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZPJx5Eh1MmetmuNlx:-Mvegrf00ZeJM9jj64xB:b-gujmji) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java#L165)):*

Use `DateTime.plus`.
</blockquote></details>

Can use https://github.com/google/nomulus/pull/1521 after it is merged.
___
*[core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1481#-Mv_WxhI827pGFOKH03F:-MwDdTW893a1UeWD-rZj:b-1o4tec) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java#L90)):*

>  the code in the action. If you want to test that the chained task gets enquequed correctly, it needs to run as part of the action.
The attempt was to rewrite the same test case. The original intention was to test the enqueuing behavior and see if the chained task could be read and written properly. Originally, the enqueueing was a method call instead. Thank you for spotting that..I now trigger enqueuing by calling ```action.run()``` to check if the original task is enqueued and chain task could be read. 
</blockquote></details>

You are still not understanding entirely what the chained task is here, I'm afraid. `UpdateDatastoreBackupAction` launches an asynchronous job to import the datastore backup into Bigquery as a table, and it then enqueues a task which launches `BigqueryPollJobAction`. This task contains a chained task which would run `UpdateSnapshotViewAction` when the Bigquery import is complete.

When `BigqueryPollJobAction` runs, its payload *is* the chained task, not a task that contains the chained task. You need to test that `BigqueryPollJobAction` can deserialize this chained task and enqueue it, which is done in `testSuccess_enqueuePollTask_withChainedTask`.

As you pointed out the enqueueing of the task that contains a chained task was done in a method call, and the method was called in `UpdateDatastoreBackupAction`. So now you will want to add a test in `UpdateDatastoreBackupActionTest` to verify that the desired task (which contains a chained task as its payload) is correctly enqueued.

Separately, as I go through this convoluted logic (I know!) it doesn't seem to me that there's a particular reason to use the chained task at all. The thinking was that the Bigquery import takes a while, so we need to poll it periodically and only enqueue a task (the chained task) when it's done. But there are two reasons why it doesn't seem necessary: 1), we could very well just pass the required parameters to construct the chained task in the task params or headers, and then construct it in place inside `BigqueryPollJobAction`, and 2) I don't see why `BigqueryPollJobAction` is needed at all, what's wrong with just enqueuing a task to launch `UpdateSnapshotViewAction`, but modify the action to check the Bigquery import status first? If it's still running, throw a `NotModifiedException` and the task will retry. Basically moving the polling logic from `BigqueryPollJobAction` to `UpdateSnapshotViewAction`.

If you look at `CheckBackupAction`, it checks if the initial datastore backup to GCS is complete before enqueueing a task to run `UpdateDatabaseBackupAction`, which suggests that we should be able to combine them into one action also.

The overall logic is like this: (1. BackupDatastoreAction) Start datstore to GCS export, enqueue a task to poll the status-> (2. CheckBackupAction) poll the status of the export, if finished, enqueue a task to start GCS to Bigquery import -> (3. UpdateDatastoreBackupAction) start GCS to Bigquery import, enqueue an action to poll status -> (4. BigqueryPollJobAction) poll the status of the import. if finished, enqueue a task to update Bigquery view -> (5. UpdateSnapshotViewAction) update Bigquery view.

You can see steps 2 and 4 don't do much other than "poll some status, and enqueue another task if the status is successful". I don't see why they cannot be merged with steps 3 and 5, respectively. That can very well help anyone who reads this code get a much better handle of what is going on.

Not saying that we should do this refactoring, certainly not in this PR. But hopefully what I wrote here can help you understand what is going on with all these interacting actions.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 7 of 11 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1481#-Mv_WxhI827pGFOKH03F:-MwhIeM008jUozZ08sEE:b-xxa44v) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java#L90)):*

You are still not understanding entirely what the chained task is here, I'm afraid. `UpdateDatastoreBackupAction` launches an asynchronous job to import the datastore backup into Bigquery as a table, and it then enqueues a task which launches `BigqueryPollJobAction`. This task contains a chained task which would run `UpdateSnapshotViewAction` when the Bigquery import is complete.

When `BigqueryPollJobAction` runs, its payload *is* the chained task, not a task that contains the chained task. You need to test that `BigqueryPollJobAction` can deserialize this chained task and enqueue it, which is done in `testSuccess_enqueuePollTask_withChainedTask`.

As you pointed out the enqueueing of the task that contains a chained task was done in a method call, and the method was called in `UpdateDatastoreBackupAction`. So now you will want to add a test in `UpdateDatastoreBackupActionTest` to verify that the desired task (which contains a chained task as its payload) is correctly enqueued.

Separately, as I go through this convoluted logic (I know!) it doesn't seem to me that there's a particular reason to use the chained task at all. The thinking was that the Bigquery import takes a while, so we need to poll it periodically and only enqueue a task (the chained task) when it's done. But there are two reasons why it doesn't seem necessary: 1), we could very well just pass the required parameters to construct the chained task in the task params or headers, and then construct it in place inside `BigqueryPollJobAction`, and 2) I don't see why `BigqueryPollJobAction` is needed at all, what's wrong with just enqueuing a task to launch `UpdateSnapshotViewAction`, but modify the action to check the Bigquery import status first? If it's still running, throw a `NotModifiedException` and the task will retry. Basically moving the polling logic from `BigqueryPollJobAction` to `UpdateSnapshotViewAction`.

If you look at `CheckBackupAction`, it checks if the initial datastore backup to GCS is complete before enqueueing a task to run `UpdateDatabaseBackupAction`, which suggests that we should be able to combine them into one action also.

The overall logic is like this: (1. BackupDatastoreAction) Start datstore to GCS export, enqueue a task to poll the status-> (2. CheckBackupAction) poll the status of the export, if finished, enqueue a task to start GCS to Bigquery import -> (3. UpdateDatastoreBackupAction) start GCS to Bigquery import, enqueue an action to poll status -> (4. BigqueryPollJobAction) poll the status of the import. if finished, enqueue a task to update Bigquery view -> (5. UpdateSnapshotViewAction) update Bigquery view.

You can see steps 2 and 4 don't do much other than "poll some status, and enqueue another task if the status is successful". I don't see why they cannot be merged with steps 3 and 5, respectively. That can very well help anyone who reads this code get a much better handle of what is going on.

Not saying that we should do this refactoring, certainly not in this PR. But hopefully what I wrote here can help you understand what is going on with all these interacting actions.
</blockquote></details>

Thanks for your detailed explanation and it definitely confirmed and expanded what I've learned from going through the code.

I took a closer look again and I was trying to write something similar to the original test case, and neglected the reason behind this test case. This should have been removed since I already removed ```BigqueryPollJobEnqueuer```. One of the reasons I deleted the case above ```testSuccess_enqueuePollTask()`` ` was that it was testing ```enqueuePOllTask()``` in ```BigQueryPollJobEnqueuer```, which was same as this one but with a different variation of the task. 
The test cases in this file should ensure that everything in ```BigqueryPollJObAction``` is tested. The test case below  ```testSuccess_chainedPayloadAndJobSucceeded_enqueuesChainedTask``` already justifies that when the action runs, the task is enqueued to the right queue, and the chained task can be read properly. 

The reason ```BigqueryPollJobEnqueuer``` was no longer needed was that enqueueing can just be done via an injection of ```CloudTasksUtil```instance, at where enqueueing happens since ```PATH``` and ```SERVICE``` can be specified. Enqueueing is currently tested in where ```BigqueryPOllJobEnqueuer``` was used which was ```UploadDatastoreBackupAction```.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 7 of 11 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java, line 165 at r5](https://reviewable.io/reviews/google/nomulus/1481#-MvZPJx5Eh1MmetmuNlx:-MxB0cOl1NOSUR7Dt8NQ:b-qtxcx4) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/main/java/google/registry/export/UploadDatastoreBackupAction.java#L165)):*

Can use https://github.com/google/nomulus/pull/1521 after it is merged.
</blockquote></details>

Because headers are involved in this particular task, there's no helper method in ```CloudTasksUtil```that would work with this. However, ```scheduleTime``` doesn't have to be declared as a variable so I could improve that here.

jianglai (changes requested):
> Reviewed 4 of 4 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java, line 90 at r5](https://reviewable.io/reviews/google/nomulus/1481#-Mv_WxhI827pGFOKH03F:-MwEm1y13ar7YrFqm7n4:b-gk8acw) ([raw file](https://github.com/google/nomulus/blob/5eb90733613157b2a1e935d66b6bf249541fb2b3/core/src/test/java/google/registry/export/BigqueryPollJobActionTest.java#L90)):*

Thanks for your detailed explanation and it definitely confirmed and expanded what I've learned from going through the code.

I took a closer look again and I was trying to write something similar to the original test case, and neglected the reason behind this test case. This should have been removed since I already removed ```BigqueryPollJobEnqueuer```. One of the reasons I deleted the case above ```testSuccess_enqueuePollTask()`` ` was that it was testing ```enqueuePOllTask()``` in ```BigQueryPollJobEnqueuer```, which was same as this one but with a different variation of the task. 
The test cases in this file should ensure that everything in ```BigqueryPollJObAction``` is tested. The test case below  ```testSuccess_chainedPayloadAndJobSucceeded_enqueuesChainedTask``` already justifies that when the action runs, the task is enqueued to the right queue, and the chained task can be read properly. 

The reason ```BigqueryPollJobEnqueuer``` was no longer needed was that enqueueing can just be done via an injection of ```CloudTasksUtil```instance, at where enqueueing happens since ```PATH``` and ```SERVICE``` can be specified. Enqueueing is currently tested in where ```BigqueryPOllJobEnqueuer``` was used which was ```UploadDatastoreBackupAction```.  
</blockquote></details>

You are welcome. This chain of actions and how they are linked is indeed confusing for most. Luckily I think we can get rid of them soon after we no longer use datastore.

You still need to add verification in `testSuccess_enqueuePollTask_withChainedTask()` below to `UploadDatastoreActionTest`. It is to test that your serialization works as expected. in `testSuccess_chainedPayloadAndJobSucceeded_enqueuesChainedTask()` you verified that de-serialization works. But since the serialization now happens in-place in `UploadDatastoreAction`, you need to verify that the enqueued tasks contain the expected chained tasks as a header.
___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 162 at r10](https://reviewable.io/reviews/google/nomulus/1481#-MxI6lvz7OkirnS7oNjY:-MxI6lvz7OkirnS7oNjZ:b-7iaaqt) ([raw file](https://github.com/google/nomulus/blob/9be76041ac49bbe4a1c03c57565c876964ee57fb/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L162)):*
> ```Java
>             .header(PROJECT_ID_HEADER, "Project-Id")
>             .header(JOB_ID_HEADER, "load-backup-2018_12_05T17_46_39_92612-three")
>             .header(CHAINED_TASK_QUEUE_HEADER, UpdateSnapshotViewAction.QUEUE)
> ```

Namely here. In addition to check the queue headers, you also need to get the payload and verify that it contains the expected chained test.



_Code quote:_
```Java
.header(CHAINED_TASK_QUEUE_HEADER, UpdateSnapshotViewAction.QUEUE)
```

jianglai (changes requested):
> Reviewed 4 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: all files reviewed, 3 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/request/RequestModule.java, line 190 at r11](https://reviewable.io/reviews/google/nomulus/1481#-MxSCj6P7tWaFRC9_rn3:-MxSCj6P7tWaFRC9_rn4:b-wodlyy) ([raw file](https://github.com/google/nomulus/blob/13c2ed68448505aa846ccc4f6649976597e0fd43/core/src/main/java/google/registry/request/RequestModule.java#L190)):*
> ```Java
>   @Provides
>   @Payload
>   static ByteString providePayloadAsByteString(HttpServletRequest req) {
> ```

Any particular reason why this provider is added?



_Code quote:_
```Java
static ByteString providePayloadAsByteString(HttpServletRequest req) {
```
___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 177 at r11](https://reviewable.io/reviews/google/nomulus/1481#-MxSADW5-O-iwAEoaW5y:-MxSADW5-O-iwAEoaW5z:b-2cf76k) ([raw file](https://github.com/google/nomulus/blob/13c2ed68448505aa846ccc4f6649976597e0fd43/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L177)):*
> ```Java
>     List<Task> enqueudTasks = cloudTasksHelper.getTestTasksFor(BigqueryPollJobAction.QUEUE);
>     // assert the chained task of each enqueud task is correct
>     assertThat(
> ```

Consider using `assertThat(enqueuedTasks.stream().map(...)).containsExactly(...).inOrder()`.



_Code quote:_
```Java
assertThat(
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 11 of 12 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/request/RequestModule.java, line 190 at r11](https://reviewable.io/reviews/google/nomulus/1481#-MxSCj6P7tWaFRC9_rn3:-Mx_z7JUCMr-8a3zvJtB:b-sqqzmo) ([raw file](https://github.com/google/nomulus/blob/13c2ed68448505aa846ccc4f6649976597e0fd43/core/src/main/java/google/registry/request/RequestModule.java#L190)):*

Any particular reason why this provider is added?
</blockquote></details>

I did it because with ```Task Queue```, payload was taken in as ```byte[]```, which makes sense to have ```byte[]``` for payload. However, with ```Cloud Task```API, ```setBody()``` takes ```ByteString`` instead of ```byte[]```. Since the original request type has not changed, I think it's okay to keep using the existing provider. I just thought it would make sense to have the type consistent with the parameter itself. 
___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 162 at r10](https://reviewable.io/reviews/google/nomulus/1481#-MxI6lvz7OkirnS7oNjY:-Mx_DiFd5jP_6iTX5oKg:b-wpwjwn) ([raw file](https://github.com/google/nomulus/blob/9be76041ac49bbe4a1c03c57565c876964ee57fb/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L162)):*

Namely here. In addition to check the queue headers, you also need to get the payload and verify that it contains the expected chained test.
</blockquote></details>

this was done below using data from ```enqueuedTasks```
___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 177 at r11](https://reviewable.io/reviews/google/nomulus/1481#-MxSADW5-O-iwAEoaW5y:-Mx_DhK890sKeJie4Pxy:b-896fix) ([raw file](https://github.com/google/nomulus/blob/13c2ed68448505aa846ccc4f6649976597e0fd43/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L177)):*

Consider using `assertThat(enqueuedTasks.stream().map(...)).containsExactly(...).inOrder()`.
</blockquote></details>

Done.

jianglai (changes requested):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 227 at r12](https://reviewable.io/reviews/google/nomulus/1481#-MxbNNeT82Uw_Sxl8iD0:-MxbNNeT82Uw_Sxl8iD1:b-5q8gcy) ([raw file](https://github.com/google/nomulus/blob/2436403c6a270b1ab3448c798ffb211c00c22750/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L227)):*
> ```Java
>         .inOrder();
>   }
>   
> ```

What is this...?



_Code quote:_
```Java
···
```
___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 228 at r12](https://reviewable.io/reviews/google/nomulus/1481#-MxbNkag4wlwaHu_Hy_h:-MxbNkag4wlwaHu_Hy_i:bxpvqyb) ([raw file](https://github.com/google/nomulus/blob/2436403c6a270b1ab3448c798ffb211c00c22750/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L228)):*
> ```Java
>   }
>   
>   Task getChainedTask(Task task) {
> ```

This can just be an in-place lambda. Also if all you want is to get around the checked exception inside the map function, why not just return a `null` in the catch block?



_Code quote:_
```Java
getChainedTask(Task task)
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 227 at r12](https://reviewable.io/reviews/google/nomulus/1481#-MxbNNeT82Uw_Sxl8iD0:-MxesSnG0Ml4FaYpKtXI:bftd85f) ([raw file](https://github.com/google/nomulus/blob/2436403c6a270b1ab3448c798ffb211c00c22750/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L227)):*

What is this...?
</blockquote></details>

no idea. my apology! 
___
*[core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java, line 228 at r12](https://reviewable.io/reviews/google/nomulus/1481#-MxbNkag4wlwaHu_Hy_h:-Mxeurtw7uBOqkfnDRfA:b5axsbe) ([raw file](https://github.com/google/nomulus/blob/2436403c6a270b1ab3448c798ffb211c00c22750/core/src/test/java/google/registry/export/UploadDatastoreBackupActionTest.java#L228)):*

This can just be an in-place lambda. Also if all you want is to get around the checked exception inside the map function, why not just return a `null` in the catch block?
</blockquote></details>

done. I tried that earlier. Must have been the parentheses that caused some issue. I was thinking to return null in the catch block. The task created there is a task with no info, so it's about choosing null or empty. I guess going with null is probably a better choice to show the error handling.

jianglai (approved):
> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1481)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

PR #1484, opened 2022-01-06
Use the service account credential to delete unused versions

jianglai:
Right now we are using the default credential of Cloud Build to run
gcloud, which does not have the necessary permission. By using the
service account credential (which is a GAE admin), we should be able to
delete unused versions as part of deployment.

Note: it looks like this never worked when it was added to the
deployment pipeline, not as a result of overground changes under our
feet.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1484)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1484)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1492, opened 2022-01-12
CommitLog handling code should call ofyTm

weiminyu:
The tm() call will use JPA transaction manager after the switch-over to
SQL. These calls would lose their transaction semantics.

Both actions are to be invoked after the switchover in case we have to
switch back to Datastore as primary.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1492)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1492)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1491, opened 2022-01-12
Remove ReportingUtils and use CloudTasksUtil to enqueue tasks in GenerateInvoicesAction and GenerateSpec11ReportAction

rachelguan:
ReportingUtils is a utility class that contains one method and it's used in two actions. `GenerateInvoicesAction` and `GenerateSpec11ReportAction`. With `CloudTasksUtil` that uses Cloud Task API, a task can be created and enqueued to a queue of your choice, without the need of using a static method. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1491)

PR #1489, opened 2022-01-12
Use CloudTasksUtil to enqueue task in IcannReportingStagingAction 

rachelguan:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1489)

PR #1493, opened 2022-01-12
Add support for delay of duration when scheduling a task

rachelguan:
Currently, there are two ways to create a task via `CloudTasksUtil`: one with no specific schedule time, and one with random delay. There needs to be a way to schedule a task by specifying the delay time.

Added `scheduleTime` as one of the matchers in `CloudTasksHelper`. `Instant` is used for time comparison since it allows us to create the time slightly easier by using millis. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1493)

### DIFF
--- a/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java
+++ b/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java
@@ -83,6 +83,7 @@ public class CommitLogCheckpointActionTest {
         QUEUE_NAME,
         new TaskMatcher()
             .url(ExportCommitLogDiffAction.PATH)
+            .scheduleTime(null)
             .param(ExportCommitLogDiffAction.LOWER_CHECKPOINT_TIME_PARAM, oneMinuteAgo.toString())
             .param(ExportCommitLogDiffAction.UPPER_CHECKPOINT_TIME_PARAM, now.toString()));
     assertThat(loadRoot().getLastWrittenTime()).isEqualTo(now);
--- a/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java
+++ b/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java
@@ -23,6 +23,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.protobuf.Timestamp;
 import google.registry.model.ofy.CommitLogCheckpoint;
 import google.registry.model.ofy.CommitLogCheckpointRoot;
 import google.registry.testing.AppEngineExtension;
@@ -67,6 +68,7 @@ public class CommitLogCheckpointActionTest {
         QUEUE_NAME,
         new TaskMatcher()
             .url(ExportCommitLogDiffAction.PATH)
+            .scheduleTime(Timestamp.getDefaultInstance())
             .param(ExportCommitLogDiffAction.LOWER_CHECKPOINT_TIME_PARAM, START_OF_TIME.toString())
             .param(ExportCommitLogDiffAction.UPPER_CHECKPOINT_TIME_PARAM, now.toString()));
     assertThat(loadRoot().getLastWrittenTime()).isEqualTo(now);
--- a/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java
+++ b/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java
@@ -23,7 +23,6 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import com.google.common.collect.ImmutableMap;
-import com.google.protobuf.Timestamp;
 import google.registry.model.ofy.CommitLogCheckpoint;
 import google.registry.model.ofy.CommitLogCheckpointRoot;
 import google.registry.testing.AppEngineExtension;
@@ -68,7 +67,6 @@ public class CommitLogCheckpointActionTest {
         QUEUE_NAME,
         new TaskMatcher()
             .url(ExportCommitLogDiffAction.PATH)
-            .scheduleTime(Timestamp.getDefaultInstance())
             .param(ExportCommitLogDiffAction.LOWER_CHECKPOINT_TIME_PARAM, START_OF_TIME.toString())
             .param(ExportCommitLogDiffAction.UPPER_CHECKPOINT_TIME_PARAM, now.toString()));
     assertThat(loadRoot().getLastWrittenTime()).isEqualTo(now);
@@ -83,7 +81,6 @@ public class CommitLogCheckpointActionTest {
         QUEUE_NAME,
         new TaskMatcher()
             .url(ExportCommitLogDiffAction.PATH)
-            .scheduleTime(null)
             .param(ExportCommitLogDiffAction.LOWER_CHECKPOINT_TIME_PARAM, oneMinuteAgo.toString())
             .param(ExportCommitLogDiffAction.UPPER_CHECKPOINT_TIME_PARAM, now.toString()));
     assertThat(loadRoot().getLastWrittenTime()).isEqualTo(now);
--- a/core/src/test/java/google/registry/testing/CloudTasksHelper.java
+++ b/core/src/test/java/google/registry/testing/CloudTasksHelper.java
@@ -39,6 +39,7 @@ import com.google.common.collect.Multimaps;
 import com.google.common.net.HttpHeaders;
 import com.google.common.net.MediaType;
 import com.google.common.truth.Truth8;
+import com.google.protobuf.Timestamp;
 import google.registry.model.ImmutableObject;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.Retrier;
@@ -195,6 +196,7 @@ public class CloudTasksHelper implements Serializable {
     // tests.
     HttpMethod method = HttpMethod.POST;
     String url;
+    Timestamp scheduleTime;
     Multimap<String, String> headers = ArrayListMultimap.create();
     Multimap<String, String> params = ArrayListMultimap.create();
 
@@ -216,6 +218,7 @@ public class CloudTasksHelper implements Serializable {
           Ascii.toLowerCase(task.getAppEngineHttpRequest().getAppEngineRouting().getService());
       method = task.getAppEngineHttpRequest().getHttpMethod();
       url = uri.getPath();
+      scheduleTime = task.getScheduleTime();
       ImmutableMultimap.Builder<String, String> headerBuilder = new ImmutableMultimap.Builder<>();
       task.getAppEngineHttpRequest()
           .getHeadersMap()
@@ -251,6 +254,7 @@ public class CloudTasksHelper implements Serializable {
       builder.put("url", url);
       builder.put("headers", headers);
       builder.put("params", params);
+      builder.put("scheduleTime", scheduleTime);
       return Maps.filterValues(builder, not(in(asList(null, "", Collections.EMPTY_MAP))));
     }
   }
@@ -293,6 +297,11 @@ public class CloudTasksHelper implements Serializable {
       return this;
     }
 
+    public TaskMatcher scheduleTime(Timestamp scheduleTime) {
+      expected.scheduleTime = scheduleTime;
+      return this;
+    }
+
     public TaskMatcher param(String key, String value) {
       checkNotNull(value, "Test error: A param can never have a null value, so don't assert it");
       expected.params.put(key, value);
@@ -324,6 +333,8 @@ public class CloudTasksHelper implements Serializable {
           && (expected.url == null || Objects.equals(expected.url, actual.url))
           && (expected.method == null || Objects.equals(expected.method, actual.method))
           && (expected.service == null || Objects.equals(expected.service, actual.service))
+          && (expected.scheduleTime == null
+              || Objects.equals(expected.scheduleTime, actual.scheduleTime))
           && containsEntries(actual.params, expected.params)
           && containsEntries(actual.headers, expected.headers);
     }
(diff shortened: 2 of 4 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 204 at r1](https://reviewable.io/reviews/google/nomulus/1493#-MtLcRFf1O2ZP-7fNnqx:-MtLcRFf1O2ZP-7fNnqy:b1fc2k3) ([raw file](https://github.com/google/nomulus/blob/4a35dfc6b2862137d3361f7fb4c32087e7ce7006/util/src/main/java/google/registry/util/CloudTasksUtils.java#L204)):*
> ```Java
>       Multimap<String, String> params,
>       Clock clock,
>       long duration) {
> ```

Probably better to pass an actual `Duration`, and the variable should be named `delay`.



_Code quote:_
```Java
long duration
```

jianglai (commented):
We should modify `TaskMatcher` to add a `scheduleTime` parameter that allows us to verify the delay as well.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493#-:-MtLfdEn-CJFmLliU1RU:bej6zui)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 204 at r1](https://reviewable.io/reviews/google/nomulus/1493#-MtLcRFf1O2ZP-7fNnqx:-MuCRK9t4DD-QfTVNDL0:buxr6ro) ([raw file](https://github.com/google/nomulus/blob/4a35dfc6b2862137d3361f7fb4c32087e7ce7006/util/src/main/java/google/registry/util/CloudTasksUtils.java#L204)):*

Probably better to pass an actual `Duration`, and the variable should be named `delay`.
</blockquote></details>

yep..changed that.

jianglai (dismissed):
Still need to modify `TaskMatcher`.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493#-:-MuDh9Iu1eyn5F6-n6Pl:bz3quo9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

rachelguan (commented):
I was thinking to do that in a separate PR (one of the followings), where the assertion has to be called that would trigger `TaskMatcher` in an action test file. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493#-:-MuE04fJ7DcUTC1BhqR0:b-qjm07a)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 224 at r3](https://reviewable.io/reviews/google/nomulus/1493#-MuI5DmD6ixTC1MLmsfO:-MuI5DmD6ixTC1MLmsfP:bgqoj6e) ([raw file](https://github.com/google/nomulus/blob/ed7469cf84f1431d16a371e9f2e0898dde2d069e/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L224)):*
> ```Java
>       Timestamp taskScheduleTime = task.getScheduleTime();
>       scheduleTime =
>           Instant.ofEpochSecond(taskScheduleTime.getSeconds() + taskScheduleTime.getNanos());
> ```

Are you sure this is right? It looks like you are adding seconds with nanoseconds without regard to the units.

You mentioned that it's easier to compare with `Instant`, but Timestamp also implements [`equals()`](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Timestamp.html#equals-java.lang.Object-). What's wrong with using that?

Even if we can't use the `equals` method, we could just compare the seconds and nanoseconds in the `test` method, right? Seems there's no downside in saving the value as `timestamp`s.



_Code quote:_
```Java
    Instant.ofEpochSecond(taskScheduleTime.getSeconds() + taskScheduleTime.getNanos());
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 224 at r3](https://reviewable.io/reviews/google/nomulus/1493#-MuI5DmD6ixTC1MLmsfO:-MuILVP166r9YODJ9aMF:b-gd1zna) ([raw file](https://github.com/google/nomulus/blob/ed7469cf84f1431d16a371e9f2e0898dde2d069e/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L224)):*

Are you sure this is right? It looks like you are adding seconds with nanoseconds without regard to the units.

You mentioned that it's easier to compare with `Instant`, but Timestamp also implements [`equals()`](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Timestamp.html#equals-java.lang.Object-). What's wrong with using that?

Even if we can't use the `equals` method, we could just compare the seconds and nanoseconds in the `test` method, right? Seems there's no downside in saving the value as `timestamp`s.
</blockquote></details>

yea, I see your reasoning. I reverted it back to `Timestamp`.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 224 at r3](https://reviewable.io/reviews/google/nomulus/1493#-MuI5DmD6ixTC1MLmsfO:-MunPAMH80DSgQnK_o7d:bb71pro) ([raw file](https://github.com/google/nomulus/blob/ed7469cf84f1431d16a371e9f2e0898dde2d069e/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L224)):*

yea, I see your reasoning. I reverted it back to `Timestamp`.
</blockquote></details>

PTAL

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java, line 71 at r5](https://reviewable.io/reviews/google/nomulus/1493#-MusN6l1EGd7SraFmpyr:-MusN6l276nKUcqGSCHH:b1pi203) ([raw file](https://github.com/google/nomulus/blob/30fb35abac7050770845d99abed01e98ebeb8629/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java#L71)):*
> ```Java
>         new TaskMatcher()
>             .url(ExportCommitLogDiffAction.PATH)
>             .scheduleTime(Timestamp.getDefaultInstance())
> ```

Does it need to be set? If you leave it to be `null`, it will not be compared with the actual value. That's how the other fields work.



_Code quote:_
```Java
Timestamp.getDefaultInstance())
```
___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 224 at r3](https://reviewable.io/reviews/google/nomulus/1493#-MuI5DmD6ixTC1MLmsfO:-MusB1Nj0VfhOmHeTa1j:bgtlxck) ([raw file](https://github.com/google/nomulus/blob/ed7469cf84f1431d16a371e9f2e0898dde2d069e/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L224)):*

PTAL
</blockquote></details>

In the future, please write PTAL or other comments that are not replies to a specific comment to the general section at the top of the PR.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java, line 71 at r5](https://reviewable.io/reviews/google/nomulus/1493#-MusN6l1EGd7SraFmpyr:-Muv1DjX7TxNWUT0L8Mm:b-ucx7ji) ([raw file](https://github.com/google/nomulus/blob/30fb35abac7050770845d99abed01e98ebeb8629/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java#L71)):*

Does it need to be set? If you leave it to be `null`, it will not be compared with the actual value. That's how the other fields work.
</blockquote></details>

so far, there's no example in the code right now that actually sets delayed timestamp except the ones with jitter seconds, which can't be compared with an expected one since it's random. Schedule time is specified in other PRs that require this one. 
___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 224 at r3](https://reviewable.io/reviews/google/nomulus/1493#-MuI5DmD6ixTC1MLmsfO:-Muv2oEV7sQFJYEAHYD7:b-mx77jz) ([raw file](https://github.com/google/nomulus/blob/ed7469cf84f1431d16a371e9f2e0898dde2d069e/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L224)):*

In the future, please write PTAL or other comments that are not replies to a specific comment to the general section at the top of the PR.
</blockquote></details>

noted.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java, line 71 at r5](https://reviewable.io/reviews/google/nomulus/1493#-MusN6l1EGd7SraFmpyr:-MuwTcxg8z3QR3MW7hF_:b-tsng6l) ([raw file](https://github.com/google/nomulus/blob/30fb35abac7050770845d99abed01e98ebeb8629/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java#L71)):*

so far, there's no example in the code right now that actually sets delayed timestamp except the ones with jitter seconds, which can't be compared with an expected one since it's random. Schedule time is specified in other PRs that require this one. 
</blockquote></details>

I added this case since ```Timestamp.getDefaultInstance()``` is ```DEFAULT_INSTANCE```, which is ```new com.google.protobuf.Timestamp()```, which is not null. So that supposedly should cover the branch in the comparison method, ```Objects.equals(expected.scheduleTime, actual.scheduleTime)```. 

Or the comparison could be tested after this PR goes in, the comparison will be added in [PR: Remove ReportingUtils](https://github.com/google/nomulus/pull/1491) and [PR: Use CloudTasksUtil to enqueue task in IcannReportingStagingAction](https://github.com/google/nomulus/pull/1489), since these two were the reason behind adding a helper that allows specifying the schedule time.

rachelguan (commented):
PTAL and thank you in advance! 

I explained in the comments why ```scheduleTime(null)``` and ```scheduleTime(Timestamp.getDefaultInstance())``` were added in existing task comparison. Schedule time comparison is not done for the delayed tasks yet since the delayed tasks are not created using Cloud Task API right now. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493#-:-MuwVDx77a08_bggr9h_:by9v1pj)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java, line 71 at r5](https://reviewable.io/reviews/google/nomulus/1493#-MusN6l1EGd7SraFmpyr:-MuxMuHUCIXjZM5PSPji:b-tce2gi) ([raw file](https://github.com/google/nomulus/blob/30fb35abac7050770845d99abed01e98ebeb8629/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java#L71)):*

I added this case since ```Timestamp.getDefaultInstance()``` is ```DEFAULT_INSTANCE```, which is ```new com.google.protobuf.Timestamp()```, which is not null. So that supposedly should cover the branch in the comparison method, ```Objects.equals(expected.scheduleTime, actual.scheduleTime)```. 

Or the comparison could be tested after this PR goes in, the comparison will be added in [PR: Remove ReportingUtils](https://github.com/google/nomulus/pull/1491) and [PR: Use CloudTasksUtil to enqueue task in IcannReportingStagingAction](https://github.com/google/nomulus/pull/1489), since these two were the reason behind adding a helper that allows specifying the schedule time.
</blockquote></details>

My point is that if you look at other fields in the matcher, say `taskName`, if you don't set it, the `equals` method will not use it for comparison. The fact that the default value of `scheduleTime` is `DEFAULT_INSTANCE` is an implementation detail that we do not control, and it could very well change in the future. We don't really care what the default value is. What we do care is that if we override it, it should be the overriding value. That's why we ignore the null fields. Does this make sense?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java, line 71 at r5](https://reviewable.io/reviews/google/nomulus/1493#-MusN6l1EGd7SraFmpyr:-Mv-Oqwk9N48tJxgFESN:bwhow1y) ([raw file](https://github.com/google/nomulus/blob/30fb35abac7050770845d99abed01e98ebeb8629/core/src/test/java/google/registry/backup/CommitLogCheckpointActionTest.java#L71)):*
> use it for comparison. The fact that the default value of scheduleTime is DEFAULT_INSTANCE is an implementation detail that we do not control, and it could very well change in the future. We don't really care what the default value is. What we do care is that if we override it, it should be the overriding value. That's why we ignore the null fields. Does this make sense?
i see you point. My intention behind the modification of the two testing cases was to show that the schedule time in matcher works since there's no unit test for that ```CloudTasksHelper```. I could remove and show the comparison of the values we have control over.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1493)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

PR #1495, opened 2022-01-19
Use a read-only replica SQL instance in RdapDomainSearchAction

gbrodman:
We can use it more places later but this can serve as a template. We
should inject the connection to the read-only replica (only created
once) to the constructor of the action, then use that instead of the
regular transaction manager.

We add a transaction manager that simulates the read-only-replica
behavior for testing purposes as well.

In addition, we set the transaction isolation level to READ COMMITTED
for this transaction manager (this is fine since we're never writing to
it). Postgres requires this for replica SQL access (it fails if we try
to use SERIALIZABLE) transactions. We didn't see this with the pipelines
before since those already had transaction isolation level overrides

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1495)

### DIFF
--- a/core/src/main/java/google/registry/module/pubapi/PubApiComponent.java
+++ b/core/src/main/java/google/registry/module/pubapi/PubApiComponent.java
@@ -30,6 +30,7 @@ import google.registry.keyring.api.KeyModule;
 import google.registry.keyring.kms.KmsModule;
 import google.registry.module.pubapi.PubApiRequestComponent.PubApiRequestComponentModule;
 import google.registry.monitoring.whitebox.StackdriverModule;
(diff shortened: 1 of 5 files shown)


gbrodman:
note: I still need to test this on alpha, will update with that's done

gbrodman:
ok this works on Alpha

mindhog (approved):
> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1495)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1496, opened 2022-01-19
Make sql-based lock more robust

weiminyu:
Currently lock/unlock transactions for SQL-based locks use the the
SERIALIZABLE isolation level. This is an overkill since locking involves
only a single table row. Using SERIALIZABLE level may cause the
transactions to rollback for irrelevant database operations.

In this PR we set lock transactions' isolation level to REPEATABLE_READ.
This is a temporary, ad hoc solution. A better transaction API being
planned will address this more elegantly.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1496)

### DIFF
--- a/core/src/main/java/google/registry/model/server/Lock.java
+++ b/core/src/main/java/google/registry/model/server/Lock.java
@@ -22,6 +22,7 @@ import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.flogger.FluentLogger;
 import com.googlecode.objectify.Key;
@@ -33,6 +34,7 @@ import google.registry.model.annotations.NotBackedUp.Reason;
 import google.registry.model.replay.DatastoreAndSqlEntity;
 import google.registry.persistence.VKey;
 import google.registry.persistence.transaction.JpaTransactionManager;
+import google.registry.persistence.transaction.ReadOnlyCheckingEntityManager;
 import google.registry.persistence.transaction.TransactionManager;
 import google.registry.util.RequestStatusChecker;
 import google.registry.util.RequestStatusCheckerImpl;
@@ -262,8 +264,9 @@ public class Lock extends ImmutableObject implements DatastoreAndSqlEntity, Seri
     // must be definitively acquired before it is used, even when called inside another transaction.
     Supplier<AcquireResult> lockAcquirer =
         () -> {
-          DateTime now = transactionManager.getTransactionTime();
+          setIsolationLevelForLockingTransaction(transactionManager);
 
+          DateTime now = transactionManager.getTransactionTime();
           // Checking if an unexpired lock still exists - if so, the lock can't be acquired.
           Lock lock =
               transactionManager
@@ -297,6 +300,15 @@ public class Lock extends ImmutableObject implements DatastoreAndSqlEntity, Seri
 
           return AcquireResult.create(now, lock, newLock, lockState);
         };
+
+    if (!transactionManager.isOfy()) {
+      // Unlike ofy, JpaTransactionManager does not implement transactNew. Need this check to
+      // catch improper usage.
+      // TODO(b/193662898): implement and use transactNew in jpa transaction API.
+      Preconditions.checkState(
+          !transactionManager.inTransaction(),
+          "SQL-backed lock cannot be acquired from inside a transaction.");
+    }
     // In ofy, backup is determined per-action, but in SQL it's determined per-transaction
     AcquireResult acquireResult =
         transactionManager.isOfy()
@@ -313,6 +325,8 @@ public class Lock extends ImmutableObject implements DatastoreAndSqlEntity, Seri
     // Just use the default clock because we aren't actually doing anything that will use the clock.
     Supplier<Void> lockReleaser =
         () -> {
+          setIsolationLevelForLockingTransaction(transactionManager);
+
           // To release a lock, check that no one else has already obtained it and if not
           // delete it. If the lock in Datastore was different then this lock is gone already;
           // this can happen if release() is called around the expiration time and the lock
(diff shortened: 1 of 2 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1496#-Mtri6UZ8LpuvDq8RV3D:-Mtri6UZ8LpuvDq8RV3E:bdel62i) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L385)):*
> ```Java
>       // REPEATABLE READ is enough since transaction reads and writes a single row.
>       entityManager
>           .createNativeQuery("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ")
> ```

How does this not affect non-lock transactions?  (I would expect it to be scoped to the connection)

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1496#-Mtri6UZ8LpuvDq8RV3D:-MtrivJM0MzhfYRJxjQ-:b-vupooc) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L385)):*

How does this not affect non-lock transactions?  (I would expect it to be scoped to the connection)
</blockquote></details>

Ah, I see.  it applies to the current transaction.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 359 at r1](https://reviewable.io/reviews/google/nomulus/1496#-MtsIagR7RI23yYTkpu4:-MtsIagR7RI23yYTkpu5:b-u2ztgz) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L359)):*
> ```Java
>           return null;
>         };
> ```

Shouldn't we also check here to make sure that we're not already in a transaction, like we do when we acquire the lock?

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 359 at r1](https://reviewable.io/reviews/google/nomulus/1496#-MtsIagR7RI23yYTkpu4:-MuByLyaDc7feRTlYUoX:bvh6vqt) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L359)):*

Lock acquire should always be in its own transaction. However, with the current code, if we do not change the transaction isolation level, 
it can execute inside a transaction without error, but the result is not what we want. Therefore we want to check for the enclosing transaction.

Lock release, however, may be in another transaction. With a richer api, we MAY even allow it if the enclosing transaction has the same isolation level. 
</blockquote></details>

I understand that lock acquisition needs to be in its own transaction, but why can lock release be contained in another transaction? I would imagine we could run in to the same issues with changing the isolation level mid-transaction.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 359 at r1](https://reviewable.io/reviews/google/nomulus/1496#-MtsIagR7RI23yYTkpu4:-MuC2wwb1fXyarugKaU2:b-xjaesm) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L359)):*

I understand that lock acquisition needs to be in its own transaction, but why can lock release be contained in another transaction? I would imagine we could run in to the same issues with changing the isolation level mid-transaction. 
</blockquote></details>

I see two differences between acquire and release:

1. Conceptually, without considering isolation level change, acquire must be in a new transaction, while release  may be grouped together with the data-access code.
2. Even with the isolation level change code added in this PR, it is possible for a transaction that includes a lock acquire to pass without notice, e.g., the following code may pass without errors, but this is not what we want:
  tm.transacti(() ->
    lock.acquireWithSql();
    other accesses.
  );
  On the other hand, if we change isolation level, lock.releaseWithSql() will always fail. If we don't change isolation level, it will also pass as intended. Therefore there is no need for extra alerting.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 359 at r1](https://reviewable.io/reviews/google/nomulus/1496#-MtsIagR7RI23yYTkpu4:-MuCAtnI1AtnBRIF1uHZ:b-yn54g7) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L359)):*

I see two differences between acquire and release:

1. Conceptually, without considering isolation level change, acquire must be in a new transaction, while release  may be grouped together with the data-access code.
2. Even with the isolation level change code added in this PR, it is possible for a transaction that includes a lock acquire to pass without notice, e.g., the following code may pass without errors, but this is not what we want:
  tm.transacti(() ->
    lock.acquireWithSql();
    other accesses.
  );
  On the other hand, if we change isolation level, lock.releaseWithSql() will always fail. If we don't change isolation level, it will also pass as intended. Therefore there is no need for extra alerting.
</blockquote></details>

I see, that makes sense thanks (mainly the reasoning behind number 2).

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/server/Lock.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1496#-MurPvK92w4Lg0XWV-Y2:-MurPvK92w4Lg0XWV-Y3:bmtw17) ([raw file](https://github.com/google/nomulus/blob/cd00f42a835744ea70e280ee6941a1c67e30ba62/core/src/main/java/google/registry/model/server/Lock.java#L385)):*
> ```Java
>       // REPEATABLE READ is enough since transaction reads and writes a single row.
>       entityManager
>           .createNativeQuery("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ")
> ```

And we don't need to use a try-with-resources pattern because SQL itself guarantees that this is wiped out after the transaction completes, right?

PR #1503, opened 2022-01-26
Use the replica jpaTm in FKI and EppResource cache methods

gbrodman:
The cached methods are only used in situations where we don't really
care about being 100% synchronously up to date (e.g. whois), and they're
not used frequently anyway, so it's safe to use the replica in these
locations.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1503)

### DIFF
--- a/core/src/main/java/google/registry/model/EppResource.java
+++ b/core/src/main/java/google/registry/model/EppResource.java
@@ -21,6 +21,7 @@ import static com.google.common.collect.Sets.union;
 import static google.registry.config.RegistryConfig.getEppResourceCachingDuration;
 import static google.registry.config.RegistryConfig.getEppResourceMaxCachedEntries;
 import static google.registry.persistence.transaction.TransactionManagerFactory.ofyTm;
+import static google.registry.persistence.transaction.TransactionManagerFactory.replicaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.CollectionUtils.nullToEmpty;
 import static google.registry.util.CollectionUtils.nullToEmptyImmutableCopy;
@@ -380,13 +381,13 @@ public abstract class EppResource extends BackupGroupRoot implements Buildable {
 
         @Override
         public EppResource load(VKey<? extends EppResource> key) {
-          return tm().doTransactionless(() -> tm().loadByKey(key));
+          return replicaTm().doTransactionless(() -> replicaTm().loadByKey(key));
         }
 
         @Override
         public Map<VKey<? extends EppResource>, EppResource> loadAll(
             Iterable<? extends VKey<? extends EppResource>> keys) {
-          return tm().doTransactionless(() -> tm().loadByKeys(keys));
+          return replicaTm().doTransactionless(() -> replicaTm().loadByKeys(keys));
         }
       };
 
--- a/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
+++ b/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java
@@ -21,6 +21,7 @@ import static google.registry.config.RegistryConfig.getEppResourceCachingDuratio
 import static google.registry.config.RegistryConfig.getEppResourceMaxCachedEntries;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.persistence.transaction.TransactionManagerFactory.replicaJpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.CollectionUtils.entriesToImmutableMap;
 import static google.registry.util.TypeUtils.instantiate;
@@ -51,6 +52,7 @@ import google.registry.model.host.HostResource;
 import google.registry.model.replay.DatastoreOnlyEntity;
 import google.registry.persistence.VKey;
 import google.registry.persistence.transaction.CriteriaQueryBuilder;
+import google.registry.persistence.transaction.JpaTransactionManager;
 import google.registry.util.NonFinalForTesting;
 import java.util.Collection;
 import java.util.Comparator;
@@ -198,7 +200,7 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
    */
   public static <E extends EppResource> ImmutableMap<String, ForeignKeyIndex<E>> load(
       Class<E> clazz, Collection<String> foreignKeys, final DateTime now) {
-    return loadIndexesFromStore(clazz, foreignKeys, true).entrySet().stream()
+    return loadIndexesFromStore(clazz, foreignKeys, true, false).entrySet().stream()
         .filter(e -> now.isBefore(e.getValue().getDeletionTime()))
         .collect(entriesToImmutableMap());
   }
@@ -217,7 +219,10 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
    */
   private static <E extends EppResource>
       ImmutableMap<String, ForeignKeyIndex<E>> loadIndexesFromStore(
-          Class<E> clazz, Collection<String> foreignKeys, boolean inTransaction) {
+          Class<E> clazz,
+          Collection<String> foreignKeys,
+          boolean inTransaction,
+          boolean useReplicaJpaTm) {
     if (tm().isOfy()) {
       Class<ForeignKeyIndex<E>> fkiClass = mapToFkiClass(clazz);
       return ImmutableMap.copyOf(
@@ -226,17 +231,18 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
               : tm().doTransactionless(() -> auditedOfy().load().type(fkiClass).ids(foreignKeys)));
     } else {
       String property = RESOURCE_CLASS_TO_FKI_PROPERTY.get(clazz);
+      JpaTransactionManager jpaTmToUse = useReplicaJpaTm ? replicaJpaTm() : jpaTm();
       ImmutableList<ForeignKeyIndex<E>> indexes =
-          tm().transact(
-                  () ->
-                      jpaTm()
-                          .criteriaQuery(
-                              CriteriaQueryBuilder.create(clazz)
-                                  .whereFieldIsIn(property, foreignKeys)
-                                  .build())
-                          .getResultStream()
-                          .map(e -> ForeignKeyIndex.create(e, e.getDeletionTime()))
-                          .collect(toImmutableList()));
+          jpaTmToUse.transact(
+              () ->
+                  jpaTmToUse
+                      .criteriaQuery(
+                          CriteriaQueryBuilder.create(clazz)
+                              .whereFieldIsIn(property, foreignKeys)
+                              .build())
+                      .getResultStream()
+                      .map(e -> ForeignKeyIndex.create(e, e.getDeletionTime()))
+                      .collect(toImmutableList()));
       // We need to find and return the entities with the maximum deletionTime for each foreign key.
       return Multimaps.index(indexes, ForeignKeyIndex::getForeignKey).asMap().entrySet().stream()
           .map(
@@ -260,7 +266,8 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
               loadIndexesFromStore(
                       RESOURCE_CLASS_TO_FKI_CLASS.inverse().get(key.getKind()),
                       ImmutableSet.of(foreignKey),
-                      false)
+                      false,
+                      true)
                   .get(foreignKey));
         }
 
@@ -276,7 +283,7 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
               Streams.stream(keys).map(v -> v.getSqlKey().toString()).collect(toImmutableSet());
           ImmutableSet<VKey<ForeignKeyIndex<?>>> typedKeys = ImmutableSet.copyOf(keys);
           ImmutableMap<String, ? extends ForeignKeyIndex<? extends EppResource>> existingFkis =
-              loadIndexesFromStore(resourceClass, foreignKeys, false);
+              loadIndexesFromStore(resourceClass, foreignKeys, false, true);
           // ofy omits keys that don't have values in Datastore, so re-add them in
           // here with Optional.empty() values.
           return Maps.asMap(
@@ -336,7 +343,7 @@ public abstract class ForeignKeyIndex<E extends EppResource> extends BackupGroup
     // Safe to cast VKey<FKI<E>> to VKey<FKI<?>>
     @SuppressWarnings("unchecked")
     ImmutableList<VKey<ForeignKeyIndex<?>>> fkiVKeys =
-        Streams.stream(foreignKeys)
+        foreignKeys.stream()
             .map(fk -> (VKey<ForeignKeyIndex<?>>) VKey.create(fkiClass, fk))
             .collect(toImmutableList());
     try {
--- a/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java
+++ b/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java
@@ -14,8 +14,8 @@
 
 package google.registry.persistence.transaction;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
+import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
 import static org.joda.time.DateTimeZone.UTC;
 
 import com.google.appengine.api.utils.SystemProperty;
@@ -47,6 +47,10 @@ public final class TransactionManagerFactory {
   private static Supplier<JpaTransactionManager> jpaTm =
       Suppliers.memoize(TransactionManagerFactory::createJpaTransactionManager);
 
+  @NonFinalForTesting
+  private static Supplier<JpaTransactionManager> replicaJpaTm =
+      Suppliers.memoize(TransactionManagerFactory::createReplicaJpaTransactionManager);
+
   private static boolean onBeam = false;
 
   private TransactionManagerFactory() {}
@@ -56,6 +60,14 @@ public final class TransactionManagerFactory {
     // by calling setJpaTm().
     if (isInAppEngine()) {
       return DaggerPersistenceComponent.create().appEngineJpaTransactionManager();
+    } else {
+      return DummyJpaTransactionManager.create();
+    }
+  }
+
+  private static JpaTransactionManager createReplicaJpaTransactionManager() {
+    if (isInAppEngine()) {
+      return DaggerPersistenceComponent.create().readOnlyReplicaJpaTransactionManager();
     } else {
       return DummyJpaTransactionManager.create();
     }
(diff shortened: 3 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuRqTZCB4lny5VrpwB1:b-8jbvwi) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*
> ```Java
>         @Override
>         public EppResource load(VKey<? extends EppResource> key) {
>           TransactionManager tm = tm().isOfy() ? tm() : replicaJpaTm();
> ```

This pattern is kind of clunky and is going to appear everywhere.

I think we need a new version of `tm()` that just always does this.  So `tm()` will be used for the main DB and then something else will be used for the replica.

What should the something else be ... how about `rtm()`, `roTm()`, or similar?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuS8t3u4L_SDu9kq0xA:b-2afbnl) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

This pattern is kind of clunky and is going to appear everywhere.

I think we need a new version of `tm()` that just always does this.  So `tm()` will be used for the main DB and then something else will be used for the replica.

What should the something else be ... how about `rtm()`, `roTm()`, or similar?
</blockquote></details>

Yeah I think that's probably a good idea. At the risk of being a bit more verbose, I think it's clearest to use `readOnlyTm()` and `readOnlyJpaTm()`

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuS9778-WZ2V75oXRaX:b-4ktjwo) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

Yeah I think that's probably a good idea. At the risk of being a bit more verbose, I think it's clearest to use `readOnlyTm()` and `readOnlyJpaTm()` 
</blockquote></details>

That might be overly verbose, considering how frequently it's gonna be used everywhere.  Note that we're already doing `tm()` instead of `transactionManager()` for exactly that reason.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuS9ICU5pWrP2gjiFpx:bv7c3f) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

That might be overly verbose, considering how frequently it's gonna be used everywhere.  Note that we're already doing `tm()` instead of `transactionManager()` for exactly that reason.
</blockquote></details>

`repTm()`?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuS9KCTCcbVxvl51W8h:bwel0lx) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

`repTm()`?
</blockquote></details>

How widely is `ro` understood as an abbreviation for "readonly"?  Anyone who's ever used `mount` or similar commands in Linux understands `ro`.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuS9emaAaXwTgAzA4Fs:b-r4fk4o) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

How widely is `ro` understood as an abbreviation for "readonly"?  Anyone who's ever used `mount` or similar commands in Linux understands `ro`.
</blockquote></details>

eh fair I suppose we can use `roTm()` and `roJpaTm()`. 

I'd prefer to use `read-only` rather than `replica` as the semantic just in case we change later on how we manage the transaction managers / backing databases. I think the former will be clearer to users of the code.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuWgUFvB9Fe6UBbZUui:b-9s5h8x) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

eh fair I suppose we can use `roTm()` and `roJpaTm()`. 

I'd prefer to use `read-only` rather than `replica` as the semantic just in case we change later on how we manage the transaction managers / backing databases. I think the former will be clearer to users of the code. 
</blockquote></details>

I think using something that means `replica` actually makes the intent more clear, though.  The replica isn't just read-only; it also potentially lags behind by a little bit.  Indeed, we do currently have a read-only transaction manager for use during the migration that does hit the _primary_ database (not the replica).  So I'm not convinced that "readonly" is even the right nomenclature here; the key concept to convey really is that it's the replica.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuWh5Rc3PiqIOxZpac2:-MuWh5Rc3PiqIOxZpac3:bdz6t20) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*
> ```Java
>         @Override
>         public EppResource load(VKey<? extends EppResource> key) {
>           TransactionManager tm = tm().isOfy() ? tm() : replicaJpaTm();
> ```

This logic of checking whether `tm().isOfy()` should be inside the `repTm()` method that I'm envisioning.  Basically, you're saying "I want to use the replica transaction manager if available", so in Cloud SQL you get it, and then in Datastore you get the normal one since it's not supported there.  Then, very soon, when we're done with the migration, you can then simply remove this `tm().isOfy()` checking once (in the implementation of `repTm()`, rather than having to remove it from all these callsites it's being added to.
___
*[core/src/main/java/google/registry/model/EppResource.java, line 386 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuWgoxr-6v5dU8Sveaw:-MuWgoxr-6v5dU8Sveax:bmar86d) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L386)):*
> ```Java
>         public EppResource load(VKey<? extends EppResource> key) {
>           TransactionManager tm = tm().isOfy() ? tm() : replicaJpaTm();
>           return tm.doTransactionless(() -> tm.loadByKey(key));
> ```

What I was expecting was something like this:

```
return repTm().doTransactionless(() -> repTm.loadByKey(key));
```

The new thing should be as easy to use inline as `tm()` itself is.  There shouldn't be a need to pull it out as an inline field.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 225 at r3](https://reviewable.io/reviews/google/nomulus/1503#-MuWhZqzBVSIU8M-IOOw:-MuWhZr-6WiM1xY-wXSI:btgawss) ([raw file](https://github.com/google/nomulus/blob/64515f06e2e78d49b570a00c7b248d088b933789/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L225)):*
> ```Java
>           Collection<String> foreignKeys,
>           boolean inTransaction,
>           boolean useReadOnlyJpaTm) {
> ```

Per above, I think this is misleading -- what's being used is the replica DB instance, not a read-only transaction manager.  We do have a read-only transaction manager, but that's not what's being used here.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuRqTZCB4lny5VrpwB0:-MuWt4bK7fX7slpouCHC:b4t6xo6) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

I think using something that means `replica` actually makes the intent more clear, though.  The replica isn't just read-only; it also potentially lags behind by a little bit.  Indeed, we do currently have a read-only transaction manager for use during the migration that does hit the _primary_ database (not the replica).  So I'm not convinced that "readonly" is even the right nomenclature here; the key concept to convey really is that it's the replica.
</blockquote></details>

well then we should use that nomenclature everywhere (which I've done here), to make it clear that we're accessing the replica. 
___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuWh5Rc3PiqIOxZpac2:-MuWtFd42Rrqvwm5mj4x:b-4nliqy) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

This logic of checking whether `tm().isOfy()` should be inside the `repTm()` method that I'm envisioning.  Basically, you're saying "I want to use the replica transaction manager if available", so in Cloud SQL you get it, and then in Datastore you get the normal one since it's not supported there.  Then, very soon, when we're done with the migration, you can then simply remove this `tm().isOfy()` checking once (in the implementation of `repTm()`, rather than having to remove it from all these callsites it's being added to.
</blockquote></details>

Again that's was was already being done here, I think you're looking at an old version of the code. 
___
*[core/src/main/java/google/registry/model/EppResource.java, line 386 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuWgoxr-6v5dU8Sveaw:-MuWtB9A7hiqiGexRUfh:b-rofc3n) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L386)):*

What I was expecting was something like this:

```
return repTm().doTransactionless(() -> repTm.loadByKey(key));
```

The new thing should be as easy to use inline as `tm()` itself is.  There shouldn't be a need to pull it out as an inline field.
</blockquote></details>

That's what's already being done here (even before this revision); I think you may be looking at an old version.
___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 225 at r3](https://reviewable.io/reviews/google/nomulus/1503#-MuWhZqzBVSIU8M-IOOw:-MuWtTlG9OJL7Zh8-TjK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/64515f06e2e78d49b570a00c7b248d088b933789/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L225)):*

Per above, I think this is misleading -- what's being used is the replica DB instance, not a read-only transaction manager.  We do have a read-only transaction manager, but that's not what's being used here.
</blockquote></details>

Done.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/EppResource.java, line 385 at r1](https://reviewable.io/reviews/google/nomulus/1503#-MuWh5Rc3PiqIOxZpac2:-MuWu2b4BXPDlcKsSaSC:b-uhl8z2) ([raw file](https://github.com/google/nomulus/blob/5f9c8721d5f217b1b9404d17dce1c68932bd4494/core/src/main/java/google/registry/model/EppResource.java#L385)):*

Again that's was was already being done here, I think you're looking at an old version of the code. 
</blockquote></details>

*what was

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 234 at r4](https://reviewable.io/reviews/google/nomulus/1503#-MuX2vbn5jKHaNRslBwY:-MuX2vbn5jKHaNRslBwZ:b-ot5xn3) ([raw file](https://github.com/google/nomulus/blob/16c72a81ab2b408f14d48256a90d1c81ab2c2228/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L234)):*
> ```Java
>     } else {
>       String property = RESOURCE_CLASS_TO_FKI_PROPERTY.get(clazz);
>       JpaTransactionManager jpaTmToUse = useReplicaJpaTm ? replicaJpaTm() : jpaTm();
> ```

Are you sure this is correct?  Currently it uses `tm()`, which could fall back to `ofyTm()` if we're still in that mode (as we are in production currently).  But with these added changes, if `useReplicaJpaTm` is set to `false`, then it can't fall back to `ofyTm()`?
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 135 at r4](https://reviewable.io/reviews/google/nomulus/1503#-MuX3Bx27bMLI7RdDm4Q:-MuX3Bx3DWHiD1mT2hQD:bmalzof) ([raw file](https://github.com/google/nomulus/blob/16c72a81ab2b408f14d48256a90d1c81ab2c2228/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L135)):*
> ```Java
>    */
>   public static TransactionManager replicaTm() {
>     TransactionManager tm = tm();
> ```

You can inline this.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/index/ForeignKeyIndex.java, line 234 at r4](https://reviewable.io/reviews/google/nomulus/1503#-MuX2vbn5jKHaNRslBwY:-MuXUn1WEqgoJSxyXUNF:b-f9p1c1) ([raw file](https://github.com/google/nomulus/blob/16c72a81ab2b408f14d48256a90d1c81ab2c2228/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L234)):*

Are you sure this is correct?  Currently it uses `tm()`, which could fall back to `ofyTm()` if we're still in that mode (as we are in production currently).  But with these added changes, if `useReplicaJpaTm` is set to `false`, then it can't fall back to `ofyTm()`?
</blockquote></details>

Check out the surrounding code; this is an `else` of a `if (tm().isOfy())`
___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 135 at r4](https://reviewable.io/reviews/google/nomulus/1503#-MuX3Bx27bMLI7RdDm4Q:-MuXV2fR58bXssL9-_kN:b-eayd1c) ([raw file](https://github.com/google/nomulus/blob/16c72a81ab2b408f14d48256a90d1c81ab2c2228/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L135)):*

You can inline this.
</blockquote></details>

i guess one extra method call isn't the end of the world? meh

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 156 at r5](https://reviewable.io/reviews/google/nomulus/1503#-MuXVf_Y7QnMCrrF6-R_:-MuXVf_Y7QnMCrrF6-Ra:b-cc8l4a) ([raw file](https://github.com/google/nomulus/blob/3f2bc3e02b5328484b44b3308393592399de46c1/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L156)):*
> ```Java
>   /** Sets the value of {@link #replicaJpaTm()} to the given {@link JpaTransactionManager}. */
>   public static void setReplicaJpaTm(Supplier<JpaTransactionManager> replicaJpaTmSupplier) {
>     checkNotNull(replicaJpaTmSupplier, "replicaJpaTmSupplier");
> ```

Use `checkArgumentNotNull()`.  It throws a more useful exception type.  NPE typically indicates an unintentionally thrown exception.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java, line 156 at r5](https://reviewable.io/reviews/google/nomulus/1503#-MuXVf_Y7QnMCrrF6-R_:-MuXYKsG2zsqDX9J3Jo1:bsuoe05) ([raw file](https://github.com/google/nomulus/blob/3f2bc3e02b5328484b44b3308393592399de46c1/core/src/main/java/google/registry/persistence/transaction/TransactionManagerFactory.java#L156)):*

Use `checkArgumentNotNull()`.  It throws a more useful exception type.  NPE typically indicates an unintentionally thrown exception.
</blockquote></details>

fine, done in all places

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1503#-:-MuXZk3D3_VJVqhrU9Hp:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1509, opened 2022-02-02
Count duplicates when comparing Databases

weiminyu:
Cursors may have duplicates in Datastore if imported across projects.
Count them instead of throwing.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1509)

### DIFF
--- a/core/src/main/java/google/registry/beam/comparedb/ValidateSqlUtils.java
+++ b/core/src/main/java/google/registry/beam/comparedb/ValidateSqlUtils.java
@@ -104,6 +104,7 @@ final class ValidateSqlUtils {
     private final HashMap<String, Counter> missingCounters = new HashMap<>();
     private final HashMap<String, Counter> unequalCounters = new HashMap<>();
     private final HashMap<String, Counter> badEntityCounters = new HashMap<>();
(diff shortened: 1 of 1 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1509)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/beam/comparedb/ValidateSqlUtils.java, line 174 at r1](https://reviewable.io/reviews/google/nomulus/1509#-Muz_2cW0BxBXQPxYLJF:-Muz_2cW0BxBXQPxYLJG:b-7wpp7k) ([raw file](https://github.com/google/nomulus/blob/035f61480391b6ca5a15636f8f6efc9832d581cb/core/src/main/java/google/registry/beam/comparedb/ValidateSqlUtils.java#L174)):*
> ```Java
>         duplicateEntityCounters.get(counterKey).inc();
>         return;
>       }
> ```

Is it worth noting when the entity size > 3?  (Either with a check fail or a log message)  We currently expect 2 datastore entities under some circumstances, but 3 would indicate a different problem.

PR #1514, opened 2022-02-03
Fix flaky RdeStagingActionDatastoreTest

weiminyu:
Fixed the most common cause that makes one method flaky (Clock and
timestamp problem). Added a TODO to rethink test case.

Also added notes on tasks potentially enqueued multiple times.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1514)

### DIFF
--- a/core/src/main/java/google/registry/beam/rde/RdeIO.java
+++ b/core/src/main/java/google/registry/beam/rde/RdeIO.java
@@ -297,6 +297,10 @@ public class RdeIO {
                 logger.atInfo().log(
                     "Rolled forward %s on %s cursor to %s.", key.cursor(), key.tld(), newPosition);
                 RdeRevision.saveRevision(key.tld(), key.watermark(), key.mode(), revision);
+                // Enqueueing a task is a side effect that is not undone if the transaction rolls
+                // back. So this may result in multiple copies of the same task being processed.
+                // This is fine because the RdeUploadAction is guarded by a lock and tracks progress
+                // by cursor. The BrdaCopyAction writes a file to GCS, which is an atomic action.
                 if (key.mode() == RdeMode.FULL) {
                   cloudTasksUtils.enqueue(
                       RDE_UPLOAD_QUEUE,
--- a/core/src/main/java/google/registry/rde/BrdaCopyAction.java
+++ b/core/src/main/java/google/registry/rde/BrdaCopyAction.java
@@ -87,6 +87,8 @@ public final class BrdaCopyAction implements Runnable {
   }
 
   private void copyAsRyde() throws IOException {
+    // TODO(b/217772483): consider guarding this action with a lock and check if there is work.
+    // Not urgent since file writes on GCS are atomic.
     int revision =
         RdeRevision.getCurrentRevision(tld, watermark, THIN)
             .orElseThrow(
--- a/core/src/main/java/google/registry/rde/RdeStagingReducer.java
+++ b/core/src/main/java/google/registry/rde/RdeStagingReducer.java
@@ -226,6 +226,10 @@ public final class RdeStagingReducer extends Reducer<PendingDeposit, DepositFrag
               logger.atInfo().log(
                   "Rolled forward %s on %s cursor to %s.", key.cursor(), tld, newPosition);
               RdeRevision.saveRevision(tld, watermark, mode, revision);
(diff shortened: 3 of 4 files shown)


rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1514)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1514#-Mv0uIaR3K42XjhdeDG0:-Mv0uIaR3K42XjhdeDG1:bi5b3yu) ([raw file](https://github.com/google/nomulus/blob/1cc18e5bcada0734f85ecf5de036fcf805da2756/core/src/main/java/google/registry/beam/rde/RdeIO.java#L300)):*
> ```Java
>                     "Rolled forward %s on %s cursor to %s.", key.cursor(), key.tld(), newPosition);
>                 RdeRevision.saveRevision(key.tld(), key.watermark(), key.mode(), revision);
>                 // Enqueueing a task is a side effect that is not undone if the transaction rolls
> ```

I'm a bit confused by the wording here. If the transaction is rolled back, the task that's already enqueued as part of the transaction cannot be removed from the task queue. Is my understanding correct?

jianglai (approved):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1514)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai, @rachelguan, and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 300 at r1](https://reviewable.io/reviews/google/nomulus/1514#-Mv0uIaR3K42XjhdeDG0:-Mv1eM10Av5ukxeUOalu:bhweyyp) ([raw file](https://github.com/google/nomulus/blob/1cc18e5bcada0734f85ecf5de036fcf805da2756/core/src/main/java/google/registry/beam/rde/RdeIO.java#L300)):*

I'm a bit confused by the wording here. If the transaction is rolled back, the task that's already enqueued as part of the transaction cannot be removed from the task queue. Is my understanding correct? 
</blockquote></details>

Before the migration, the Task Queue API is datastore transaction-aware (because both datastore and task queue are part of the GAE environment), meaning that if the transaction is cancelled, the enqueueing is also cancelled.

Now with Cloud Tasks we don't have that guarantee. However as Weimin stated it's OK even if the task ended up running, because the resulting action checks the cursor before uploading. When the transaction is cancelled, the cursor is rolled back, so the action would not run.
___
*[core/src/test/java/google/registry/rde/RdeStagingActionDatastoreTest.java, line 476 at r1](https://reviewable.io/reviews/google/nomulus/1514#-Mv1ffXi-2Lx3aKWz5pf:-Mv1ffXi-2Lx3aKWz5pg:b-s1gsl1) ([raw file](https://github.com/google/nomulus/blob/1cc18e5bcada0734f85ecf5de036fcf805da2756/core/src/test/java/google/registry/rde/RdeStagingActionDatastoreTest.java#L476)):*
> ```Java
>     action.run();
>     executeTasksUntilEmpty("mapreduce", clock);
>     // TODO(b/217773051): duplicate tasks are possible though unlikely. Consider if below calls are
> ```

Are duplicates still possible in this case when the clock auto advances?



_Code quote:_
```Java
duplicate tasks are possible
```

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1514)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[core/src/test/java/google/registry/rde/RdeStagingActionDatastoreTest.java, line 476 at r1](https://reviewable.io/reviews/google/nomulus/1514#-Mv1ffXi-2Lx3aKWz5pf:-Mv4Nh720HTwXy9Ujqo0:bwre5u6) ([raw file](https://github.com/google/nomulus/blob/1cc18e5bcada0734f85ecf5de036fcf805da2756/core/src/test/java/google/registry/rde/RdeStagingActionDatastoreTest.java#L476)):*

Are duplicates still possible in this case when the clock auto advances?
</blockquote></details>

Theoretically possible, if the reduce transaction retries for some reason.

PR #1515, opened 2022-02-04
Fix create/update timestamp replay problems

mindhog:
When CreateAutoTimestamp and UpdateAutoTimestamp are inserted into a
Transaction, their values are not populated in the same way as when they are
stored in the course of an SQL commit.  This results in different timestamp
values between SQL and datastore during the SQL -> DS replay.

Fix this by providing these values from the JPA transaction time when we're
doing transaction serialization.

This change also removes the initialization of the Ofy clock in
ExpandRecurringBillingEventsActionTest.  It's not necessary as the
ReplayExtension already takes care of this and doing it after the
ReplayExtension as we were breaks a test now that the update timestamps are
correct.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1515)

### DIFF
--- a/core/src/main/java/google/registry/model/translators/CreateAutoTimestampTranslatorFactory.java
+++ b/core/src/main/java/google/registry/model/translators/CreateAutoTimestampTranslatorFactory.java
@@ -14,12 +14,10 @@
 
 package google.registry.model.translators;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-import static google.registry.persistence.transaction.TransactionManagerFactory.ofyTm;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static org.joda.time.DateTimeZone.UTC;
 
 import google.registry.model.CreateAutoTimestamp;
-import google.registry.persistence.transaction.Transaction;
 import java.util.Date;
 import org.joda.time.DateTime;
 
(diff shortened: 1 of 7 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1515)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @mindhog and @weiminyu)

___
*[core/src/test/java/google/registry/model/domain/DomainBaseSqlTest.java, line 676 at r1](https://reviewable.io/reviews/google/nomulus/1515#-MvKxvKEEu4q1B3EHeF3:-MvKxvKF2ixvMM8rvNYu:by25w7o) ([raw file](https://github.com/google/nomulus/blob/d6c3264904d2547e1ba920090278ac87cc1de7d3/core/src/test/java/google/registry/model/domain/DomainBaseSqlTest.java#L676)):*
> ```Java
> 
>   // @TestSqlOnly
>   // void testUpdateTimeAfterNameserverUpdate() {
> ```

Is this method meant to be kept or discarded?



_Code quote:_
```Java
  // @TestSqlOnly
  // void testUpdateTimeAfterNameserverUpdate() {
```

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1515)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/test/java/google/registry/model/domain/DomainBaseSqlTest.java, line 676 at r1](https://reviewable.io/reviews/google/nomulus/1515#-MvKxvKEEu4q1B3EHeF3:-MvL8iId-czY6fhoxofX:b-ed3vch) ([raw file](https://github.com/google/nomulus/blob/d6c3264904d2547e1ba920090278ac87cc1de7d3/core/src/test/java/google/registry/model/domain/DomainBaseSqlTest.java#L676)):*

Is this method meant to be kept or discarded?
</blockquote></details>

Sorry, I meant to exclude that from the check-in.  Removed.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1515)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1517, opened 2022-02-08
Fix update timestamps for DomainContent types

mindhog:
We expect update timestamps to be updated whenever a containing entity is
modified and persisted, but unfortunately Hibernate doesn't seem to do this --
instead it appears to regard such an entity as unchanged.

To work around this, we explicitly reset the update timestamp whenever a
nested collection is modified in the Builder.

Note that this change only solves the problem for DomainContent.  All other
entitities containing UpdateAutoTimestamp will need to be audited and
instrumented with a similar change.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1517)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1517)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1519, opened 2022-02-10
Inject CloudTasksUtils to DomainLockUtils

rachelguan:
The main mission of Task Queue Migration is to convert from ```Task Queue``` to ```Cloud Task```. The first phase is to convert all the push queue calls APIs. There's a mix of push and pull queue methods in ```AsyncTaskEnqueuer```. This purpose of this PR is to convert ```enqueueDomainRelock``` from using ```Task Queue``` to ```Cloud Task```. 

The method ```enqueueDomainRelock```is used in ```RelockDomainAction```, ```RegistryLockVerifyAction```, and ```UnlockDomainCommand```, and all three files have an instance of ```DomainLockUtils``` injected, which makes ```DomainLockUtils``` the better place to store this method. 
 
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1519)

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1519)

### DIFF
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -162,20 +162,16 @@ public final class AsyncTaskEnqueuer {
         lock.getRelockDuration().isPresent(),
         "Lock with ID %s not configured for relock",
         lock.getRevisionId());
-    enqueueDomainRelock(lock.getRelockDuration().get(), lock.getRevisionId(), 0);
-  }
-
-  /** Enqueues a task to asynchronously re-lock a registry-locked domain after it was unlocked. */
-  void enqueueDomainRelock(Duration countdown, long lockRevisionId, int previousAttempts) {
-    String backendHostname = appEngineServiceUtils.getServiceHostname("backend");
     addTaskToQueueWithRetry(
         asyncActionsPushQueue,
         TaskOptions.Builder.withUrl(RelockDomainAction.PATH)
             .method(Method.POST)
-            .header("Host", backendHostname)
-            .param(RelockDomainAction.OLD_UNLOCK_REVISION_ID_PARAM, String.valueOf(lockRevisionId))
-            .param(RelockDomainAction.PREVIOUS_ATTEMPTS_PARAM, String.valueOf(previousAttempts))
-            .countdownMillis(countdown.getMillis()));
+            .header("Host", appEngineServiceUtils.getServiceHostname("backend"))
+            .param(
+                RelockDomainAction.OLD_UNLOCK_REVISION_ID_PARAM,
+                String.valueOf(lock.getRevisionId()))
+            .param(RelockDomainAction.PREVIOUS_ATTEMPTS_PARAM, String.valueOf(0))
+            .countdownMillis(lock.getRelockDuration().get().getMillis()));
   }
 
   /**
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -26,7 +26,6 @@ import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.EppResource;
-import google.registry.model.domain.RegistryLock;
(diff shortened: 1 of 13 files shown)


rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1519)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/tools/DomainLockUtils.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1519#-MwFB27a3dJSJ5psRxYG:-MwFI_jW5OYAPiZROrk4:bll842w) ([raw file](https://github.com/google/nomulus/blob/3f10791494e6025be248ce40dd0b1f38f46a54c7/core/src/main/java/google/registry/tools/DomainLockUtils.java#L61)):*

Why not just inject a `CloudTasksUtils` here? Then you don't have to inject the clock (remember the clock is now part of it after #1521. The methods also don't need to take a `CloudTasksUtils` as an argument.
</blockquote></details>

I did not inject ```CloudTasksUtils`` here because
 1) I was thinking to have ```CloudTasksUtils``` in action file like other action files,  for code consistency 
2) Logically, it didn't seem like a right decision to have a utility class injected in anther utility class since util classes should be on the independent from each other. But for this use case, it does seem ok 
___
*[core/src/test/java/google/registry/batch/RelockDomainActionTest.java, line 270 at r1](https://reviewable.io/reviews/google/nomulus/1519#-MwFCv1e6epKdz5SFetQ:-MwFKTPzBXoS4IXMgyOp:bskj0sa) ([raw file](https://github.com/google/nomulus/blob/3f10791494e6025be248ce40dd0b1f38f46a54c7/core/src/test/java/google/registry/batch/RelockDomainActionTest.java#L270)):*

Shouldn't this test be in `DomainLockUtilsTest` as well?
</blockquote></details>

yes this one was supposed to be in ```DomainLockUtilTest``` and I think the existing test cases in this file cover all the cases for enqueuing relock.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1519)* status: 2 of 13 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/tools/DomainLockUtils.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1519#-MwFB27a3dJSJ5psRxYG:-Mwhdt_K4l-6jMy1grFa:bqny4hr) ([raw file](https://github.com/google/nomulus/blob/3f10791494e6025be248ce40dd0b1f38f46a54c7/core/src/main/java/google/registry/tools/DomainLockUtils.java#L61)):*

I did not inject ```CloudTasksUtils`` here because
 1) I was thinking to have ```CloudTasksUtils``` in action file like other action files,  for code consistency 
2) Logically, it didn't seem like a right decision to have a utility class injected in anther utility class since util classes should be on the independent from each other. But for this use case, it does seem ok 
</blockquote></details>

Modified the previous solution and now an instance of ```CloudTasksUtils``` is injected to ```DomainLockUtils```. It seems neater  since it avoids ```CloudTasksUtils``` being injected in multiple action files.

jianglai (approved):
> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1519)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/tools/DomainLockUtils.java, line 61 at r1](https://reviewable.io/reviews/google/nomulus/1519#-MwFB27a3dJSJ5psRxYG:-Mwj2YqpCbZWaKzPA6qv:b-2teahn) ([raw file](https://github.com/google/nomulus/blob/3f10791494e6025be248ce40dd0b1f38f46a54c7/core/src/main/java/google/registry/tools/DomainLockUtils.java#L61)):*

Modified the previous solution and now an instance of ```CloudTasksUtils``` is injected to ```DomainLockUtils```. It seems neater  since it avoids ```CloudTasksUtils``` being injected in multiple action files.  
</blockquote></details>

Technically there's nothing inherently wrong with injecting a util class into another. We have nested injections all over the place and utils are nothing special. In fact the beauty of using a dependency injection framework is that you don't have to worry about all these. You just request the thing that you immediately need in your class and the framework will make sure that the correct ones are injected, and the scopes and life cycles of the injected objects are correct.

PR #1520, opened 2022-02-10
Remove enqueueAsyncResave from AsyncTasksEnqueuer 

rachelguan:
For Task Queue Migration, ```AsyncTasksEnqueuer``` is the last file needed to be modified to complete Task Queue Migration for push queue. This case is trickier than the previous ones since the ```enqueueAsyncResave()``` is being used in ```Flow```classes. The existing set up for test classes involves more than just remove ```AsyncTasksEnqueuer``` and add ```CloudTasksUtil``` to enqueue.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1520)

### DIFF
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -15,14 +15,11 @@
 package google.registry.batch;
 
 import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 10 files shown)


rachelguan:
Because of the existing set up for flow test cases, it's difficult to simply just remove AsyncTasksEnqueuer for some flows and not the others. One of the easier ways to handle this is to inject CloudTasksUtil to AsyncTasksEnqueuer.

PR #1521, opened 2022-02-11
Make a few quality-of-life improvements in CloudTasksUtils

jianglai:
1. Update the method names. There are too many overloaded methods and it
     is hard to figure out which one does which without checking the
     javadoc.

  2. Added a method in the task matcher to specify the delay time in
     DateTime, so the caller does not need to convert it to Timestamp.

  3. Remove the explicit dependency on a clock when enqueueing a task with
     delay, the clock is now injected directly into the util instance
     itself.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1521)

### DIFF
--- a/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java
+++ b/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java
@@ -96,7 +96,7 @@ public final class CommitLogCheckpointAction implements Runnable {
                   // Enqueue a diff task between previous and current checkpoints.
                   cloudTasksUtils.enqueue(
                       QUEUE_NAME,
-                      CloudTasksUtils.createPostTask(
+                      cloudTasksUtils.createPostTask(
                           ExportCommitLogDiffAction.PATH,
                           Service.BACKEND.toString(),
                           ImmutableMultimap.of(
--- a/core/src/main/java/google/registry/beam/rde/RdeIO.java
+++ b/core/src/main/java/google/registry/beam/rde/RdeIO.java
@@ -304,7 +304,7 @@ public class RdeIO {
                 if (key.mode() == RdeMode.FULL) {
                   cloudTasksUtils.enqueue(
                       RDE_UPLOAD_QUEUE,
-                      CloudTasksUtils.createPostTask(
+                      cloudTasksUtils.createPostTask(
                           RdeUploadAction.PATH,
                           Service.BACKEND.getServiceId(),
                           ImmutableMultimap.of(
@@ -315,7 +315,7 @@ public class RdeIO {
                 } else {
                   cloudTasksUtils.enqueue(
                       BRDA_QUEUE,
-                      CloudTasksUtils.createPostTask(
+                      cloudTasksUtils.createPostTask(
                           BrdaCopyAction.PATH,
                           Service.BACKEND.getServiceId(),
                           ImmutableMultimap.of(
--- a/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java
+++ b/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java
@@ -21,6 +21,7 @@ import dagger.Module;
 import dagger.Provides;
 import google.registry.config.CredentialModule.DefaultCredential;
 import google.registry.config.RegistryConfig.Config;
+import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.CloudTasksUtils.GcpCloudTasksClient;
 import google.registry.util.CloudTasksUtils.SerializableCloudTasksClient;
@@ -46,8 +47,9 @@ public abstract class CloudTasksUtilsModule {
       @Config("projectId") String projectId,
       @Config("locationId") String locationId,
       SerializableCloudTasksClient client,
-      Retrier retrier) {
-    return new CloudTasksUtils(retrier, projectId, locationId, client);
+      Retrier retrier,
+      Clock clock) {
+    return new CloudTasksUtils(retrier, clock, projectId, locationId, client);
   }
 
   // Provides a supplier instead of using a Dagger @Provider because the latter is not serializable.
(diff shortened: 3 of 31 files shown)


rachelguan (changes requested):
> Reviewed 31 of 31 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1521)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 311 at r3](https://reviewable.io/reviews/google/nomulus/1521#-MvuEuJc8IZIj7Cgvjxv:-MvuEuJd95Hk0s-A4V1g:bxqyj4u) ([raw file](https://github.com/google/nomulus/blob/243c7a4d7d711da954f1fd6b901d9e243bd7ca73/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L311)):*
> ```Java
>     }
> 
>     public TaskMatcher scheduleTime(Timestamp scheduleTime) {
> ```

Since ```DateTime``` is now introduced to do time comparison, is it necessary to keep ```Timestamp``` schedule time method then?
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 227 at r3](https://reviewable.io/reviews/google/nomulus/1521#-MvuFN1c8OxmgVnX-vVa:-MvuFN1d1yo7QL80X5Z5:b-b0s2vn) ([raw file](https://github.com/google/nomulus/blob/243c7a4d7d711da954f1fd6b901d9e243bd7ca73/util/src/main/java/google/registry/util/CloudTasksUtils.java#L227)):*
> ```Java
>   public Task createPostTaskWithJitter(
>       String path,
>       String service,
> ```

I would like to know the reason behind removing static. I think that It's actually good that an instance of ```CloudTasksUtils``` is not required to create a task. A scenario could be, let's say you want to just create a Task object to do Task comparison.. without enqueueing or deleting a task from the queue.

jianglai (commented):
> Reviewed 31 of 31 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1521)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/testing/CloudTasksHelper.java, line 311 at r3](https://reviewable.io/reviews/google/nomulus/1521#-MvuEuJc8IZIj7Cgvjxv:-Mvy1Qhi1FMcuPoEykOZ:baf3ihw) ([raw file](https://github.com/google/nomulus/blob/243c7a4d7d711da954f1fd6b901d9e243bd7ca73/core/src/test/java/google/registry/testing/CloudTasksHelper.java#L311)):*

Since ```DateTime``` is now introduced to do time comparison, is it necessary to keep ```Timestamp``` schedule time method then?
</blockquote></details>

It doesn't hurt to keep it around. In general it make sense to have at least the setter that takes the exact same type and optionally there can be more setters that does implicit conversion of the arguments.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 227 at r3](https://reviewable.io/reviews/google/nomulus/1521#-MvuFN1c8OxmgVnX-vVa:-Mvy12ZtDlVImFt4jRZu:b-a2rrh9) ([raw file](https://github.com/google/nomulus/blob/243c7a4d7d711da954f1fd6b901d9e243bd7ca73/util/src/main/java/google/registry/util/CloudTasksUtils.java#L227)):*

I would like to know the reason behind removing static. I think that It's actually good that an instance of ```CloudTasksUtils``` is not required to create a task. A scenario could be, let's say you want to just create a Task object to do Task comparison.. without enqueueing or deleting a task from the queue.
</blockquote></details>

It's because all of the methods rely on the injected `clock`. It's conceivable that we can have static methods that create vanilla tasks and then non-static methods that create tasks with delays. But are there actually use cases where we'd create tasks but not compare them?

rachelguan (changes requested):
The PR itself seems fine but there's a test case ```GenerateSpec11ReportActionTest.testSuccess()``` failing. Once that's fixed, this PR should be good to go! 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1521#-:-Mvz45HhBM45CawyKTun:bbhf8qh)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 227 at r3](https://reviewable.io/reviews/google/nomulus/1521#-MvuFN1c8OxmgVnX-vVa:-Mvyv2nB-rEw6epddHe8:b7dwt1s) ([raw file](https://github.com/google/nomulus/blob/243c7a4d7d711da954f1fd6b901d9e243bd7ca73/util/src/main/java/google/registry/util/CloudTasksUtils.java#L227)):*

It's because all of the methods rely on the injected `clock`. It's conceivable that we can have static methods that create vanilla tasks and then non-static methods that create tasks with delays. But are there actually use cases where we'd create tasks but not compare them?
</blockquote></details>

I don't think there's any use case that would do that..before using Cloud Task, when a task method whose purpose is to create a certain task without being enqueued. I think it does make sense to have to have ```CloudTaskUtil``` instance to call all the methods in the class.

jianglai (commented):
> Reviewed 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1521)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 227 at r3](https://reviewable.io/reviews/google/nomulus/1521#-MvuFN1c8OxmgVnX-vVa:-MwC9XMH9rtjgf69Leol:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/243c7a4d7d711da954f1fd6b901d9e243bd7ca73/util/src/main/java/google/registry/util/CloudTasksUtils.java#L227)):*

I don't think there's any use case that would do that..before using Cloud Task, when a task method whose purpose is to create a certain task without being enqueued. I think it does make sense to have to have ```CloudTaskUtil``` instance to call all the methods in the class. 
</blockquote></details>

Ack.

jianglai (commented):
Fixed the test. PTAL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1521#-:-MwDhRmRCCDxTQ3bWcNG:b-z3wg1x)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

PR #1522, opened 2022-02-11
Inject CloudTasksUtil to AsyncTaskEnqueuer

rachelguan:
The purpose of this PR is to convert ```enqueueAsyncResave``` from using ```Task Queue``` to ```Cloud Task```. This method is used in both ```Action``` and ```Flow``` files. Because of the way test cases are set up for all the ```Flow``` test cases, it's not easy to Inject ```CloudTasksUtil``` in only flows that require enqueueing. ```Task Queue``` is part of ```App Engine``` feature, and an instance of ```AsyncTaskEnqueuer``` is part of ```FakesAndMocksModule``` which is part of the [test set up](https://github.com/google/nomulus/blob/master/core/src/test/java/google/registry/flows/FlowTestCase.java#L253) to run  all the flows listed in [FlowComponent](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/flows/FlowComponent.java). 

One way to bypass the issue with flow test cases is to inject ```CloudTasksUtil``` to ```AsyncTaskEnqueuer```, which allows the method to stay in the same file, maintaining its reusability. 
 
Because of this code change, ```QUEUE_ASYNC_ACTION```will no longer needed to be part of ```AsyncTaskEnqueuer``` constructor. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1522)

### DIFF
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -62,7 +62,6 @@ public final class AsyncTaskEnqueuer {
   private static final Duration MAX_ASYNC_ETA = Duration.standardDays(30);
 
   private final Duration asyncDeleteDelay;
-  private final Queue asyncActionsPushQueue;
   private final Queue asyncDeletePullQueue;
   private final Queue asyncDnsRefreshPullQueue;
   private final AppEngineServiceUtils appEngineServiceUtils;
@@ -71,15 +70,12 @@ public final class AsyncTaskEnqueuer {
 
   @Inject
   public AsyncTaskEnqueuer(
-      // TODO: (b/220926392) remove QUEUE_AYNSC_ACTIONS after PR#1519 is merged
-      @Named(QUEUE_ASYNC_ACTIONS) Queue asyncActionsPushQueue,
       @Named(QUEUE_ASYNC_DELETE) Queue asyncDeletePullQueue,
       @Named(QUEUE_ASYNC_HOST_RENAME) Queue asyncDnsRefreshPullQueue,
       @Config("asyncDeleteFlowMapreduceDelay") Duration asyncDeleteDelay,
       AppEngineServiceUtils appEngineServiceUtils,
       CloudTasksUtils cloudTasksUtils,
       Retrier retrier) {
-    this.asyncActionsPushQueue = asyncActionsPushQueue;
     this.asyncDeletePullQueue = asyncDeletePullQueue;
     this.asyncDnsRefreshPullQueue = asyncDnsRefreshPullQueue;
     this.asyncDeleteDelay = asyncDeleteDelay;
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -22,14 +22,19 @@ import com.google.appengine.api.taskqueue.TaskOptions;
 import com.google.appengine.api.taskqueue.TaskOptions.Method;
 import com.google.appengine.api.taskqueue.TransientFailureException;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.Multimap;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.EppResource;
 import google.registry.model.eppcommon.Trid;
 import google.registry.model.host.HostResource;
 import google.registry.persistence.VKey;
+import google.registry.request.Action.Service;
 import google.registry.util.AppEngineServiceUtils;
+import google.registry.util.Clock;
+import google.registry.util.CloudTasksUtils;
 import google.registry.util.Retrier;
 import javax.inject.Inject;
 import javax.inject.Named;
@@ -62,6 +67,8 @@ public final class AsyncTaskEnqueuer {
   private final Queue asyncDeletePullQueue;
   private final Queue asyncDnsRefreshPullQueue;
   private final AppEngineServiceUtils appEngineServiceUtils;
+  private CloudTasksUtils cloudTasksUtils;
+  private Clock clock;
   private final Retrier retrier;
 
   @Inject
@@ -71,12 +78,16 @@ public final class AsyncTaskEnqueuer {
       @Named(QUEUE_ASYNC_HOST_RENAME) Queue asyncDnsRefreshPullQueue,
       @Config("asyncDeleteFlowMapreduceDelay") Duration asyncDeleteDelay,
       AppEngineServiceUtils appEngineServiceUtils,
+      CloudTasksUtils cloudTasksUtils,
+      Clock clock,
       Retrier retrier) {
     this.asyncActionsPushQueue = asyncActionsPushQueue;
     this.asyncDeletePullQueue = asyncDeletePullQueue;
     this.asyncDnsRefreshPullQueue = asyncDnsRefreshPullQueue;
     this.asyncDeleteDelay = asyncDeleteDelay;
     this.appEngineServiceUtils = appEngineServiceUtils;
+    this.cloudTasksUtils = cloudTasksUtils;
+    this.clock = clock;
     this.retrier = retrier;
   }
 
@@ -94,7 +105,7 @@ public final class AsyncTaskEnqueuer {
   public void enqueueAsyncResave(
       VKey<?> entityKey, DateTime now, ImmutableSortedSet<DateTime> whenToResave) {
     DateTime firstResave = whenToResave.first();
-    checkArgument(isBeforeOrAt(now, firstResave), "Can't enqueue a resave to run in the past");
+    checkArgument(isBeforeOrAt(now, firstResave), "Can't enqueue a resave " + "to run in the past");
     Duration etaDuration = new Duration(now, firstResave);
     if (etaDuration.isLongerThan(MAX_ASYNC_ETA)) {
       logger.atInfo().log(
@@ -102,19 +113,17 @@ public final class AsyncTaskEnqueuer {
           entityKey, firstResave, MAX_ASYNC_ETA);
       return;
     }
-    logger.atInfo().log("Enqueuing async re-save of %s to run at %s.", entityKey, whenToResave);
-    String backendHostname = appEngineServiceUtils.getServiceHostname("backend");
-    TaskOptions task =
-        TaskOptions.Builder.withUrl(ResaveEntityAction.PATH)
-            .method(Method.POST)
-            .header("Host", backendHostname)
-            .countdownMillis(etaDuration.getMillis())
-            .param(PARAM_RESOURCE_KEY, entityKey.stringify())
-            .param(PARAM_REQUESTED_TIME, now.toString());
+    Multimap<String, String> params = ArrayListMultimap.create();
+    params.put(PARAM_RESOURCE_KEY, entityKey.stringify());
+    params.put(PARAM_REQUESTED_TIME, now.toString());
     if (whenToResave.size() > 1) {
-      task.param(PARAM_RESAVE_TIMES, Joiner.on(',').join(whenToResave.tailSet(firstResave, false)));
+      params.put(PARAM_RESAVE_TIMES, Joiner.on(',').join(whenToResave.tailSet(firstResave, false)));
     }
-    addTaskToQueueWithRetry(asyncActionsPushQueue, task);
+    logger.atInfo().log("Enqueuing async re-save of %s to run at %s.", entityKey, whenToResave);
+    cloudTasksUtils.enqueue(
+        QUEUE_ASYNC_ACTIONS,
+        CloudTasksUtils.createPostTask(
+            ResaveEntityAction.PATH, Service.BACKEND.toString(), params, clock, etaDuration));
   }
 
   /** Enqueues a task to asynchronously delete a contact or host, by key. */
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -33,7 +33,6 @@ import google.registry.model.host.HostResource;
 import google.registry.persistence.VKey;
 import google.registry.request.Action.Service;
 import google.registry.util.AppEngineServiceUtils;
-import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.Retrier;
 import javax.inject.Inject;
@@ -68,18 +67,17 @@ public final class AsyncTaskEnqueuer {
   private final Queue asyncDnsRefreshPullQueue;
   private final AppEngineServiceUtils appEngineServiceUtils;
   private CloudTasksUtils cloudTasksUtils;
-  private Clock clock;
   private final Retrier retrier;
 
   @Inject
   public AsyncTaskEnqueuer(
+      // TODO: (b/220926392) remove QUEUE_AYNSC_ACTIONS after PR#1519 is merged
       @Named(QUEUE_ASYNC_ACTIONS) Queue asyncActionsPushQueue,
       @Named(QUEUE_ASYNC_DELETE) Queue asyncDeletePullQueue,
       @Named(QUEUE_ASYNC_HOST_RENAME) Queue asyncDnsRefreshPullQueue,
       @Config("asyncDeleteFlowMapreduceDelay") Duration asyncDeleteDelay,
       AppEngineServiceUtils appEngineServiceUtils,
       CloudTasksUtils cloudTasksUtils,
-      Clock clock,
       Retrier retrier) {
     this.asyncActionsPushQueue = asyncActionsPushQueue;
     this.asyncDeletePullQueue = asyncDeletePullQueue;
@@ -87,7 +85,6 @@ public final class AsyncTaskEnqueuer {
     this.asyncDeleteDelay = asyncDeleteDelay;
     this.appEngineServiceUtils = appEngineServiceUtils;
     this.cloudTasksUtils = cloudTasksUtils;
-    this.clock = clock;
     this.retrier = retrier;
   }
 
@@ -105,7 +102,7 @@ public final class AsyncTaskEnqueuer {
   public void enqueueAsyncResave(
       VKey<?> entityKey, DateTime now, ImmutableSortedSet<DateTime> whenToResave) {
     DateTime firstResave = whenToResave.first();
-    checkArgument(isBeforeOrAt(now, firstResave), "Can't enqueue a resave " + "to run in the past");
+    checkArgument(isBeforeOrAt(now, firstResave), "Can't enqueue a resave to run in the past");
     Duration etaDuration = new Duration(now, firstResave);
     if (etaDuration.isLongerThan(MAX_ASYNC_ETA)) {
       logger.atInfo().log(
@@ -122,8 +119,8 @@ public final class AsyncTaskEnqueuer {
     logger.atInfo().log("Enqueuing async re-save of %s to run at %s.", entityKey, whenToResave);
     cloudTasksUtils.enqueue(
         QUEUE_ASYNC_ACTIONS,
-        CloudTasksUtils.createPostTask(
-            ResaveEntityAction.PATH, Service.BACKEND.toString(), params, clock, etaDuration));
+        cloudTasksUtils.createPostTaskWithDelay(
+            ResaveEntityAction.PATH, Service.BACKEND.toString(), params, etaDuration));
   }
 
   /** Enqueues a task to asynchronously delete a contact or host, by key. */
--- a/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
+++ b/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java
@@ -32,7 +32,6 @@ import google.registry.model.eppcommon.Trid;
 import google.registry.model.host.HostResource;
 import google.registry.persistence.VKey;
 import google.registry.request.Action.Service;
-import google.registry.util.AppEngineServiceUtils;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.Retrier;
 import javax.inject.Inject;
@@ -64,7 +63,6 @@ public final class AsyncTaskEnqueuer {
   private final Duration asyncDeleteDelay;
   private final Queue asyncDeletePullQueue;
   private final Queue asyncDnsRefreshPullQueue;
-  private final AppEngineServiceUtils appEngineServiceUtils;
   private CloudTasksUtils cloudTasksUtils;
   private final Retrier retrier;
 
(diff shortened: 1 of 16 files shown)


jianglai (commented):
I don't understand why are we only changing `enqueueAsyncResave`, but not the other methods like `enqueueAsyncDelete`? Also please rebase to #1521 which should make a lot of the conversion less tedious.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522#-:-MwFF97C5oVofbppmwgJ:b-wjmggl)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (changes requested):
Duh, of course. Rebase is still needed, though.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522#-:-MwP7OYaC4jDP8N5IX4q:biwcxe0)* status: all files reviewed, 8 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPA0qq9Wum8qz-mFz7:-MwPA0qr31tD_kvvlbqh:bs27cc1) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java#L77)):*
> ```Java
>   @Inject
>   public AsyncTaskEnqueuer(
>       @Named(QUEUE_ASYNC_ACTIONS) Queue asyncActionsPushQueue,
> ```

This field, and its provider method method can be deleted, after https://github.com/google/nomulus/pull/1519 is merged.



_Code quote:_
```Java
asyncActionsPushQueue
```
___
*[core/src/main/java/google/registry/batch/ResaveEntityAction.java, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPB32O2SHsjmK_xlq3:-MwPB32PEIOrWmhZnIid:b-ueinqh) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/ResaveEntityAction.java#L84)):*
> ```Java
>                           : entity);
>             });
>     if (!resaveTimes.isEmpty()) {
> ```

Just because the enqueueing action does not require any information from the transaction doesn't mean that it can be done outside the transaction. Have we verified that it is indeed OK to de-couple them?



_Code quote:_
```Java
if (!resaveTimes.isEmpty()) {
```
___
*[core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java, line 104 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPBcJsDClybqRwA1F_:-MwPBcJsDClybqRwA1Fa:b-67h9or) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java#L104)):*
> ```Java
>             .url(ResaveEntityAction.PATH)
>             .method("POST")
>             .header("Host", "backend.hostname.fake")
> ```

Don't we need to verify that the service is "backend" in the new matcher?



_Code quote:_
```Java
.header("Host", "backend.hostname.fake")
```
___
*[core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPBpgA0h9lctSw-FbY:-MwPBpgA0h9lctSw-FbZ:brhumzt) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java#L126)):*
> ```Java
>             .url(ResaveEntityAction.PATH)
>             .method("POST")
>             .header("Host", "backend.hostname.fake")
> ```

Same here.



_Code quote:_
```Java
.header("Host", "backend.hostname.fake")
```
___
*[core/src/test/java/google/registry/batch/ResaveEntityActionTest.java, line 151 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPDgDo53-j9loWONrA:-MwPDgDo53-j9loWONrB:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/batch/ResaveEntityActionTest.java#L151)):*
> ```Java
>             .url(ResaveEntityAction.PATH)
>             .method("POST")
>             .header("Host", "backend.hostname.fake")
> ```

Ditto.



_Code quote:_
```Java
.header("Host", "backend.hostname.fake")
```
___
*[core/src/test/java/google/registry/flows/EppTestComponent.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPD3yj2GWAPj2eNKtu:-MwPD3yj2GWAPj2eNKtv:b-lx8shz) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/test/java/google/registry/flows/EppTestComponent.java#L68)):*
> ```Java
>     private Sleeper sleeper;
> 
>     public static FakesAndMocksModule create() {
> ```

This class needs some re-thinking now that the cloud tasks utils also contains a clock, we cannot have two fake clocks that works independently -- too much confusion.



_Code quote:_
```Java
 public static FakesAndMocksModule create() {
```
___
*[core/src/test/java/google/registry/flows/domain/DomainDeleteFlowTest.java, line 322 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPDTOa7KelUfXra15A:-MwPDTOa7KelUfXra15B:brhumzt) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/flows/domain/DomainDeleteFlowTest.java#L322)):*
> ```Java
>             .url(ResaveEntityAction.PATH)
>             .method("POST")
>             .header("Host", "backend.hostname.fake")
> ```

Same here.



_Code quote:_
```Java
.header("Host", "backend.hostname.fake")
```
___
*[core/src/test/java/google/registry/flows/domain/DomainTransferRequestFlowTest.java, line 522 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPDY6y-Hz9o83kbMHq:-MwPDY6y-Hz9o83kbMHr:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/flows/domain/DomainTransferRequestFlowTest.java#L522)):*
> ```Java
>             .url(ResaveEntityAction.PATH)
>             .method("POST")
>             .header("Host", "backend.hostname.fake")
> ```

Ditto.



_Code quote:_
```Java
.header("Host", "backend.hostname.fake")
```

rachelguan (commented):
yes yes ,, first thing to do today after that CloudTasksUtil PR went in on friday :) 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522#-:-MwXvYf-2l73ZiFn6b2Y:b-9u7qsy)* status: all files reviewed, 8 unresolved discussions (waiting on @rachelguan)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522)* status: 1 of 14 files reviewed, 8 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPA0qq9Wum8qz-mFz7:-MwYm0Gn5CUnHg7XVn0_:b-vsmd6q) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java#L77)):*

This field, and its provider method method can be deleted, after https://github.com/google/nomulus/pull/1519 is merged.
</blockquote></details>

added a TO-DO along with the bug number to track this. 
___
*[core/src/main/java/google/registry/batch/ResaveEntityAction.java, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPB32O2SHsjmK_xlq3:-MwZa3ZJDTrL5iydtbtb:brqblzj) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/ResaveEntityAction.java#L84)):*

Just because the enqueueing action does not require any information from the transaction doesn't mean that it can be done outside the transaction. Have we verified that it is indeed OK to de-couple them?
</blockquote></details>

I see your point. I took a look at the test cases and there's no existing test cases to make sure that enqueueing shouldn't happen if the transaction fails. I also looked at the log message which is the first line in ```run()``` "Re-saving entity [resourceKey] which was enqueued at [requestedTime] which made me think that it's referring to ```enqueueAsyncResave()```. This log message seemed to have only served for debugging purposes, wording is ambiguous.

After diving deeper into the flow logic and talked with Ben who originally wrote this, it doesn't make a huge difference whether this should be inside or outside of the transaction, as resave is idempotent So this is going to stay the way it is. 
___
*[core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java, line 104 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPBcJsDClybqRwA1F_:-MwZhR5nBdsLcKKLxtfq:b-t948zx) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java#L104)):*

Don't we need to verify that the service is "backend" in the new matcher?
</blockquote></details>

added.
___
*[core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java, line 126 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPBpgA0h9lctSw-FbY:-MwZhSXd77OYbHdCASmA:b-7tfmjp) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/batch/AsyncTaskEnqueuerTest.java#L126)):*

Same here.
</blockquote></details>

added
___
*[core/src/test/java/google/registry/batch/ResaveEntityActionTest.java, line 151 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPDgDo53-j9loWONrA:-MwZhTDl3dcAeyy1DNga:b-7tfmjp) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/batch/ResaveEntityActionTest.java#L151)):*

Ditto.
</blockquote></details>

added
___
*[core/src/test/java/google/registry/flows/EppTestComponent.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPD3yj2GWAPj2eNKtu:-MwZhUPQ0lJb9BGn7GbP:b-ds1sqp) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/test/java/google/registry/flows/EppTestComponent.java#L68)):*

This class needs some re-thinking now that the cloud tasks utils also contains a clock, we cannot have two fake clocks that works independently -- too much confusion.
</blockquote></details>

I took a closer look at the use case of FakesAndMocksModule.create() and it seems like similar problem is actually in the code ..
For example, I checked the code that uses ```public static FakesAndMocksModule create(FakeClock clock, EppMetric.Builder metricBuilder)``` and metricBuilder requires clock as well. I removed the dependency on clock but adding a getter to get the clock of the `CloudTaskHelper` instance. 
___
*[core/src/test/java/google/registry/flows/domain/DomainDeleteFlowTest.java, line 322 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPDTOa7KelUfXra15A:-MwZX2924VX3MVEavBQh:b-7ql1td) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/flows/domain/DomainDeleteFlowTest.java#L322)):*

Same here.
</blockquote></details>

fixed
___
*[core/src/test/java/google/registry/flows/domain/DomainTransferRequestFlowTest.java, line 522 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPDY6y-Hz9o83kbMHq:-MwZX35SBXHxX069kS2C:b-7ql1td) ([raw file](https://github.com/google/nomulus/blob/d3fc6063c997a4a69c9445b08752c7c01020b63c/core/src/test/java/google/registry/flows/domain/DomainTransferRequestFlowTest.java#L522)):*

Ditto.
</blockquote></details>

fixed

jianglai (changes requested):
> Reviewed 14 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522)* status: all files reviewed, 4 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPA0qq9Wum8qz-mFz7:-Mwj43uC8WHxZuS7y5NN:bt0ue3) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java#L77)):*

added a TO-DO along with the bug number to track this. 
</blockquote></details>

If you have a PR that's based on another PR, depending on how much dependency there is, you might actually want to have the dependent PR branch off of the base PR. Note that this is not always recommended because it might become harder later on to rebase correctly.
___
*[core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java, line 75 at r2](https://reviewable.io/reviews/google/nomulus/1522#-Mwj4S1C0_Ee5Zen-RjL:-Mwj4S1C0_Ee5Zen-RjM:bhfa620) ([raw file](https://github.com/google/nomulus/blob/12a9af04c33a3481328bcc40ebc878fb66103b6d/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java#L75)):*
> ```Java
>   @Inject
>   public AsyncTaskEnqueuer(
>       // TODO: (b/220926392) remove QUEUE_AYNSC_ACTIONS after PR#1519 is merged
> ```

Just "#1519" is enough. The numbers are unique in a repo, so you will not mistake a PR with an issue, for example.



_Code quote:_
```Java
PR#1519
```
___
*[core/src/test/java/google/registry/flows/EppTestComponent.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPD3yj2GWAPj2eNKtu:-Mwj810YBacm7U48Ri5x:byof0jl) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/test/java/google/registry/flows/EppTestComponent.java#L68)):*

I took a closer look at the use case of FakesAndMocksModule.create() and it seems like similar problem is actually in the code ..
For example, I checked the code that uses ```public static FakesAndMocksModule create(FakeClock clock, EppMetric.Builder metricBuilder)``` and metricBuilder requires clock as well. I removed the dependency on clock but adding a getter to get the clock of the `CloudTaskHelper` instance. 
</blockquote></details>

If you check all the call sites for the three `create` methods you'll see that all of them are actually using the same clocks (I glanced over them, so I might be wrong), so there's no reason to have all these static methods, which only adds confusion. It looks like all you need is a `create(clock)` public method, which creates the necessary intermediate objects based on the clock and calls the `create(Clock, EppMetric.Builder, TmchXmlSignature, CloudTasksHelper)` private method. The module then can expose the `CloudTasksHelper` through a getter method to be used in `FlowTestCase`.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 58 at r2](https://reviewable.io/reviews/google/nomulus/1522#-Mwj57NDCdEdn1qQ0LU0:-Mwj57NDCdEdn1qQ0LU1:b-vg3oky) ([raw file](https://github.com/google/nomulus/blob/12a9af04c33a3481328bcc40ebc878fb66103b6d/util/src/main/java/google/registry/util/CloudTasksUtils.java#L58)):*
> ```Java
>   private final String locationId;
>   private final SerializableCloudTasksClient client;
>   private final Clock clock;
> ```

Accidental change?



_Code quote:_
```Java
 private final Clock clock;
```
___
*[core/src/main/java/google/registry/batch/ResaveEntityAction.java, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPB32O2SHsjmK_xlq3:-MwjAyFvAucI0yA3ef8Q:be0hrnf) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/ResaveEntityAction.java#L84)):*

I see your point. I took a look at the test cases and there's no existing test cases to make sure that enqueueing shouldn't happen if the transaction fails. I also looked at the log message which is the first line in ```run()``` "Re-saving entity [resourceKey] which was enqueued at [requestedTime] which made me think that it's referring to ```enqueueAsyncResave()```. This log message seemed to have only served for debugging purposes, wording is ambiguous.

After diving deeper into the flow logic and talked with Ben who originally wrote this, it doesn't make a huge difference whether this should be inside or outside of the transaction, as resave is idempotent So this is going to stay the way it is. 

</blockquote></details>

Your conclusion is correct. However you cannot arrive at it by looking at the tests, because the transactionality of task queues inside datastore transactions is guaranteed by the runtime.  Therefore they are assumed to be correct and not tested in unit tests. Just like we don't test SQL transactionality in unit tests usually -- we may test that, but the absence of the such tests should not constitute proof that the transactionality is not needed for a certain operation.

Also,  if it doesn't matter whether they are in a transaction or not, I think it's better to de-couple them. The simpler the better, and it makes clear to the next reader that they are not transactionally linked.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522)* status: 1 of 16 files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java, line 77 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPA0qq9Wum8qz-mFz7:-Mwmcm2R1ZxcdQ91P9a2:b6vpamg) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java#L77)):*

If you have a PR that's based on another PR, depending on how much dependency there is, you might actually want to have the dependent PR branch off of the base PR. Note that this is not always recommended because it might become harder later on to rebase correctly.
</blockquote></details>

originally I was doing that and it got a bit confusing..and my original implementation was very different from these two PRs. I decided to split them and start fresh off of the master instead. Thank you for the suggestion though! 
___
*[core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java, line 75 at r2](https://reviewable.io/reviews/google/nomulus/1522#-Mwj4S1C0_Ee5Zen-RjL:-MwmcUCxCFdwIfQt3OIm:b-agw65k) ([raw file](https://github.com/google/nomulus/blob/12a9af04c33a3481328bcc40ebc878fb66103b6d/core/src/main/java/google/registry/batch/AsyncTaskEnqueuer.java#L75)):*

Just "#1519" is enough. The numbers are unique in a repo, so you will not mistake a PR with an issue, for example.
</blockquote></details>

cool. I removed it since #1519  was already merged.
___
*[core/src/test/java/google/registry/flows/EppTestComponent.java, line 68 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPD3yj2GWAPj2eNKtu:-MwsjgpeAvM_HXBdaYLD:b-seemwu) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/test/java/google/registry/flows/EppTestComponent.java#L68)):*

If you check all the call sites for the three `create` methods you'll see that all of them are actually using the same clocks (I glanced over them, so I might be wrong), so there's no reason to have all these static methods, which only adds confusion. It looks like all you need is a `create(clock)` public method, which creates the necessary intermediate objects based on the clock and calls the `create(Clock, EppMetric.Builder, TmchXmlSignature, CloudTasksHelper)` private method. The module then can expose the `CloudTasksHelper` through a getter method to be used in `FlowTestCase`.
</blockquote></details>

I looked into it and now there's only one ```create()``` method which takes only ```clock``` to guarantee everything dependent on ```clock``` is using the same clock. 
___
*[core/src/main/java/google/registry/batch/ResaveEntityAction.java, line 84 at r1](https://reviewable.io/reviews/google/nomulus/1522#-MwPB32O2SHsjmK_xlq3:-Mwmi8DP7VOLCQHyRlHA:b-jxqjej) ([raw file](https://github.com/google/nomulus/blob/c58ff48fd66ce0d14a12b18ce569d4926d0305de/core/src/main/java/google/registry/batch/ResaveEntityAction.java#L84)):*
>  queues inside datastore transactions is guaranteed by the runtime.  Therefore they are assumed to be correct and not tested in unit tests. Just like we don't test SQL transactionality in unit tests usually -- we may test that, but the absence of the such tests should not constitute proof that the transactionality is not needed for a certain operation.
> Also,  if it doesn't matter whether they are in a transaction or not, I think it's better to de-couple them. The simpler the better, and it makes clear to the next reader that they are not transactionally linked.
___
*[util/src/main/java/google/registry/util/CloudTasksUtils.java, line 58 at r2](https://reviewable.io/reviews/google/nomulus/1522#-Mwj57NDCdEdn1qQ0LU0:-MwmhTgc72CQ6F0mxUEu:b-3xem34) ([raw file](https://github.com/google/nomulus/blob/12a9af04c33a3481328bcc40ebc878fb66103b6d/util/src/main/java/google/registry/util/CloudTasksUtils.java#L58)):*

Accidental change?
</blockquote></details>

Originally, I was actually trying to see if setting the clock is necessary so I had to remove the keyword `final` to introduce a setter. I realized it's not necessary to do that so I removed it and I didn't move it back to where `final` variables are.

jianglai (changes requested):
> Reviewed 15 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/test/java/google/registry/flows/FlowTestCase.java, line 91 at r3](https://reviewable.io/reviews/google/nomulus/1522#-Mx2WRnq4I6bDcI0RyD1:-Mx2WRnq4I6bDcI0RyD2:b-jsd4go) ([raw file](https://github.com/google/nomulus/blob/d98d387e2c7e28bee85c81ac68a36151ccf4a3ef/core/src/test/java/google/registry/flows/FlowTestCase.java#L91)):*
> ```Java
>   protected CloudTasksHelper cloudTasksHelper;
> 
>   private FakesAndMocksModule fakesAndMocksModule;
> ```

This doesn't have to be a field. It's can be a temporary variable.



_Code quote:_
```Java
fakesAndMocksModule;
```
___
*[core/src/test/java/google/registry/flows/domain/DomainDeleteFlowTest.java, line 582 at r2](https://reviewable.io/reviews/google/nomulus/1522#-Mx2WdP713kZ1gt9u6hA:-Mx2WdP713kZ1gt9u6hB:b-6jt4oh) ([raw file](https://github.com/google/nomulus/blob/12a9af04c33a3481328bcc40ebc878fb66103b6d/core/src/test/java/google/registry/flows/domain/DomainDeleteFlowTest.java#L582)):*
> ```Java
>     clock.advanceOneMilli();
>     runFlowAssertResponse(loadFile("domain_delete_response_autorenew_fee.xml", FEE_11_MAP));
>     cloudTasksHelper.assertTasksEnqueued(
> ```

Why is this deleted?



_Code quote:_
```Java
 cloudTasksHelper.assertTasksEnqueued(
```

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1522)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

PR #1524, opened 2022-02-15
Do full database comparison during replay tests

mindhog:
Replay tests currently only verify that the contents of a transaction
can be successfully replicated to the other database.  They do not verify that
the contents of both databases are equivalent.  As a result, we miss any
changes omitted from the transaction (as was the case with entity deletions).
    
This change adds a final database comparison to ReplayExtension so we can
safely say that the databases are in the same state at the end of a test.

Doing a full database comparison surfaces a number of problems.  Some of these
problems are fixed in this PR (notably, the lack of replication during 
`jpaTm().delete(entity)`, and update timestamp issues when changing a Registrar's
TLDs or DomainContent grace periods).  Others are simply suppressed by disabling
full database comparison for the specific test or disabling the comparison of the
problematic entity type.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1524)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainContent.java
+++ b/core/src/main/java/google/registry/model/domain/DomainContent.java
@@ -1035,17 +1035,20 @@ public class DomainContent extends EppResource
 
     public B setGracePeriods(ImmutableSet<GracePeriod> gracePeriods) {
       getInstance().gracePeriods = gracePeriods;
+      getInstance().resetUpdateTimestamp();
       return thisCastToDerived();
     }
 
     public B addGracePeriod(GracePeriod gracePeriod) {
       getInstance().gracePeriods = union(getInstance().getGracePeriods(), gracePeriod);
+      getInstance().resetUpdateTimestamp();
       return thisCastToDerived();
     }
 
     public B removeGracePeriod(GracePeriod gracePeriod) {
       getInstance().gracePeriods =
           CollectionUtils.difference(getInstance().getGracePeriods(), gracePeriod);
+      getInstance().resetUpdateTimestamp();
       return thisCastToDerived();
     }
 
--- a/core/src/main/java/google/registry/model/ofy/CommitLoggedWork.java
+++ b/core/src/main/java/google/registry/model/ofy/CommitLoggedWork.java
@@ -41,7 +41,7 @@ import org.joda.time.DateTime;
 
 /** Wrapper for {@link Supplier} that associates a time with each attempt. */
 @DeleteAfterMigration
-class CommitLoggedWork<R> implements Runnable {
+public class CommitLoggedWork<R> implements Runnable {
 
   private final Supplier<R> work;
   private final Clock clock;
--- a/core/src/main/java/google/registry/model/registrar/Registrar.java
+++ b/core/src/main/java/google/registry/model/registrar/Registrar.java
@@ -46,6 +46,7 @@ import static google.registry.util.X509Utils.loadCertificate;
 import static java.util.Comparator.comparing;
 import static java.util.function.Predicate.isEqual;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
@@ -824,6 +825,7 @@ public class Registrar extends ImmutableObject
 
     public Builder setAllowedTlds(Set<String> allowedTlds) {
       getInstance().allowedTlds = ImmutableSortedSet.copyOf(assertTldsExist(allowedTlds));
+      getInstance().lastUpdateTime = UpdateAutoTimestamp.create(null);
       return this;
     }
 
@@ -988,6 +990,16 @@ public class Registrar extends ImmutableObject
 
     public Builder setRegistryLockAllowed(boolean registryLockAllowed) {
       getInstance().registryLockAllowed = registryLockAllowed;
+      return this;
+    }
+
+    /**
+     * This lets tests set the update timestamp in cases where setting fields resets the timestamp
+     * and breaks the verification that an object has not been updated since it was copied.
+     */
+    @VisibleForTesting
+    public Builder setLastUpdateTime(DateTime timestamp) {
+      getInstance().lastUpdateTime = UpdateAutoTimestamp.create(timestamp);
       return this;
     }
 
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -30,6 +30,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 import com.google.common.flogger.FluentLogger;
+import com.googlecode.objectify.Key;
 import google.registry.model.ImmutableObject;
 import google.registry.model.common.DatabaseMigrationStateSchedule;
 import google.registry.model.common.DatabaseMigrationStateSchedule.ReplayDirection;
@@ -604,6 +605,7 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
       managedEntity = getEntityManager().merge(entity);
     }
     getEntityManager().remove(managedEntity);
+    transactionInfo.get().addDelete(VKey.from(Key.create(entity)));
     return managedEntity;
   }
 
--- a/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
+++ b/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java
@@ -605,7 +605,13 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
       managedEntity = getEntityManager().merge(entity);
     }
     getEntityManager().remove(managedEntity);
-    transactionInfo.get().addDelete(VKey.from(Key.create(entity)));
+
+    // We check shouldReplicate() in TransactionInfo.addDelete(), but we have to check it here as
+    // well prior to attempting to create a datastore key because a non-replicated entity may not
+    // have one.
+    if (shouldReplicate(entity.getClass())) {
+      transactionInfo.get().addDelete(VKey.from(Key.create(entity)));
+    }
     return managedEntity;
   }
 
@@ -829,6 +835,12 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
     replaySqlToDatastoreOverrideForTest.set(Optional.empty());
   }
 
+  /** Returns true if the entity class should be replicated from SQL to datastore. */
+  private static boolean shouldReplicate(Class<?> entityClass) {
+    return !NonReplicatedEntity.class.isAssignableFrom(entityClass)
+        && !SqlOnlyEntity.class.isAssignableFrom(entityClass);
+  }
+
   private static class TransactionInfo {
     ReadOnlyCheckingEntityManager entityManager;
     boolean inTransaction = false;
@@ -883,12 +895,6 @@ public class JpaTransactionManagerImpl implements JpaTransactionManager {
       if (contentsBuilder != null && shouldReplicate(key.getKind())) {
         contentsBuilder.addDelete(key);
       }
-    }
-
-    /** Returns true if the entity class should be replicated from SQL to datastore. */
-    private boolean shouldReplicate(Class<?> entityClass) {
-      return !NonReplicatedEntity.class.isAssignableFrom(entityClass)
-          && !SqlOnlyEntity.class.isAssignableFrom(entityClass);
     }
 
     private void recordTransaction() {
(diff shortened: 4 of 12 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1524)* status: all files reviewed, 5 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/registrar/Registrar.java, line 828 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzGHTI9rbBTXaLqUsH:-MvzGHTI9rbBTXaLqUsI:b4i399o) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/main/java/google/registry/model/registrar/Registrar.java#L828)):*
> ```Java
>     public Builder setAllowedTlds(Set<String> allowedTlds) {
>       getInstance().allowedTlds = ImmutableSortedSet.copyOf(assertTldsExist(allowedTlds));
>       getInstance().lastUpdateTime = UpdateAutoTimestamp.create(null);
> ```

I'm  a bit confused. Is this the fix for whatever registrar-update-timestamp issues you were seeing in the comparison? 

I'm surprised that the update timestamp wouldn't be changing whenever the set of allowed TLDs is changed. It's just a text[] field in Postgres. 
___
*[core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java, line 2193 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzHjgL6Pn0QZatIpbx:-MvzHjgL6Pn0QZatIpby:b-17ljhg) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2193)):*
> ```Java
>     assertAboutEppExceptions().that(thrown).marshalsToXml();
>   }
> ```

nit, extra newline
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzLObk1AWpQVrv_5DY:-MvzLObk1AWpQVrv_5DZ:bv26oso) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L138)):*
> ```Java
>   List<DomainBase> expectedUpdates = new ArrayList<>();
>   boolean enableDomainTimestampChecks;
>   boolean enableDatabaseCompare = true;
> ```

this seems a bit confusingly named that we have both `compare` and `enableDatabaseCompare`



actually i think we can just remove `compare`? Is it ever false? I don't think it is
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 371 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzPDSS49kv-v3rjDBq:-MvzPDSS49kv-v3rjDBr:b-y60wul) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L371)):*
> ```Java
> 
>     // Build a map containing all of the SQL entities indexed by their key.
>     HashMap<Object, Object> sqlEntities = new HashMap<>();
> ```

can this be `HashMap<Object, ImmutableObject>`? 
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 417 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzNnLP7l4p8tsOCLs2:-MvzNnLP7l4p8tsOCLs3:b-r649mp) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L417)):*
> ```Java
>     }
> 
>     // Report any objects that
> ```

this comment got cut off

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1524)* status: 10 of 12 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/model/registrar/Registrar.java, line 828 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzGHTI9rbBTXaLqUsH:-Mw1am5g1Q5T-x4b5Ksc:ba8a5i8) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/main/java/google/registry/model/registrar/Registrar.java#L828)):*

I'm  a bit confused. Is this the fix for whatever registrar-update-timestamp issues you were seeing in the comparison? 

I'm surprised that the update timestamp wouldn't be changing whenever the set of allowed TLDs is changed. It's just a text[] field in Postgres. 
</blockquote></details>

Yep, me too :-)

Apparently Hiberrnate treats arrays similarly to collections decomposed into associated tables and does not modify the scalar fields of the owning record when they are modified.

One of the things I want to do today is try a more exhaustive experiment to make sure that 1) timestamp fields always get updated when all scalar field types are modified and 2) timestamp fields always _don't_ get updated when non-scalar fields are modified.  I'm less concerned about 2, but gaining some certainty here would be nice.

To be completely safe, unless JPA fully specifies the behavior we should arguably reset the timestamp for _any_ record updates, given that it's always possible for an implementation to update only modified fields in SQL.
___
*[core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java, line 2193 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzHjgL6Pn0QZatIpbx:-Mw1d5vA7DV29NR4xGm6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2193)):*

nit, extra newline
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzLObk1AWpQVrv_5DY:-Mw1dEmQEIzVQjgUYQwX:b-x7k8hl) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L138)):*

this seems a bit confusingly named that we have both `compare` and `enableDatabaseCompare`



actually i think we can just remove `compare`? Is it ever false? I don't think it is
</blockquote></details>

Yeah, it is a bit confusing, and we should just remove it (it was originally for a similar purpose to enableDatabaseCompare -- allowing us to fix failures of individual change comparisons during replay incrementally).

However, I'm reluctant to remove it in this PR given that (in spite of the name) it's unrelated.  Also ReplayExtension goes away after migration.  There are only two instances of at-DoNotCompare remaining, so I'm inclined to just leave it as is and remove the enableDatabaseCompare when those are fixed. 
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 371 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzPDSS49kv-v3rjDBq:-Mw1iFOi9_m7MqUBDeEF:b7citto) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L371)):*

can this be `HashMap<Object, ImmutableObject>`? 
</blockquote></details>

It could, but I'd have to introduce another type-cast where I populate it -- it's populated from `Stream<T> loadStreamOf(Class<T>)` where `Class<T>` is unconstrained :-/ It would save the other cast where I retrieve it...  seems like a wash?
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 417 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzNnLP7l4p8tsOCLs2:-Mw1jK7H0p5juddonQPB:b-njvruu) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L417)):*

this comment got cut off
</blockquote></details>

My train of thought derailed ;-)  Done.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1524)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/registrar/Registrar.java, line 828 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzGHTI9rbBTXaLqUsH:-Mw2K_t6AmZtsAiQQsgc:b-22h52w) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/main/java/google/registry/model/registrar/Registrar.java#L828)):*

Yep, me too :-)

Apparently Hiberrnate treats arrays similarly to collections decomposed into associated tables and does not modify the scalar fields of the owning record when they are modified.

One of the things I want to do today is try a more exhaustive experiment to make sure that 1) timestamp fields always get updated when all scalar field types are modified and 2) timestamp fields always _don't_ get updated when non-scalar fields are modified.  I'm less concerned about 2, but gaining some certainty here would be nice.

To be completely safe, unless JPA fully specifies the behavior we should arguably reset the timestamp for _any_ record updates, given that it's always possible for an implementation to update only modified fields in SQL.
</blockquote></details>

Does this mean that some of the tests in this PR would actually fail without this change?  I'm hoping so :) 
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzLObk1AWpQVrv_5DY:-Mw2Kh8l0-shqICuMBLI:b-wskkkf) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L138)):*

Yeah, it is a bit confusing, and we should just remove it (it was originally for a similar purpose to enableDatabaseCompare -- allowing us to fix failures of individual change comparisons during replay incrementally).

However, I'm reluctant to remove it in this PR given that (in spite of the name) it's unrelated.  Also ReplayExtension goes away after migration.  There are only two instances of at-DoNotCompare remaining, so I'm inclined to just leave it as is and remove the enableDatabaseCompare when those are fixed. 
</blockquote></details>

It's not really unrelated since we're adding a similarly-named variable. `compare` is never used, let's just remove it here (it's always true)
___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 371 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzPDSS49kv-v3rjDBq:-Mw2L9qV57oPdCgvA1mn:b-y3auwe) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L371)):*

It could, but I'd have to introduce another type-cast where I populate it -- it's populated from `Stream<T> loadStreamOf(Class<T>)` where `Class<T>` is unconstrained :-/ It would save the other cast where I retrieve it...  seems like a wash?
</blockquote></details>

Yeah, eh. Annoying that ALL_JPA_ENTITIES is a set of `Class<?>` and not `Class<? extends ImmutableObject>`

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1524)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/model/registrar/Registrar.java, line 828 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzGHTI9rbBTXaLqUsH:-Mw346EQEViFeodquz-c:bb77ufg) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/main/java/google/registry/model/registrar/Registrar.java#L828)):*

Does this mean that some of the tests in this PR would actually fail without this change?  I'm hoping so :) 
</blockquote></details>

Yes.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1524)* status: 11 of 12 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/testing/ReplayExtension.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1524#-MvzLObk1AWpQVrv_5DY:-Mw37tdt40yXsfAtUOEB:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b540f6b93bac6927110b8f4cabef6c1c5e1783aa/core/src/test/java/google/registry/testing/ReplayExtension.java#L138)):*

It's not really unrelated since we're adding a similarly-named variable. `compare` is never used, let's just remove it here (it's always true)
</blockquote></details>

Done.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1524)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1525, opened 2022-02-16
Disable prober data deletion cron job in prod & sandbox

CydeWeys:
This is going to unnecessarily make the database migration more complex, and we
don't need them that badly. We'll re-enable these cron jobs once we've written
the new version of this action that handles Cloud SQL correctly (the current
version only does Datastore anyway).

### DIFF
--- a/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml
+++ b/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml
@@ -254,16 +254,6 @@
   </cron>
 
   <cron>
-    <url><![CDATA[/_dr/cron/fanout?queue=retryable-cron-tasks&endpoint=/_dr/task/deleteProberData&runInEmpty]]></url>
-    <description>
-      This job clears out data from probers and runs once a week.
-    </description>
-    <schedule>every monday 14:00</schedule>
-    <timezone>UTC</timezone>
-    <target>backend</target>
-  </cron>
-
-  <cron>
     <url><![CDATA[/_dr/cron/fanout?queue=retryable-cron-tasks&endpoint=/_dr/task/exportReservedTerms&forEachRealTld]]></url>
     <description>
       Reserved terms export to Google Drive job for creating once-daily exports.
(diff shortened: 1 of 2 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1525)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml, line 257 at r1](https://reviewable.io/reviews/google/nomulus/1525#-Mw3DhDQAo16aCETA8_z:-Mw3DhDR3n3L4p3dyhSc:b7snvii) ([raw file](https://github.com/google/nomulus/blob/453af876154cb36fa5f73ee7164b0e40867393ac/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml#L257)):*
> ```XML
> 
>   <cron>
>     <url><![CDATA[/_dr/cron/fanout?queue=retryable-cron-tasks&endpoint=/_dr/task/deleteProberData&runInEmpty]]></url>
> ```

Only suggestion is maybe just comment these out with this explanation so we don't have to dig through history to retrieve them (I'm assuming the endpoints will be the same in SQL primary)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1525)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml, line 257 at r1](https://reviewable.io/reviews/google/nomulus/1525#-Mw3DhDQAo16aCETA8_z:-Mw7OBcb5e4JuJ8_vU94:b-t6vcjs) ([raw file](https://github.com/google/nomulus/blob/453af876154cb36fa5f73ee7164b0e40867393ac/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml#L257)):*

Only suggestion is maybe just comment these out with this explanation so we don't have to dig through history to retrieve them (I'm assuming the endpoints will be the same in SQL primary)
</blockquote></details>

I thought about doing that, but commenting stuff out like that is discouraged by the style guide, and it's easy enough to retrieve them from git history anway.

PR #1532, opened 2022-02-25
Don't reset the update time for TLD updates

mindhog:
It turns out that the reason that the Registrar update timestamp isn't updated
for some of the tests is because the record is updated unchanged.  We can
avoid this problem by not trying to update the registrar to the same value.
So in this case, if the registrar alreay contains the TLD we're adding, don't
try to add it.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1532)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1532)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1535, opened 2022-02-28
Use built-in Java URL connections instead of UrlFetchService

gbrodman:
- Use the standard HttpsURLConnection to write/read data
- Rewrite RdeReporter, Nordn*Action, and Marksdb classes and related
  tests to conform to the new format
- Remove FakeURLFetchService and ForwardingUrlFetchService as they weren't used
- Refactor UrlFetchException to UrlConnectionException
- Refactor UrlFetchUtils to UrlConnectionUtils

I will need to test this on Alpha. Fortunately the connections that
don't require auth (e.g. TMDB downloading) should be testable.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1535)


gbrodman:
Most of these use authorization or real data (so they can't be tested on alpha) however the TmchCrlAction doesn't. I tested that on alpha and it works; it pulls down the list correctly.

jianglai (changes requested):
> Reviewed 8 of 25 files at r1, 2 of 6 files at r2, 15 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/main/java/google/registry/rde/RdeReporter.java, line 82 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7X9MaDHfEFg3Av1wQ:-Mx7X9MaDHfEFg3Av1wR:b-ra2u2u) ([raw file](https://github.com/google/nomulus/blob/178702ded3ea7b60dc0f6bc2214f9776b1a1b619/core/src/main/java/google/registry/rde/RdeReporter.java#L82)):*
> ```Java
>     req.addHeader(new HTTPHeader(AUTHORIZATION, "Basic " + token));
>     req.setPayload(reportBytes);
>     logger.atInfo().log("Sending report:\n%s", new String(reportBytes, UTF_8));
> ```

Do we not want this logging message?



_Code quote:_
```Java
logger.atInfo().log("Sending report:\n%s", new String(reportBytes, UTF_8));
```
___
*[core/src/main/java/google/registry/request/Modules.java, line 47 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7VHBl4sezU7wblxKv:-Mx7VHBl4sezU7wblxKw:b-ey28rn) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/request/Modules.java#L47)):*
> ```Java
>   }
> 
>   /** Dagger module for {@link UrlConnectionService}, injectable for testing. */
> ```

This part seems to be wrong. The interface itself is injectable, but the module is not.



_Code quote:_
```Java
, injectable for testing.
```
___
*[core/src/main/java/google/registry/tmch/Marksdb.java, line 59 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7ZuT75JKH_J8XO6eZ:-Mx7ZuT75JKH_J8XO6e_:b67e6pi) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/tmch/Marksdb.java#L59)):*
> ```Java
> 
>   @Inject UrlConnectionService urlConnectionService;
> ```

Extra line?
___
*[core/src/main/java/google/registry/tmch/NordnUploadAction.java, line 98 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7_6Gn2dn_3Fr_Kw91:-Mx7_6Gn2dn_3Fr_Kw92:b-ooqyi6) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L98)):*
> ```Java
>   @Inject LordnRequestInitializer lordnRequestInitializer;
>   @Inject UrlConnectionService urlConnectionService;
> ```

Is this line added by the formatter?
___
*[core/src/main/java/google/registry/tmch/NordnVerifyAction.java, line 70 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7_dHR3_oucEqI2cQL:-Mx7_dHR3_oucEqI2cQM:brhumzt) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/tmch/NordnVerifyAction.java#L70)):*
> ```Java
>   @Inject Response response;
>   @Inject UrlConnectionService urlConnectionService;
> ```

Same here.
___
*[core/src/test/java/google/registry/rde/RdeReportActionTest.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7fl_R3onXwUGkd3A5:-Mx7fl_R3onXwUGkd3A6:bszp4im) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/rde/RdeReportActionTest.java#L102)):*
> ```Java
>           return httpUrlConnection;
>         };
>     when(httpUrlConnection.getOutputStream()).thenReturn(connectionOutputStream);
> ```

This is repeated in almost all tests that uses a `UrlConnectionService`. Is it better to make it a separate `FakeUrlConnectionService` that returns a mocked `HttpUrlConnection`? The mocking on `getUrl()` and `getOutputStream()` can then be done just once.



_Code quote:_
```Java
    reporter.urlConnectionService =
        (url) -> {
          when(httpUrlConnection.getURL()).thenReturn(url);
          return httpUrlConnection;
        };
    when(httpUrlConnection.getOutputStream()).thenReturn(connectionOutputStream);
```
___
*[core/src/test/java/google/registry/request/UrlConnectionUtilsTest.java, line 46 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7gsfk4u7K8M3v5s2m:-Mx7gsfk4u7K8M3v5s2n:b-p8gtxc) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/request/UrlConnectionUtilsTest.java#L46)):*
> ```Java
> public class UrlConnectionUtilsTest {
> 
>   @RegisterExtension final AppEngineExtension appEngine = AppEngineExtension.builder().build();
> ```

Is this extension still needed?



_Code quote:_
```Java
@RegisterExtension final AppEngineExtension appEngine = AppEngineExtension.builder().build();
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: all files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/rde/RdeReporter.java, line 82 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7X9MaDHfEFg3Av1wQ:-MxA_-Ov3vyYx1uF6Q2D:b-126f0p) ([raw file](https://github.com/google/nomulus/blob/178702ded3ea7b60dc0f6bc2214f9776b1a1b619/core/src/main/java/google/registry/rde/RdeReporter.java#L82)):*

Do we not want this logging message?
</blockquote></details>

ah, we should probably keep it around 
___
*[core/src/main/java/google/registry/request/Modules.java, line 47 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7VHBl4sezU7wblxKv:-MxAU-wiDrbnaz3NPd5q:bddqqs4) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/request/Modules.java#L47)):*

This part seems to be wrong. The interface itself is injectable, but the module is not.
</blockquote></details>

yeah we probably need to only mention that in the interface 
___
*[core/src/main/java/google/registry/tmch/Marksdb.java, line 59 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7ZuT75JKH_J8XO6eZ:-MxATdKtBHmR7687cExE:b-9d0li6) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/tmch/Marksdb.java#L59)):*

Extra line?
</blockquote></details>

added by the formatter :/ 
___
*[core/src/main/java/google/registry/tmch/NordnUploadAction.java, line 98 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7_6Gn2dn_3Fr_Kw91:-MxATcQ6DWf0h5W1i_q2:bea5qju) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L98)):*

Is this line added by the formatter?
</blockquote></details>

yep :/ 
___
*[core/src/main/java/google/registry/tmch/NordnVerifyAction.java, line 70 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7_dHR3_oucEqI2cQL:-MxATehE0f429gHpN3aN:bowhoe6) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/main/java/google/registry/tmch/NordnVerifyAction.java#L70)):*

Same here.
</blockquote></details>

added by the formatter :/
___
*[core/src/test/java/google/registry/rde/RdeReportActionTest.java, line 102 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7fl_R3onXwUGkd3A5:-MxAZ_If3FwW_mTfZ4oQ:b67g4zj) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/rde/RdeReportActionTest.java#L102)):*

This is repeated in almost all tests that uses a `UrlConnectionService`. Is it better to make it a separate `FakeUrlConnectionService` that returns a mocked `HttpUrlConnection`? The mocking on `getUrl()` and `getOutputStream()` can then be done just once.
</blockquote></details>

We can do this but only to some degree. Not all the tests use the output stream (since not all actions are a POST/PUT request). This is done. 
___
*[core/src/test/java/google/registry/request/UrlConnectionUtilsTest.java, line 46 at r3](https://reviewable.io/reviews/google/nomulus/1535#-Mx7gsfk4u7K8M3v5s2m:-MxAVK_BAjAja6Cu3Ior:b-qjrggk) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/request/UrlConnectionUtilsTest.java#L46)):*

Is this extension still needed?
</blockquote></details>

it is not!

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: 17 of 26 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[core/src/main/java/google/registry/tmch/LordnRequestInitializer.java, line 42 at r4](https://reviewable.io/reviews/google/nomulus/1535#-MxAovLj8a9gbXRFTbMf:-MxAovLj8a9gbXRFTbMg:b4xd6v3) ([raw file](https://github.com/google/nomulus/blob/c1df0fed74aa25e7f2ead601519eb074c5fd8294/core/src/main/java/google/registry/tmch/LordnRequestInitializer.java#L42)):*
> ```Java
>   void initialize(HttpURLConnection connection, String tld) {
>     getMarksDbLordnCredentials(tld)
>         .ifPresent(login -> UrlConnectionUtils.setBasicAuth(connection, login));
> ```

It seems like there's a potential error condition here that's not being handled?  If the credentials don't exist then this method is a noop -- is that expected from all callers?
___
*[core/src/main/java/google/registry/tmch/NordnUploadAction.java, line 210 at r4](https://reviewable.io/reviews/google/nomulus/1535#-MxApJ7v7WYooRfPNNq9:-MxApJ7v7WYooRfPNNqA:b-wv0u3k) ([raw file](https://github.com/google/nomulus/blob/c1df0fed74aa25e7f2ead601519eb074c5fd8294/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L210)):*
> ```Java
>       if (logger.atInfo().isEnabled()) {
>         String responseContent = new String(getResponseBytes(connection), US_ASCII);
>         if (responseContent.isEmpty()) {
> ```

And we're sure this can't be `null`?  That used to be a possbility.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: 17 of 26 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/main/java/google/registry/tmch/LordnRequestInitializer.java, line 42 at r4](https://reviewable.io/reviews/google/nomulus/1535#-MxAovLj8a9gbXRFTbMf:-MxBWShB0s2aodo4I2CK:b-toqail) ([raw file](https://github.com/google/nomulus/blob/c1df0fed74aa25e7f2ead601519eb074c5fd8294/core/src/main/java/google/registry/tmch/LordnRequestInitializer.java#L42)):*

Yes, and this is how it was before. I don't really know what situations it wouldn't be configured in (?) 
</blockquote></details>

Well, it's probably not configured on some of our environments (probably anything that's not prod or sandbox, I'd guess).  Anyway, looks like it's intended that it should try to make the request even if it doesn't have basic auth to attach.

jianglai (changes requested):
> Reviewed 7 of 9 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[core/src/test/java/google/registry/testing/FakeUrlConnectionService.java, line 26 at r5](https://reviewable.io/reviews/google/nomulus/1535#-MxD02KW31aZVQyfHXdC:-MxD02KW31aZVQyfHXdD:b-j4uf9h) ([raw file](https://github.com/google/nomulus/blob/74688f4ce686497a216e22aa0bde33e8ba94942f/core/src/test/java/google/registry/testing/FakeUrlConnectionService.java#L26)):*
> ```Java
> public class FakeUrlConnectionService implements UrlConnectionService {
> 
>   private final HttpURLConnection connection;
> ```

Since you are calling "when()" on the connection, it will have to be a mock. Why not just construct a mock in place and return it via a getter so it can be further stubbed in each tests?

Also, I understand that not all requests are PUT/POST, but does it hurt to provide an output stream? If it's a GET, the output stream will simply not be used.

The same goes for tracking the URLs attempted. The fake can always keep a list, and if it is not needed, it is simply the not used.

Basically the fake should be the "common denominator" that accommodates different test cases. It makes the tests easier to understand and write, as it sort of provides some level of abstraction as to exactly which interactions with the connection does a certain test method involve.



_Code quote:_
```Java
private final HttpURLConnection connection;
```
___
*[core/src/test/java/google/registry/tmch/TmchActionTestCase.java, line 58 at r3](https://reviewable.io/reviews/google/nomulus/1535#-MxD2J8v7t1EKCREUyjQ:-MxD2J8w912XLZVj9DGr:b-lj09vh) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/tmch/TmchActionTestCase.java#L58)):*
> ```Java
>     marksdb.urlConnectionService =
>         (url) -> {
>           when(httpUrlConnection.getURL()).thenReturn(url);
> ```

For example, here I have to wonder why is this mocking not needed (which is answered in `TmchCrlActionTest`). This kind of "optimization" seems to introduce more confusion than if we just have a fake that "does it all". What do you think?



_Code quote:_
```Java
when(httpUrlConnection.getURL()).thenReturn(url);
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: 18 of 26 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[core/src/test/java/google/registry/testing/FakeUrlConnectionService.java, line 26 at r5](https://reviewable.io/reviews/google/nomulus/1535#-MxD02KW31aZVQyfHXdC:-MxGCMjw-WM0klFmNIDO:bdyzblo) ([raw file](https://github.com/google/nomulus/blob/74688f4ce686497a216e22aa0bde33e8ba94942f/core/src/test/java/google/registry/testing/FakeUrlConnectionService.java#L26)):*

Since you are calling "when()" on the connection, it will have to be a mock. Why not just construct a mock in place and return it via a getter so it can be further stubbed in each tests?

Also, I understand that not all requests are PUT/POST, but does it hurt to provide an output stream? If it's a GET, the output stream will simply not be used.

The same goes for tracking the URLs attempted. The fake can always keep a list, and if it is not needed, it is simply the not used.

Basically the fake should be the "common denominator" that accommodates different test cases. It makes the tests easier to understand and write, as it sort of provides some level of abstraction as to exactly which interactions with the connection does a certain test method involve.
</blockquote></details>

Yeah I see what you mean. I could see it going either way -- it's encapsulating the mocking logic in one place, but it does mean that there's a bit of indirection in tests (where I think we tend to bias more towards plainness at the cost of verbosity) and it means we need to enable lenient mocking in a lot of places. 
___
*[core/src/test/java/google/registry/tmch/TmchActionTestCase.java, line 58 at r3](https://reviewable.io/reviews/google/nomulus/1535#-MxD2J8v7t1EKCREUyjQ:-MxGE-528J9pUO9mu1Ct:b-kfumru) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/tmch/TmchActionTestCase.java#L58)):*

For example, here I have to wonder why is this mocking not needed (which is answered in `TmchCrlActionTest`). This kind of "optimization" seems to introduce more confusion than if we just have a fake that "does it all". What do you think?
</blockquote></details>

possibly, though it does so at the cost of potentially having indirection. Thoughts on this revision?

CydeWeys (approved):
:lgtm: pending Lai's issues

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535#-:-MxGJhOWByk7GLQ1EM82:bryqqky)* status: 18 of 26 files reviewed, 2 unresolved discussions (waiting on @jianglai)

jianglai (approved):
> Reviewed 8 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[core/src/test/java/google/registry/testing/FakeUrlConnectionService.java, line 26 at r5](https://reviewable.io/reviews/google/nomulus/1535#-MxD02KW31aZVQyfHXdC:-MxHvTTh26xxeDvYIQ9r:b-q7urht) ([raw file](https://github.com/google/nomulus/blob/74688f4ce686497a216e22aa0bde33e8ba94942f/core/src/test/java/google/registry/testing/FakeUrlConnectionService.java#L26)):*

Yeah I see what you mean. I could see it going either way -- it's encapsulating the mocking logic in one place, but it does mean that there's a bit of indirection in tests (where I think we tend to bias more towards plainness at the cost of verbosity) and it means we need to enable lenient mocking in a lot of places. 
</blockquote></details>

Good point regarding strictness. It looks like Mockito itself is biased toward removing unnecessary mocking by setting strict mocking as default. Now like you I can see it go either way also.
___
*[core/src/test/java/google/registry/tmch/NordnUploadActionTest.java, line 106 at r6](https://reviewable.io/reviews/google/nomulus/1535#-MxHwSi34oFLpkcSAfbK:-MxHwSi34oFLpkcSAfbL:b-lz1lyn) ([raw file](https://github.com/google/nomulus/blob/beaa706b60ac42a8d73b0dfb46fb471f0b045191/core/src/test/java/google/registry/tmch/NordnUploadActionTest.java#L106)):*
> ```Java
>   void beforeEach() throws Exception {
>     inject.setStaticField(Ofy.class, "clock", clock);
>     urlConnectionService = new FakeUrlConnectionService();
> ```

You can probably have `FakeUrlConnectionService`, `HttpUrlConnection`, `ByteArrayOutputStream` all as fields and save a lot of calls to `urlConnectionService.getMockConnection()` and such.

Same comment to the other files as well.



_Code quote:_
```Java
urlConnectionService = new FakeUrlConnectionService();
```
___
*[core/src/test/java/google/registry/tmch/TmchActionTestCase.java, line 58 at r3](https://reviewable.io/reviews/google/nomulus/1535#-MxD2J8v7t1EKCREUyjQ:-MxHxrksCoR5xeImE57l:b-a9zbv0) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/tmch/TmchActionTestCase.java#L58)):*

possibly, though it does so at the cost of potentially having indirection. Thoughts on this revision? 
</blockquote></details>

As I said in the other comment, now that you brought up mocking strictness, I do see things go both ways. If you decide to keep this revision, consider my suggestion about using field variables to reduce unnecessary calls to methods. Or if you would like to preserve strictness, I think the previous revision where the fake does the bare minimum is also fine.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: 18 of 26 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[core/src/test/java/google/registry/testing/FakeUrlConnectionService.java, line 26 at r5](https://reviewable.io/reviews/google/nomulus/1535#-MxD02KW31aZVQyfHXdC:-MxL1MsPEsztbfUNweo4:bsmfnc3) ([raw file](https://github.com/google/nomulus/blob/74688f4ce686497a216e22aa0bde33e8ba94942f/core/src/test/java/google/registry/testing/FakeUrlConnectionService.java#L26)):*

Good point regarding strictness. It looks like Mockito itself is biased toward removing unnecessary mocking by setting strict mocking as default. Now like you I can see it go either way also.
</blockquote></details>

see other comment 
___
*[core/src/test/java/google/registry/tmch/NordnUploadActionTest.java, line 106 at r6](https://reviewable.io/reviews/google/nomulus/1535#-MxHwSi34oFLpkcSAfbK:-MxL1OKL0CCv3LcQpWk-:b-6oabos) ([raw file](https://github.com/google/nomulus/blob/beaa706b60ac42a8d73b0dfb46fb471f0b045191/core/src/test/java/google/registry/tmch/NordnUploadActionTest.java#L106)):*

You can probably have `FakeUrlConnectionService`, `HttpUrlConnection`, `ByteArrayOutputStream` all as fields and save a lot of calls to `urlConnectionService.getMockConnection()` and such.

Same comment to the other files as well.
</blockquote></details>

Yeah, I think keeping with the principle of "tests should be obvious" it's best to keep everything as separate fields as much as possible. 
___
*[core/src/test/java/google/registry/tmch/TmchActionTestCase.java, line 58 at r3](https://reviewable.io/reviews/google/nomulus/1535#-MxD2J8v7t1EKCREUyjQ:-MxL1V-N-m55c_7z6F0R:bflwoc4) ([raw file](https://github.com/google/nomulus/blob/62d710a06425ebb8755b07648204befcf75800e7/core/src/test/java/google/registry/tmch/TmchActionTestCase.java#L58)):*

As I said in the other comment, now that you brought up mocking strictness, I do see things go both ways. If you decide to keep this revision, consider my suggestion about using field variables to reduce unnecessary calls to methods. Or if you would like to preserve strictness, I think the previous revision where the fake does the bare minimum is also fine.
</blockquote></details>

Yeah, I think it's best to have the fake url connection service do fairly little. I think the revision that I have pushed here now keeps the best balance between keeping tests simple/verbose and having the fake url connection service do some of the common work.

jianglai (approved):
> Reviewed 7 of 8 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1535)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1549, opened 2022-03-08
Revise Host index on inet_addresses

weiminyu:
The index on the 'inet_addresses array column should be of gin or gist
type, which index individual array elements. We use gin for now since
host updates are not often, and gin has better accuracy.

Since flyway script V108__... has not been deployed, we  edit the file
in place instead of adding a new script.

This will be followed up with a modified query that can take advantage
of the gin index. Until then we don't expect to see performance
improvement.

The suspected bottlenect query in the whois path is:

select * from "Host" where 'non-ip-string' = any(inet_address) and
deletion_time < now();

It needs to be revised into:

select * from "Host" where array['non-ip-string'] <@ inet_address and
deletion_time < now();

The combined change reduces the query time from 90ms to 30ms in Sandbox,
and from 150ms to 40ms in production.

It is unclear if this solves all problem with whois latency.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1549)

### DIFF
--- a/db/src/main/resources/sql/flyway/V108__add_host_indexes_for_whois.sql
+++ b/db/src/main/resources/sql/flyway/V108__add_host_indexes_for_whois.sql
@@ -12,6 +12,4 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
-
-CREATE INDEX IDXovmntef6l45tw2bsfl56tcugx ON "Host" (deletion_time);
-CREATE INDEX IDXrc77s1ndiemi2vwwudchye214 ON "Host" (inet_addresses);
+CREATE INDEX IDXrc77s1ndiemi2vwwudchye214 ON "Host" USING GIN (inet_addresses);
--- a/db/src/main/resources/sql/flyway/V108__add_host_indexes_for_whois.sql
+++ b/db/src/main/resources/sql/flyway/V108__add_host_indexes_for_whois.sql
@@ -12,4 +12,5 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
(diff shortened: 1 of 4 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1549)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

db/src/main/resources/sql/flyway/V108__add_host_indexes_for_whois.sql
```diff
@@ -12,6 +12,5 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
-
-CREATE INDEX IDXovmntef6l45tw2bsfl56tcugx ON "Host" (deletion_time);
```
CydeWeys, line 16: Why did you remove the `deletion_time` index?  It's still going to be useful to have to be able to discern between hosts that are active and those that are not (even if for ad hoc queries and analytics if nothing else).  Not having an index on the column that is used to determine whether or not a given entity is deleted or not in a database that uses a soft-deletion paradigm seems like a mistake.

weiminyu: Query analyzer says that the index on deletion_time is not used in this and similar queries.

I believe the only use cases where an index on deletion_time is useful are:

1. Specific time is given: deletion_time = ...
2. Deletion time is the only parameter: .. where deletion_time > ...;

Otherwise the cardinality of the intermediate result is usually too large to be useful.

PR #1551, opened 2022-03-09
Reorganize new schema changes

weiminyu:
Reorganized new schema changes and make each flyway script update a
single table.

Each flyway script is executed in a single database transaction so that
the script can be rolled back in one shot. It acquires a shared lock on
all tables touched by the script. This is deadlock-prone because in a
busy database, there may be user queries that attempt to lock the same
set of tables, but in different order. By limiting each script to one
table, we avoid the problem.

We should have some a presubmit check to enforce this rule.

All changes have been deployed to Sandbox out-of-band. When doing so,
we changed all CREATE INDEX statements to CREATE INDEX IF NOT EXISTS.

Future deployments should be able to proceed normally.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1551)


CydeWeys (approved):
Approve so long as this method of rewriting existing Flyway files will still apply cleanly to sandbox (I assume none of the files being modified have yet been successfully applied there?).

weiminyu (commented):
Yes, this will apply to sandbox successfully.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1551#-:-MxlVTmv47e7vCG85RXm:b-g0pmdf)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @mindhog)

PR #1553, opened 2022-03-14
Create a Dataflow pipeline to resave EPP resources

gbrodman:
This has two modes.

If `isFast` is false, then we will just load all EPP resources, project them to the current time, and save them.

If `isFast` is true, we will attempt to intelligently load and save only resources that we expect to have changes applied when we project them to the current time. This means resources with pending transfers that have expired, domains with expired grace periods, and non-deleted domains that have expired (we expect that they autorenewed).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1553)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -799,6 +799,11 @@ if (environment == 'alpha') {
               mainClass: 'google.registry.beam.comparedb.ValidateDatabasePipeline',
               metaData: 'google/registry/beam/validate_database_pipeline_metadata.json'
           ],
+      resaveAllEppResources:
+          [
+              mainClass: 'google.registry.beam.resave.ResaveAllEppResourcesPipeline',
+              metaData: 'google/registry/beam/resave_all_epp_resources_pipeline_metadata.json'
+          ],
   ]
   project.tasks.create("stageBeamPipelines") {
     doLast {
--- a/core/src/main/java/google/registry/batch/BatchModule.java
+++ b/core/src/main/java/google/registry/batch/BatchModule.java
@@ -113,6 +113,12 @@ public class BatchModule {
   }
 
   @Provides
+  @Parameter(ResaveAllEppResourcesPipelineAction.PARAM_FAST)
+  static Optional<Boolean> provideIsFast(HttpServletRequest req) {
+    return extractOptionalBooleanParameter(req, ResaveAllEppResourcesPipelineAction.PARAM_FAST);
+  }
+
+  @Provides
   @Named(QUEUE_ASYNC_ACTIONS)
   static Queue provideAsyncActionsPushQueue() {
     return getQueue(QUEUE_ASYNC_ACTIONS);
--- a/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java
+++ b/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java
@@ -1,0 +1,129 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.batch;
+
+import static google.registry.beam.BeamUtils.createJobName;
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
+
+import com.google.api.services.dataflow.Dataflow;
+import com.google.api.services.dataflow.model.LaunchFlexTemplateParameter;
+import com.google.api.services.dataflow.model.LaunchFlexTemplateRequest;
+import com.google.api.services.dataflow.model.LaunchFlexTemplateResponse;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.flogger.FluentLogger;
+import com.google.common.net.MediaType;
+import google.registry.config.RegistryConfig.Config;
+import google.registry.config.RegistryEnvironment;
+import google.registry.request.Action;
+import google.registry.request.Parameter;
+import google.registry.request.Response;
+import google.registry.request.auth.Auth;
+import google.registry.util.Clock;
+import java.util.Optional;
+import javax.inject.Inject;
+
+/**
+ * Starts a Dataflow pipeline that resaves all EPP resources projected to the current time.
+ *
+ * <p>This is useful for a few situations. First, as a fallback option for resource transfers that
+ * have expired pending transfers (this will resolve them), just in case the enqueued action fails.
+ * Second, it will reflect domain autorenews that have happened. Third, it will remove any expired
+ * grace periods.
+ *
+ * <p>There's also the general principle that it's good to have the data in the database remain as
+ * current as is reasonably possible.
+ *
+ * <p>If the <code>?isFast=true</code> query string parameter is passed as true, the pipeline will
+ * only attempt to load, project, and resave entities where we expect one of the previous situations
+ * has occurred. Otherwise, we will load, project, and resave all EPP resources.
+ *
+ * <p>This runs the {@link google.registry.beam.resave.ResaveAllEppResourcesPipeline}.
+ */
+@Action(
+    service = Action.Service.BACKEND,
+    path = ResaveAllEppResourcesPipelineAction.PATH,
+    auth = Auth.AUTH_INTERNAL_OR_ADMIN)
+public class ResaveAllEppResourcesPipelineAction implements Runnable {
+
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+  static final String PATH = "/_dr/task/resaveAllEppResourcesPipeline";
+  static final String PIPELINE_NAME = "resave_all_epp_resources_pipeline";
+
+  public static final String PARAM_FAST = "fast";
+
+  private final String projectId;
+  private final String jobRegion;
+  private final String stagingBucketUrl;
+  private final boolean fast;
+  private final Clock clock;
+  private final Response response;
+  private final Dataflow dataflow;
+
+  @Inject
+  ResaveAllEppResourcesPipelineAction(
+      @Config("projectId") String projectId,
+      @Config("defaultJobRegion") String jobRegion,
+      @Config("beamStagingBucketUrl") String stagingBucketUrl,
+      @Parameter(PARAM_FAST) Optional<Boolean> fast,
+      Clock clock,
+      Response response,
+      Dataflow dataflow) {
+    this.projectId = projectId;
+    this.jobRegion = jobRegion;
+    this.stagingBucketUrl = stagingBucketUrl;
+    this.fast = fast.orElse(false);
+    this.clock = clock;
+    this.response = response;
+    this.dataflow = dataflow;
+  }
+
+  @Override
+  public void run() {
+    response.setContentType(MediaType.PLAIN_TEXT_UTF_8);
+    logger.atInfo().log("Launching ResaveAllEppResourcesPipeline");
+    try {
+      LaunchFlexTemplateParameter parameter =
+          new LaunchFlexTemplateParameter()
+              .setJobName(createJobName("resave-all-epp-resources", clock))
+              .setContainerSpecGcsPath(
+                  String.format("%s/%s_metadata.json", stagingBucketUrl, PIPELINE_NAME))
+              .setParameters(
+                  new ImmutableMap.Builder<String, String>()
+                      .put(PARAM_FAST, Boolean.toString(fast))
+                      .put("registryEnvironment", RegistryEnvironment.get().name())
+                      .build());
+      LaunchFlexTemplateResponse launchResponse =
+          dataflow
+              .projects()
+              .locations()
+              .flexTemplates()
+              .launch(
+                  projectId,
+                  jobRegion,
+                  new LaunchFlexTemplateRequest().setLaunchParameter(parameter))
+              .execute();
+      logger.atInfo().log("Got response: %s", launchResponse.getJob().toPrettyString());
+      String jobId = launchResponse.getJob().getId();
+      response.setStatus(SC_OK);
+      response.setPayload(String.format("Launched resaveAllEppResources pipeline: %s", jobId));
+    } catch (Exception e) {
+      logger.atSevere().withCause(e).log("Template Launch failed.");
+      response.setStatus(SC_INTERNAL_SERVER_ERROR);
+      response.setPayload(String.format("Pipeline launch failed: %s", e.getMessage()));
+    }
+  }
+}
--- a/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
+++ b/core/src/main/java/google/registry/beam/initsql/InitSqlPipeline.java
@@ -158,7 +158,7 @@ public class InitSqlPipeline implements Serializable {
                     .addAll(toKindStrings(PHASE_TWO_ORDERED))
                     .build()));
 
-    // Set up the pipeline to write entity kinds from PHASE_ONE_ORDERED to SQL. Return a object
+    // Set up the pipeline to write entity kinds from PHASE_ONE_ORDERED to SQL. Return an object
     // that signals the completion of the phase.
     PCollection<Void> blocker =
         scheduleOnePhaseWrites(datastoreSnapshot, PHASE_ONE_ORDERED, Optional.empty(), null);
--- a/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java
+++ b/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java
@@ -1,0 +1,174 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.beam.resave;
+
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static org.apache.beam.sdk.values.TypeDescriptors.integers;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import google.registry.beam.common.RegistryJpaIO;
+import google.registry.beam.common.RegistryJpaIO.Read;
+import google.registry.model.EppResource;
+import google.registry.model.contact.ContactResource;
+import google.registry.model.domain.DomainBase;
+import google.registry.model.host.HostResource;
+import google.registry.persistence.PersistenceModule.TransactionIsolationLevel;
+import google.registry.persistence.transaction.CriteriaQueryBuilder;
+import google.registry.util.DateTimeUtils;
+import java.io.Serializable;
+import java.util.concurrent.ThreadLocalRandom;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.PipelineResult;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.GroupIntoBatches;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.WithKeys;
+import org.apache.beam.sdk.util.ShardedKey;
+import org.apache.beam.sdk.values.KV;
+import org.joda.time.DateTime;
+
+/**
+ * A Dataflow Flex pipeline that resaves changed EPP resources in SQL.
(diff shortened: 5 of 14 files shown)


CydeWeys (commented):
Are you sure about that?  What happened to the issue of update times?  Won't this give all resources an update time set to now?

Also the bug mentioned being smarter than the existing Datastore implementation and using queries to only run over resources that actually have pending state needing persisting -- what happened to that?

We want this to be low-impact enough to run at least daily, if not potentially more often than that.  Loading up everything in the DB each time it's run doesn't seem to meet that need.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553#-:-My8WBGzAKLCxCSv5wVr:bibktil)* status: 0 of 3 files reviewed, all discussions resolved

gbrodman (commented):
1. This is something we've talked about before. If an object hasn't changed and we resave it, it's a no-op. This is different from the Datastore behavior, but we were fine with that because it's impossible to change it neatly and it's more technically correct anyway. 

2. Hibernate seemingly gives us that for free, which is good because any hand-written SQL queries we try to use to suss out which resources might have pending actions to resolve could be error-prone or out of date at some point. The Datastore implementation runs over all EPP resources anyway. 

3. This should be the same amount of work as Datastore. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553#-:-My8XKtf1XIzPxAzayG7:bmz1wuu)* status: 0 of 3 files reviewed, all discussions resolved

CydeWeys (commented):
1. So will it or won't it change the auto update timestamp?  Does "no-op" mean the latter?  Also, note how the existing Datastore implementation (in fast mode anyway) checks to see if there are any changes prior to re-saving them -- are you sure that leaving it to Hibernate is as performant?  Would that require a round-trip to the DB or is that check happening in-memory too?

2. Hibernate is not giving it to use for free.  This implementation requires loading all EPP resources from the database.  We can do substantially better than that by specifically querying for only the resources that have pending state that has 'come due'.  So we should.  And it's not liable to go out of date because we don't add any pending state.  The EPP object life cycle is entirely fixed and has been for two decades.

3. Yes, the idea is to improve on Datastore.  We want to be able to run this more often without adding lots of load on the DB.  The entire point of this is to do it better than we did in Datastore now that we have a fully queryable DB that allows us to do so, rather than replicating the same braindead Datastore NoSQL approach.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553#-:-My9K88r6UQl1OvH2Izw:b-uha154)* status: 0 of 3 files reviewed, all discussions resolved

mindhog (commented):
Note on update timestamps: These won't be changed.  They only get written (and updated) when something else in the object has changed.  I expect this is done in memory, doing so with a database reload would be a huge performance hit.

The one caveat is that there are a few places where we now force an update specifically to get the update timestamp to update when it otherwise wouldn't have (specifically when setting values that are broken out into other tables).  I do not believe this to be an issue for this PR, though.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553#-:-MyCPR9k-jntSzsKUtn3:bo87320)* status: 0 of 3 files reviewed, all discussions resolved

gbrodman (commented):
Yeah, this should be better now. The "update timestamps don't get changed" definitely is better behavior and fortunately shouldn't be relevant in this PR. Well, except for the tests -- we have to spy the transaction manager to see if we're attempting to save the unchanged resources even if they don't actually get changed / saved in the DB.

This should be better now. We have a fast mode that handles "expected" out-of-date resources, and a slow mode that handles everything. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553#-:-MyJR40608isCCb2Ea6o:b-lqx8xc)* status: 0 of 4 files reviewed, all discussions resolved

gbrodman:
OK I think this should be good to go

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553)* status: 0 of 14 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1553#-Mzag_qiDI7T1qAbz5DS:-Mzag_qiDI7T1qAbz5DT:b-n6v2lq) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L67)):*
> ```Java
>   static final String PIPELINE_NAME = "resave_all_epp_resources_pipeline";
> 
>   public static final String PARAM_IS_FAST = "isFast";
> ```

`fast` seems like a better name for a URL query parameter than `isFast` (and is consistent with the previous mapreduce that already exists).  If I'm ever calling this endpoint manually (which definitely is an expected use case), `?fast=true` would come to me more naturally than `?isFast=true`.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 72 at r3](https://reviewable.io/reviews/google/nomulus/1553#-Mzah-f0BP5fYRGGF2yo:-Mzah-f0BP5fYRGGF2yp:b11huz5) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L72)):*
> ```Java
>   private final String jobRegion;
>   private final String stagingBucketUrl;
>   private final Optional<Boolean> isFast;
> ```

There's no need for this to be `Optional`.  Just make it a `boolean` and default it to the correct default value (`true`) in the constructor if not specified.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 107 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahM_L4aACJYlXLfM5:-MzahM_L4aACJYlXLfM6:b3dhroh) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L107)):*
> ```Java
>               .setParameters(
>                   new ImmutableMap.Builder<String, String>()
>                       .put(PARAM_IS_FAST, isFast.orElse(Boolean.FALSE).toString())
> ```

Yeah, see this would be better handled in the constructor than here, and it allows the field to be of a simpler type.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahWR7E9-kj4KPBaPP:-MzahWR7E9-kj4KPBaPQ:bny5qde) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L124)):*
> ```Java
>       response.setStatus(SC_OK);
>       response.setPayload(String.format("Launched resaveAllEppResources pipeline: %s", jobId));
> ```

Why only `IOException`?  Surely there's some runtime exceptions that might be thrown here too, that would necessitate proper handling?
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 64 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzamPg-5NOADalSRWav:-MzamPg-5NOADalSRWaw:by9jlkk) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L64)):*
> ```Java
>    */
>   private static final String DOMAINS_TO_PROJECT_QUERY =
>       "FROM Domain d WHERE (d.transferData.transferStatus = 'PENDING' AND"
> ```

And you're sure these are all the relevant cases?
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahwLhBBCMYTZSkObK:-MzahwLhBBCMYTZSkObL:b7kmukv) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L67)):*
> ```Java
>           + " d.transferData.pendingTransferExpirationTime < current_timestamp()) OR"
>           + " (d.registrationExpirationTime < current_timestamp() AND d.deletionTime ="
>           + " (:END_OF_TIME)) OR (EXISTS (SELECT 1 FROM GracePeriod gp WHERE gp.domainRepoId ="
> ```

This seems to me like it could be quite non-performant, if you're doing the `SELECT` as a subquery separately for each domain?  It seems to me like this would be more performant as querying on Domain `JOIN` GracePeriod?
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 146 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzakzAMA-sPpLj1i2eC:-MzakzAMA-sPpLj1i2eD:bz6o46h) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L146)):*
> ```Java
>   private static class MapToNowFunction<T extends EppResource> extends DoFn<T, EppResource> {
> 
>     private final DateTime currentTime;
> ```

I don't think you can pass this in and have it be the same for everything. I think you have to project each resource to `now` where `now` is the _actual_ transaction time of the actual transaction that is re-saving the resource.  Otherwise, you risk having issues with concurrent updates, where a resource has already been modified after the query found it, and now you're attempting to project it to a time earlier than its state already actually is.  That seems prone to errors.

Note that it's only safe to project a resource to the current transaction time or further into the future; projecting it into the past (which you're doing here) is unsafe.  Note that, for this reason, the existing `ResaveAllEppResourcesAction` projects all resources to the current transaction time that is part of the re-save.  You should continue doing it that way here.
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipelineOptions.java, line 23 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzaiG4g3wVyrnrD3LOY:-MzaiG4g3wVyrnrD3LOZ:b-no3zhg) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipelineOptions.java#L23)):*
> ```Java
> 
>   @Description("True if we should attempt to run only over potentially out-of-date EPP resources")
>   boolean getIsFast();
> ```

`getIsFast()` is a redundant name.  The whole point of using the `is` prefix in a name is that `is` is the action.  Is this something we have control over or are we at the whims of auto-generated code?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553)* status: 0 of 14 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1553#-Mzag_qiDI7T1qAbz5DS:-MzvYJDD4fSiIh8mc8YB:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L67)):*

`fast` seems like a better name for a URL query parameter than `isFast` (and is consistent with the previous mapreduce that already exists).  If I'm ever calling this endpoint manually (which definitely is an expected use case), `?fast=true` would come to me more naturally than `?isFast=true`.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 72 at r3](https://reviewable.io/reviews/google/nomulus/1553#-Mzah-f0BP5fYRGGF2yo:-MzvYeMK-Y3z-aT_K4GD:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L72)):*

There's no need for this to be `Optional`.  Just make it a `boolean` and default it to the correct default value (`true`) in the constructor if not specified.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 107 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahM_L4aACJYlXLfM5:-MzvYk7zC7tTwmzFNZnv:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L107)):*

Yeah, see this would be better handled in the constructor than here, and it allows the field to be of a simpler type.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 124 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahWR7E9-kj4KPBaPP:-Mzv_mQ-80Y4guMAr5qK:bnjt9h0) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L124)):*

Why only `IOException`?  Surely there's some runtime exceptions that might be thrown here too, that would necessitate proper handling?
</blockquote></details>

eh fair enough, we can catch Exception. Seems like the existing codebase is a bit of a mix of these.
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 64 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzamPg-5NOADalSRWav:-MzvaC-kEmSZlNdwgWjd:bx9xtj9) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L64)):*

And you're sure these are all the relevant cases?
</blockquote></details>

I think I'm about as confident as I could be, comparing this to `DomainContent`'s `cloneProjectedAtPointInTime` 
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahwLhBBCMYTZSkObK:-N--VBUm2UcTT6jINtDW:b-nbjhbm) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L67)):*

This seems to me like it could be quite non-performant, if you're doing the `SELECT` as a subquery separately for each domain?  It seems to me like this would be more performant as querying on Domain `JOIN` GracePeriod?
</blockquote></details>

Nope, it's actually roughly the same efficiency or faster to use `EXISTS`.  With a join we'd have to worry about deduping result rows, which we don't need to do with EXISTS. 

On production:
`EXPLAIN ANALYZE SELECT d.* FROM "Domain" d WHERE EXISTS (SELECT 1 FROM "GracePeriod" gp WHERE gp.expiration_time < current_timestamp AND gp.domain_repo_id = d.repo_id);`

shows about 3900 - 4000 milliseconds

`EXPLAIN ANALYZE SELECT d.* FROM "Domain" d INNER JOIN "GracePeriod" gp ON gp.domain_repo_id = d.repo_id WHERE gp.expiration_time < current_timestamp;`

shows about 4000-4100 milliseconds which is slower to begin with, but crucially it returns duplicate rows since we don't do a DISTINCT on the repo id. If we do that, it increases the time dramatically. 

https://stackoverflow.com/a/7082510 has more words on the subject.
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 146 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzakzAMA-sPpLj1i2eC:-N--ZhNv0GA_iNxFpuZy:bygusi3) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L146)):*

I don't think you can pass this in and have it be the same for everything. I think you have to project each resource to `now` where `now` is the _actual_ transaction time of the actual transaction that is re-saving the resource.  Otherwise, you risk having issues with concurrent updates, where a resource has already been modified after the query found it, and now you're attempting to project it to a time earlier than its state already actually is.  That seems prone to errors.

Note that it's only safe to project a resource to the current transaction time or further into the future; projecting it into the past (which you're doing here) is unsafe.  Note that, for this reason, the existing `ResaveAllEppResourcesAction` projects all resources to the current transaction time that is part of the re-save.  You should continue doing it that way here.
</blockquote></details>

Ugh yeah, that makes sense. I was hoping to avoid having to open up a transaction for every single resource that we're processing, but I suppose that's the only way. We can just do that and use whatever transaction time we find.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 67 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzahwLhBBCMYTZSkObK:-N-00guR3VqFhSDf17qB:b-y0ip8p) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L67)):*

Nope, it's actually roughly the same efficiency or faster to use `EXISTS`.  With a join we'd have to worry about deduping result rows, which we don't need to do with EXISTS. 

On production:
`EXPLAIN ANALYZE SELECT d.* FROM "Domain" d WHERE EXISTS (SELECT 1 FROM "GracePeriod" gp WHERE gp.expiration_time < current_timestamp AND gp.domain_repo_id = d.repo_id);`

shows about 3900 - 4000 milliseconds

`EXPLAIN ANALYZE SELECT d.* FROM "Domain" d INNER JOIN "GracePeriod" gp ON gp.domain_repo_id = d.repo_id WHERE gp.expiration_time < current_timestamp;`

shows about 4000-4100 milliseconds which is slower to begin with, but crucially it returns duplicate rows since we don't do a DISTINCT on the repo id. If we do that, it increases the time dramatically. 

https://stackoverflow.com/a/7082510 has more words on the subject.
</blockquote></details>

Huh, good to know.
___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 146 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzakzAMA-sPpLj1i2eC:-N-00PnyBgMVAiVse2sQ:b-prsqi4) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L146)):*

Ugh yeah, that makes sense. I was hoping to avoid having to open up a transaction for every single resource that we're processing, but I suppose that's the only way. We can just do that and use whatever transaction time we find. 
</blockquote></details>

We shouldn't have to open up a separate transaction for every resource being saved.  We can batch them.  As currently written, with one transaction per row, won't this absolutely slam the DB and reduce throughput?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java, line 146 at r3](https://reviewable.io/reviews/google/nomulus/1553#-MzakzAMA-sPpLj1i2eC:-N-4TsD-6c5RcNHjbbG8:b-yek773) ([raw file](https://github.com/google/nomulus/blob/5c3d65db28bd089e943abeb51879c48897364b05/core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java#L146)):*

We shouldn't have to open up a separate transaction for every resource being saved.  We can batch them.  As currently written, with one transaction per row, won't this absolutely slam the DB and reduce throughput?
</blockquote></details>

Possibly! Either way, we are capable of batching this up, which I've done here.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 124 at r5](https://reviewable.io/reviews/google/nomulus/1553#-N-cgGrJ1hihmIUZ8pft:-N-cgGrKExg2ZXjiXTKx:b-v0rlgk) ([raw file](https://github.com/google/nomulus/blob/8c71f84de615c823062463cf5f37fc701db08cee/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L124)):*
> ```Java
>       response.setPayload(String.format("Launched resaveAllEppResources pipeline: %s", jobId));
>     } catch (Exception e) {
>       logger.atWarning().withCause(e).log("Template Launch failed.");
> ```

This seems like it should be logged at severe?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1553)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java, line 124 at r5](https://reviewable.io/reviews/google/nomulus/1553#-N-cgGrJ1hihmIUZ8pft:-N-civkA-xhBYaqLlQRs:b-9qmi90) ([raw file](https://github.com/google/nomulus/blob/8c71f84de615c823062463cf5f37fc701db08cee/core/src/main/java/google/registry/batch/ResaveAllEppResourcesPipelineAction.java#L124)):*

This seems like it should be logged at severe?
</blockquote></details>

yeah fair enough

PR #1554, opened 2022-03-14
Add a a delay to task in CommitLogCheckpointAction 

rachelguan:
There is an significant increase of `VerifyException` at at `ExportCommitLogDiffAction.run`. After the switch to `Cloud Task` API, there's a chance that the task gets invoked before the `Datastore` transaction commits. Adding a delay to each submitted task could prevent a task gets invoked before the the transaction is committed.

### DIFF
--- a/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java
+++ b/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java
@@ -28,11 +28,11 @@ import google.registry.model.ofy.CommitLogCheckpointRoot;
 import google.registry.request.Action;
 import google.registry.request.Action.Service;
 import google.registry.request.auth.Auth;
-import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import java.util.Optional;
 import javax.inject.Inject;
 import org.joda.time.DateTime;
+import org.joda.time.Duration;
 
 /**
  * Action that saves commit log checkpoints to Datastore and kicks off a diff export task.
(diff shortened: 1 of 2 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1554)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rachelguan)

jianglai (approved):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1554)* status: all files reviewed, 1 unresolved discussion (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java, line 69 at r1](https://reviewable.io/reviews/google/nomulus/1554#-MyAf-Bj4_261iza867U:-MyAf-Bj4_261iza867V:b-2uh21z) ([raw file](https://github.com/google/nomulus/blob/8cce064b8d0d764a11675351e04d658dfe6081f8/core/src/main/java/google/registry/backup/CommitLogCheckpointAction.java#L69)):*
> ```Java
>    * com.google.common.base.VerifyException}.
>    *
>    * <p>In order to invoke the task after the transaction commits, a reasonable delay should be
> ```

I think we should also comment on why the enqueueing cannot simply be placed after the transaction.



_Code quote:_
```Java
   * <p>In order to invoke the task after the transaction commits, a reasonable delay should be
   * added to each task. The latency of the request is mostly in the range of 4-6 seconds; Choosing
   * a value 30% greater than the upper bound should solve the issue invoking a task before the
   * transaction commits.
```

PR #1508, opened 2022-02-02
Copy the latest revision of BRDA during upload

jianglai:
The revision was hardcoded to 0, which caused problem when we need to
re-run BRDA.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1508)

PR #1556, opened 2022-03-15
Fix a subtle issue in BRDA copy caused by Cloud Tasks

jianglai:
After the Cloud Tasks migration and #1508, the BRDA copy job now
routinely fail on the first try because the revision update is not
commited by the time the Cloud Tasks job enqueued in the same
transaction runs for  the first time. This is because the enqueueing is
a side effect and not part of the transaction. The job eventually
succeeds because of retries.

This PR attempts to mitigate the initial failure by saving the revisions
in a separate transaction before the enqueueing happens, insuring that
it is updated by the time the job runs.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1556)

### DIFF
--- a/core/src/main/java/google/registry/beam/rde/RdeIO.java
+++ b/core/src/main/java/google/registry/beam/rde/RdeIO.java
@@ -68,6 +68,7 @@ import org.apache.beam.sdk.values.PDone;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.bouncycastle.openpgp.PGPPublicKey;
 import org.joda.time.DateTime;
+import org.joda.time.Duration;
 
 public class RdeIO {
 
@@ -261,6 +262,7 @@ public class RdeIO {
 
     private static final FluentLogger logger = FluentLogger.forEnclosingClass();
     private static final long serialVersionUID = 5822176227753327224L;
+    private static final Duration ENQUEUE_DELAY = Duration.standardMinutes(1);
 
     private final CloudTasksUtils cloudTasksUtils;
 
@@ -271,19 +273,11 @@ public class RdeIO {
     @ProcessElement
     public void processElement(
         @Element KV<PendingDeposit, Integer> input, PipelineOptions options) {
-      PendingDeposit key = input.getKey();
-      // We need to save the revision in a separate transaction because the subsequent upload/copy
-      // action reads the most current revision from the database. If it is done in the same
-      // transaction with the enqueueing, the action might start running before the transaction is
-      // committed, due to Cloud Tasks not being transaction aware. The downside is that if for some
-      // reason the second transaction is rolled back, the revision update is not undone. But this
-      // should be fine since the next run will just increment the revision and start over.
-      tm().transact(
-              () ->
-                  RdeRevision.saveRevision(
-                      key.tld(), key.watermark(), key.mode(), input.getValue()));
+
       tm().transact(
               () -> {
+                PendingDeposit key = input.getKey();
+                int revision = input.getValue();
                 Registry registry = Registry.get(key.tld());
                 Optional<Cursor> cursor =
                     transactIfJpaTm(
@@ -305,25 +299,30 @@ public class RdeIO {
                 tm().put(Cursor.create(key.cursor(), newPosition, registry));
                 logger.atInfo().log(
                     "Rolled forward %s on %s cursor to %s.", key.cursor(), key.tld(), newPosition);
+                RdeRevision.saveRevision(key.tld(), key.watermark(), key.mode(), input.getValue());
                 // Enqueueing a task is a side effect that is not undone if the transaction rolls
                 // back. So this may result in multiple copies of the same task being processed.
                 // This is fine because the RdeUploadAction is guarded by a lock and tracks progress
-                // by cursor. The BrdaCopyAction writes a file to GCS, which is an atomic action.
+                // by cursor. The BrdaCopyAction writes a file to GCS, which is an atomic action. It
+                // is also guarded by a cursor to not run before the cursor is updated. We also
+                // include a delay to minimize the chance that the enqueued job executes before the
+                // transaction is committed, which triggers a retry.
                 if (key.mode() == RdeMode.FULL) {
                   cloudTasksUtils.enqueue(
                       RDE_UPLOAD_QUEUE,
-                      cloudTasksUtils.createPostTask(
+                      cloudTasksUtils.createPostTaskWithDelay(
                           RdeUploadAction.PATH,
                           Service.BACKEND.getServiceId(),
                           ImmutableMultimap.of(
                               RequestParameters.PARAM_TLD,
                               key.tld(),
                               RdeModule.PARAM_PREFIX,
-                              options.getJobName() + '/')));
+                              options.getJobName() + '/'),
+                          ENQUEUE_DELAY));
                 } else {
                   cloudTasksUtils.enqueue(
                       BRDA_QUEUE,
-                      cloudTasksUtils.createPostTask(
+                      cloudTasksUtils.createPostTaskWithDelay(
                           BrdaCopyAction.PATH,
                           Service.BACKEND.getServiceId(),
                           ImmutableMultimap.of(
@@ -332,7 +331,8 @@ public class RdeIO {
                               RdeModule.PARAM_WATERMARK,
                               key.watermark().toString(),
                               RdeModule.PARAM_PREFIX,
-                              options.getJobName() + '/')));
+                              options.getJobName() + '/'),
+                          ENQUEUE_DELAY));
                 }
               });
     }
--- a/core/src/main/java/google/registry/beam/rde/RdeIO.java
+++ b/core/src/main/java/google/registry/beam/rde/RdeIO.java
@@ -271,10 +271,19 @@ public class RdeIO {
     @ProcessElement
     public void processElement(
         @Element KV<PendingDeposit, Integer> input, PipelineOptions options) {
+      PendingDeposit key = input.getKey();
+      // We need to save the revision in a separate transaction because the subsequent upload/copy
+      // action reads the most current revision from the database. If it is done in the same
+      // transaction with the enqueueing, the action might start running before the transaction is
+      // committed, due to Cloud Tasks not being transaction aware. The downside is that if for some
+      // reason the second transaction is rolled back, the revision update is not undone. But this
+      // should be fine since the next run will just increment the revision and start over.
+      tm().transact(
+              () ->
+                  RdeRevision.saveRevision(
+                      key.tld(), key.watermark(), key.mode(), input.getValue()));
       tm().transact(
               () -> {
-                PendingDeposit key = input.getKey();
-                int revision = input.getValue();
                 Registry registry = Registry.get(key.tld());
                 Optional<Cursor> cursor =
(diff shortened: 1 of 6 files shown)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyFS5A9AYOZv4c6eIaA:b-n4s609) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

I agree and it seems like the amount of delay doesn't have to be as long as the task in go/r3pr/1554, based on the logs of tasks that trigger `RdeUploadAction`
</blockquote></details>

The cursor only records the time that a certain should run next. It does not concern itself with revisions as far as I can tell. In fact, with BRDA, the copy action is not guarded by a lock. So if the revision advanced and the copy job enqueued, but the cursor advancement failed, the entire process will run again. Should there be a new revision for the new run? With this PR, there will be. If we add a delay instead, there won't be because the revision is rolled back. It seems more reasonable to have a new revision so to not override whatever is already copied to the escrow provider by the copy job. WDYT?

Also I have to say that the lack of transactionality is starting to bug me more and more. Maybe we should think of a way to guarantee it? I don't have anything on top of my head, but all the scrutinizing over edge cases is bad as we are always making some compromise, not to mention the mental toll :P

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyHkl7B9NFkLB6Cd6aA:bma9vmy) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

The cursor only records the time that a certain should run next. It does not concern itself with revisions as far as I can tell. In fact, with BRDA, the copy action is not guarded by a lock. So if the revision advanced and the copy job enqueued, but the cursor advancement failed, the entire process will run again. Should there be a new revision for the new run? With this PR, there will be. If we add a delay instead, there won't be because the revision is rolled back. It seems more reasonable to have a new revision so to not override whatever is already copied to the escrow provider by the copy job. WDYT?

Also I have to say that the lack of transactionality is starting to bug me more and more. Maybe we should think of a way to guarantee it? I don't have anything on top of my head, but all the scrutinizing over edge cases is bad as we are always making some compromise, not to mention the mental toll :P
</blockquote></details>

Not to belabor the point, but you are making a case for rolling our own task-queues ;-)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyIAF2WC_yxM_vm2aO8:b-xuakh5) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*
> belabor
Yes, I am leaning more and more toward that given the subtleties that we are discovering more and more.

But I also recognize that this is not a small undertaking and probably require a design doc, and that we'd probably be reinventing the wheel somewhere. I wonder if we should raise this question with Cloud Tasks. I'd bet we are not the first ones to want the transactionality back.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyILOkg9mVgfXMtZ2Ws:b-8dt336) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

> belabor
Yes, I am leaning more and more toward that given the subtleties that we are discovering more and more.

But I also recognize that this is not a small undertaking and probably require a design doc, and that we'd probably be reinventing the wheel somewhere. I wonder if we should raise this question with Cloud Tasks. I'd bet we are not the first ones to want the transactionality back.
</blockquote></details>

Isn't this code called in the RDE pipeline that generates the data escrow files?
The cursors would be the per-tld RDE_STAGING cursors and the revision number
would be used in the generated file names.

With the new code, if revision is updated and cursor is not, we will rerun RDE staging and 
the revision would have a gap. In one scenario, the very first copy we send to IronMountain
and Icann would be R1 instead of R0.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyINjdrCkGCLXc3ibvJ:b-exrdxa) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

Isn't this code called in the RDE pipeline that generates the data escrow files?
The cursors would be the per-tld RDE_STAGING cursors and the revision number
would be used in the generated file names.

With the new code, if revision is updated and cursor is not, we will rerun RDE staging and 
the revision would have a gap. In one scenario, the very first copy we send to IronMountain
and Icann would be R1 instead of R0.
</blockquote></details>

That's true for RDE, but for BRDA it's the opposite, because BRDA copy is not guarded by a cursor. So if the BRDA revision is updated and the copy job is enqueued, R0 will be sent. If they cursor updates fails, the job reruns and R1 will be generated and sent.

If we add a delay instead and keep everything in one transaction, nothing will be updated but the copy job will still be enqueued. If the revision is unset, we then have a copy job that keep retrying. If a revision is already set, then the copy job would be copying the old (un-updated revision), which probably is not the right thing to do.

So it looks like if we want to be exact, we need to put the BRDA copy action behind a cursor and do the delayed enqueueing in one transaction.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyIYlTL8GDzB--ZXiWp:b-jzifgv) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

That's true for RDE, but for BRDA it's the opposite, because BRDA copy is not guarded by a cursor. So if the BRDA revision is updated and the copy job is enqueued, R0 will be sent. If they cursor updates fails, the job reruns and R1 will be generated and sent.

If we add a delay instead and keep everything in one transaction, nothing will be updated but the copy job will still be enqueued. If the revision is unset, we then have a copy job that keep retrying. If a revision is already set, then the copy job would be copying the old (un-updated revision), which probably is not the right thing to do.

So it looks like if we want to be exact, we need to put the BRDA copy action behind a cursor and do the delayed enqueueing in one transaction.
</blockquote></details>

Yes. I think we can add a delay here (and in RdeStagingReducer) and update BrdaCopyAction in another PR.
As of now, it is just some annoying error messages and nothing truly bad, right.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyJ86tADSuQ-fl75WsG:b-iy3rt2) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

Yes. I think we can add a delay here (and in RdeStagingReducer) and update BrdaCopyAction in another PR.
As of now, it is just some annoying error messages and nothing truly bad, right. 
</blockquote></details>

Yes, for now it's just annoyance. I added the delay and also the check for cursor in `BrdaCopyAction`. I don't think a lock is necessary as the copy to GCS is atomic. PTAL.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyNb0gz0AkSEcXcmiyy:b-hcujpj) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

Yes, for now it's just annoyance. I added the delay and also the check for cursor in `BrdaCopyAction`. I don't think a lock is necessary as the copy to GCS is atomic. PTAL.
</blockquote></details>

Have you pushed a new revision? I only see r1 here.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1556)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[core/src/main/java/google/registry/beam/rde/RdeIO.java, line 313 at r1](https://reviewable.io/reviews/google/nomulus/1556#-MyDVOSi9DGayVoZzoB5:-MyO0tFX4r69CGpr6g8c:b-32ck94) ([raw file](https://github.com/google/nomulus/blob/e0972580ee4426b67920843680de84e991940874/core/src/main/java/google/registry/beam/rde/RdeIO.java#L313)):*

Have you pushed a new revision? I only see r1 here.
</blockquote></details>

I think GitHub is somehow broked by the outage this morning. If you go to my branch you can see it's updated: https://github.com/jianglai/nomulus/tree/brda-copy.

PR #1557, opened 2022-03-15
Track and replay Transaction table gaps

mindhog:
Id gaps in the Transaction table can be the result of a transactions committed
out of order.  To deal with this, keep track of gaps for up to five minutes
and check to see if they've been back-filled prior to applying the next batch
of transactions during reply.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1557)

### DIFF
--- a/core/src/main/java/google/registry/model/EntityClasses.java
+++ b/core/src/main/java/google/registry/model/EntityClasses.java
@@ -40,6 +40,7 @@ import google.registry.model.rde.RdeRevision;
 import google.registry.model.registrar.Registrar;
 import google.registry.model.registrar.RegistrarContact;
 import google.registry.model.replay.LastSqlTransaction;
+import google.registry.model.replay.ReplayGap;
 import google.registry.model.reporting.HistoryEntry;
 import google.registry.model.server.Lock;
 import google.registry.model.server.ServerSecret;
@@ -86,6 +87,7 @@ public final class EntityClasses {
           Registrar.class,
           RegistrarContact.class,
           Registry.class,
+          ReplayGap.class,
           ServerSecret.class);
 
   private EntityClasses() {}
--- a/core/src/main/java/google/registry/model/replay/ReplayGap.java
+++ b/core/src/main/java/google/registry/model/replay/ReplayGap.java
@@ -27,9 +27,9 @@ import org.joda.time.DateTime;
  * Tracks gaps in transaction ids when replicating from SQL to datastore.
  *
  * <p>SQL -&gt; DS replication uses a Transaction table indexed by a SEQUENCE column, which normally
- * increments monotonically for each committed transaction.  Gaps in this sequence can occur when a
+ * increments monotonically for each committed transaction. Gaps in this sequence can occur when a
  * transaction is rolled back or when a transaction has been initiated but not committed to the
- * table at the time of a query.  To protect us from the latter scenario, we need to keep track of
+ * table at the time of a query. To protect us from the latter scenario, we need to keep track of
  * these gaps and replay any of them that have been filled in since we processed their batch.
  */
 @DeleteAfterMigration
--- a/core/src/main/java/google/registry/model/replay/ReplayGap.java
+++ b/core/src/main/java/google/registry/model/replay/ReplayGap.java
@@ -1,0 +1,60 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1557)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @mindhog and @weiminyu)

___
*[core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1557#-MyDXL8SBt3l7TCgizeg:-MyDXL8SBt3l7TCgizeh:b-s8wmb4) ([raw file](https://github.com/google/nomulus/blob/d98f8b70c795308c1ed41e6be395d40c551619f4/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java#L79)):*
> ```Java
>    * milliseconds, after which, the gap record will be deleted.
>    */
>   public static final long MAX_GAP_RETENTION = 300000;
> ```

MAX_GAP_RETENTION_MILLIS or make var of type Duration?



_Code quote:_
```Java
MAX_GAP_RETENTION
```
___
*[core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java, line 115 at r1](https://reviewable.io/reviews/google/nomulus/1557#-MyDXf4_--Onr7LMlgHS:-MyDXf4_--Onr7LMlgHT:b-9sdp7x) ([raw file](https://github.com/google/nomulus/blob/d98f8b70c795308c1ed41e6be395d40c551619f4/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java#L115)):*
> ```Java
>                 // Unless we're doing a snapshot, we'll want to fill in any gaps in the transaction
>                 // log that have since become available before processing the next batch.
>                 if (!snapshotId.isPresent()) {
> ```

This seems problematic to me.

There can be missed transactions during the previous cron-invocation of the Replay action right? If so, during db comparison, we want to catch those.

Wouldn't it be sufficient if you call applyMissingTransaction with the snapshot? You may need to skip the purge of expired gap id if there is a snapshot.
 



_Code quote:_
```Java
if (!snapshotId.isPresent())
```

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1557)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java, line 79 at r1](https://reviewable.io/reviews/google/nomulus/1557#-MyDXL8SBt3l7TCgizeg:-MyDafK55akjzae8S-Fn:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d98f8b70c795308c1ed41e6be395d40c551619f4/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java#L79)):*

MAX_GAP_RETENTION_MILLIS or make var of type Duration?
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java, line 115 at r1](https://reviewable.io/reviews/google/nomulus/1557#-MyDXf4_--Onr7LMlgHS:-MyE21085_HmVJPy7fMR:bijho6n) ([raw file](https://github.com/google/nomulus/blob/d98f8b70c795308c1ed41e6be395d40c551619f4/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java#L115)):*

This seems problematic to me.

There can be missed transactions during the previous cron-invocation of the Replay action right? If so, during db comparison, we want to catch those.

Wouldn't it be sufficient if you call applyMissingTransaction with the snapshot? You may need to skip the purge of expired gap id if there is a snapshot.
 
</blockquote></details>

Yeah, you're right -- we want to capture everything now available in the snapshot transaction log.  Done.

I don't think we need to worry about purging expired gaps.  Even if the portion of the database sync that occurs after `replayAllTransactions`  takes more than 5 minutes (and `replicateToDatastore` doesn't get called during that time) we'll still end up trying to apply the missing transactions before purging the gap records.

Also, since this is called in the same transaction as the batch retrieval, we can just set the snapshot id before calling it.

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1557)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java, line 180 at r2](https://reviewable.io/reviews/google/nomulus/1557#-MyE6h-f2NEHFQFdisWs:-MyEA-_22hW0ij2vjQ_-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ab545b76f90709df1f20792a1d5cdacc44d7a4e5/core/src/main/java/google/registry/model/replay/ReplicateToDatastoreAction.java#L180)):*

Move this line to the beginning of this method, in case the query time is a non-trivial portion of the retention time?
</blockquote></details>

Done.

PR #1558, opened 2022-03-15
Improve cache loading in Registries.java

weiminyu:
The loader for the TLD cache in Registries.java unnecessarily reads from
another cache when running with SQL, potentially triggering additional
database access. This code runs in the whois query path, and contributes
to the periodical high latency we are seeing in sandbox.

In crash with production data, this change reduces about 1 second in latency on cache miss.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1558)

### DIFF
--- a/core/src/main/java/google/registry/model/tld/Registries.java
+++ b/core/src/main/java/google/registry/model/tld/Registries.java
@@ -72,10 +72,11 @@ public final class Registries {
                         return Registry.getAll(tlds).stream()
                             .map(e -> Maps.immutableEntry(e.getTldStr(), e.getTldType()))
                             .collect(entriesToImmutableMap());
+                      } else {
+                        return tm().loadAllOf(Registry.class).stream()
+                            .map(e -> Maps.immutableEntry(e.getTldStr(), e.getTldType()))
+                            .collect(entriesToImmutableMap());
                       }
-                      return tm().loadAllOf(Registry.class).stream()
-                          .map(e -> Maps.immutableEntry(e.getTldStr(), e.getTldType()))
-                          .collect(entriesToImmutableMap());
                     }));
   }
 
--- a/core/src/main/java/google/registry/model/tld/Registries.java
+++ b/core/src/main/java/google/registry/model/tld/Registries.java
@@ -58,21 +58,22 @@ public final class Registries {
         () ->
             tm().doTransactionless(
                     () -> {
-                      ImmutableSet<String> tlds =
-                          tm().isOfy()
-                              ? auditedOfy()
-                                  .load()
-                                  .type(Registry.class)
-                                  .ancestor(getCrossTldKey())
-                                  .keys()
-                                  .list()
-                                  .stream()
-                                  .map(Key::getName)
-                                  .collect(toImmutableSet())
-                              : tm().loadAllOf(Registry.class).stream()
-                                  .map(Registry::getTldStr)
-                                  .collect(toImmutableSet());
-                      return Registry.getAll(tlds).stream()
+                      if (tm().isOfy()) {
+                        ImmutableSet<String> tlds =
+                            auditedOfy()
+                                .load()
+                                .type(Registry.class)
+                                .ancestor(getCrossTldKey())
+                                .keys()
+                                .list()
+                                .stream()
+                                .map(Key::getName)
+                                .collect(toImmutableSet());
+                        return Registry.getAll(tlds).stream()
+                            .map(e -> Maps.immutableEntry(e.getTldStr(), e.getTldType()))
+                            .collect(entriesToImmutableMap());
+                      }
+                      return tm().loadAllOf(Registry.class).stream()
                           .map(e -> Maps.immutableEntry(e.getTldStr(), e.getTldType()))
                           .collect(entriesToImmutableMap());
                     }));
--- a/core/src/main/java/google/registry/model/tld/Registries.java
+++ b/core/src/main/java/google/registry/model/tld/Registries.java
@@ -24,6 +24,7 @@ import static com.google.common.collect.Maps.filterValues;
 import static google.registry.model.CacheUtils.memoizeWithShortExpiration;
 import static google.registry.model.common.EntityGroupRoot.getCrossTldKey;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.CollectionUtils.entriesToImmutableMap;
 import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
@@ -38,6 +39,8 @@ import com.google.common.net.InternetDomainName;
 import com.googlecode.objectify.Key;
 import google.registry.model.tld.Registry.TldType;
 import java.util.Optional;
+import java.util.stream.Stream;
+import javax.persistence.EntityManager;
 
 /** Utilities for finding and listing {@link Registry} entities. */
 public final class Registries {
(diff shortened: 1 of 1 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE917q1hkup9JxznG7:-MyE917q1hkup9JxznG8:bmx7nos) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L75)):*
> ```Java
>                             .map(e -> Maps.immutableEntry(e.getTldStr(), e.getTldType()))
>                             .collect(entriesToImmutableMap());
>                       }
> ```

Adding an ` else {` here might make the rest of this slightly more clear.
___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE93sp15GpgbURgjN4:-MyE93sp15GpgbURgjN5:bi0pmw) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L76)):*
> ```Java
>                             .collect(entriesToImmutableMap());
>                       }
>                       return tm().loadAllOf(Registry.class).stream()
> ```

Why load up the entirety of the row?  Why not do something like this?  (pseudocode)

```
SELECT tldStr, tldType FROM Tld;
```

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @mindhog and @weiminyu)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE93sp15GpgbURgjN4:-MyEA5u70i0-Ealt0ur4:b-vx4k6o) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L76)):*

Why load up the entirety of the row?  Why not do something like this?  (pseudocode)

```
SELECT tldStr, tldType FROM Tld;
```
</blockquote></details>

If we just do this it should be quite fast.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 75 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE917q1hkup9JxznG7:-MyEA4qD3e4GDmnOMiga:be60xvn) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L75)):*

Adding an ` else {` here might make the rest of this slightly more clear.
</blockquote></details>

Done.

Personally I don't like else if the IF block has a return but...
___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE93sp15GpgbURgjN4:-MyEAHWlDySOSZ-AepWv:b-ofq62u) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L76)):*

If we just do this it should be quite fast.
</blockquote></details>

Entity query is type-safe and in this case is already fast.

With text query you have to figure out the property name, and breaking changes can only be discovered after running tests.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mindhog and @weiminyu)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE93sp15GpgbURgjN4:-MyEC88xDymeIm9ZXPGy:b-9l824f) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L76)):*

Entity query is type-safe and in this case is already fast.

With text query you have to figure out the property name, and breaking changes can only be discovered after running tests.
</blockquote></details>

How fast is it?  It would be substantially faster still if it only loaded up the 2 columns it needed, rather than all of the columns that currently exist on `Registry` (which is a large number and contains many arrays).

As for breaking changes, we're highly unlikely to ever rename these fields/columns, and even if we did for some reason, you'd still need to run all the tests anyway.

It seems to me that this is really simple a "select two columns from this table" query.  Like, a one-liner.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE93sp15GpgbURgjN4:-MyER4oc3qABm1E7wzLR:bkilckk) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L76)):*

How fast is it?  It would be substantially faster still if it only loaded up the 2 columns it needed, rather than all of the columns that currently exist on `Registry` (which is a large number and contains many arrays).

As for breaking changes, we're highly unlikely to ever rename these fields/columns, and even if we did for some reason, you'd still need to run all the tests anyway.

It seems to me that this is really simple a "select two columns from this table" query.  Like, a one-liner.
</blockquote></details>

Done.

Not a one-liner. 8-)

Note that there is another cache in Registry.java with same expiry that caches full tld entities. It seems to me that it should be merged with this one.  If we do that, we need to revert to entity query.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @mindhog)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 76 at r1](https://reviewable.io/reviews/google/nomulus/1558#-MyE93sp15GpgbURgjN4:-MyEhaaz9eFgLmLDa6SS:bx3glbe) ([raw file](https://github.com/google/nomulus/blob/d2cf7850d0959d07bd902d7a720ef447372c1bd2/core/src/main/java/google/registry/model/tld/Registries.java#L76)):*

Done.

Not a one-liner. 8-)

Note that there is another cache in Registry.java with same expiry that caches full tld entities. It seems to me that it should be merged with this one.  If we do that, we need to revert to entity query.
</blockquote></details>

It's not entirely clear to me that the two can be combined.  This cache is a lightweight cache of all TLDs that exist along with their types, but just those two pieces of info, whereas the other one is a cache of entire individual `Registry` objects (but not necessarily all of them simultaneously).  Note that we have _a lot_ of Registry objects corresponding to unlaunched/prober/test-data TLDs, which are hardly ever loaded, and it doesn't particularly make sense to always keep all of them in memory.

So we do have a use case for "Tell me which TLDs exist" independently of "Give me all of the data about all of the TLDs that exist".  A very common use case is to simply check a TLD for existence, but otherwise not do anything else with any of the information on the `Registry` object.  So this lightweight Map is perfect for that.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558#-:-MyEnopy0k6uRi9tNGOJ:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @mindhog)

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1558)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/tld/Registries.java, line 79 at r3](https://reviewable.io/reviews/google/nomulus/1558#-MyEnxGPCMR0wrKOtevC:-MyFRLwH4YUcWQEcwd8C:b-f0zijk) ([raw file](https://github.com/google/nomulus/blob/e5133c24137a3567c1f30551885209a99bf748e4/core/src/main/java/google/registry/model/tld/Registries.java#L79)):*

This entire `else` could be a single multi-line statement.  You can inline `entityManager` and `resultStream`; they're both only being used once.
</blockquote></details>

That's due to the quirk of Java generics. Query.getResultStream() returns a raw Stream. We must convert it to Stream<?> for the compiler to
deduce the type of  .map(e -> ((Object[]) e).  Alternative is using (Stream<?>) to avoid the intermediate variable resultStream.

PR #1560, opened 2022-03-16
Add domain repo ID and token type indexes to AllocationToken table

CydeWeys:
These are useful for the purposes of filtering by one-time/multi-use tokens, and
for determining which one-time tokens have been used (and if so, for which
domain).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1560)


CydeWeys (commented):
Analytics queries for now, but also supporting possible future registrar/admin console tables.  If we're going to have a way to manage tokens in the console (which would be nice), we'll want to be able to filter them by token type and used/unused.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1560#-:-MyJ-e683OkmG6BR0i_5:b-qfqtk7)* status: 0 of 7 files reviewed, all discussions resolved

PR #1562, opened 2022-03-18
Bump flogger and beam dependency versions

gbrodman:
Beam 2.34.0 -> 2.37.0
Flogger 0.7.3 -> 0.7.4
Bouncycastle 1.61.0 -> 1.67.0

Intellij keeps getting confused about which version of Flogger we're
bringing in. Even though we had previously locked Flogger to 0.7.3, for
some reason it was still bringing in the Beam transitive dependency of
0.6.0 which was causing the a bunch of class initialization errors.

Bumping Beam to 2.34.0 bumps the transitive dependency to 0.7.4 so we
can always use that.

There were some awkward interactions with the Bouncycastle dependency too, so we bumped that to the same version as the Beam dependency (1.67.0) 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1562)


weiminyu (approved):
> Reviewed 114 of 114 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1562)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1564, opened 2022-03-21
Use TLS v1.3 explicitly in RDE reporting

gbrodman:
The default Java 1.8 TLS version is 1.2 which isn't supported by the
ICANN upload site.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1564)

### DIFF
--- a/core/src/main/java/google/registry/request/Modules.java
+++ b/core/src/main/java/google/registry/request/Modules.java
@@ -29,6 +29,8 @@ import dagger.Module;
 import dagger.Provides;
 import java.net.HttpURLConnection;
 import javax.inject.Singleton;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
 
 /** Dagger modules for App Engine services and other vendor classes. */
 public final class Modules {
@@ -49,7 +51,16 @@ public final class Modules {
   public static final class UrlConnectionServiceModule {
     @Provides
     static UrlConnectionService provideUrlConnectionService() {
-      return url -> (HttpURLConnection) url.openConnection();
+      return url -> {
+        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+        if (connection instanceof HttpsURLConnection) {
+          HttpsURLConnection httpsConnection = (HttpsURLConnection) connection;
+          SSLContext tls13Context = SSLContext.getInstance("TLSv1.3");
+          tls13Context.init(null, null, null);
+          httpsConnection.setSSLSocketFactory(tls13Context.getSocketFactory());
+        }
+        return connection;
+      };
     }
   }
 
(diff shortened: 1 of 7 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1564)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rde/RdeReporter.java, line 86 at r1](https://reviewable.io/reviews/google/nomulus/1564#-MymKUnv5KUl9laVxcVe:-MymKUnv5KUl9laVxcVf:bdpcjem) ([raw file](https://github.com/google/nomulus/blob/ca0bc47488813d8d6085acb7b7f6b9271d25b3c1/core/src/main/java/google/registry/rde/RdeReporter.java#L86)):*
> ```Java
>             () -> {
>               HttpURLConnection connection = urlConnectionService.createConnection(url);
>               if (connection instanceof HttpsURLConnection) {
> ```

It looks like calling `SSLContext.getInstance(...)` just sets the default, the SSL engine should still be able to negotiate a lower version if need be.

https://stackoverflow.com/questions/18237569/httpclient-supporting-multiple-tls-protocols

Maybe this should be the default behavior for our HTTP client. Also all of our HTTP outbound connection should be HTTPS. We should make that the default (and only) behavior.



_Code quote:_
```Java
if (connection instanceof HttpsURLConnection) {
```

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1564)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[core/src/main/java/google/registry/rde/RdeReporter.java, line 86 at r1](https://reviewable.io/reviews/google/nomulus/1564#-MymKUnv5KUl9laVxcVe:-MyoMdMS0oOtXuvAkGYR:b-ulqdh) ([raw file](https://github.com/google/nomulus/blob/ca0bc47488813d8d6085acb7b7f6b9271d25b3c1/core/src/main/java/google/registry/rde/RdeReporter.java#L86)):*

I'm not sure what you mean in the first part. The link you posted seems to suggest that providing an argument to `getInstance` simply sets the default level which would then be negotiated if the server doesn't support TLS v1.3. The default TLS level in Java 8 is 1.2 (which won't work).

Maybe they should be all HTTPS but they aren't. Specifically, the TMCH CRL isn't, and cannot be (I tried at first to make them all HTTPS because it would be better, but it didn't work). 


</blockquote></details>

Sorry if I wasn't being clear. I meant to say that 1) we should default to TLS 1.3 for all HTTPS calls (since it downgrades to 1.2 or lower if 1.3 is not available), and 2) we should only do HTTPS calls.

Good point on CRL. They are signed so the transport media does not have to be secure, plus you if you'll get in to an infinite loop if use HTTPS because CRL itself is needed to establish an HTTPS connection. But I still think we should make the default behavior to be HTTPS only, and have an override when we have to use HTTP.

jianglai (approved):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1564)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1567, opened 2022-03-23
Ignore trivial differences when comparing DB

weiminyu:
Some data difference are due to entity model differences and also
harmless. We should igore them when comparing Datastore and SQL.

This PR ignores the following diffs:
- null vs empty collection
- the empty string in Address.stree field, which is a list


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1567)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1568, opened 2022-03-23
Add renewal columns in BillingRecurrence

rachelguan:
Before making changes to the renewal process for anchor tenants and internal registration, there needs to be a schema change added and deployed before any code change. This PR is the first one for the new feature: **Improve Anchor Tenants and Internal Registrations renewal process**.  

The following columns will be added: 

- `renewal_price_behavior`: expected behaviors include `STANDARD_PRICE`, `DEFAULT_PRICE` and `RENEWAL_PRICE` (the price specified under `renewal_price_amount` with the currency under `renewal_price_currency`), 
- `renewal_price_currency` : currency of the renewal price,
- `renewal_price_amount`: the numeric value of the renewal price  

There's three tables related to billing: `BillingEvent`, `BillingRecurrence`, and `BillingCancellation`. The columns should be added to `BillingRecurrence` since they are used for renewal which ideally shouldn't be a one time event. The columns could be added to `BillingEvent` which is for one time events such as creating the domain initially along with cost; this additional info could be useful but not necessary at the moment . There's no reason to add to `BillingCancellation`; once it's cancelled, the deal will be gone. 

`DomainRenewFlow` is the flow for both manually and automatically renewing a domain and the `BillingRecurrence` entity can be accessed for both renewal cases.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1568)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myvznu_6_YKCJzWnKCC:-Myvznu_6_YKCJzWnKCD:b-ims8dx) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*
> ```TSQL
> 
> alter table "BillingRecurrence" add column "always_renew_at_standard_price" boolean default false;
> ```

These names are kind of long.  It seems like you could drop the "always_" prefix without impacting comprehensibility, right?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx0rdI9GpIb138V0as:-Myx0rdI9GpIb138V0at:bz1g6mv) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*
> ```TSQL
> 
> alter table "BillingRecurrence" add column "always_renew_at_standard_price" boolean default false;
> ```

Two separate columns with boolean values doesn't seem like the right call, given that having `true` for both wouldn't make any sense.  A single column that is an enum would make more sense.  Something like:

`renewal_price_behavior` with possible values of `NULL`, `"STANDARD_PRICE"`, and `"INITIAL_PRICE"`.

I don't see how `"INITIAL_PRICE"` works here given that you aren't actually recording anywhere what the initial price *was*.  You need to add data onto the Recurrence event that stores this.

So there's really three possible states:
* Default behavior (consult premium list for renewal price)
* Renew at standard price (which consults the `Registry` entity to see what the standard price for this TLD is)
* Renew at _this price_ where the price is stored as data on the Recurrence (which can be set to the initial price but might also be something else, depending).
___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 17 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx1b4J8SupNlT7lqs0:-Myx1b4KDh2HEp6T29QU:brxvo6w) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L17)):*
> ```TSQL
> alter table "BillingRecurrence" add column "always_renew_at_standard_price" boolean default false;
> alter table "BillingRecurrence" add column "always_renew_at_initial_price" boolean default false;
> alter table "BillingEvent" add column "always_renew_at_standard_price" boolean default false;
> ```

These column names don't really make sense on `BillingEvent`, as `always_renew_*` implies having an effect on future behavior, but the `BillingEvent` `OneTime`s are strictly a record of past actions.  My `renewal_price_behavior` suggestion from above makes a little more sense.

Note here that, unlike `BillingRecurrence`, we wouldn't separate need to record the price renewed at, as that's already recorded on the `BillingEvent` itself.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, @mindhog, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx0rdI9GpIb138V0as:-MyxS_vk3bVz8P2dZ_16:bayom54) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*

Two separate columns with boolean values doesn't seem like the right call, given that having `true` for both wouldn't make any sense.  A single column that is an enum would make more sense.  Something like:

`renewal_price_behavior` with possible values of `NULL`, `"STANDARD_PRICE"`, and `"INITIAL_PRICE"`.

I don't see how `"INITIAL_PRICE"` works here given that you aren't actually recording anywhere what the initial price *was*.  You need to add data onto the Recurrence event that stores this.

So there's really three possible states:
* Default behavior (consult premium list for renewal price)
* Renew at standard price (which consults the `Registry` entity to see what the standard price for this TLD is)
* Renew at _this price_ where the price is stored as data on the Recurrence (which can be set to the initial price but might also be something else, depending).
</blockquote></details>

yea, the case of having `true` for both wouldn't be assigned but technically could happen, which wouldn't be a safe practice. There are three states and ENUM could do the job which also would allow the possibility to add more states in the future. As you mentioned, the price (a number) would have to be stored as well for internal registrations. 

Is there a reason why initial price should not be stored in `BilllingRecurrence`?  Regarding the initial price, we could try to retrieve the initial fee from the database  using the the original one time billing event when the domain was created.  Or add another column to `BilllingRecurrence` to store initial price. It could potentially be used for more complicated discount system in the future (maybe some day for standard users, renewal , initial price for another year, hypothetically speaking). 

I think the word "this" seems ambiguous ..if we want to expand this ENUM in the future, we could potentially have a slit of price categories and then "this".

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myvznu_6_YKCJzWnKCC:-MyxZKE12NIEp8ewqHpe:b-d2dr3t) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*

These names are kind of long.  It seems like you could drop the "always_" prefix without impacting comprehensibility, right?
</blockquote></details>

that is indeed a good point..if these two columns are being kept, then "always" would go away. Looks like the condition would become of one of the states in an new ENUM.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx0rdI9GpIb138V0as:-Myxe7BbE0Be_bTyL4Ff:bx6jjkq) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*

yea, the case of having `true` for both wouldn't be assigned but technically could happen, which wouldn't be a safe practice. There are three states and ENUM could do the job which also would allow the possibility to add more states in the future. As you mentioned, the price (a number) would have to be stored as well for internal registrations. 

Is there a reason why initial price should not be stored in `BilllingRecurrence`?  Regarding the initial price, we could try to retrieve the initial fee from the database  using the the original one time billing event when the domain was created.  Or add another column to `BilllingRecurrence` to store initial price. It could potentially be used for more complicated discount system in the future (maybe some day for standard users, renewal , initial price for another year, hypothetically speaking). 

I think the word "this" seems ambiguous ..if we want to expand this ENUM in the future, we could potentially have a slit of price categories and then "this". 


</blockquote></details>

for the number that should be stored, i think the word rate would be more suitable than price..maybe it still depends on the industry. The total price takes the number of years into consideration. I'm not sure if I could assume the rate would always be an integer so I'm gonna use float8 for now.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx0rdI9GpIb138V0as:-MyxgB-I1EuciIiXNqy6:b21ffqh) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*

for the number that should be stored, i think the word rate would be more suitable than price..maybe it still depends on the industry. The total price takes the number of years into consideration. I'm not sure if I could assume the rate would always be an integer so I'm gonna use float8 for now.  
</blockquote></details>

actually I would use the word price since it's used everywhere in the codebase.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 121 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1L3FK1tZOsgCeXr6L:-Mz1L3FK1tZOsgCeXr6M:by1zyvj) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/schema/nomulus.golden.sql#L121)):*
> ```TSQL
>     recurrence_time_of_year text,
>     renewal_price_type text,
>     initial_price double precision
> ```

This is the wrong type (and it's also missing the currency).  See other examples in our codebase for how prices are stored in the DB schema.  I think you're going to need two columns here, one a decimal (NOT a float) and the other a string to store the 3-letter currency code.

Note that, as a general rule, using a floating point type to store a monetary value is a code smell.  See e.g.: https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio
___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx0rdI9GpIb138V0as:-Mz1M3Hr5rdv4Vnlskpy:b-aj42cr) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*

actually I would use the word price since it's used everywhere in the codebase. 
</blockquote></details>

`renewal_price_behavior` to me more accurately communicates what this field is actually doing vs `renewal_price_type`.  The word `type` is just a little bit generic -- people might confuse it for the history entry's `type` (which, keep in mind, every billing event is parented directly on a history entry): https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/reporting/HistoryEntry.java;l=93
___
*[db/src/main/resources/sql/flyway/V115__add_renewal_columns_to_billing_event.sql, line 16 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1LYaM5f98tdfIK84w:-Mz1LYaM5f98tdfIK84x:bi0ozd2) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/flyway/V115__add_renewal_columns_to_billing_event.sql#L16)):*
> ```TSQL
> 
> alter table "BillingRecurrence" add column renewal_price_type text;
> alter table "BillingRecurrence" add column initial_price float8;
> ```

`renewal_price` makes a lot more sense to me than `initial_price` here, as it supports the use case of changing the renewal price to be something other than the initial price, and does a better job of communicating the intent of why we're storing this data and what it's being used for.  If you really ever need to find the initial price, it's on the `BillingEvent` (`OneTime`) entity with type `DOMAIN_CREATE` for that domain.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 121 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1L3FK1tZOsgCeXr6L:-Mz1PsYp-f54dT-xI2ni:b-3qjtqz) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/schema/nomulus.golden.sql#L121)):*

This is the wrong type (and it's also missing the currency).  See other examples in our codebase for how prices are stored in the DB schema.  I think you're going to need two columns here, one a decimal (NOT a float) and the other a string to store the 3-letter currency code.

Note that, as a general rule, using a floating point type to store a monetary value is a code smell.  See e.g.: https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio
</blockquote></details>

noted and changed it. 
___
*[db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql, line 15 at r1](https://reviewable.io/reviews/google/nomulus/1568#-Myx0rdI9GpIb138V0as:-Mz1Pis9-k-hh2V2iyAH:b4hd1pq) ([raw file](https://github.com/google/nomulus/blob/08362dded4a6ea004552bfdb07f774215bf8a695/db/src/main/resources/sql/flyway/V115__add_renewal_boolean_flags_to_billing_event.sql#L15)):*

`renewal_price_behavior` to me more accurately communicates what this field is actually doing vs `renewal_price_type`.  The word `type` is just a little bit generic -- people might confuse it for the history entry's `type` (which, keep in mind, every billing event is parented directly on a history entry): https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/reporting/HistoryEntry.java;l=93
</blockquote></details>

ok changed it to 'behavior'; 
___
*[db/src/main/resources/sql/flyway/V115__add_renewal_columns_to_billing_event.sql, line 16 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1LYaM5f98tdfIK84w:-Mz1SSQU97rCpeQS1wek:b-o53deu) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/flyway/V115__add_renewal_columns_to_billing_event.sql#L16)):*

`renewal_price` makes a lot more sense to me than `initial_price` here, as it supports the use case of changing the renewal price to be something other than the initial price, and does a better job of communicating the intent of why we're storing this data and what it's being used for.  If you really ever need to find the initial price, it's on the `BillingEvent` (`OneTime`) entity with type `DOMAIN_CREATE` for that domain.
</blockquote></details>

noted. Originally, I was actually thinking,,,since we the standard price for AT is not gonna be stored; we could just get the initial price which would be the renewal price for internal registrations from the one time `BillingEvent` entity with type `DOMAIN_CREATE` for the domain instead of storing them.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 121 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1L3FK1tZOsgCeXr6L:-Mz1Ux_10kvdwgn5S8nE:b-khvhp2) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/schema/nomulus.golden.sql#L121)):*

noted and changed it. 
</blockquote></details>

I think `renewal_price_amount` makes more sense here.  That way it parallels the related fields on `BillingEvent`, which are:

```
        cost_amount numeric(19, 2),
        cost_currency text,
```
___
*[db/src/main/resources/sql/flyway/V115__add_renewal_columns_to_billing_event.sql, line 16 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1LYaM5f98tdfIK84w:-Mz1VR813T-wHoSvLV6D:b-5akia4) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/flyway/V115__add_renewal_columns_to_billing_event.sql#L16)):*

noted. Originally, I was actually thinking,,,since we the standard price for AT is not gonna be stored; we could just get the initial price which would be the renewal price for internal registrations from the one time `BillingEvent` entity with type `DOMAIN_CREATE` for the domain instead of storing them. 
</blockquote></details>

The thing is, that's not the most efficient operation to do, and crucially, because `BillingEvent`s are immutable, you could not change the future renewal price even if it were necessary to do so (because it's _also_ a record of past events).

By recording the future renewal price here, it divorces it from whatever happened in the past, so that it's possible to change it.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 121 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1L3FK1tZOsgCeXr6L:-Mz1VxBN7sPJaSfHTeAw:b-xmnnyg) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/schema/nomulus.golden.sql#L121)):*

I think `renewal_price_amount` makes more sense here.  That way it parallels the related fields on `BillingEvent`, which are:

```
        cost_amount numeric(19, 2),
        cost_currency text,
```
</blockquote></details>

And more generally, the first pattern here makes more sense the the second pattern:

```
abc_xyz_foo
abc_xyz_bar
abc_xyz_baz
```

```
abc_xyz_foo
abc_xyz_bar
abc_xyz
```

In the second one, it's not clear what the third field is, because unlike the other two, it's not named.

CydeWeys (commented):
And it just occurs to me that, while the `BillingRecurrence` is the obvious and intuitive choice for where to store the cost of autorenewal, we're also going to need to load and use this for the cost of a manual renewal.  I think that's OK (note the existence of the`DomainContent.autorenewBillingEvent` pointer), but it probably needs to be added to the design docs and such.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568#-:-Mz1X4KW2svPG5v7OVHJ:bdr9dd4)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, @rachelguan, and @weiminyu)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/schema/nomulus.golden.sql, line 121 at r3](https://reviewable.io/reviews/google/nomulus/1568#-Mz1L3FK1tZOsgCeXr6L:-Mz1XJDR5OgU9Xy6dZ93:b-xks35k) ([raw file](https://github.com/google/nomulus/blob/23eead5881df35a72311c0c381b9d4139d743f0c/db/src/main/resources/sql/schema/nomulus.golden.sql#L121)):*

And more generally, the first pattern here makes more sense the the second pattern:

```
abc_xyz_foo
abc_xyz_bar
abc_xyz_baz
```

```
abc_xyz_foo
abc_xyz_bar
abc_xyz
```

In the second one, it's not clear what the third field is, because unlike the other two, it's not named.
</blockquote></details>

i noticed that too.. thank you for the suggestion.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, @rachelguan, and @weiminyu)

___
*[db/src/main/resources/sql/flyway.txt, line 115 at r5](https://reviewable.io/reviews/google/nomulus/1568#-Mz1yLl0CEi7xebr6WP8:-Mz1yLl0CEi7xebr6WP9:b-ki9sgy) ([raw file](https://github.com/google/nomulus/blob/e993a0cb1dff4cf2a7bf93caccdf1b65481bfb33/db/src/main/resources/sql/flyway.txt#L115)):*
> ```help
> V113__add_host_missing_indexes.sql
> V114__add_allocation_token_indexes.sql
> V115__add_renewal_columns_to_billing_event.sql
> ```

This filename isn't quite right anymore.  It should be `billing_recurrence`, not `billing_event`

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568#-:-Mz1yWvJ9ZvxtWTTYlpM:bnfp4nl)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @jianglai, @mindhog, @rachelguan, and @weiminyu)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @jianglai, @mindhog, and @weiminyu)

___
*[db/src/main/resources/sql/flyway.txt, line 115 at r5](https://reviewable.io/reviews/google/nomulus/1568#-Mz1yLl0CEi7xebr6WP8:-Mz28JXXBf_E0IA-Qdeq:bblcvru) ([raw file](https://github.com/google/nomulus/blob/e993a0cb1dff4cf2a7bf93caccdf1b65481bfb33/db/src/main/resources/sql/flyway.txt#L115)):*

This filename isn't quite right anymore.  It should be `billing_recurrence`, not `billing_event`
</blockquote></details>

I used billing_event originally because both `BillingRecurrence` and `BillingEvent` tables are in `billing/BillingEvent.java`. I'm indeed gonna change it to reflect the exact table that will be modified in db. Thanks!

CydeWeys (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568#-:-Mz2C1Xo2j0RUdOfrq4C:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jianglai, @mindhog, and @weiminyu)

CydeWeys (commented):
Just to be clear though, recurring billing events exist to service autorenewals.  They're not otherwise currently used by manual renewals, which is why this is a change that needs to be thought through properly and documented.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1568#-:-Mz2CDFR0lGPwyGXKI1Q:b-ru3h43)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jianglai, @mindhog, and @weiminyu)

PR #1570, opened 2022-03-25
Terminate with exit code of 1 on format check fail

mindhog:
The format check script currently outputs "true" if there were files that need
reformatting and "false" if not, which is useful for gradle but less so for
other applications (notably commit hooks).  Terminate with an exit code of 1
if the format check fails.

TESTED: Tried this from both a pre-commit hook and from the gradle build.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1570)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1570)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1573, opened 2022-03-29
Add new columns to BillingEvent 

rachelguan:
After the schema change to `BillingRecurrence` (see [PR#1568](https://github.com/google/nomulus/pull/1568/files)), the corresponding code change could go in. The code changes happens in `BillingEvent`, where `BillingRecurrence` resides.  

The three renewal price behaviors  are in an `enum`, which are `DEFAULT`, `NONPREMIUM` and `SPECIFIED`. The price columns for internal registrations are `renewal_price_currency` and `renewal_price_amount`, and together they form `renewalPrice`.

There is an issue with the `Money` column. All the existing use cases have the amount and currency code as required columns. However, renewal price columns can have null values (either both contain values or both as null), and current method for `Money` construction does not support that.  `Money` construction happens via `JodaMoneyType`. In the override method `nullSafeGet`, `CurrencyUnit.of()` throws null currency unit which prevents the method to return `null`. `CurrencyUnit.of()` also handles unknown currency by throwing `IllegalCurrencyException`. If `CurrencyUnit.of()` returns a value, it would be a valid currency. The way to fix this is do a null check on the currency string before passing to `CurrencyUnit.of()`. 

 **tl;dr: corresponding schema change:** 

- [PR#1568: Add renewal columns in BillingRecurrence](https://github.com/google/nomulus/pull/1568)
- [PR#1575: Add default value to renewal_price_behavior](https://github.com/google/nomulus/pull/1575) <--did not add default value, which was why two PRs were made for schema change. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1573)

### DIFF
--- a/core/src/main/java/google/registry/model/billing/BillingEvent.java
+++ b/core/src/main/java/google/registry/model/billing/BillingEvent.java
@@ -126,7 +126,7 @@ public abstract class BillingEvent extends ImmutableObject
      * This flag will be added to any {@link OneTime} events that are created via, e.g., an
      * automated process to expand {@link Recurring} events.
      */
-    SYNTHETIC
+    SYNTHETIC;
   }
 
   /**
@@ -142,20 +142,24 @@ public abstract class BillingEvent extends ImmutableObject
      * <p>By default, if the domain is premium, then premium price will be used. Otherwise, the
      * standard price of the TLD will be used.
      */
-    DEFAULT_PRICE,
+    DEFAULT,
     /**
-     * This indicates that standard price will be used for domain renewal.
+     * This indicates the domain will be renewed at standard price even if it's a premium domain.
+     *
+     * <p>We chose to name this "NONPREMIUM" rather than simply "STANDARD" to avoid confusion
+     * between "STANDARD" and "DEFAULT".
      *
      * <p>This price behavior is used with anchor tenants.
      */
-    STANDARD_PRICE,
+    NONPREMIUM,
     /**
-     * This indicates that the renewalPrice in {@link Recurring} will be used for domain renewal.
+     * This indicates that the renewalPrice in {@link BillingEvent.Recurring} will be used for
+     * domain renewal.
      *
-     * <p>The renewalPrice has a value iff the price behavior is set to SPECIFIED_PRICE. This
+     * <p>The renewalPrice has a non-null value iff the price behavior is set to "SPECIFIED". This
      * behavior is used with internal registrations.
      */
-    SPECIFIED_PRICE
+    SPECIFIED;
   }
 
   /** Entity id. */
@@ -588,16 +592,17 @@ public abstract class BillingEvent extends ImmutableObject
      * The renewal price for domain renewal if and only if it's specified.
      *
      * <p>This price column remains null except when the renewal price behavior of the billing is
-     * SPECIFIED_PRICE. This column is used for internal registrations.
+     * SPECIFIED. This column is used for internal registrations.
      */
     @Nullable
     @Type(type = JodaMoneyType.TYPE_NAME)
     @Columns(
         columns = {@Column(name = "renewalPriceAmount"), @Column(name = "renewalPriceCurrency")})
-    Optional<Money> renewalPrice;
+    Money renewalPrice;
 
-    @Column(name = "renewalPriceBehavior")
-    RenewalPriceBehavior renewalPriceBehavior;
+    @Enumerated(EnumType.STRING)
+    @Column(name = "renewalPriceBehavior", nullable = false)
+    RenewalPriceBehavior renewalPriceBehavior = RenewalPriceBehavior.DEFAULT;
 
     public DateTime getRecurrenceEndTime() {
       return recurrenceEndTime;
@@ -612,7 +617,7 @@ public abstract class BillingEvent extends ImmutableObject
     }
 
     public Optional<Money> getRenewalPrice() {
-      return renewalPrice;
+      return Optional.ofNullable(renewalPrice);
     }
 
     @Override
@@ -648,9 +653,8 @@ public abstract class BillingEvent extends ImmutableObject
         return this;
       }
 
-      public Builder setRenewalPrice(Money renewalPrice) {
-        checkNotNull(renewalPrice, "Renewal price is not specified");
-        getInstance().renewalPrice = Optional.of(renewalPrice);
+      public Builder setRenewalPrice(@Nullable Money renewalPrice) {
+        getInstance().renewalPrice = renewalPrice;
         return this;
       }
 
@@ -659,12 +663,11 @@ public abstract class BillingEvent extends ImmutableObject
         Recurring instance = getInstance();
         checkNotNull(instance.eventTime);
         checkNotNull(instance.reason);
-        if (instance.renewalPrice.isPresent()) {
-          checkArgument(
-              instance.renewalPriceBehavior == RenewalPriceBehavior.SPECIFIED_PRICE,
-              "Renewal price can have a value if an only if the renewal price behavior is"
-                  + " SPECIFIED_PRICE");
-        }
+        checkArgument(
+            (instance.renewalPriceBehavior == RenewalPriceBehavior.SPECIFIED)
+                ^ (instance.renewalPrice == null),
+            "Renewal price can have a value if and only if the renewal price behavior is"
+                + " SPECIFIED");
         instance.recurrenceTimeOfYear = TimeOfYear.fromDateTime(instance.eventTime);
         instance.recurrenceEndTime =
             Optional.ofNullable(instance.recurrenceEndTime).orElse(END_OF_TIME);
--- a/core/src/main/java/google/registry/model/billing/BillingEvent.java
+++ b/core/src/main/java/google/registry/model/billing/BillingEvent.java
@@ -129,6 +129,35 @@ public abstract class BillingEvent extends ImmutableObject
     SYNTHETIC
   }
 
+  /**
+   * Sets of renewal price behaviors that can be applied to billing recurrences.
+   *
+   * <p>When a client renews a domain, they could be charged differently, depending on factors such
+   * as the client type and the domain itself.
+   */
+  public enum RenewalPriceBehavior {
+    /**
+     * This indicates the renewal price is the default price.
+     *
+     * <p>By default, if the domain is premium, then premium price will be used. Otherwise, the
+     * standard price of the TLD will be used.
+     */
+    DEFAULT_PRICE,
+    /**
+     * This indicates that standard price will be used for domain renewal.
+     *
+     * <p>This price behavior is used with anchor tenants.
+     */
+    STANDARD_PRICE,
+    /**
+     * This indicates that the renewalPrice in {@link Recurring} will be used for domain renewal.
+     *
+     * <p>The renewalPrice has a value iff the price behavior is set to SPECIFIED_PRICE. This
+     * behavior is used with internal registrations.
+     */
+    SPECIFIED_PRICE
+  }
+
   /** Entity id. */
   @Id @javax.persistence.Id Long id;
 
@@ -555,12 +584,35 @@ public abstract class BillingEvent extends ImmutableObject
     })
     TimeOfYear recurrenceTimeOfYear;
 
+    /**
+     * The renewal price for domain renewal if and only if it's specified.
+     *
+     * <p>This price column remains null except when the renewal price behavior of the billing is
+     * SPECIFIED_PRICE. This column is used for internal registrations.
+     */
+    @Nullable
+    @Type(type = JodaMoneyType.TYPE_NAME)
+    @Columns(
+        columns = {@Column(name = "renewalPriceAmount"), @Column(name = "renewalPriceCurrency")})
+    Optional<Money> renewalPrice;
+
+    @Column(name = "renewalPriceBehavior")
+    RenewalPriceBehavior renewalPriceBehavior;
+
     public DateTime getRecurrenceEndTime() {
       return recurrenceEndTime;
     }
 
     public TimeOfYear getRecurrenceTimeOfYear() {
       return recurrenceTimeOfYear;
+    }
+
+    public RenewalPriceBehavior getRenewalPriceBehavior() {
+      return renewalPriceBehavior;
+    }
+
+    public Optional<Money> getRenewalPrice() {
+      return renewalPrice;
     }
 
     @Override
@@ -591,11 +643,28 @@ public abstract class BillingEvent extends ImmutableObject
         return this;
       }
 
+      public Builder setRenewalPriceBehavior(RenewalPriceBehavior renewalPriceBehavior) {
+        getInstance().renewalPriceBehavior = renewalPriceBehavior;
+        return this;
+      }
+
+      public Builder setRenewalPrice(Money renewalPrice) {
+        checkNotNull(renewalPrice, "Renewal price is not specified");
+        getInstance().renewalPrice = Optional.of(renewalPrice);
+        return this;
+      }
+
       @Override
       public Recurring build() {
         Recurring instance = getInstance();
         checkNotNull(instance.eventTime);
         checkNotNull(instance.reason);
+        if (instance.renewalPrice.isPresent()) {
+          checkArgument(
+              instance.renewalPriceBehavior == RenewalPriceBehavior.SPECIFIED_PRICE,
+              "Renewal price can have a value if an only if the renewal price behavior is"
+                  + " SPECIFIED_PRICE");
+        }
         instance.recurrenceTimeOfYear = TimeOfYear.fromDateTime(instance.eventTime);
         instance.recurrenceEndTime =
             Optional.ofNullable(instance.recurrenceEndTime).orElse(END_OF_TIME);
--- a/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java
+++ b/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java
@@ -121,13 +121,17 @@ public class JodaMoneyType implements CompositeUserType {
       ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner)
       throws HibernateException, SQLException {
     BigDecimal amount = StandardBasicTypes.BIG_DECIMAL.nullSafeGet(rs, names[AMOUNT_ID], session);
-    CurrencyUnit currencyUnit =
-        CurrencyUnit.of(StandardBasicTypes.STRING.nullSafeGet(rs, names[CURRENCY_ID], session));
-    if (amount != null && currencyUnit != null) {
-      return Money.of(currencyUnit, amount.stripTrailingZeros());
-    }
-    if (amount == null && currencyUnit == null) {
+    String currencyUnitString =
+        StandardBasicTypes.STRING.nullSafeGet(rs, names[CURRENCY_ID], session);
+    // It is allowable for a Money object to be null, but only if both the currency unit and the
+    // amount are null
+    if (amount == null && currencyUnitString == null) {
       return null;
+    }
+    if (amount != null) {
+      // CurrencyUnit.of() throws an IllegalCurrencyException for unknown currency, which means the
+      // currency is valid if it returns a value
+      return Money.of(CurrencyUnit.of(currencyUnitString), amount.stripTrailingZeros());
     }
     throw new HibernateException("Mismatching null state between currency and amount.");
   }
--- a/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java
+++ b/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java
@@ -136,7 +136,7 @@ public class JodaMoneyType implements CompositeUserType {
     } else {
       throw new HibernateException(
           String.format(
-              "Mismatching null state between currency '%s' and amount '%s'.",
+              "Mismatching null state between currency '%s' and amount '%s'",
               currencyUnitString, amount));
     }
   }
@@ -149,7 +149,7 @@ public class JodaMoneyType implements CompositeUserType {
     String currencyUnit = value == null ? null : ((Money) value).getCurrencyUnit().getCode();
 
     if ((amount == null && currencyUnit != null) || (amount != null && currencyUnit == null)) {
-      throw new HibernateException("Mismatching null state between currency and amount.");
+      throw new HibernateException("Mismatching null state between currency and amount");
     }
     StandardBasicTypes.BIG_DECIMAL.nullSafeSet(st, amount, index, session);
     StandardBasicTypes.STRING.nullSafeSet(st, currencyUnit, index + 1, session);
--- a/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java
+++ b/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java
@@ -21,6 +21,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Objects;
+import javax.annotation.Nullable;
 import org.hibernate.HibernateException;
 import org.hibernate.engine.spi.SharedSessionContractImplementor;
 import org.hibernate.type.StandardBasicTypes;
@@ -116,6 +117,7 @@ public class JodaMoneyType implements CompositeUserType {
     return Objects.hashCode(x);
   }
 
+  @Nullable
   @Override
   public Object nullSafeGet(
       ResultSet rs, String[] names, SharedSessionContractImplementor session, Object owner)
@@ -127,13 +129,16 @@ public class JodaMoneyType implements CompositeUserType {
     // amount are null
     if (amount == null && currencyUnitString == null) {
       return null;
-    }
-    if (amount != null) {
+    } else if (amount != null && currencyUnitString != null) {
       // CurrencyUnit.of() throws an IllegalCurrencyException for unknown currency, which means the
       // currency is valid if it returns a value
       return Money.of(CurrencyUnit.of(currencyUnitString), amount.stripTrailingZeros());
+    } else {
+      throw new HibernateException(
+          String.format(
+              "Mismatching null state between currency '%s' and amount '%s'.",
+              currencyUnitString, amount));
     }
-    throw new HibernateException("Mismatching null state between currency and amount.");
   }
 
   @Override
--- a/core/src/test/java/google/registry/model/billing/BillingEventTest.java
+++ b/core/src/test/java/google/registry/model/billing/BillingEventTest.java
@@ -524,7 +524,7 @@ public class BillingEventTest extends EntityTestCase {
             .build());
     assertThat(loadByEntity(recurringEvent).getRenewalPriceBehavior())
         .isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(loadByEntity(recurringEvent).getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
+    assertThat(loadByEntity(recurringEvent).getRenewalPrice()).hasValue(Money.of(USD, 100));
   }
 
   @TestOfyAndSql
@@ -574,7 +574,7 @@ public class BillingEventTest extends EntityTestCase {
             .build());
     assertThat(loadByEntity(recurringEvent).getRenewalPriceBehavior())
         .isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(loadByEntity(recurringEvent).getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
+    assertThat(loadByEntity(recurringEvent).getRenewalPrice()).hasValue(Money.of(USD, 100));
   }
 
   @TestOfyAndSql
@@ -614,7 +614,7 @@ public class BillingEventTest extends EntityTestCase {
                     .setRenewalPrice(Money.of(USD, 100))
                     .setRecurrenceEndTime(END_OF_TIME)));
     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(recurringEvent.getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
+    assertThat(recurringEvent.getRenewalPrice()).hasValue(Money.of(USD, 100));
     BillingEvent.Recurring loadedEntity = loadByEntity(recurringEvent);
     assertThat(loadedEntity).isEqualTo(recurringEvent);
     persistResource(
@@ -642,7 +642,7 @@ public class BillingEventTest extends EntityTestCase {
                     .setRenewalPrice(Money.of(USD, 100))
                     .setRecurrenceEndTime(END_OF_TIME)));
     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(recurringEvent.getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
+    assertThat(recurringEvent.getRenewalPrice()).hasValue(Money.of(USD, 100));
     BillingEvent.Recurring loadedEntity = loadByEntity(recurringEvent);
     assertThat(loadedEntity).isEqualTo(recurringEvent);
     persistResource(
@@ -796,7 +796,7 @@ public class BillingEventTest extends EntityTestCase {
                     .setRenewalPrice(Money.of(USD, 100))
                     .setRecurrenceEndTime(END_OF_TIME)));
     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(recurringEvent.getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
+    assertThat(recurringEvent.getRenewalPrice()).hasValue(Money.of(USD, 100));
     BillingEvent.Recurring loadedEntity = loadByEntity(recurringEvent);
     assertThat(loadedEntity).isEqualTo(recurringEvent);
     IllegalArgumentException thrown =
@@ -828,7 +828,7 @@ public class BillingEventTest extends EntityTestCase {
                     .setRenewalPrice(Money.of(USD, 100))
                     .setRecurrenceEndTime(END_OF_TIME)));
     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(recurringEvent.getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
+    assertThat(recurringEvent.getRenewalPrice()).hasValue(Money.of(USD, 100));
     BillingEvent.Recurring loadedEntity = loadByEntity(recurringEvent);
     assertThat(loadedEntity).isEqualTo(recurringEvent);
     IllegalArgumentException thrown =
@@ -860,8 +860,7 @@ public class BillingEventTest extends EntityTestCase {
                     .setRenewalPrice(Money.of(USD, BigDecimal.valueOf(100)))
                     .setRecurrenceEndTime(END_OF_TIME)));
     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(recurringEvent.getRenewalPrice().get())
-        .isEqualTo(Money.of(USD, BigDecimal.valueOf(100)));
+    assertThat(recurringEvent.getRenewalPrice()).hasValue(Money.of(USD, 100));
   }
 
   @TestOfyAndSql
@@ -894,8 +893,7 @@ public class BillingEventTest extends EntityTestCase {
                     .setRenewalPrice(Money.of(USD, BigDecimal.valueOf(100)))
                     .setRecurrenceEndTime(END_OF_TIME)));
     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
-    assertThat(recurringEvent.getRenewalPrice().get())
-        .isEqualTo(Money.of(USD, BigDecimal.valueOf(100)));
+    assertThat(recurringEvent.getRenewalPrice()).hasValue(Money.of(USD, 100));
   }
 
   @TestOfyAndSql
--- a/core/src/test/java/google/registry/model/billing/BillingEventTest.java
+++ b/core/src/test/java/google/registry/model/billing/BillingEventTest.java
@@ -15,6 +15,7 @@
 package google.registry.model.billing;
 
 import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth8.assertThat;
 import static google.registry.model.domain.token.AllocationToken.TokenType.UNLIMITED_USE;
 import static google.registry.model.ofy.ObjectifyService.auditedOfy;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
@@ -36,6 +37,7 @@ import com.googlecode.objectify.Key;
 import google.registry.model.EntityTestCase;
 import google.registry.model.billing.BillingEvent.Flag;
 import google.registry.model.billing.BillingEvent.Reason;
+import google.registry.model.billing.BillingEvent.RenewalPriceBehavior;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.domain.DomainHistory;
 import google.registry.model.domain.GracePeriod;
@@ -49,6 +51,7 @@ import google.registry.testing.TestOfyAndSql;
 import google.registry.testing.TestOfyOnly;
 import google.registry.testing.TestSqlOnly;
 import google.registry.util.DateTimeUtils;
+import java.math.BigDecimal;
 import org.joda.money.Money;
 import org.joda.time.DateTime;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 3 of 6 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 8 unresolved discussions (waiting on @mindhog and @rachelguan)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM7rkF6528xuKvng86:-MzM7rkF6528xuKvng87:b-4e93fl) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L132)):*
> ```Java
>   }
> 
>   /** The expected renewal behaviors that can be applied to billing recurrences. */
> ```

These comments need to be worded better.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 134 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8YUp9jcQRTdqb11C:-MzM8YUp9jcQRTdqb11D:beg3tks) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L134)):*
> ```Java
>   /** The expected renewal behaviors that can be applied to billing recurrences. */
>   public enum RenewalPriceBehavior {
>     /** The default price behavior to renew a domain. */
> ```

Comment needs to mention that it will use the premium price for the label if there is one, otherwise the TLD's standard price.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM7u593MbdIlh_3ta8:-MzM7u593MbdIlh_3ta9:b-4wdon3) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L136)):*
> ```Java
>     /** The default price behavior to renew a domain. */
>     DEFAULT_PRICE,
>     /** The price behavior to renew a domain for an anchor tenant. */
> ```

This isn't necessarily only for use with anchor tenants.  Just describe what it does.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM7ycuBes3VMvrNnM0:-MzM7ycuBes3VMvrNnM1:b-oqvt5f) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L138)):*
> ```Java
>     /** The price behavior to renew a domain for an anchor tenant. */
>     STANDARD_PRICE,
>     /** The price behavior to renew a domain for an internal registration. */
> ```

There's no mention here that `renewalPrice` has a value if and only if the renewal price behavior is set to this value.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 139 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8FC60yFhkjW5UUPU:-MzM8FC60yFhkjW5UUPV:bmepu78) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L139)):*
> ```Java
>     STANDARD_PRICE,
>     /** The price behavior to renew a domain for an internal registration. */
>     DISCOUNTED_PRICE
> ```

`DISCOUNTED_PRICE` isn't a good name; a better name would be `SPECIFIED_PRICE`.  The intent of this behavior is that we can override the renewals to happen at a price that we specify, which may or may not be less than the price it otherwise would be.  There are potential use cases here for why we might set this to be greater.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 535 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM6ag034_VQMmJnHhJ:-MzM6ag034_VQMmJnHhK:b-f5h0ne) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L535)):*
> ```Java
>         @javax.persistence.Index(columnList = "domainRepoId"),
>         @javax.persistence.Index(columnList = "recurrenceEndTime"),
>         @javax.persistence.Index(columnList = "renewalPriceBehavior"),
> ```

Why are these indexes necessary?
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 577 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM6cZx2v5aF34f0FzK:-MzM6cZx2v5aF34f0FzL:b-rraobx) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L577)):*
> ```Java
>     Money renewalPrice;
> 
>     RenewalPriceBehavior renewalPriceBehavior;
> ```

This is going to be a data migration problem -- you didn't assign a correct default value in the Flyway script, so all existing data is going to have `null` here, rather than the expected value of `DEFAULT_PRICE`.  Any time you're adding a new column you have to think about how the existing data is going to be handled.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 641 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM86G73PYYPOq4hBTN:-MzM86G88M3rR8cHx9Dr:b-szof4t) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L641)):*
> ```Java
>         instance.recurrenceEndTime =
>             Optional.ofNullable(instance.recurrenceEndTime).orElse(END_OF_TIME);
>         return super.build();
> ```

You need to add error checking here that `renewalPrice` can have a value if and only if the renewal price behavior is `DISCOUNTED_PRICE`.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 132 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM7rkF6528xuKvng86:-MzMC_-SAamBFuB6LrRA:b-pqo3aj) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L132)):*

These comments need to be worded better.
</blockquote></details>

I tried to follow the existing comments for the `enums` in the same file. The name of the `enum` type already seems self-explanatory to me but I'm happy to add more details to explain the purpose of this `enum`. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 134 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8YUp9jcQRTdqb11C:-MzMCZKa50jCi3ZDvnGU:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L134)):*

Comment needs to mention that it will use the premium price for the label if there is one, otherwise the TLD's standard price.
</blockquote></details>

done
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 136 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM7u593MbdIlh_3ta8:-MzMCXJK0VImJHkaXb7d:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L136)):*

This isn't necessarily only for use with anchor tenants.  Just describe what it does.
</blockquote></details>

done
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 138 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM7ycuBes3VMvrNnM0:-MzMGcdH7XSFjcy8JP46:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L138)):*

There's no mention here that `renewalPrice` has a value if and only if the renewal price behavior is set to this value.
</blockquote></details>

done
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 139 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8FC60yFhkjW5UUPU:-MzMCVziDSDNUr9Szzts:b-mx77jz) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L139)):*

`DISCOUNTED_PRICE` isn't a good name; a better name would be `SPECIFIED_PRICE`.  The intent of this behavior is that we can override the renewals to happen at a price that we specify, which may or may not be less than the price it otherwise would be.  There are potential use cases here for why we might set this to be greater.
</blockquote></details>

noted.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 535 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM6ag034_VQMmJnHhJ:-MzMH2OL0IWfYD5KD6wv:b-nflu33) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L535)):*

Why are these indexes necessary?
</blockquote></details>

Not necessary but I was trying to stay consistent with the existing table. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 571 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8mws8zOMIA1mTpTB:-MzMNhbt-HrU2JjLyTU4:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L571)):*

Expand on when this can have a value.
</blockquote></details>

done
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 575 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8oj9CMARHzCS0ep6:-MzMMatD2jiW1l3gIn9-:b-rnrapz) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L575)):*

Needs to be `@Nullable`.
</blockquote></details>

done. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 577 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM6cZx2v5aF34f0FzK:-MzMM4QT8DVOmxhNNtJS:bkxf2a9) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L577)):*

This is going to be a data migration problem -- you didn't assign a correct default value in the Flyway script, so all existing data is going to have `null` here, rather than the expected value of `DEFAULT_PRICE`.  Any time you're adding a new column you have to think about how the existing data is going to be handled.
</blockquote></details>

i will create a PR to add that. I was thinking that the existing behavior could stay the way it was with null value. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 577 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM93XA9IzJYNSYjjqq:-MzMNmZ549fG3dUdMZN5:b-x9epf6) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L577)):*

Not necessarily part of this PR (indeed it cannot be until the related DB change is made), but you're intending to add this to the `BillingEvent` table as well, right?
</blockquote></details>

will create PR before this. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 591 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8qiR0sB8cq5LYR8U:-MzMJBq29zzMRFqETTxI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L591)):*

Needs to return `Optional`, as this will usually be null.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 628 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8ye21au5yZlWZf0K:-MzMKY_-2DC9xDZr1a-1:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L628)):*

Needs handling for null/absent values here, as "renewalPrice not specified" is a valid state.
</blockquote></details>

done
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 641 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM86G73PYYPOq4hBTN:-MzMLu404ytwk--N7Uiz:b-iiltaw) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L641)):*

You need to add error checking here that `renewalPrice` can have a value if and only if the renewal price behavior is `DISCOUNTED_PRICE`.
</blockquote></details>

done and changed it to `SPECIFIED_PRICE`

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 575 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8oj9CMARHzCS0ep6:-MzRGj3_3qG1ZThGIdm7:bclruci) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L575)):*

done. 
</blockquote></details>

This seems wrong -- you'd normally never want to combine at-Nullable and Optional, I don't think the presence of the other annotations changes that.  I initially assumed Ben might have meant "nullable = true" in the column definitions, though I would expect that would be inferred from the Optional and, indeed, it appears to have been since the generated code hasn't changed.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 12 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 575 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8oj9CMARHzCS0ep6:-MzRLR8Z9i_Jcz8CTRFt:byqbdo5) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L575)):*

To clarify, this should have been either at-Nullable or Optional, not both.
</blockquote></details>

yea makes sense if it's Optional.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 535 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM6ag034_VQMmJnHhJ:-MzRPpm32kbaX7OZbcob:b5fjd7b) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L535)):*

Not necessary but I was trying to stay consistent with the existing table. 
</blockquote></details>

Being consistent with the existing table would be only adding indexes that are necessary, not adding indexes on everything.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 575 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8oj9CMARHzCS0ep6:-MzRRkuD4ooWLcThkAZn:b-fn982q) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L575)):*

yea makes sense if it's Optional. 
</blockquote></details>

It can't be `Optional` -- that doesn't work for data model fields.  Look at other entity classes as an example.  This field needs to be `@Nullable Money renewalPrice`.  And then the getters and setters need to be typed with `Optional<Money>` and do the conversion themselves between `Optional` and `null`.  As an example see: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/billing/BillingEvent.java;l=392-394;drc=3ef1e6c6a480444f77edbe2abbd8ec4a4a53a54c?q=billingevent&ss=nomulus%2Fnomulus
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 577 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM6cZx2v5aF34f0FzK:-MzRTcxf0tNaB-D87sfQ:b-5ariss) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L577)):*

i will create a PR to add that. I was thinking that the existing behavior could stay the way it was with null value. 
</blockquote></details>

Yeah, that would be much worse than only having a single possible value representing a single case in business logic.  Having both `DEFAULT` and `(null)` be valid states in the DB, and do the same thing, would introduce a lot more potential for errors all over, and would make SQL queries quite ugly (as you'd need to query for both possible values every time just to get all domains that have normal renewal behavior).
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 628 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8ye21au5yZlWZf0K:-MzRU0GNBWKApf1fBlOE:bux3j6c) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L628)):*

done
</blockquote></details>

To be clear, the handling for null/absent values that it needs is that they should be _allowed_ and handled correctly.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 138 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRPxlFE9dgkJWmai_g:-MzRPxlFE9dgkJWmai_h:bqptj9t) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L138)):*
> ```Java
>    * as the client type and the domain itself.
>    */
>   public enum RenewalPriceBehavior {
> ```

Thinking further about the names here, and especially after seeing what the data actually looks like in the DB per PR #1575, I think we want to do something shorter, just 1 word, that will look nicer in SQL queries and also not be redundant.  Like, `RenewalPriceBehavior.DEFAULT_PRICE` has the redundant word `price` in both parts.

Anyway, my suggestions are:

```
NORMAL
NONPREMIUM
SPECIFIED
```

It is kind of hard to come up with wording that doesn't use `STANDARD`/`DEFAULT` (which is a bit confusing), but I think this works.

NORMAL is just the normal pricing behavior, i.e. what applies to 99.9% of domains.  NONPREMIUM would be for anchor tenants and maybe other stuff, and it indicates that the premium list isn't applied, and then SPECIFIED is, of course, this domain renews at the specified price.

I think that works?
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 158 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRPcLuEhNUM4P9bXNx:-MzRPcLuEhNUM4P9bXNy:b8ackn3) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L158)):*
> ```Java
>      * behavior is used with internal registrations.
>      */
>     SPECIFIED_PRICE
> ```

Missing semicolon after the last one.  (It being missing here would be a compilation error if there were any further logic in the enum)
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 652 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRTxKo5SiwZNtb21kS:-MzRTxKo5SiwZNtb21kT:b-q61glq) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L652)):*
> ```Java
> 
>       public Builder setRenewalPrice(Money renewalPrice) {
>         checkNotNull(renewalPrice, "Renewal price is not specified");
> ```

"renewalPrice not specified" is a valid state, yet here you're making it an error.  I don't think you understood my previous comment.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 662 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRU7tT4MWbgGx9JCeU:-MzRU7tT4MWbgGx9JCeV:b-2ujxgm) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L662)):*
> ```Java
>         checkNotNull(instance.eventTime);
>         checkNotNull(instance.reason);
>         if (instance.renewalPrice.isPresent()) {
> ```

This does not correctly handle the if-and-only-if error condition.  Having renewal price be absent when the renewal price behavior is `SPECIFIED_PRICE` is also an error, yet you are not throwing an error for that condition here.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 575 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8oj9CMARHzCS0ep6:-MzWGfiUEvKR-CtBP5--:b-qrigx6) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L575)):*

It can't be `Optional` -- that doesn't work for data model fields.  Look at other entity classes as an example.  This field needs to be `@Nullable Money renewalPrice`.  And then the getters and setters need to be typed with `Optional<Money>` and do the conversion themselves between `Optional` and `null`.  As an example see: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/billing/BillingEvent.java;l=392-394;drc=3ef1e6c6a480444f77edbe2abbd8ec4a4a53a54c?q=billingevent&ss=nomulus%2Fnomulus
</blockquote></details>

noted and fixed.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 628 at r1](https://reviewable.io/reviews/google/nomulus/1573#-MzM8ye21au5yZlWZf0K:-MzWNOmK6oy2X6uUIIcw:bn602zm) ([raw file](https://github.com/google/nomulus/blob/bcbd13c18be3078f01a5a2f6161335c111e96597/core/src/main/java/google/registry/model/billing/BillingEvent.java#L628)):*

To be clear, the handling for null/absent values that it needs is that they should be _allowed_ and handled correctly.
</blockquote></details>

not sure if there's a way to check absent but added null check .. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 138 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRPxlFE9dgkJWmai_g:-MzRZetl4TGd4hh34T8Y:bvlccer) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L138)):*

Thinking further about the names here, and especially after seeing what the data actually looks like in the DB per PR #1575, I think we want to do something shorter, just 1 word, that will look nicer in SQL queries and also not be redundant.  Like, `RenewalPriceBehavior.DEFAULT_PRICE` has the redundant word `price` in both parts.

Anyway, my suggestions are:

```
NORMAL
NONPREMIUM
SPECIFIED
```

It is kind of hard to come up with wording that doesn't use `STANDARD`/`DEFAULT` (which is a bit confusing), but I think this works.

NORMAL is just the normal pricing behavior, i.e. what applies to 99.9% of domains.  NONPREMIUM would be for anchor tenants and maybe other stuff, and it indicates that the premium list isn't applied, and then SPECIFIED is, of course, this domain renews at the specified price.

I think that works?
</blockquote></details>

going with `DEFAULT`, `NONPREMIUM` and `SPECIFIED`. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 158 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRPcLuEhNUM4P9bXNx:-MzRZPJ32nb9LV655nJq:bgs63yw) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L158)):*
> NONPREMIUM
I actually think it would work; I was referencing the Flag `enum` above which does not have a semicolon. 

https://stackoverflow.com/questions/24614393/why-is-the-semicolon-not-required-but-allowed-at-the-end-of-a-class-definition

I added them tho after your comment. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 652 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRTxKo5SiwZNtb21kS:-MzWJh8B9RM8Sr30JAOn:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L652)):*

"renewalPrice not specified" is a valid state, yet here you're making it an error.  I don't think you understood my previous comment.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 662 at r2](https://reviewable.io/reviews/google/nomulus/1573#-MzRU7tT4MWbgGx9JCeU:-MzWNYCo1JtbQailMupb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3d3676680c45e91030f3051abc69789652f29c07/core/src/main/java/google/registry/model/billing/BillingEvent.java#L662)):*

This does not correctly handle the if-and-only-if error condition.  Having renewal price be absent when the renewal price behavior is `SPECIFIED_PRICE` is also an error, yet you are not throwing an error for that condition here.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 166 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazIjL8a6g9hM5Ft91:-MzazIjL8a6g9hM5Ft92:b-lf7d6u) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L166)):*
> ```Java
>      */
>     SPECIFIED
>   };
> ```

The semicolon doesn't go here, it goes after the last value in the enum.  It's not strictly required until you add additional logic after the enum values themselves, but we tend to use it in all enums across the codebase.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 659 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazXz63JVo8w-_WKS4:-MzazXz63JVo8w-_WKS5:b-jq0b4s) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L659)):*
> ```Java
> 
>       public Builder setRenewalPrice(@Nullable Money renewalPrice) {
>         if (renewalPrice != null) {
> ```

This still doesn't work.  What if the current state is that the renewal price is 5, and we call the setter with `null` because we want to set the renewal price to null?  Then your code here isn't doing anything.

You're overthinking things.  Just remove the conditional.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 671 at r4](https://reviewable.io/reviews/google/nomulus/1573#-Mzazoz2EBkxKQF3qHtP:-Mzazoz2EBkxKQF3qHtQ:bdamnid) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L671)):*
> ```Java
>         checkNotNull(instance.reason);
>         checkArgument(
>             ((instance.renewalPriceBehavior == RenewalPriceBehavior.SPECIFIED
> ```

Look into the `^` operator in Java.  It will make this terser.  https://www.baeldung.com/java-xor-operator

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 166 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazIjL8a6g9hM5Ft91:-MzpIG1x5394LP8s-uDq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L166)):*

The semicolon doesn't go here, it goes after the last value in the enum.  It's not strictly required until you add additional logic after the enum values themselves, but we tend to use it in all enums across the codebase.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 659 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazXz63JVo8w-_WKS4:-MzpM7Tu1Tug5rfv_f6Q:b-z2wbua) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L659)):*

This still doesn't work.  What if the current state is that the renewal price is 5, and we call the setter with `null` because we want to set the renewal price to null?  Then your code here isn't doing anything.

You're overthinking things.  Just remove the conditional.
</blockquote></details>

that is not the one of the use cases we are considering, even tho that technically could be a case. and ok removed it.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 671 at r4](https://reviewable.io/reviews/google/nomulus/1573#-Mzazoz2EBkxKQF3qHtP:-MzpQRsw4H5lwIkkSGJB:b-896fix) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L671)):*

Look into the `^` operator in Java.  It will make this terser.  https://www.baeldung.com/java-xor-operator
</blockquote></details>

Done.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: all files reviewed, 2 unresolved discussions (waiting on @rachelguan)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 659 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazXz63JVo8w-_WKS4:-MzqNr-v0OqLHjP-UTSi:btcxrj1) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L659)):*

that is not the one of the use cases we are considering, even tho that technically could be a case. and ok removed it.
</blockquote></details>

Why isn't it a use case we're considering? Let's say that a recurrence already has a behavior of `SPECIFIED`, but needs to be changed to something else for some reason.  Since it was `SPECIFIED`, then it had a value for `renewalPrice`, but now that we're changing the behavior, the `renewalPrice` needs to be `null`ed out.  Hence why we have to support it here.  If `null` is a valid value for this field (and it absolutely is; the vast majority of `BillingRecurrence`s in the entire DB will be `null`), then the setter needs to support that.  Otherwise you get into a situation where its value is set but then cannot be unset.
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 671 at r4](https://reviewable.io/reviews/google/nomulus/1573#-Mzazoz2EBkxKQF3qHtP:-MzqNcLD8jXtWM2GLEMw:b-80al8z) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L671)):*

Done.
</blockquote></details>

Are you sure that the order of operations here is correct?  It would be clearer to use parentheses.  Also, this PR doesn't have any tests?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 659 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazXz63JVo8w-_WKS4:-MzqZ2Nw2XJ9dQYlpKMC:b-y2viur) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L659)):*

Why isn't it a use case we're considering? Let's say that a recurrence already has a behavior of `SPECIFIED`, but needs to be changed to something else for some reason.  Since it was `SPECIFIED`, then it had a value for `renewalPrice`, but now that we're changing the behavior, the `renewalPrice` needs to be `null`ed out.  Hence why we have to support it here.  If `null` is a valid value for this field (and it absolutely is; the vast majority of `BillingRecurrence`s in the entire DB will be `null`), then the setter needs to support that.  Otherwise you get into a situation where its value is set but then cannot be unset.
</blockquote></details>

"changed to something else for some reason". It remains null unless it's specified. There's no use case, at least not mentioned in the design doc that the specified value will be modified. For internal registrations, we want to renew at the initial price and I assume initial price won't change? 

Anyway, removed the conditional. 
___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 671 at r4](https://reviewable.io/reviews/google/nomulus/1573#-Mzazoz2EBkxKQF3qHtP:-MzqWxp76E5ojQNX3LFY:b-tlu7lc) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L671)):*

Are you sure that the order of operations here is correct?  It would be clearer to use parentheses.  Also, this PR doesn't have any tests?
</blockquote></details>

Added test cases to test this argument and added parenthesis for readability.  

For the order of operations. 
Let A be renewalPrice is SPECIFIED, and B be renewalPrice is null 
we want To return true for when 
  1.renewalPriceBehavior is SPECIFIED  and renewalPrice is NOT null  (A =true, B= false,  which results true) 
  2. renewalPriceBehavior is NOT SPECIFIED and renewalPrice is null (A =false, B= true, which results true) 
 and the other two cases return false: 
  1. renewalPriceBehavior is SPECIFIED and renewalPrice is null  (A is true and B is true) 
  2. renewalPriceBehavior is NOT SPECIFIED and renewalPrice is NOT null (A = false, B = false)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/model/billing/BillingEvent.java, line 659 at r4](https://reviewable.io/reviews/google/nomulus/1573#-MzazXz63JVo8w-_WKS4:-N-00zea5WHyvmmyxlhj:b-69kcnd) ([raw file](https://github.com/google/nomulus/blob/fbb352c97e0270dcab36447f9036bb6048aa7e76/core/src/main/java/google/registry/model/billing/BillingEvent.java#L659)):*

"changed to something else for some reason". It remains null unless it's specified. There's no use case, at least not mentioned in the design doc that the specified value will be modified. For internal registrations, we want to renew at the initial price and I assume initial price won't change? 

Anyway, removed the conditional. 
</blockquote></details>

It's not a good idea to have `null` as an allowed state, but just have that be the default, and then not actually allow it to be explicitly set to that.  It's confusing.  You could imagine someone calling it from a callsite thinking they're explicitly setting it as `null`, and then seeing that as an error, and then thinking `null` is not a valid value when in fact it is.  Don't just rely on the default; if it's valid, it should be passable.

And, more concretely, it's a relatively common pattern in tests to set up an object with a set of default values, and then modify them after the fact in individual tests.
___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 130 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-01hNLAdzWtbWmZoPs:-N-01hNM3Q4lxlz4lUXN:b-4p0bdg) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L130)):*
> ```Java
>         return Money.of(currencyUnit, amount.stripTrailingZeros());
>       }
>     } catch (NullPointerException e) {
> ```

Catching an exception only to swallow it as an expected value is generally a code smell.  Also, you're not differentiating *why* `NullPointerException` is being thrown here; it's possible that it's being thrown by something else unexpected, in which case swallowing it is the wrong thing to do.

It's better to explicitly detect the error condition in a conditional.

Also this probably merits a comment explaining what exactly it is doing (namely, it is allowable for a money to be null, but only if the currency unit and the amount are null together).

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 483 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49dpWAR4rgVKAuatE:-N-49dpWAR4rgVKAuatF:b7topy3) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L483)):*
> ```Java
>   void testSuccess_defaultRenewalPriceBehavior_assertsIsDefault() {
>     assertThat(recurring.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.DEFAULT);
>     assertThat(recurring.getRenewalPrice().isEmpty()).isTrue();
> ```

Statically import `Truth8.assertThat` and then us `assertThat(...).isEmpty()` .  There's plenty of other examples in our codebase of doing this.  The error messages are better this way.
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 499 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49nPf9u-8zodBZr_W:-N-49nPf9u-8zodBZr_X:b-89a4l7) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L499)):*
> ```Java
>                     .setRecurrenceEndTime(END_OF_TIME)));
>     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.NONPREMIUM);
>     assertThat(recurringEvent.getRenewalPrice().isEmpty()).isTrue();
> ```

Ditto
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 515 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49oAD41WtBcol2ULd:-N-49oAD41WtBcol2ULe:b-89a4l7) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L515)):*
> ```Java
>                     .setRecurrenceEndTime(END_OF_TIME)));
>     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.DEFAULT);
>     assertThat(recurringEvent.getRenewalPrice().isEmpty()).isTrue();
> ```

Ditto
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 542 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49tTF9LQY5KMsdJbS:-N-49tTF9LQY5KMsdJbT:b-agup8d) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L542)):*
> ```Java
>                     .setRecurrenceEndTime(END_OF_TIME)));
>     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.DEFAULT);
>     assertThat(recurringEvent.getRenewalPrice().isEmpty()).isTrue();
> ```

I won't flag the rest of them, but all of them in this file need addressing.
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 678 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-4AflY2aMx0mozFCP-:-N-4AflY2aMx0mozFCP0:b3gnt4n) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L678)):*
> ```Java
>             IllegalArgumentException.class,
>             () ->
>                 persistResource(
> ```

No need to attempt to persist the resource here -- the error is being thrown by the `build()` method.  You can remove this here and from a lot of the rest of the tests.
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 84 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-4AsSL-xhQ5DOcchTi:-N-4AsSM9FzCKQvpj4Nl:b-36nk9f) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/db/src/main/resources/sql/schema/db-schema.sql.generated#L84)):*
> ```
>         renewal_price_amount numeric(19, 2),
>         renewal_price_currency text,
>         renewal_price_behavior text,
> ```

Should be `NOT NULL`.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 130 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-01hNLAdzWtbWmZoPs:-N-3zKKG2SIWuVkiMK8s:b-6osuxv) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L130)):*

Catching an exception only to swallow it as an expected value is generally a code smell.  Also, you're not differentiating *why* `NullPointerException` is being thrown here; it's possible that it's being thrown by something else unexpected, in which case swallowing it is the wrong thing to do.

It's better to explicitly detect the error condition in a conditional.

Also this probably merits a comment explaining what exactly it is doing (namely, it is allowable for a money to be null, but only if the currency unit and the amount are null together).
</blockquote></details>

The goal here is to allow `renewalPrice` as an `Money` object or a `null` object.  If both amount and currency columns contain value, then we return a `Money` object, otherwise a `null` object. However, the override method `nullSafeGet` contains a line that calls `CurrencyUnit.of()`, which throws an error that prevents the object to be constructed. 

`CurrencyUnit.of()` throws 1) NullPointerException for null currency code, which is the case we are dealing with,  2) IllegalCurrencyException for unknown currency. 
If the currency code are not thrown then currency code is returned. 

The existing code actually had a bug because I do not see it returning a null object under the conditional `if (amount==null || currencyUnit == null)` since an error would be thrown before it gets there. 

added explanation and re-arranged the conditionals. 
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 483 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49dpWAR4rgVKAuatE:-N-4L49T6LN9jkp_b6lP:b-nicyga) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L483)):*

Statically import `Truth8.assertThat` and then us `assertThat(...).isEmpty()` .  There's plenty of other examples in our codebase of doing this.  The error messages are better this way.
</blockquote></details>

done and thanks!
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 499 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49nPf9u-8zodBZr_W:-N-4ExJl3eFxW3pgLqrK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L499)):*

Ditto
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 515 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49oAD41WtBcol2ULd:-N-4Ewf3DInvhSozzDZp:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L515)):*

Ditto
</blockquote></details>

Done.
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 542 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-49tTF9LQY5KMsdJbS:-N-4ExV_D_rZ7xraa9XY:b-vzs1sl) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L542)):*

I won't flag the rest of them, but all of them in this file need addressing.
</blockquote></details>

gotcha 
___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 678 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-4AflY2aMx0mozFCP-:-N-4LBoY4kYcCVzInuAv:bfhhjr7) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L678)):*

No need to attempt to persist the resource here -- the error is being thrown by the `build()` method.  You can remove this here and from a lot of the rest of the tests.
</blockquote></details>

ok and did that for all failure test cases. 
___
*[db/src/main/resources/sql/schema/db-schema.sql.generated, line 84 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-4AsSL-xhQ5DOcchTi:-N-4LDAx0kCZjK-jDkub:bmeiihg) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/db/src/main/resources/sql/schema/db-schema.sql.generated#L84)):*

Should be `NOT NULL`.
</blockquote></details>

added the annotation so this is fixed. thanks

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 130 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-01hNLAdzWtbWmZoPs:-N-9Ot-LApvmLV6jGfpS:br3n1wg) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L130)):*

The goal here is to allow `renewalPrice` as an `Money` object or a `null` object.  If both amount and currency columns contain value, then we return a `Money` object, otherwise a `null` object. However, the override method `nullSafeGet` contains a line that calls `CurrencyUnit.of()`, which throws an error that prevents the object to be constructed. 

`CurrencyUnit.of()` throws 1) NullPointerException for null currency code, which is the case we are dealing with,  2) IllegalCurrencyException for unknown currency. 
If the currency code are not thrown then currency code is returned. 

The existing code actually had a bug because I do not see it returning a null object under the conditional `if (amount==null || currencyUnit == null)` since an error would be thrown before it gets there. 

added explanation and re-arranged the conditionals. 
</blockquote></details>

Add relevants tests to `JodaMoneyConverterTest.java`.  If you'd had them you would've already caught this error.
___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 120 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9ObfqC_hxGtJWRJec:-N-9ObfqC_hxGtJWRJed:b-eh39ji) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L120)):*
> ```Java
> 
>   @Override
>   public Object nullSafeGet(
> ```

Should be annotated with ` @Nullable`.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 136 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9P_d352OaxS3zUs9_:-N-9P_d352OaxS3zUs9a:b-f8sn80) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L136)):*
> ```Java
>       return Money.of(CurrencyUnit.of(currencyUnitString), amount.stripTrailingZeros());
>     }
>     throw new HibernateException("Mismatching null state between currency and amount.");
> ```

Does this exception actually get correctly hit in some relevant cases?  What about when `amount` isn't null but `currencyUnit` is -- won't that error with an NPE inside `CurrencyUnit.of(...)` in the above conditional?

The most sensible way to structure this overall seems to be:

```
if both are NULL:
  return null
else if neither are NULL:
  return money
else:
  throw exception
```

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 130 at r8](https://reviewable.io/reviews/google/nomulus/1573#-N-01hNLAdzWtbWmZoPs:-N-9T3DbCFlIE7Fo1XJZ:b-rdi8uy) ([raw file](https://github.com/google/nomulus/blob/4a250973b428ff6bb925d64ceba8ce0d8b22d435/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L130)):*

Add relevants tests to `JodaMoneyConverterTest.java`.  If you'd had them you would've already caught this error.
</blockquote></details>

Agreed if the test cases were written along with the method. There's currently no test cases for these null cases nor for the exceptions. And I'm gonna add them
___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 120 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9ObfqC_hxGtJWRJec:-N-9SwrY-SLTNlM9K6nX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L120)):*

Should be annotated with ` @Nullable`.
</blockquote></details>

Done.
___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 136 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9P_d352OaxS3zUs9_:-N-9VgYvDzh8HgIZ-E2M:bq3wfvp) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L136)):*

Does this exception actually get correctly hit in some relevant cases?  What about when `amount` isn't null but `currencyUnit` is -- won't that error with an NPE inside `CurrencyUnit.of(...)` in the above conditional?

The most sensible way to structure this overall seems to be:

```
if both are NULL:
  return null
else if neither are NULL:
  return money
else:
  throw exception
```
</blockquote></details>

I didn't do the check explicitly since `CurrencyUnit.of()` throws exception when the currenccy string is null or invalid. I cannot think of a case where there's a null state mismatch between currency and amount.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 136 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9P_d352OaxS3zUs9_:-N-9pAdk5IOWJoP9c8uk:b-p8ia7y) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L136)):*

I didn't do the check explicitly since `CurrencyUnit.of()` throws exception when the currenccy string is null or invalid. I cannot think of a case where there's a null state mismatch between currency and amount. 
</blockquote></details>

Actually I figured out a case for the null mismatch and im currently working on the test cases.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 136 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9P_d352OaxS3zUs9_:-N-OZAKa3Nu1PRun2QCz:b-667w46) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L136)):*

> if both are NULL:
>   return null
> else if neither are NULL:
>   return money
> else:
>   throw exception

This still isn't implemented?
</blockquote></details>

I was thinking to just let `CurrentUnit.of()` to handle the errors and keep the original implementation since the case I wanted to take care of originally was when both columns are null.

Anyways, I changed it to the way you asked. 
___
*[core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java, line 196 at r11](https://reviewable.io/reviews/google/nomulus/1573#-N-OAz3F2nzhN7mM5OEf:-N-OIbaV2xcg-3t9gNEo:b-pymnfi) ([raw file](https://github.com/google/nomulus/blob/0697d5df3bfbb03f29342b991b132adc8e0fb07b/core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java#L196)):*

Why are you merely making an assertion about the number of elements rather than their actual contents?  You can make a much stronger assertion here -- `.containsExactly(....)`.  Same for everywhere else.

And then you won't need the subsequent assert at all.
</blockquote></details>

I used existing test cases as reference. The two assertions are doing different things. 
The first assertion you are referring to is to assert the insertion that a new row of data is inserted. The second assertion uses a different SQL query because that's to test the Money object construction, which is why `Select money` is used in the query.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java, line 196 at r11](https://reviewable.io/reviews/google/nomulus/1573#-N-OAz3F2nzhN7mM5OEf:-N-OgMu95JDYEp2vanch:b-co338m) ([raw file](https://github.com/google/nomulus/blob/0697d5df3bfbb03f29342b991b132adc8e0fb07b/core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java#L196)):*

I used existing test cases as reference. The two assertions are doing different things. 
The first assertion you are referring to is to assert the insertion that a new row of data is inserted. The second assertion uses a different SQL query because that's to test the Money object construction, which is why `Select money` is used in the query.  
</blockquote></details>

I added another assertion to verify the added row contains the expected values.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 136 at r9](https://reviewable.io/reviews/google/nomulus/1573#-N-9P_d352OaxS3zUs9_:-N-P6dzvAy4DswZUI6fC:beu8ezi) ([raw file](https://github.com/google/nomulus/blob/c176cd7266bfa1faa91bfa4f507264036ca09fec/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L136)):*

I was thinking to just let `CurrentUnit.of()` to handle the errors and keep the original implementation since the case I wanted to take care of originally was when both columns are null.

Anyways, I changed it to the way you asked. 
</blockquote></details>

It's better if we explicitly handle it though. It makes the code more clear and it lets us throw a better error message. Since the problem is caused by a null value in one place but not in another, it's best to throw an exception indicating that fact.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 137 at r12](https://reviewable.io/reviews/google/nomulus/1573#-N-P6xapAWwuBgFFGMH5:-N-P6xapAWwuBgFFGMH6:b-h14x1g) ([raw file](https://github.com/google/nomulus/blob/b4f7c587fec2f4c7d151795272018fe9fcf39e4d/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L137)):*
> ```Java
>       return Money.of(CurrencyUnit.of(currencyUnitString), amount.stripTrailingZeros());
>     } else {
>       throw new HibernateException("Mismatching null state between currency and amount.");
> ```

This will be clearer (and easier to debug) if you include the values that caused it to fail.  Something like: `"Mismatching null state between currency '%s' and amount '%'"`

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java, line 195 at r12](https://reviewable.io/reviews/google/nomulus/1573#-N-P7J0bAiIoDqUyM_4a:-N-P7J0bAiIoDqUyM_4b:b-o179pa) ([raw file](https://github.com/google/nomulus/blob/b4f7c587fec2f4c7d151795272018fe9fcf39e4d/core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java#L195)):*
> ```Java
>                             "SELECT amount, currency FROM \"TestEntity\" WHERE name = 'id'")
>                         .getResultList());
>     assertThat(result.size()).isEqualTo(1);
> ```

The better way to write this is `assertThat(result).hasSize(1)` (here and everywhere else).  Please read the Truth docs and familiarize yourself with the available methods; you frequently fail to use the correct ones.  This is important as the error messages when tests fail are much better when the correct methods are used.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 137 at r12](https://reviewable.io/reviews/google/nomulus/1573#-N-P6xapAWwuBgFFGMH5:-N-PI5b-5gZhbUy8LWQy:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/b4f7c587fec2f4c7d151795272018fe9fcf39e4d/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L137)):*

This will be clearer (and easier to debug) if you include the values that caused it to fail.  Something like: `"Mismatching null state between currency '%s' and amount '%'"`
</blockquote></details>

done
___
*[core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java, line 195 at r12](https://reviewable.io/reviews/google/nomulus/1573#-N-P7J0bAiIoDqUyM_4a:-N-PJRr255mQrbmE0gYa:bbw6wly) ([raw file](https://github.com/google/nomulus/blob/b4f7c587fec2f4c7d151795272018fe9fcf39e4d/core/src/test/java/google/registry/persistence/converter/JodaMoneyConverterTest.java#L195)):*

The better way to write this is `assertThat(result).hasSize(1)` (here and everywhere else).  Please read the Truth docs and familiarize yourself with the available methods; you frequently fail to use the correct ones.  This is important as the error messages when tests fail are much better when the correct methods are used.
</blockquote></details>

i changed those and the existing cases that use `assertThat(result.size()).isEqualTo(1)` in this file as well.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @mindhog, and @rachelguan)

___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 645 at r13](https://reviewable.io/reviews/google/nomulus/1573#-N-Pbvk19Q-QG5BBlzHu:-N-Pbvk2Cm1zNsrX7txj:b-80edx8) ([raw file](https://github.com/google/nomulus/blob/a1ef169322acdc26f43a9211de3659120a2ea062/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L645)):*
> ```Java
>                     .setRecurrenceEndTime(END_OF_TIME)));
>     assertThat(recurringEvent.getRenewalPriceBehavior()).isEqualTo(RenewalPriceBehavior.SPECIFIED);
>     assertThat(recurringEvent.getRenewalPrice().get()).isEqualTo(Money.of(USD, 100));
> ```

Use `assertThat(optionalVariable).hasValue()` from Truth8 here and elsewhere.  You get a worse error message (and potential NPE) from manually calling `.get()`.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[core/src/test/java/google/registry/model/billing/BillingEventTest.java, line 645 at r13](https://reviewable.io/reviews/google/nomulus/1573#-N-Pbvk19Q-QG5BBlzHu:-N-PmVSK91LFz5XS3LAE:br3950n) ([raw file](https://github.com/google/nomulus/blob/a1ef169322acdc26f43a9211de3659120a2ea062/core/src/test/java/google/registry/model/billing/BillingEventTest.java#L645)):*

Use `assertThat(optionalVariable).hasValue()` from Truth8 here and elsewhere.  You get a worse error message (and potential NPE) from manually calling `.get()`.
</blockquote></details>

replaced.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573#-:-N-PoOQjCLat0mdIwf69:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @CydeWeys and @mindhog)

___
*[core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java, line 137 at r12](https://reviewable.io/reviews/google/nomulus/1573#-N-P6xapAWwuBgFFGMH5:-N-PoEVbE7LbHAv3g8Pw:bwejukh) ([raw file](https://github.com/google/nomulus/blob/b4f7c587fec2f4c7d151795272018fe9fcf39e4d/core/src/main/java/google/registry/persistence/converter/JodaMoneyType.java#L137)):*

done
</blockquote></details>

You added a period to my suggestion -- exception messages shouldn't generally have terminal periods.

rachelguan (commented):
The period actually came from the original exception message and I removed the period in both `nullSafeGet()` methods. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1573#-:-N-PtE0136rOLV_weJNx:b-msz3sf)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @CydeWeys and @mindhog)

PR #1574, opened 2022-03-29
Make a best effort guess on the RDE folder name (prefix) when not provided.

jianglai:
We have a cron job that runs the RDE upload action every 4 hours for all
TLD. Normally this should be a no-op beacuse a RDE upload is scheduled
after RDE staging is completed, and when it fails with non-2XX status it
will retry. However if for some reason it failed due to 20X status (like
waiting for the SFTP cursor), it will not retry but rely on the cron job to
catch up.

With the BEAM RDE pipeline every staging job saves all its deposits in a
uniquely named folder to avoid the need to use a lock, which is not
practical in BEAM. However the cron job has no way of knowing what the
prefixes are for each TLD so it will fail in SQL mode.

In this PR we implemented a logic to guess what the prefix should be and
use it, if we are in SQL mode and a prefix is not provided.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1574)

### DIFF
--- a/core/src/main/java/google/registry/rde/RdeUploadAction.java
+++ b/core/src/main/java/google/registry/rde/RdeUploadAction.java
@@ -32,6 +32,7 @@ import static java.util.Arrays.asList;
 import com.google.cloud.storage.BlobId;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Ordering;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.io.ByteStreams;
 import com.jcraft.jsch.JSch;
@@ -122,7 +123,6 @@ public final class RdeUploadAction implements Runnable, EscrowTask {
   @Inject @Key("rdeSigningKey") PGPKeyPair signingKey;
   @Inject @Key("rdeStagingDecryptionKey") PGPPrivateKey stagingDecryptionKey;
   @Inject RdeUploadAction() {}
-
   @Override
   public void run() {
     logger.atInfo().log("Attempting to acquire RDE upload lock for TLD '%s'.", tld);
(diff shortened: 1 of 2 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1574)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[core/src/main/java/google/registry/rde/RdeUploadAction.java, line 155 at r2](https://reviewable.io/reviews/google/nomulus/1574#-MzNoGZC6kWg7AkRRdXX:-MzNoGZC6kWg7AkRRdXY:b-2bzli9) ([raw file](https://github.com/google/nomulus/blob/5a3b5d2ba24dc95b4c1239dc351a8f52eb61b6ca/core/src/main/java/google/registry/rde/RdeUploadAction.java#L155)):*
> ```Java
>               .sorted()
>               .reduce((first, second) -> second)
>               .orElse(null);
> ```

The following may be more readable:

.max(Comparator.naturalOrder()) or
.max(String::compare) 



_Code quote:_
```Java
              .sorted()
              .reduce((first, second) -> second)
              .orElse(null);
```

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1574)_ status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)
> 
> _[core/src/main/java/google/registry/rde/RdeUploadAction.java, line 155 at r2](https://reviewable.io/reviews/google/nomulus/1574#-MzNoGZC6kWg7AkRRdXX:-MzNoGZC6kWg7AkRRdXY:b-2bzli9) ([raw file](https://github.com/google/nomulus/blob/5a3b5d2ba24dc95b4c1239dc351a8f52eb61b6ca/core/src/main/java/google/registry/rde/RdeUploadAction.java#L155)):_
> 
> > ```java
> >               .sorted()
> >               .reduce((first, second) -> second)
> >               .orElse(null);
> > ```
> 
> The following may be more readable:
> 
> .max(Comparator.naturalOrder()) or .max(String::compare)
> 
> _Code quote:_
> 
> ```java
>               .sorted()
>               .reduce((first, second) -> second)
>               .orElse(null);
> ```

Good point. Done.

weiminyu (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1574)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

PR #1586, opened 2022-04-09
Build Java8-compatible release

weiminyu:
Use the new options.release Gradle property to make sure builds are
compatible with Java 8, which is the runtime on Appengine.

This new property replaces sourceCompatibility, targetCompatibility, and
bootclasspath (wasn't previously set, which is the reason why we
couldn't detect Java9 api usage when building).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1586)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -331,6 +331,18 @@ subprojects {
 
   apply from: "${rootDir.path}/java_common.gradle"
 
+  if (project.name == 'docs') {
+    compileJava {
+      // TODO: Remove this once we are on Java 11 across the board.
+      options.release = 11
+    }
+  } else {
+    compileJava {
+      // TODO: Remove this once we migrate off AppEngine.
+      options.release = 8
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


jianglai:
You have the following error when compiling:

```
> Task :docs:compileJava FAILED
error: exporting a package from system module jdk.compiler is not allowed with --release
error: exporting a package from system module jdk.compiler is not allowed with --release
error: exporting a package from system module jdk.javadoc is not allowed with --release
3 errors
```

weiminyu (commented):
Fixed. PTAL

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1586#-:-N-Dh2b2BFNoBiW8pvsh:b6n7j80)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @gbrodman, @jianglai, and @mindhog)

jianglai (approved):

PR #1587, opened 2022-04-11
fdosRemove dos.xml from the configs

weiminyu:
We don't have dos config right now, and applying dos from "gcloud app
deploy" is deprecated and has started causing problems.

If we add dos configs, it should be using "gcloud app firewall-rules".

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1587)


jianglai:
You have some unintended characters in the PR title.

jianglai (approved):

PR #1592, opened 2022-04-13
Add renewal support for AT and internal registrations in DomainRenewFlow

rachelguan:
The purpose of this PR is to add renewal support for anchor tenants and internal registrations in the renew flow. The calculation was updated from a different PR, [PR#1610](https://github.com/google/nomulus/pull/1610). The helpers from `DomainFlowUtils` and `DomainPricingLogic` will be removed once [PR#1610](https://github.com/google/nomulus/pull/1610) is merged. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1592)

### DIFF
--- a/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
+++ b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
@@ -28,7 +28,7 @@ import static google.registry.persistence.transaction.QueryComposer.Comparator.E
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
-import static google.registry.pricing.PricingEngineProxy.getDomainRenewPrice;
+import static google.registry.pricing.PricingEngineProxy.getDomainRenewCost;
 import static google.registry.util.CollectionUtils.union;
 import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static google.registry.util.DateTimeUtils.earliestOf;
@@ -70,6 +70,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
+import org.joda.money.Money;
 import org.joda.time.DateTime;
 
 /**
@@ -430,12 +431,13 @@ public class ExpandRecurringBillingEventsAction implements Runnable {
       historyEntriesBuilder.add(historyEntry);
 
       DateTime eventTime = billingTime.minus(tld.getAutoRenewGracePeriodLength());
+      // Determine the cost for a one-year renewal.
+      Money renewCost = getDomainRenewCost(recurring.getTargetId(), eventTime, 1);
       syntheticOneTimesBuilder.add(
           new OneTime.Builder()
               .setBillingTime(billingTime)
               .setRegistrarId(recurring.getRegistrarId())
-              // Determine the cost for a one-year renewal.
-              .setCost(getDomainRenewPrice(recurring.getTargetId(), eventTime, recurring))
+              .setCost(renewCost)
               .setEventTime(eventTime)
               .setFlags(union(recurring.getFlags(), Flag.SYNTHETIC))
               .setParent(historyEntry)
--- a/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
+++ b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
@@ -28,7 +28,7 @@ import static google.registry.persistence.transaction.QueryComposer.Comparator.E
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
-import static google.registry.pricing.PricingEngineProxy.getDomainRenewCost;
+import static google.registry.pricing.PricingEngineProxy.getDomainRenewPrice;
 import static google.registry.util.CollectionUtils.union;
 import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static google.registry.util.DateTimeUtils.earliestOf;
@@ -70,7 +70,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
-import org.joda.money.Money;
 import org.joda.time.DateTime;
 
 /**
@@ -431,13 +430,12 @@ public class ExpandRecurringBillingEventsAction implements Runnable {
       historyEntriesBuilder.add(historyEntry);
 
       DateTime eventTime = billingTime.minus(tld.getAutoRenewGracePeriodLength());
-      // Determine the cost for a one-year renewal.
-      Money renewCost = getDomainRenewCost(recurring.getTargetId(), eventTime, 1);
       syntheticOneTimesBuilder.add(
           new OneTime.Builder()
               .setBillingTime(billingTime)
               .setRegistrarId(recurring.getRegistrarId())
-              .setCost(renewCost)
+              // Determine the cost for a one-year renewal.
+              .setCost(getDomainRenewPrice(recurring.getTargetId(), eventTime, recurring))
               .setEventTime(eventTime)
               .setFlags(union(recurring.getFlags(), Flag.SYNTHETIC))
               .setParent(historyEntry)
--- a/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
@@ -36,6 +36,7 @@ import google.registry.model.domain.fee.Fee;
 import google.registry.model.domain.token.AllocationToken;
 import google.registry.model.pricing.PremiumPricingEngine.DomainPrices;
 import google.registry.model.tld.Registry;
+
 import java.math.RoundingMode;
 import java.util.Optional;
 import javax.annotation.Nullable;
@@ -104,6 +105,7 @@ public final class DomainPricingLogic {
             .build());
   }
 
+
   /** Returns a new renewal cost for the pricer. */
   FeesAndCredits getRenewPrice(
       Registry registry,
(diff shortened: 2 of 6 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 683 at r4](https://reviewable.io/reviews/google/nomulus/1592#-N1Q-H0q4dA1FjtXJXQe:-N1Q-H0r2xhdwBzKMdlp:b-wmtkdk) ([raw file](https://github.com/google/nomulus/blob/6171c8d06170d7904459bc0e2091bfb7f8608247/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L683)):*
> ```Java
>       case RENEW:
>         builder.setAvailIfSupported(true);
>         // TODO(rachelguan): remove after PR#1610 for renewal cost calculation gets merged
> ```

I'm confused. This PR has TODO dependencies on #1592, but then also vice-versa?  Which of these 2 PRs is going in first?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 683 at r4](https://reviewable.io/reviews/google/nomulus/1592#-N1Q-H0q4dA1FjtXJXQe:-N2IP03e6kYA_SaU50Ty:bwz6mto) ([raw file](https://github.com/google/nomulus/blob/6171c8d06170d7904459bc0e2091bfb7f8608247/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L683)):*

I'm confused. This PR has TODO dependencies on #1592, but then also vice-versa?  Which of these 2 PRs is going in first?
</blockquote></details>

So the action file and the renew flow are both depending on the new method for the renew price, therefore this PR (1592) should go in first..There's already a `getRenewPrice` method in `DomainPricingLogic` and this is used in the renew flow. Therefore, there's going to be two `goRenewPrice` in `DomainPricingLogic` until the PR for renew flow modification goes in.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 683 at r4](https://reviewable.io/reviews/google/nomulus/1592#-N1Q-H0q4dA1FjtXJXQe:-N2ISsnbAELhnoD45lSB:b-s3s0nz) ([raw file](https://github.com/google/nomulus/blob/6171c8d06170d7904459bc0e2091bfb7f8608247/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L683)):*

So the action file and the renew flow are both depending on the new method for the renew price, therefore this PR (1592) should go in first..There's already a `getRenewPrice` method in `DomainPricingLogic` and this is used in the renew flow. Therefore, there's going to be two `goRenewPrice` in `DomainPricingLogic` until the PR for renew flow modification goes in. 
</blockquote></details>

I thought this was a message for the PR for the`DomainPricingLogic`.. here's the situation: 

So the action file and the renew flow are both depending on the new method for the renew price, therefore the PR for domain pricing logic (1610) should go in first..There's already a getRenewPrice method in DomainPricingLogic and this is used in the renew flow. Therefore, there's going to be two goRenewPrice in DomainPricingLogic after the merge of 1610...This PR then only would include renew flow related modification.. not renew price calculation.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 186 at r5](https://reviewable.io/reviews/google/nomulus/1592#-N36QVcJ0COb7HF8NuBv:-N36QVcJ0COb7HF8NuBw:bxweso) ([raw file](https://github.com/google/nomulus/blob/5aee82d8716f0d3c302ff6cc95bff11d705f9a7d/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L186)):*
> ```Java
>             .setRenewalPrice(
>                 existingRecurringBillingEvent.getRenewalPrice().isPresent()
>                     ? existingRecurringBillingEvent.getRenewalPrice().get()
> ```

Is this correct?  Doesn't it need to check the renewal price behavior too?  It seems to me that `NONPREMIUM` is not handled correctly here?
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 394 at r5](https://reviewable.io/reviews/google/nomulus/1592#-N36QxuW0OVazrs6oEl1:-N36QxuW0OVazrs6oEl2:b-5e4qb5) ([raw file](https://github.com/google/nomulus/blob/5aee82d8716f0d3c302ff6cc95bff11d705f9a7d/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L394)):*
> ```Java
>         "NewRegistrar",
>         UserPrivileges.SUPERUSER,
>         ImmutableMap.of("DOMAIN", "example.tld", "EXDATE", "2005-04-03T22:00:00.0Z"),
> ```

I don't think this test does what you think it does, as `example.tld` isn't a premium domain name.  I think this is only working by accident.

Use a domain name that is on the premium list for this check, and ensure that the `NONPREMIUM` behavior is *overriding* the premium price.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 186 at r5](https://reviewable.io/reviews/google/nomulus/1592#-N36QVcJ0COb7HF8NuBv:-N36SDmfDlJ-lJsuPkp2:btizeuu) ([raw file](https://github.com/google/nomulus/blob/5aee82d8716f0d3c302ff6cc95bff11d705f9a7d/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L186)):*

Is this correct?  Doesn't it need to check the renewal price behavior too?  It seems to me that `NONPREMIUM` is not handled correctly here?
</blockquote></details>

Why? Renewal price exists only if renewal price behavior is `SPECIFIED`. For `NONPREMIUM`, the renewal price will be null. This is only for the billing event. The recurring billing event should be the same as the previous one, right?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 394 at r5](https://reviewable.io/reviews/google/nomulus/1592#-N36QxuW0OVazrs6oEl1:-N36ofwU0zrGzQHptQ0N:b-cy5muy) ([raw file](https://github.com/google/nomulus/blob/5aee82d8716f0d3c302ff6cc95bff11d705f9a7d/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L394)):*

I don't think this test does what you think it does, as `example.tld` isn't a premium domain name.  I think this is only working by accident.

Use a domain name that is on the premium list for this check, and ensure that the `NONPREMIUM` behavior is *overriding* the premium price.
</blockquote></details>

added new test cases

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 173 at r6](https://reviewable.io/reviews/google/nomulus/1592#-N3Padwh62UoC9eY8A_Y:-N3Padwh62UoC9eY8A_Z:bthswzp) ([raw file](https://github.com/google/nomulus/blob/d6aa4516e68f97d00101767eb906e753c4dc6a89/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L173)):*
> ```Java
>     // if the renewal price behavior is SPECIFIED, it means the renewal price comes from the
>     // existing billing event regardless of the status of the domain being premium or standard.
>     if (existingRecurringBillingEvent.getRenewalPriceBehavior() == DEFAULT) {
> ```

I don't think this is right.  We should always be validating the fee challenge.  If the registrar passes it in, and it is wrong, then an error needs to be returned.  Otherwise something is happening that they are not expecting.

E.g. if a domain is premium, and they send a fee of $100 on the renew, but the cost of the renew is actually something other than $100 because of `NONPREMIUM` or `SPECIFIED`, then we need to return an error here, and they need to either send a renew with the correct value of the fee in the fee extension, or in the case of a `NONPREMIUM` renewal, they don't need to send the fee extension at all.  But if they do send it, the values in it need to be correct, and the only case in which they can omit it is in the case of a standard priced renewal.

CydeWeys (commented):
Doesn't need to be part of this PR, but we're going to want some lifecycle tests (in `EppLifecycleDomainTest`) that test situations like:

1. A domain is created with special renewal behavior, is then manually renewed, and then is manually renewed *again* still at the correct price behavior (this tests that the attributes on the recurrence are copied over correctly on renew).  With some domain checks in there as well.
2. A domain is created with special renewal behavior, is deleted, and is then restored, and still has the correct renewal price behavior.
3. A domain is created with special renewal behavior, is transferred, and still has the correct renewal price behavior.

Heck, all of those could be combined into a single mega lifecycle test ... domain is created, then renewed, then transferred, then deleted, then restored, then renewed again, all with checks interspersed.  Am I forgetting any other commands that affect recurrences and need to have this data copied over?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592#-:-N3QY--uEt1NOgEnNG9F:b4mcctp)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @CydeWeys)

rachelguan (commented):
> Doesn't need to be part of this PR, but we're going to want some lifecycle tests (in EppLifecycleDomainTest) that test situations like:
> 
> A domain is created with special renewal behavior, is then manually renewed, and then is manually renewed again still at the correct price behavior (this tests that the attributes on the recurrence are copied over correctly on renew).  With some domain checks in there as well.
> A domain is created with special renewal behavior, is deleted, and is then restored, and still has the correct renewal price behavior.
> A domain is created with special renewal behavior, is transferred, and still has the correct renewal price behavior.
> 
> Heck, all of those could be combined into a single mega lifecycle test ... domain is created, then renewed, then transferred, then deleted, then restored, then renewed again, all with checks interspersed.  Am I forgetting any other commands that affect recurrences and need to have this data copied over?
>           1 day ago
>         CydeWeysBen McIlwain

created a bug: 
 http://b/234637061 life cycle tests in EppLifecycleDomainTest for anchor tenants and internal registrations 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1592#-:-N3W6y-47fXJGrqZ1QGx:b-ba5w0x)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1596, opened 2022-04-19
Add renewal logic in allocation token related commands 

rachelguan:
The purpose of this PR is to set up the renewal behavior in allocation token related commands, which will be used in domain create flow. 

The changes to `GenerateAllocationTokensCommand` and `UpdateAllocationTokensCommand` are in this PR since they are very similar. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1596)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -373,6 +373,9 @@ public class AllocationToken extends BackupGroupRoot implements Buildable, Datas
     }
 
     public Builder setRenewalPriceBehavior(RenewalPriceBehavior renewalPriceBehavior) {
+      checkArgument(
+          renewalPriceBehavior != RenewalPriceBehavior.NONPREMIUM,
+          "NONPREMIUM is not a supported renewal price behavior in allocation token");
       getInstance().renewalPriceBehavior = renewalPriceBehavior;
       return this;
     }
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -373,9 +373,6 @@ public class AllocationToken extends BackupGroupRoot implements Buildable, Datas
     }
 
     public Builder setRenewalPriceBehavior(RenewalPriceBehavior renewalPriceBehavior) {
-      checkArgument(
-          renewalPriceBehavior != RenewalPriceBehavior.NONPREMIUM,
-          "NONPREMIUM is not a supported renewal price behavior in allocation token");
       getInstance().renewalPriceBehavior = renewalPriceBehavior;
       return this;
     }
--- a/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java
@@ -25,6 +25,7 @@ import static google.registry.util.CollectionUtils.nullToEmpty;
 import static google.registry.util.StringGenerator.DEFAULT_PASSWORD_LENGTH;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
+import avro.shaded.com.google.common.base.Ascii;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import com.google.appengine.tools.remoteapi.RemoteApiException;
@@ -37,6 +38,7 @@ import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
 import com.google.common.io.Files;
+import google.registry.model.billing.BillingEvent.RenewalPriceBehavior;
 import google.registry.model.domain.token.AllocationToken;
 import google.registry.model.domain.token.AllocationToken.TokenStatus;
 import google.registry.model.domain.token.AllocationToken.TokenType;
@@ -142,6 +144,13 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
   private ImmutableSortedMap<DateTime, TokenStatus> tokenStatusTransitions;
 
   @Parameter(
+      names = {"--renewal_price_behavior"},
+      description =
+          "Type of renewal price behavior, either DEFAULT (default) or SPECIFIED. This indicates"
+              + " how a domain should be charged for renewal.")
+  private String renewalPriceBehavior;
+
+  @Parameter(
       names = {"--dry_run"},
       description = "Do not actually persist the tokens; defaults to false")
   boolean dryRun;
@@ -174,7 +183,6 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
               .collect(Collectors.toCollection(ArrayDeque::new));
       numTokens = domainNames.size();
     }
-
     int tokensSaved = 0;
     do {
       ImmutableSet<AllocationToken> tokens =
@@ -184,6 +192,7 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
                     AllocationToken.Builder token =
                         new AllocationToken.Builder()
                             .setToken(t)
+                            .setRenewalPriceBehavior(getRenewalPriceBehavior(renewalPriceBehavior))
                             .setTokenType(tokenType == null ? SINGLE_USE : tokenType)
                             .setAllowedRegistrarIds(
                                 ImmutableSet.copyOf(nullToEmpty(allowedClientIds)))
@@ -299,5 +308,18 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
             tm().loadByKeysIfPresent(existingTokenKeys).values().stream()
                 .map(AllocationToken::getToken)
                 .collect(toImmutableSet()));
+  }
+
+  private RenewalPriceBehavior getRenewalPriceBehavior(String renewalPriceBehaviorStr) {
+    if (renewalPriceBehaviorStr == null) {
+      return RenewalPriceBehavior.DEFAULT;
+    } else {
+      try {
+        return RenewalPriceBehavior.valueOf(Ascii.toUpperCase(renewalPriceBehaviorStr));
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
+            String.format("Invalid renewal price behavior: '%s'", renewalPriceBehaviorStr));
+      }
+    }
   }
 }
--- a/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java
@@ -25,7 +25,6 @@ import static google.registry.util.CollectionUtils.nullToEmpty;
 import static google.registry.util.StringGenerator.DEFAULT_PASSWORD_LENGTH;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-import avro.shaded.com.google.common.base.Ascii;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import com.google.appengine.tools.remoteapi.RemoteApiException;
@@ -146,9 +145,9 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
   @Parameter(
       names = {"--renewal_price_behavior"},
       description =
-          "Type of renewal price behavior, either DEFAULT (default) or SPECIFIED. This indicates"
-              + " how a domain should be charged for renewal.")
-  private String renewalPriceBehavior;
+          "Type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
+              + " This indicates how a domain should be charged for renewal.")
+  private RenewalPriceBehavior renewalPriceBehavior = RenewalPriceBehavior.DEFAULT;
 
   @Parameter(
       names = {"--dry_run"},
@@ -183,6 +182,7 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
               .collect(Collectors.toCollection(ArrayDeque::new));
       numTokens = domainNames.size();
     }
+
     int tokensSaved = 0;
     do {
       ImmutableSet<AllocationToken> tokens =
@@ -192,7 +192,7 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
                     AllocationToken.Builder token =
                         new AllocationToken.Builder()
                             .setToken(t)
-                            .setRenewalPriceBehavior(getRenewalPriceBehavior(renewalPriceBehavior))
+                            .setRenewalPriceBehavior(renewalPriceBehavior)
                             .setTokenType(tokenType == null ? SINGLE_USE : tokenType)
                             .setAllowedRegistrarIds(
                                 ImmutableSet.copyOf(nullToEmpty(allowedClientIds)))
@@ -308,18 +308,5 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
             tm().loadByKeysIfPresent(existingTokenKeys).values().stream()
                 .map(AllocationToken::getToken)
                 .collect(toImmutableSet()));
-  }
-
-  private RenewalPriceBehavior getRenewalPriceBehavior(String renewalPriceBehaviorStr) {
-    if (renewalPriceBehaviorStr == null) {
-      return RenewalPriceBehavior.DEFAULT;
-    } else {
-      try {
-        return RenewalPriceBehavior.valueOf(Ascii.toUpperCase(renewalPriceBehaviorStr));
-      } catch (IllegalArgumentException e) {
-        throw new IllegalArgumentException(
-            String.format("Invalid renewal price behavior: '%s'", renewalPriceBehaviorStr));
-      }
-    }
   }
 }
--- a/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java
+++ b/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java
@@ -17,6 +17,7 @@ package google.registry.tools;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.Sets.difference;
+import static google.registry.model.billing.BillingEvent.RenewalPriceBehavior.DEFAULT;
 import static google.registry.model.domain.token.AllocationToken.TokenType.SINGLE_USE;
 import static google.registry.model.domain.token.AllocationToken.TokenType.UNLIMITED_USE;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
@@ -145,9 +146,12 @@ class GenerateAllocationTokensCommand implements CommandWithRemoteApi {
   @Parameter(
       names = {"--renewal_price_behavior"},
       description =
-          "Type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
-              + " This indicates how a domain should be charged for renewal.")
-  private RenewalPriceBehavior renewalPriceBehavior = RenewalPriceBehavior.DEFAULT;
+          "The type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
+              + " This indicates how a domain should be charged for renewal. By default, a domain"
+              + " will be renewed at the renewal price from the pricing engine. If the renewal"
+              + " price behavior is set to SPECIFIED, it means that the renewal cost will be the"
+              + " same as the domain's calculated create price.")
+  private RenewalPriceBehavior renewalPriceBehavior = DEFAULT;
 
   @Parameter(
       names = {"--dry_run"},
--- a/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java
+++ b/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java
@@ -21,9 +21,9 @@ import static com.google.common.collect.Streams.stream;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 
-import avro.shaded.com.google.common.base.Ascii;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
+import com.beust.jcommander.internal.Nullable;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -93,9 +93,10 @@ final class UpdateAllocationTokensCommand extends UpdateOrDeleteAllocationTokens
   @Parameter(
       names = {"--renewal_price_behavior"},
       description =
-          "Type of renewal price behavior, either DEFAULT (default) or SPECIFIED. This indicates"
-              + " how a domain should be charged for renewal.")
-  private String renewalPriceBehavior;
+          "Type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
+              + " This indicates how a domain should be charged for renewal.")
+  @Nullable
+  private RenewalPriceBehavior renewalPriceBehavior;
 
   private static final int BATCH_SIZE = 20;
   private static final Joiner JOINER = Joiner.on(", ");
@@ -152,17 +153,8 @@ final class UpdateAllocationTokensCommand extends UpdateOrDeleteAllocationTokens
     Optional.ofNullable(discountYears).ifPresent(builder::setDiscountYears);
     Optional.ofNullable(tokenStatusTransitions).ifPresent(builder::setTokenStatusTransitions);
     Optional.ofNullable(renewalPriceBehavior)
-        .ifPresent(behavior -> builder.setRenewalPriceBehavior(getRenewalPriceBehavior(behavior)));
+        .ifPresent(behavior -> builder.setRenewalPriceBehavior(renewalPriceBehavior));
     return builder.build();
-  }
-
-  private RenewalPriceBehavior getRenewalPriceBehavior(String renewalPriceBehaviorStr) {
-    try {
-      return RenewalPriceBehavior.valueOf(Ascii.toUpperCase(renewalPriceBehaviorStr));
-    } catch (IllegalArgumentException e) {
-      throw new IllegalArgumentException(
-          String.format("Invalid renewal price behavior: '%s'", renewalPriceBehaviorStr));
-    }
   }
 
   private long saveBatch(ImmutableList<AllocationToken> batch) {
--- a/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java
+++ b/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java
@@ -21,12 +21,14 @@ import static com.google.common.collect.Streams.stream;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 
+import avro.shaded.com.google.common.base.Ascii;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
+import google.registry.model.billing.BillingEvent.RenewalPriceBehavior;
 import google.registry.model.domain.token.AllocationToken;
 import google.registry.model.domain.token.AllocationToken.TokenStatus;
 import google.registry.tools.params.TransitionListParameter.TokenStatusTransitions;
@@ -88,6 +90,13 @@ final class UpdateAllocationTokensCommand extends UpdateOrDeleteAllocationTokens
               + "form <time>=<status>[,<time>=<status>]* where each status represents the status.")
   private ImmutableSortedMap<DateTime, TokenStatus> tokenStatusTransitions;
 
+  @Parameter(
+      names = {"--renewal_price_behavior"},
+      description =
+          "Type of renewal price behavior, either DEFAULT (default) or SPECIFIED. This indicates"
+              + " how a domain should be charged for renewal.")
+  private String renewalPriceBehavior;
+
   private static final int BATCH_SIZE = 20;
   private static final Joiner JOINER = Joiner.on(", ");
 
@@ -142,7 +151,18 @@ final class UpdateAllocationTokensCommand extends UpdateOrDeleteAllocationTokens
     Optional.ofNullable(discountPremiums).ifPresent(builder::setDiscountPremiums);
     Optional.ofNullable(discountYears).ifPresent(builder::setDiscountYears);
     Optional.ofNullable(tokenStatusTransitions).ifPresent(builder::setTokenStatusTransitions);
+    Optional.ofNullable(renewalPriceBehavior)
+        .ifPresent(behavior -> builder.setRenewalPriceBehavior(getRenewalPriceBehavior(behavior)));
     return builder.build();
+  }
+
+  private RenewalPriceBehavior getRenewalPriceBehavior(String renewalPriceBehaviorStr) {
+    try {
+      return RenewalPriceBehavior.valueOf(Ascii.toUpperCase(renewalPriceBehaviorStr));
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(
+          String.format("Invalid renewal price behavior: '%s'", renewalPriceBehaviorStr));
+    }
   }
 
   private long saveBatch(ImmutableList<AllocationToken> batch) {
--- a/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java
+++ b/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java
@@ -93,8 +93,11 @@ final class UpdateAllocationTokensCommand extends UpdateOrDeleteAllocationTokens
   @Parameter(
       names = {"--renewal_price_behavior"},
       description =
-          "Type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
-              + " This indicates how a domain should be charged for renewal.")
+          "The type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
+              + " This indicates how a domain should be charged for renewal. By default, a domain"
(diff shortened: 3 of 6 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[core/src/main/java/google/registry/model/domain/token/AllocationToken.java, line 378 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXLBt4FDgV7Jy5TiS:-N0CXLBt4FDgV7Jy5TiT:b8b25vt) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L378)):*
> ```Java
>       checkArgument(
>           renewalPriceBehavior != RenewalPriceBehavior.NONPREMIUM,
>           "NONPREMIUM is not a supported renewal price behavior in allocation token");
> ```

Why not?
___
*[core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java, line 151 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXUEO49fepDFgrL6X:-N0CXUEO49fepDFgrL6Y:b-2x06j1) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L151)):*
> ```Java
>           "Type of renewal price behavior, either DEFAULT (default) or SPECIFIED. This indicates"
>               + " how a domain should be charged for renewal.")
>   private String renewalPriceBehavior;
> ```

Use the enum type directly, not a String.
___
*[core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXdDLDe3wgKxvg3zW:-N0CXdDLDe3wgKxvg3zX:b-y3w0fs) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L98)):*
> ```Java
>           "Type of renewal price behavior, either DEFAULT (default) or SPECIFIED. This indicates"
>               + " how a domain should be charged for renewal.")
>   private String renewalPriceBehavior;
> ```

Use the enum type directly here.
___
*[core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java, line 245 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXptvBEsmJ44JvSVR:-N0CXptvBEsmJ44JvSVS:b-j93p7k) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L245)):*
> ```Java
>     AllocationToken token =
>         persistResource(
>             builderWithPromo(RenewalPriceBehavior.DEFAULT)
> ```

Rather than calling `DEFAULT` in a huge number of places, sometimes it's easier to just let the default actually be the default, and then only override it in the few cases that you need to.  (Same for the generate tests.)

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java, line 151 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXUEO49fepDFgrL6X:-N0CcPfv4f2s4rJIreu2:baiofxw) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L151)):*

Use the enum type directly, not a String.
</blockquote></details>

I mentioned in the comment that I chose String over enum because `RenewalPriceBehavior.valueOf()` does the magic which is great but if there's an issue with the input..it generates an error and the error message just states the acceptable values..and I added the check the disallow setting the behavior to `NONPREMIUM`..if that's the case the the error message shouldn't show a possible value that's not even allowed . 

Below was what I wrote in the description, just for the record on why I used `String` and a check on `NONPREMIUM` 

"
There are three states in `RenewalPriceToken` but `AllocationToken` should only accept `DEFAULT` and `SPECIFIED`, at least for now, since there's no use case for `NONPREMIUM` yet. Therefore, a check to see if the behavior is `NONPREMIUM` is added in `setAllocationToken` of `AllocationToken`. 


The input type for `renewalPriceBehavior` is `String` instead of `RenewalPriceBehavior` for better error handling. The default `RenewalPriceBehavior.valueOf()` error message shows the allowed values which contains `NONPREMIUM`, which is not supported in `AllocationToken`. We also don't want to expose the full path of `RenewalPriceBehavior` in the error message. 
 "
___
*[core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java, line 98 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXdDLDe3wgKxvg3zW:-N0Ccuh_AeQTQiWOLn8c:bognxb1) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L98)):*

Use the enum type directly here.
</blockquote></details>

provided explanation above and in the description. I could definitely change it..
___
*[core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java, line 245 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXptvBEsmJ44JvSVR:-N0Cd0D21O5uxbfMLNgd:b-4cfe3e) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L245)):*

Rather than calling `DEFAULT` in a huge number of places, sometimes it's easier to just let the default actually be the default, and then only override it in the few cases that you need to.  (Same for the generate tests.)
</blockquote></details>

hmm..i saw an example previously where there's two methods that were essentially the same except one takes in one additional parameter. I thought it would be nice to see what parameters are being passed in which allows easier comparison..
___
*[core/src/main/java/google/registry/model/domain/token/AllocationToken.java, line 378 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXLBt4FDgV7Jy5TiS:-N0Cbnt15sR6YOb--0gV:bc0xns1) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L378)):*

Why not?
</blockquote></details>

Because there's no use case yet? Anchor Tenant status doesn't come from allocation token. If there's use case for that, then this should be removed. It's just a check to make sure only the cases (so far just DEFAULT and SPECIFIED) are allowed . 

I gave my reason in one of the comments. I changed it though.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java, line 151 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXUEO49fepDFgrL6X:-N0GvAWn5GgunTDLVwwj:btryw92) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L151)):*

I mentioned in the comment that I chose String over enum because `RenewalPriceBehavior.valueOf()` does the magic which is great but if there's an issue with the input..it generates an error and the error message just states the acceptable values..and I added the check the disallow setting the behavior to `NONPREMIUM`..if that's the case the the error message shouldn't show a possible value that's not even allowed . 

Below was what I wrote in the description, just for the record on why I used `String` and a check on `NONPREMIUM` 

"
There are three states in `RenewalPriceToken` but `AllocationToken` should only accept `DEFAULT` and `SPECIFIED`, at least for now, since there's no use case for `NONPREMIUM` yet. Therefore, a check to see if the behavior is `NONPREMIUM` is added in `setAllocationToken` of `AllocationToken`. 


The input type for `renewalPriceBehavior` is `String` instead of `RenewalPriceBehavior` for better error handling. The default `RenewalPriceBehavior.valueOf()` error message shows the allowed values which contains `NONPREMIUM`, which is not supported in `AllocationToken`. We also don't want to expose the full path of `RenewalPriceBehavior` in the error message. 
 "
</blockquote></details>

Yeah, see, this implementation is much simpler and nicer.
___
*[core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java, line 245 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXptvBEsmJ44JvSVR:-N0Gw3Oy2r5u_qTFvbbs:bp1dg1q) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L245)):*

hmm..i saw an example previously where there's two methods that were essentially the same except one takes in one additional parameter. I thought it would be nice to see what parameters are being passed in which allows easier comparison..
</blockquote></details>

Overloads are fine.  You haven't really implemented it correctly below though -- the overloaded methods should be calling each other, not implementing everything from scratch each time.  You also potentially don't need overloads at all; just build it with default values and then explicitly set the 1 thing under test at callsites (this 1 overload for the renewal price behavior is fine though).

When one parameter has the same default value 90+% of the time, adding it in at all the callsites actually makes it less clear what's going on, because it's reducing the signal-to-noise ratio.  Imagine a hypothetical scenario where an object under test has many different fields.  Which of these is clearer to show what's really going on, always splatting out every single field at every callsite, or just showing the one field of interest that's different for each test case?

```
test1() {
  createTestObject("abcd", DEFAULT, FOO_ENUM_PARAM, POST, 135135, STANDARD, SPECIFIED);
  .....
}

test2() {
  createTestObject("abcd", DEFAULT, FOO_ENUM_PARAM, POST, 135135, PREMIUM, SPECIFIED);
  .....
}

test3() {
  createTestObject("abcd", DEFAULT, FOO_ENUM_PARAM, GET, 135135, STANDARD, SPECIFIED);
  .....
}

test4() {
  createTestObject("abcd", DEFAULT, BAR_ENUM_PARAM, POST, 135135, STANDARD, SPECIFIED);
  .....
}
```

vs:


```
test1() {
  createTestObject();
  .....
}

test2() {
  createTestObject.asBuilder().setPriceBehavior(PREMIUM).build();
  .....
}

test3() {
  createTestObject.asBuilder().setHttpMethod(GET).build();
  .....
}

test4() {
  createTestObject.asBuilder().setParamName(BAR_ENUM_PARAM).build();
  .....
}
```
___
*[core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java, line 303 at r4](https://reviewable.io/reviews/google/nomulus/1596#-N0Gw6kQBZ7VQZRe5dZn:-N0Gw6kQBZ7VQZRe5dZo:bw7tvp6) ([raw file](https://github.com/google/nomulus/blob/ec663f3286e52cad6e76ae53136ec32ece06badf/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L303)):*
> ```Java
>   private static AllocationToken.Builder builderWithPromo(
>       RenewalPriceBehavior renewalPriceBehavior) {
>     DateTime now = DateTime.now(UTC);
> ```

There's no reason to duplicate all of this logic.  Leave this method as-is, and then change the `builderWithPromo()` method below to simply be:

```
return builderWithPromo(DEFAULT);
```
___
*[core/src/main/java/google/registry/model/domain/token/AllocationToken.java, line 378 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXLBt4FDgV7Jy5TiS:-N0GvPW3ChEBpcW5k7NW:bma5mz9) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L378)):*

Because there's no use case yet? Anchor Tenant status doesn't come from allocation token. If there's use case for that, then this should be removed. It's just a check to make sure only the cases (so far just DEFAULT and SPECIFIED) are allowed . 

I gave my reason in one of the comments. I changed it though. 
</blockquote></details>

Generally the YAGNI principle means that you don't spend a lot of extra work to support a use case that there isn't yet a need for.  What you did is the opposite though -- you put in a lot of extra work to explicitly not support a use case when supporting it was free and much simpler.  There's no reason in principle we wouldn't want the capability for a token to trigger domain renewal at standard prices (and indeed we might actually end up using that for internal registrations over specified price, now that I think about it), so why not support it, especially when explicitly not supporting it is more work?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java, line 246 at r4](https://reviewable.io/reviews/google/nomulus/1596#-N0Gxujs7tzGOre-IBfJ:-N0Gxujs7tzGOre-IBfK:b-lhdjhh) ([raw file](https://github.com/google/nomulus/blob/ec663f3286e52cad6e76ae53136ec32ece06badf/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L246)):*
> ```Java
>         assertThrows(
>             ParameterException.class,
>             () -> {
> ```

Unnecessary curly braces here and elsewhere.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java, line 151 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXUEO49fepDFgrL6X:-N0Gxzo5-pUHmrqF0qy5:bjhp6n1) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L151)):*

Yeah, see, this implementation is much simpler and nicer.
</blockquote></details>

And it's not creating a lot of throw-away String parsing that's just used for this one method.  If you were going to go that route, it should've been abstracted out in a way that is reusable across all commands that need enum params ... like `EnumParameter` which you are now currently using: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/params/EnumParameter.java;l=1?q=enumparam&sq=&ss=nomulus%2Fnomulus
___
*[core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java, line 410 at r4](https://reviewable.io/reviews/google/nomulus/1596#-N0GyLDe6BnHekhS9QuR:-N0GyLDe6BnHekhS9QuS:b-jjpm5b) ([raw file](https://github.com/google/nomulus/blob/ec663f3286e52cad6e76ae53136ec32ece06badf/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L410)):*
> ```Java
>   }
> 
>   private AllocationToken createToken(
> ```

You've duplicated most of the `createToken()` logic here -- one of these overloads should be calling the other.  See lots of examples in `DatabaseHelper` of that, e.g.:  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/DatabaseHelper.java;l=180-213?q=databasehelp&ss=nomulus%2Fnomulus

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java` line 246 at r4](https://reviewable.io/reviews/google/nomulus/1596#-N0Gxujs7tzGOre-IBfJ:-N0b2w_k4HeNVfFQarab:b-l0nwmh) ([raw file](https://github.com/google/nomulus/blob/ec663f3286e52cad6e76ae53136ec32ece06badf/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L246)):*

Unnecessary curly braces here and elsewhere.
</blockquote></details>

Done.
My question now is, why do we remove curly braces for one liners but not for one liner if statements. 
___
*[`core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java` line 410 at r4](https://reviewable.io/reviews/google/nomulus/1596#-N0GyLDe6BnHekhS9QuR:-N0b4hA7AJBr_PsCI3mO:b-c50aqa) ([raw file](https://github.com/google/nomulus/blob/ec663f3286e52cad6e76ae53136ec32ece06badf/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L410)):*

You've duplicated most of the `createToken()` logic here -- one of these overloads should be calling the other.  See lots of examples in `DatabaseHelper` of that, e.g.:  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/DatabaseHelper.java;l=180-213?q=databasehelp&ss=nomulus%2Fnomulus
</blockquote></details>

Utilized the other `createToken` method now. 
___
*[`core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java` line 245 at r1](https://reviewable.io/reviews/google/nomulus/1596#-N0CXptvBEsmJ44JvSVR:-N0bLYC30k2GA87Ij-gr:b-2e5tuu) ([raw file](https://github.com/google/nomulus/blob/808936f2d6202b976edf5b6e7b756e1b6e2e5e62/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L245)):*

Overloads are fine.  You haven't really implemented it correctly below though -- the overloaded methods should be calling each other, not implementing everything from scratch each time.  You also potentially don't need overloads at all; just build it with default values and then explicitly set the 1 thing under test at callsites (this 1 overload for the renewal price behavior is fine though).

When one parameter has the same default value 90+% of the time, adding it in at all the callsites actually makes it less clear what's going on, because it's reducing the signal-to-noise ratio.  Imagine a hypothetical scenario where an object under test has many different fields.  Which of these is clearer to show what's really going on, always splatting out every single field at every callsite, or just showing the one field of interest that's different for each test case?

```
test1() {
  createTestObject("abcd", DEFAULT, FOO_ENUM_PARAM, POST, 135135, STANDARD, SPECIFIED);
  .....
}

test2() {
  createTestObject("abcd", DEFAULT, FOO_ENUM_PARAM, POST, 135135, PREMIUM, SPECIFIED);
  .....
}

test3() {
  createTestObject("abcd", DEFAULT, FOO_ENUM_PARAM, GET, 135135, STANDARD, SPECIFIED);
  .....
}

test4() {
  createTestObject("abcd", DEFAULT, BAR_ENUM_PARAM, POST, 135135, STANDARD, SPECIFIED);
  .....
}
```

vs:


```
test1() {
  createTestObject();
  .....
}

test2() {
  createTestObject.asBuilder().setPriceBehavior(PREMIUM).build();
  .....
}

test3() {
  createTestObject.asBuilder().setHttpMethod(GET).build();
  .....
}

test4() {
  createTestObject.asBuilder().setParamName(BAR_ENUM_PARAM).build();
  .....
}
```
</blockquote></details>

I originally was thinking to expand the existing method that allows more flexibility when it comes to creating test objects. I decided to include two methods, one with a default behavior without modifying too much of the existing code ..

Your reasoning makes sense too. I modified and used `asBuilder()` to set different behaviors instead. 
___
*[`core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java` line 303 at r4](https://reviewable.io/reviews/google/nomulus/1596#-N0Gw6kQBZ7VQZRe5dZn:-N0b79sI4T7p4tdRytcw:b-jo0szh) ([raw file](https://github.com/google/nomulus/blob/ec663f3286e52cad6e76ae53136ec32ece06badf/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L303)):*

There's no reason to duplicate all of this logic.  Leave this method as-is, and then change the `builderWithPromo()` method below to simply be:

```
return builderWithPromo(DEFAULT);
```
</blockquote></details>

had no idea why i kept doing that, thanks !

CydeWeys (changes requested):
I'm still finding errors that indicate to me you aren't looking carefully at the diffs after making changes.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596#-:-N0fdmT21dDAoGPo6_vs:b-uh636f)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java` line 177 at r5](https://reviewable.io/reviews/google/nomulus/1596#-N0fdSeWD422XyHWlhJA:-N0fdSeWD422XyHWlhJB:bydz78h) ([raw file](https://github.com/google/nomulus/blob/0dfabe1c64a7e9295f7705f9a7c06b433bb75d6a/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L177)):*
> ```Java
>       numTokens = domainNames.size();
>     }
> ```

I don't see a reason to remove this newline.
___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 97 at r5](https://reviewable.io/reviews/google/nomulus/1596#-N0fdZDIDWzcqtW6418P:-N0fdZDIDWzcqtW6418Q:b5zu6k3) ([raw file](https://github.com/google/nomulus/blob/ea4974c3792a8c37bb126cbb4e65484001f6fbc7/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L97)):*
> ```Java
>           "Type of renewal price behavior, either DEFAULT (default), NONPREMIUM, or SPECIFIED."
>               + " This indicates how a domain should be charged for renewal.")
>   private RenewalPriceBehavior renewalPriceBehavior;
> ```

Should be `@Nullable` (because it will usually be null).
___
*[`core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java` line 387 at r5](https://reviewable.io/reviews/google/nomulus/1596#-N0fdi9y57FIWgyzEkOr:-N0fdi9y57FIWgyzEkOs:b9qpwvd) ([raw file](https://github.com/google/nomulus/blob/ea4974c3792a8c37bb126cbb4e65484001f6fbc7/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L387)):*
> ```Java
>         .isEqualTo("For UNLIMITED_USE tokens, must specify --token_status_transitions");
>   }
> ```

Unnecessary newline added.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 97 at r5](https://reviewable.io/reviews/google/nomulus/1596#-N0fdZDIDWzcqtW6418P:-N0fiHZ--jYAgDjcr5fY:b8a0dga) ([raw file](https://github.com/google/nomulus/blob/ea4974c3792a8c37bb126cbb4e65484001f6fbc7/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L97)):*

Should be `@Nullable` (because it will usually be null).
</blockquote></details>

It's set to `DEFAULT` by default. There's no case when it would be null.  In the schema, it is `alter table "AllocationToken" add column renewal_price_behavior text default 'DEFAULT' not null;`

See files for more details: 
[AllocationToken.java - line 157](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/domain/token/AllocationToken.java;l=157?q=allocatio&ss=nomulus%2Fnomulus)
[V116__add_renewal_column_to_allocation_token.sql - line 15](https://cs.opensource.google/nomulus/nomulus/+/master:db/src/main/resources/sql/flyway/V116__add_renewal_column_to_allocation_token.sql;l=15?q=renewal_price_behavior&ss=nomulus%2Fnomulus)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 97 at r5](https://reviewable.io/reviews/google/nomulus/1596#-N0fdZDIDWzcqtW6418P:-N0lbOlj30DU9b2e8rqp:bg82s3o) ([raw file](https://github.com/google/nomulus/blob/ea4974c3792a8c37bb126cbb4e65484001f6fbc7/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L97)):*

It's set to `DEFAULT` by default. There's no case when it would be null.  In the schema, it is `alter table "AllocationToken" add column renewal_price_behavior text default 'DEFAULT' not null;`

See files for more details: 
[AllocationToken.java - line 157](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/domain/token/AllocationToken.java;l=157?q=allocatio&ss=nomulus%2Fnomulus)
[V116__add_renewal_column_to_allocation_token.sql - line 15](https://cs.opensource.google/nomulus/nomulus/+/master:db/src/main/resources/sql/flyway/V116__add_renewal_column_to_allocation_token.sql;l=15?q=renewal_price_behavior&ss=nomulus%2Fnomulus)
</blockquote></details>

It's not set to `DEFAULT` by default.  If it were, then this line of code would look like:

```
private RenewalPriceBehavior renewalPriceBehavior = DEFAULT;
```

(but that wouldn't be correct for what it's actually doing)

However, since it does not look like that, this parameter does indeed default to `null` and thus needs to be`@Nullable`.  Note that this parameter to this command is not the same thing as the column in the DB.  And note that this parameter can be null, where null means that the existing value on the token should be retained.
___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 153 at r6](https://reviewable.io/reviews/google/nomulus/1596#-N0lbpVjBkC3bEDUW_r8:-N0lbpVjBkC3bEDUW_r9:b-l2s22f) ([raw file](https://github.com/google/nomulus/blob/134515a85a0140402be89f612f186fb5f31caec8/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L153)):*
> ```Java
>     Optional.ofNullable(discountYears).ifPresent(builder::setDiscountYears);
>     Optional.ofNullable(tokenStatusTransitions).ifPresent(builder::setTokenStatusTransitions);
>     Optional.ofNullable(renewalPriceBehavior)
> ```

You're literally correctly handling the null case for `renewalPriceBehavior` right here, so the command parameter needs to be `@Nullable`.
___
*[`core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java` line 177 at r6](https://reviewable.io/reviews/google/nomulus/1596#-N0lc0ZV9pboLvRYeviq:-N0lc0ZV9pboLvRYevir:bcmxzeq) ([raw file](https://github.com/google/nomulus/blob/134515a85a0140402be89f612f186fb5f31caec8/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L177)):*
> ```Java
>         .hasMessageThat()
>         .isEqualTo(
>             "Invalid value for --renewal_price_behavior parameter. Allowed values:[DEFAULT,"
> ```

There's a missing space in this error message after the colon.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 97 at r5](https://reviewable.io/reviews/google/nomulus/1596#-N0fdZDIDWzcqtW6418P:-N0lveAa01BW3lysQPmF:b-r0dzgz) ([raw file](https://github.com/google/nomulus/blob/ea4974c3792a8c37bb126cbb4e65484001f6fbc7/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L97)):*

It's not set to `DEFAULT` by default.  If it were, then this line of code would look like:

```
private RenewalPriceBehavior renewalPriceBehavior = DEFAULT;
```

(but that wouldn't be correct for what it's actually doing)

However, since it does not look like that, this parameter does indeed default to `null` and thus needs to be`@Nullable`.  Note that this parameter to this command is not the same thing as the column in the DB.  And note that this parameter can be null, where null means that the existing value on the token should be retained.
</blockquote></details>

my bad and fixed.
___
*[`core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java` line 177 at r6](https://reviewable.io/reviews/google/nomulus/1596#-N0lc0ZV9pboLvRYeviq:-N0lxtgl5fJ5iRebZiBw:b-sjdy4t) ([raw file](https://github.com/google/nomulus/blob/134515a85a0140402be89f612f186fb5f31caec8/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L177)):*

There's a missing space in this error message after the colon.
</blockquote></details>

Do you mean there's no space here `Allowed values:[DEFAULT,` ? That's the default enum message

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 99 at r7](https://reviewable.io/reviews/google/nomulus/1596#-N1A8a3N0IJ40J76ZLGu:-N1A8a3N0IJ40J76ZLGv:b-fwrh1w) ([raw file](https://github.com/google/nomulus/blob/87e35c3f92d3d3321dfd7ba76b409cd81e713223/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L99)):*
> ```Java
>               + " This indicates how a domain should be charged for renewal.")
>   @Nullable
>   private RenewalPriceBehavior renewalPriceBehavior;
> ```

It seems like something is missing here.  If the renewal price behavior is set to `SPECIFIED`, then how is the price actually being specified?  It seems like that's missing here and missing on the `AllocationToken` entity itself.  You already added it to `BillingRecurrence` at least: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/billing/BillingEvent.java;l=591-601?q=billingevent&ss=nomulus%2Fnomulus

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 99 at r7](https://reviewable.io/reviews/google/nomulus/1596#-N1A8a3N0IJ40J76ZLGu:-N1AOInz4AZVeSOZ1YBh:bcc55dq) ([raw file](https://github.com/google/nomulus/blob/87e35c3f92d3d3321dfd7ba76b409cd81e713223/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L99)):*

It seems like something is missing here.  If the renewal price behavior is set to `SPECIFIED`, then how is the price actually being specified?  It seems like that's missing here and missing on the `AllocationToken` entity itself.  You already added it to `BillingRecurrence` at least: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/billing/BillingEvent.java;l=591-601?q=billingevent&ss=nomulus%2Fnomulus
</blockquote></details>

Never mind. Per in-person conversation, we don't need to add this here (at least not now), as it's pretty much duplicative of how the existing `discountFraction` on allocation tokens works.

We should, however, add some wording here to the parameter help message explaining that `SPECIFIED` in this context means that the domain will renew at the same cost as the initial create, taking into account the discount fraction.

Note that this is different from how it works at the level of the `BillingRecurrence` -- there it really does mean "renew at this specified price" where that specified price is stored alongside and could be anything (though in our existing use cases, it will, so far, be the same as the domain's calculated create price).

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java` line 177 at r6](https://reviewable.io/reviews/google/nomulus/1596#-N0lc0ZV9pboLvRYeviq:-N1AWwUf24sWLHRxszNh:bwcobry) ([raw file](https://github.com/google/nomulus/blob/134515a85a0140402be89f612f186fb5f31caec8/core/src/test/java/google/registry/tools/UpdateAllocationTokensCommandTest.java#L177)):*

Do you mean there's no space here `Allowed values:[DEFAULT,` ? That's the default enum message 
</blockquote></details>

Oh darn. I'd prefer a space in there, but if it's coming from the framework/library, guess we're stuck with it.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java` line 99 at r7](https://reviewable.io/reviews/google/nomulus/1596#-N1A8a3N0IJ40J76ZLGu:-N1AiqfsATKAOBGrXpH7:b8ek0lh) ([raw file](https://github.com/google/nomulus/blob/87e35c3f92d3d3321dfd7ba76b409cd81e713223/core/src/main/java/google/registry/tools/UpdateAllocationTokensCommand.java#L99)):*

Never mind. Per in-person conversation, we don't need to add this here (at least not now), as it's pretty much duplicative of how the existing `discountFraction` on allocation tokens works.

We should, however, add some wording here to the parameter help message explaining that `SPECIFIED` in this context means that the domain will renew at the same cost as the initial create, taking into account the discount fraction.

Note that this is different from how it works at the level of the `BillingRecurrence` -- there it really does mean "renew at this specified price" where that specified price is stored alongside and could be anything (though in our existing use cases, it will, so far, be the same as the domain's calculated create price).
</blockquote></details>

updated the description for both commands.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1596#-:-N1KkG9e1XqoxhskVyl8:bnfp4nl)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1602, opened 2022-04-21
Change check for root directory during rollback

mindhog:
`rollback_tool` tries to infer the root of the nomulus tree by checking for a
directory named "nomulus".  This is potentially problematic (and, indeed, was
for me) since there is no guarantee what that directory will be named.

There are a number of features that characterize the root directory.  Check
for the presence of the `rollback_tool` wrapper script, as this is both at
root level and tightly coupled to the python code, so hopefully we won't
move it without testing that the script still works.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1602)

### DIFF
--- a/release/rollback/common.py
+++ b/release/rollback/common.py
@@ -108,7 +108,7 @@ def get_nomulus_root() -> str:
         The absolute path to the Nomulus root directory.
     """
     for folder in pathlib.Path(__file__).parents:
-        if folder.name != 'nomulus':
+        if not folder.joinpath('rollback_tool').exists():
             continue
         if not folder.joinpath('settings.gradle').exists():
             continue


mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1602)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`release/rollback/common.py` line 114 at r1](https://reviewable.io/reviews/google/nomulus/1602#-N0WR4Ko7O9oiGkwie6-:-N0WS5Z07FVw0nXC2GDq:blvtfn4) ([raw file](https://github.com/google/nomulus/blob/7f923cd09d490be851b2aea3e2b2eebccecfa4a7/release/rollback/common.py#L114)):*

I wonder if this check is sufficient by itself.
</blockquote></details>

It currently is, as is the one I added.  My hunch is that we're better off checking for `rollback_tool`, since if anyone moves it they'll hopefully do a test run which will reveal the need to update this script.

Shall I remove the check for settings.gradle?  Or is there a reason to prefer that over `rollback_tool`?

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1602)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1607, opened 2022-04-27
Ignore version UIDs during txn deserialization

mindhog:
When deserializing transactions for replay to datastore, ignore class version
UIDs that don't match those of the local classes and just use the local class
descriptors instead.  This is a simple solution for the problem of persisted
VKeys containing references to classes where the class has been updated and
the serial version UID has changed.

Also add a "replay_txns" command that replays the transactions from a given
start point so we can verify all transactions are deserializable.

TESTED:
    Ran replay_txns against all transactions on sandbox beginning with
    transaction id 1828385, which includes Recurring billing events containing
    both the old and the new serial version UIDs.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1607)

### DIFF
--- a/core/src/main/java/google/registry/persistence/transaction/Transaction.java
+++ b/core/src/main/java/google/registry/persistence/transaction/Transaction.java
@@ -33,8 +33,10 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.EOFException;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.io.ObjectStreamClass;
 
 /**
  * A SQL transaction that can be serialized and stored in its own table.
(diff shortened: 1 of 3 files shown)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1607)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mindhog and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/ReplayTxnsCommand.java` line 54 at r1](https://reviewable.io/reviews/google/nomulus/1607#-N0gEPUW4roTSSqfkCSL:-N0gEPUW4roTSSqfkCSM:b6wutoq) ([raw file](https://github.com/google/nomulus/blob/538d9f8f673bc14790076b9b4aa8bc3baab811a1/core/src/main/java/google/registry/tools/ReplayTxnsCommand.java#L54)):*
> ```Java
>     } while (txns.size() > 0);
>   }
> }
> ```

Missing newline?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1607)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/tools/ReplayTxnsCommand.java` line 54 at r1](https://reviewable.io/reviews/google/nomulus/1607#-N0gEPUW4roTSSqfkCSL:-N0gJ9bv5f91XpbVWoho:b-896fix) ([raw file](https://github.com/google/nomulus/blob/538d9f8f673bc14790076b9b4aa8bc3baab811a1/core/src/main/java/google/registry/tools/ReplayTxnsCommand.java#L54)):*

Missing newline?
</blockquote></details>

Done.

PR #1609, opened 2022-04-28
Add AT and IR support in ExpandRecurringBillingEventsAction 

rachelguan:
Because of the additional support for anchor tenants and internal registrations in domain renewal, `ExpandRecurringBillingEventsAction` needs to be modified for generating the correct billing events. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1609)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @rachelguan)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N1A7juD4G8sRLl7_7qR:b5y16d) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*
> ```Java
> import com.google.common.collect.Streams;
> import com.google.common.flogger.FluentLogger;
> import google.registry.flows.domain.DomainFlowUtils;
> ```

Per my comment on the other PR suggesting that this functionality should be in `DomainPricingLogic` and now in `DomainFlowUtils`, that is bolstered by seeing this import here.  It doesn't make sense to import flow utils in class that is not a flow (as this one isn't) outside the flows package.  `DomainFlowUtils` should only be used by domain flows, whereas here you've written something (calculating a domain's renew price) that is clearly much more generally applicable, and thus should be in `DomainPricingLogic` which is suitable for use more broadly.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N2rvxAR7QMuEPaKOu8A:bxzq4jf) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*

I had some discussion with Rachel last week regarding this. Unfortunately `DomainPricingLogic` is scoped at `@FlowScope` so it also can only be injected to flows. From what Rachel said, the renewal price that she wants does not require a `CustomLogic`, and thus does not actually need to be injected. It seems to me the easiest (and least messy) way is to put the generic renewal logic in a public static function that can be used in non-flow actions (like this one), and have the `getRenewPrice()` function call `CustomLogic` to customize the standard renewal priced returned by the static function.
</blockquote></details>

I initially put the method in `DomainFlowUtils`because it seems like a reasonable place to put a method that's usable. I also wanted to avoid injecting `DomainPricingLogic` to code that's not flow. I tried to create two versions of `getRenewPrice()` (one returns `Money` and another returns `FeesAndCredit`) but that seems confusing. 

I had issues injecting `DomainPricingLogic` and as jianglai@ mentioned above, it can only be injected to flows. Therefore, I extracted the logic and created a static method. They don't have the same name and I hope that's less confusing for readers.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N2x0xSIEnktZFPfGbnf:b-t4pvv4) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*

I initially put the method in `DomainFlowUtils`because it seems like a reasonable place to put a method that's usable. I also wanted to avoid injecting `DomainPricingLogic` to code that's not flow. I tried to create two versions of `getRenewPrice()` (one returns `Money` and another returns `FeesAndCredit`) but that seems confusing. 

I had issues injecting `DomainPricingLogic` and as jianglai@ mentioned above, it can only be injected to flows. Therefore, I extracted the logic and created a static method. They don't have the same name and I hope that's less confusing for readers. 
</blockquote></details>

`DomainPricingLogic` does not need to be scoped at `@FlowScope`.  That annotation barely does anything anyway (and it really doesn't matter here as `DomainPricingLogic` does not maintain any internal state that is expensive to construct anyway).  So it seems like the simplest thing to do here is just remove `@FlowScope` from `DomainPricingLogic` and then just use everything through Dagger normally.
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 136 at r2](https://reviewable.io/reviews/google/nomulus/1609#-N2x1ClJ7LZxzVBkcnEW:-N2x1ClJ7LZxzVBkcnEX:b-ov49fy) ([raw file](https://github.com/google/nomulus/blob/3f01105546dde61f22b6958bb3340972e5778f64/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L136)):*
> ```Java
>    * google.registry.batch.ExpandRecurringBillingEventsAction}.
>    */
>   public static FeesAndCredits getNonCustomRenewPrice(
> ```

Per comment above, we don't need a 2nd method for getting the renew price.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N2x2je-4ixT8uGQ0B69:b-pe37j3) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*

`DomainPricingLogic` does not need to be scoped at `@FlowScope`.  That annotation barely does anything anyway (and it really doesn't matter here as `DomainPricingLogic` does not maintain any internal state that is expensive to construct anyway).  So it seems like the simplest thing to do here is just remove `@FlowScope` from `DomainPricingLogic` and then just use everything through Dagger normally.
</blockquote></details>

I've tried the following but I didn't get far: 
1. Remove `FlowScope` in `DomainPricingLogic` and it threw an error 

```error: [Dagger/MissingBinding] google.registry.flows.custom.DomainPricingCustomLogic cannot be provided without an @Inject constructor or an @Provides-annotated method.```

2. Since `CustomLogicModule` contains `provideDomainPricingCustomLogic`, I added that to `BackendComponent` and the errors were 

```
/usr/local/google/home/rachelguan/git/nomulus/core/src/main/java/google/registry/module/backend/BackendComponent.java:93: error: [Dagger/MissingBinding] google.registry.model.eppinput.EppInput cannot be provided without an @Inject constructor or an @Provides-annotated method.

/usr/local/google/home/rachelguan/git/nomulus/core/src/main/java/google/registry/module/backend/BackendComponent.java:93: error: [Dagger/MissingBinding] google.registry.flows.SessionMetadata cannot be provided without an @Provides-annotated method.

/usr/local/google/home/rachelguan/git/nomulus/core/src/main/java/google/registry/module/backend/BackendComponent.java:93: error: [Dagger/MissingBinding] google.registry.flows.FlowMetadata cannot be provided without an @Provides-annotated method.


```
In the full error message, it indicates tho, `  A binding for google.registry.flows.FlowMetadata exists in google.registry.flows.FlowComponent:`, ` A binding for google.registry.flows.SessionMetadata exists in google.registry.flows.FlowComponent:` and `A binding for google.registry.model.eppinput.EppInput exists in google.registry.flows.FlowComponent:`

I tried to add `FlowModule.class` to `BackendComponent` and I ran into more binding issues.

**Review by CydeWeys — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N36Q5q0AxzGHNzDkf-l:b-tllr1j) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*

I've tried the following but I didn't get far: 
1. Remove `FlowScope` in `DomainPricingLogic` and it threw an error 

```error: [Dagger/MissingBinding] google.registry.flows.custom.DomainPricingCustomLogic cannot be provided without an @Inject constructor or an @Provides-annotated method.```

2. Since `CustomLogicModule` contains `provideDomainPricingCustomLogic`, I added that to `BackendComponent` and the errors were 

```
/usr/local/google/home/rachelguan/git/nomulus/core/src/main/java/google/registry/module/backend/BackendComponent.java:93: error: [Dagger/MissingBinding] google.registry.model.eppinput.EppInput cannot be provided without an @Inject constructor or an @Provides-annotated method.

/usr/local/google/home/rachelguan/git/nomulus/core/src/main/java/google/registry/module/backend/BackendComponent.java:93: error: [Dagger/MissingBinding] google.registry.flows.SessionMetadata cannot be provided without an @Provides-annotated method.

/usr/local/google/home/rachelguan/git/nomulus/core/src/main/java/google/registry/module/backend/BackendComponent.java:93: error: [Dagger/MissingBinding] google.registry.flows.FlowMetadata cannot be provided without an @Provides-annotated method.


```
In the full error message, it indicates tho, `  A binding for google.registry.flows.FlowMetadata exists in google.registry.flows.FlowComponent:`, ` A binding for google.registry.flows.SessionMetadata exists in google.registry.flows.FlowComponent:` and `A binding for google.registry.model.eppinput.EppInput exists in google.registry.flows.FlowComponent:`

I tried to add `FlowModule.class` to `BackendComponent` and I ran into more binding issues. 

</blockquote></details>

Now that I've submitted PR #1648, please try again.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N36Sj7SChI0x_HJYVM4:b2ve83t) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*

Now that I've submitted PR #1648, please try again.
</blockquote></details>

merged that and now handling all the non static/static issues introduced to the action file. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 136 at r2](https://reviewable.io/reviews/google/nomulus/1609#-N2x1ClJ7LZxzVBkcnEW:-N36SrYX8CwpH8cdlfZY:bp5yzku) ([raw file](https://github.com/google/nomulus/blob/3f01105546dde61f22b6958bb3340972e5778f64/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L136)):*

Per comment above, we don't need a 2nd method for getting the renew price.
</blockquote></details>

removed that.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/1609#-N1A7juC7UzanJzRPTPr:-N37-8eF0VvDpfC58Ve7:bb33o13) ([raw file](https://github.com/google/nomulus/blob/289a6a639cbb90dc1813b9774a3ae71d443351e7/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L43)):*

merged that and now handling all the non static/static issues introduced to the action file. 
</blockquote></details>

Because `getRenewPrice()` method is non static and the methods that call `getRenewPrice()` in the action file are static, I have to wrap the method calls in a try catch block. I'm not sure what is the best exception error to throw so I have `RunTimeException()` right now. 

I thought creating a static method that returns a `Money` actually would make this action file cleaner and as @jianglai mentioned, there's no custom logic in it..Yes, the three parameters for setting up `DomainPricingCustomLogic` could be null but wrapping the methods call with try catch seem unnecessary to me. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 136 at r2](https://reviewable.io/reviews/google/nomulus/1609#-N2x1ClJ7LZxzVBkcnEW:-N37-6vNEbU0nP4C8Abd:b3ksc) ([raw file](https://github.com/google/nomulus/blob/3f01105546dde61f22b6958bb3340972e5778f64/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L136)):*

removed that. 
</blockquote></details>

a

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 188 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3Pbd-S894flQqI7-yj:-N3Pbd-S894flQqI7-yk:b-p690ry) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L188)):*
> ```Java
>                         billingEventsSaved =
>                             expandBillingEvent(recurring, executeTime, cursorTime, isDryRun);
>                       } catch (EppException e) {
> ```

What's going on here?  Why is `expandBillingEvent()` throwing errors?  We need to handle this better, not use a sentinel value indicating an error condition (while swallowing the actual exception making it impossible to debug).
___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 194 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3PbmuLA4D3shAi55qt:-N3PbmuLA4D3shAi55qu:b-h1ppvp) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L194)):*
> ```Java
>                         throw new RuntimeException(
>                             String.format(
>                                 "Error while expanding billing events in batches () for %d",
> ```

If we are going to re-throw an exception here (which I'm not convinced is what should be happening), then you would need to wrap the actual root exception as an inner exception here.  As currently written, this code is swallowing the actual cause of the error, making debugging impossible.
___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 311 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3Pc2sM2Sh6G8ExVCxK:-N3Pc2sM2Sh6G8ExVCxL:b-5d3bxx) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L311)):*
> ```Java
>                         return expandBillingEvent(recurring, executeTime, cursorTime, isDryRun);
>                       } catch (EppException e) {
>                         return -1;
> ```

Same issue here as above.
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 54 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3PbO2tC2n-R14RHScD:-N3PbO2u2OuYK5uNcJs_:by55vul) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L54)):*
> ```Java
> public final class DomainPricingLogic {
> 
>   private DomainPricingCustomLogic customLogic;
> ```

This should be `final` too.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 188 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3Pbd-S894flQqI7-yj:-N3Qd1Pc3LzCtXvyT3Mk:b-qssebl) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L188)):*

Anyway, so I looked into it more and the `EppException` is part of the return signature of `DomainPricingCustomLogic.customizeRenewPrice()`, which never actually throws an Exception in practice.  The simplest solution here is to simply remove `throws EppException` from the signatures of those `DomainPricingCustomLogic` methods, as they're not used, were speculative in nature, and now that we're using `DomainPricingLogic` outside the context of EPP flows anyway, it doesn't even make sense to use the EPP-specific `EppException` there.  So just remove the unused `throws` clauses, and then in the future, on the off chance that we do want to re-add a checked exception to these methods, a more suitable type could be selected, and then they'd have to add handling at callsites like we see here.  But for now it's not necessary.
</blockquote></details>

fixed.
___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 194 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3PbmuLA4D3shAi55qt:-N3QfkDS4jP4PRUGjEUe:b-qssebl) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L194)):*

If we are going to re-throw an exception here (which I'm not convinced is what should be happening), then you would need to wrap the actual root exception as an inner exception here.  As currently written, this code is swallowing the actual cause of the error, making debugging impossible.
</blockquote></details>

fixed.
___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 311 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3Pc2sM2Sh6G8ExVCxK:-N3QflGK0qsEkGJfO_nZ:blkwbr5) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L311)):*

Same issue here as above.
</blockquote></details>

fixed. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 54 at r4](https://reviewable.io/reviews/google/nomulus/1609#-N3PbO2tC2n-R14RHScD:-N3QfmL27hh7ERe3CH9v:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/0d342a3e01d31aed22b4af84ceb98631d21d21bd/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L54)):*

This should be `final` too.
</blockquote></details>

done

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 259 at r5](https://reviewable.io/reviews/google/nomulus/1609#-N3ydFnrAl9BZLrrp_nq:-N3ydFns5mXJlMnNAwmi:bb270p3) ([raw file](https://github.com/google/nomulus/blob/a012526728b9fad8de0361520a48b0c040d4276b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L259)):*
> ```Java
> 
>   /** Mapper to expand {@link Recurring} billing events into synthetic {@link OneTime} events. */
>   public class ExpandRecurringBillingEventsMapper extends Mapper<Recurring, DateTime, DateTime> {
> ```

Why did you remove the `static` keyword on the inner class here?  This may introduce an error?  (Though it won't matter much since we don't run in Datastore anymore, it's still an unnecessary change.)

If the issue is that you need to be able to call `domainPricingLogic`, then that would need to be passed in to the constructor to `ExpandRecurringBillingEventsMapper`.  Alternatively you can just delete the mapreduce entirely as it supports the Ofy mode and we simply don't need that.

In general, static inner classes are preferred to non-static inner classes: https://dev.to/kylec32/effective-java-favor-static-members-classes-over-non-static-odj

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1609)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 259 at r5](https://reviewable.io/reviews/google/nomulus/1609#-N3ydFnrAl9BZLrrp_nq:-N3zAUtcCevL0TMpdvXm:b-qssebl) ([raw file](https://github.com/google/nomulus/blob/a012526728b9fad8de0361520a48b0c040d4276b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L259)):*

Why did you remove the `static` keyword on the inner class here?  This may introduce an error?  (Though it won't matter much since we don't run in Datastore anymore, it's still an unnecessary change.)

If the issue is that you need to be able to call `domainPricingLogic`, then that would need to be passed in to the constructor to `ExpandRecurringBillingEventsMapper`.  Alternatively you can just delete the mapreduce entirely as it supports the Ofy mode and we simply don't need that.

In general, static inner classes are preferred to non-static inner classes: https://dev.to/kylec32/effective-java-favor-static-members-classes-over-non-static-odj
</blockquote></details>

fixed.

CydeWeys:
Submitted as #1687.

PR #1610, opened 2022-04-28
Add renewal cost logic to DomainPricingLogic

rachelguan:
The purpose of this PR is to add renewal support for anchor tenants and internal registrations in the renew flow. This will be used in `ExpandRecurringBillingEventsAction` ([PR#1609](https://github.com/google/nomulus/pull/1609)), `DomainRenewFlow` ([PR#1592](https://github.com/google/nomulus/pull/1592)), and `DomainFlowUtil` ([PR#1637](https://github.com/google/nomulus/pull/1637)). 

While I was working on this PR, I found out that there's no test file for `DomainPricingLogic.java` -  Most of the logic files are wrappers but `DomainPricingLogic` contains calculations of different kinds of prices such as `renew` and `transfer`. I added test cases for `getRenewPrice()` and there should be cases for other calculation methods as well. I opened a bug to track that. In order to create an instance of `DomainPricingLogic`, the constructor of `DomainPricingCustomLogic` needs to be made public. 

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1610)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
@@ -128,6 +128,50 @@ public final class DomainPricingLogic {
             .build());
   }
 
+  /**
+   * Calculates the renewal cost of a domain.
+   *
+   * <p>This method is set as static which allows it to be used without instantiating a {@link
+   * DomainPricingLogic} object to get the renewal price in other files. If this method is non-
+   * static, it would be difficult for static methods in files such as {@link
+   * google.registry.batch.ExpandRecurringBillingEventsAction} to get the renewal cost. Moreover,
+   * it's easier to get the renewal price as an {@link Money} object by simply calling
+   * DomainPricingLogic.getRenewPrice() instead of injecting domainPricingLogic then calling
+   * domainPricingLogic.getRenewPrice().getRenewPrice().
+   */
+  @SuppressWarnings("unused")
+  public static Money getRenewPrice(
+      String domainName, DateTime dateTime, int years, @Nullable Recurring recurringBillingEvent) {
+    checkArgument(years > 0, "Number of years must be positive");
+    if (recurringBillingEvent != null) {
+      switch (recurringBillingEvent.getRenewalPriceBehavior()) {
+        case DEFAULT:
+          return PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+          // if the renewal price behavior is specified, then the renewal price should be the same
+          // as the creation price, which is stored in the billing event as the renewal price
+        case SPECIFIED:
+          checkArgument(
+              recurringBillingEvent.getRenewalPrice().isPresent(),
+              "Unexpected behavior: renewal price cannot be null when renewal behavior is"
+                  + " SPECIFIED");
+          return recurringBillingEvent.getRenewalPrice().get().multipliedBy(years);
+          // if the renewal price behavior is nonpremium, it means that the domain should be renewed
+          // at standard price of domains at the time, even if the domain is premium
+        case NONPREMIUM:
+          return Registry.get(getTldFromDomainName(domainName))
+              .getStandardRenewCost(dateTime)
+              .multipliedBy(years);
+        default:
+          throw new IllegalArgumentException(
+              String.format(
+                  "Unknown RenewalPriceBehavior enum value: %s",
+                  recurringBillingEvent.getRenewalPriceBehavior()));
+      }
+    } else {
+      return PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+    }
+  }
+
   /** Returns a new renewal cost for the pricer. */
   @SuppressWarnings("unused")
   FeesAndCredits getRenewPrice(
@@ -138,32 +182,7 @@ public final class DomainPricingLogic {
       @Nullable Recurring recurringBillingEvent)
       throws EppException {
     checkArgument(years > 0, "Number of years must be positive");
-    Money renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
-    if (recurringBillingEvent != null) {
-      switch (recurringBillingEvent.getRenewalPriceBehavior()) {
-        case DEFAULT:
-          renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
-          break;
-        case SPECIFIED:
-          checkArgument(
-              recurringBillingEvent.getRenewalPrice().isPresent(),
-              "Unexpected behavior: renewal price cannot be null when renewal behavior is"
-                  + " SPECIFIED");
-          renewCost = recurringBillingEvent.getRenewalPrice().get().multipliedBy(years);
-          break;
-        case NONPREMIUM:
-          renewCost =
-              Registry.get(getTldFromDomainName(domainName))
-                  .getStandardRenewCost(dateTime)
-                  .multipliedBy(years);
-          break;
-        default:
-          throw new IllegalArgumentException(
-              String.format(
-                  "Unknown RenewalPriceBehavior enum value: %s",
-                  recurringBillingEvent.getRenewalPriceBehavior()));
-      }
-    }
+    Money renewCost = getRenewPrice(domainName, dateTime, years, recurringBillingEvent);
     return customLogic.customizeRenewPrice(
         RenewPriceParameters.newBuilder()
             .setFeesAndCredits(
--- a/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
@@ -14,8 +14,10 @@
 
 package google.registry.flows.domain;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.flows.domain.DomainFlowUtils.zeroInCurrency;
 import static google.registry.pricing.PricingEngineProxy.getPricesForDomainName;
+import static google.registry.util.DomainNameUtils.getTldFromDomainName;
 
 import com.google.common.net.InternetDomainName;
 import google.registry.flows.EppException;
@@ -27,15 +29,18 @@ import google.registry.flows.custom.DomainPricingCustomLogic.RenewPriceParameter
 import google.registry.flows.custom.DomainPricingCustomLogic.RestorePriceParameters;
 import google.registry.flows.custom.DomainPricingCustomLogic.TransferPriceParameters;
 import google.registry.flows.custom.DomainPricingCustomLogic.UpdatePriceParameters;
+import google.registry.model.billing.BillingEvent.Recurring;
 import google.registry.model.domain.fee.BaseFee;
 import google.registry.model.domain.fee.BaseFee.FeeType;
 import google.registry.model.domain.fee.Fee;
 import google.registry.model.domain.token.AllocationToken;
 import google.registry.model.pricing.PremiumPricingEngine.DomainPrices;
 import google.registry.model.tld.Registry;
+import google.registry.pricing.PricingEngineProxy;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import org.joda.money.CurrencyUnit;
 import org.joda.money.Money;
@@ -102,6 +107,7 @@ public final class DomainPricingLogic {
             .build());
   }
 
+  // TODO (rachelguan): this will be replaced by the method getRenewPrice below in PR#1592
   /** Returns a new renew price for the pricer. */
   @SuppressWarnings("unused")
   FeesAndCredits getRenewPrice(Registry registry, String domainName, DateTime dateTime, int years)
@@ -114,6 +120,60 @@ public final class DomainPricingLogic {
                 new FeesAndCredits.Builder()
                     .setCurrency(registry.getCurrency())
                     .addFeeOrCredit(Fee.create(renewCost, FeeType.RENEW, domainPrices.isPremium()))
+                    .build())
+            .setRegistry(registry)
+            .setDomainName(InternetDomainName.from(domainName))
+            .setAsOfDate(dateTime)
+            .setYears(years)
+            .build());
+  }
+
+  /** Returns a new renewal cost for the pricer. */
+  @SuppressWarnings("unused")
+  FeesAndCredits getRenewPrice(
+      Registry registry,
+      String domainName,
+      DateTime dateTime,
+      int years,
+      @Nullable Recurring recurringBillingEvent)
+      throws EppException {
+    checkArgument(years > 0, "Number of years must be positive");
+    Money renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+    if (recurringBillingEvent != null) {
+      switch (recurringBillingEvent.getRenewalPriceBehavior()) {
+        case DEFAULT:
+          renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+          break;
+        case SPECIFIED:
+          checkArgument(
+              recurringBillingEvent.getRenewalPrice().isPresent(),
+              "Unexpected behavior: renewal price cannot be null when renewal behavior is"
+                  + " SPECIFIED");
+          renewCost = recurringBillingEvent.getRenewalPrice().get().multipliedBy(years);
+          break;
+        case NONPREMIUM:
+          renewCost =
+              Registry.get(getTldFromDomainName(domainName))
+                  .getStandardRenewCost(dateTime)
+                  .multipliedBy(years);
+          break;
+        default:
+          throw new IllegalArgumentException(
+              String.format(
+                  "Unknown RenewalPriceBehavior enum value: %s",
+                  recurringBillingEvent.getRenewalPriceBehavior()));
+      }
+    }
+    return customLogic.customizeRenewPrice(
+        RenewPriceParameters.newBuilder()
+            .setFeesAndCredits(
+                new FeesAndCredits.Builder()
+                    .setCurrency(renewCost.getCurrencyUnit())
+                    .addFeeOrCredit(
+                        Fee.create(
+                            renewCost.getAmount(),
+                            FeeType.RENEW,
+                            PricingEngineProxy.isDomainPremium(domainName, dateTime)))
                     .build())
             .setRegistry(registry)
             .setDomainName(InternetDomainName.from(domainName))
--- a/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
@@ -128,50 +128,6 @@ public final class DomainPricingLogic {
             .build());
   }
 
-  /**
-   * Calculates the renewal cost of a domain.
-   *
-   * <p>This method is set as static which allows it to be used without instantiating a {@link
-   * DomainPricingLogic} object to get the renewal price in other files. If this method is non-
-   * static, it would be difficult for static methods in files such as {@link
-   * google.registry.batch.ExpandRecurringBillingEventsAction} to get the renewal cost. Moreover,
-   * it's easier to get the renewal price as an {@link Money} object by simply calling
-   * DomainPricingLogic.getRenewPrice() instead of injecting domainPricingLogic then calling
-   * domainPricingLogic.getRenewPrice().getRenewPrice().
-   */
-  @SuppressWarnings("unused")
-  public static Money getRenewPrice(
-      String domainName, DateTime dateTime, int years, @Nullable Recurring recurringBillingEvent) {
-    checkArgument(years > 0, "Number of years must be positive");
-    if (recurringBillingEvent != null) {
-      switch (recurringBillingEvent.getRenewalPriceBehavior()) {
-        case DEFAULT:
-          return PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
-          // if the renewal price behavior is specified, then the renewal price should be the same
-          // as the creation price, which is stored in the billing event as the renewal price
-        case SPECIFIED:
-          checkArgument(
-              recurringBillingEvent.getRenewalPrice().isPresent(),
-              "Unexpected behavior: renewal price cannot be null when renewal behavior is"
-                  + " SPECIFIED");
-          return recurringBillingEvent.getRenewalPrice().get().multipliedBy(years);
-          // if the renewal price behavior is nonpremium, it means that the domain should be renewed
-          // at standard price of domains at the time, even if the domain is premium
-        case NONPREMIUM:
-          return Registry.get(getTldFromDomainName(domainName))
-              .getStandardRenewCost(dateTime)
-              .multipliedBy(years);
-        default:
-          throw new IllegalArgumentException(
-              String.format(
-                  "Unknown RenewalPriceBehavior enum value: %s",
-                  recurringBillingEvent.getRenewalPriceBehavior()));
-      }
-    } else {
-      return PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
-    }
-  }
-
   /** Returns a new renewal cost for the pricer. */
   @SuppressWarnings("unused")
   FeesAndCredits getRenewPrice(
@@ -182,7 +138,36 @@ public final class DomainPricingLogic {
       @Nullable Recurring recurringBillingEvent)
       throws EppException {
     checkArgument(years > 0, "Number of years must be positive");
-    Money renewCost = getRenewPrice(domainName, dateTime, years, recurringBillingEvent);
+    Money renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+    if (recurringBillingEvent != null) {
+      switch (recurringBillingEvent.getRenewalPriceBehavior()) {
+        case DEFAULT:
+          renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+          break;
+          // if the renewal price behavior is specified, then the renewal price should be the same
+          // as the creation price, which is stored in the billing event as the renewal price
+        case SPECIFIED:
+          checkArgument(
+              recurringBillingEvent.getRenewalPrice().isPresent(),
+              "Unexpected behavior: renewal price cannot be null when renewal behavior is"
+                  + " SPECIFIED");
+          renewCost = recurringBillingEvent.getRenewalPrice().get().multipliedBy(years);
+          break;
+          // if the renewal price behavior is nonpremium, it means that the domain should be renewed
+          // at standard price of domains at the time, even if the domain is premium
+        case NONPREMIUM:
+          renewCost =
+              Registry.get(getTldFromDomainName(domainName))
+                  .getStandardRenewCost(dateTime)
+                  .multipliedBy(years);
+          break;
+        default:
+          throw new IllegalArgumentException(
+              String.format(
+                  "Unknown RenewalPriceBehavior enum value: %s",
+                  recurringBillingEvent.getRenewalPriceBehavior()));
+      }
+    }
     return customLogic.customizeRenewPrice(
         RenewPriceParameters.newBuilder()
             .setFeesAndCredits(
--- a/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
@@ -18,6 +18,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.flows.domain.DomainFlowUtils.zeroInCurrency;
 import static google.registry.pricing.PricingEngineProxy.getPricesForDomainName;
 import static google.registry.util.DomainNameUtils.getTldFromDomainName;
+import static google.registry.util.PreconditionsUtils.checkArgumentPresent;
 
 import com.google.common.net.InternetDomainName;
 import google.registry.flows.EppException;
@@ -36,8 +37,6 @@ import google.registry.model.domain.fee.Fee;
 import google.registry.model.domain.token.AllocationToken;
 import google.registry.model.pricing.PremiumPricingEngine.DomainPrices;
 import google.registry.model.tld.Registry;
-import google.registry.pricing.PricingEngineProxy;
-import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.Optional;
 import javax.annotation.Nullable;
@@ -107,29 +106,7 @@ public final class DomainPricingLogic {
             .build());
   }
 
-  // TODO (rachelguan): this will be replaced by the method getRenewPrice below in PR#1592
-  /** Returns a new renew price for the pricer. */
-  @SuppressWarnings("unused")
-  FeesAndCredits getRenewPrice(Registry registry, String domainName, DateTime dateTime, int years)
-      throws EppException {
-    DomainPrices domainPrices = getPricesForDomainName(domainName, dateTime);
-    BigDecimal renewCost = domainPrices.getRenewCost().multipliedBy(years).getAmount();
-    return customLogic.customizeRenewPrice(
-        RenewPriceParameters.newBuilder()
-            .setFeesAndCredits(
-                new FeesAndCredits.Builder()
-                    .setCurrency(registry.getCurrency())
-                    .addFeeOrCredit(Fee.create(renewCost, FeeType.RENEW, domainPrices.isPremium()))
-                    .build())
-            .setRegistry(registry)
-            .setDomainName(InternetDomainName.from(domainName))
-            .setAsOfDate(dateTime)
-            .setYears(years)
-            .build());
-  }
-
   /** Returns a new renewal cost for the pricer. */
-  @SuppressWarnings("unused")
   FeesAndCredits getRenewPrice(
       Registry registry,
       String domainName,
@@ -138,20 +115,30 @@ public final class DomainPricingLogic {
       @Nullable Recurring recurringBillingEvent)
       throws EppException {
     checkArgument(years > 0, "Number of years must be positive");
-    Money renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
-    if (recurringBillingEvent != null) {
+    Money renewCost;
+    boolean isRenewCostPremiumPrice;
+    // recurring billing event is null if the domain is still available. Billing events are created
+    // in the process of domain creation.
+    if (recurringBillingEvent == null) {
+      DomainPrices domainPrices = getPricesForDomainName(domainName, dateTime);
+      renewCost = domainPrices.getRenewCost().multipliedBy(years);
+      isRenewCostPremiumPrice = domainPrices.isPremium();
+    } else {
       switch (recurringBillingEvent.getRenewalPriceBehavior()) {
         case DEFAULT:
-          renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+          DomainPrices domainPrices = getPricesForDomainName(domainName, dateTime);
+          renewCost = domainPrices.getRenewCost().multipliedBy(years);
+          isRenewCostPremiumPrice = domainPrices.isPremium();
           break;
           // if the renewal price behavior is specified, then the renewal price should be the same
           // as the creation price, which is stored in the billing event as the renewal price
         case SPECIFIED:
-          checkArgument(
-              recurringBillingEvent.getRenewalPrice().isPresent(),
+          checkArgumentPresent(
+              recurringBillingEvent.getRenewalPrice(),
               "Unexpected behavior: renewal price cannot be null when renewal behavior is"
                   + " SPECIFIED");
           renewCost = recurringBillingEvent.getRenewalPrice().get().multipliedBy(years);
+          isRenewCostPremiumPrice = false;
           break;
           // if the renewal price behavior is nonpremium, it means that the domain should be renewed
           // at standard price of domains at the time, even if the domain is premium
@@ -160,6 +147,7 @@ public final class DomainPricingLogic {
               Registry.get(getTldFromDomainName(domainName))
                   .getStandardRenewCost(dateTime)
                   .multipliedBy(years);
+          isRenewCostPremiumPrice = false;
           break;
         default:
           throw new IllegalArgumentException(
@@ -174,10 +162,7 @@ public final class DomainPricingLogic {
                 new FeesAndCredits.Builder()
                     .setCurrency(renewCost.getCurrencyUnit())
                     .addFeeOrCredit(
-                        Fee.create(
-                            renewCost.getAmount(),
-                            FeeType.RENEW,
-                            PricingEngineProxy.isDomainPremium(domainName, dateTime)))
+                        Fee.create(renewCost.getAmount(), FeeType.RENEW, isRenewCostPremiumPrice))
                     .build())
             .setRegistry(registry)
             .setDomainName(InternetDomainName.from(domainName))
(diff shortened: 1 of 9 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @rachelguan and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 611 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A5R4tC-2DSj4NxyQu:-N1A5R4tC-2DSj4NxyQv:b-gn4ady) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L611)):*
> ```Java
> 
>   /** Returns the billing price for renewing the specified domain name for one year. */
>   public static Money getDomainRenewPrice(
> ```

This feels duplicative of `DomainPricingLogic.getRenewPrice()` (and `DomainFlowUtils` is a worse place for this logic to go than `DomainPricingLogic`).  Merge the additional functionality you've added here into `DomainPricingLogic.getRenewPrice()` and then remove that.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 706 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A61GH07HbDgO_MStK:-N1A61GH07HbDgO_MStL:b-fngjsl) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L706)):*
> ```Java
>       case RENEW:
>         builder.setAvailIfSupported(true);
>         // There are cases when the domain does not contain billing event such as checking the fee
> ```

Can you expand on this reasoning exactly?  How can a domain exist yet not have an autorenew billing event?  (Also you should clarify that it's the autorenew billing even you're talking about in this comment.)  Presumably the billing event exists, it's just not loaded correctly?  But if it does exist, which it must, and you don't use it, then aren't you returning the wrong information?
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 707 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A5wO7CAb7t_12xZKO:-N1A5wO7CAb7t_12xZKP:bb64e1a) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L707)):*
> ```Java
>         builder.setAvailIfSupported(true);
>         // There are cases when the domain does not contain billing event such as checking the fee
>         // extension of the renew command via {@link DomainInfoFlow}
> ```

Missing period.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 712 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A6NpRE__MRFVnrL0n:-N1A6NpRE__MRFVnrL0o:b-amddyz) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L712)):*
> ```Java
>         // event, unless it's loaded for a restore check. The renew price in both cases would be
>         // whatever coming from the pricing engine.
>         if (domain.isPresent() && domain.get().getAutorenewBillingEvent() != null) {
> ```

How is `domain.get().getAutorenewBillingEvent()` ever `null`?  I can see the `domain` object not yet existing during a domain check flow, but if the domain exists then surely it must have an autorenew billing event?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 611 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A5R4tC-2DSj4NxyQu:-N1Aw2GqCCdaWWBx6PeJ:b-exujz4) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L611)):*

This feels duplicative of `DomainPricingLogic.getRenewPrice()` (and `DomainFlowUtils` is a worse place for this logic to go than `DomainPricingLogic`).  Merge the additional functionality you've added here into `DomainPricingLogic.getRenewPrice()` and then remove that.
</blockquote></details>

The main reason I added here instead of pricing logic was that there's no test file for `DomainPricingLogic` and other logic files and this logic was also used in the util file. I could certainly add them to `DomainPricingLogic`.Another reason was the name of the existing method.. since there seems to be a difference between price and cost and price is the amount per year and cost is the price x number of years. The existing method is `getDomainRenewPrice` that takes `years` as one of the parameters, which shouldn't be the case. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 706 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A61GH07HbDgO_MStK:-N1AwuL09G9blySjK345:b-plxvvq) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L706)):*

Can you expand on this reasoning exactly?  How can a domain exist yet not have an autorenew billing event?  (Also you should clarify that it's the autorenew billing even you're talking about in this comment.)  Presumably the billing event exists, it's just not loaded correctly?  But if it does exist, which it must, and you don't use it, then aren't you returning the wrong information?
</blockquote></details>

I added these comments after going over the test cases for `DomainCheckFlow` and `DomainInfoFlow`, where some of the test cases require the domain to be available (not persisted) for renew flow as well, or no billing event. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 707 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A5wO7CAb7t_12xZKO:-N1B1_SZ9VY4iGHueAA5:b-yuhh12) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L707)):*

Missing period.
</blockquote></details>

will fix that once we determine what explanation I need to include here. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 712 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A6NpRE__MRFVnrL0n:-N1AxDV0CLzGF9PGSqpK:b-p1n1mw) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L712)):*

How is `domain.get().getAutorenewBillingEvent()` ever `null`?  I can see the `domain` object not yet existing during a domain check flow, but if the domain exists then surely it must have an autorenew billing event?
</blockquote></details>

I discovered that via testing both `DomainCheckFlowTest` and `DomainInfoFlowTest`. 

In `DomainInfoFlowTest.java`, two cases failed because of NullPointerException. The reason was that there expected to be no history and billing according to the test case. 

```  @TestOfyAndSql
  void testFeeExtension_renewCommand() throws Exception {
    setEppInput(
        "domain_info_fee.xml",
        updateSubstitutions(
            SUBSTITUTION_BASE,
            "COMMAND", "renew",
            "PERIOD", "2"));
    persistTestEntities(false);
    doSuccessfulTest(
        "domain_info_fee_response.xml",
        false,
        ImmutableMap.of(
            "COMMAND", "renew",
            "DESCRIPTION", "renew",
            "PERIOD", "2",
            "FEE", "22.00"));
  }
```

which calls `doSuccessfulTest` which is 
```
 private void doSuccessfulTest(
      String expectedXmlFilename, boolean inactive, ImmutableMap<String, String> substitutions)
      throws Exception {
    doSuccessfulTest(expectedXmlFilename, inactive, substitutions, false);
  }
``` 

I only found one test case in this file that has `expectHistoryAndBilling` set to `true` which is ` testSuccess_autoRenewGracePeriod`

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 611 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A5R4tC-2DSj4NxyQu:-N1LIJoxE8aGKc9qL0_I:b-am1vk7) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L611)):*

Feel free to add tests for `DomainPricingLogic`.  That's the place where this stuff belongs.  Also, to the extent that the current method's signature is sub-optimal, feel free to improve it. I don't believe it's yet being called from everywhere. But generally applicable domain pricing logic should live there, not in the utils class specific to domain flows.

Also, you could just call it with a value of `1` for the years parameter for the create use case and then with the correct actual value on renews (can be up to 9) and transfers (can also be up to 9 on superuser transfers).  Having that parameter in there is reasonable, and you can imagine plausible scenarios in which a domain pricing custom logic might want to give some sort of discount for a longer-term registration, so it makes sense to have the years of renewal be passed in to that logic.  Also note that creates can be for up to 10 years, and that that is passed into `DomainPricingLogic.getCreatePrice()` as well.

The overall way the layers of this onion are structured is:

```
domain flows -> DomainFlowUtils -> DomainPricingLogic -> PricingEngineProxy -> pricing engine (which so far is always StaticPremiumListPricingEngine)
anything else doing pricing such as expand recurring billing events -> DomainPricingLogic -> PricingEngineProxy -> pricing engine
```

You've taken that existing structure and inverted the ordering of it and now added a case where `DomainPricingLogic` is instead calling `DomainFlowUtils`, which is confusing the overall design.
</blockquote></details>

understand. I moved this to `DomainPricingLogic` and added a test file along with test cases for this. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 706 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A61GH07HbDgO_MStK:-N1LIeJl0O0SRIY3fork:b-g1vzxb) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L706)):*

Can you clarify that in this code comment then? Something like "When the renewal price is being checked on a non-existent domain, the billing recurrence is not present."

The existing comment doesn't actually explain this use case, and it also talks about `DomainInfoFlow`, but DomainInfoFlow is only for domains that exist; it will fail if the domain doesn't exist; see: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/domain/DomainInfoFlow.java;l=100-101?q=domaininfoflow&ss=nomulus%2Fnomulus

So the comment needs to be talking about the domain check flow, not the domain info flow.  That's what threw me off here.
</blockquote></details>

Revisited the existing test cases and fixed them by persisting recurring events
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 712 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A6NpRE__MRFVnrL0n:-N1LIncc4Xi4FCS3X5Dy:b74f1me) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L712)):*

`DomainInfoFlow` is supposed to fail early on if the domain doesn't actually exist. That should be happening prior to handling the fee extension. So it sounds like some reordering of execution may be required here?

But a basic invariant of the system is that, if the domain exists, then the billing recurrence must too also exist.  There are no situations in which the domain exists but the billing recurrence does not. To the extent that is happening in tests, is because the tests are wrong, and aren't setting up the associated billing recurrence events as well (which they may not have previously needed, if we weren't previously loading them in this flow because we didn't need them to display the correct price). But the correct resolution to this is not to make the business logic handle the bad tests, but to fix the tests so they more accurately represent business logic situations.
</blockquote></details>

I understand this logic but I did put in more faith in test cases, thinking maybe that's supposed to be the logic for the check and info flow. Fixed them.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 712 at r1](https://reviewable.io/reviews/google/nomulus/1610#-N1A6NpRE__MRFVnrL0n:-N1Px_bw4GhOn62NEKmb:b-1epbi0) ([raw file](https://github.com/google/nomulus/blob/2a5197f71263722e047350d9d57cd76f1b4e2166/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L712)):*

I understand this logic but I did put in more faith in test cases, thinking maybe that's supposed to be the logic for the check and info flow. Fixed them. 
</blockquote></details>

The ultimate decider of what is correct is what the intended business logic is. Sometimes the tests are simply incomplete, and only mock up the immediate stuff relevant to a given test, not the entire intended state of the system.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 5 unresolved discussions (waiting on @rachelguan and @weiminyu)

___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 133 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzL_h83Sq77Dgplub:-N1PzL_h83Sq77Dgpluc:bf4fgn4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L133)):*
> ```Java
>                 .setDomain(domain)
>                 .setType(HistoryEntry.Type.DOMAIN_CREATE)
>                 .setModificationTime(clock.nowUtc())
> ```

This should be the creation time of the domain.
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 143 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzM2L4ZHmCFmuowdP:-N1PzM2L4ZHmCFmuowdQ:bf4fgn4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L143)):*
> ```Java
>                 .setTargetId(domain.getDomainName())
>                 .setRegistrarId("TheRegistrar")
>                 .setEventTime(clock.nowUtc())
> ```

This should be the creation time of the domain.
___
*[`core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java` line 219 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzBSR1M4iHXpEV3zR:-N1PzBSR1M4iHXpEV3zS:bf4fgn4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java#L219)):*
> ```Java
>                 .setRegistrarId(domain.getCreationRegistrarId())
>                 .setType(HistoryEntry.Type.DOMAIN_CREATE)
>                 .setModificationTime(clock.nowUtc())
> ```

This should be the creation time of the domain.
___
*[`core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java` line 227 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzDxe9PDg7I4FmSh8:-N1PzDxe9PDg7I4FmSh9:bf4fgn4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java#L227)):*
> ```Java
>                 .setParent(historyEntry)
>                 .setRegistrarId(domain.getCreationRegistrarId())
>                 .setEventTime(clock.nowUtc())
> ```

This should be the creation time of the domain.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 1 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Pzpap0s_el7mwrzOk:-N1Pzpap0s_el7mwrzOl:b-cv6hdp) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L1)):*
> ```XML
> <epp xmlns:domain="urn:ietf:params:xml:ns:domain-1.0" xmlns:contact="urn:ietf:params:xml:ns:contact-1.0" xmlns:fee="urn:ietf:params:xml:ns:fee-0.6" xmlns="urn:ietf:params:xml:ns:epp-1.0" xmlns:rgp="urn:ietf:params:xml:ns:rgp-1.0" xmlns:fee11="urn:ietf:params:xml:ns:fee-0.11" xmlns:fee12="urn:ietf:params:xml:ns:fee-0.12" xmlns:launch="urn:ietf:params:xml:ns:launch-1.0" xmlns:secDNS="urn:ietf:params:xml:ns:secDNS-1.1" xmlns:host="urn:ietf:params:xml:ns:host-1.0">
> ```

The changes in this file aren't making sense to me.  Why are you pulling in the `fee12` namespace when this file is specifically about testing the v0.6 version of the fee extension?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 6 unresolved discussions (waiting on @rachelguan and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 110 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Q-VC7DjWzk3syVzPv:-N1Q-VC7DjWzk3syVzPw:bl3p325) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L110)):*
> ```Java
>   }
> 
>   // TODO (rachelguan): this will be replaced by the method getRenewPrice below in PR#1592
> ```

That PR has TODOs blocking on this PR going in. Which one is going in first?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @rachelguan, and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 110 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Q-VC7DjWzk3syVzPv:-N1Q5BFt0D_-0E8DuSAn:b-26v2ad) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L110)):*

That PR has TODOs blocking on this PR going in. Which one is going in first?
</blockquote></details>

so .. this one should go in first ..because the new `getRenewPrice()` will be used.because of an additional parameter, it affects the other PRs (renew flow and expand recurring billing event action).. so the original `getRenewPrice()` will be here for now .. the new one will be used and the old one will be removed in both PR. 
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 133 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzL_h83Sq77Dgplub:-N1Q7BGi5NFCxTiOKDYI:bq1dsl4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L133)):*

This should be the creation time of the domain.
</blockquote></details>

Done and fixed similar issue.
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 143 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzM2L4ZHmCFmuowdP:-N1Q7GVd1wOzzn6PdRFj:bq1dsl4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L143)):*

This should be the creation time of the domain.
</blockquote></details>

Done and fixed similar issue.
___
*[`core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java` line 219 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzBSR1M4iHXpEV3zR:-N1Q7SBFEKbfa9w5CWe2:bq1dsl4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java#L219)):*

This should be the creation time of the domain.
</blockquote></details>

Done and fixed similar issue.
___
*[`core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java` line 227 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1PzDxe9PDg7I4FmSh8:-N1Q7RlHClNPw4-VECKH:bq1dsl4) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java#L227)):*

This should be the creation time of the domain.
</blockquote></details>

Done and fixed similar issue.

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 1 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Pzpap0s_el7mwrzOk:-N1Q8pRjAf-YMOeqooeK:b38ey1t) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L1)):*

The changes in this file aren't making sense to me.  Why are you pulling in the `fee12` namespace when this file is specifically about testing the v0.6 version of the fee extension?
</blockquote></details>

I am actually not sure how to fix this ..after I added billing event..this test case broke and i looked into the data..the data remain the same but the tag info goes to epp.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @rachelguan and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 110 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Q-VC7DjWzk3syVzPv:-N1p1ITU0O2moj8DtIy_:b8i1p0n) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L110)):*

so .. this one should go in first ..because the new `getRenewPrice()` will be used.because of an additional parameter, it affects the other PRs (renew flow and expand recurring billing event action).. so the original `getRenewPrice()` will be here for now .. the new one will be used and the old one will be removed in both PR. 
</blockquote></details>

Why not remove the old one here and now?  It's not used by anything, right?
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 134 at r4](https://reviewable.io/reviews/google/nomulus/1610#-N1p0ERkDs0MCFDkdDnc:-N1p0ERkDs0MCFDkdDnd:b-if0g3) ([raw file](https://github.com/google/nomulus/blob/7e6c51cc89039745c3413e75ff798670532b36d3/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L134)):*
> ```Java
>    * Calculates the renewal cost of a domain.
>    *
>    * <p>This method is set as static which allows it to be used without instantiating a {@link
> ```

Make the method that calls it non-static.

The way Dagger works is that you use instances, not static methods, and dependencies are injected as instance fields.  You should continue using that pattern here.

The correct solution here is to use Dagger and fix whatever issue you were having with injecting `DomainPricingLogic`, not just giving up on using Dagger for one particular use case.
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 177 at r4](https://reviewable.io/reviews/google/nomulus/1610#-N1p0PVcD0IZSfifP7FU:-N1p0PVcD0IZSfifP7FV:bi2e1rr) ([raw file](https://github.com/google/nomulus/blob/7e6c51cc89039745c3413e75ff798670532b36d3/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L177)):*
> ```Java
>   /** Returns a new renewal cost for the pricer. */
>   @SuppressWarnings("unused")
>   FeesAndCredits getRenewPrice(
> ```

You've now got _three_ `getRenewPrice()` methods now.  I'm really not understanding what's going on.  There should only be one.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 1 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Pzpap0s_el7mwrzOk:-N1p0rui6to2FrWHbXf-:b-f8t135) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L1)):*

I am actually not sure how to fix this ..after I added billing event..this test case broke and i looked into the data..the data remain the same but the tag info goes to epp. 
</blockquote></details>

It's not inconsistent with some other test cases, anyway.  Maybe it's OK.  Is this test case still failing or is it working?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 110 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Q-VC7DjWzk3syVzPv:-N1uZY891u4Ei2paxPtX:b-bz6o9a) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L110)):*
> make sense. 
Used by `DomainRenewFlow` 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 134 at r4](https://reviewable.io/reviews/google/nomulus/1610#-N1p0ERkDs0MCFDkdDnc:-N1uYwfeENmXoTqNPlOl:b-jzqc40) ([raw file](https://github.com/google/nomulus/blob/7e6c51cc89039745c3413e75ff798670532b36d3/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L134)):*

Make the method that calls it non-static.

The way Dagger works is that you use instances, not static methods, and dependencies are injected as instance fields.  You should continue using that pattern here.

The correct solution here is to use Dagger and fix whatever issue you were having with injecting `DomainPricingLogic`, not just giving up on using Dagger for one particular use case.
</blockquote></details>

I was trying to figure out a way to inject `DomainPricingLogic` to an action file and I ran into issues. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 177 at r4](https://reviewable.io/reviews/google/nomulus/1610#-N1p0PVcD0IZSfifP7FU:-N1uZo05CVh7rOMe1bjm:b-qff6so) ([raw file](https://github.com/google/nomulus/blob/7e6c51cc89039745c3413e75ff798670532b36d3/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L177)):*

You've now got _three_ `getRenewPrice()` methods now.  I'm really not understanding what's going on.  There should only be one.
</blockquote></details>

indeed..one of them shall be removed since it will be used by an action file and renew method.. 
Two were there because the return type was different.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 1 at r3](https://reviewable.io/reviews/google/nomulus/1610#-N1Pzpap0s_el7mwrzOk:-N1p2r6o8Y6qRHKjAEZe:b3livu9) ([raw file](https://github.com/google/nomulus/blob/0630d2710eb4ab8b29166b3e2b03fbfeacc5fade/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L1)):*

It's not inconsistent with some other test cases, anyway.  Maybe it's OK.  Is this test case still failing or is it working?
</blockquote></details>

It fails if I don't change the xml tags to to include other versions in the tag.

rachelguan (commented):
Moved check and flow info related code caused by `DomainFlowUtil` to a separate PR..This PR now only contains code related to adding `getRenewPrice()` to `DomainPricingLogic` and the old one will be removed in the PR for renew flow, which is [PR#1592](https://github.com/google/nomulus/pull/1592/files)

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610#-:-N2Ng89j6WbvD0Ry3qxN:b-yceeji)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @weiminyu)

core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
```diff
@@ -102,6 +107,7 @@ FeesAndCredits getCreatePrice(
             .build());
   }
 
+  // TODO (rachelguan): this will be replaced by the method getRenewPrice below in PR#1592
```
CydeWeys, line 110: Can you just have one `getRenewPrice()` method?  Seeing as how the existing one is unused, I don't see why you're retaining it.  Just delete it and replace it with the new one.

core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
```diff
@@ -122,6 +128,64 @@ FeesAndCredits getRenewPrice(Registry registry, String domainName, DateTime date
...
+    Money renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+    if (recurringBillingEvent != null) {
+      switch (recurringBillingEvent.getRenewalPriceBehavior()) {
+        case DEFAULT:
+          renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
+          break;
+          // if the renewal price behavior is specified, then the renewal price should be the same
+          // as the creation price, which is stored in the billing event as the renewal price
+        case SPECIFIED:
+          checkArgument(
```
CydeWeys, line 150: Use `checkArgumentPresent()`

core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
```diff
@@ -122,6 +128,64 @@ FeesAndCredits getRenewPrice(Registry registry, String domainName, DateTime date
...
+  @SuppressWarnings("unused")
+  FeesAndCredits getRenewPrice(
+      Registry registry,
+      String domainName,
+      DateTime dateTime,
+      int years,
+      @Nullable Recurring recurringBillingEvent)
+      throws EppException {
+    checkArgument(years > 0, "Number of years must be positive");
+    Money renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
```
CydeWeys, line 141: Don't initialize this here. It's confusing, as it's always being overwritten later on (by even the same thing in the `DEFAULT` case, so you're calling the pricing proxy twice for no reason on that path).

core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
```diff
@@ -122,6 +128,64 @@ FeesAndCredits getRenewPrice(Registry registry, String domainName, DateTime date
...
+    return customLogic.customizeRenewPrice(
+        RenewPriceParameters.newBuilder()
+            .setFeesAndCredits(
+                new FeesAndCredits.Builder()
+                    .setCurrency(renewCost.getCurrencyUnit())
+                    .addFeeOrCredit(
+                        Fee.create(
+                            renewCost.getAmount(),
+                            FeeType.RENEW,
+                            PricingEngineProxy.isDomainPremium(domainName, dateTime)))
```
CydeWeys, line 180: I think this might do the wrong thing for the `SPECIFIED` and `NONPREMIUM` cases.  It seems to me it should only be possible for `isPremium` to be true on the `DEFAULT` renewal behavior?

CydeWeys: Also, this needs test coverage.  Your tests below are only making assertions on the monetary amount, but aren't testing the fee type or the `isPremium` boolean.

core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
CydeWeys: Yeah, I think you should only be calling this in the else.  So it'd look something like:

```
Money renewCost;
if (recurringBillingEvent == null) {
  renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
} else {
  switch(...) {
    ......
  }
}
```

core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java
```diff
@@ -122,6 +128,64 @@ FeesAndCredits getRenewPrice(Registry registry, String domainName, DateTime date
             .build());
   }
 
+  /** Returns a new renewal cost for the pricer. */
+  @SuppressWarnings("unused")
```
CydeWeys, line 132: Once this is being used you shouldn't need this (also technically we already no longer need it on the existing `getRenewPrice()`, as that method is in fact used in 2 places, so its presence is already confusing).

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610)* status: 0 of 9 files reviewed, 10 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 110 at r5](https://reviewable.io/reviews/google/nomulus/1610#-N0lUqMnX86LUWlE9Jzl-r5-110:-N2NnKVE6J4bgSNZaxiF:bf0ili5) ([raw file](https://github.com/google/nomulus/blob/87546ea0d09857fd1d87398ba7ce6662254b039d/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L110)):*

Can you just have one `getRenewPrice()` method?  Seeing as how the existing one is unused, I don't see why you're retaining it.  Just delete it and replace it with the new one.
</blockquote></details>

Replaced . 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 132 at r5](https://reviewable.io/reviews/google/nomulus/1610#-N0lUqMnX86LUWlE9Jzl-r5-132:-N2Ns13R2XM5JXp6NnAs:b-5dfryo) ([raw file](https://github.com/google/nomulus/blob/87546ea0d09857fd1d87398ba7ce6662254b039d/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L132)):*

Once this is being used you shouldn't need this (also technically we already no longer need it on the existing `getRenewPrice()`, as that method is in fact used in 2 places, so its presence is already confusing).
</blockquote></details>

removed it
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 141 at r5](https://reviewable.io/reviews/google/nomulus/1610#-N0lUqMnX86LUWlE9Jzl-r5-141:-N2Nneiu84QUBNjLOF-g:b-nyzyog) ([raw file](https://github.com/google/nomulus/blob/87546ea0d09857fd1d87398ba7ce6662254b039d/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L141)):*

Yeah, I think you should only be calling this in the else.  So it'd look something like:

```
Money renewCost;
if (recurringBillingEvent == null) {
  renewCost = PricingEngineProxy.getDomainRenewCost(domainName, dateTime, years);
} else {
  switch(...) {
    ......
  }
}
```
</blockquote></details>

yes..changed to that. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 142 at r5](https://reviewable.io/reviews/google/nomulus/1610#-N0lUqMnX86LUWlE9Jzl-r5-142:-N2NqyI76g6KZ3s3SmYm:b-rnrapz) ([raw file](https://github.com/google/nomulus/blob/87546ea0d09857fd1d87398ba7ce6662254b039d/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L142)):*

Add a comment here explaining under what circumstances this is expected to be null / not null.
</blockquote></details>

done. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 150 at r5](https://reviewable.io/reviews/google/nomulus/1610#-N0lUqMnX86LUWlE9Jzl-r5-150:-N2NrOJdBlzxGf99uJgG:b-896fix) ([raw file](https://github.com/google/nomulus/blob/87546ea0d09857fd1d87398ba7ce6662254b039d/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L150)):*

Use `checkArgumentPresent()`
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 180 at r5](https://reviewable.io/reviews/google/nomulus/1610#-N0lUqMnX86LUWlE9Jzl-r5-180:-N2NrQ8gDc99p9jJT8vx:buakcl7) ([raw file](https://github.com/google/nomulus/blob/87546ea0d09857fd1d87398ba7ce6662254b039d/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L180)):*

Also, this needs test coverage.  Your tests below are only making assertions on the monetary amount, but aren't testing the fee type or the `isPremium` boolean.
</blockquote></details>

There's no really indicator if it's supposed to reflect the type of the fee or the domain. But since it's part of `FeesAndCredits` which is an extended class of `BaseFee`. It makes sense to set that to false for anchor tenants and internal registrations. 

And replaced all the assertions on the amount to the assertions o n `FeesAndCredit` objects.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1610#-:-N2Sribm9Hq1EqZlv3fu:bnfp4nl)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @rachelguan and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 119 at r8](https://reviewable.io/reviews/google/nomulus/1610#-N2SrGWc-L4W-Qnc9oGz:-N2SrGWc-L4W-Qnc9oH-:b1fli0b) ([raw file](https://github.com/google/nomulus/blob/6d2d96c6728f8d458a213fb18623bfb2880e35f6/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L119)):*
> ```Java
>     checkArgument(years > 0, "Number of years must be positive");
>     Money renewCost;
>     boolean isRenewCostPremiumPrice;
> ```

`isPremium` works too and is shorter, fwiw

PR #1618, opened 2022-05-04
Add test for Java 8 Compatibility of third party jars

weiminyu:
Add a test to check for Java 8 compatibility of all jars deployed to
AppEngine. This is a follow-up of pr 1586, and ensures that third
party dependencies are also compatible with Java 8

It is not enough to run existing tests with Java 8 VM, since many API
jars are not exercised by tests. For example, those for GCP services
like the SecretManager.

We take the conservative approach and verify that every class in every
jar are compiled for Java 8.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1618)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1618)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[`java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java` line 76 at r1](https://reviewable.io/reviews/google/nomulus/1618#-N1EQqLR7TIzAn3_YA_E:-N1EQqLR7TIzAn3_YA_F:bz49kus) ([raw file](https://github.com/google/nomulus/blob/8baaf4960e5ef0c211356c5d5adcb0e1593a3201/java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java#L76)):*
> ```Java
>   private static Stream<Arguments> provideJarNames() {
>     return jarsPaths.stream()
>         .filter(pathStr -> !Objects.equals(pathStr, "/usr/lib/jvm/lib/tools.jar"))
> ```

Are we sure that this location will remain the same in all contexts?  We might just check pathStr.endsWith("/tools.jar") for portability?



_Code quote:_
```Java
"/usr/lib/jvm/lib/tools.jar")
```

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1618)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @mindhog)

___
*[`java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java` line 76 at r1](https://reviewable.io/reviews/google/nomulus/1618#-N1EQqLR7TIzAn3_YA_E:-N1EXBZUDavTcAObyovJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/8baaf4960e5ef0c211356c5d5adcb0e1593a3201/java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java#L76)):*

Are we sure that this location will remain the same in all contexts?  We might just check pathStr.endsWith("/tools.jar") for portability?
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1618)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @mindhog and @weiminyu)

___
*[`java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java` line 105 at r2](https://reviewable.io/reviews/google/nomulus/1618#-N1E_sMz6MNLK79553Cf:-N1E_sMz6MNLK79553Cg:b-fqnr3e) ([raw file](https://github.com/google/nomulus/blob/ca6669a1c975c53e543a2e2b5af692dd60eab68a/java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java#L105)):*
> ```Java
>     try (InputStream inputStream = jarFile.getInputStream(entry)) {
>       byte[] header = inputStream.readNBytes(HEADER_LENGTH);
>       Verify.verify(
> ```

Any particular reason you're using `verify`, both here and below in `throw new VerifyException`, rather than just using `assertThat()` statements?
___
*[`java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/1618#-N1EaBJW8vbwAJYph9sM:-N1EaBJW8vbwAJYph9sN:ba5r5ww) ([raw file](https://github.com/google/nomulus/blob/ca6669a1c975c53e543a2e2b5af692dd60eab68a/java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java#L113)):*
> ```Java
>           String.format("Incompatible with Java 8: Class %s in %s.", entry.getName(), jarPath));
>       throw new RuntimeException(e);
> ```

Rather than rewrapping this to be a non-checked-exception, I think it's cleaner if you just declare the method signature as `throws IOException`, and then from where this is called above, use a `for` loop as necessary if the Stream API can't handle checked exceptions.  It's still simpler overall.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1618)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

___
*[`java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java` line 105 at r2](https://reviewable.io/reviews/google/nomulus/1618#-N1E_sMz6MNLK79553Cf:-N1ElLzY5xSo-gXOtHvR:b-e0iu0i) ([raw file](https://github.com/google/nomulus/blob/ca6669a1c975c53e543a2e2b5af692dd60eab68a/java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java#L105)):*

Any particular reason you're using `verify`, both here and below in `throw new VerifyException`, rather than just using `assertThat()` statements?
</blockquote></details>

Done.

I originally wrote it as a tool, then moved the code to test.
___
*[`java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/1618#-N1EaBJW8vbwAJYph9sM:-N1ElLWX6M7VS0CkbdTK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca6669a1c975c53e543a2e2b5af692dd60eab68a/java8compatibility/src/test/java/google/registry/java8compatibility/ByteCodeVersionTest.java#L113)):*

Rather than rewrapping this to be a non-checked-exception, I think it's cleaner if you just declare the method signature as `throws IOException`, and then from where this is called above, use a `for` loop as necessary if the Stream API can't handle checked exceptions.  It's still simpler overall.
</blockquote></details>

Done.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1618)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @mindhog)

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1618#-:-N1F_EsIEDF_GLB3bbvo:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

PR #1619, opened 2022-05-04
Do not delete build cache when building release candidates

jianglai:
We would like to re-use the build cache when building RCs for different
environments. There's not much practical use in doing a "clean" for
every build when Gradle should be able to figure out which artifacts
need to be rebuilt. It also does not make sense to build each
environment in a separate step, which also introduces redunency because
not all artifacts are cached across steps. The build cache is enabled by
default.

Lastly, the cache needs to be inside the /workspace folder, which is the
default persisted storage location.

TESTED=tried to build the RCs on alpha and saved about 10 min.


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1619)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1625, opened 2022-05-05
Remove to-be-deprecated OOB OAuth flow in nomulus login

jianglai:
The Out-of-Band flow will be deprecated in October 2022. Unfortunately
there is no alternative provided that can perform login on a remote host
(which is probably WAI as the OOB flow is susceptible to remote phishing
attacks).

In this PR we switch to using the same loopback IP address redirect URI that a
normal local login session uses, and added instructions on how to set up
SSH port forwarding to complete the flow on a local host when the nomulus
tool runs remotely.

See https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html for
more details on the deprecation.

TESTED=run the new login flow from a remote host and verified it worked.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1625)

### DIFF
--- a/core/src/main/java/google/registry/tools/ForwardingServerReceiver.java
+++ b/core/src/main/java/google/registry/tools/ForwardingServerReceiver.java
@@ -1,0 +1,72 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools;
+
+import com.google.api.client.extensions.java6.auth.oauth2.VerificationCodeReceiver;
+import com.google.api.client.extensions.jetty.auth.oauth2.LocalServerReceiver;
+import java.io.IOException;
+
+/**
+ * A thin wrapper around {@link LocalServerReceiver} which points the redirect URI to a different
+ * port (the forwarding port) while still listening on the random unused port (the remote port)
+ * nomulus itself picks. This allows us to run the nomulus tool on a remote host (which one can SSH
+ * into) while performing the OAuth 3-legged login flow in a local browser (from the lost host where
+ * the SSH client resides).
+ *
+ * <p>When performing the login flow, an HTTP server will be listening on the remote port and have a
+ * redirect_uri of <code>http://localhost:remote_port</code>, which is only accessible from the
+ * remote host. By changing the redirect_uri to <code>http://localhost:forwarding_port</code>, it
+ * becomes accessible from the local host, if <code>local_host:forwarding_port</code> is forwarded
+ * to <code>remote_host:remote_port</code>.
+ *
+ * <p>Note that port forwarding is <b>required</b>. We cannot use the remote host's IP or reverse
+ * DNS address in the redirect URI, even if they are directly accessible from the local host,
+ * because the only allowed redirect URI scheme for desktops apps when sending a request to the
+ * Google OAuth server is the loopback address with a port.
+ *
+ * @see <href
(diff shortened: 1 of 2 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1625)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/LoginCommand.java` line 35 at r1](https://reviewable.io/reviews/google/nomulus/1625#-N1Q6qADEW4RSduQMT1-:-N1Q6qADEW4RSduQMT10:b7bw223) ([raw file](https://github.com/google/nomulus/blob/cb44a9feeac015378b30dee27361b3c77e3f1923/core/src/main/java/google/registry/tools/LoginCommand.java#L35)):*
> ```Java
>       names = "--port",
>       description =
>           "A free port on the local host. When set, it is assumed that the nomulus tool runs on a"
> ```

Why not just pick a free one randomly, or use a default value here that is unlikely to normally be in use?

Forcing the user to pick a port isn't a great user experience.  Usually this isn't something you'd have to deal with.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1625)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/LoginCommand.java` line 57 at r1](https://reviewable.io/reviews/google/nomulus/1625#-N1Q794J1vs2NorO5iFS:-N1Q794J1vs2NorO5iFT:b-d0069i) ([raw file](https://github.com/google/nomulus/blob/cb44a9feeac015378b30dee27361b3c77e3f1923/core/src/main/java/google/registry/tools/LoginCommand.java#L57)):*
> ```Java
>                     "Please first run the following command in a separate terminal on"
>                         + " your local host:\n\n"
>                         + "  ssh -L "
> ```

Use better String formatting than this, e.g. `System.out.printf()`

jianglai (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1625)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/LoginCommand.java` line 35 at r1](https://reviewable.io/reviews/google/nomulus/1625#-N1Q6qADEW4RSduQMT1-:-N1RTCNw5AgrIWROzDXg:b-l9go7p) ([raw file](https://github.com/google/nomulus/blob/cb44a9feeac015378b30dee27361b3c77e3f1923/core/src/main/java/google/registry/tools/LoginCommand.java#L35)):*

Why not just pick a free one randomly, or use a default value here that is unlikely to normally be in use?

Forcing the user to pick a port isn't a great user experience.  Usually this isn't something you'd have to deal with.
</blockquote></details>

That would have been nice, wouldn't it :package: 

But this command runs on the remote host, and what we need is to initiate an SSH tunnel from the local host and forward the local port  to the remote host. There is no way for the remote host to know which port is available on the local host.

What we can do is to perform the tunneling and port forwarding from the other side with in Java, i. e. from the remote host to the local host,  to eliminate the need for the user to have to run a command in terminal. But that involves setting up credentials for the remote host to access the local host (not to mention whether the local host is accessible), and still does not solve the problem that it needs to find a free local port. I guess if we really really want to make the user experience nice, we can randomly generate a port number and try to establish the tunnel from the remote host, and retry until we find a free local port.

That said, what I described above is probably over-engineering for minimal benefits...
___
*[`core/src/main/java/google/registry/tools/LoginCommand.java` line 57 at r1](https://reviewable.io/reviews/google/nomulus/1625#-N1Q794J1vs2NorO5iFS:-N1RX-Pj53EOZQrU2kdB:b-896fix) ([raw file](https://github.com/google/nomulus/blob/cb44a9feeac015378b30dee27361b3c77e3f1923/core/src/main/java/google/registry/tools/LoginCommand.java#L57)):*

Use better String formatting than this, e.g. `System.out.printf()`
</blockquote></details>

Done.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1625#-:-N1efgQn72CSWeIpwSz5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1626, opened 2022-05-11
Fix new checkstyle violations

mindhog:
Looks like the new port-forwarding stuff has some naming issues.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1626)

### DIFF
--- a/core/src/main/java/google/registry/tools/ForwardingServerReceiver.java
+++ b/core/src/main/java/google/registry/tools/ForwardingServerReceiver.java
@@ -26,10 +26,10 @@ import java.io.IOException;
  * the SSH client resides).
  *
  * <p>When performing the login flow, an HTTP server will be listening on the remote port and have a
- * redirectUri of <code>http://localhost:remote_port</code>, which is only accessible from the
- * remote host. By changing the redirectUri to <code>http://localhost:forwardingPort</code>, it
- * becomes accessible from the local host, if <code>local_host:forwardingPort</code> is forwarded to
- * <code>remote_host:remote_port</code>.
+ * redirect_uri of <code>http://localhost:remote_port</code>, which is only accessible from the
+ * remote host. By changing the redirect_uri to <code>http://localhost:forwarding_port</code>, it
+ * becomes accessible from the local host, if <code>local_host:forwarding_port</code> is forwarded
+ * to <code>remote_host:remote_port</code>.
  *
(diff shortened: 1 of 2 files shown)


jianglai (changes requested):
Sorry, I don't know how this could have escaped the format checking tool...

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1626#-:-N1ns8sL6qyCrxsD7PF0:b-n88wt8)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[`core/src/main/java/google/registry/tools/ForwardingServerReceiver.java` line 28 at r1](https://reviewable.io/reviews/google/nomulus/1626#-N1ns-tk4mqzjov6j__o:-N1ns-tl2YzYH0wUQTmm:b-puu9t2) ([raw file](https://github.com/google/nomulus/blob/42012273629849670824ed7f7b0e3fb34ecbac42/core/src/main/java/google/registry/tools/ForwardingServerReceiver.java#L28)):*
> ```Java
>  * the SSH client resides).
>  *
>  * <p>When performing the login flow, an HTTP server will be listening on the remote port and have a
> ```

We shouldn't need to change the javadoc. Especially the URL in `href` tag.



_Code quote:_
```Java
 * <p>When performing the login flow, an HTTP server will be listening on the remote port and have a
 * redirectUri of <code>http://localhost:remote_port</code>, which is only accessible from the
 * remote host. By changing the redirectUri to <code>http://localhost:forwardingPort</code>, it
 * becomes accessible from the local host, if <code>local_host:forwardingPort</code> is forwarded to
 * <code>remote_host:remote_port</code>.
 *
 * <p>Note that port forwarding is <b>required</b>. We cannot use the remote host's IP or reverse
 * DNS address in the redirect URI, even if they are directly accessible from the local host,
 * because the only allowed redirect URI scheme for desktops apps when sending a request to the
 * Google OAuth server is the loopback address with a port.
 *
 * @see <href
 *     a=https://developers.google.com/identity/protocols/oauth2/native-app#request-parameter-redirectUri>
 *     redirectUri values </href>
 */
```

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1626)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1628, opened 2022-05-13
Track DNS-Publish retry count

sarahcaseybot:
This is the first PR to implement go/terminatednsupdates

This PR manually changes the result to a SC_NO_CONTENT (204) so that app engine does not automatically retry failing batches. Right now, The retry count is simply incremented, and the task is requeued, but in a follow up PR I will add functionality to split the batch or terminate based on the size of the batch and the number of retries.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1628)

### DIFF
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -14,6 +14,8 @@
 
 package google.registry.dns;
 
+import static google.registry.batch.RelockDomainAction.PREVIOUS_ATTEMPTS_PARAM;
+import static google.registry.dns.DnsConstants.DNS_PUBLISH_PUSH_QUEUE_NAME;
 import static google.registry.dns.DnsModule.PARAM_DNS_WRITER;
 import static google.registry.dns.DnsModule.PARAM_DOMAINS;
 import static google.registry.dns.DnsModule.PARAM_HOSTS;
@@ -25,6 +27,7 @@ import static google.registry.request.Action.Method.POST;
 import static google.registry.request.RequestParameters.PARAM_TLD;
 import static google.registry.util.CollectionUtils.nullToEmpty;
 
+import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
 import google.registry.config.RegistryConfig.Config;
@@ -34,11 +37,13 @@ import google.registry.dns.DnsMetrics.PublishStatus;
 import google.registry.dns.writer.DnsWriter;
 import google.registry.model.tld.Registry;
 import google.registry.request.Action;
+import google.registry.request.Action.Service;
 import google.registry.request.HttpException.ServiceUnavailableException;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
 import google.registry.request.lock.LockHandler;
 import google.registry.util.Clock;
+import google.registry.util.CloudTasksUtils;
 import google.registry.util.DomainNameUtils;
 import java.util.Set;
 import java.util.concurrent.Callable;
@@ -57,6 +62,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
 
   public static final String PATH = "/_dr/task/publishDnsUpdates";
   public static final String LOCK_NAME = "DNS updates";
+  public static final String PREVIOUS_ATTEMPTS_PARAM = "previousAttempts";
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
@@ -83,14 +89,37 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   @Parameter(PARAM_REFRESH_REQUEST_CREATED)
   DateTime itemsCreateTime;
 
-  @Inject @Parameter(PARAM_LOCK_INDEX) int lockIndex;
-  @Inject @Parameter(PARAM_NUM_PUBLISH_LOCKS) int numPublishLocks;
-  @Inject @Parameter(PARAM_DOMAINS) Set<String> domains;
-  @Inject @Parameter(PARAM_HOSTS) Set<String> hosts;
-  @Inject @Parameter(PARAM_TLD) String tld;
-  @Inject LockHandler lockHandler;
-  @Inject Clock clock;
-  @Inject PublishDnsUpdatesAction() {}
+  private final int previousAttempts;
+  private final int lockIndex;
+  private final int numPublishLocks;
+  private final Set<String> domains;
+  private final Set<String> hosts;
+  private final String tld;
+  private final LockHandler lockHandler;
+  private final Clock clock;
+  private CloudTasksUtils cloudTasksUtils;
+
+  @Inject
+  PublishDnsUpdatesAction(
+      @Parameter(PREVIOUS_ATTEMPTS_PARAM) int previousAttempts,
+      @Parameter(PARAM_LOCK_INDEX) int lockIndex,
+      @Parameter(PARAM_NUM_PUBLISH_LOCKS) int numPublishLocks,
+      @Parameter(PARAM_DOMAINS) Set<String> domains,
+      @Parameter(PARAM_HOSTS) Set<String> hosts,
+      @Parameter(PARAM_TLD) String tld,
+      LockHandler lockHandler,
+      Clock clock,
+      CloudTasksUtils cloudTasksUtils) {
+    this.previousAttempts = previousAttempts;
+    this.lockIndex = lockIndex;
+    this.numPublishLocks = numPublishLocks;
+    this.domains = domains;
+    this.hosts = hosts;
+    this.tld = tld;
+    this.lockHandler = lockHandler;
+    this.clock = clock;
+    this.cloudTasksUtils = cloudTasksUtils;
+  }
 
   private void recordActionResult(ActionStatus status) {
     DateTime now = clock.nowUtc();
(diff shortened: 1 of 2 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1628)* status: all files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 17 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2SuJYt7i7SDS7KEQmb:-N2SuJYt7i7SDS7KEQmc:b-dcqhp9) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L17)):*
> ```Java
> package google.registry.dns;
> 
> import static google.registry.batch.RelockDomainAction.PREVIOUS_ATTEMPTS_PARAM;
> ```

If we are reusing this parameter, it is better to move it to a more generic class/module, like `RequestParameters`.



_Code quote:_
```Java
google.registry.batch.RelockDomainAction.PREVIOUS_ATTEMPTS_PARAM;
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 109 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2SurG49QooAjbUc-fb:-N2SurG49QooAjbUc-fc:b-nrvgxs) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L109)):*
> ```Java
>   @Inject
>   PublishDnsUpdatesAction(
>       @Parameter(PREVIOUS_ATTEMPTS_PARAM) int previousAttempts,
> ```

Is it not going to error out when you don't set the query parameter? The initially enqueued task will not have this parameter set, and `Integer.parseInt` will try to parse an empty string and throw. Will it not?

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/RequestParameters.java;l=101;drc=c4cf1288444a4eff908c1c66f152a4e4ee205af5



_Code quote:_
```Java
previousAttempts
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 194 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2St4Bc9HwtZ3d1sjk5:-N2St4Bd3LU0IRqBJXbK:bldxtxv) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L194)):*
> ```Java
>      * See https://cloud.google.com/appengine/docs/standard/java/taskqueue/push/retrying-tasks
>      * for more details on retry behavior. */
>     response.setStatus(SC_NO_CONTENT);
> ```

Is there a reason to set the response status here again? When you call `executeWithLocks()` in `run()`, the response is already set to 204.



_Code quote:_
```Java
    response.setStatus(SC_NO_CONTENT);
    response.setContentType(MediaType.PLAIN_TEXT_UTF_8);
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 288 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2SvRKkBxde4u17Y1Zo:-N2SvRKl8RcrXHO4fPo2:bd23plc) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L288)):*
> ```Java
>     } finally {
>       if (commitStatus.equals(CommitStatus.FAILURE)) {
>         handleCommitFailure();
> ```

Does this actually work? When a retriable exception occurs when writing to Cloud DNS, a [`ZoneStateException`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/dns/writer/clouddns/CloudDnsWriter.java;l=408;drc=ec5c2cdb6890323839731209539da8096a604ac1;bpv=1;bpt=1?q=clouddns&ss=nomulus%2Fnomulus) is thrown.

This is a `RuntimeException` and is not handled by the `try` block, so the `finally` block will run, followed by the default exception handling mechanism for unhandled exceptions, which when running in a Java servlet is to set the response status code to `500`, if I remember correctly. So your response code would not be `204`. This would not be caught by unit test because the behavior of java servlet is not tested in unit tests.

Also if you don't set the response code, it will be `200`, again set by Java servlet runtime.

Overall it seems to be an anti-pattern to set the default response code to `204` and only change it to `200` when things are successful. A better (and easier to understand approach) should be:

```
try {
  ...
} catch (`ZoneStateException` e) {
  // handle rcommit failure
} finally {
  ...
}
```

We should also consider what to do with other runtime errors that Cloud DNS does not consider retriable, if left unhandled, they'll make the response `500` and cause Cloud Tasks to retry, which is probably not what we want.



_Code quote:_
```Java
handleCommitFailure();
```
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 76 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2UHycR7Cl22HrZMJoF:-N2UHycSAUY6on7yX2w7:bdqnc4l) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L76)):*
> ```Java
>   private final DnsMetrics dnsMetrics = mock(DnsMetrics.class);
>   private final DnsQueue dnsQueue = mock(DnsQueue.class);
>   private final CloudTasksUtils cloudTasksUtils = mock(CloudTasksUtils.class);
> ```

Use [`CloudTasksHelper`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/CloudTasksHelper.java).



_Code quote:_
```Java
mock(CloudTasksUtils.class);
```
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 166 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2UJCoGD2F-lSc84So7:-N2UJCoGD2F-lSc84So8:b-dp6hgp) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L166)):*
> ```Java
>     verifyNoMoreInteractions(dnsMetrics);
>     verifyNoMoreInteractions(dnsQueue);
>     assertThat(response.getStatus()).isEqualTo(SC_OK);
> ```

As mentioned above, there's no need to explicitly check the response status. If it is left untouched, the servlet will set it to `200`. Most of our code base is based on this assumption.



_Code quote:_
```Java
assertThat(response.getStatus()).isEqualTo(SC_OK);
```
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 252 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2UJYFk0b564cpiRA1K:-N2UJYFl2FaY3EsKPC81:b-evqsd5) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L252)):*
> ```Java
>                 cloudTasksUtils,
>                 response));
>     doThrow(new RuntimeException()).when(dnsWriter).commit();
> ```

You shouldn't expect a `RuntimeException` to be thrown here, for the reasons explained above.



_Code quote:_
```Java
doThrow(new RuntimeException()).when(dnsWriter).commit();
```
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 275 at r1](https://reviewable.io/reviews/google/nomulus/1628#-N2UJj1S1ULGy97M6uKg:-N2UJj1S1ULGy97M6uKh:bokt6pt) ([raw file](https://github.com/google/nomulus/blob/5f5566181b15c6cd306b8b6fe619b55682422f83/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L275)):*
> ```Java
>     ImmutableMultimap.Builder<String, String> mapBuilder = ImmutableMultimap.builder();
> 
>     verify(cloudTasksUtils)
> ```

`CloudTasksHelper` provides methods to check enqueued tasks.



_Code quote:_
```Java
verify(cloudTasksUtils)
```

CydeWeys (commented):
So the way I'd recommend doing this is:

Take advantage of Cloud Task's built-in retry count rather than implementing your own.  It's impossible to ensure that we'll always be able to count errors correctly if our action is being aborted uncleanly (e.g. if it infinite loops and times out), but the Cloud Tasks built-in retry count should be able to count those correctly.  Also, the built-in retry count is exposed in the Cloud Tasks pane in the GCP web console, and we want what we're using to be the same as what's viewable there for easy debugging.

So the overall logic should be:

```
runAction() {
  try {
    [ ... do DNS updates .. ]
  } catch (Throwable e) {
    if (retry count on task from task queue < MAX_RETRIES) {
      throw e;  // this will return a 500-level HTTP status code that will cause Cloud Tasks to retry
    } else {
      log exception at severe and say that we are permanently failing this task;
      // do not re-throw -- we are purposely swallowing this exception
  }
  return HTTP status code 200;  // this will NOT retry, either because it was successful or max retries was reached
}
```

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1628#-:-N2woxDSEqkxaknNo3V9:bqnjx5b)* status: all files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

CydeWeys (commented):
Also, as discussed, you can test this by running a domain create / update, which will enqueue a task in the task queue, and then you can edit said task to make its DNS information invalid, at which point it should always fail.  You may find pausing the queue to be helpful in order to get said edit through before the task itself executes.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1628#-:-N2wrrkt2HOA4wp_dnK5:bulg9gv)* status: all files reviewed, 8 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

PR #1629, opened 2022-05-16
Cleanup gpg-agent instances and home directories

mindhog:
The GpgSystemCommandException leaks home directories, but more importantly it
leaks gpg-agent instances.  This can cause problems with inotify limits, since
the agent seems to make use of inotify.  Do a proper cleanup in afterEach().


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1629)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

mindhog:
I suspect that gpgconf may not be available on some of the build systems.  It's not clear that there's anything we can do in this case, so I've chosen to simply log the failure and continue.

PR #1631, opened 2022-05-16
Remove Datastore-specific tests and classes

CydeWeys:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1631)


CydeWeys:
This was way too much to bite off at once and it's subsequently been tackled in smaller parts by other people.

PR #1634, opened 2022-05-17
Make Caffeine cache loading work in non-GAE thread

jianglai:
Caffeine cache loading happens asynchronously in a different thread than
the main app engine thread. When a VKey is created in that thread, it
tries to ascertain that the current thread is an GAE thread and fails.

Of all CacheLoader::load implementations, only the one in Registry
creates a VKey, so we wrap the call to VKey.create() in a method that
sets and unsets the thread local field used to indicate whether the
current thread is an GAE thread.

### DIFF
--- a/core/src/main/java/google/registry/model/AppEngineEnvironment.java
+++ b/core/src/main/java/google/registry/model/AppEngineEnvironment.java
@@ -116,4 +116,9 @@ public class AppEngineEnvironment {
               }
             });
   }
+
+  /** Returns true if the current thread is in an App Engine Environment.*/
+  public static boolean isInAppEngineEnvironment() {
+    return ApiProxy.getCurrentEnvironment() != null;
+  }
 }
--- a/core/src/main/java/google/registry/model/tld/Registry.java
+++ b/core/src/main/java/google/registry/model/tld/Registry.java
@@ -48,6 +48,7 @@ import com.googlecode.objectify.annotation.Id;
 import com.googlecode.objectify.annotation.Mapify;
 import com.googlecode.objectify.annotation.OnSave;
 import com.googlecode.objectify.annotation.Parent;
+import google.registry.model.AppEngineEnvironment;
 import google.registry.model.Buildable;
 import google.registry.model.CacheUtils;
 import google.registry.model.CreateAutoTimestamp;
@@ -119,6 +120,8 @@ public class Registry extends ImmutableObject
 
   /** The suffix that identifies roids as belonging to this specific tld, e.g. -HOW for .how. */
   String roidSuffix;
+
+  private static final AppEngineEnvironment environment = new AppEngineEnvironment();
 
   /** Default values for all the relevant TLD parameters. */
   public static final TldState DEFAULT_TLD_STATE = TldState.PREDELEGATION;
@@ -288,7 +291,15 @@ public class Registry extends ImmutableObject
               });
 
   public static VKey<Registry> createVKey(String tld) {
-    return VKey.create(Registry.class, tld, Key.create(getCrossTldKey(), Registry.class, tld));
+    VKey<Registry> vkey;
+    if (environment.isInAppEngineEnvironment()) {
+      environment.setEnvironmentForCurrentThread();
+    }
+    vkey = VKey.create(Registry.class, tld, Key.create(getCrossTldKey(), Registry.class, tld));
+    if (environment.isInAppEngineEnvironment()) {
+      environment.unsetEnvironmentForCurrentThread();
+    }
+    return vkey;
   }
 
   public static VKey<Registry> createVKey(Key<Registry> key) {


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/google/nomulus/1634#-N2IkY3zDd3pMf7U7XDp):*
There are places where VKey is implicitly created. E.g., in the ForeignKeyIndex cache.
Some resource classes embeds vkeys and the VKeyConverter will be called on them.



_Code quote:_
```
  Of all CacheLoader::load implementations, only the one in Registry
  creates a VKey, so we wrapp the call to VKey.create() in a method that
```

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 295 at r2](https://reviewable.io/reviews/google/nomulus/1634#-N2Ir0337Bx7vmNNw9s3:-N2Ir0345s0823RtLyus:b-aauor8) ([raw file](https://github.com/google/nomulus/blob/2fadc02b739e338c77c3b0fb556d4e806e44c033/core/src/main/java/google/registry/model/tld/Registry.java#L295)):*
> ```Java
>   public static VKey<Registry> createVKey(String tld) {
>     VKey<Registry> vkey;
>     if (!AppEngineEnvironment.isInAppEngineEnvironment()) {
> ```

Is this all even necessary?  Why not use `VKey.createSql()` and then we don't have to do any Datastore stuff at all?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 295 at r2](https://reviewable.io/reviews/google/nomulus/1634#-N2Ir0337Bx7vmNNw9s3:-N2IruCS9ujmGqNMJ0zO:bp331jk) ([raw file](https://github.com/google/nomulus/blob/2fadc02b739e338c77c3b0fb556d4e806e44c033/core/src/main/java/google/registry/model/tld/Registry.java#L295)):*

Is this all even necessary?  Why not use `VKey.createSql()` and then we don't have to do any Datastore stuff at all?
</blockquote></details>

I think this would require an audit of every time that we call `VKey::getOfyKey` (or the similar `maybeGetOfyKey` to make sure that we aren't ever depending on it. 

For instance, PollMessage calls it a few times but only in the Datastore/Objectify OnLoad code path, so that's not relevant fortunately

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/google/nomulus/1634#-N2IkY3zDd3pMf7U7XDp):*

EppResource cache is the last one.
</blockquote></details>

`ForeignKeyIndex` is only used by Ofy so we don't care about it anymore. 'EppResource' cache already takes keys.  I also I don't see any warnings about either of them in the log.
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 295 at r2](https://reviewable.io/reviews/google/nomulus/1634#-N2Ir0337Bx7vmNNw9s3:-N2Mg7bt7qlCY7QNTMOt:bilkwpd) ([raw file](https://github.com/google/nomulus/blob/2fadc02b739e338c77c3b0fb556d4e806e44c033/core/src/main/java/google/registry/model/tld/Registry.java#L295)):*

I think this would require an audit of every time that we call `VKey::getOfyKey` (or the similar `maybeGetOfyKey` to make sure that we aren't ever depending on it. 

For instance, PollMessage calls it a few times but only in the Datastore/Objectify OnLoad code path, so that's not relevant fortunately 
</blockquote></details>

In theory we should not be depending on Ofy key anymore. But like Gus said we need to do an audit to not break things, which in the end would be a waste of time because we will tear down Ofy related stuff anyway.

This PR is intended to be a quick fix to enable background caching, so I'm focused on doing the least intrusive thing to make it work.

weiminyu (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/google/nomulus/1634#-N2IkY3zDd3pMf7U7XDp):*

Search for "No API environment" AND "ForeignKeyIndex"

Here is one sample stack trace:


BoundedLocalCache.java:1221com.github.benmanes.caffeine.cache.BoundedLocalCache lambda$refreshIfNeeded$5: Exception thrown during refresh
java.util.concurrent.CompletionException: javax.persistence.PersistenceException: Error attempting to apply AttributeConverter
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
	at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1582)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
Caused by: javax.persistence.PersistenceException: Error attempting to apply AttributeConverter
	at org.hibernate.type.descriptor.converter.AttributeConverterSqlTypeDescriptorAdapter$2.doConversion(AttributeConverterSqlTypeDescriptorAdapter.java:148)
	at org.hibernate.type.descriptor.converter.AttributeConverterSqlTypeDescriptorAdapter$2.extract(AttributeConverterSqlTypeDescriptorAdapter.java:121)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:257)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:253)
	at org.hibernate.type.AbstractStandardBasicType.nullSafeGet(AbstractStandardBasicType.java:243)
	at org.hibernate.type.AbstractStandardBasicType.hydrate(AbstractStandardBasicType.java:329)
	at org.hibernate.persister.entity.AbstractEntityPersister.hydrate(AbstractEntityPersister.java:3214)
	at org.hibernate.loader.Loader.loadFromResultSet(Loader.java:1887)
	at org.hibernate.loader.Loader.hydrateEntityState(Loader.java:1811)
	at org.hibernate.loader.Loader.instanceNotYetLoaded(Loader.java:1784)
	at org.hibernate.loader.Loader.getRow(Loader.java:1624)
	at org.hibernate.loader.Loader.getRowFromResultSet(Loader.java:748)
	at org.hibernate.loader.Loader.getRowFromResultSet(Loader.java:710)
	at org.hibernate.loader.Loader.loadSingleRow(Loader.java:396)
	at org.hibernate.internal.ScrollableResultsImpl.prepareCurrentRow(ScrollableResultsImpl.java:197)
	at org.hibernate.internal.ScrollableResultsImpl.next(ScrollableResultsImpl.java:102)
	at org.hibernate.query.internal.ScrollableResultsIterator.hasNext(ScrollableResultsIterator.java:33)
	at java.util.Iterator.forEachRemaining(Iterator.java:115)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at org.hibernate.query.spi.StreamDecorator.collect(StreamDecorator.java:209)
	at google.registry.model.index.ForeignKeyIndex.lambda$loadIndexesFromStore$3(ForeignKeyIndex.java:244)
	at google.registry.persistence.transaction.JpaTransactionManagerImpl.lambda$transact$1(JpaTransactionManagerImpl.java:206)
	at google.registry.util.Retrier.callWithRetry(Retrier.java:153)
	at google.registry.util.Retrier.callWithRetry(Retrier.java:74)
	at google.registry.persistence.transaction.JpaTransactionManagerImpl.transact(JpaTransactionManagerImpl.java:195)
	at google.registry.persistence.transaction.JpaTransactionManagerImpl.transact(JpaTransactionManagerImpl.java:186)
	at google.registry.model.index.ForeignKeyIndex.loadIndexesFromStore(ForeignKeyIndex.java:235)
	at google.registry.model.index.ForeignKeyIndex.access$100(ForeignKeyIndex.java:71)
	at google.registry.model.index.ForeignKeyIndex$1.load(ForeignKeyIndex.java:265)
	at google.registry.model.index.ForeignKeyIndex$1.load(ForeignKeyIndex.java:259)
	at com.github.benmanes.caffeine.cache.CacheLoader.reload(CacheLoader.java:166)
	at com.github.benmanes.caffeine.cache.CacheLoader.lambda$asyncReload$2(CacheLoader.java:190)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1590)
	... 5 more
Caused by: java.lang.NullPointerException: No API environment is registered for this thread.
</blockquote></details>

It might be easier to change the VKeyConverter

jianglai (commented):
> Reviewed 5 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/google/nomulus/1634#-N2IkY3zDd3pMf7U7XDp):*

It might be easier to change the VKeyConverter
</blockquote></details>

Found a way to masquerade the background thread as GAE thread in the refresh method. PTAL.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/CacheUtils.java` line 97 at r4](https://reviewable.io/reviews/google/nomulus/1634#-N2NHyZe-pR4HFOVFDd8:-N2NHyZe-pR4HFOVFDd9:bqdi7cj) ([raw file](https://github.com/google/nomulus/blob/4d842f995dd20125ac6cd84c3af8c4895199f3f7/core/src/main/java/google/registry/model/CacheUtils.java#L97)):*
> ```Java
>       }
>       value = load(key);
>       if (!AppEngineEnvironment.isInAppEngineEnvironment()) {
> ```

I don't think this will ever trigger, right? Given that we set the environment on the current thread above...

jianglai:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)_ status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)
> 
> _[`core/src/main/java/google/registry/model/CacheUtils.java` line 97 at r4](https://reviewable.io/reviews/google/nomulus/1634#-N2NHyZe-pR4HFOVFDd8:-N2NHyZe-pR4HFOVFDd9:bqdi7cj) ([raw file](https://github.com/google/nomulus/blob/4d842f995dd20125ac6cd84c3af8c4895199f3f7/core/src/main/java/google/registry/model/CacheUtils.java#L97)):_
> 
> > ```java
> >       }
> >       value = load(key);
> >       if (!AppEngineEnvironment.isInAppEngineEnvironment()) {
> > ```
> 
> I don't think this will ever trigger, right? Given that we set the environment on the current thread above...

Oops, you are absolutely right. That's a stupid bug.

gbrodman (changes requested):
> Reviewed 4 of 6 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/CacheUtils.java` line 86 at r5](https://reviewable.io/reviews/google/nomulus/1634#-N2NLrAMDubXc4aXaBk7:-N2NLrAMDubXc4aXaBk8:bxmblt6) ([raw file](https://github.com/google/nomulus/blob/ac70c2e3295d590147ba74c530304ce26d0e6923/core/src/main/java/google/registry/model/CacheUtils.java#L86)):*
> ```Java
>    */
>   @DeleteAfterMigration
>   public abstract static class MasqueradedCacheLoader<K, V> implements CacheLoader<K, V> {
> ```

it might not be clear from the name what exactly this cache loader is doing -- thoughts on a name like `AppEngineEnvironmentCacheLoader`?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/CacheUtils.java` line 86 at r5](https://reviewable.io/reviews/google/nomulus/1634#-N2NLrAMDubXc4aXaBk7:-N2NW2ZY-aJXSK8hIrDU:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/ac70c2e3295d590147ba74c530304ce26d0e6923/core/src/main/java/google/registry/model/CacheUtils.java#L86)):*

it might not be clear from the name what exactly this cache loader is doing -- thoughts on a name like `AppEngineEnvironmentCacheLoader`? 
</blockquote></details>

Done

gbrodman (approved):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

core/src/main/java/google/registry/model/CacheUtils.java
```diff
@@ -73,4 +77,29 @@ public static Caffeine<Object, Object> newCacheBuilder(Duration expireAfterWrite
     }
     return caffeine;
   }
+
+  /**
+   * A {@link CacheLoader} that automatically masquerade the background thread where the refresh
+   * action runs in to be an GAE thread.
+   */
+  @DeleteAfterMigration
+  public abstract static class AppEngineEnvironmentCacheLoader<K, V> implements CacheLoader<K, V> {
```
CydeWeys, line 86: Will any of this be necessary once we get rid of Ofy?

Also, why is this only necessary for some caches, rather than all of them?

jianglai: None of these would be necessary, that's why this class is annotated with `@DeleteAfterMigration` (granted, this annotation is probably overloaded, it could mean after migration to SQL, which is already completed, or after migration off of Ofy).

I've actually changed all usages of `CacheLoader`, except for one used in testing only. This might already be an overkill as some of the `CacheLoader` usages do not require the creation of a `VKey`.

https://cs.opensource.google/search?q=cacheloader&sq=&ss=nomulus%2Fnomulus

CydeWeys: So, considering that either way we go is going to be reverted after the imminent deactivation of Datastore stuff, why not just go with PR #1638, which is simpler?

CydeWeys: Simpler and less risky.

jianglai: With this PR we do get the async refresh feature, which was the main reason to switch to caffeine. Plus this fix is not as intrusive as originally thought. Also we don't need a bug or a TODO to remind us to turn back on async refresh -- we will delete the annotated classes, and then we will have to rename anonymous `AppEngineEnvironmentCacheLoader` back to `CacheLoader`.

So it seems to be a better solution in the meantime.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1634#-:-N2O-T5m2FaVnMVdKM2M:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

PR #1635, opened 2022-05-18
Optimize RDAP entity event query

weiminyu:
For each EPP entity, directly load the latest HistoryEntry per event type
instead of loading all events through the HistoryEntryDao.

Although most entities have a small number of history entries, there are
a few entities with many entries, enough to cause OutOfMemory error.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1635)

### DIFF
--- a/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java
+++ b/core/src/main/java/google/registry/model/reporting/HistoryEntryDao.java
@@ -211,7 +211,7 @@ public class HistoryEntryDao {
         jpaTm().criteriaQuery(criteriaQuery).getResultList());
   }
 
-  private static Class<? extends HistoryEntry> getHistoryClassFromParent(
+  public static Class<? extends HistoryEntry> getHistoryClassFromParent(
       Class<? extends EppResource> parent) {
     if (!RESOURCE_TYPES_TO_HISTORY_TYPES.containsKey(parent)) {
       throw new IllegalArgumentException(
@@ -220,7 +220,7 @@ public class HistoryEntryDao {
     return RESOURCE_TYPES_TO_HISTORY_TYPES.get(parent);
   }
 
-  private static String getRepoIdFieldNameFromHistoryClass(
+  public static String getRepoIdFieldNameFromHistoryClass(
       Class<? extends HistoryEntry> historyClass) {
     if (!REPO_ID_FIELD_NAMES.containsKey(historyClass)) {
       throw new IllegalArgumentException(
--- a/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java
+++ b/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java
@@ -20,11 +20,14 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.ImmutableSetMultimap.toImmutableSetMultimap;
 import static google.registry.model.EppResourceUtils.isLinked;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.persistence.transaction.TransactionManagerUtil.transactIfJpaTm;
 import static google.registry.rdap.RdapIcannStandardInformation.CONTACT_REDACTED_VALUE;
 import static google.registry.util.CollectionUtils.union;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -81,6 +84,7 @@ import java.util.Set;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import javax.persistence.Entity;
 import org.joda.time.DateTime;
 
 /**
@@ -149,6 +153,19 @@ public class RdapJsonFormatter {
      */
     INTERNAL
   }
+
+  /**
+   * JPQL query template for finding the latest history entry per event type for an EPP entity.
+   *
+   * <p>User should replace '%entityName%', '%repoIdField%', and '%repoIdValue%' with valid values.
+   * A DomainHistory query may look like below: {@code select e from DomainHistory e where
+   * domainRepoId = '17-Q9JYB4C' and modificationTime in (select max(modificationTime) from
+   * DomainHistory where domainRepoId = '17-Q9JYB4C' and type is not null group by type)}
+   */
+  private static final String GET_LAST_HISTORY_BY_TYPE_JPQL_TEMPLATE =
+      "select e from %entityName% e where %repoIdField% = '%repoIdValue%' and modificationTime in "
+          + " (select max(modificationTime) from %entityName% where "
+          + " %repoIdField% = '%repoIdValue%' and type is not null group by type)";
 
   /** Map of EPP status values to the RDAP equivalents. */
   private static final ImmutableMap<StatusValue, RdapStatus> STATUS_TO_RDAP_STATUS_MAP =
@@ -855,17 +872,8 @@ public class RdapJsonFormatter {
     return rolesBuilder.build();
   }
 
-  /**
-   * Creates the list of optional events to list in domain, nameserver, or contact replies.
-   *
-   * <p>Only has entries for optional events that won't be shown in "SUMMARY" versions of these
-   * objects. These are either stated as optional in the RDAP Response Profile 15feb19, or not
-   * mentioned at all but thought to be useful anyway.
-   *
-   * <p>Any required event should be added elsewhere, preferably without using HistoryEntries (so
-   * that we don't need to load HistoryEntries for "summary" responses).
-   */
-  private ImmutableList<Event> makeOptionalEvents(EppResource resource) {
+  @VisibleForTesting
+  ImmutableMap<EventAction, HistoryEntry> getLastHistoryEntryByType(EppResource resource) {
     HashMap<EventAction, HistoryEntry> lastEntryOfType = Maps.newHashMap();
     // Events (such as transfer, but also create) can appear multiple times. We only want the last
     // time they appeared.
@@ -878,8 +886,26 @@ public class RdapJsonFormatter {
     // 2.3.2.3 An event of *eventAction* type *transfer*, with the last date and time that the
     // domain was transferred. The event of *eventAction* type *transfer* MUST be omitted if the
     // domain name has not been transferred since it was created.
-    Iterable<? extends HistoryEntry> historyEntries =
-        HistoryEntryDao.loadHistoryObjectsForResource(resource.createVKey());
+    Iterable<? extends HistoryEntry> historyEntries;
+    if (tm().isOfy()) {
+      historyEntries = HistoryEntryDao.loadHistoryObjectsForResource(resource.createVKey());
+    } else {
+      VKey<? extends EppResource> resourceVkey = resource.createVKey();
+      Class<? extends HistoryEntry> historyClass =
+          HistoryEntryDao.getHistoryClassFromParent(resourceVkey.getKind());
+      String entityName = historyClass.getAnnotation(Entity.class).name();
+      if (Strings.isNullOrEmpty(entityName)) {
+        entityName = historyClass.getSimpleName();
+      }
+      String repoIdFieldName = HistoryEntryDao.getRepoIdFieldNameFromHistoryClass(historyClass);
+      String jpql =
+          GET_LAST_HISTORY_BY_TYPE_JPQL_TEMPLATE
+              .replace("%entityName%", entityName)
+              .replace("%repoIdField%", repoIdFieldName)
+              .replace("%repoIdValue%", resourceVkey.getSqlKey().toString());
+      historyEntries =
+          jpaTm().transact(() -> jpaTm().getEntityManager().createQuery(jpql).getResultList());
+    }
     for (HistoryEntry historyEntry : historyEntries) {
       EventAction rdapEventAction =
           HISTORY_ENTRY_TYPE_TO_RDAP_EVENT_ACTION_MAP.get(historyEntry.getType());
@@ -889,6 +915,21 @@ public class RdapJsonFormatter {
       }
       lastEntryOfType.put(rdapEventAction, historyEntry);
     }
+    return ImmutableMap.copyOf(lastEntryOfType);
+  }
+
+  /**
+   * Creates the list of optional events to list in domain, nameserver, or contact replies.
+   *
+   * <p>Only has entries for optional events that won't be shown in "SUMMARY" versions of these
+   * objects. These are either stated as optional in the RDAP Response Profile 15feb19, or not
+   * mentioned at all but thought to be useful anyway.
+   *
+   * <p>Any required event should be added elsewhere, preferably without using HistoryEntries (so
+   * that we don't need to load HistoryEntries for "summary" responses).
+   */
+  private ImmutableList<Event> makeOptionalEvents(EppResource resource) {
+    ImmutableMap<EventAction, HistoryEntry> lastEntryOfType = getLastHistoryEntryByType(resource);
     ImmutableList.Builder<Event> eventsBuilder = new ImmutableList.Builder<>();
     DateTime creationTime = resource.getCreationTime();
     DateTime lastChangeTime =
--- a/core/src/test/java/google/registry/rdap/RdapJsonFormatterTest.java
+++ b/core/src/test/java/google/registry/rdap/RdapJsonFormatterTest.java
@@ -15,6 +15,7 @@
 package google.registry.rdap;
 
 import static com.google.common.truth.Truth.assertThat;
+import static google.registry.rdap.RdapDataStructures.EventAction.TRANSFER;
 import static google.registry.rdap.RdapTestHelper.assertThat;
 import static google.registry.testing.DatabaseHelper.createTld;
 import static google.registry.testing.DatabaseHelper.persistResource;
@@ -28,7 +29,9 @@ import static google.registry.testing.TestDataHelper.loadFile;
 import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Maps;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import google.registry.model.contact.ContactResource;
@@ -51,6 +54,7 @@ import google.registry.testing.DualDatabaseTest;
 import google.registry.testing.FakeClock;
 import google.registry.testing.InjectExtension;
 import google.registry.testing.TestOfyAndSql;
+import google.registry.testing.TestSqlOnly;
 import org.joda.time.DateTime;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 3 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @mindhog, and @weiminyu)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 165 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2Na1gEAb36ERBLO__X:-N2Na1gEAb36ERBLO__Y:b-ji3uig) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L165)):*
> ```Java
>    * DomainHistory where domainRepoId = '17-Q9JYB4C' and type is not null group by type)}
>    */
>   private static final String GET_LAST_HISTORY_BY_TYPE_JPQL_TEMPLATE =
> ```

I'm not 100% sure how this would work in HQL, but in raw Postgres you can do 

`SELECT DISTINCT ON (history_type) * FROM "%entityName%" WHERE %repoIdField = '%repoIdValue' ORDER BY history_type, history_modification_time DESC;`

and it gets the same results without having nested statements 
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 890 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2NZRtMEuS-oV5Sib9w:-N2NZRtNB6n3OzPZgTJ7:b5zjptg) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L890)):*
> ```Java
>     // domain name has not been transferred since it was created.
>     Iterable<? extends HistoryEntry> historyEntries;
>     if (tm().isOfy()) {
> ```

We don't need the `isOfy` check anymore, right? 
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 897 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_EnkDJT8tUg48oIj:-N2N_EnkDJT8tUg48oIk:b-tnxu48) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L897)):*
> ```Java
>           HistoryEntryDao.getHistoryClassFromParent(resourceVkey.getKind());
>       String entityName = historyClass.getAnnotation(Entity.class).name();
>       if (Strings.isNullOrEmpty(entityName)) {
> ```

I don't think this is the case for any of our history classes right?
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 902 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_Xv0CoRgE31Ta5g4:-N2N_Xv0CoRgE31Ta5g5:b-n8e8s5) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L902)):*
> ```Java
>       String repoIdFieldName = HistoryEntryDao.getRepoIdFieldNameFromHistoryClass(historyClass);
>       String jpql =
>           GET_LAST_HISTORY_BY_TYPE_JPQL_TEMPLATE
> ```

thoughts on using `String::format`?

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 165 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2Na1gEAb36ERBLO__X:-N2NgVdWE5QfWarFsKQR:b-i5b4kl) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L165)):*

I'm not 100% sure how this would work in HQL, but in raw Postgres you can do 

`SELECT DISTINCT ON (history_type) * FROM "%entityName%" WHERE %repoIdField = '%repoIdValue' ORDER BY history_type, history_modification_time DESC;`

and it gets the same results without having nested statements 
</blockquote></details>

select distinct **ON (column)** is postgresql-specific, so unlikely to be supported by JPQL.
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 890 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2NZRtMEuS-oV5Sib9w:-N2NczXb0rJBK7EVZDVx:b-43az7n) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L890)):*

We don't need the `isOfy` check anymore, right? 
</blockquote></details>

Until we remove TestOfyAndSql tags from all Rdap tests, we need it.
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 897 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_EnkDJT8tUg48oIj:-N2NdPxYAEIGTwLfP_HK:bjui0jh) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L897)):*

I don't think this is the case for any of our history classes right?
</blockquote></details>

It is. All of them are annotated with a plain Entity annotation without name.

I can call getSimpleName directly, but feel that calling getAnnotation first is a good habit.
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 902 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_Xv0CoRgE31Ta5g4:-N2Nep0E08W6oRWx_tO4:bc5r6t) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L902)):*

thoughts on using `String::format`?
</blockquote></details>

I think this makes the template more readable. Even if we use argument indexes, the template itself can be confusing unless it is put next to the place when the actual query is generated.

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman, @mindhog, and @weiminyu)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 165 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2Na1gEAb36ERBLO__X:-N2SCz4V4J6YlsXBCZtD:b-5raa8t) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L165)):*

select distinct **ON (column)** is postgresql-specific, so unlikely to be supported by JPQL.
</blockquote></details>

If the `SELECT DISTINCT ON (column)` is more performant, then we should probably just use that even though it's PostgreSQL-specific.  (Is it more performant?)

We're already using SQL arrays, so we're already locked into PostgreSQL anyway.
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 897 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_EnkDJT8tUg48oIj:-N2SDJipBqDBHIGB9MRN:blmjlkx) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L897)):*

It is. All of them are annotated with a plain Entity annotation without name.

I can call getSimpleName directly, but feel that calling getAnnotation first is a good habit.


</blockquote></details>

Is there anything in `TypeUtils` that is useful here?  `getClassFromString()` perhaps?
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 902 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_Xv0CoRgE31Ta5g4:-N2SDQrF2rHnVIhBN7fQ:bdyfmue) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L902)):*

I think this makes the template more readable. Even if we use argument indexes, the template itself can be confusing unless it is put next to the place when the actual query is generated.
</blockquote></details>

Any reason this replacement isn't happening using built-in SQL 
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 907 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2SD_X21-ul6S-XxL_P:-N2SD_X21-ul6S-XxL_Q:b-v5ipzi) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L907)):*
> ```Java
>               .replace("%repoIdValue%", resourceVkey.getSqlKey().toString());
>       historyEntries =
>           jpaTm().transact(() -> jpaTm().getEntityManager().createQuery(jpql).getResultList());
> ```

Why not use `setParameter()` here instead of String munging from above?  Does it not work on column names?  `repoIdValue` at least is not a column name -- it's actual data.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, @mindhog, and @weiminyu)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 165 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2Na1gEAb36ERBLO__X:-N2SJa-fCdpDw4S6oX4Z:b-gtgk7w) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L165)):*

If the `SELECT DISTINCT ON (column)` is more performant, then we should probably just use that even though it's PostgreSQL-specific.  (Is it more performant?)

We're already using SQL arrays, so we're already locked into PostgreSQL anyway.
</blockquote></details>

I didn't actually know which one was superior before but I just did a test using separate repo IDs on prod (if I don't use separate repo IDs, whichever query runs last is faster due to caching):

`explain analyze select distinct on (history_type) * FROM "DomainHistory" WHERE domain_repo_id = '339578985-DEV' ORDER BY history_type, history_modification_time DESC;` Planning Time: 0.177 ms
 Execution Time: 0.364 ms
(8 rows)


`explain analyze select e from "DomainHistory" e where domain_repo_id = '436E9C8CE-DEV' and history_modification_time in (select max(history_modification_time) from "DomainHistory" where domain_repo_id = '436E9C8CE-DEV' and history_type is not null group by history_type);`
 Planning Time: 0.240 ms
 Execution Time: 8.925 ms
(16 rows)

Interestingly, it seems like the second one (with the select(max)) has a lot more variance when testing it with various domains. The highest execution time I saw with a few different choices was 35ms, and the lowest was about 1ms 

I'm not sure if this is a meaningful difference. There is a tangible benefit to writing HQL (what already exists here), but there's also probably a benefit in a simpler shorter SQL statement. 
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 897 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_EnkDJT8tUg48oIj:-N2NhlNH8PkqrrKZBK-L:b-rzudru) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L897)):*

It is. All of them are annotated with a plain Entity annotation without name.

I can call getSimpleName directly, but feel that calling getAnnotation first is a good habit.


</blockquote></details>

what I meant was that our classes don't have a name in the Entity annotation, so this doesn't seem necessary

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, @mindhog, and @weiminyu)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 890 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2NZRtMEuS-oV5Sib9w:-N2SKxMD3Z5FppqIb0Vj:b-ygm5bs) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L890)):*

Until we remove TestOfyAndSql tags from all Rdap tests, we need it.
</blockquote></details>

boo :/

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, @mindhog, and @weiminyu)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 911 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2SduDI480As9NF4SoR:-N2SduDI480As9NF4SoS:blcg8gp) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L911)):*
> ```Java
>     for (HistoryEntry historyEntry : historyEntries) {
>       EventAction rdapEventAction =
>           HISTORY_ENTRY_TYPE_TO_RDAP_EVENT_ACTION_MAP.get(historyEntry.getType());
> ```

Looking at the contents of this map, it seems like we can make our query above smarter, by only querying for the relevant types from the DB, rather than querying for all distinct ones and ignroing them.  In particular, `HISTORY_ENTRY_TYPE_TO_RDAP_EVENT_ACTION_MAP` does not even contain UPDATEs, which are the types causing us all these problems in the first place!

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman, @mindhog, and @weiminyu)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 165 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2Na1gEAb36ERBLO__X:-N2Sg3OeEWnTt2ROGrvJ:b-kwmtry) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L165)):*

I didn't actually know which one was superior before but I just did a test using separate repo IDs on prod (if I don't use separate repo IDs, whichever query runs last is faster due to caching):

`explain analyze select distinct on (history_type) * FROM "DomainHistory" WHERE domain_repo_id = '339578985-DEV' ORDER BY history_type, history_modification_time DESC;` Planning Time: 0.177 ms
 Execution Time: 0.364 ms
(8 rows)


`explain analyze select e from "DomainHistory" e where domain_repo_id = '436E9C8CE-DEV' and history_modification_time in (select max(history_modification_time) from "DomainHistory" where domain_repo_id = '436E9C8CE-DEV' and history_type is not null group by history_type);`
 Planning Time: 0.240 ms
 Execution Time: 8.925 ms
(16 rows)

Interestingly, it seems like the second one (with the select(max)) has a lot more variance when testing it with various domains. The highest execution time I saw with a few different choices was 35ms, and the lowest was about 1ms 

I'm not sure if this is a meaningful difference. There is a tangible benefit to writing HQL (what already exists here), but there's also probably a benefit in a simpler shorter SQL statement. 

</blockquote></details>

In my measurements comparing domains with similar number of history entries, the sub-select query is usually faster. E.g., 334745225-DEV (select distint) took 170ms while 34040F2D0-DEV (sub select) took 130ms, on cold starts. When cache is loaded, it's 50ms vs 40ms, still in subselect's favor. Note that explain analyze does not execute the query. The Execution Time reported is the analysis cost.

Performance is a smaller matter. Only the sub-select query can be executed in JPQL. If we use select distinct with native query, we will have to 
cast string to Enum and DateTime.  


PS: Here is someone else's report of select distinct on being a bit slower: https://stackoverflow.com/questions/54810250/postgresql-latest-row-in-distinct-on-less-performant-than-max-row-in-group-by

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 902 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_Xv0CoRgE31Ta5g4:-N2SkBaD6KO7pl9sVoD8:bv65dx7) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L902)):*

Any reason this replacement isn't happening using built-in SQL 
</blockquote></details>

If only parameter **values** are undetermined, we can use Query.setParameter to set them.
Here the table and column names are variables. We have to replace them before creation the query.
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 907 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2SD_X21-ul6S-XxL_P:-N2SkcA30FfqU1vq5Nys:bjw7gtn) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L907)):*

Why not use `setParameter()` here instead of String munging from above?  Does it not work on column names?  `repoIdValue` at least is not a column name -- it's actual data.
</blockquote></details>

Entityname and repoIdFieldName cannot be set with setParameter(). It's an extra mental cost to repoIdValue separately.
___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 911 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2SduDI480As9NF4SoR:-N2SlBcd4L5AY8ChblP-:b-daw641) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L911)):*

Looking at the contents of this map, it seems like we can make our query above smarter, by only querying for the relevant types from the DB, rather than querying for all distinct ones and ignroing them.  In particular, `HISTORY_ENTRY_TYPE_TO_RDAP_EVENT_ACTION_MAP` does not even contain UPDATEs, which are the types causing us all these problems in the first place!
</blockquote></details>

The query as it stands  takes < 10s  for the million-row repoId on cold start, and 1s if cache is loaded. For the others it is at millisecond level.
We don't need to spend too much time on it right now.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @mindhog)

___
*[`core/src/main/java/google/registry/rdap/RdapJsonFormatter.java` line 897 at r1](https://reviewable.io/reviews/google/nomulus/1635#-N2N_EnkDJT8tUg48oIj:-N2SmNdI0i0745tiExVn:b-wezghw) ([raw file](https://github.com/google/nomulus/blob/e8412bb8c9d5941bf6385f962cfdc746747d1aba/core/src/main/java/google/registry/rdap/RdapJsonFormatter.java#L897)):*

what I meant was that our classes don't have a name in the Entity annotation, so this doesn't seem necessary 
</blockquote></details>

There is a getEntityName() utility method in some test utils. We should move it to production and replace the few places entity name is needed. That can be left for later.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635#-:-N2T8XMM3rmHi-DQdODl:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @mindhog)

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1635)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1637, opened 2022-05-18
Use new renew cost calculation in handleFeeRequest()

rachelguan:
The purpose of this PR is to change the renewal cost calculation in `handleFeeRequest()` in `DomainFlowUtil` and fix existing test cases in check and info flow test after adding the new renew price calculation to `DomainFlowUtil`. 

- Missing billing events in existing test cases of `DomainInfoFlowTest.Java`:  two of the existing test cases for renewal have `eventBillingAndHistory` in `doSuccessTest()` as false (which is the default setting). However, if the domain is persisted, corresponding billing event should be there as well. I added a helper method that creates sample `DomainHistory` and `Recurring`  for an existing domain and the changes were made in those two test cases. 

- Missing billing events in existing test cases of `DomainCheckFlowTest.Java`: similar to the issue in `DomainInfoFlowTest` where `DomainHistory` and `Recurring` billing event objects are missing. A helper method was added that could be used to persist billing event for an existing domain. Because of this change, one of the test cases failed because of mismatching xml response file. The content remains the same except the tags are now added in the first `<epp>` tag.  See the diff in `...s/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml`. This file currently has only been used once in all the test cases. 
Moreover, the tags include v11 and v12 and the expiration date ends with .001 second instead of .002.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1637)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -31,6 +31,7 @@ import static google.registry.flows.domain.DomainFlowUtils.validateDomainNameWit
 import static google.registry.flows.domain.DomainFlowUtils.verifyNotInPredelegation;
 import static google.registry.model.tld.Registry.TldState.START_DATE_SUNRISE;
 import static google.registry.model.tld.label.ReservationType.getTypeOfHighestSeverity;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -267,16 +268,21 @@ public final class DomainCheckFlow implements Flow {
     for (FeeCheckCommandExtensionItem feeCheckItem : feeCheck.getItems()) {
       for (String domainName : getDomainNamesToCheckForFee(feeCheckItem, domainNames.keySet())) {
         FeeCheckResponseExtensionItem.Builder<?> builder = feeCheckItem.createResponseBuilder();
+        Optional<DomainBase> domainBase =
+            Optional.ofNullable((DomainBase) domainObjs.get(domainName));
         handleFeeRequest(
             feeCheckItem,
             builder,
             domainNames.get(domainName),
-            Optional.ofNullable((DomainBase) domainObjs.get(domainName)),
+            domainBase,
             feeCheck.getCurrency(),
             now,
             pricingLogic,
             allocationToken,
-            availableDomains.contains(domainName));
+            availableDomains.contains(domainName),
+            domainBase.isPresent()
+                ? tm().transact(() -> tm().loadByKey(domainBase.get().getAutorenewBillingEvent()))
+                : null);
         responseItems.add(builder.setDomainNameIfSupported(domainName).build());
       }
     }
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -680,7 +680,19 @@ public class DomainFlowUtils {
         break;
       case RENEW:
         builder.setAvailIfSupported(true);
-        fees = pricingLogic.getRenewPrice(registry, domainNameString, now, years, null).getFees();
+        if (domain.isPresent()) {
+          fees =
+              pricingLogic
+                  .getRenewPrice(
+                      registry,
+                      domainNameString,
+                      now,
+                      years,
+                      tm().transact(() -> tm().loadByKey(domain.get().getAutorenewBillingEvent())))
+                  .getFees();
+        } else {
+          fees = pricingLogic.getRenewPrice(registry, domainNameString, now, years, null).getFees();
+        }
         break;
       case RESTORE:
         // The minimum allowable period per the EPP spec is 1, so, strangely, 1 year still has to be
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -633,7 +633,8 @@ public class DomainFlowUtils {
       DateTime currentDate,
       DomainPricingLogic pricingLogic,
       Optional<AllocationToken> allocationToken,
-      boolean isAvailable)
+      boolean isAvailable,
+      @Nullable Recurring recurringBillingEvent)
       throws EppException {
     DateTime now = currentDate;
     // Use the custom effective date specified in the fee check request, if there is one.
@@ -680,19 +681,10 @@ public class DomainFlowUtils {
         break;
       case RENEW:
         builder.setAvailIfSupported(true);
-        if (domain.isPresent()) {
-          fees =
-              pricingLogic
-                  .getRenewPrice(
-                      registry,
-                      domainNameString,
-                      now,
-                      years,
-                      tm().transact(() -> tm().loadByKey(domain.get().getAutorenewBillingEvent())))
-                  .getFees();
-        } else {
-          fees = pricingLogic.getRenewPrice(registry, domainNameString, now, years, null).getFees();
-        }
+        fees =
+            pricingLogic
+                .getRenewPrice(registry, domainNameString, now, years, recurringBillingEvent)
+                .getFees();
         break;
       case RESTORE:
         // The minimum allowable period per the EPP spec is 1, so, strangely, 1 year still has to be
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -682,9 +682,7 @@ public class DomainFlowUtils {
       case RENEW:
         builder.setAvailIfSupported(true);
         fees =
-            pricingLogic
-                .getRenewPrice(registry, domainNameString, now, years, recurringBillingEvent)
-                .getFees();
+            pricingLogic.getRenewPrice(registry, domainNameString, now, years, recurringBillingEvent).getFees();
         break;
       case RESTORE:
         // The minimum allowable period per the EPP spec is 1, so, strangely, 1 year still has to be
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -682,7 +682,9 @@ public class DomainFlowUtils {
       case RENEW:
         builder.setAvailIfSupported(true);
         fees =
-            pricingLogic.getRenewPrice(registry, domainNameString, now, years, recurringBillingEvent).getFees();
+            pricingLogic
+                .getRenewPrice(registry, domainNameString, now, years, recurringBillingEvent)
+                .getFees();
         break;
       case RESTORE:
         // The minimum allowable period per the EPP spec is 1, so, strangely, 1 year still has to be
--- a/core/src/main/java/google/registry/flows/domain/DomainInfoFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainInfoFlow.java
@@ -169,7 +169,8 @@ public final class DomainInfoFlow implements Flow {
           now,
           pricingLogic,
           Optional.empty(),
-          false);
+          false,
+          tm().transact(() -> tm().loadByKey(domain.getAutorenewBillingEvent())));
       extensions.add(builder.build());
     }
     return extensions.build();
--- a/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java
+++ b/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java
@@ -867,17 +867,16 @@ class DomainCheckFlowTest extends ResourceCheckFlowTestCase<DomainCheckFlow, Dom
   @TestOfyAndSql
   void testFeeExtension_premium_eap_v06_withRenewalOnRestore() throws Exception {
     createTld("example");
-    DomainBase domainBase = persistActiveDomain("rich.example");
-    setUpBillingEventForExistingDomain(domainBase);
-    setEppInput("domain_check_fee_premium_v06.xml");
-    clock.setTo(DateTime.parse("2010-01-01T10:00:00Z"));
     persistResource(
-        loadByEntity(domainBase)
+        persistActiveDomain("rich.example")
             .asBuilder()
             .setDeletionTime(clock.nowUtc().plusDays(25))
             .setRegistrationExpirationTime(clock.nowUtc().minusDays(1))
             .setStatusValues(ImmutableSet.of(StatusValue.PENDING_DELETE))
             .build());
+    persistPendingDeleteDomain("rich.example");
+    setEppInput("domain_check_fee_premium_v06.xml");
+    clock.setTo(DateTime.parse("2010-01-01T10:00:00Z"));
     persistResource(
         Registry.get("example")
             .asBuilder()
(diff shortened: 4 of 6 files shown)


rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1637)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 683 at r1](https://reviewable.io/reviews/google/nomulus/1637#-N3Pf3EoDctagLv8i3Nh:-N3VvtyDAzGsSodoKkkP:bdqhpk6) ([raw file](https://github.com/google/nomulus/blob/e3acc8486ca8155dcc459cfa42d6ed243bd5171a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L683)):*

And I just realized, in the case of `DomainCheckFlow`, this is inefficient as it will load every single recurring billing event sequentially (up to 50 in a row on a max-size domain check), substantially reducing the speed of the overall flow.  You're going to need to change the method signature here to add the Recurring as a parameter, and then load them up in bulk in the check flow and pass the correct one to each `handleFeeRequest()` individually.

The key thing to notice here is that `handleFeeRequest()` was not previously doing any database interactions, and by adding them, you are adding potential sources of slow-down.  This method should continue to not have DB interactions.
</blockquote></details>

right now, I'm passing the event from `Flow` instead. 
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1637#-N3PgoRD3fSnaghu3uUd:-N3VwJ9C0DFlu6-bkweK:b-hwo8ji) ([raw file](https://github.com/google/nomulus/blob/e3acc8486ca8155dcc459cfa42d6ed243bd5171a/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L97)):*

This file seems to be missing tests validating that a `NONPREMIUM` or `SPECIFIED` renewal price is handled correctly as seen in a check.
</blockquote></details>

It actually doesn't seem necessary to add test cases in `Info` since the domains are not supposed to be all existing domains. `handleFeeRequest()` is used for existing domains for checking restore fees. more details: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java;l=265?q=handleFeeRequest&ss=nomulus%2Fnomulus

```
 ImmutableMap<String, EppResource> domainObjs =
        loadDomainsForRestoreChecks(feeCheck, domainNames, existingDomains);

    for (FeeCheckCommandExtensionItem feeCheckItem : feeCheck.getItems()) {
      for (String domainName : getDomainNamesToCheckForFee(feeCheckItem, domainNames.keySet())) {
        FeeCheckResponseExtensionItem.Builder<?> builder = feeCheckItem.createResponseBuilder();
        Optional<DomainBase> domainBase =
            Optional.ofNullable((DomainBase) domainObjs.get(domainName));
        handleFeeRequest(
            feeCheckItem,
            builder,
            domainNames.get(domainName),
            domainBase,
            feeCheck.getCurrency(),
            now,
            pricingLogic,
            allocationToken,
            availableDomains.contains(domainName),
            domainBase.isPresent()
                ? tm().transact(() -> tm().loadByKey(domainBase.get().getAutorenewBillingEvent()))
                : null);
        responseItems.add(builder.setDomainNameIfSupported(domainName).build());
      }


additional test cases would be needed for `NONPREMIUM` and `SPECIFIED` for restoring a domain and there's no additional code for restore flow supporting anchor tenants and internal registrations. 
```

___
*[`core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java` line 84 at r1](https://reviewable.io/reviews/google/nomulus/1637#-N3PhdMN8CYLqJh5LK3e:-N3RGjfU-xtYUfC1Nw-q:b-t9490b) ([raw file](https://github.com/google/nomulus/blob/e3acc8486ca8155dcc459cfa42d6ed243bd5171a/core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java#L84)):*

This file seems to be missing tests validating that a `NONPREMIUM` or `SPECIFIED` renewal price is handled correctly on a domain info flow (for the 1 version of the fee extension that works with the info flow anyway).
</blockquote></details>

added

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1637)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1637#-N3PgoRD3fSnaghu3uUd:-N3ykNMN43uHJ61_XDpt:b-xnx9e9) ([raw file](https://github.com/google/nomulus/blob/e3acc8486ca8155dcc459cfa42d6ed243bd5171a/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L97)):*

It actually doesn't seem necessary to add test cases in `Info` since the domains are not supposed to be all existing domains. `handleFeeRequest()` is used for existing domains for checking restore fees. more details: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java;l=265?q=handleFeeRequest&ss=nomulus%2Fnomulus

```
 ImmutableMap<String, EppResource> domainObjs =
        loadDomainsForRestoreChecks(feeCheck, domainNames, existingDomains);

    for (FeeCheckCommandExtensionItem feeCheckItem : feeCheck.getItems()) {
      for (String domainName : getDomainNamesToCheckForFee(feeCheckItem, domainNames.keySet())) {
        FeeCheckResponseExtensionItem.Builder<?> builder = feeCheckItem.createResponseBuilder();
        Optional<DomainBase> domainBase =
            Optional.ofNullable((DomainBase) domainObjs.get(domainName));
        handleFeeRequest(
            feeCheckItem,
            builder,
            domainNames.get(domainName),
            domainBase,
            feeCheck.getCurrency(),
            now,
            pricingLogic,
            allocationToken,
            availableDomains.contains(domainName),
            domainBase.isPresent()
                ? tm().transact(() -> tm().loadByKey(domainBase.get().getAutorenewBillingEvent()))
                : null);
        responseItems.add(builder.setDomainNameIfSupported(domainName).build());
      }


additional test cases would be needed for `NONPREMIUM` and `SPECIFIED` for restoring a domain and there's no additional code for restore flow supporting anchor tenants and internal registrations. 
```

</blockquote></details>

Are you sure?  See here:  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/flows/domain/DomainInfoFlowTest.java;l=711-728?q=domain_info_fee_response.xml&ss=nomulus%2Fnomulus

The primary use case for the `<domain:info>` command is definitely to use it on domains that exist, and if you then also pass the right version of the fee extension and ask for the renewal price, it should return the correct one.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1637)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @jianglai and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 284 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ykf494n8vzPCZ5mV1:-N3ykf494n8vzPCZ5mV2:b-twn04n) ([raw file](https://github.com/google/nomulus/blob/aee62d95e016ce299cb4cf9cf556698baf0f693c/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L284)):*
> ```Java
>             availableDomains.contains(domainName),
>             domainBase.isPresent()
>                 ? tm().transact(() -> tm().loadByKey(domainBase.get().getAutorenewBillingEvent()))
> ```

This is still performing individual loads in separate transactions per my other comment.

You need to do a single batch load of all the relevant recurrences before the `for` loop.  It needs to be O(1) loads, not O(n) loads.
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 879 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ymLb_9VzQdKmzaNJb:-N3ymLb_9VzQdKmzaNJc:b-ii4cqx) ([raw file](https://github.com/google/nomulus/blob/aee62d95e016ce299cb4cf9cf556698baf0f693c/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L879)):*
> ```Java
>     persistPendingDeleteDomain("rich.example");
>     setEppInput("domain_check_fee_premium_v06.xml");
>     clock.setTo(DateTime.parse("2010-01-01T10:00:00Z"));
> ```

I think the reason this test is messed up now is that you're now setting the time after persisting the resources.  You need to set the time first.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 10 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ylNIe6Mkrvbpc_VqI:-N3ylNIe6Mkrvbpc_VqJ:bfi1q4m) ([raw file](https://github.com/google/nomulus/blob/2377774bf9d6cd2e80253733fc76720ba023f096/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L10)):*
> ```XML
>         <domain:cd>
>           <domain:name avail="false">rich.example</domain:name>
>           <domain:reason>In use</domain:reason>
> ```

Why is this changing?  I'm looking at the test that's using it, and in said test, the domain is supposed to be in pending delete (meaning it's not yet available for re-registration).

On a broader note, you can't just blindly change test files to match actual output without thinking in depth about whether the changes actually make sense (here, they do not).  You made an error in your test above, and instead of fixing said error which was caught here by the expected test data, you just modified the test data to accommodate, so now the test isn't even testing what it's supposed to be testing.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 47 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ylG8Y4g3tOulK82zT:-N3ylG8_23dvTD5BUnRz:bc7plbm) ([raw file](https://github.com/google/nomulus/blob/2377774bf9d6cd2e80253733fc76720ba023f096/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L47)):*
> ```XML
>           <fee:command>restore</fee:command>
>           <fee:period unit="y">1</fee:period>
>           <fee:fee description="renew">100.00</fee:fee>
> ```

This removal seems quite wrong to me.  The file name has `eap` in it, and $100 is the EAP fee.  Why is the EAP fee being removed in your change?

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1637)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 284 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ykf494n8vzPCZ5mV1:-N3zSBI19__abwQia21V:bic40sa) ([raw file](https://github.com/google/nomulus/blob/aee62d95e016ce299cb4cf9cf556698baf0f693c/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L284)):*

This is still performing individual loads in separate transactions per my other comment.

You need to do a single batch load of all the relevant recurrences before the `for` loop.  It needs to be O(1) loads, not O(n) loads.
</blockquote></details>

I actually don't understand why it should be done before the `for` loop since the recurring billing event comes from the domain base? 
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1637#-N3PgoRD3fSnaghu3uUd:-N3zSovfEz0gq7EWP9D_:b-qjqe8s) ([raw file](https://github.com/google/nomulus/blob/e3acc8486ca8155dcc459cfa42d6ed243bd5171a/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L97)):*
> em necessary to add test cases in Info since the domains are not supposed to be all existing do
that was a typo..i meant in the check flow. I  added cases in the check flow test

rachelguan (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1637)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 879 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ymLb_9VzQdKmzaNJb:-N3zUaYxAqQ_vLnQNYqe:bbfsk9f) ([raw file](https://github.com/google/nomulus/blob/aee62d95e016ce299cb4cf9cf556698baf0f693c/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L879)):*

I think the reason this test is messed up now is that you're now setting the time after persisting the resources.  You need to set the time first.
</blockquote></details>

I actually think DomainCheckFlowTest changes might have to wait for restore/transfer flow changes to come in
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 10 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ylNIe6Mkrvbpc_VqI:-N3zUmKl3njozgxJGm9X:b-12x6ke) ([raw file](https://github.com/google/nomulus/blob/2377774bf9d6cd2e80253733fc76720ba023f096/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L10)):*

Why is this changing?  I'm looking at the test that's using it, and in said test, the domain is supposed to be in pending delete (meaning it's not yet available for re-registration).

On a broader note, you can't just blindly change test files to match actual output without thinking in depth about whether the changes actually make sense (here, they do not).  You made an error in your test above, and instead of fixing said error which was caught here by the expected test data, you just modified the test data to accommodate, so now the test isn't even testing what it's supposed to be testing.
</blockquote></details>

understood and I didn't mean to change the test case like that..I've been investigating why the tags had to be done this way. 
___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml` line 47 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ylG8Y4g3tOulK82zT:-N3zUswSCwWPyXvWBvAc:b-m2tz8k) ([raw file](https://github.com/google/nomulus/blob/2377774bf9d6cd2e80253733fc76720ba023f096/core/src/test/resources/google/registry/flows/domain/domain_check_fee_premium_eap_response_v06_with_renewal.xml#L47)):*

This removal seems quite wrong to me.  The file name has `eap` in it, and $100 is the EAP fee.  Why is the EAP fee being removed in your change?
</blockquote></details>

I believe ..restore and transfer flow code need to be modified in order for this to work normally

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1637)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @jianglai, and @rachelguan)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 284 at r3](https://reviewable.io/reviews/google/nomulus/1637#-N3ykf494n8vzPCZ5mV1:-N5v3RnS3xCc1Lqv4-46:bujc5e3) ([raw file](https://github.com/google/nomulus/blob/aee62d95e016ce299cb4cf9cf556698baf0f693c/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L284)):*

I actually don't understand why it should be done before the `for` loop since the recurring billing event comes from the domain base? 
</blockquote></details>

Because that way you're only doing 1 DB transaction instead of N DB transactions.  The domain bases are already loaded; they're not being loaded individually inside the for loop, so they are already present to do a single batch load of the recurrings ahead of time too rather than one at a time.  See PR ##1694 for the fix.

CydeWeys:
This has been merged, with modifications, as #1694.

PR #1638, opened 2022-05-18
Temporarily disable Caffeine refreshAfterWrite()

CydeWeys:
We will re-enable this by reverting this commit in the near future (a few weeks
from now), but for the present, we need to disable the async refresh behavior on
the cache because unmanaged threading does not play along well with App Engine,
resulting in Objectify errors when it can't get the GAE environment during Ofy
Key construction as part of making the VKeys for the entities being
cached. Right now this code is sadly just throwing lots of warnings into the
logs without actually successfully refreshing anything.

As soon as we are no longer using Objectify this will no longer be an issue, and
we can revert this commit and then go back to refreshing asynchronously in
background.

### DIFF
--- a/core/src/main/java/google/registry/model/CacheUtils.java
+++ b/core/src/main/java/google/registry/model/CacheUtils.java
@@ -62,15 +62,6 @@ public class CacheUtils {
    * cost of the load will never be incurred during the read.
    */
   public static Caffeine<Object, Object> newCacheBuilder(Duration expireAfterWrite) {
-    Duration refreshAfterWrite = expireAfterWrite.dividedBy(2);
-    Caffeine<Object, Object> caffeine = Caffeine.newBuilder().expireAfterWrite(expireAfterWrite);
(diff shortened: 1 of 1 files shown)


CydeWeys:
http://b/233092977 for the TODO to revert this once Ofy is gone.

jianglai:
See https://github.com/google/nomulus/pull/1634, there does appear to be an easy way to keep the background refresh running.

ben-manes:
FYI, you could use a caller-runs policy instead and behave like Guava does. We default to async since there is now a platform threadpool. You have two options,

1. `Caffeine.executor(Runnable::run)` will delegate all work (maintenance, listeners, etc) to the caller's thread.
2. Return a completed future in `CacheLoader.asyncReload`. In 3.1.0 this will inspect and return the new value to the caller, though not yet backported to v2 (which still assumes a fire-and-forget mode). As of 3.1, it then mirror's Guava's behavior where the default refresh is caller-runs but can be overridden to return an incomplete future.

Hope that helps
-Ben

PR #1642, opened 2022-05-23
Slightly improve performance of ExpandRecurringBillingEventsAction

CydeWeys:
We don't need to log every single no-op batch of 50 Recurrences that are
processed (considering we have 1.5M total in our system), and we also don't need
to process Recurrences that already ended prior to the Cursor time (this gets us
down to 420k from 1.5M).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1642)

### DIFF
--- a/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
+++ b/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java
@@ -67,6 +67,7 @@ import google.registry.util.Clock;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import org.joda.money.Money;
 import org.joda.time.DateTime;
@@ -155,10 +156,12 @@ public class ExpandRecurringBillingEventsAction implements Runnable {
                                     + "WHERE eventTime <= :executeTime "
                                     + "AND eventTime < recurrenceEndTime "
                                     + "AND id > :maxProcessedRecurrenceId "
+                                    + "AND recurrenceEndTime > :cursorTime "
                                     + "ORDER BY id ASC",
                                 Recurring.class)
                             .setParameter("executeTime", executeTime)
                             .setParameter("maxProcessedRecurrenceId", prevMaxProcessedRecurrenceId)
+                            .setParameter("cursorTime", cursorTime)
                             .setMaxResults(batchSize)
                             .getResultList();
                     for (Recurring recurring : recurrings) {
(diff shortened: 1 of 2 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1642)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 210 at r1](https://reviewable.io/reviews/google/nomulus/1642#-N2nXZny64OQbnaEkAr9:-N2nXZny64OQbnaEkArA:b86i64q) ([raw file](https://github.com/google/nomulus/blob/0ef60fa8c7afc4af04361f61c2fbb305b540cda4/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L210)):*
> ```Java
>         // then only log one no-op every so often as a good balance between letting the user track
>         // that the action is still running while also not spamming the logs incessantly.
> ```

logger.atInfo().atMostEvery(1, TimeUnit.Minute(3)).log(...) behave more or less the same?



_Code quote:_
```Java
logger.atInfo().log(
```

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1642)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java` line 210 at r1](https://reviewable.io/reviews/google/nomulus/1642#-N2nXZny64OQbnaEkAr9:-N2rxM1k6FPGIVnn0jqp:b-poeylv) ([raw file](https://github.com/google/nomulus/blob/0ef60fa8c7afc4af04361f61c2fbb305b540cda4/core/src/main/java/google/registry/batch/ExpandRecurringBillingEventsAction.java#L210)):*

logger.atInfo().atMostEvery(1, TimeUnit.Minute(3)).log(...) behave more or less the same?
</blockquote></details>

Done. It's not exactly the same semantics (as now it'll be outputting potentially immediately after a non-noop batch), but that's probably fine.  And the trade-off is that the implementation is simpler.

CydeWeys (commented):
PTAL

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1642#-:-N2rxUpf1Rd8_QBLCT_S:bb71pro)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1642)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1644, opened 2022-05-24
Disable Ofy tests.

mindhog:
This change just turns off the Ofy tests at the root, by removing processing
for dual tests and disassociating the TestOfyOnly annotation from test
annotations.

This is far less comprehensive than #1631, but it's probably worth entering as
a stopgap solution just because it should speed up our test runs and unblock a
lot of other cleanup work.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1644)

### DIFF
--- a/core/src/test/java/google/registry/testing/DualDatabaseTestInvocationContextProvider.java
+++ b/core/src/test/java/google/registry/testing/DualDatabaseTestInvocationContextProvider.java
@@ -1,3 +1,4 @@
+
 // Copyright 2020 The Nomulus Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


CydeWeys (commented):
This doesn't do much though, right?  It doesn't affect `@TestOfyAndSql` annotations, which are the vast majority of all ofy usages?

(Or is that fine, this is just low hanging fruit?)

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1644#-:-N2rO0tC-Nw0FtMocCMo:bfvlzrs)* status: 0 of 2 files reviewed, all discussions resolved

mindhog (commented):
I believe it completely disables Ofy tests within `@DualDatabaseTest`s, both those defined by `@TestOfyOnly` and by `@TestOfyAndSql` (the latter by simply removing the Ofy context from the list of contexts the test is executed in).

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1644#-:-N2rOzIv4L0yjCMwiwKW:br3ovpu)* status: 0 of 2 files reviewed, all discussions resolved

CydeWeys (approved):
LGTM so long as it actually does what it says on the tin (i.e. the total number of tests executed now is actually substantially less).

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1644#-:-N2rndZT1sXLSL8nzYi_:baghy4f)* status: 0 of 3 files reviewed, all discussions resolved

PR #1646, opened 2022-05-25
Invoicing pipeline hack for April billing fix

sarahcaseybot:
**DO NOT SUBMIT**

This modifies the invoicing pipeline to create a new April invoice so that it only invoices for Auto renews that occurred after 2:59:59 on April 19th, 2022. See b/233666109 for more details.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1646)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -768,7 +768,7 @@ createUberJar(
 // alpha use the pipelines staged on alpha deployment project.
 //
 // User should install gcloud and login to GCP before invoking this tasks.
-if (environment == 'alpha') {
+if (true) {
   def pipelines = [
       initSql            :
           [
--- a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java
+++ b/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java
@@ -96,7 +96,6 @@ public class InvoicingPipeline implements Serializable {
         options.getDatabase().equals("DATASTORE")
             ? readFromBigQuery(options, pipeline)
             : readFromCloudSql(options, pipeline);
-
     saveInvoiceCsv(billingEvents, options);
 
     saveDetailedCsv(billingEvents, options);
@@ -189,10 +188,11 @@ public class InvoicingPipeline implements Serializable {
             TextIO.write()
                 .to(
                     String.format(
-                        "%s/%s/%s/%s-%s",
+                        "%s/%s/%s-%s/%s-%s",
                         options.getBillingBucketUrl(),
                         BillingModule.INVOICES_DIRECTORY,
                         options.getYearMonth(),
+                        "Missing-Auto-Renews",
                         options.getInvoiceFilePrefix(),
                         options.getYearMonth()))
                 .withHeader(InvoiceGroupingKey.invoiceHeader())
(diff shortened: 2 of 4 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1646)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/resources/google/registry/beam/invoicing/sql/april_fix.sql` line 22 at r1](https://reviewable.io/reviews/google/nomulus/1646#-N2wwePf5AZmqBdPR6sU:-N2wwePf5AZmqBdPR6sV:bwpoa2) ([raw file](https://github.com/google/nomulus/blob/d58cdff94f3721a168a62fe744bdf8650304c3eb/core/src/main/resources/google/registry/beam/invoicing/sql/april_fix.sql#L22)):*
> ```TSQL
> LEFT JOIN BillingCancellation cr ON b.cancellationMatchingBillingEvent = cr.refRecurring.billingId
> WHERE r.billingAccountMap IS NOT NULL
> AND b.flags IS NOT NULL
> ```

What about where billing event type is `RENEW`?
___
*[`core/src/main/resources/google/registry/beam/invoicing/sql/april_fix.sql` line 25 at r1](https://reviewable.io/reviews/google/nomulus/1646#-N2wwoSwBL69tx-XJ1Ah:-N2wwoSx66rAdG4Lii6S:b1hol1e) ([raw file](https://github.com/google/nomulus/blob/d58cdff94f3721a168a62fe744bdf8650304c3eb/core/src/main/resources/google/registry/beam/invoicing/sql/april_fix.sql#L25)):*
> ```TSQL
> AND r.type = 'REAL'
> AND t.invoicingEnabled IS TRUE
> AND b.billingTime BETWEEN CAST('2022-04-19 03:00:00' AS timestamp) AND CAST('%LAST_TIMESTAMP_OF_MONTH%' AS timestamp)
> ```

There's no timezone on this timestamp -- are you sure it's doing the correct thing?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1646)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/resources/google/registry/beam/invoicing/sql/april_fix.sql` line 22 at r1](https://reviewable.io/reviews/google/nomulus/1646#-N2wwePf5AZmqBdPR6sU:-N2wx1geBC8ocTQs8hU1:b-ygx9jd) ([raw file](https://github.com/google/nomulus/blob/d58cdff94f3721a168a62fe744bdf8650304c3eb/core/src/main/resources/google/registry/beam/invoicing/sql/april_fix.sql#L22)):*

What about where billing event type is `RENEW`?
</blockquote></details>

Or is flags only not null for autorenews anyway?

sarahcaseybot (commented):
Deployed the pipeline as it is in the PR today and it correctly recreated the summary invoice. I loaded the summary invoice into a spreadsheet and verified that the totals are correct. However, the hacked pipeline overwrote the /2022-04 bucket even though it renamed it to /2022-04-Missing-Auto-Renews. The modified pipeline also failed to created the detailed invoices. What we know from this is that the modified query is correct, but I still need to make a few more adjustments to the actual pipeline to make the correct invoice files.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1646#-:-N31142q9rzc96RRrz6m:bh3190p)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

PR #1650, opened 2022-05-31
Add a new recurrenceLastExpansion column to the BillingRecurrence table

CydeWeys:
This will be used to determine which recurrences are in scope for
expansion. Anything that has already been expanded within the past year is
automatically out of scope.

Note that the default value is set to just over a year ago, so that, initially,
everything is in scope for expansion, and then will gradually be winnowed down
over time so that most recurrences at any given point are out of scope. Newly
created recurrings (after the subsequent code change goes in) will have their
last expansion time set to the same as the event time of when the recurring is
written, such that they'll first be considered in-scope precisely one year
later.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1650)


CydeWeys:
See https://b/233070128

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1650)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1651, opened 2022-05-31
Add recurrenceLastExpansion billing event logic code

CydeWeys:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1651)


CydeWeys:
I haven't gotten to the tests yet, so it's not ready for LGTM, but please look at the approach in this code and let me know if you can think of any issues with it.

And yes, as currently written, it's going to take a full year until all existing recurrings are updated so that they don't all count as in scope, as the existing recurring is only updated when a new one time billing event is being expanded.  I might change so that so that the recurrenceLastExpansion is updated (to replace for the default value of 2021-06-01 in the DB), even if no one time billing event is yet being written.  E.g., if the recurrenceLastExpansion has its default value of 2021-06-01, but this runs and determines that the last one-time billing event that was expanded is from 2022-03-04, then the recurrenceLastExpansion should be updated to 2022-03-04 as well.  I guess I'll add in that logic too (but even without it, it's still doing the correct thing, albeit with more work for the first year).

CydeWeys (commented):
Getting back to the tests this week.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1651#-:-N56-geQ-wySW_B37ccy:b-dh4sst)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @jianglai and @weiminyu)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1651)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
What is the status on this PR? I see Weimin approved it two months ago and there has been no movement since. Is it abandoned? Should I also take a look at it?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1651#-:-NC0PIyZ3vkngr_BoTix:b68faiv)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @weiminyu)

CydeWeys (commented):
The status is that it's not quite finished yet on my end.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1651#-:-NC0Pz3jA56AExZWFp4t:b-bey2jb)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1653, opened 2022-06-02
Remove Ofy code from various flow-related classes

gbrodman:
This included removing ofy-specific code from various tests. Also, some
of the other tests (e.g. RdapDomainActionTest) had to be configured to
use only SQL -- otherwise, as it currently stands, they were trying to
use ofy.

We also delete the CreateSyntheticHistoryEntriesAction and pipeline
because they're no longer relevant, and impossible to test (the goal of
the actions were to create objects in ofy which doesn't happen any
more).


mindhog:
Ok if I take the review on this one?

mindhog (changes requested):
> Reviewed 21 of 30 files at r1, 3 of 3 files at r2, 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1653)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/rdap/RdapDomainActionTest.java` line 252 at r3](https://reviewable.io/reviews/google/nomulus/1653#-N3y4cuV2JCxb8dkZmsN:-N3y4cuV2JCxb8dkZmsO:b4l75hr) ([raw file](https://github.com/google/nomulus/blob/b11741fd71ad43e648e84d85c27775bf6caaddac/core/src/test/java/google/registry/rdap/RdapDomainActionTest.java#L252)):*
> ```Java
>   }
> 
>   @TestOfyAndSql
> ```

Why not TestSqlOnly?  (It doesn't matter much, we're going to have to replace all of these anyway)
___
*[`core/src/test/java/google/registry/webdriver/TestServerExtension.java` line 212 at r3](https://reviewable.io/reviews/google/nomulus/1653#-N3y5VYBC_IhaSjF3mVI:-N3y5VYC9kSDsZOYOwuw:b-h9s69n) ([raw file](https://github.com/google/nomulus/blob/b11741fd71ad43e648e84d85c27775bf6caaddac/core/src/test/java/google/registry/webdriver/TestServerExtension.java#L212)):*
> ```Java
>     void runInner() throws Exception {
>       // Clear the SQL database and set it as primary (we have to do this out of band because the
>       // AppEngineException can't natively do it for us yet
> ```

"Extension?"



_Code quote:_
```Java
Exception
```

PR #1656, opened 2022-06-03
Use getRenewPrice() in DomainPricingLogic in transfer flow

rachelguan:
The new implementation of getRenewPrice() should be used in getTransferPrice() in DomainPricingLogic in order to support the anchor tenants and internal registrations. Therefore, an additional parameter recurringBillingEvent was added as @nullable for clients with existing billing events (mainly for anchor tenants and internal registrations).

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1656)


CydeWeys:
This will be submitted in PR #1701.

PR #1660, opened 2022-06-07
Remove ofy scaffolding from DomainContent and core classes

mindhog:
Remove the Ofy field restoration code from a number of core classes, including DomainContent, GracePeriod, BillingEvent, PollEvent and TransferData.

Warning: this is a very invasive PR which removes a lof the basic mechanisms for restoring data.  While I believe it to be correct, thar be dragons here.  Please review with caution and rollback at the first sign of trouble.

Among other things, this changes the `PollEvent.setParent*()` methods to `setHistoryEntry()`, `set*RepoId()` and `set*HistoryRevisionId()`.  This seemed to be better naming, as the specific repo id and revision id methods were desirable in some circumstances, the use of `setHistoryEntry()` seemed more consistent with that  and the notion of a parent is more typically associated with homogenous data types (even though a history record has historically been modeled as a "parent" to its associated data).  We may wish to consider dropping the "parent" nomenclature for other cases.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1660)


mindhog (commented):
Replaced PollMessage repo ids and history ids with *HistoryId setters, per discussion.  PTAL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660#-:-N5FfZch4NWRmgbzAA-N:b-mhpmy5)* status: 0 of 66 files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 1 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660)* status: 1 of 66 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @mindhog)

___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 197 at r2](https://reviewable.io/reviews/google/nomulus/1660#-N5GayA65eF3_1M5nmGP:-N5GayA7AvX5luYu3HaE:bglj464) ([raw file](https://github.com/google/nomulus/blob/0a56f5d5aea08e1ef102aaa3119e610616d6ba6a/core/src/main/java/google/registry/model/poll/PollMessage.java#L197)):*
> ```Java
>   public String getDomainRepoId() {
>     checkArgument(getType() == Type.DOMAIN);
>     return domainRepoId;
> ```

I was thinking that we'd store the actual `DomainHistoryId` object in the PollMessage class, rather than continuing to have the split `long domainHistoryRevisionId` and `String domainRepoId`, thoughts?

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660)* status: 1 of 66 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 197 at r2](https://reviewable.io/reviews/google/nomulus/1660#-N5GayA65eF3_1M5nmGP:-N5Gl9j29FK1L0JgBnEQ:b4k88mw) ([raw file](https://github.com/google/nomulus/blob/0a56f5d5aea08e1ef102aaa3119e610616d6ba6a/core/src/main/java/google/registry/model/poll/PollMessage.java#L197)):*

I was thinking that we'd store the actual `DomainHistoryId` object in the PollMessage class, rather than continuing to have the split `long domainHistoryRevisionId` and `String domainRepoId`, thoughts?  
</blockquote></details>

Agree in principle, but I held off on this because it's a little more involved (the History classes themselves don't even do this, and maybe never will) and this PR is already pretty scary.  Added a todo and a bug, though.

gbrodman (changes requested):
> Reviewed 43 of 63 files at r1, 11 of 13 files at r2, 11 of 11 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660)* status: all files reviewed, 12 unresolved discussions (waiting on @mindhog)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 887 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5b9XIDBvJMKRoJhNM6:-N5b9XIEEHTcroX_VRR2:b-3jun4z) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/domain/DomainContent.java#L887)):*
> ```Java
>     }
> 
>     public B setAutorenewPollMessage(
> ```

Are there any situations where the caller of this method doesn't have the full `Autorenew` object? We could just pass that in instead of having two parameters here
___
*[`core/src/main/java/google/registry/model/domain/DomainHistory.java` line 331 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bHZWA1zbO6qy9YbAH:-N5bHZWA1zbO6qy9YbAI:b-1pksvb) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/domain/DomainHistory.java#L331)):*
> ```Java
>      * Returns the domain repository id.
>      *
>      * <p>This method is private because it is only used by Hibernate.
> ```

probably want to remove this part of each of the comments heh 
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 197 at r2](https://reviewable.io/reviews/google/nomulus/1660#-N5GayA65eF3_1M5nmGP:-N5aN-EV3waIEGgi6usx:b-967hq) ([raw file](https://github.com/google/nomulus/blob/0a56f5d5aea08e1ef102aaa3119e610616d6ba6a/core/src/main/java/google/registry/model/poll/PollMessage.java#L197)):*

Agree in principle, but I held off on this because it's a little more involved (the History classes themselves don't even do this, and maybe never will) and this PR is already pretty scary.  Added a todo and a bug, though.
</blockquote></details>

eh all right yeah 
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 118 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aMvhq-CARavHpuUcK:-N5aMvhq-CARavHpuUcL:b-z4tu7l) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L118)):*
> ```Java
>     }
> 
>     public long getId() {
> ```

Should document what this means (that it's part of the external key) 
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 467 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aQ69u6bFaKg9LAs-4:-N5aQ69u6bFaKg9LAs-5:bpa48mx) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L467)):*
> ```Java
>     void postLoad() {
>       if (pendingActionNotificationResponse != null) {
>         // Promote the pending action notification response to its specialized type.
> ```

we should really refactor PendingActionNotificationResponse, or at least the subclass constructors 
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 560 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aRWw630TUX71JdPzv:-N5aRWw630TUX71JdPzw:buc7bng) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L560)):*
> ```Java
>         } else if (instance.pendingActionNotificationResponse
>             instanceof HostPendingActionNotificationResponse) {
>           instance.hostId = instance.pendingActionNotificationResponse.nameOrId.value;
> ```

wait, "hostId"? This, according to the HostPending..., is the fully qualified host name. This seems odd right?
___
*[`core/src/main/java/google/registry/model/transfer/DomainTransferData.java` line 113 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bHF1z7o6j0OtI8aqX:-N5bHF1z7o6j0OtI8aqY:bsnxnxo) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L113)):*
> ```Java
> 
>   /**
>    * Autorenew history, which we need to preserve because it's often used in contexts where we don't
> ```

nit, typo, remove don't 
___
*[`core/src/main/java/google/registry/model/transfer/TransferData.java` line 185 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aocZj5SMBISS5lau0:-N5aocZj5SMBISS5lau1:b-v2oaba) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/TransferData.java#L185)):*
> ```Java
>     }
> 
>     public B setServerApproveEntities(
> ```

kinda seems weird to be setting three fields in a single method called `setServerApproveEntiities` right? 
___
*[`core/src/main/java/google/registry/model/transfer/TransferData.java` line 197 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aUZ4u9nuAfGO5pxzC:-N5aUZ4u9nuAfGO5pxzD:b-gxzbwx) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/TransferData.java#L197)):*
> ```Java
>     @Override
>     public T build() {
> ```

nit, extra newline I think unless the UI is messing up
___
*[`core/src/test/java/google/registry/model/domain/DomainBaseTest.java` line 246 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bEnaV5V09NRKPLLTh:-N5bEnaV5V09NRKPLLTi:beg3xbu) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/test/java/google/registry/model/domain/DomainBaseTest.java#L246)):*
> ```Java
>   }
> 
>   @TestSqlOnly
> ```

I wouldn't expect these to require the dual-database tests given that those don't really do anything nowadays, do you know why these are added? I'd expect the AppEngineExtension to be sufficient. 
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 271 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bH-hu605HbTerGAbs:-N5bH-hu605HbTerGAbt:bz7sexk) ([raw file](https://github.com/google/nomulus/blob/1719d066cf22a2db18798092c3baca5ab993bdef/db/src/main/resources/sql/schema/db-schema.sql.generated#L271)):*
> ```
>         statuses text[],
>         admin_contact text,
>         auth_info_repo_id text,
> ```

What about a Flyway file? 
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 445 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bG62LAYjeVYysQwd9:-N5bG62LAYjeVYysQwdA:bbyukug) ([raw file](https://github.com/google/nomulus/blob/1719d066cf22a2db18798092c3baca5ab993bdef/db/src/main/resources/sql/schema/db-schema.sql.generated#L445)):*
> ```
>     create table "GracePeriodHistory" (
>        grace_period_history_revision_id int8 not null,
>         billing_event_id int8,
> ```

Just want to double-check that all these removals are safe. The principle is that the int8 id should be sufficient to identify a row in a table without the history or repo ID?
___
*[`util/src/main/java/google/registry/util/NullSafeCollectionBuilder.java` line 25 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aLB0z54Lmc1dIu_xE:-N5aLB1-9Ypcks3daHvK:bg9j38z) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/util/src/main/java/google/registry/util/NullSafeCollectionBuilder.java#L25)):*
> ```Java
>  * <p>Useful because we do this in a number of cases.
>  */
> public class NullSafeCollectionBuilder<E, B extends ImmutableCollection.Builder<E>> {
> ```

we should probably call this something like `NullIgnoringCollectionBuilder`, `NullDroppingCollectionBuilder`, or `NullSwallowingCollectionBuilder` since i think "safe" implies that they're not just dropped

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660)* status: all files reviewed, 10 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 887 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5b9XIDBvJMKRoJhNM6:-N5jSlfO8_soNQz53kuM:b-iincfm) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/domain/DomainContent.java#L887)):*

Are there any situations where the caller of this method doesn't have the full `Autorenew` object? We could just pass that in instead of having two parameters here
</blockquote></details>

Yes, when we [copy DomainBase](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/domain/DomainBase.java#L193)

We could certainly add a second method for it?
___
*[`core/src/main/java/google/registry/model/domain/DomainHistory.java` line 331 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bHZWA1zbO6qy9YbAH:-N5jTieX54uwnFh8pIBU:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/domain/DomainHistory.java#L331)):*

probably want to remove this part of each of the comments heh 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 118 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aMvhq-CARavHpuUcK:-N5jVYcm5NyCAZ6AfYDZ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L118)):*

Should document what this means (that it's part of the external key) 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 467 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aQ69u6bFaKg9LAs-4:-N5jssJ3DUakhtjnDoUZ:b-399gdm) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L467)):*

we should really refactor PendingActionNotificationResponse, or at least the subclass constructors 
</blockquote></details>

Yeah.  If you have thoughts on how it should look, let's write up a bug for it.
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 560 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aRWw630TUX71JdPzv:-N5jrtqDBPvtwOiHnyIp:b2nrv5y) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L560)):*

wait, "hostId"? This, according to the HostPending..., is the fully qualified host name. This seems odd right?
</blockquote></details>

Yes.  This is actually how it worked in the original code, I didn't name it :-)

It  would be easy enough for me to rename this here, if you like?  It's not publicly visible.
___
*[`core/src/main/java/google/registry/model/transfer/DomainTransferData.java` line 113 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bHF1z7o6j0OtI8aqX:-N5jq0AZBSJtpUUtWoNC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/DomainTransferData.java#L113)):*

nit, typo, remove don't 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/transfer/TransferData.java` line 185 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aocZj5SMBISS5lau0:-N5jmVNv5lJmo4bnC94k:bps31ak) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/TransferData.java#L185)):*

kinda seems weird to be setting three fields in a single method called `setServerApproveEntiities` right? 
</blockquote></details>

Yes, it does.  Originally, I had three setters.  However, repo id and history id _must be set_ in lockstep with the server approve entities and the best way to enforce that (as with most things) is at compile time.

This was more straightforward when `serverApproveEntities` was a field.  I don't think there's a common naming convention for a method that sets a batch of related attributes (which is what this does, even without `repoId` and `historyId`).  It arguably flies against the conventions of the Builder pattern, but the Builder pattern seems kind of deficient to me so I don't mind shaking it up a bit :-) 
___
*[`core/src/main/java/google/registry/model/transfer/TransferData.java` line 197 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aUZ4u9nuAfGO5pxzC:-N5jmPkB1p24gUJtqps6:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/TransferData.java#L197)):*

nit, extra newline I think unless the UI is messing up
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/domain/DomainBaseTest.java` line 246 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bEnaV5V09NRKPLLTh:-N5kSkmq4KCCnz8uHnNa:b-fol8ck) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/test/java/google/registry/model/domain/DomainBaseTest.java#L246)):*

I wouldn't expect these to require the dual-database tests given that those don't really do anything nowadays, do you know why these are added? I'd expect the AppEngineExtension to be sufficient. 
</blockquote></details>

Force of habit.    Done.

(Important safety note: these need to be `withDatastoreAndSql()` because some of the test infrastructure requires the construction of ofy keys and thus ofy Entity registration.  Killed an hour or two on that one...)
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 271 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bH-hu605HbTerGAbs:-N5jVyOHEU3isfTjS9CY:bgy87uh) ([raw file](https://github.com/google/nomulus/blob/1719d066cf22a2db18798092c3baca5ab993bdef/db/src/main/resources/sql/schema/db-schema.sql.generated#L271)):*

What about a Flyway file? 
</blockquote></details>

We can't change the deployed schema until the uses have been removed.
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 445 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bG62LAYjeVYysQwd9:-N5jWICY5Rn7quon7udL:b-iovg72) ([raw file](https://github.com/google/nomulus/blob/1719d066cf22a2db18798092c3baca5ab993bdef/db/src/main/resources/sql/schema/db-schema.sql.generated#L445)):*

Just want to double-check that all these removals are safe. The principle is that the int8 id should be sufficient to identify a row in a table without the history or repo ID?
</blockquote></details>

Yes.  BillingEvent just uses its id as its primary key, we only use the other field to reconstruct ofy keys.
___
*[`util/src/main/java/google/registry/util/NullSafeCollectionBuilder.java` line 25 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aLB0z54Lmc1dIu_xE:-N5jivE84tXDhGX_cO1h:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/util/src/main/java/google/registry/util/NullSafeCollectionBuilder.java#L25)):*

we should probably call this something like `NullIgnoringCollectionBuilder`, `NullDroppingCollectionBuilder`, or `NullSwallowingCollectionBuilder` since i think "safe" implies that they're not just dropped
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 17 of 17 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660)* status: all files reviewed, 3 unresolved discussions (waiting on @mindhog)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 887 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5b9XIDBvJMKRoJhNM6:-N5kfUoQ4kN4b14Ka2JC:b-fq77f9) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/domain/DomainContent.java#L887)):*

Yes, when we [copy DomainBase](https://github.com/google/nomulus/blob/master/core/src/main/java/google/registry/model/domain/DomainBase.java#L193)

We could certainly add a second method for it?
</blockquote></details>

eh, yeah this is probably fine then no worries 
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 467 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aQ69u6bFaKg9LAs-4:-N5kfxk7DP9kgoL9J_mf:b-7rl0jq) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L467)):*

Yeah.  If you have thoughts on how it should look, let's write up a bug for it.
</blockquote></details>

done!
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 560 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aRWw630TUX71JdPzv:-N5kg-3j5tYTreH5JuQ0:bin4795) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L560)):*

Yes.  This is actually how it worked in the original code, I didn't name it :-)

It  would be easy enough for me to rename this here, if you like?  It's not publicly visible.
</blockquote></details>

ugh yeah not worth doing here then, unfortunately 
___
*[`core/src/main/java/google/registry/model/transfer/TransferData.java` line 185 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aocZj5SMBISS5lau0:-N5khVkE8BW6nX2o6GLy:b-wa2dst) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/transfer/TransferData.java#L185)):*

Yes, it does.  Originally, I had three setters.  However, repo id and history id _must be set_ in lockstep with the server approve entities and the best way to enforce that (as with most things) is at compile time.

This was more straightforward when `serverApproveEntities` was a field.  I don't think there's a common naming convention for a method that sets a batch of related attributes (which is what this does, even without `repoId` and `historyId`).  It arguably flies against the conventions of the Builder pattern, but the Builder pattern seems kind of deficient to me so I don't mind shaking it up a bit :-) 
</blockquote></details>

I mean, yeah maybe the name could probably be changed somehow but I see what you mean. Fair enough then 
___
*[`core/src/test/java/google/registry/model/domain/DomainBaseTest.java` line 246 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bEnaV5V09NRKPLLTh:-N5kgHK3BRwTF6cOmpfI:b-lz2p7c) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/test/java/google/registry/model/domain/DomainBaseTest.java#L246)):*

Force of habit.    Done.

(Important safety note: these need to be `withDatastoreAndSql()` because some of the test infrastructure requires the construction of ofy keys and thus ofy Entity registration.  Killed an hour or two on that one...)
</blockquote></details>

oof yeah -- my other PR should help with that as we just do entity registration when the "withCloudSql" flag is set (as eventually we'll remove entity registration under the hood)  
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 271 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5bH-hu605HbTerGAbs:-N5kh-rgBbfS5tMa1_0X:b74ngpt) ([raw file](https://github.com/google/nomulus/blob/1719d066cf22a2db18798092c3baca5ab993bdef/db/src/main/resources/sql/schema/db-schema.sql.generated#L271)):*

We can't change the deployed schema until the uses have been removed.
</blockquote></details>

yeah fair point, filed https://b.corp.google.com/issues/237551857 so that I don't forget (you'll probably have left by then I assume)

mindhog (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1660)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @mindhog)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 887 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5b9XIDBvJMKRoJhNM6:-N5kiqcu67gybYYzu-rA:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/domain/DomainContent.java#L887)):*

eh, yeah this is probably fine then no worries 
</blockquote></details>

Ack.
___
*[`core/src/main/java/google/registry/model/poll/PollMessage.java` line 560 at r4](https://reviewable.io/reviews/google/nomulus/1660#-N5aRWw630TUX71JdPzv:-N5kj3fIDsq9r7d6z4C-:bb6sh2i) ([raw file](https://github.com/google/nomulus/blob/9e457b591dd866f9e732798b21b24b5fe2612180/core/src/main/java/google/registry/model/poll/PollMessage.java#L560)):*

ugh yeah not worth doing here then, unfortunately 
</blockquote></details>

Ack.

PR #1663, opened 2022-06-09
Add logging for nameserver deletion

jianglai:
The main purpose of this PR is to help debug b/234189023, where a
registrar reported that in sandbox they observed seemingly successful EPP
update responses to delete NS records, which are not actually deleted after
the commands executed.

To actually load the persisted domain resource after an update would
require us to execute another transaction immediately after the update
transaction and that can only be achieved outside the flow (i. e. in
FlowRunner or EppController) and we need to test for the type of flows
before logging, which seems unnecessarily complex.

For now we are just adding logs inside the update transaction itself to
validate that:

1. The NS records to delete are as expected.
2. The Current NS records are as expected.
3. The new NS records to persist are as expected.

The EPP success reply is the default reply when no errors are thrown in
a transaction. If we see a success reply (which means that the
transaction finished successfully) and expected logs from the transaction, the
only explanation could be that somewhere in the ORM layer the java
representation of what the entity is is different from what is being
presented to the database. I think that signals a much bigger and
fundamental problem, which is quite unlikely given how isolated the
issue under consideration is.

In any case we would like to add the logging functionality in sandbox and ask
the registrar to report again when they see similar issues.

Also made some typo and linting fixes.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1663)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
@@ -48,7 +48,9 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
+import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
+import google.registry.config.RegistryEnvironment;
 import google.registry.dns.DnsQueue;
 import google.registry.flows.EppException;
 import google.registry.flows.ExtensionManager;
@@ -146,6 +148,9 @@ public final class DomainUpdateFlow implements TransactionalFlow {
   private static final ImmutableSet<StatusValue> UPDATE_DISALLOWED_STATUSES =
       ImmutableSet.of(StatusValue.PENDING_DELETE, StatusValue.SERVER_UPDATE_PROHIBITED);
 
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+  private boolean isNsDelete;
   @Inject ResourceCommand resourceCommand;
   @Inject ExtensionManager extensionManager;
   @Inject EppInput eppInput;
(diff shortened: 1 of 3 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1663)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 208 at r1](https://reviewable.io/reviews/google/nomulus/1663#-N48ggSQ4pCu3vUcWdoQ:-N48ggSQ4pCu3vUcWdoR:bbspsuk) ([raw file](https://github.com/google/nomulus/blob/26dd77af9963b5b612a0230e04f220d69a9fc04e/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L208)):*
> ```Java
>                 .build());
>     persistEntityChanges(entityChanges);
>     // We would like to reload the persisted entity and show what is actually persisted, but that
> ```

This explanation needs some edits.

Reloading in the current session will only give back the cached version in Hibernate. It is impossible to see what is actually persisted in the DB because PSQL doesn't support READ_UNCOMMITTED. So even if we call flush here and query with another entitymanager at READ_UNCOMMITTED isolation level, PSQL won't show us the new data.



_Code quote:_
```Java
 We would like to reload the persisted entity and show what is actually persisted
```

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1663)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 208 at r1](https://reviewable.io/reviews/google/nomulus/1663#-N48ggSQ4pCu3vUcWdoQ:-N48lEoWCG6qWk3JmktY:byinwqx) ([raw file](https://github.com/google/nomulus/blob/26dd77af9963b5b612a0230e04f220d69a9fc04e/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L208)):*
> Reloading in the current session will only give back the cached version in Hibernate. It is impossible to see what is actually persisted in the DB because PSQL doesn't support READ_UNCOMMITTED. So even if we call flush here and query with another entitymanager at READ_UNCOMMITTED isolation level, PSQL won't show us the new data.

That's a very good explanation. Thanks!

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1663)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @weiminyu)

PR #1664, opened 2022-06-09
Split failing dns update batches and kill after 10 retries

sarahcaseybot:
This implements go/terminatednsupdates. I already tested this in crash and verified that it properly split the batch after 3 retries and then terminated after 10. I will add an email feature in a follow up PR that emails partners to inform them that their dns update failed. This implementation was first attempted in go/r3pr/1628 but I closed that PR in favor of this one.

Summary of go/terminatednsupdates

- Identify batches of failing dns updates 
- After 3 retries of a batch, split the batch in half and retry
- Continue to split failing batch in half until size == 1
- Kill updates of batches with size == 1 after 10 retries

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1664)

### DIFF
--- a/core/src/main/java/google/registry/dns/DnsMetrics.java
+++ b/core/src/main/java/google/registry/dns/DnsMetrics.java
@@ -46,7 +46,7 @@ public class DnsMetrics {
     LOCK_FAILURE,
     BAD_WRITER,
     BAD_LOCK_INDEX,
-    PERMANENT_FAILURE
+    MAX_RETRIES_EXCEEDED
   }
 
   private static final ImmutableSet<LabelDescriptor> LABEL_DESCRIPTORS_FOR_PUBLISH_REQUESTS =
--- a/core/src/main/java/google/registry/dns/DnsMetrics.java
+++ b/core/src/main/java/google/registry/dns/DnsMetrics.java
@@ -40,7 +40,14 @@ public class DnsMetrics {
   public enum CommitStatus { SUCCESS, FAILURE }
 
   /** Disposition of the publish action. */
-  public enum ActionStatus { SUCCESS, COMMIT_FAILURE, LOCK_FAILURE, BAD_WRITER, BAD_LOCK_INDEX }
+  public enum ActionStatus {
+    SUCCESS,
+    COMMIT_FAILURE,
+    LOCK_FAILURE,
+    BAD_WRITER,
+    BAD_LOCK_INDEX,
+    TIMED_OUT
+  }
 
   private static final ImmutableSet<LabelDescriptor> LABEL_DESCRIPTORS_FOR_PUBLISH_REQUESTS =
       ImmutableSet.of(
--- a/core/src/main/java/google/registry/dns/DnsMetrics.java
+++ b/core/src/main/java/google/registry/dns/DnsMetrics.java
@@ -46,7 +46,7 @@ public class DnsMetrics {
     LOCK_FAILURE,
     BAD_WRITER,
     BAD_LOCK_INDEX,
-    TIMED_OUT
+    PERMANENT_FAILURE
   }
 
   private static final ImmutableSet<LabelDescriptor> LABEL_DESCRIPTORS_FOR_PUBLISH_REQUESTS =
--- a/core/src/main/java/google/registry/dns/DnsModule.java
+++ b/core/src/main/java/google/registry/dns/DnsModule.java
@@ -16,10 +16,8 @@ package google.registry.dns;
 
 import static google.registry.dns.DnsConstants.DNS_PUBLISH_PUSH_QUEUE_NAME;
 import static google.registry.dns.DnsConstants.DNS_PULL_QUEUE_NAME;
-import static google.registry.dns.PublishDnsUpdatesAction.RETRY_HEADER;
 import static google.registry.request.RequestParameters.extractEnumParameter;
 import static google.registry.request.RequestParameters.extractIntParameter;
-import static google.registry.request.RequestParameters.extractRequiredHeader;
 import static google.registry.request.RequestParameters.extractRequiredParameter;
 import static google.registry.request.RequestParameters.extractSetOfParameters;
 
@@ -32,7 +30,6 @@ import dagger.Module;
 import dagger.Provides;
 import google.registry.dns.DnsConstants.TargetType;
 import google.registry.dns.writer.DnsWriterZone;
-import google.registry.request.Header;
 import google.registry.request.Parameter;
 import google.registry.request.RequestParameters;
 import java.util.Set;
@@ -131,11 +128,5 @@ public abstract class DnsModule {
   @Parameter("type")
   static TargetType provideType(HttpServletRequest req) {
     return extractEnumParameter(req, TargetType.class, "type");
-  }
-
-  @Provides
-  @Header(RETRY_HEADER)
-  static int provideRetryCount(HttpServletRequest req) {
-    return Integer.parseInt(extractRequiredHeader(req, RETRY_HEADER));
   }
 }
--- a/core/src/main/java/google/registry/dns/DnsModule.java
+++ b/core/src/main/java/google/registry/dns/DnsModule.java
@@ -16,8 +16,10 @@ package google.registry.dns;
 
 import static google.registry.dns.DnsConstants.DNS_PUBLISH_PUSH_QUEUE_NAME;
 import static google.registry.dns.DnsConstants.DNS_PULL_QUEUE_NAME;
+import static google.registry.dns.PublishDnsUpdatesAction.RETRY_HEADER;
 import static google.registry.request.RequestParameters.extractEnumParameter;
 import static google.registry.request.RequestParameters.extractIntParameter;
+import static google.registry.request.RequestParameters.extractRequiredHeader;
 import static google.registry.request.RequestParameters.extractRequiredParameter;
 import static google.registry.request.RequestParameters.extractSetOfParameters;
 
@@ -30,6 +32,7 @@ import dagger.Module;
 import dagger.Provides;
 import google.registry.dns.DnsConstants.TargetType;
 import google.registry.dns.writer.DnsWriterZone;
+import google.registry.request.Header;
 import google.registry.request.Parameter;
 import google.registry.request.RequestParameters;
 import java.util.Set;
@@ -128,5 +131,11 @@ public abstract class DnsModule {
   @Parameter("type")
   static TargetType provideType(HttpServletRequest req) {
     return extractEnumParameter(req, TargetType.class, "type");
+  }
+
+  @Provides
+  @Header(RETRY_HEADER)
+  static int provideRetryCount(HttpServletRequest req) {
+    return Integer.parseInt(extractRequiredHeader(req, RETRY_HEADER));
   }
 }
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -25,7 +25,7 @@ import static google.registry.dns.DnsModule.PARAM_REFRESH_REQUEST_CREATED;
 import static google.registry.request.Action.Method.POST;
 import static google.registry.request.RequestParameters.PARAM_TLD;
 import static google.registry.util.CollectionUtils.nullToEmpty;
-import static java.lang.Math.max;
+import static javax.servlet.http.HttpServletResponse.SC_ACCEPTED;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
@@ -43,11 +43,13 @@ import google.registry.request.Action.Service;
 import google.registry.request.Header;
 import google.registry.request.HttpException.ServiceUnavailableException;
 import google.registry.request.Parameter;
+import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import google.registry.request.lock.LockHandler;
 import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.DomainNameUtils;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import javax.inject.Inject;
@@ -77,8 +79,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   private final DnsWriterProxy dnsWriterProxy;
   private final DnsMetrics dnsMetrics;
   private final Duration timeout;
-  private final int appEngineRetryCount;
-  private final int cloudTasksRetryCount;
+  private final int retryCount;
 
   /**
    * The DNS writer to use for this batch.
@@ -100,6 +101,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   private final LockHandler lockHandler;
   private final Clock clock;
   private final CloudTasksUtils cloudTasksUtils;
+  private final Response response;
 
   @Inject
   public PublishDnsUpdatesAction(
@@ -112,20 +114,21 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
       @Parameter(PARAM_HOSTS) Set<String> hosts,
       @Parameter(PARAM_TLD) String tld,
       @Config("publishDnsUpdatesLockDuration") Duration timeout,
-      @Header(APP_ENGINE_RETRY_HEADER) int appEngineRetryCount,
-      @Header(CLOUD_TASKS_RETRY_HEADER) int cloudTasksRetryCount,
+      @Header(APP_ENGINE_RETRY_HEADER) Optional<String> appEngineRetryCount,
+      @Header(CLOUD_TASKS_RETRY_HEADER) Optional<String> cloudTasksRetryCount,
       DnsQueue dnsQueue,
       DnsWriterProxy dnsWriterProxy,
       DnsMetrics dnsMetrics,
       LockHandler lockHandler,
       Clock clock,
-      CloudTasksUtils cloudTasksUtils) {
+      CloudTasksUtils cloudTasksUtils,
+      Response response) {
     this.dnsQueue = dnsQueue;
     this.dnsWriterProxy = dnsWriterProxy;
     this.dnsMetrics = dnsMetrics;
     this.timeout = timeout;
-    this.appEngineRetryCount = appEngineRetryCount;
-    this.cloudTasksRetryCount = cloudTasksRetryCount;
+    this.retryCount =
+        Integer.parseInt(cloudTasksRetryCount.orElse(appEngineRetryCount.orElse("0")));
     this.dnsWriter = dnsWriter;
     this.enqueuedTime = enqueuedTime;
     this.itemsCreateTime = itemsCreateTime;
@@ -137,6 +140,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
     this.lockHandler = lockHandler;
     this.clock = clock;
     this.cloudTasksUtils = cloudTasksUtils;
+    this.response = response;
   }
 
   private void recordActionResult(ActionStatus status) {
@@ -189,7 +193,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
       processBatch();
     } catch (Throwable e) {
       // Retry the batch 3 times
-      if (max(appEngineRetryCount, cloudTasksRetryCount) < 3) {
+      if (retryCount < 3) {
         throw e;
       }
       // After 3 retries, split the batch
@@ -202,13 +206,14 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
         enqueue(ImmutableList.of(), ImmutableList.copyOf(hosts));
       }
       // If the batch only contains 1 name, allow it more retries
-      else if (max(appEngineRetryCount, cloudTasksRetryCount) < RETRIES_BEFORE_PERMANENT_FAILURE) {
+      else if (retryCount < RETRIES_BEFORE_PERMANENT_FAILURE) {
         throw e;
       }
       // If we get here, we should terminate this task as it is likely a perpetually failing task.
       // TODO(b/237302821): Send an email notifying partner the dns update failed
       recordActionResult(ActionStatus.TIMED_OUT);
-      logger.atSevere().log("Terminated task after too many retries");
+      response.setStatus(SC_ACCEPTED);
+      logger.atSevere().withCause(e).log("Terminated task after too many retries");
     }
     return null;
   }
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -137,23 +137,24 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   @Override
   public void run() {
     try {
-    if (!validLockParams()) {
-      recordActionResult(ActionStatus.BAD_LOCK_INDEX);
-      requeueBatch();
-      return;
-    }
-    // If executeWithLocks fails to get the lock, it does not throw an exception, simply returns
-    // false. We need to make sure to take note of this error; otherwise, a failed lock might result
-    // in the update task being dequeued and dropped. A message will already have been logged
-    // to indicate the problem.
-    if (!lockHandler.executeWithLocks(
-        this,
-        tld,
-        timeout,
-        String.format("%s-lock %d of %d", LOCK_NAME, lockIndex, numPublishLocks))) {
-      recordActionResult(ActionStatus.LOCK_FAILURE);
-      throw new ServiceUnavailableException("Lock failure");
-    }
+      if (!validLockParams()) {
+        recordActionResult(ActionStatus.BAD_LOCK_INDEX);
+        requeueBatch();
+        return;
+      }
+      // If executeWithLocks fails to get the lock, it does not throw an exception, simply returns
+      // false. We need to make sure to take note of this error; otherwise, a failed lock might
+      // result
+      // in the update task being dequeued and dropped. A message will already have been logged
+      // to indicate the problem.
+      if (!lockHandler.executeWithLocks(
+          this,
+          tld,
+          timeout,
+          String.format("%s-lock %d of %d", LOCK_NAME, lockIndex, numPublishLocks))) {
+        recordActionResult(ActionStatus.LOCK_FAILURE);
+        throw new ServiceUnavailableException("Lock failure");
+      }
     } catch (Throwable e) {
       // Retry the batch 3 times
       if (retryCount < 3) {
@@ -198,6 +199,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
     return null;
   }
 
+  /** Splits the domains and hosts in a batch into 2 batches and adds them to the queue. */
   private void splitBatch() {
     Set<String> set1Domain = new HashSet<>();
     Set<String> set2Domain = new HashSet<>();
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -14,6 +14,7 @@
 
 package google.registry.dns;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static google.registry.dns.DnsConstants.DNS_PUBLISH_PUSH_QUEUE_NAME;
 import static google.registry.dns.DnsModule.PARAM_DNS_WRITER;
 import static google.registry.dns.DnsModule.PARAM_DOMAINS;
@@ -25,9 +26,11 @@ import static google.registry.dns.DnsModule.PARAM_REFRESH_REQUEST_CREATED;
 import static google.registry.request.Action.Method.POST;
 import static google.registry.request.RequestParameters.PARAM_TLD;
 import static google.registry.util.CollectionUtils.nullToEmpty;
+import static java.lang.Math.max;
 
 import com.google.appengine.api.taskqueue.Queue;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
@@ -47,8 +50,6 @@ import google.registry.request.lock.LockHandler;
 import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.DomainNameUtils;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import javax.inject.Inject;
@@ -67,7 +68,11 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
 
   public static final String PATH = "/_dr/task/publishDnsUpdates";
   public static final String LOCK_NAME = "DNS updates";
-  public static final String RETRY_HEADER = "X-AppEngine-TaskRetryCount";
+  // TODO(b/236726584): Remove App Engine header once CloudTasksUtils is refactored to create HTTP
+  // tasks.
+  public static final String APP_ENGINE_RETRY_HEADER = "X-AppEngine-TaskRetryCount";
+  public static final String CLOUD_TASKS_RETRY_HEADER = "X-CloudTasks-TaskRetryCount";
+  public static final int RETRIES_BEFORE_PERMANENT_FAILURE = 10;
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
@@ -77,8 +82,12 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   @Inject @Config("publishDnsUpdatesLockDuration") Duration timeout;
 
   @Inject
-  @Header(RETRY_HEADER)
-  int retryCount;
+  @Header(APP_ENGINE_RETRY_HEADER)
+  int appEngineRetryCount;
+
+  @Inject
+  @Header(CLOUD_TASKS_RETRY_HEADER)
+  int cloudTasksRetryCount;
 
   /**
    * The DNS writer to use for this batch.
@@ -138,54 +147,60 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   /** Runs the task. */
   @Override
   public void run() {
-    try {
-      if (!validLockParams()) {
-        recordActionResult(ActionStatus.BAD_LOCK_INDEX);
-        requeueBatch();
-        return;
-      }
-      // If executeWithLocks fails to get the lock, it does not throw an exception, simply returns
-      // false. We need to make sure to take note of this error; otherwise, a failed lock might
-      // result in the update task being dequeued and dropped. A message will already have been
-      // logged to indicate the problem.
-      if (!lockHandler.executeWithLocks(
-          this,
-          tld,
-          timeout,
-          String.format("%s-lock %d of %d", LOCK_NAME, lockIndex, numPublishLocks))) {
-        recordActionResult(ActionStatus.LOCK_FAILURE);
-        throw new ServiceUnavailableException("Lock failure");
-      }
-    } catch (Throwable e) {
-      // Retry the batch 3 times
-      if (retryCount < 3) {
-        throw e;
-      }
-      // After 3 retries, split the batch
-      if (domains.size() > 1 || hosts.size() > 1) {
-        // split batch and requeue
-        splitBatch();
-      }
-      // If the batch only contains 1 name, allow it 10 retries
-      else if (retryCount < 10) {
-        throw e;
-      }
-      // If we get here, we should terminate this task as it is likely a perpetually failing task.
-      // TODO(sarahbot): Send an email notifying partner the dns update failed
+    if (!validLockParams()) {
+      recordActionResult(ActionStatus.BAD_LOCK_INDEX);
+      requeueBatch();
+      return;
+    }
+    // If executeWithLocks fails to get the lock, it does not throw an exception, simply returns
+    // false. We need to make sure to take note of this error; otherwise, a failed lock might
+    // result in the update task being dequeued and dropped. A message will already have been
+    // logged to indicate the problem.
+    if (!lockHandler.executeWithLocks(
+        this,
+        tld,
+        timeout,
+        String.format("%s-lock %d of %d", LOCK_NAME, lockIndex, numPublishLocks))) {
+      recordActionResult(ActionStatus.LOCK_FAILURE);
+      throw new ServiceUnavailableException("Lock failure");
     }
   }
 
   /** Runs the task, with the lock. */
   @Override
   public Void call() {
-    processBatch();
+    try {
+      processBatch();
+    } catch (Throwable e) {
+      // Retry the batch 3 times
+      if (max(appEngineRetryCount, cloudTasksRetryCount) < 3) {
+        throw e;
+      }
+      // After 3 retries, split the batch
+      if (domains.size() > 1 || hosts.size() > 1) {
+        // split batch and requeue
+        splitBatch();
+      } else if (domains.size() == 1 && hosts.size() == 1) {
+        // Enqueue the single domain and single host separately
+        enqueue(domains.stream().collect(toImmutableList()), ImmutableList.of());
+        enqueue(ImmutableList.of(), hosts.stream().collect(toImmutableList()));
+      }
+      // If the batch only contains 1 name, allow it more retries
+      else if (max(appEngineRetryCount, cloudTasksRetryCount) < RETRIES_BEFORE_PERMANENT_FAILURE) {
+        throw e;
+      }
+      // If we get here, we should terminate this task as it is likely a perpetually failing task.
+      // TODO(sarahbot): Send an email notifying partner the dns update failed
+      recordActionResult(ActionStatus.TIMED_OUT);
+      logger.atSevere().log("Terminated task after too many retries");
+    }
     return null;
   }
 
   /** Splits the domains and hosts in a batch into smaller batches and adds them to the queue. */
   private void splitBatch() {
-    List<String> domainList = new ArrayList<>(domains);
-    List<String> hostList = new ArrayList<>(hosts);
+    ImmutableList<String> domainList = domains.stream().collect(toImmutableList());
+    ImmutableList<String> hostList = hosts.stream().collect(toImmutableList());
 
     enqueue(domainList.subList(0, domains.size() / 2), hostList.subList(0, hosts.size() / 2));
     enqueue(
@@ -193,7 +208,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
         hostList.subList(hosts.size() / 2, hosts.size()));
   }
 
-  private void enqueue(List<String> domains, List<String> hosts) {
+  private void enqueue(ImmutableList<String> domains, ImmutableList<String> hosts) {
     cloudTasksUtils.enqueue(
         DNS_PUBLISH_PUSH_QUEUE_NAME,
         cloudTasksUtils.createPostTask(
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -190,7 +190,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
         throw e;
       }
       // If we get here, we should terminate this task as it is likely a perpetually failing task.
-      // TODO(sarahbot): Send an email notifying partner the dns update failed
+      // TODO(b/237302821): Send an email notifying partner the dns update failed
       recordActionResult(ActionStatus.TIMED_OUT);
       logger.atSevere().log("Terminated task after too many retries");
     }
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -14,6 +14,7 @@
 
 package google.registry.dns;
 
+import static google.registry.dns.DnsConstants.DNS_PUBLISH_PUSH_QUEUE_NAME;
 import static google.registry.dns.DnsModule.PARAM_DNS_WRITER;
 import static google.registry.dns.DnsModule.PARAM_DOMAINS;
 import static google.registry.dns.DnsModule.PARAM_HOSTS;
@@ -25,6 +26,8 @@ import static google.registry.request.Action.Method.POST;
 import static google.registry.request.RequestParameters.PARAM_TLD;
 import static google.registry.util.CollectionUtils.nullToEmpty;
 
+import com.google.appengine.api.taskqueue.Queue;
+import com.google.appengine.api.taskqueue.TaskOptions;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
 import google.registry.config.RegistryConfig.Config;
@@ -34,15 +37,20 @@ import google.registry.dns.DnsMetrics.PublishStatus;
 import google.registry.dns.writer.DnsWriter;
 import google.registry.model.tld.Registry;
 import google.registry.request.Action;
+import google.registry.request.Header;
 import google.registry.request.HttpException.ServiceUnavailableException;
 import google.registry.request.Parameter;
 import google.registry.request.auth.Auth;
 import google.registry.request.lock.LockHandler;
 import google.registry.util.Clock;
 import google.registry.util.DomainNameUtils;
+import google.registry.util.TaskQueueUtils;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.stream.Collectors;
 import javax.inject.Inject;
+import javax.inject.Named;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
 
@@ -57,6 +65,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
 
   public static final String PATH = "/_dr/task/publishDnsUpdates";
   public static final String LOCK_NAME = "DNS updates";
+  public static final String RETRY_HEADER = "X-AppEngine-TaskRetryCount";
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
@@ -64,6 +73,10 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   @Inject DnsWriterProxy dnsWriterProxy;
   @Inject DnsMetrics dnsMetrics;
   @Inject @Config("publishDnsUpdatesLockDuration") Duration timeout;
+
+  @Inject
+  @Header(RETRY_HEADER)
+  int retryCount;
 
   /**
    * The DNS writer to use for this batch.
@@ -88,8 +101,15 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   @Inject @Parameter(PARAM_DOMAINS) Set<String> domains;
   @Inject @Parameter(PARAM_HOSTS) Set<String> hosts;
   @Inject @Parameter(PARAM_TLD) String tld;
+
+  @Inject
+  @Named(DNS_PUBLISH_PUSH_QUEUE_NAME)
+  Queue dnsPublishPushQueue;
+
   @Inject LockHandler lockHandler;
   @Inject Clock clock;
+  @Inject TaskQueueUtils taskQueueUtils;
+
   @Inject PublishDnsUpdatesAction() {}
 
   private void recordActionResult(ActionStatus status) {
@@ -116,6 +136,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   /** Runs the task. */
   @Override
   public void run() {
+    try {
     if (!validLockParams()) {
       recordActionResult(ActionStatus.BAD_LOCK_INDEX);
       requeueBatch();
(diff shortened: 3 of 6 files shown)


gbrodman (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 3 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DX3_PAq6V1gE-wZYg:-N4DX3_PAq6V1gE-wZYh:bmb7t77) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L68)):*
> ```Java
>   public static final String PATH = "/_dr/task/publishDnsUpdates";
>   public static final String LOCK_NAME = "DNS updates";
>   public static final String RETRY_HEADER = "X-AppEngine-TaskRetryCount";
> ```

do you happen to know what we'll do once we move away from appengine? 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 147 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4D_mwQ3o4zESOQ9zKa:-N4D_mwQ3o4zESOQ9zKb:b4eb4jk) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L147)):*
> ```Java
>       // If executeWithLocks fails to get the lock, it does not throw an exception, simply returns
>       // false. We need to make sure to take note of this error; otherwise, a failed lock might
>       // result
> ```

nit: comment formatting 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Ddejr0KSWrW2bBcIK:-N4Ddejr0KSWrW2bBcIL:b7o92eu) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L158)):*
> ```Java
>         throw new ServiceUnavailableException("Lock failure");
>       }
>     } catch (Throwable e) {
> ```

Don't we already handle this in the `call` method below? 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 169 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Dd8HP9K2LvdLQpEcC:-N4Dd8HP9K2LvdLQpEcD:b-mtycc9) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L169)):*
> ```Java
>       }
>       // If the batch only contains 1 name, allow it 10 retries
>       else if (retryCount < 10) {
> ```

nit, switch the ordering of this line and the previous (comment) line 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 203 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DeKhW5en7WutSejee:-N4DeKhW5en7WutSejef:bi5jqra) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L203)):*
> ```Java
> 
>   /** Splits the domains and hosts in a batch into 2 batches and adds them to the queue. */
>   private void splitBatch() {
> ```

Is there any reason why we can't just use `Iterables::partition` to split the domains and hosts? 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 225 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DeQcX9ZpRGWtgVClG:-N4DeQcX9ZpRGWtgVClH:b-69lxml) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L225)):*
> ```Java
>     }
> 
>     // Enqueue set 1
> ```

to be safe we can have this be a method to which we pass `domains` and `hosts` so that we don't accidentally mess up task enqueueing 
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Dea1i8jMmLiV7QxxI:-N4Dea1i8jMmLiV7QxxJ:b-gnw7kt) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L113)):*
> ```Java
>     action.clock = clock;
>     action.taskQueueUtils = taskQueueUtils;
>     action.retryCount = 1;
> ```

I'm not sure if this matters but the retry count starts at 0 right? 
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 256 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Dekjo1yp_J98aTG69:-N4Dekjo1yp_J98aTG6A:bbvbvf0) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L256)):*
> ```Java
>     verify(taskQueueUtils).enqueue(queue, taskOptions2);
>   }
> ```

Shouldn't we also test that we stop after the 10th failure?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DX3_PAq6V1gE-wZYg:-N4Ty3G40YJi60mhbn-Z:b-rpmxxt) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L68)):*

do you happen to know what we'll do once we move away from appengine? 
</blockquote></details>

I don't know what the plan is for that. Do we know yet what we will do with the task queues?
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Ddejr0KSWrW2bBcIK:-N4TUAGQ4PrAVRNPpl9S:b-fohfg6) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L158)):*

Don't we already handle this in the `call` method below? 
</blockquote></details>

This is needed here too since we only call run() in tests
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 203 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DeKhW5en7WutSejee:-N4TtoZ529aU3ksnZRdA:bag1bd) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L203)):*

Is there any reason why we can't just use `Iterables::partition` to split the domains and hosts? 
</blockquote></details>

Using partition could work, but there is a lot of extra handling that needs to be added since domains or hosts could be empty lists which would cause an IllegalArgumentException in partition. 

Also, partition would split into 2 or 3 lists depending on if the size is even or odd. Then, there may be more lists for host than there are for domains or vice versa. Ex: there may be no lists for host if there are no hosts in the original batch but if the domain list was originally of size 3 it would be split into 3 lists of size 1 which would leave us with zero host lists and 3 domain lists. 

Obviously all that could all be handled, but it just seemed cleaner to do it this way instead.  
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 225 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DeQcX9ZpRGWtgVClG:-N4TvrJz-HhXESgznyUk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L225)):*

to be safe we can have this be a method to which we pass `domains` and `hosts` so that we don't accidentally mess up task enqueueing 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Dea1i8jMmLiV7QxxI:-N4TvyEKATjSFRAT3pYs:bm8h2v8) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L113)):*

I'm not sure if this matters but the retry count starts at 0 right? 
</blockquote></details>

yes, I didn't think it mattered so I just randomly chose 1, but I changed it now to zero.
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 256 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Dekjo1yp_J98aTG69:-N4Ty1NM9Gb3EbmBW36M:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L256)):*

Shouldn't we also test that we stop after the 10th failure? 
</blockquote></details>

Done.

jianglai (changes requested):
It's better to include a summary of what this PR does in the description. Not everyone has access to the design doc.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664#-:-N4YZXu0EFUBuctruqCv:b-qyewxo)* status: 1 of 3 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/DnsModule.java` line 137 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4YXshd7c1LhFgk3HQe:-N4YXshd7c1LhFgk3HQf:bt3ms37) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/DnsModule.java#L137)):*
> ```Java
> 
>   @Provides
>   @Header(RETRY_HEADER)
> ```

I think it is better to provide this binding in `RequestModule`, as the header is generic to all requests.



_Code quote:_
```Java
  @Provides
  @Header(RETRY_HEADER)
  static int provideRetryCount(HttpServletRequest req) {
    return Integer.parseInt(extractRequiredHeader(req, RETRY_HEADER));
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DX3_PAq6V1gE-wZYg:-N4YX4CP-WadmS9Bn5B4:b-gw3eao) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L68)):*

I don't know what the plan is for that. Do we know yet what we will do with the task queues?
</blockquote></details>

https://cloud.google.com/tasks/docs/creating-http-target-tasks

It looks to be `X-CloudTasks-TaskRetryCount`. Not really a surprise here.



_Code quote:_
```Java
X-AppEngine-TaskRetryCount
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Ddejr0KSWrW2bBcIK:-N4Y_zve2r1_6atc6ZTw:b-oiz38i) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L158)):*

This is needed here too since we only call run() in tests
</blockquote></details>

If that's the case, why do we need the try block in the `call` method? Is it called anywhere other than inside the `run` method?
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 111 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4YXFd58sgZeZSXHhxY:-N4YXFd58sgZeZSXHhxZ:bzbrnwh) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L111)):*
> ```Java
>   @Inject LockHandler lockHandler;
>   @Inject Clock clock;
>   @Inject TaskQueueUtils taskQueueUtils;
> ```

We should not be using TaskQueueUtils anymore. Use `CloudTasksUtils` instead.



_Code quote:_
```Java
TaskQueueUtils taskQueueUtil
```

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 203 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DeKhW5en7WutSejee:-N4Zw3gz4qRtCmiWyM3f:b-psip5g) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L203)):*

Using partition could work, but there is a lot of extra handling that needs to be added since domains or hosts could be empty lists which would cause an IllegalArgumentException in partition. 

Also, partition would split into 2 or 3 lists depending on if the size is even or odd. Then, there may be more lists for host than there are for domains or vice versa. Ex: there may be no lists for host if there are no hosts in the original batch but if the domain list was originally of size 3 it would be split into 3 lists of size 1 which would leave us with zero host lists and 3 domain lists. 

Obviously all that could all be handled, but it just seemed cleaner to do it this way instead.  
</blockquote></details>

actually a better solution that would solve all of this would be to just convert them to a list (or inject them as a list, whichever) then just use `domains.subList(0, domains.size() / 2)` and `domains.subList(domains.size() / 2, domains.size())` as your two lists 

then, to create the strings, we can use the better-supported `Joiner.on(",").join(...)` 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 111 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4YXFd58sgZeZSXHhxY:-N4Zug6jBoDb1Epo-gIt:b7pembw) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L111)):*

We should not be using TaskQueueUtils anymore. Use `CloudTasksUtils` instead.
</blockquote></details>

in addition to doing this, while you're at it maybe mark TaskQueueUtils as `@Deprecated` so that others don't run into the same thing? 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 163 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4ZwmjQABOCM89ljdBO:-N4ZwmjQABOCM89ljdBP:bs4iouq) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L163)):*
> ```Java
>       }
>       // After 3 retries, split the batch
>       if (domains.size() > 1 || hosts.size() > 1) {
> ```

What happens if we have a ton of domains but no hosts, but it's still failing? Wouldn't we still want to split the domains up? 

(also we'd want isEmpty checks here anyway)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Ddejr0KSWrW2bBcIK:-N4c0ixb0AGdpD-KDGSX:b-3l0lae) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L158)):*

If that's the case, why do we need the try block in the `call` method? Is it called anywhere other than inside the `run` method?
</blockquote></details>

right, the thinking is that we only want to handle exceptions in one place

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 111 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4YXFd58sgZeZSXHhxY:-N4cTv8t89wTVhc0iorI:b-mctn33) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L111)):*

in addition to doing this, while you're at it maybe mark TaskQueueUtils as `@Deprecated` so that others don't run into the same thing? 
</blockquote></details>

#1669 is now merged. We can rebase this PR, replace `TaskQueueUtils` with `CloudTasksUtils` and deploy to alpha to test.

jianglai (changes requested):
> Reviewed 2 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 70 at r3](https://reviewable.io/reviews/google/nomulus/1664#-N55q8og4g59h5opmZLM:-N55q8og4g59h5opmZLN:bt22cva) ([raw file](https://github.com/google/nomulus/blob/9d0a48d4b91113688b24b177895cbe56be5a4ffe/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L70)):*
> ```Java
>   public static final String PATH = "/_dr/task/publishDnsUpdates";
>   public static final String LOCK_NAME = "DNS updates";
>   public static final String RETRY_HEADER = "X-CloudTasks-TaskRetryCount";
> ```

Let's link to the source:

https://cloud.google.com/tasks/docs/creating-http-target-tasks

Also, have you tried this on alpha yet? My concern is that Cloud Tasks allows you to create either an App Engine task or an HTTP task. We currently still create the former because we still use GAE, but this header is listed in the documentation for the latter. So we should make sure that App Engine tasks created by Cloud Tasks contains this header.

Ultimately once we move off of GAE, we will need to refactor `CloudTasksUtils` to create HTTP tasks instead and this header will definitely be the one to use, but I'm not sure for the time being.



_Code quote:_
```Java
public static final String RETRY_HEADER = "X-CloudTasks-TaskRetryCount";
```

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4Ddejr0KSWrW2bBcIK:-N56YLd5Cp8EWrXHnCqJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L158)):*

right, the thinking is that we only want to handle exceptions in one place 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 203 at r1](https://reviewable.io/reviews/google/nomulus/1664#-N4DeKhW5en7WutSejee:-N4nGyfx8KovVosXqlDB:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d31e295220b39898029e283c200f0bd49ce06614/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L203)):*

actually a better solution that would solve all of this would be to just convert them to a list (or inject them as a list, whichever) then just use `domains.subList(0, domains.size() / 2)` and `domains.subList(domains.size() / 2, domains.size())` as your two lists 

then, to create the strings, we can use the better-supported `Joiner.on(",").join(...)` 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 163 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4ZwmjQABOCM89ljdBO:-N4n9Jdf1k5soGUQj1u-:buvy7dt) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L163)):*

What happens if we have a ton of domains but no hosts, but it's still failing? Wouldn't we still want to split the domains up? 

(also we'd want isEmpty checks here anyway) 
</blockquote></details>

I think this handles the situation you're describing.  If we have say...

domains.size() = 5
hosts.size() = 0
retryCount = 3

this code would call splitBatch()

I don't think isEmpty checks are necessary since a .size() call on an empty set returns 0
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 70 at r3](https://reviewable.io/reviews/google/nomulus/1664#-N55q8og4g59h5opmZLM:-N56Y3sP73dpvEcOg3Gi:bo1gxjl) ([raw file](https://github.com/google/nomulus/blob/9d0a48d4b91113688b24b177895cbe56be5a4ffe/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L70)):*

Let's link to the source:

https://cloud.google.com/tasks/docs/creating-http-target-tasks

Also, have you tried this on alpha yet? My concern is that Cloud Tasks allows you to create either an App Engine task or an HTTP task. We currently still create the former because we still use GAE, but this header is listed in the documentation for the latter. So we should make sure that App Engine tasks created by Cloud Tasks contains this header.

Ultimately once we move off of GAE, we will need to refactor `CloudTasksUtils` to create HTTP tasks instead and this header will definitely be the one to use, but I'm not sure for the time being.
</blockquote></details>

Changed this back to the App Engine Header as the Cloud Tasks header did not work in Alpha. Added a TODO in CloudTasksUtils to change the header once CloudTasksUtils is refactored to create HTTP tasks
___
*[`core/src/main/java/google/registry/dns/DnsModule.java` line 137 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4YXshd7c1LhFgk3HQe:-N4iGSUA27N1hgf3yW03:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/DnsModule.java#L137)):*

I think it is better to provide this binding in `RequestModule`, as the header is generic to all requests.
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 5 files reviewed, 4 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 70 at r3](https://reviewable.io/reviews/google/nomulus/1664#-N55q8og4g59h5opmZLM:-N5FrH5_59YMyjQPkmhM:b-q7eim6) ([raw file](https://github.com/google/nomulus/blob/9d0a48d4b91113688b24b177895cbe56be5a4ffe/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L70)):*

Changed this back to the App Engine Header as the Cloud Tasks header did not work in Alpha. Added a TODO in CloudTasksUtils to change the header once CloudTasksUtils is refactored to create HTTP tasks
</blockquote></details>

It would be more resilient to check _both_, and use whichever happens to be present, with a TODO to remove the App Engine one post-migration.  That method should be more resilient across the migration boundary.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 5 files reviewed, 4 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 163 at r2](https://reviewable.io/reviews/google/nomulus/1664#-N4ZwmjQABOCM89ljdBO:-N5Fv5XM1VdimR6G8S-S:bp7ea0n) ([raw file](https://github.com/google/nomulus/blob/e7938837ab90d575c81d71e67899f3e9e3ad62b4/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L163)):*

I think this handles the situation you're describing.  If we have say...

domains.size() = 5
hosts.size() = 0
retryCount = 3

this code would call splitBatch()

I don't think isEmpty checks are necessary since a .size() call on an empty set returns 0
</blockquote></details>

What if there's 1 domain and 1 host?  Shouldn't that still be split?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 170 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5FvA5c-VQ7GwxC3Rtc:-N5FvA5c-VQ7GwxC3Rtd:bky9zie) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L170)):*
> ```Java
>       }
>       // If the batch only contains 1 name, allow it 10 retries
>       else if (retryCount < 10) {
> ```

Don't hard-code a magic number like 10.  This should, at minimum, be a `static final int`, and then possibly, configurable.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 5 files reviewed, 6 unresolved discussions (waiting on @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 239 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5Fvg2UChPC_vYyn9St:-N5Fvg2UChPC_vYyn9Su:b-ftutdg) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L239)):*
> ```Java
>     action.domains =
>         ImmutableSet.of(
>             "example1.xn--q9jyb4c",
> ```

Can you make this 5 domains?  I want to see how `subList()` handles an uneven split.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 5 files reviewed, 7 unresolved discussions (waiting on @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 187 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5FwH073ERmQzTmozMX:-N5FwH073ERmQzTmozMY:brpla7r) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L187)):*
> ```Java
>   /** Splits the domains and hosts in a batch into smaller batches and adds them to the queue. */
>   private void splitBatch() {
>     List<String> domainList = new ArrayList<>(domains);
> ```

Use `ImmutableList` here instead of `ArrayList`.  Basic guidance on Guava's immutable collections is to always use them unless for some reason you can't (e.g. you need to continually modify the contents, which isn't required here).

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 173 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5Fwx_PAXVLnCoL9YR5:-N5Fwx_PAXVLnCoL9YR6:b37ftqv) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L173)):*
> ```Java
>         throw e;
>       }
>       // If we get here, we should terminate this task as it is likely a perpetually failing task.
> ```

Set the HTTP response code to something different in this case to distinguish it from the `200` response that would indicate total success.  Maybe `202` is suitable?  This should still cause App Engine task queue to not retry, but will have a different status code we could e.g. search out in the logs.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 173 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5FxszW8x7XS592WJ2c:-N5FxszW8x7XS592WJ2d:bdkxu4g) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L173)):*
> ```Java
>         throw e;
>       }
>       // If we get here, we should terminate this task as it is likely a perpetually failing task.
> ```

There's a missing log statement here.  Log the permanently failed task at `SEVERE`.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 3 of 6 files reviewed, 11 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 84 at r6](https://reviewable.io/reviews/google/nomulus/1664#-N5eeGSS-ehp-f1y7AQA:-N5kjCtx4b3IDdt5dg8y:b-896fix) ([raw file](https://github.com/google/nomulus/blob/22720af9fa0af930ff37a9f3e50d63041f1f2a95/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L84)):*

Suggestion: Ever time we touch  a class that uses member injection, we should switch it to constructor injection. It is considered better practice to use constructor injection as it does not require you to expose otherwise private fields. Also the formatter always wanted to reformat the changed lines, resulting in an ugly file.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 185 at r6](https://reviewable.io/reviews/google/nomulus/1664#-N5bJ1Lx1x4NiFzasnFf:-N5ffeZ_1yWAoi7NED8r:b-896fix) ([raw file](https://github.com/google/nomulus/blob/22720af9fa0af930ff37a9f3e50d63041f1f2a95/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L185)):*

copyOf takes any collection 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 255 at r6](https://reviewable.io/reviews/google/nomulus/1664#-N5bJvVRA6TIWIvEutdI:-N5fYD0J3TKiYHmhIMmd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/22720af9fa0af930ff37a9f3e50d63041f1f2a95/core/src/main/java/google/registry/request/RequestModule.java#L255)):*

A bit nitty, but this and the other one will be better as 

`return extractOptionalHeader(...).map(Integer::parseInt).orElse(0)`

that way you don't have to write out the string 0
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 3 of 6 files reviewed, 8 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 173 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5Fwx_PAXVLnCoL9YR5:-N5usYvf09tzJvhtUSkK:b-82l2x9) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L173)):*

I could not figure out how to set a specific response code, but I added a new action status "TIMED_OUT" that is recorded here.
</blockquote></details>

See https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/Response.java;l=29;bpv=1;bpt=1?q=response.setstatus&ss=nomulus%2Fnomulus

`response.setStatus()` is a common pattern across our codebase with many invocations.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 127 at r7](https://reviewable.io/reviews/google/nomulus/1664#-N5us6anBpXrrhboqfIT:-N5us6anBpXrrhboqfIU:bjpfvsu) ([raw file](https://github.com/google/nomulus/blob/4d32b71841e3c8f9c698f2d90fcdd198b5dbcffb/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L127)):*
> ```Java
>     this.dnsMetrics = dnsMetrics;
>     this.timeout = timeout;
>     this.appEngineRetryCount = appEngineRetryCount;
> ```

I think you can handle the logic here.  Rather than having two member fields, `appEngineRetryCount` and `cloudTasksRetryCount`, and having to choose between them each time one is used, just have one local variable set as follows:

`this.retryCount = cloudTasksRetryCount.orElse(appEngineRetryCount)`

(and see below on comment for making the headers `Optional`)
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 211 at r7](https://reviewable.io/reviews/google/nomulus/1664#-N5utTcf3ik5GQ8Di8gK:-N5utTcf3ik5GQ8Di8gL:b-fgzw67) ([raw file](https://github.com/google/nomulus/blob/4d32b71841e3c8f9c698f2d90fcdd198b5dbcffb/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L211)):*
> ```Java
>       // TODO(b/237302821): Send an email notifying partner the dns update failed
>       recordActionResult(ActionStatus.TIMED_OUT);
>       logger.atSevere().log("Terminated task after too many retries");
> ```

You are missing `.withCause(e)`.  Whenever there is a relevant exception to log, it should always be logged.  As written you are still swallowing the exception and never logging its actual contents.

As a general rule of thumb, if there's a log messages inside a `catch` block it should always log the exception unless the exception is being re-thrown.
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 254 at r7](https://reviewable.io/reviews/google/nomulus/1664#-N5urmtqBBxJzyOpepoL:-N5urmtqBBxJzyOpepoM:b-5sod7y) ([raw file](https://github.com/google/nomulus/blob/4d32b71841e3c8f9c698f2d90fcdd198b5dbcffb/core/src/main/java/google/registry/request/RequestModule.java#L254)):*
> ```Java
>   @Provides
>   @Header(APP_ENGINE_RETRY_HEADER)
>   static int provideAppEngineRetryCount(HttpServletRequest req) {
> ```

Why isn't this returning `Optional<Integer>`?  Same for the one below.  It doesn't make sense to me to default these to 0.

Then at the callsite that uses these it'd simply be `cloudTasksRetryCount.orElse(appEngineRetryCount)`

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 3 of 6 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 173 at r4](https://reviewable.io/reviews/google/nomulus/1664#-N5Fwx_PAXVLnCoL9YR5:-N5vNLqZ8rcNcvAluESP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2fcf73ebfcac1cb42933265a259ad8d7adffe30f/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L173)):*

See https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/Response.java;l=29;bpv=1;bpt=1?q=response.setstatus&ss=nomulus%2Fnomulus

`response.setStatus()` is a common pattern across our codebase with many invocations.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 127 at r7](https://reviewable.io/reviews/google/nomulus/1664#-N5us6anBpXrrhboqfIT:-N5vGVml0FSPNFLvQpY3:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d32b71841e3c8f9c698f2d90fcdd198b5dbcffb/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L127)):*

I think you can handle the logic here.  Rather than having two member fields, `appEngineRetryCount` and `cloudTasksRetryCount`, and having to choose between them each time one is used, just have one local variable set as follows:

`this.retryCount = cloudTasksRetryCount.orElse(appEngineRetryCount)`

(and see below on comment for making the headers `Optional`)
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 211 at r7](https://reviewable.io/reviews/google/nomulus/1664#-N5utTcf3ik5GQ8Di8gK:-N5vNL_aEUqZuEr9c0hF:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d32b71841e3c8f9c698f2d90fcdd198b5dbcffb/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L211)):*

You are missing `.withCause(e)`.  Whenever there is a relevant exception to log, it should always be logged.  As written you are still swallowing the exception and never logging its actual contents.

As a general rule of thumb, if there's a log messages inside a `catch` block it should always log the exception unless the exception is being re-thrown.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 254 at r7](https://reviewable.io/reviews/google/nomulus/1664#-N5urmtqBBxJzyOpepoL:-N5vGZBG3cr5x-Z6RoIY:b-ita26m) ([raw file](https://github.com/google/nomulus/blob/4d32b71841e3c8f9c698f2d90fcdd198b5dbcffb/core/src/main/java/google/registry/request/RequestModule.java#L254)):*

Why isn't this returning `Optional<Integer>`?  Same for the one below.  It doesn't make sense to me to default these to 0.

Then at the callsite that uses these it'd simply be `cloudTasksRetryCount.orElse(appEngineRetryCount)`
</blockquote></details>

Made this Optional<String> cause otherwise it still needs a mapping and default to 0. Moved the type conversion to the action class

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 3 of 6 files reviewed, 4 unresolved discussions (waiting on @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 214 at r8](https://reviewable.io/reviews/google/nomulus/1664#-N6JTX97-qXrspEoRsDo:-N6JTX97-qXrspEoRsDp:ba3ms18) ([raw file](https://github.com/google/nomulus/blob/70f6cca54702b85954959c74036543f0c11014fc/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L214)):*
> ```Java
>       // If we get here, we should terminate this task as it is likely a perpetually failing task.
>       // TODO(b/237302821): Send an email notifying partner the dns update failed
>       recordActionResult(ActionStatus.TIMED_OUT);
> ```

`TIMED_OUT` doesn't quite convey the correct thing to me.  It's not that this task has timed out (Cloud DNS API calls can time out, but that's not what's going on here).  What's happened is that the task has permanently failed after too many retries, which is typically caused by the task itself containing invalid DS data that will cause failures from the Cloud DNS API regardless of how many times it's retried.  This isn't a time out.

So something like `PERMANENT_FAILURE` or `PERMANENT_FAILURE_MAX_RETRIES_ATTEMPTED` seems like it would encapsulate the correct meaning here, and fit in naming-wise with the other existing values of `ActionStatus`.  The latter might be too verbose.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 3 of 6 files reviewed, 4 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 214 at r8](https://reviewable.io/reviews/google/nomulus/1664#-N6JTX97-qXrspEoRsDo:-N6P5sJm1e0s4dBTdKsV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/70f6cca54702b85954959c74036543f0c11014fc/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L214)):*

`TIMED_OUT` doesn't quite convey the correct thing to me.  It's not that this task has timed out (Cloud DNS API calls can time out, but that's not what's going on here).  What's happened is that the task has permanently failed after too many retries, which is typically caused by the task itself containing invalid DS data that will cause failures from the Cloud DNS API regardless of how many times it's retried.  This isn't a time out.

So something like `PERMANENT_FAILURE` or `PERMANENT_FAILURE_MAX_RETRIES_ATTEMPTED` seems like it would encapsulate the correct meaning here, and fit in naming-wise with the other existing values of `ActionStatus`.  The latter might be too verbose.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 255 at r6](https://reviewable.io/reviews/google/nomulus/1664#-N5bJvVRA6TIWIvEutdI:-N6P5Rrt9RlMWjQ__rMt:b-896fix) ([raw file](https://github.com/google/nomulus/blob/22720af9fa0af930ff37a9f3e50d63041f1f2a95/core/src/main/java/google/registry/request/RequestModule.java#L255)):*

If it is done, it hasn't been pushed yet.

And yes I agree with Gus, "retry count" logically makes sense as an integer rather than as an integer stored unparsed in a string, so parse it to the correct type as early as possible (i.e. here) and then use it as an int everywhere else in the codebase.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 6 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 257 at r9](https://reviewable.io/reviews/google/nomulus/1664#-N6PVzvf8l8usUkGzpkG:-N6PVzvf8l8usUkGzpkH:b-r25in5) ([raw file](https://github.com/google/nomulus/blob/077d9b5bb1241645a9b30404a8616881ae0ef7a1/core/src/main/java/google/registry/request/RequestModule.java#L257)):*
> ```Java
>   static Optional<Integer> provideAppEngineRetryCount(HttpServletRequest req) {
>     return extractOptionalHeader(req, APP_ENGINE_RETRY_HEADER)
>         .map(s -> Optional.of(Integer.parseInt(s)))
> ```

You're not using `map()` correctly here.  See documentations and other examples in our codebase.  This can be written more succinctly as:

```
return extractOptionalHeader(req, APP_ENGINE_RETRY_HEADER).map(Integer::parseInt)
```

Note that `map()` is already doing the work of returning an empty `Optional` if the `Optional` it's called on is already empty.



_Code quote:_
```Java
    return extractOptionalHeader(req, APP_ENGINE_RETRY_HEADER)
        .map(s -> Optional.of(Integer.parseInt(s)))
        .orElse(Optional.empty());
```
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 264 at r9](https://reviewable.io/reviews/google/nomulus/1664#-N6PWF33EnINlYn-Ga3-:-N6PWF33EnINlYn-Ga30:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/077d9b5bb1241645a9b30404a8616881ae0ef7a1/core/src/main/java/google/registry/request/RequestModule.java#L264)):*
> ```Java
>   @Header(CLOUD_TASKS_RETRY_HEADER)
>   static Optional<Integer> provideCloudTasksRetryCount(HttpServletRequest req) {
>     return extractOptionalHeader(req, CLOUD_TASKS_RETRY_HEADER)
> ```

Ditto.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 6 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 257 at r9](https://reviewable.io/reviews/google/nomulus/1664#-N6PVzvf8l8usUkGzpkG:-N6PWQ1H0RgKX8XOO8aF:be4ty2y) ([raw file](https://github.com/google/nomulus/blob/077d9b5bb1241645a9b30404a8616881ae0ef7a1/core/src/main/java/google/registry/request/RequestModule.java#L257)):*

You're not using `map()` correctly here.  See documentations and other examples in our codebase.  This can be written more succinctly as:

```
return extractOptionalHeader(req, APP_ENGINE_RETRY_HEADER).map(Integer::parseInt)
```

Note that `map()` is already doing the work of returning an empty `Optional` if the `Optional` it's called on is already empty.
</blockquote></details>

I highly recommend reading the relevant Javadocs in this and all other circumstances: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#map-java.util.function.Function-

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 6 files reviewed, 6 unresolved discussions (waiting on @gbrodman, @jianglai, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 214 at r8](https://reviewable.io/reviews/google/nomulus/1664#-N6JTX97-qXrspEoRsDo:-N6PWpwq4tLz2IwAXtZs:bndhmun) ([raw file](https://github.com/google/nomulus/blob/70f6cca54702b85954959c74036543f0c11014fc/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L214)):*

Done.
</blockquote></details>

Ugh, sorry to come back to this, but I think `MAX_RETRIES_EXCEEDED` is actually the best possible name for this, and the verbiage that is most widely used in our industry.  It's more descriptive than "permanent failure" too.  Naming things is hard.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 2 of 6 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 214 at r8](https://reviewable.io/reviews/google/nomulus/1664#-N6JTX97-qXrspEoRsDo:-N6PZOVXBxLGe3w2X-cU:b-896fix) ([raw file](https://github.com/google/nomulus/blob/70f6cca54702b85954959c74036543f0c11014fc/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L214)):*

Ugh, sorry to come back to this, but I think `MAX_RETRIES_EXCEEDED` is actually the best possible name for this, and the verbiage that is most widely used in our industry.  It's more descriptive than "permanent failure" too.  Naming things is hard.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 257 at r9](https://reviewable.io/reviews/google/nomulus/1664#-N6PVzvf8l8usUkGzpkG:-N6PZAp96rxV-dEXBwKm:b-896fix) ([raw file](https://github.com/google/nomulus/blob/077d9b5bb1241645a9b30404a8616881ae0ef7a1/core/src/main/java/google/registry/request/RequestModule.java#L257)):*

I highly recommend reading the relevant Javadocs in this and all other circumstances: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#map-java.util.function.Function-
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/RequestModule.java` line 264 at r9](https://reviewable.io/reviews/google/nomulus/1664#-N6PWF33EnINlYn-Ga3-:-N6PZBA1EI_xJed62DnP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/077d9b5bb1241645a9b30404a8616881ae0ef7a1/core/src/main/java/google/registry/request/RequestModule.java#L264)):*

Ditto.
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 1 of 4 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 130 at r10](https://reviewable.io/reviews/google/nomulus/1664#-N7GlHCCDaCmtVaIu4ZW:-N7Hd709EdqyEmvjT10B:brj0wad) ([raw file](https://github.com/google/nomulus/blob/87521ef7b7fc1e54732393d27a034ed29ecdb777/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L130)):*

This might cause intended issues if one accidentally sets the cloud tasks retry header before we switch away from app engine. Maybe we should just hardcode to use app engine retry header now. We will have to remember to switch the header to set when the runtime changes, so it's not likely that we'd forget to change where header to look for here.
</blockquote></details>

This is not a header we're using, this is a header that GCP uses.  I don't see what the problem is?  Is the concern that both might be present somehow?

CydeWeys (approved):
:lgtm: once use of `CloudTasksHelper` is addresssed.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664#-:-N7HdXCeC2ywS5h8Z9Uk:bz2gb4x)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 130 at r10](https://reviewable.io/reviews/google/nomulus/1664#-N7GlHCCDaCmtVaIu4ZW:-N7I1djQEb2uTORlne6L:bmgbrw6) ([raw file](https://github.com/google/nomulus/blob/87521ef7b7fc1e54732393d27a034ed29ecdb777/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L130)):*

This is not a header we're using, this is a header that GCP uses.  I don't see what the problem is?  Is the concern that both might be present somehow?
</blockquote></details>

Hmm, I think you are right. I don't believe GCP would set both headers. If we use the Task Queue API, it would only set the app engine header. If we use the Cloud Tasks API, depending on whether we are creating an app engine task, or a generic HTTP task, it should use the corresponding header.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 130 at r10](https://reviewable.io/reviews/google/nomulus/1664#-N7GlHCCDaCmtVaIu4ZW:-N7N2kWwAy4nl4pJmtLx:b1xd6s5) ([raw file](https://github.com/google/nomulus/blob/87521ef7b7fc1e54732393d27a034ed29ecdb777/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L130)):*

Hmm, I think you are right. I don't believe GCP would set both headers. If we use the Task Queue API, it would only set the app engine header. If we use the Cloud Tasks API, depending on whether we are creating an app engine task, or a generic HTTP task, it should use the corresponding header.
</blockquote></details>

Leaving this with handling for both since only 1 should ever be set at one time
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 93 at r10](https://reviewable.io/reviews/google/nomulus/1664#-N7GvlxMEeQjQ99dKdo-:-N7N8fB-CTlcGCgiUzgb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/87521ef7b7fc1e54732393d27a034ed29ecdb777/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L93)):*

Use `CloudTasksHelper` to create and manage fake `CloudTasksUtils`.

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/testing/CloudTasksHelper.java
</blockquote></details>

Done.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @gbrodman, and @jianglai)

___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 93 at r10](https://reviewable.io/reviews/google/nomulus/1664#-N7GvlxMEeQjQ99dKdo-:-N7XWpmi-2sFEc0bwJGQ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/87521ef7b7fc1e54732393d27a034ed29ecdb777/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L93)):*

You don't need the spy any more. Check other test classes that use `CloudTasksHelper`. The helper should provide you with all the verification methods you need. If there are something that the helper cannot do, we should consider augmenting it, instead of using spies.
</blockquote></details>

Done.

jianglai (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1664)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys and @gbrodman)

PR #1665, opened 2022-06-10
Fix EPP update NS delete bug

jianglai:
When removing nameservers during an EPP update, we use the
Sets.difference(), which depends on the equals() method of the set
elements (VKey<HoseResource>), to compute the resulting NS set. The
equality check for a VKey checks for equality of all of its significant fields.

We are seeing strange issues where the removal of NS appeared to
have succeeded, but did not actually happen, because the equality check failed
for two VKey<HostResource> objects with identical SQL key, causing the VKeys to
not be removed by the Set.difference() method when they should have been.

The most likely cause is that the Ofy key field in VKey is no longer up-to-date,
now that we stopped replaying to datastore, and are actively tearing down
Ofy-related pieces.

While I did not check how the ofy keys were different between the VKeys loaded
from as part of the  domain (the existing NSes) and the VKeys loaded directly
from host resource, as specified in the EPP update command (the NSes to delete);
I did verify that by designating the field as @Insignificant, which causes it
to be ignored when checking for equality, the issue disappeared when repeatedly
adding and removing nameservers. Previously I was able to reproduce the issue
by adding/removing nameservers within a few iterations.

Also removed the logging that was put in place to help debug the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1665)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
@@ -50,7 +50,6 @@ import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
-import google.registry.config.RegistryEnvironment;
 import google.registry.dns.DnsQueue;
 import google.registry.flows.EppException;
 import google.registry.flows.ExtensionManager;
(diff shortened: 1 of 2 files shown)


mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1665)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1668, opened 2022-06-13
Convert GenerateZoneFilesAction to SQL

gbrodman:
I'm not 100% sure that this is strictly necessary, but for now we can
replicate the ability to generate zonefiles for any point in time in the
recent past.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1668)

### DIFF
--- a/core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java
+++ b/core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java
@@ -15,27 +15,21 @@
 package google.registry.tools.server;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.common.collect.Iterators.filter;
 import static com.google.common.io.BaseEncoding.base16;
-import static google.registry.mapreduce.inputs.EppResourceInputs.createEntityInput;
 import static google.registry.model.EppResourceUtils.loadAtPointInTime;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.Action.Method.POST;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.joda.time.DateTimeZone.UTC;
 
-import com.google.appengine.tools.mapreduce.Mapper;
-import com.google.appengine.tools.mapreduce.Reducer;
-import com.google.appengine.tools.mapreduce.ReducerInput;
 import com.google.cloud.storage.BlobId;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.gcs.GcsUtils;
 import google.registry.mapreduce.MapreduceRunner;
-import google.registry.mapreduce.inputs.NullInput;
-import google.registry.model.EppResource;
 import google.registry.model.domain.DomainBase;
 import google.registry.model.domain.secdns.DelegationSignerData;
 import google.registry.model.host.HostResource;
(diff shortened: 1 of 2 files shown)


mindhog (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1668)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java` line 143 at r1](https://reviewable.io/reviews/google/nomulus/1668#-N4X8Pzv7q5f2qb53P59:-N4X8Pzw9ThSeA4kCWkj:b6jxxkf) ([raw file](https://github.com/google/nomulus/blob/3c5d5d66c5574c80aad4ba4ff0ff5f13a61ee734/core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java#L143)):*
> ```Java
>   private void generateForTld(String tld, DateTime exportTime) {
>     ImmutableList<String> stanzas = jpaTm().transact(() -> getStanzasForTld(tld, exportTime));
>     BlobId filename = BlobId.of(bucket, String.format(FILENAME_FORMAT, tld, exportTime));
> ```

`fileId` or even `blobId` would be a better name for this (the file name is what you've String.formatted). 
___
*[`core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java` line 196 at r1](https://reviewable.io/reviews/google/nomulus/1668#-N4XBan60ZnoTdd1nBfp:-N4XBan60ZnoTdd1nBfq:b-j7ax6k) ([raw file](https://github.com/google/nomulus/blob/3c5d5d66c5574c80aad4ba4ff0ff5f13a61ee734/core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java#L196)):*
> ```Java
>       for (HostResource unprojectedHost : tm().loadByKeys(domain.getNameservers()).values()) {
>         HostResource host = loadAtPointInTime(unprojectedHost, exportTime);
>         // A null means the host was deleted (or not created) at this time.
> ```

How can this happen?  If we're loading the domain at a given point in time and we're loading the subordinate hosts at the same point in time, shouldn't they always be present?  If so, it seems like the null host case should be an error.
Unless there's still something non-transactional about nameserver updates?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1668)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

___
*[`core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java` line 143 at r1](https://reviewable.io/reviews/google/nomulus/1668#-N4X8Pzv7q5f2qb53P59:-N4Y86-DC7H3aBEqagID:b-rvk1e0) ([raw file](https://github.com/google/nomulus/blob/3c5d5d66c5574c80aad4ba4ff0ff5f13a61ee734/core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java#L143)):*

`fileId` or even `blobId` would be a better name for this (the file name is what you've String.formatted). 
</blockquote></details>

yeah sure (just copied what was there before) 
___
*[`core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java` line 196 at r1](https://reviewable.io/reviews/google/nomulus/1668#-N4XBan60ZnoTdd1nBfp:-N4YJfl85zJ5rTIrcGx6:buy75qx) ([raw file](https://github.com/google/nomulus/blob/3c5d5d66c5574c80aad4ba4ff0ff5f13a61ee734/core/src/main/java/google/registry/tools/server/GenerateZoneFilesAction.java#L196)):*

How can this happen?  If we're loading the domain at a given point in time and we're loading the subordinate hosts at the same point in time, shouldn't they always be present?  If so, it seems like the null host case should be an error.
Unless there's still something non-transactional about nameserver updates?
</blockquote></details>

Maybe this is related to one of the tasks that previously used to be asynchronous but is no longer? I don't know. This, like the other area, was just copied from the previously-existing code :P

mindhog (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1668)* status: all files reviewed, 2 unresolved discussions (waiting on @mindhog)

PR #1672, opened 2022-06-17
Remove Ofy support from Cursor

jianglai:
Cursor was originally envisioned to support arbitary ImmutableObject
scopes. However, in practice only the Registry scope is used. The SQL
representation of Cursor assumes that and the schema uses a composite ID
with a string column for the primary key of the scope object. Without a
schema migration to persist the VKey of the scope, we cannot support any
ImmutableObject other than those with a primitive string primary key.

Given the complexity involved and the limited use case, the scope is now
explictly limited to Registry only.

Also removed mapreduces that depends on Ofy keys of Cursors, and made
some code quality improvement based on IntelliJ suggestions on modified
files.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1672)

### DIFF
--- a/core/src/main/java/google/registry/beam/rde/RdeIO.java
+++ b/core/src/main/java/google/registry/beam/rde/RdeIO.java
@@ -279,7 +279,7 @@ public class RdeIO {
                     transactIfJpaTm(
                         () ->
                             tm().loadByKeyIfPresent(
-                                    Cursor.createVKey(key.cursor(), registry.getTldStr())));
+                                    Cursor.createScopedVKey(key.cursor(), registry)));
                 DateTime position = getCursorTimeOrStartOfTime(cursor);
                 checkState(key.interval() != null, "Interval must be present");
                 DateTime newPosition = key.watermark().plus(key.interval());
@@ -292,7 +292,7 @@ public class RdeIO {
                     "Partial ordering of RDE deposits broken: %s %s",
                     position,
                     key);
-                tm().put(Cursor.create(key.cursor(), newPosition, registry));
+                tm().put(Cursor.createScoped(key.cursor(), newPosition, registry));
                 logger.atInfo().log(
                     "Rolled forward %s on %s cursor to %s.", key.cursor(), key.tld(), newPosition);
                 RdeRevision.saveRevision(key.tld(), key.watermark(), key.mode(), input.getValue());
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -27,7 +27,6 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Ascii;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
-import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedMap;
@@ -42,7 +41,9 @@ import java.net.URL;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.inject.Named;
@@ -1266,7 +1267,7 @@ public final class RegistryConfig {
                       e.getKey().equals("START_OF_TIME")
                           ? START_OF_TIME
                           : DateTime.parse(e.getKey()),
-                  e -> e.getValue()));
+                  Entry::getValue));
     }
 
     @Provides
--- a/core/src/main/java/google/registry/model/EntityClasses.java
+++ b/core/src/main/java/google/registry/model/EntityClasses.java
@@ -17,7 +17,6 @@ package google.registry.model;
 import com.google.common.collect.ImmutableSet;
 import google.registry.model.annotations.DeleteAfterMigration;
 import google.registry.model.billing.BillingEvent;
-import google.registry.model.common.Cursor;
 import google.registry.model.common.EntityGroupRoot;
 import google.registry.model.common.GaeUserIdConverter;
 import google.registry.model.contact.ContactHistory;
@@ -54,7 +53,6 @@ public final class EntityClasses {
           BillingEvent.Recurring.class,
           ContactHistory.class,
           ContactResource.class,
-          Cursor.class,
           DomainBase.class,
           DomainHistory.class,
           EntityGroupRoot.class,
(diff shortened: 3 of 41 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/model/common/Cursor.java` line 139 at r4](https://reviewable.io/reviews/google/nomulus/1672#-N5epb5Q8YCxS2psliEZ:-N5epb5Q8YCxS2psliE_:b8z6if9) ([raw file](https://github.com/google/nomulus/blob/2a91406ee7b780ea28dde8e1159e3b5e2c39ec0b/core/src/main/java/google/registry/model/common/Cursor.java#L139)):*
> ```Java
>   }
> 
>   public static VKey<Cursor> createVKey(CursorType type, String scope) {
> ```

I suggest we 
1. Let createGlobalVkey do its own implementation.
2. Change 'scope' parameter to tld
3. Overload this method with one that takes Registry as parameter.

We may consider drop 2 and just do 3, since there are only two production cases that do not have registry object at call site. then we can get rid of the Registry.get() call.



_Code quote:_
```Java
public static VKey<Cursor> createVKey(CursorType type, String scope)
```

jianglai (commented):
> Reviewed 4 of 16 files at r1, 12 of 12 files at r2, 1 of 1 files at r3, 4 of 4 files at r4, 29 of 29 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[`core/src/main/java/google/registry/model/common/Cursor.java` line 139 at r4](https://reviewable.io/reviews/google/nomulus/1672#-N5epb5Q8YCxS2psliEZ:-N5fH5PL5TvULkhA4W0B:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2a91406ee7b780ea28dde8e1159e3b5e2c39ec0b/core/src/main/java/google/registry/model/common/Cursor.java#L139)):*

I suggest we 
1. Let createGlobalVkey do its own implementation.
2. Change 'scope' parameter to tld
3. Overload this method with one that takes Registry as parameter.

We may consider drop 2 and just do 3, since there are only two production cases that do not have registry object at call site. then we can get rid of the Registry.get() call.



</blockquote></details>

Done.

weiminyu (approved):
> Reviewed 3 of 16 files at r1, 3 of 12 files at r2, 4 of 4 files at r4, 28 of 29 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 3 of 16 files at r1, 12 of 12 files at r2, 4 of 4 files at r4, 28 of 29 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/EppResource.java` line 60 at r6](https://reviewable.io/reviews/google/nomulus/1672#-N5fM1Q-5sEkQRf79DAS:-N5fM1Q06WF7iSgnkMRV:b-en677y) ([raw file](https://github.com/google/nomulus/blob/54adc89ee68c402b8ee7561ecc4d608e45a2d1ec/core/src/main/java/google/registry/model/EppResource.java#L60)):*
> ```Java
> public abstract class EppResource extends BackupGroupRoot implements Buildable {
> 
>   private static final long serialVersionUID = -252782773382339534L;
> ```

Out of curiosity, why add the serial version IDs?
___
*[`core/src/main/java/google/registry/model/EppResource.java` line 204 at r6](https://reviewable.io/reviews/google/nomulus/1672#-N5fLzUIEeXiqp4TC9b5:-N5fLzUIEeXiqp4TC9b6:bgly59r) ([raw file](https://github.com/google/nomulus/blob/54adc89ee68c402b8ee7561ecc4d608e45a2d1ec/core/src/main/java/google/registry/model/EppResource.java#L204)):*
> ```Java
> 
>   /** Create the VKey for the specified EPP resource. */
>   @Override
> ```

Weird that this didn't have an override annotation before, right? 
___
*[`core/src/main/java/google/registry/model/common/Cursor.java` line 113 at r6](https://reviewable.io/reviews/google/nomulus/1672#-N5fMMyAECudh3-_JPsJ:-N5fMMyAECudh3-_JPsK:b-sxm0rz) ([raw file](https://github.com/google/nomulus/blob/63e4f4f10a455f1f1a526287505cd3d64684a2d1/core/src/main/java/google/registry/model/common/Cursor.java#L113)):*
> ```Java
>     /** Cursor for tracking monthly uploads of MANIFEST.txt to ICANN. No longer used. */
>     @Deprecated
>     ICANN_UPLOAD_MANIFEST(EntityGroupRoot.class);
> ```

Have we verified that this doesn't exist in the SQL database? Just double-checking
___
*[`core/src/main/java/google/registry/rde/RdeUploadAction.java` line 181 at r6](https://reviewable.io/reviews/google/nomulus/1672#-N5fL5d76VJRhO6qz8o3:-N5fL5d76VJRhO6qz8o4:b-wsa38s) ([raw file](https://github.com/google/nomulus/blob/54adc89ee68c402b8ee7561ecc4d608e45a2d1ec/core/src/main/java/google/registry/rde/RdeUploadAction.java#L181)):*
> ```Java
>                 () ->
>                     tm().loadByKeyIfPresent(
>                             Cursor.createScopedVKey(RDE_UPLOAD_SFTP, Registry.get(tld))))
> ```

nitty but we only need to call Registry.get() once in this method

gbrodman (approved):
> Dismissed @jianglai from a discussion.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/rde/RdeUploadAction.java` line 181 at r6](https://reviewable.io/reviews/google/nomulus/1672#-N5fL5d76VJRhO6qz8o3:-N5fOLiL5yTqyUOAaO_k:b-tlfcga) ([raw file](https://github.com/google/nomulus/blob/54adc89ee68c402b8ee7561ecc4d608e45a2d1ec/core/src/main/java/google/registry/rde/RdeUploadAction.java#L181)):*

Yes, but `Registry.get()` is cached, so calling it multiple times doesn't actually matter :D
</blockquote></details>

hah fair enough

gbrodman (approved):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1672)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #1695, opened 2022-07-01
Add a registration_behavior column to AllocationToken

gbrodman:
This is, as of now, unused but we can use it for b/237683906 and
b/237800445 in the future to allow for special behavior dictated by
allocation tokens rather than having to reserve specific domains.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1695)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -86,6 +86,22 @@ public class AllocationToken extends BackupGroupRoot implements Buildable {
           .putAll(VALID, ENDED, CANCELLED)
           .build();
 
+  /** Any special behavior that should be used when registering domains using this token. */
+  public enum RegistrationBehavior {
+    /** No special behavior */
+    DEFAULT,
+    /**
+     * Bypasses the TLD state check, e.g. allowing registration during QUIET_PERIOD.
+     *
+     * <p>NB: while this means that, for instance, one can register non-trademarked domains in the
+     * sunrise period, any trademarked-domain registrations in the sunrise period must still include
+     * the proper signed marks. In other words, this only bypasses the TLD state check.
+     */
+    BYPASS_TLD_STATE,
+    /** Bypasses most checks and creates the domain as an anchor tenant, with all that implies. */
+    ANCHOR_TENANT
+  }
+
   /** Single-use tokens are invalid after use. Infinite-use tokens, predictably, are not. */
   public enum TokenType {
     SINGLE_USE,
@@ -152,6 +168,10 @@ public class AllocationToken extends BackupGroupRoot implements Buildable {
   @Enumerated(EnumType.STRING)
   @Column(name = "renewalPriceBehavior", nullable = false)
   RenewalPriceBehavior renewalPriceBehavior = RenewalPriceBehavior.DEFAULT;
+
+  @Enumerated(EnumType.STRING)
+  @Column(nullable = false)
+  RegistrationBehavior registrationBehavior = RegistrationBehavior.DEFAULT;
 
   // TODO: Remove onLoad once all allocation tokens are migrated to have a discountYears of 1.
   @OnLoad
@@ -225,6 +245,10 @@ public class AllocationToken extends BackupGroupRoot implements Buildable {
     return renewalPriceBehavior;
   }
 
+  public RegistrationBehavior getRegistrationBehavior() {
+    return registrationBehavior;
+  }
+
   @Override
   public VKey<AllocationToken> createVKey() {
     return VKey.create(AllocationToken.class, getToken(), Key.create(this));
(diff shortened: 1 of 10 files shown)


gbrodman:
oops, need to do the sql changes only first

gbrodman:
this should be good to go hopefully

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1695)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 95 at r1](https://reviewable.io/reviews/google/nomulus/1695#-N7N8dnI-7m6GKxcyS4C:-N7N8dnI-7m6GKxcyS4D:bt6qe4t) ([raw file](https://github.com/google/nomulus/blob/da21f99a5d84fb9281f59311d2b7e53487400a12/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L95)):*
> ```Java
>     /** No special behavior */
>     DEFAULT,
>     /** Bypasses the TLD state check, e.g. allowing registration during QUIET_PERIOD. */
> ```

Add a note to the effect that, while this also allows registration of non-TMed domains during sunrise, it does not bypass the requirement that names in the TMDB must have associated signed marks being sent as part of the create.

And of course ensure that this is actually encoded into the logic correctly when implementing the rest of this functionality.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1695#-N7N8MRxDWv5I7W6We9Y:-N7N8MRxDWv5I7W6We9Z:b80t5yp) ([raw file](https://github.com/google/nomulus/blob/da21f99a5d84fb9281f59311d2b7e53487400a12/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L97)):*
> ```Java
>     /** Bypasses the TLD state check, e.g. allowing registration during QUIET_PERIOD. */
>     BYPASS_TLD_STATE,
>     /** Bypasses most checks for anchor tenants. TODO(b/237800445) implement this. */
> ```

I think this is not quite accurate.  Not only does it bypass most checks, it also has price implications, right?  Anchor tenant domains should be $0 for the first 2 years and then renew at the base price thereafter.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1695)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 95 at r1](https://reviewable.io/reviews/google/nomulus/1695#-N7N8dnI-7m6GKxcyS4C:-N7NAR3MEhHOe7kcaT1y:bdwbr2g) ([raw file](https://github.com/google/nomulus/blob/da21f99a5d84fb9281f59311d2b7e53487400a12/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L95)):*

Add a note to the effect that, while this also allows registration of non-TMed domains during sunrise, it does not bypass the requirement that names in the TMDB must have associated signed marks being sent as part of the create.

And of course ensure that this is actually encoded into the logic correctly when implementing the rest of this functionality.
</blockquote></details>

Sure (I already have tests for this scenario in the branch that actually implements this logic) 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1695#-N7N8MRxDWv5I7W6We9Y:-N7NAWlSDFPLHgvSMJ1o:bbk7vfx) ([raw file](https://github.com/google/nomulus/blob/da21f99a5d84fb9281f59311d2b7e53487400a12/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L97)):*

I think this is not quite accurate.  Not only does it bypass most checks, it also has price implications, right?  Anchor tenant domains should be $0 for the first 2 years and then renew at the base price thereafter.
</blockquote></details>

Yeah, more generally it's just treated as an anchor tenant in every way.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1695)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 99 at r2](https://reviewable.io/reviews/google/nomulus/1695#-N7NBNhx4H7SM_UB81gR:-N7NCK6W93g20dtr9vxb:bu2eumt) ([raw file](https://github.com/google/nomulus/blob/6bf3ba326aae9a2d057466981b57e378befe9448/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L99)):*

```
any trademarked-domain registrations must (in all periods) still include the proper signed marks
```

This isn't accurate.  Signed marks are _only_ required (or even allowed) during sunrise.  During all other phases, sending a signed mark is an error.  Now it is true that during the claims period you need to send a claims key, but that's a different beast entirely.  And it's also true that we happen to use indefinite claims period by policy for all of our launched open TLDs, but this isn't enforced in the code; it's configurable through `nomulus update_tld` to not do so (and we're leaving ourselves the option of not doing so in the future).
</blockquote></details>

ugh yeah i mixed it up with the claims period

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1695#-:-N7NDf13ED3BVK7nBQbY:bnfp4nl)* status: 0 of 10 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1696, opened 2022-07-01
Deal with some getOfyKey() references

mindhog:
Now that ofy keys aren't necessarily being restored, it seemed prudent to
audit existing uses to verify that we aren't relying on any keys that may now
be null.

This fixes one case that appeared to be potentially problematic (in
ResourceFlowUtils), removes a few methods that call getOfyKey() but are no
longer used, and adds comments to one use of the key that appears to be safe
after visual inspection.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1696)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1696)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mindhog)

PR #1707, opened 2022-07-18
Add base object classes for new user/role permissioning model

gbrodman:
- Adds the permissions themselves
- Adds the six roles that a user may have -- three global, three
  per-registrar
- Adds the mapping from role -> set of permissions
- Adds a UserRoles object to encapsulate the answer to the question of
  "does this user have this permission?"
- Adds a User class as a base to show how we will use the new UserRoles
  object

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1707)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1707)* status: all files reviewed, 13 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 18 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdkDf7cNLaG0eMEV3:-N8OdkDf7cNLaG0eMEV4:bvw4ltn) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L18)):*
> ```Java
> 
> /** Permissions that users may have in the UI, either per-registrar or globally. */
> public enum ConsolePermission {
> ```

The list of permissions seems too unwieldy. Is it better to introduce some sort of hierarchy for them to be more organized? Relevant permissions can be grouped together into static nested classes (can they?) or named accordingly for better discoverability.

Honestly I don't have a concrete idea of how to make them more presentable, but I just feel like a flat structure like this is probably not the best way. At least these permissions can be classified as either for registrar or for registry, I think? Or maybe some of them are only for registry, whereas some others can be for both.



_Code quote:_
```Java
public enum ConsolePermission {
```
___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 22 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdA3J-HCyP6zBHKBG:-N8OdA3J-HCyP6zBHKBH:b-q0r328) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L22)):*
> ```Java
>   MANAGE_USERS,
>   /** Add, update, or remove registrars. */
>   MANAGE_REGISTRARS,
> ```

I'm not sure what exactly is the difference between these two roles.



_Code quote:_
```Java
  /** Add, update, or remove registrars. */
  MANAGE_REGISTRARS,
  /** Manage related registrars (possibly TBD). */
  MANAGE_ACCREDITATION,
```
___
*[`core/src/main/java/google/registry/model/console/ConsoleRoleDefinitions.java` line 25 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OmvEiDxA1XarWT6xy:-N8OmvEiDxA1XarWT6xz:b-j9i2l2) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsoleRoleDefinitions.java#L25)):*
> ```Java
>  * where each role has all the permissions of the role below it plus some others.
>  */
> public class ConsoleRoleDefinitions {
> ```

Add a private constructor to signal that this is a utility class that cannot be instantiated.



_Code quote:_
```Java
public class ConsoleRoleDefinitions {
```
___
*[`core/src/main/java/google/registry/model/console/User.java` line 29 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OoqDWD54PafYVJxGU:-N8OoqDWD54PafYVJxGV:b-n07dq5) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L29)):*
> ```Java
> 
> /** A console user, either a registry employee or a registrar partner. */
> public class User extends ImmutableObject implements Buildable {
> ```

Is this class supposed to be an entity? If so, are you waiting to add the table before making it an `@Entity`?



_Code quote:_
```Java
public class User extends ImmutableObject implements Buildable 
```
___
*[`core/src/main/java/google/registry/model/console/User.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OpF_U6uTXobFBKDsD:-N8OpF_U6uTXobFBKDsE:b-976e2h) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L47)):*
> ```Java
>    * console. This will be set to false on contact creation and when the user sets a password.
>    */
>   boolean allowedToSetRegistryLockPassword = false;
> ```

Why is this a separate field instead of being one of the permissions? Is this value supposed to be persisted in the database?



_Code quote:_
```Java
allowedToSetRegistryLockPassword 
```
___
*[`core/src/main/java/google/registry/model/console/User.java` line 127 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OwMtr5ZzzFPiDDwSn:-N8OwMtr5ZzzFPiDDwSo:b-tydzor) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L127)):*
> ```Java
> 
>     public Builder setAllowedToSetRegistryLockPassword(boolean allowedToSetRegistryLockPassword) {
>       if (allowedToSetRegistryLockPassword) {
> ```

I don't understand why setting `allowedToSetRegistryLockPassword` should re-set the saved password (and hash).

It looks like this value is always set to true before setting the password, which is then automatically reverted to false after the password. What exactly is the use case here? Does it just mean that one should always call this method prior to setting the password? Is it more like a "unlockRegistryLockPassword" method?



_Code quote:_
```Java
      if (allowedToSetRegistryLockPassword) {
        getInstance().registryLockPasswordSalt = null;
        getInstance().registryLockPasswordHash = null;
      }
```
___
*[`core/src/main/java/google/registry/model/console/UserRoles.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8Oz8-vD-RSDPxE_wZq:-N8Oz8-vD-RSDPxE_wZr:b-ok6i8d) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/UserRoles.java#L26)):*
> ```Java
> 
> /** Contains the global and per-registrar roles for a given user. */
> public class UserRoles extends ImmutableObject implements Buildable {
> ```

Can this be an `AutoValue`? There's no Id to allocate so we don't actually need `Buildable`.

Also, how is this class going to be saved in the database? An an embedded entity spanning multiple columns? I know there a design doc for this but I don't recall if all the details are ironed out. We should also link to the design doc in the javadoc.



_Code quote:_
```Java
public class UserRoles extends ImmutableObject implements Buildable
```
___
*[`core/src/main/java/google/registry/model/console/UserRoles.java` line 31 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OyVTQ0HyMuFhLukvT:-N8OyVTQ0HyMuFhLukvU:b-utu2eg) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/UserRoles.java#L31)):*
> ```Java
>   private boolean isAdmin = false;
> 
>   /** The global role (e.g. support agent) that the user has across all registrars. */
> ```

use `{@link SupportAgent}`



_Code quote:_
```Java
(e.g. support agent)
```
___
*[`core/src/test/java/google/registry/model/console/ConsoleRoleDefinitionsTest.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P1SUo27Tlxo5_c1o9:-N8P1SUo27Tlxo5_c1oA:bl3ujv4) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/ConsoleRoleDefinitionsTest.java#L26)):*
> ```Java
>   @Test
>   void testHierarchicalPermissions_registry() {
>     // Note: we can't use Truth's IterableSubject to check all the subset/superset restrictions
> ```

Might be better to explicitly state why.



_Code quote:_
```Java
// Note: we can't use Truth's IterableSubject to check all the subset/superset restrictions
```
___
*[`core/src/test/java/google/registry/model/console/UserRolesTest.java` line 39 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P3Nmn8JeoIhU41XYV:-N8P3Nmn8JeoIhU41XYW:b-sii2k0) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserRolesTest.java#L39)):*
> ```Java
>   void testAdmin_overridesAll() {
>     UserRoles userRoles = new UserRoles.Builder().setIsAdmin(true).build();
>     assertThat(userRoles.hasGlobalPermission(ConsolePermission.MANAGE_USERS)).isTrue();
> ```

Shouldn't we loop over all the `ConsolePermission`s like we did in the previous test?



_Code quote:_
```Java
assertThat(userRoles.hasGlobalPermission(ConsolePermission.MANAGE_USERS)).isTrue();
```
___
*[`core/src/test/java/google/registry/model/console/UserRolesTest.java` line 46 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P4-Ut7GZ6hiLqWB2_:-N8P4-Ut7GZ6hiLqWB2a:b904o97) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserRolesTest.java#L46)):*
> ```Java
>   void testRegistrarPermission_withGlobal() {
>     UserRoles userRoles = new UserRoles.Builder().setGlobalRole(GlobalRole.FTE).build();
>     assertThat(userRoles.hasGlobalPermission(ConsolePermission.MANAGE_USERS)).isTrue();
> ```

Same question about looping over the permissions.



_Code quote:_
```Java
assertThat(userRoles.hasGlobalPermission(ConsolePermission.MANAGE_USERS)).isTrue();
```
___
*[`core/src/test/java/google/registry/model/console/UserRolesTest.java` line 58 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P4ChE4KQ446Osrsol:-N8P4ChE4KQ446Osrsom:b-x9lugr) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserRolesTest.java#L58)):*
> ```Java
>             .setRegistrarRoles(ImmutableMap.of("TheRegistrar", RegistrarRole.PRIMARY_CONTACT))
>             .build();
>     assertThat(userRoles.hasPermission("TheRegistrar", ConsolePermission.MANAGE_USERS)).isTrue();
> ```

We should also test that this user role does not grant to "TheRegistrar" permissions not covered by `MANAGED_UERS`.



_Code quote:_
```Java
assertThat(userRoles.hasPermission("TheRegistrar", ConsolePermission.MANAGE_USERS)).isTrue();
```
___
*[`core/src/test/java/google/registry/model/console/UserTest.java` line 27 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P61UJ8TlLMlkuQbBR:-N8P61UK4Z-8ffNdl5zd:b-kauhyo) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserTest.java#L27)):*
> ```Java
>   void testFailure_badInputs() {
>     User.Builder builder = new User.Builder();
>     assertThrows(IllegalArgumentException.class, () -> builder.setGaiaId(null));
> ```

Should assert on the error messages as well.



_Code quote:_
```Java
    assertThrows(IllegalArgumentException.class, () -> builder.setGaiaId(null));
    assertThrows(IllegalArgumentException.class, () -> builder.setEmailAddress(""));
    assertThrows(NullPointerException.class, () -> builder.setEmailAddress(null));
    assertThrows(IllegalArgumentException.class, () -> builder.setEmailAddress(""));
    assertThrows(IllegalArgumentException.class, () -> builder.setUserRoles(null));
    assertThrows(IllegalArgumentException.class, builder::build);
    builder.setGaiaId("gaiaId");
    assertThrows(IllegalArgumentException.class, builder::build);
    assertThrows(IllegalArgumentException.class, () -> builder.setEmailAddress("invalidEmail"));
    builder.setEmailAddress("email@email.com");
    assertThrows(IllegalArgumentException.class, builder::build);
    builder.setUserRoles(new UserRoles.Builder().build());
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1707)* status: all files reviewed, 13 unresolved discussions (waiting on @Embedded and @jianglai)

___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 18 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdkDf7cNLaG0eMEV3:-N8UnuNiEiRNIDI06qen:b-4ohlb0) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L18)):*

The list of permissions seems too unwieldy. Is it better to introduce some sort of hierarchy for them to be more organized? Relevant permissions can be grouped together into static nested classes (can they?) or named accordingly for better discoverability.

Honestly I don't have a concrete idea of how to make them more presentable, but I just feel like a flat structure like this is probably not the best way. At least these permissions can be classified as either for registrar or for registry, I think? Or maybe some of them are only for registry, whereas some others can be for both.
</blockquote></details>

1. Unfortunately it doesn't seem like there's any strict hierarchy that cuts across both registrar and registry users. You can see the kinda scattershot graph in the appendix of https://docs.google.com/document/d/1_RLjUP5_e8EpTcs8AZzzdaEFlf6dUvoE-ovCa-OE2Ks/edit?resourcekey=0-F232AvNAbmc4VEzUM8Dg-w  (which comes from the BRD)

2. yeah, some are global and some are either global or per-registrar. It doesn't seem like that affects the actual structure though. 
___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 22 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdA3J-HCyP6zBHKBG:-N8UkZKg8btogIdvQINH:b3upq3y) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L22)):*

I'm not sure what exactly is the difference between these two roles.
</blockquote></details>

MANAGE_REGISTRARS is sort of the overarching registrar-management role. MANAGE_ACCREDITATION is apparently related to managing how one registrar is related to another, e.g. GoDaddy and one of the registrars that they've acquired. 
___
*[`core/src/main/java/google/registry/model/console/ConsoleRoleDefinitions.java` line 25 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OmvEiDxA1XarWT6xy:-N8UlZGe3fuDTakKGty_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsoleRoleDefinitions.java#L25)):*

Add a private constructor to signal that this is a utility class that cannot be instantiated.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/console/User.java` line 29 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OoqDWD54PafYVJxGU:-N8UpBzg40zw7hiClAum:b55t8a8) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L29)):*

Is this class supposed to be an entity? If so, are you waiting to add the table before making it an `@Entity`?
</blockquote></details>

Yep. I figured we should finalize the structure before creating tables (then, after the tables are created, we can make this an Entity). 
___
*[`core/src/main/java/google/registry/model/console/User.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OpF_U6uTXobFBKDsD:-N8UldUy1zz8yc0MbobX:b-y6ohb0) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L47)):*

Why is this a separate field instead of being one of the permissions? Is this value supposed to be persisted in the database?
</blockquote></details>

yeah, this whole paradigm is copied from the existing RegistrarPoc stuff. Basically, it's a temporary permission that someone powerful can give to a less-powerful user that allows that user to set their own registry lock password in the UI. We do this because the powerful user shouldn't be the one setting the lesser user's password (since they shouldn't know it). 
___
*[`core/src/main/java/google/registry/model/console/User.java` line 127 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OwMtr5ZzzFPiDDwSn:-N8Um7KmCudh_DmBeoE1:b-6hrhxb) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L127)):*

I don't understand why setting `allowedToSetRegistryLockPassword` should re-set the saved password (and hash).

It looks like this value is always set to true before setting the password, which is then automatically reverted to false after the password. What exactly is the use case here? Does it just mean that one should always call this method prior to setting the password? Is it more like a "unlockRegistryLockPassword" method?
</blockquote></details>

yeah, see the other comment -- this is so that powerful users can essentially reset lesser users' passwords without the powerful user actually knowing the password 
___
*[`core/src/main/java/google/registry/model/console/UserRoles.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8Oz8-vD-RSDPxE_wZq:-N8UpivPC_TF9-l6rWXM:b-uqjoie) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/UserRoles.java#L26)):*

Can this be an `AutoValue`? There's no Id to allocate so we don't actually need `Buildable`.

Also, how is this class going to be saved in the database? An an embedded entity spanning multiple columns? I know there a design doc for this but I don't recall if all the details are ironed out. We should also link to the design doc in the javadoc.
</blockquote></details>

It'll be @Embedded. 

and nah we want this to be buildable so we can do some checks when building it (the same as all the other @Embedded entities) 

added link
___
*[`core/src/main/java/google/registry/model/console/UserRoles.java` line 31 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OyVTQ0HyMuFhLukvT:-N8Uq2dA62qcLkqGMcVn:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/UserRoles.java#L31)):*

use `{@link SupportAgent}`
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/console/ConsoleRoleDefinitionsTest.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P1SUo27Tlxo5_c1o9:-N8UnEnfA9ccyuEauZMS:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/ConsoleRoleDefinitionsTest.java#L26)):*

Might be better to explicitly state why.
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/console/UserRolesTest.java` line 39 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P3Nmn8JeoIhU41XYV:-N8UrTvQClT59YfYepVu:b-e8513m) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserRolesTest.java#L39)):*

Shouldn't we loop over all the `ConsolePermission`s like we did in the previous test?
</blockquote></details>

good point 
___
*[`core/src/test/java/google/registry/model/console/UserRolesTest.java` line 46 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P4-Ut7GZ6hiLqWB2_:-N8UrdWyCUDFGXXfb_pf:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserRolesTest.java#L46)):*

Same question about looping over the permissions.
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/console/UserRolesTest.java` line 58 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P4ChE4KQ446Osrsol:-N8UsTIgDIV2qM_bG0RN:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserRolesTest.java#L58)):*

We should also test that this user role does not grant to "TheRegistrar" permissions not covered by `MANAGED_UERS`.
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/console/UserTest.java` line 27 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8P61UJ8TlLMlkuQbBR:-N8Utm9U9PXkz_pmFcC8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/test/java/google/registry/model/console/UserTest.java#L27)):*

Should assert on the error messages as well.
</blockquote></details>

Done.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1707)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 18 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdkDf7cNLaG0eMEV3:-N8Z8vca8byAmga9pyyx:b-swbxv8) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L18)):*

1. Unfortunately it doesn't seem like there's any strict hierarchy that cuts across both registrar and registry users. You can see the kinda scattershot graph in the appendix of https://docs.google.com/document/d/1_RLjUP5_e8EpTcs8AZzzdaEFlf6dUvoE-ovCa-OE2Ks/edit?resourcekey=0-F232AvNAbmc4VEzUM8Dg-w  (which comes from the BRD)

2. yeah, some are global and some are either global or per-registrar. It doesn't seem like that affects the actual structure though. 
</blockquote></details>

Well, like I said I can't think of a better way, but I still feel this will get even messier over time :D
___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 22 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdA3J-HCyP6zBHKBG:-N8Z8kwlARE_U0zS0c_1:bfv1ctp) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L22)):*

MANAGE_REGISTRARS is sort of the overarching registrar-management role. MANAGE_ACCREDITATION is apparently related to managing how one registrar is related to another, e.g. GoDaddy and one of the registrars that they've acquired. 
</blockquote></details>

Ah, that make sense. Better to include this explanation in the javadoc.
___
*[`core/src/main/java/google/registry/model/console/User.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OpF_U6uTXobFBKDsD:-N8Z6VWc4Gbju_LCFMD0:b-4zwbhb) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L47)):*

yeah, this whole paradigm is copied from the existing RegistrarPoc stuff. Basically, it's a temporary permission that someone powerful can give to a less-powerful user that allows that user to set their own registry lock password in the UI. We do this because the powerful user shouldn't be the one setting the lesser user's password (since they shouldn't know it). 
</blockquote></details>

If the lesser user is the one whose's supposed to set the password, why should they be granted special permission to do so? Shouldn't they be able to do it all the time?

The separation of privileges seems weird here, can you give a more concrete example of who these two types of users should be?
___
*[`core/src/main/java/google/registry/model/console/User.java` line 127 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OwMtr5ZzzFPiDDwSn:-N8Z7IjF-QRjmg3uWjcr:b3jg6fz) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L127)):*

yeah, see the other comment -- this is so that powerful users can essentially reset lesser users' passwords without the powerful user actually knowing the password 
</blockquote></details>

The behavior should definitely be documented in javadoc.

Also, if we want the powerful user to reset the registry lock password (this is not the lesser user's password, if I understand correctly, this is per registrar/registry pair password, regardless of how many users a registrar might have in our system), we could just have a reset function. Allowing the lesser user to reset the password and resetting the password itself seems to be two orthogonal functions that do not need to happen at the same time.
___
*[`core/src/main/java/google/registry/model/console/UserRoles.java` line 31 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OyVTQ0HyMuFhLukvT:-N8Z8FEC2QH6v0l4NAqf:bhw90w9) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/UserRoles.java#L31)):*

Done.
</blockquote></details>

It looks like you changed the one below, but not this one itself.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1707)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/console/ConsolePermission.java` line 22 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OdA3J-HCyP6zBHKBG:-N8ZgXGT1ZGZjwPF_NYh:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/ConsolePermission.java#L22)):*

Ah, that make sense. Better to include this explanation in the javadoc.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/console/User.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OpF_U6uTXobFBKDsD:-N8ZglNDBTXM-bG7vjhe:bwfd3h9) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L47)):*

If the lesser user is the one whose's supposed to set the password, why should they be granted special permission to do so? Shouldn't they be able to do it all the time?

The separation of privileges seems weird here, can you give a more concrete example of who these two types of users should be?
</blockquote></details>

again, this is just copying over the way that the existing password stuff works. 

enabling a user to use the registry lock feature or forcing a password reset are features done by more powerful registrar users (or by support agents) 
___
*[`core/src/main/java/google/registry/model/console/User.java` line 127 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OwMtr5ZzzFPiDDwSn:-N8ZhQX24QhDKGqjcOS2:b-nsh1uo) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L127)):*

The behavior should definitely be documented in javadoc.

Also, if we want the powerful user to reset the registry lock password (this is not the lesser user's password, if I understand correctly, this is per registrar/registry pair password, regardless of how many users a registrar might have in our system), we could just have a reset function. Allowing the lesser user to reset the password and resetting the password itself seems to be two orthogonal functions that do not need to happen at the same time.
</blockquote></details>

again, this is just how the existing code works in relation to the current console, and it's set up so that registrar admins can grant permissions to other users at that registrar to onboard themselves to the registry lock feature 
___
*[`core/src/main/java/google/registry/model/console/UserRoles.java` line 31 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OyVTQ0HyMuFhLukvT:-N8ZghEX0rUf4Ni8ZDIO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/UserRoles.java#L31)):*

It looks like you changed the one below, but not this one itself.
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1707)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/console/User.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OpF_U6uTXobFBKDsD:-N8eTjW_3kDCHTjFD9Vl:b-bkhsoq) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L47)):*

again, this is just copying over the way that the existing password stuff works. 

enabling a user to use the registry lock feature or forcing a password reset are features done by more powerful registrar users (or by support agents) 
</blockquote></details>

See below, I will approve this PR, but I think we need someone from the business side to reconsider whether this is the most intuitive behavior.
___
*[`core/src/main/java/google/registry/model/console/User.java` line 127 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OwMtr5ZzzFPiDDwSn:-N8eTQia0oFPw3DfBkXo:b-x2xtsg) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L127)):*

again, this is just how the existing code works in relation to the current console, and it's set up so that registrar admins can grant permissions to other users at that registrar to onboard themselves to the registry lock feature 
</blockquote></details>

I understand this is copying existing behavior. However, just because they exist doesn't mean they are correct, or the most intuitive way of doing this. As I mentioned, it is not clear why allowing the lesser user to reset the password would reset the password immediately.

I'm OK with the PR getting merged as is, since it's not changing behaviors. But this behavior itself should probably be revisited with more UX involvement.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1707)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/console/User.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OpF_U6uTXobFBKDsD:-N8jD4WjB4WS10fMYTBc:b-1hsdbv) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L47)):*

See below, I will approve this PR, but I think we need someone from the business side to reconsider whether this is the most intuitive behavior.
</blockquote></details>

Yeah, this was originally written with input from Bruno and Kirsten from the business team. It's definitely not perfect. 
___
*[`core/src/main/java/google/registry/model/console/User.java` line 127 at r1](https://reviewable.io/reviews/google/nomulus/1707#-N8OwMtr5ZzzFPiDDwSn:-N8jD8_LBLSVzc6DWFqC:bjpmcg7) ([raw file](https://github.com/google/nomulus/blob/dd8ef78c889894071d7da288f8c7562a72eadee8/core/src/main/java/google/registry/model/console/User.java#L127)):*

I understand this is copying existing behavior. However, just because they exist doesn't mean they are correct, or the most intuitive way of doing this. As I mentioned, it is not clear why allowing the lesser user to reset the password would reset the password immediately.

I'm OK with the PR getting merged as is, since it's not changing behaviors. But this behavior itself should probably be revisited with more UX involvement.
</blockquote></details>

Yeah to clarify hopefully, the goal was that users are allowed to set their registry lock password through the UI, but it requires a power user to reset the password or to enable it 

So the power user says "user X can set their password" which, in and of itself, doesn't give the user access to use registry lock. That comes after the user logs on to the UI and actually does set their password. 

Normally we'd use a role-based system -- and we already have the roles for this in the new UI, so we'll delete this before tooooo long hopefully.

PR #1709, opened 2022-07-20
Ignore TLD state on domain create when using corresponding token

gbrodman:
For non-trademarked domains, this basically means that we can register them at any time with an allocation token that bypasses TLD state. 

For trademarked domains, signed marks are still required in the sunrise state. In addition, we forbid any registrations of trademarked domains in any quiet period that occurs before sunrise, but we allow registration of these domains in any quiet period that occurs after sunrise (assuming one has a token that bypasses TLD state). 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1709)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 280 at r1](https://reviewable.io/reviews/google/nomulus/1709#-N7qrVeF4SAFRViTqqnK:-N7qrVeF4SAFRViTqqnL:b-8n1zbh) ([raw file](https://github.com/google/nomulus/blob/64a7d3e7db1ddf07bf024591f344e384e2dd15c1/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L280)):*
> ```Java
>       checkHasBillingAccount(registrarId, registry.getTldStr());
>       boolean isValidReservedCreate = isValidReservedCreate(domainName, allocationToken);
>       ClaimsList claimsList = ClaimsListDao.get();
> ```

This uncached load operation is unnecessarily expensive to perform during every domain create.  It should be cached (the `ClaimsList` revision rarely changes, only once every 12 hours, and I believe then only if the claims list has actually changed, which it often doesn't).

Also, this shouldn't be here.  It should be done inside of `verifyIsGaOrSpecialCase()`, only when actually needed (i.e. inside the conditional).  Same for inside of `verifyClaimsNoticeIfAndOnlyIfNeeded()`.  Once it's cached, it'll be a noop to potentially load it twice.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 495 at r1](https://reviewable.io/reviews/google/nomulus/1709#-N7qsXOi7gkJJVn_l2NC:-N7qsXOi7gkJJVn_l2ND:bc1hfrb) ([raw file](https://github.com/google/nomulus/blob/64a7d3e7db1ddf07bf024591f344e384e2dd15c1/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L495)):*
> ```Java
>         || behavior.equals(RegistrationBehavior.ANCHOR_TENANT)) {
>       // Trademarked and non-trademarked names get treated differently in sunrise or quiet periods
>       boolean isTrademarkedName = claimsList.getClaimKey(domainLabel).isPresent();
> ```

There's no need to attempt to load this if sunrise has already finished.  May as well not add unnecessary loads during the performance-critical domain create flow.

So I think you want to invert the handling of the conditionals here.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 280 at r1](https://reviewable.io/reviews/google/nomulus/1709#-N7qrVeF4SAFRViTqqnK:-N7qtxmp3rklrk6ro5VE:b-rloruj) ([raw file](https://github.com/google/nomulus/blob/64a7d3e7db1ddf07bf024591f344e384e2dd15c1/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L280)):*

This uncached load operation is unnecessarily expensive to perform during every domain create.  It should be cached (the `ClaimsList` revision rarely changes, only once every 12 hours, and I believe then only if the claims list has actually changed, which it often doesn't).

Also, this shouldn't be here.  It should be done inside of `verifyIsGaOrSpecialCase()`, only when actually needed (i.e. inside the conditional).  Same for inside of `verifyClaimsNoticeIfAndOnlyIfNeeded()`.  Once it's cached, it'll be a noop to potentially load it twice.
</blockquote></details>

Unfortunately the claims list isn't really cached unless you follow one very specific code path (where you choose to load ALL of the claims entries for a given claims list revision). And even then it's only cached for the specific ClaimsList Java object you have. Caching isn't really relevant at all (at least, with how ClaimsList is written at the moment)

We're already doing the ClaimsList load every time the DomainCreateFlow is called, since our claims periods are indefinite, `verifyClaimsNoticeIfAndOnlyIfNeeded` is always called. This doesn't add any additional loads, and removes one call to `ClaimsListDao.get()` which isn't cached. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 495 at r1](https://reviewable.io/reviews/google/nomulus/1709#-N7qsXOi7gkJJVn_l2NC:-N7qvyJw4cw4iL9LnM-4:bwm4jzq) ([raw file](https://github.com/google/nomulus/blob/64a7d3e7db1ddf07bf024591f344e384e2dd15c1/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L495)):*

There's no need to attempt to load this if sunrise has already finished.  May as well not add unnecessary loads during the performance-critical domain create flow.

So I think you want to invert the handling of the conditionals here.
</blockquote></details>

Sure (it's not quite as simple as swapping them, but it basically is)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 280 at r1](https://reviewable.io/reviews/google/nomulus/1709#-N7qrVeF4SAFRViTqqnK:-N7qzEcUA2wk3MxQFHX1:b-jflk3h) ([raw file](https://github.com/google/nomulus/blob/64a7d3e7db1ddf07bf024591f344e384e2dd15c1/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L280)):*

Unfortunately the claims list isn't really cached unless you follow one very specific code path (where you choose to load ALL of the claims entries for a given claims list revision). And even then it's only cached for the specific ClaimsList Java object you have. Caching isn't really relevant at all (at least, with how ClaimsList is written at the moment)

We're already doing the ClaimsList load every time the DomainCreateFlow is called, since our claims periods are indefinite, `verifyClaimsNoticeIfAndOnlyIfNeeded` is always called. This doesn't add any additional loads, and removes one call to `ClaimsListDao.get()` which isn't cached. 
</blockquote></details>

OK, so if we're already loading it all the time anyway then it should definitely be cached (though, admittedly that doesn't have to be part of this CL).

And caching definitely is relevant.  What you'd be caching is the topmost revision, so that when you perform a query on a specific label, you only need to query that specific label, not first query the revision and then use that to query the label.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 495 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7qzoi2A94pllkWrq2z:-N7qzoi2A94pllkWrq3-:b4gga7e) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L495)):*
> ```Java
>         || behavior.equals(RegistrationBehavior.ANCHOR_TENANT)) {
>       // If bypassing TLD state checks, a post-sunrise quiet period is always fine
>       if (currentState.equals(QUIET_PERIOD)
> ```

Why is it necessary to have `currentState.equals(QUIET_PERIOD)` in this conditional?  Why not only check if sunrise is already over?  If there were some other state besides `QUIET_PERIOD` we'd still expect that to be bypassed by `BYPASS_TLD_STATE` as well.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 512 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7r-UFTAMhXo4zrdqFS:-N7r-UFTAMhXo4zrdqFT:bjh3gxd) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L512)):*
> ```Java
>     }
> 
>     // Anchor tenant overrides all other considerations to allow registration
> ```

I don't this comment as written is correct.  You've now moved the `isAnchorTenant` check to below the previous check, so you are now causing anchor tenant registrations to fail on TMed terms when previously they would have succeeded.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 280 at r1](https://reviewable.io/reviews/google/nomulus/1709#-N7qrVeF4SAFRViTqqnK:-N7r0vTR5MKH6vZ4MKlf:b-8fri7t) ([raw file](https://github.com/google/nomulus/blob/64a7d3e7db1ddf07bf024591f344e384e2dd15c1/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L280)):*

OK, so if we're already loading it all the time anyway then it should definitely be cached (though, admittedly that doesn't have to be part of this CL).

And caching definitely is relevant.  What you'd be caching is the topmost revision, so that when you perform a query on a specific label, you only need to query that specific label, not first query the revision and then use that to query the label.
</blockquote></details>

yeah my point is that the concept of caching isn't super relevant to this PR since ClaimsList isn't really cached at the moment 

and yeah what the PR is doing now is "caching" the top-level object so we don't need to do the revisionId query twice
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 495 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7qzoi2A94pllkWrq2z:-N7r2SwHD-ZGv_wmmCA4:blls47s) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L495)):*

Why is it necessary to have `currentState.equals(QUIET_PERIOD)` in this conditional?  Why not only check if sunrise is already over?  If there were some other state besides `QUIET_PERIOD` we'd still expect that to be bypassed by `BYPASS_TLD_STATE` as well.
</blockquote></details>

yeah fair, i guess the main point is that it is not currently sunrise, done
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 512 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7r-UFTAMhXo4zrdqFS:-N7r2vgh2LdagLzruuDK:b-92zxy6) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L512)):*

I don't this comment as written is correct.  You've now moved the `isAnchorTenant` check to below the previous check, so you are now causing anchor tenant registrations to fail on TMed terms when previously they would have succeeded.
</blockquote></details>

Depending on how you parse it yeah it could be incorrect. I clarified it, "overrides any remaining considerations" 

The use case here is an anchor tenant, regardless of trademark status, created in a quiet period before sunrise. 

While we may wish to change it so that TMed anchor tenants cannot be created before sunrise, that's out of the scope of this PR.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 512 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7r-UFTAMhXo4zrdqFS:-N7rOHmBBh4crPUQbBJh:b-kkoqg5) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L512)):*
> While we may wish to change it so that TMed anchor tenants cannot be created before sunrise, that's out of the scope of this PR.

That is what you have in fact done here, for the current code path of anchor tenants (i.e. the one where anchor-tenant-ness comes from the reserved list reservation type rather than the state on the token).  This change seems unintended?  Admittedly, we ultimately want to remove `isAnchorTenant` from the method signature entirely, and perform that check solely based on the status of the token.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 512 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7r-UFTAMhXo4zrdqFS:-N7rOgHW2qWB3_TnnfJA:b815qaf) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L512)):*

> While we may wish to change it so that TMed anchor tenants cannot be created before sunrise, that's out of the scope of this PR.

That is what you have in fact done here, for the current code path of anchor tenants (i.e. the one where anchor-tenant-ness comes from the reserved list reservation type rather than the state on the token).  This change seems unintended?  Admittedly, we ultimately want to remove `isAnchorTenant` from the method signature entirely, and perform that check solely based on the status of the token.
</blockquote></details>

If you don't want to change the current behavior, then the `isAnchorTenant` short circuit needs to be at the beginning of the method, as it previously was.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 512 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7r-UFTAMhXo4zrdqFS:-N7vL5F45i_BxmtLb_h4:b-4z137m) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L512)):*

If you don't want to change the current behavior, then the `isAnchorTenant` short circuit needs to be at the beginning of the method, as it previously was.
</blockquote></details>

The only behavioral change in this code is that anchor tenant domains that are created in sunrise (note: sunrise only, doesn't apply to quiet periods) require signed marks. You can see that in the tests too. 

You can still, in this code, create anchor tenants during any quiet period.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 512 at r3](https://reviewable.io/reviews/google/nomulus/1709#-N7r-UFTAMhXo4zrdqFS:-N7vMCdx7uht29XJrN46:byaxqzx) ([raw file](https://github.com/google/nomulus/blob/d18076decbda69a59d8a895e69410c995af11fae/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L512)):*

The only behavioral change in this code is that anchor tenant domains that are created in sunrise (note: sunrise only, doesn't apply to quiet periods) require signed marks. You can see that in the tests too. 

You can still, in this code, create anchor tenants during any quiet period. 
</blockquote></details>

well, my first sentence should have said "The only behavioral change with regards to anchor tenants created before GA is that..."

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1709#-:-N7vW30HEP9UODq-K9-9:bnfp4nl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1710, opened 2022-07-20
Remove ofy support from BillingEvent

jianglai:
This PR turns out to be more massive than I would have liked but it
  deals with all billing event related stuff, which are more or link all
  intertwined:

  * Remove all billing events as Ofy entities.
  * Add a temporary annotation to allow BillingEvent's ID to be
    auto-allocated by ofy while not lacking the Ofy @Id annotation.
  * Remove Modification, which is only used in ofy.
  * Remove BillingVKey, as we do not need to store the ofy key parent
     information anymore. The VKey for a billing event now just contain
     its primary key, and can be converted by VKeyConverter.
  * Remove BigQuery related code in the billing pipeline.

  Note that after BillingVKey is removed, several columns in
  BillingCancellation are no longer needed. The change to database schema
  will be handled in https://github.com/google/nomulus/pull/1721 after
  this PR is deployed to production.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1710)


weiminyu (approved):
> Reviewed 8 of 48 files at r1, 16 of 41 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1710)* status: 26 of 71 files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1720, opened 2022-07-26
Add CurrentPackageToken to Domain table

sarahcaseybot:
Do not merge until go/r3pr/1719 is deployed. 

This adds an allocation_token column to the Domain table to be used to track which package a domain belongs to if it belongs to one. More details in go/registry-package-promo.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1720)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1720)* status: 3 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1720#-N8-9BBS52V13kcEQpLH:-N8-9BBT43iOybQFkyiz:b-nnrkua) ([raw file](https://github.com/google/nomulus/blob/9b45faee5ce7cbfde1b70c004a6fecb71e55519a/core/src/main/java/google/registry/model/domain/DomainContent.java#L291)):*
> ```Java
> 
>   /**
>    * The {@link AllocationToken} used during domain creation to include this domain in a package.
> ```

I don't think this comment is accurate.  This should simply be the allocation token used to create the domain, if one was used.  So not just specific to packages.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1720)* status: 3 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1720#-N8-9BBS52V13kcEQpLH:-N80FVrp8O10pw7BBOSd:bx0qal0) ([raw file](https://github.com/google/nomulus/blob/9b45faee5ce7cbfde1b70c004a6fecb71e55519a/core/src/main/java/google/registry/model/domain/DomainContent.java#L291)):*

I was planning to only store the allocation token used on creation if it is a PackagePromotion allocation token. I assumed it was not necessary to store other allocation tokens used in domain creations since we had not done that historically. Is there a reason why it would be useful to store allocation tokens otherwise? I think this would make it harder to discern based on a simple domain lookup if a domain is part of a package simply if it has an allocation token stored on it.  
</blockquote></details>

If this is to be used only to determine what package a domain belongs to, then the name needs to be different.  `allocationToken` is misleading that case.

And it will change after creation if the domain is removed from the package, right?  Then the name should be something like `currentPackageToken` or similar.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1720)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/model/domain/DomainContent.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1720#-N8-9BBS52V13kcEQpLH:-N80koqM-2oyr3w-npUq:b6io7x8) ([raw file](https://github.com/google/nomulus/blob/9b45faee5ce7cbfde1b70c004a6fecb71e55519a/core/src/main/java/google/registry/model/domain/DomainContent.java#L291)):*

If this is to be used only to determine what package a domain belongs to, then the name needs to be different.  `allocationToken` is misleading that case.

And it will change after creation if the domain is removed from the package, right?  Then the name should be something like `currentPackageToken` or similar.
</blockquote></details>

Updated column name to currentPackageToken

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1720)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

___
*[`db/src/main/resources/sql/flyway/V122__add_current_package_token_to_domain.sql` line 18 at r2](https://reviewable.io/reviews/google/nomulus/1720#-N89bTEz5JIseUMgnST0:-N89bTF-7nfMCbEkkb5Q:b-ph4yfo) ([raw file](https://github.com/google/nomulus/blob/c28247aefe34310b6c332392ce4a3d7ade94dd49/db/src/main/resources/sql/flyway/V122__add_current_package_token_to_domain.sql#L18)):*
> ```TSQL
> alter table "DomainHistory" add column current_package_token text;
> 
> ALTER TABLE "Domain" ADD CONSTRAINT fk_domain_current_package_token FOREIGN KEY (current_package_token) REFERENCES "AllocationToken"(token) DEFERRABLE INITIALLY DEFERRED;
> ```

There's a missing foreign key here (but I guess it'll turn up when you sync from the other one).

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1720#-:-N89bZjnCfAWZi2jdcSp:bnfp4nl)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @CydeWeys, @sarahcaseybot, and @weiminyu)

PR #1730, opened 2022-08-02
Fix GAE build under Gradle 7

jianglai:
For some reason after the upgrade to Gradle, the core.jar file is no
longer included in the generated WAR, even though the deploy_jar
configuration is specified as a dependency.

I could not figure out a way to tweak the configuration dependency to
have core.jar pulled into the .war, so I decided to just explicitly pick
it from its known location.

TESTED=deployed to alpha and verified that the instances can start.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1730)

### DIFF
--- a/appengine_war.gradle
+++ b/appengine_war.gradle
@@ -26,6 +26,7 @@ project.convention.plugins['war'].webAppDirName =
 apply plugin: 'com.google.cloud.tools.appengine'
 
 def coreResourcesDir = "${rootDir}/core/build/resources/main"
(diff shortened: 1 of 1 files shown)


weiminyu (approved):
> Reviewed 1 of 13 files at r1, 2 of 2 files at r2, 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1730)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1731, opened 2022-08-02
Add caches to ClaimsListDao and ClaimsList

gbrodman:
We cache the ClaimsList Java object for six hours (we don't expect it to
change frequently, and the cron job to update it only runs every twelve
hours). Subsequent calls to ClaimsListDao::get will return the cached
value.

Within the ClaimsList Java object itself, we cache any labels that we
have retrieved. While we already have a form of a cache here in the
"labelsToKeys" map, that only handles situations where we've loaded the
entire map from the database. We want to have a non-guaranteed cache in
order to make repeated calls to getClaimKey fast.

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1731)

### DIFF
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1427,6 +1427,11 @@ public final class RegistryConfig {
     return CONFIG_SETTINGS.get().caching.eppResourceMaxCachedEntries;
   }
 
+  /** Returns the amount of time that a particular claims list should be cached. */
+  public static java.time.Duration getClaimsListCacheDuration() {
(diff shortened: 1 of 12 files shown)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1731#-:-N8VSJKd24UHq6P3QAtn:bnfp4nl)* status: 0 of 10 files reviewed, all discussions resolved

gbrodman:
sneaky review!

gbrodman:
pushed changes are 
1. a rebase
2. adding the test cache extension to a couple tests that failed due to caches sticking around 
3. changing the name of the claim key cache to match what checkstyle expects

PR #1735, opened 2022-08-08
Allow anchor tenant creation via allocation token behavior

gbrodman:
This also enforces that non-superusers cannot create registrations on
trademarked names prior to the sunrise period, even if they have an
allocation token with ANCHOR_TENANT behavior.

Further details for context:

Assuming that we're trying to register an anchor tenant or use a BYPASS_TLD_STATE token, trademarked domains can only be registered during or after a sunrise period (requiring a signed mark if in sunrise, and a claims notice otherwise). Non-trademarked domains can be registered at any time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1735)


jianglai (changes requested):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1735)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 505 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IGcsh2xWRzjEYBoYJ:-N9IGcsh2xWRzjEYBoYK:bi6wwkh) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L505)):*
> ```Java
>         return;
>       }
>       if (!currentState.equals(START_DATE_SUNRISE)) {
> ```

Is there a reason why this check is moved after the claims list check? I think it can be modified in-place.



_Code quote:_
```Java
      if (!currentState.equals(START_DATE_SUNRISE)) {
        // If bypassing TLD state checks, a post-sunrise state is always fine
        if (registry.getTldStateTransitions().headMap(now).containsValue(START_DATE_SUNRISE)) {
          return;
        } else {
          // If sunrise hasn't happened yet, trademarked domains are unavailable
          throw new NoTrademarkedRegistrationsBeforeSunriseException(domainLabel);
        }
      }
    }
```
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 510 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IHAlY9KoRP6A8qKDX:-N9IHAlY9KoRP6A8qKDY:b3ywim1) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L510)):*
> ```Java
>           return;
>         } else {
>           // If sunrise hasn't happened yet, trademarked domains are unavailable
> ```

So, is the assumption here that the sunrise state always exists in the TLD state transition map? Therefore, if it is not before the current state, it must be after it (meaning sunrise hasn't happened yet).

It's better to spell this out in comments, for the benefit of future readers.



_Code quote:_
```Java
// If sunrise hasn't happened yet, trademarked domains are unavailable
```
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2851 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IHz-6DeZ6kzKIxeGA:-N9IHz-6DeZ6kzKIxeGB:b-88014e) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2851)):*
> ```Java
> 
>   @Test
>   void testSuccess_anchorTenant_qpBeforeSunrise_nonTrademarked_viaToken() throws Exception {
> ```

Does "qp" stand for "quiet period"? In other tests they are spelled out. We should do the same here.



_Code quote:_
```Java
qp
```
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2870 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IIFXF8ZBrLUZikVpD:-N9IIFXF8ZBrLUZikVpE:b8732dc) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2870)):*
> ```Java
> 
>   @Test
>   void testFailure_anchorTenantInQuietPeriodBeforeSunriseWithoutClaims_viaToken() {
> ```

The naming style should be consistent with the previous test.

Also, there are wayyyy to many tests here due to combinatorial explosion to understand what is being tested in each case. Can you please add some comments at the beginning of the set of tests to state what are the (binary) factors that are involved that resulted in the 2^n test cases. And make sure the test names clearly refer to each combination (probably better to separate each condition with underscores.



_Code quote:_
```Java
anchorTenantInQuietPeriodBeforeSunriseWithoutClaims
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1735)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 505 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IGcsh2xWRzjEYBoYJ:-N9ISeqL3TRqbauDR7dW:b-xrfb8s) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L505)):*

Is there a reason why this check is moved after the claims list check? I think it can be modified in-place.
</blockquote></details>

this is moved because we want to add the explicit exception in the case where you're trying to register a trademarked domain before sunrise. The claims list check should come first because if the domain isn't in the claims list (isn't trademarked) then it's always (when using these types of tokens) going to be ok 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 510 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IHAlY9KoRP6A8qKDX:-N9ITfAz-8CrcB3a-Q3b:b6vxkj8) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L510)):*

So, is the assumption here that the sunrise state always exists in the TLD state transition map? Therefore, if it is not before the current state, it must be after it (meaning sunrise hasn't happened yet).

It's better to spell this out in comments, for the benefit of future readers.
</blockquote></details>

Yes, the rules are "Trademarked domains cannot be registered until after the sunrise period has ended, unless a valid signed mark is provided." and "Signed marks can only be provided during sunrise."
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2851 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IHz-6DeZ6kzKIxeGA:-N9IUN6_E3BqPb7qiOnP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2851)):*

Does "qp" stand for "quiet period"? In other tests they are spelled out. We should do the same here.
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2870 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IIFXF8ZBrLUZikVpD:-N9IVt9Y-SSDMSB4AnTF:bsm6jtg) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2870)):*

The naming style should be consistent with the previous test.

Also, there are wayyyy to many tests here due to combinatorial explosion to understand what is being tested in each case. Can you please add some comments at the beginning of the set of tests to state what are the (binary) factors that are involved that resulted in the 2^n test cases. And make sure the test names clearly refer to each combination (probably better to separate each condition with underscores.
</blockquote></details>

yeah fair point -- I standardized the names and added headers to each of the four sections

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1735)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 505 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IGcsh2xWRzjEYBoYJ:-N9RQ1_c05gtL5HsdUtp:b-b65br9) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L505)):*

this is moved because we want to add the explicit exception in the case where you're trying to register a trademarked domain before sunrise. The claims list check should come first because if the domain isn't in the claims list (isn't trademarked) then it's always (when using these types of tokens) going to be ok 
</blockquote></details>

Ah, yes. Good point!
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2870 at r1](https://reviewable.io/reviews/google/nomulus/1735#-N9IIFXF8ZBrLUZikVpD:-N9RW3eIDqJhxoVaR0IQ:bovplaa) ([raw file](https://github.com/google/nomulus/blob/7131ecdc69bb15750bc4e98e6ded6a8b3ebe7756/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2870)):*

yeah fair point -- I standardized the names and added headers to each of the four sections
</blockquote></details>

Thank you! This is much better.

PR #1744, opened 2022-08-15
Allow UserAuthInfo to contain either old GAE Users or new console Users

gbrodman:
This means that LegacyAuthenticationMechanism or a to-be-created
OAuth2AuthenticationMechanism) can return a UserAuthInfo object that
contains either the GAE User or the console User as appropriate. The
goal is that the non-auth flows shouldn't have to know about which user
type it is. Note: the registry lock flow (for now) needs to know about
the separate types of auth because it is a separate level of auth from
the standard AuthenticatedRegistrarAccessor.

The AuthenticatedRegistrarAccessor code is a bit odd because the new
role system doesn't quite fit neatly into the old registrar ->
OWNER,ADMIN system but this is a fine approximation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1744)

### DIFF
--- a/core/src/main/java/google/registry/rdap/RdapModule.java
+++ b/core/src/main/java/google/registry/rdap/RdapModule.java
@@ -108,7 +108,7 @@ public final class RdapModule {
     if (userAuthInfo.isUserAdmin()) {
       return RdapAuthorization.ADMINISTRATOR_AUTHORIZATION;
     }
-    ImmutableSet<String> clientIds = registrarAccessor.getAllClientIdWithRoles().keySet();
+    ImmutableSet<String> clientIds = registrarAccessor.getAllRegistrarIdsWithRoles().keySet();
     if (clientIds.isEmpty()) {
       logger.atWarning().log("Couldn't find registrar for User %s.", authResult.userIdForLogging());
       return RdapAuthorization.PUBLIC_AUTHORIZATION;
--- a/core/src/main/java/google/registry/request/auth/AuthResult.java
+++ b/core/src/main/java/google/registry/request/auth/AuthResult.java
@@ -42,7 +42,7 @@ public abstract class AuthResult {
             userAuthInfo ->
                 String.format(
                     "%s %s",
-                    userAuthInfo.isUserAdmin() ? "admin" : "user", userAuthInfo.user().getEmail()))
+                    userAuthInfo.isUserAdmin() ? "admin" : "user", userAuthInfo.getEmailAddress()))
         .orElse("<logged-out user>");
   }
 
--- a/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java
+++ b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java
@@ -40,8 +40,8 @@ import javax.inject.Inject;
  * <p>A user has OWNER role on a Registrar if there exists a {@link RegistrarPoc} with that user's
  * gaeId and the registrar as a parent.
  *
- * <p>An "admin" has in addition OWNER role on {@code #registryAdminClientId} and to all non-{@code
- * REAL} registrars (see {@link Registrar#getType}).
+ * <p>An "admin" has in addition OWNER role on {@code #registryAdminRegistrarId} and to all
+ * non-{@code REAL} registrars (see {@link Registrar#getType}).
  *
  * <p>An "admin" also has ADMIN role on ALL registrars.
  *
@@ -76,7 +76,7 @@ public class AuthenticatedRegistrarAccessor {
   private final boolean isAdmin;
 
   /**
-   * Gives all roles a user has for a given clientId.
+   * Gives all roles a user has for a given registrar ID.
    *
    * <p>The order is significant, with "more specific to this user" coming first.
    *
@@ -107,13 +107,13 @@ public class AuthenticatedRegistrarAccessor {
   @Inject
   public AuthenticatedRegistrarAccessor(
       AuthResult authResult,
-      @Config("registryAdminClientId") String registryAdminClientId,
+      @Config("registryAdminClientId") String registryAdminRegistrarId,
       @Config("gSuiteSupportGroupEmailAddress") Optional<String> gSuiteSupportGroupEmailAddress,
       Lazy<GroupsConnection> lazyGroupsConnection) {
     this.isAdmin = userIsAdmin(authResult, gSuiteSupportGroupEmailAddress, lazyGroupsConnection);
 
     this.userIdForLogging = authResult.userIdForLogging();
-    this.roleMap = createRoleMap(authResult, this.isAdmin, registryAdminClientId);
+    this.roleMap = createRoleMap(authResult, this.isAdmin, registryAdminRegistrarId);
 
     logger.atInfo().log("%s has the following roles: %s", userIdForLogging(), roleMap);
   }
@@ -129,7 +129,7 @@ public class AuthenticatedRegistrarAccessor {
    * Creates a "logged-in user" accessor with a given role map, used for tests.
    *
    * <p>The user will be allowed to create Registrars (and hence do OT&amp;E setup) iff they have
-   * the role of ADMIN for at least one clientId.
+   * the role of ADMIN for at least one registrar ID.
    *
    * <p>The user's "name" in logs and exception messages is "TestUserId".
    */
@@ -148,59 +148,62 @@ public class AuthenticatedRegistrarAccessor {
   }
 
   /**
-   * A map that gives all roles a user has for a given clientId.
+   * A map that gives all roles a user has for a given registrar ID.
    *
    * <p>Throws a {@link RegistrarAccessDeniedException} if the user is not logged in.
    *
    * <p>The result is ordered starting from "most specific to this user".
    *
-   * <p>If you want to load the {@link Registrar} object from these (or any other) {@code clientId},
-   * in order to perform actions on behalf of a user, you must use {@link #getRegistrar} which makes
-   * sure the user has permissions.
+   * <p>If you want to load the {@link Registrar} object from these (or any other) {@code
+   * registrarId}, in order to perform actions on behalf of a user, you must use {@link
+   * #getRegistrar} which makes sure the user has permissions.
    *
    * <p>Note that this is an OPTIONAL step in the authentication - only used if we don't have any
-   * other clue as to the requested {@code clientId}. It is perfectly OK to get a {@code clientId}
-   * from any other source, as long as the registrar is then loaded using {@link #getRegistrar}.
+   * other clue as to the requested {@code registrarId}. It is perfectly OK to get a {@code
+   * registrarId} from any other source, as long as the registrar is then loaded using {@link
+   * #getRegistrar}.
    */
-  public ImmutableSetMultimap<String, Role> getAllClientIdWithRoles() {
+  public ImmutableSetMultimap<String, Role> getAllRegistrarIdsWithRoles() {
     return roleMap;
   }
 
   /**
    * Returns all the roles the current user has on the given registrar.
    *
-   * <p>This is syntactic sugar for {@code getAllClientIdWithRoles().get(clientId)}.
+   * <p>This is syntactic sugar for {@code getAllRegistrarIdsWithRoles().get(registrarId)}.
    */
-  public ImmutableSet<Role> getRolesForRegistrar(String clientId) {
-    return getAllClientIdWithRoles().get(clientId);
+  public ImmutableSet<Role> getRolesForRegistrar(String registrarId) {
+    return getAllRegistrarIdsWithRoles().get(registrarId);
   }
 
   /**
    * Checks if we have a given role for a given registrar.
    *
-   * <p>This is syntactic sugar for {@code getAllClientIdWithRoles().containsEntry(clientId, role)}.
+   * <p>This is syntactic sugar for {@code getAllRegistrarIdsWithRoles().containsEntry(registrarId,
+   * role)}.
    */
-  public boolean hasRoleOnRegistrar(Role role, String clientId) {
-    return getAllClientIdWithRoles().containsEntry(clientId, role);
+  public boolean hasRoleOnRegistrar(Role role, String registrarId) {
+    return getAllRegistrarIdsWithRoles().containsEntry(registrarId, role);
   }
 
   /**
    * "Guesses" which client ID the user wants from all those they have access to.
    *
-   * <p>If no such ClientIds exist, throws a RegistrarAccessDeniedException.
+   * <p>If no such registrar IDs exist, throws a RegistrarAccessDeniedException.
    *
-   * <p>This should be the ClientId "most likely wanted by the user".
+   * <p>This should be the registrar ID "most likely wanted by the user".
    *
-   * <p>If you want to load the {@link Registrar} object from this (or any other) {@code clientId},
-   * in order to perform actions on behalf of a user, you must use {@link #getRegistrar} which makes
-   * sure the user has permissions.
+   * <p>If you want to load the {@link Registrar} object from this (or any other) {@code
+   * registrarId}, in order to perform actions on behalf of a user, you must use {@link
+   * #getRegistrar} which makes sure the user has permissions.
    *
    * <p>Note that this is an OPTIONAL step in the authentication - only used if we don't have any
-   * other clue as to the requested {@code clientId}. It is perfectly OK to get a {@code clientId}
-   * from any other source, as long as the registrar is then loaded using {@link #getRegistrar}.
+   * other clue as to the requested {@code registrarId}. It is perfectly OK to get a {@code
+   * registrarId} from any other source, as long as the registrar is then loaded using {@link
+   * #getRegistrar}.
    */
-  public String guessClientId() throws RegistrarAccessDeniedException {
-    return getAllClientIdWithRoles().keySet().stream()
+  public String guessRegistrarId() throws RegistrarAccessDeniedException {
+    return getAllRegistrarIdsWithRoles().keySet().stream()
         .findFirst()
         .orElseThrow(
             () ->
@@ -227,7 +230,7 @@ public class AuthenticatedRegistrarAccessor {
 
     if (!registrarId.equals(registrar.getRegistrarId())) {
       logger.atSevere().log(
-          "registrarLoader.apply(clientId) returned a Registrar with a different clientId. "
+          "registrarLoader.apply(registrarId) returned a Registrar with a different registrarId. "
               + "Requested: %s, returned: %s.",
           registrarId, registrar.getRegistrarId());
       throw new RegistrarAccessDeniedException("Internal error - please check logs");
@@ -237,7 +240,7 @@ public class AuthenticatedRegistrarAccessor {
   }
 
   public void verifyAccess(String registrarId) throws RegistrarAccessDeniedException {
-    ImmutableSet<Role> roles = getAllClientIdWithRoles().get(registrarId);
+    ImmutableSet<Role> roles = getAllRegistrarIdsWithRoles().get(registrarId);
 
     if (roles.isEmpty()) {
       throw new RegistrarAccessDeniedException(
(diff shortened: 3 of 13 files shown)


jianglai (changes requested):
To be clear, the current `OAuthAuthenticationMechanism` is using OAuth 2. I think we will be modifying it, not creating a new one.

Also, did I understand correctly that this PR only makes it possible to handle both cases, but we are not creating any `UserAuthInfo` with console users just yet?

> Reviewed 12 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1744#-:-N9wm5iH65uH69pNlWiq:b27nbki)* status: all files reviewed, 7 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 292 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAG64s8JwwsjHyqqHB:-NAAG64s8JwwsjHyqqHC:bcngmke) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L292)):*
> ```Java
>     String emailAddress =
>         userAuthInfo.user().isPresent()
>             ? userAuthInfo.user().get().getEmail()
> ```

Does `AutoValue` allow us to add arbitrary methods? It seems like we should have a `getEmail` method on the `UserAuthInfo` object.



_Code quote:_
```Java
            ? userAuthInfo.user().get().getEmail()
            : userAuthInfo.consoleUser().get().getEmailAddress();
```
___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 303 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAHPH163ZqsUrVRN2e:-NAAHPH163ZqsUrVRN2f:b-a3kcm8) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L303)):*
> ```Java
> 
>   /**
>    * Returns a map of registrar client IDs to roles for all registrars that the user has access to.
> ```

registrar IDs



_Code quote:_
```Java
registrar client IDs
```
___
*[`core/src/main/java/google/registry/request/auth/UserAuthInfo.java` line 26 at r2](https://reviewable.io/reviews/google/nomulus/1744#-N9wpxHUDopMZZsHKPbR:-N9wpxHUDopMZZsHKPbS:b4b10wh) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/request/auth/UserAuthInfo.java#L26)):*
> ```Java
> 
>   /** User object from the AppEngine Users API. */
>   public abstract Optional<User> user();
> ```

Maybe this should be renamed to `appEngineUser()`?

Also, what would this method return in the post-GAE world? Would we have another user type to represent authentication results for non-console user logins? Do you envision this new type and the console user type would implement the same interface, therefore removing the needs to have two separate optional users here?



_Code quote:_
```Java
user
```
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAJA9SAi4XlVFrorCQ:-NAAJA9SAi4XlVFrorCR:bb3bjgy) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L161)):*
> ```Java
>     boolean isRegistryLockAllowed;
>     String relevantEmail;
>     if (authResult.userAuthInfo().get().user().isPresent()) {
> ```

It feels like we are doing a lot of this "check if this field is present, then do this, or do that" on the `UserAutoInfo` object...

If we cannot add methods to the object itself, due to `AutoValue` limitations, to make it easier to extract information like username and email, maybe we can just have some static helper methods to do it?



_Code quote:_
```Java
if (authResult.userAuthInfo().get().user().isPresent()) {
```
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 173 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAN8Bs2PoEB_0q1Xmr:-NAAN8Bs2PoEB_0q1Xms:b-qyzzbu) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L173)):*
> ```Java
>     } else {
>       google.registry.model.console.User user = authResult.userAuthInfo().get().consoleUser().get();
>       isRegistryLockAllowed =
> ```

Why don't we check `isAdmin` in this case?



_Code quote:_
```Java
 isRegistryLockAllowed =
```
___
*[`core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java` line 434 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAALxbk32xU7nE8tNSd:-NAALxbl389qrGN1RNaH:b-93h9v9) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java#L434)):*
> ```Java
> 
>   @Test
>   void testConsoleUser_globalRole() {
> ```

What happens when a console user is not admin but their global role is `FTE` instead of `SUPPORT_AGENT`?



_Code quote:_
```Java
globalRole
```
___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 228 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAOORm9XvOuQ1xueRL:-NAAOORm9XvOuQ1xueRM:b-81lu6t) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L228)):*
> ```Java
> 
>   @Test
>   void testSuccess_consoleUser() throws Exception {
> ```

Should have tests for when the console user is admin, and for when the password check fails.



_Code quote:_
```Java
void testSuccess_consoleUser
```

gbrodman (commented):
For now creating a new one instead of modifying the existing one, because the existing one is still used. 

And yes, that is correct. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1744#-:-NAAT1V383eqU7FBoCfs:b-5767mn)* status: all files reviewed, 7 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 292 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAG64s8JwwsjHyqqHB:-NAARyHW9TkI4RDjvtqW:b-kbx6vh) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L292)):*

Does `AutoValue` allow us to add arbitrary methods? It seems like we should have a `getEmail` method on the `UserAuthInfo` object.
</blockquote></details>

Yep, as long as they're not abstract. Good point. 
___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 303 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAHPH163ZqsUrVRN2e:-NAATBsR5WJIujtOhxwb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L303)):*

registrar IDs
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/auth/UserAuthInfo.java` line 26 at r2](https://reviewable.io/reviews/google/nomulus/1744#-N9wpxHUDopMZZsHKPbR:-NAATFhP6tB9VNx_9top:b-vybzos) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/request/auth/UserAuthInfo.java#L26)):*

Maybe this should be renamed to `appEngineUser()`?

Also, what would this method return in the post-GAE world? Would we have another user type to represent authentication results for non-console user logins? Do you envision this new type and the console user type would implement the same interface, therefore removing the needs to have two separate optional users here?
</blockquote></details>

In the post-GAE world, we should be able to just remove this entirely and make `consoleUser()` return not-an-Optional. 
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAJA9SAi4XlVFrorCQ:-NAAaYxy2xyaVi_-GQZm:bwr368o) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L161)):*

It feels like we are doing a lot of this "check if this field is present, then do this, or do that" on the `UserAutoInfo` object...

If we cannot add methods to the object itself, due to `AutoValue` limitations, to make it easier to extract information like username and email, maybe we can just have some static helper methods to do it?
</blockquote></details>

We can add methods, however in situations like this where the logic is actually branching, it doesn't help us very much. The old user objects require that we match up the registry lock email address with the `RegistrarPoc` object unfortunately. 
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 173 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAN8Bs2PoEB_0q1Xmr:-NAATqYLApjQWSVKz143:b-1dnp23) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L173)):*

Why don't we check `isAdmin` in this case?
</blockquote></details>

It's encapsulated in the `hasPermission` check. If it wasn't, we'd have to make every permission check something like `if (user.isAdmin() || user.hasPermission(...)` which would be annoying.

Ideally the callers of `hasPermission` don't even need to know that an "admin" functionality exists. 
___
*[`core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java` line 434 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAALxbk32xU7nE8tNSd:-NAAblIy5v_j8bN9p9Gx:b-idk4zo) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java#L434)):*

What happens when a console user is not admin but their global role is `FTE` instead of `SUPPORT_AGENT`?
</blockquote></details>

I don't think this would really happen (FTEs will be admins, the difference is that we might have service account non-FTE admins, possibly). 

The thinking is that currently the AuthenticatedRegistrarAccessor, if you're not admin, requires that you have a RegistrarPoc object in every registrar you want to access -- so we do the same here. 
___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 228 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAOORm9XvOuQ1xueRL:-NAAg8lpDFnHiXXgQgHi:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L228)):*

Should have tests for when the console user is admin, and for when the password check fails.
</blockquote></details>

Done.

jianglai (changes requested):
I don't believe there needs to be any client-side changes for the new OAuth authentication mechanism. You basically just call the Google API itself (like we did in our little exercise) instead of relying on the App Engine SDK to verify the bearer token supplied by the client. So there shouldn't be any reason why you can't just change the existing class in-place, AFAICT.

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1744#-:-NABHyW7AWWxMgcxznb4:b9bvmep)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/registrar/HtmlAction.java` line 91 at r3](https://reviewable.io/reviews/google/nomulus/1744#-NABKMoZAZ_EgVfStqCV:-NABKMoZAZ_EgVfStqCW:bni0g8o) ([raw file](https://github.com/google/nomulus/blob/b206971b14dbbcc21339609f99d296ae56b02337/core/src/main/java/google/registry/ui/server/registrar/HtmlAction.java#L91)):*
> ```Java
>     UserAuthInfo authInfo = authResult.userAuthInfo().get();
>     String username;
>     if (authInfo.appEngineUser().isPresent()) {
> ```

This should be a `getUsername()` method on `UserAuthInfo` as well.



_Code quote:_
```Java
    if (authInfo.appEngineUser().isPresent()) {
      username = authInfo.appEngineUser().get().getNickname();

    } else {
      username = authInfo.consoleUser().get().getEmailAddress();

    }
```
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAJA9SAi4XlVFrorCQ:-NABKYFl54wcGNRu7Xlo:btq1j02) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L161)):*

We can add methods, however in situations like this where the logic is actually branching, it doesn't help us very much. The old user objects require that we match up the registry lock email address with the `RegistrarPoc` object unfortunately. 
</blockquote></details>

I think a `isConsoleUser()` method is more descriptive than calling `.consoleUser().isPresent()`.
___
*[`core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java` line 434 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAALxbk32xU7nE8tNSd:-NABQAUw-w0y5ZoHa6CP:bxtwjm4) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java#L434)):*

I don't think this would really happen (FTEs will be admins, the difference is that we might have service account non-FTE admins, possibly). 

The thinking is that currently the AuthenticatedRegistrarAccessor, if you're not admin, requires that you have a RegistrarPoc object in every registrar you want to access -- so we do the same here. 
</blockquote></details>

I'm a bit confused here. There doesn't seem to be a global role that corresponds to a service account. Also, I think if a user is an admin, it doesn't matter what roles it has. So we have to set a role here because the builder requires it, and it doesn't actually matter which role we set?
___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 255 at r3](https://reviewable.io/reviews/google/nomulus/1744#-NABNXBg9D8WmLnJH9hC:-NABNXBg9D8WmLnJH9hD:b7oz0p4) ([raw file](https://github.com/google/nomulus/blob/b206971b14dbbcc21339609f99d296ae56b02337/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L255)):*
> ```Java
>             .setGaiaId("gaiaId")
>             .setUserRoles(new UserRoles.Builder().setIsAdmin(true).build())
>             .setRegistryLockPassword("hi")
> ```

You don't need to set the password for admin, right?



_Code quote:_
```Java
.setRegistryLockPassword("hi")
```

gbrodman (commented):
Yeah, I was thinking of the additional CookieAuthenticationMechanism that we'll need 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1744#-:-NABRP150HIaUGNfUMm5:b-bsutb0)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/ui/server/registrar/HtmlAction.java` line 91 at r3](https://reviewable.io/reviews/google/nomulus/1744#-NABKMoZAZ_EgVfStqCV:-NABRwD31kgpzFRMGpBC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/b206971b14dbbcc21339609f99d296ae56b02337/core/src/main/java/google/registry/ui/server/registrar/HtmlAction.java#L91)):*

This should be a `getUsername()` method on `UserAuthInfo` as well.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAJA9SAi4XlVFrorCQ:-NABSX-0CgU6XdybrkZ-:b142j45) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L161)):*

I think a `isConsoleUser()` method is more descriptive than calling `.consoleUser().isPresent()`.
</blockquote></details>

Is it really that better, since it seems like it's not saving much code? One downside is that it causes warnings on the `userAuthInfo.consoleUser().get()` call afterward because it's not preceded by a call to `isPresent` 
___
*[`core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java` line 434 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAALxbk32xU7nE8tNSd:-NABS2_t--LsfdkigEbw:bn4lqw) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java#L434)):*

I'm a bit confused here. There doesn't seem to be a global role that corresponds to a service account. Also, I think if a user is an admin, it doesn't matter what roles it has. So we have to set a role here because the builder requires it, and it doesn't actually matter which role we set?
</blockquote></details>

no, the user in this test isn't an admin so they only have partial access 

service accounts will be admins as far as I can tell
___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 255 at r3](https://reviewable.io/reviews/google/nomulus/1744#-NABNXBg9D8WmLnJH9hC:-NABRZHWB_2Oiv4pLL6p:bp5gce3) ([raw file](https://github.com/google/nomulus/blob/b206971b14dbbcc21339609f99d296ae56b02337/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L255)):*

You don't need to set the password for admin, right?
</blockquote></details>

nope, doesn't matter either way -- we can remove it here

jianglai (approved):
Oh, yes, we'll need one to check for the cookies when logging in from the console. For programmatic access from the proxy, we just need to update the existing `OAuthAuthenticationMechanism`. Although both of them are using OAuth, so it might be a bit confusing to call one of them OAuth. Maybe we can name them `OAuthCookieAuthenticationMechanism` and `OAuthHeaderAuthenticatioMechanism`, respectively?

Note that it is also possible to just have one class and have it check for both the cookie and the header.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1744#-:-NAF6HDl4o53JizylJDN:bg9mlq1)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAJA9SAi4XlVFrorCQ:-NAF75H75MD7ICKIchuj:b-qttp1u) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L161)):*

Is it really that better, since it seems like it's not saving much code? One downside is that it causes warnings on the `userAuthInfo.consoleUser().get()` call afterward because it's not preceded by a call to `isPresent` 
</blockquote></details>

lol yes, IntelliJ is pretty annoying in that respect. Personally I turned of the warning for "acting on an Optional without calling isPresent() first".

My thought was that for a reader without intimate knowledge of `UserAuthInfo`, they would not necessarily know that console user and app engine user are exclusive of each other, so you can check the existence of one to know the type of user the object represents.

But either way is fine with me.
___
*[`core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java` line 434 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAALxbk32xU7nE8tNSd:-NAF7zj35xu5O6SH6K5N:bc8w1x1) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java#L434)):*

no, the user in this test isn't an admin so they only have partial access 

service accounts will be admins as far as I can tell
</blockquote></details>

Sorry I meant for the test above, where you set the user to be admin and also give it the global role of FTE. Does it matter if it's FTE or SUPPORT_AGENT in that case?

gbrodman (commented):
It might be a bit trickier than just one class given that in one instance we might be checking for the ID token and in the other checking for the access token. But yeah. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1744#-:-NAFN3w_9HtapEstiYns:b-ewmn6)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAAJA9SAi4XlVFrorCQ:-NAFNBu198VhAKJMBkCI:br45s2q) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L161)):*

lol yes, IntelliJ is pretty annoying in that respect. Personally I turned of the warning for "acting on an Optional without calling isPresent() first".

My thought was that for a reader without intimate knowledge of `UserAuthInfo`, they would not necessarily know that console user and app engine user are exclusive of each other, so you can check the existence of one to know the type of user the object represents.

But either way is fine with me.
</blockquote></details>

yeah, there's not really a nice built-in way in Java to enforce one or the other, and to retrieve whichever one exists. 
___
*[`core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java` line 434 at r2](https://reviewable.io/reviews/google/nomulus/1744#-NAALxbk32xU7nE8tNSd:-NAFNIcy-bZQsawqy5iZ:b-s5tkg1) ([raw file](https://github.com/google/nomulus/blob/2260f3bca5f438e8bcfa1f5536dc237f1f58f520/core/src/test/java/google/registry/request/auth/AuthenticatedRegistrarAccessorTest.java#L434)):*

Sorry I meant for the test above, where you set the user to be admin and also give it the global role of FTE. Does it matter if it's FTE or SUPPORT_AGENT in that case?
</blockquote></details>

Ah. No, then it doesn't matter. Admin overrides all

PR #1745, opened 2022-08-16
Add the PackagePromotion table

sarahcaseybot:
Technical design for the packaging promotion is detailed in go/package-promo-manual-tiers-design

DO NOT MERGE until flyway files from go/r3pr/1746 are deployed

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1745)


gbrodman (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cYX6DEtJVWRIWFkh-:-N9cYX6DEtJVWRIWFkh0:b-hp63t) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L291)):*
> ```Java
>           getInstance().tokenType != TokenType.PACKAGE
>               || getInstance().allowedClientIds.size() == 1,
>           "PACKAGE tokens must only have one allowed client");
> ```

bit nitty, but I think it's a bit more strict / clear if we say something like "PACKAGE tokens must have exactly one allowed client registrar" (trying to get away from "clientId" anyway) 
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 41 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cX8j2DeMpl1L9yLmv:-N9cX8j2DeMpl1L9yLmw:b-ytr3oo) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L41)):*
> ```Java
> 
>   /** The allocation token string for the package. */
>   @Id String token;
> ```

Unless I'm missing something, instead of having the token be the ID, we should have an autogenerated long ID. That way, we can change the token later if we decide that we need that for some reason. It would also allow us to store the token as a `VKey<AllocationToken>` which seems better anyway 
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 55 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cXe147rcqLBLQuWpe:-N9cXe147rcqLBLQuWpf:bvt540g) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L55)):*
> ```Java
>   Money packagePrice;
> 
>   /** The next billing date of the package. */
> ```

when / why would this ever be null?

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: 6 of 14 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 55 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cXe147rcqLBLQuWpe:-N9hLtLvD6SEmgYd_vD2:b-bfhj3f) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L55)):*

Right now the plan is to have a daily cron job that checks the next billing date and sends an email to the Registry when it is time to bill for that package. It is probably not a good idea to delete these objects when a package promotion contract ends, but we also don't want the cron job to keep alerting us to bill for that package if the set nextBillingDate value has passed. So, when we end a package we could either set the nextBillingDate to END_OF_TIME or null. 

I guess an alternative would be to have the cron job check if the allocation token on that packagePromotion is still active, but that would require an extra read which is probably not efficient. 
</blockquote></details>

Yeah I guess the thing I'm comparing it to in my head is Domain's deletion time, where we set it to be END OF TIME rather than null if the domain is active. It might also make queries easier if we don't have to worry about what null means?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: 6 of 14 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cYX6DEtJVWRIWFkh-:-N9h3S7I-iSbmaNqTjGT:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L291)):*

bit nitty, but I think it's a bit more strict / clear if we say something like "PACKAGE tokens must have exactly one allowed client registrar" (trying to get away from "clientId" anyway) 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 41 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cX8j2DeMpl1L9yLmv:-N9hFG_c6PNmKbPlRnBH:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L41)):*

Unless I'm missing something, instead of having the token be the ID, we should have an autogenerated long ID. That way, we can change the token later if we decide that we need that for some reason. It would also allow us to store the token as a `VKey<AllocationToken>` which seems better anyway 
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 6 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 44 at r2](https://reviewable.io/reviews/google/nomulus/1745#-N9hN4RS2guTLYF4CHht:-N9hN4RT2R31fNyFMo45:be3spt4) ([raw file](https://github.com/google/nomulus/blob/9414a075b76b3a012cccadfc877194c6123e73e5/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L44)):*
> ```Java
>   @Id long packagePromotionId;
> 
>   /** The allocation token string for the package. */
> ```

should be non-null as well

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: 8 of 14 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 55 at r1](https://reviewable.io/reviews/google/nomulus/1745#-N9cXe147rcqLBLQuWpe:-N9lZyFF1jFAkM8ZhKiU:b-74lmgu) ([raw file](https://github.com/google/nomulus/blob/f978904fa73d380a76f1adf90fbd58e7c71a1198/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L55)):*

Yeah I guess the thing I'm comparing it to in my head is Domain's deletion time, where we set it to be END OF TIME rather than null if the domain is active. It might also make queries easier if we don't have to worry about what null means?
</blockquote></details>

Changed to not null
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 44 at r2](https://reviewable.io/reviews/google/nomulus/1745#-N9hN4RS2guTLYF4CHht:-N9lZxWB7Q2wx7pszWom:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9414a075b76b3a012cccadfc877194c6123e73e5/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L44)):*

should be non-null as well 
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 1 of 8 files at r2, 1 of 6 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: all files reviewed, 4 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 45 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5TgymCUn2V2fG9Ryo:-NA5TgymCUn2V2fG9Ryp:b-5y1oa7) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L45)):*
> ```Java
>   @Id long packagePromotionId;
> 
>   /** The allocation token string for the package. */
> ```

The non-null fields like this one should have the `@Column(nullable = false)` annotation to make the generated file accurate. 
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 82 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5UDQ18dCMzRPWvaXM:-NA5UDQ18dCMzRPWvaXN:b-y7ufn8) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L82)):*
> ```Java
>   }
> 
>   public Optional<DateTime> getNextBillingDate() {
> ```

this doesn't have to be an optional now 
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 137 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5Tuyf9s_THrftkByl:-NA5Tuyf9s_THrftkBym:b3c65dj) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L137)):*
> ```Java
> 
>     public Builder setNextBillingDate(@Nullable DateTime nextBillingDate) {
>       if (nextBillingDate == null) {
> ```

Instead of doing it here, just set the default to END_OF_TIME when you declare it, then the input must be non-null
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 144 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5U9wa8VnlcDw8POsn:-NA5U9wa8VnlcDw8POso:b5sgtfh) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L144)):*
> ```Java
>     }
> 
>     public Builder setLastNotificationSent(@Nullable DateTime lastNotificationSent) {
> ```

Is there ever any situation where we'd want this to be null?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: 11 of 14 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 45 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5TgymCUn2V2fG9Ryo:-NA625bx6MF847NTcvuy:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L45)):*

The non-null fields like this one should have the `@Column(nullable = false)` annotation to make the generated file accurate. 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 82 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5UDQ18dCMzRPWvaXM:-NA6266j4KVpyUjz1nbP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L82)):*

this doesn't have to be an optional now 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 137 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5Tuyf9s_THrftkByl:-NA62Krq5g-64D59uhXJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L137)):*

Instead of doing it here, just set the default to END_OF_TIME when you declare it, then the input must be non-null
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 144 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5U9wa8VnlcDw8POsn:-NA69eAt7G06n7Bm1F6Z:b-b6dmrj) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L144)):*

Is there ever any situation where we'd want this to be null?  
</blockquote></details>

Yes, as mentioned in the Active Domains Limit Exceeded section of go/package-promo-manual-tiers-design the lastNotificationSent will be set to null when a tier is upgraded or if a notification has never been sent

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1745)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 144 at r4](https://reviewable.io/reviews/google/nomulus/1745#-NA5U9wa8VnlcDw8POsn:-NA6C5l7CMHw6Ul3e6B3:b9e0y60) ([raw file](https://github.com/google/nomulus/blob/d772ad432fefc5c3486771ad2c339ce985992f9d/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L144)):*

Yes, as mentioned in the Active Domains Limit Exceeded section of go/package-promo-manual-tiers-design the lastNotificationSent will be set to null when a tier is upgraded or if a notification has never been sent
</blockquote></details>

eh yeah I was imagining that if we're setting it to null here, it must be because there was once a non-null `lastNotificationSent` and now it's null, which I can't imagine being the case. I suppose we can't guarantee that we'll never want that though

PR #1755, opened 2022-08-22
Add registry email to bcc for outgoing DNS failure emails

ptkach:
This adds registry support email to bcc for when DNS failure occurs and we send DNS failure emails. This should help us to track how many and what kind of emails are getting sent by us and make sure partners don't get flooded with emails.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1755)

### DIFF
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1320,8 +1320,14 @@ public final class RegistryConfig {
 
     @Provides
     @Config("registrySupportEmail")
-    public static String provideRegistrySupportEmail(RegistryConfigSettings config) {
-      return config.dnsUpdate.registrySupportEmail;
+    public static InternetAddress provideRegistrySupportEmail(RegistryConfigSettings config) {
+      return parseEmailAddress(config.dnsUpdate.registrySupportEmail);
+    }
+
+    @Provides
+    @Config("registryCcEmail")
+    public static InternetAddress provideRegistryCcEmail(RegistryConfigSettings config) {
+      return parseEmailAddress(config.dnsUpdate.registryCcEmail);
     }
 
     @Provides
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -254,5 +254,6 @@ public class RegistryConfigSettings {
     public String dnsUpdateFailEmailBodyText;
     public String dnsUpdateFailRegistryName;
     public String registrySupportEmail;
+    public String registryCcEmail;
   }
 }
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -481,6 +481,7 @@ contactHistory:
 dnsUpdate:
   dnsUpdateFailRegistryName: Example name
   registrySupportEmail: email@example.com
+  registryCcEmail: email@example.com
   # Email subject text template to notify partners after repeatedly failing DNS update
   dnsUpdateFailEmailSubjectText: "[ACTION REQUIRED]: Incomplete DNS Update"
   # Email body text template for failing DNS update that accepts 5 parameters:
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -34,6 +34,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
+import dagger.Lazy;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.dns.DnsMetrics.ActionStatus;
 import google.registry.dns.DnsMetrics.CommitStatus;
@@ -52,14 +53,16 @@ import google.registry.request.Parameter;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
 import google.registry.request.lock.LockHandler;
-import google.registry.ui.server.SendEmailUtils;
 import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import google.registry.util.DomainNameUtils;
+import google.registry.util.EmailMessage;
+import google.registry.util.SendEmailService;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import javax.inject.Inject;
+import javax.mail.internet.InternetAddress;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
 
@@ -109,11 +112,13 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   private final Clock clock;
   private final CloudTasksUtils cloudTasksUtils;
   private final Response response;
-  private final SendEmailUtils sendEmailUtils;
+  private final SendEmailService sendEmailService;
   private final String dnsUpdateFailEmailSubjectText;
   private final String dnsUpdateFailEmailBodyText;
   private final String dnsUpdateFailRegistryName;
-  private final String registrySupportEmail;
+  private final Lazy<InternetAddress> registrySupportEmail;
+  private final Lazy<InternetAddress> registryCcEmail;
+  private final InternetAddress gSuiteOutgoingEmailAddress;
 
   @Inject
   public PublishDnsUpdatesAction(
@@ -129,7 +134,9 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
       @Config("dnsUpdateFailEmailSubjectText") String dnsUpdateFailEmailSubjectText,
       @Config("dnsUpdateFailEmailBodyText") String dnsUpdateFailEmailBodyText,
       @Config("dnsUpdateFailRegistryName") String dnsUpdateFailRegistryName,
-      @Config("registrySupportEmail") String registrySupportEmail,
+      @Config("registrySupportEmail") Lazy<InternetAddress> registrySupportEmail,
+      @Config("registryCcEmail") Lazy<InternetAddress> registryCcEmail,
+      @Config("gSuiteOutgoingEmailAddress") InternetAddress gSuiteOutgoingEmailAddress,
       @Header(APP_ENGINE_RETRY_HEADER) Optional<Integer> appEngineRetryCount,
       @Header(CLOUD_TASKS_RETRY_HEADER) Optional<Integer> cloudTasksRetryCount,
       DnsQueue dnsQueue,
@@ -138,13 +145,13 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
       LockHandler lockHandler,
       Clock clock,
       CloudTasksUtils cloudTasksUtils,
-      SendEmailUtils sendEmailUtils,
+      SendEmailService sendEmailService,
       Response response) {
     this.dnsQueue = dnsQueue;
     this.dnsWriterProxy = dnsWriterProxy;
     this.dnsMetrics = dnsMetrics;
     this.timeout = timeout;
-    this.sendEmailUtils = sendEmailUtils;
+    this.sendEmailService = sendEmailService;
     this.retryCount =
         cloudTasksRetryCount.orElse(
             appEngineRetryCount.orElseThrow(
@@ -165,6 +172,8 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
     this.dnsUpdateFailEmailSubjectText = dnsUpdateFailEmailSubjectText;
     this.dnsUpdateFailRegistryName = dnsUpdateFailRegistryName;
     this.registrySupportEmail = registrySupportEmail;
+    this.registryCcEmail = registryCcEmail;
+    this.gSuiteOutgoingEmailAddress = gSuiteOutgoingEmailAddress;
   }
 
   private void recordActionResult(ActionStatus status) {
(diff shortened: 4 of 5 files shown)


jianglai (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 143 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAAAxiwASmcivejY8Oh:-NAAAxiwASmcivejY8Oi:bh0brhj) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L143)):*
> ```Java
>       Clock clock,
>       CloudTasksUtils cloudTasksUtils,
>       SendEmailService sendEmailService,
> ```

This is probably the better way as `SendEmailUtils` seems to be only used in the UI backend.



_Code quote:_
```Java
 SendEmailService sendEmailService
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 281 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABNQVA352YyIP4kFZ:-NAABNQW4lLsUkn_-qrB:b-nlchtd) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L281)):*
> ```Java
>           String.format(
>               "Could not parse email contact %s to send DNS failure notification", email));
>       return null;
> ```

Is it going to be a problem if we pass `null` to `SendEmailService`?

I think we should just inject `InternetAddress` for `registrySupportEmail` and `registryCcemail`, like how `gSuiteOutgoingEmailAddress` is injected.



_Code quote:_
```Java
return null;
```
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NAABXw43I_yHrBV8jFj:b-4r2mha) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*
> ```Java
>     try {
>       outgoingRegistry = new InternetAddress("outgoing@registry.example");
>     } catch (Exception e) {
> ```

I'm not sure what this does? You could just add `throws Exception` to the function signature. It is OK to not handle exceptions in tests when you know the exception will not be thrown. We do this all the time in tests to save having to write all the try-catch blocks.



_Code quote:_
```Java
    } catch (Exception e) {
      System.out.println("");
    }
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 143 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAAAxiwASmcivejY8Oh:-NAADPHdARJrs1FmfNMQ:b66s4da) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L143)):*

This is probably the better way as `SendEmailUtils` seems to be only used in the UI backend.
</blockquote></details>

Yeah, I agree, with bcc requirement "util" is not a fit anymore. 
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 281 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABNQVA352YyIP4kFZ:-NAAEZ378IgPzWaB1U0k:b-yd5eh1) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L281)):*

Is it going to be a problem if we pass `null` to `SendEmailService`?

I think we should just inject `InternetAddress` for `registrySupportEmail` and `registryCcemail`, like how `gSuiteOutgoingEmailAddress` is injected.
</blockquote></details>

No, it shouldn't be a problem, this is essentially what SendEmailUtils does under the hood when converting 
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NAADkt1AtdiuwNwMLAq:bijk3l9) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*

I'm not sure what this does? You could just add `throws Exception` to the function signature. It is OK to not handle exceptions in tests when you know the exception will not be thrown. We do this all the time in tests to save having to write all the try-catch blocks.
</blockquote></details>

Unfortunately simple "throws" to the function is not gonna do the trick as then it also requires updates across all the places that call this function to either handle or throw and so on. I figured, it's guaranteed not to throw and probably the shortest way to handle it is this.

sarahcaseybot (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 55 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAA_zIp6-RWxQTk8q4Q:-NAA_zIp6-RWxQTk8q4R:bj0v1bf) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L55)):*
> ```Java
> import google.registry.request.auth.Auth;
> import google.registry.request.lock.LockHandler;
> import google.registry.util.*;
> ```

Don't use wildcard imports - see Java style guide: go/java-style#s3.3.1-wildcard-imports
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NAAcno17amVCCNKBNZf:b4dfcos) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*

Unfortunately simple "throws" to the function is not gonna do the trick as then it also requires updates across all the places that call this function to either handle or throw and so on. I figured, it's guaranteed not to throw and probably the shortest way to handle it is this.
</blockquote></details>

I agree with Lai, you should change each of the tests that call createAction to have `throws Exception` in the function signature

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 55 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAA_zIp6-RWxQTk8q4Q:-NAB00EpBsCkdymmfil7:b-886gsx) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L55)):*

Don't use wildcard imports - see Java style guide: go/java-style#s3.3.1-wildcard-imports
</blockquote></details>

Fixed
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NAB-yqK2Clbfc5n2lT0:b-bu91xm) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*

I agree with Lai, you should change each of the tests that call createAction to have `throws Exception` in the function signature
</blockquote></details>

Updated

sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 281 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABNQVA352YyIP4kFZ:-NABRPpF-zljgWQ0592N:botv8xk) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L281)):*

No, it shouldn't be a problem, this is essentially what SendEmailUtils does under the hood when converting 
</blockquote></details>

Still, is there a reason not to inject an `InternetAddress`? We already have a [parseEmailClass](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/config/RegistryConfig.java;drc=6dd96c247a78e8f1ea26a77628d003c794fe71b1;l=1570?q=registryconfig&ss=nomulus%2Fnomulus) method in `RegistryConfig` to do exactly this.
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NABSOD9BtKA3UaCPvNR:b3ovcx5) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*

Updated
</blockquote></details>

You can create the `InternetAccess` in a `@BeforeEach` method (with `throws Exception`) and use it here, this way you don't have to add `throws Exception` to every test methods.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 281 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABNQVA352YyIP4kFZ:-NABpccR91MR3tTJ2-UW:bns2umt) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L281)):*

Still, is there a reason not to inject an `InternetAddress`? We already have a [parseEmailClass](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/config/RegistryConfig.java;drc=6dd96c247a78e8f1ea26a77628d003c794fe71b1;l=1570?q=registryconfig&ss=nomulus%2Fnomulus) method in `RegistryConfig` to do exactly this.
</blockquote></details>

registrySupportEmail is not used as InternetAddress, it's used as a String in email template, so it has to stay as is. Recipients is derived from registrar contacts, which is a String, so transformation in this class is still necessary (unless we expose  parseEmailClass). To sum it up only registryCcemail that can currently be injected as InternetAddress, but on a flip side it will cause a transformation being done ahead of time, wasting computation, unless we do it lazy. Please let me know if you think it's necessary and I will update. 
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NABtOKYCMMEHd525j4X:bdcv6kv) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*

You can create the `InternetAccess` in a `@BeforeEach` method (with `throws Exception`) and use it here, this way you don't have to add `throws Exception` to every test methods.
</blockquote></details>

Good point!

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 281 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABNQVA352YyIP4kFZ:-NAF8iSC29HWka66jXQx:b-l4e3nn) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L281)):*

registrySupportEmail is not used as InternetAddress, it's used as a String in email template, so it has to stay as is. Recipients is derived from registrar contacts, which is a String, so transformation in this class is still necessary (unless we expose  parseEmailClass). To sum it up only registryCcemail that can currently be injected as InternetAddress, but on a flip side it will cause a transformation being done ahead of time, wasting computation, unless we do it lazy. Please let me know if you think it's necessary and I will update. 
</blockquote></details>

It actually might not be a bad idea to inject registrySupportEmail as an `InternetAddress` precisely because it does validation on the address. You can then call `InternetAddress.getAddress()` to get the string.

The transformation should be pretty lightweight, but good point on not wasting resources when we don't have to. We can lazy inject all three of them.



_Code quote:_
```Java
javax.mail.internet.InternetAddress
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 281 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABNQVA352YyIP4kFZ:-NAGp2PG8kRZhZQmnMOI:b7c0akr) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L281)):*

It actually might not be a bad idea to inject registrySupportEmail as an `InternetAddress` precisely because it does validation on the address. You can then call `InternetAddress.getAddress()` to get the string.

The transformation should be pretty lightweight, but good point on not wasting resources when we don't have to. We can lazy inject all three of them.
</blockquote></details>

There's just 2 to inject, which I've now updated as per your suggestion. Done!
___
*[`core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/1755#-NAABXw43I_yHrBV8jFi:-NAGpBTJ7dFrtb-ApyUx:b-896fja) ([raw file](https://github.com/google/nomulus/blob/a9c602e86247cfc1f8d32a7eb5acda1f06252ed2/core/src/test/java/google/registry/dns/PublishDnsUpdatesActionTest.java#L139)):*

Good point!
</blockquote></details>

Done!

jianglai (approved):
Remember to change the corresponding file in the private repo.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1755#-:-NAI2EJz92EMG_mMRNpw:b-q43iho)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #1758, opened 2022-08-24
Prohibit renewals of package domains unless REMOVEPACKAGE token is included

ptkach:


gbrodman (commented):
Wasn't there discussion about using an EPP extension rather than an allocation for this? I'm not sure where that landed. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758#-:-NAGYNJ12alT2OxGeBmY:btcsf8u)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sarahcaseybot)

ptkach (commented):
There was and we agreed that it's likely won't happen since Bruno and Ben wanted it this way, but we agreed to return to this discussion later (I've listed different options in the plan), however since implementation is pretty easy we agreed to move forward with current plan for now

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758#-:-NAGb8sk3C7qerWD71EG:b-to2cvl)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @sarahcaseybot)

sarahcaseybot (changes requested):
Looking at this, I think it would make a lot more sense to use the packageToken EPP extension I am writing in go/r3pr/1760 instead. Perhaps you should hold off on this PR until Ben gets back tomorrow to discuss.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758#-:-NALTXU326srpC6mhNwb:blnumej)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALNXK52BL5bGRT2Whe:-NALNXK52BL5bGRT2Whf:b-xuv8f0) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L158)):*
> ```Java
>   @Inject DomainRenewFlow() {}
> 
>   private EppResponse process(
> ```

Why did you separate this from run()? this seems like an unnecessary abstraction
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 148 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALRFb65cqPHHaZSrFB:-NALRFb65cqPHHaZSrFC:b-gq29y1) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L148)):*
> ```Java
>     }
>     // Static tokens are not stored in DB
>     if (Enums.getIfPresent(StaticTokens.class, token).isPresent()) {
> ```

We should probably add checks somewhere to make sure that allocation tokens of other types can't be made with the same token string as a static token.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 201 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALPn-Z5_6YOwsZt72-:-NALPn-Z5_6YOwsZt720:b9z4vem) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L201)):*
> ```Java
>   }
> 
>   public static void verifyTokenAllowedOnDomain(
> ```

This would not throw anything in the case where a domain is in a package but no allocation token is present on the request. Renew requests should only succeed on package domains if they include the REMOVEPACKAGE token. 

This would also not throw anything in the case where a REMOVEPACKAGE token is included for a non-package domain. We probably should reject in that case as well.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 115 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALSLbq-1hja3moe893:-NALSLbq-1hja3moe894:b-ghtr9q) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L115)):*
> ```Java
>     SINGLE_USE,
>     UNLIMITED_USE,
>     STATIC
> ```

Enum values should be listed in alphabetical order
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1192 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALSuIv6h022A7sR19c:-NALSuIv6h022A7sR19d:bxw32pz) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1192)):*
> ```Java
> 
>   @Test
>   void testFailsToRenewPackageDomainWithoutRemovePackageToken() throws Exception {
> ```

should also have test for no allocation token on a package domain, success with the REMOVEPACKAGE token, and use of the REMOVEPACKAGE domain on a non-package domain

gbrodman (changes requested):
we can review this now but let's maybe wait to merge until we talk it through with Ben and Bruno/Michael 

> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758#-:-NAKd1fE3CQtk5q94pNe:bm9tu2p)* status: 1 of 5 files reviewed, 10 unresolved discussions (waiting on @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKdhOq9595p3mWh7dl:-NAKdhOrAdqbQlzDaa2f:bvwjw1a) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L158)):*
> ```Java
>   @Inject DomainRenewFlow() {}
> 
>   private EppResponse process(
> ```

any particular reason for moving this into a separate function? It doesn't seem like we're calling it anywhere 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 117 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKfTGmEbY3XPG6AyBp:-NAKfTGnE3_OXpR9eU0y:bi9um7i) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L117)):*
> ```Java
>     STATIC
>   }
> ```

Thoughts on using `__REMOVEPACKAGE__` instead of `REMOVEPACKAGE`? It removes any possibility of collision and makes it clear that they're passing in a non-standard token. 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALT4TP8GZDx-j8r1AU:-NALT4TP8GZDx-j8r1AV:b-56myug) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L131)):*
> ```Java
> 
>   /** Static tokens are not stored in DB, but used as a signals */
>   public enum StaticTokens {
> ```

I'm not precisely sure what the best solution is, but having separate structures here, where we check enum values against the input in multiple places, seems not ideal. 

Maybe something like removing the STATIC TokenType, replacing StaticTokens with some sort of "TokenBehavior"-type of name, and then having a map from string (token) to singleton static instances of tokens with particular behavior? 

eg

```
private static final Map<String, AllocationToken> BEHAVIORAL_TOKENS = ImmutableMap.of(
    "__REMOVEPACKAGE__", new AllocationToken.Builder().setTokenBehavior(TokenBehavior.REMOVEPACKAGE).build()); // likely with other stuff in the token builder too

public enum TokenBehavior {
  DEFAULT,
  REMOVEPACKAGE
}

@Column(nullable = false)
TokenBehavior tokenBehavior = TokenBehavior.DEFAULT; // this defaults it
```

then when loading tokens, use the static singleton instance if they provide a matching name. Then everywhere we need, we can just check the TokenBehavior 

thoughts? 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 266 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKvsIn6YvmclGbgJ5X:-NAKvsIn6YvmclGbgJ5Y:b-pjgryz) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L266)):*
> ```Java
>   }
> 
>   public boolean isPackageRemove() {
> ```

If we go the static token route we might end up with several types of static tokens. One method per static token would likely end up too unwieldy, so it should probably just be the job of the caller to check the actual token 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 292 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALMwTs47K1xniED-6S:-NALMwTs47K1xniED-6T:blq0tor) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L292)):*
> ```Java
>     @Override
>     public AllocationToken build() {
>       checkArgument(
> ```

probably want to move this below the token-type-check, then you can just say 

`checkArgument(!getInstance().tokenType.equals(TokenType.STATIC) || Enums.getIfPresent...), "Static allocation...")`

possibly

CydeWeys (commented):
The problem with using EPP extensions is that our partners aren't going to implement them.  It was already hard to get partners onboard with implementing the allocation token extension (which is a proper RFC used by other registries).  Asking them to implement a new custom extension just for this likely isn't going to happen, hence shoehorning it into the allocation token extension.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758#-:-NAP0r6yEwEAA6gS1hTg:bpw2lpm)* status: 1 of 5 files reviewed, 10 unresolved discussions (waiting on @gbrodman, @ptkach, and @sarahcaseybot)

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 0 of 6 files reviewed, 10 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALNXK52BL5bGRT2Whe:-NARB5W59xLCLUXLiZCo:brppg7w) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L158)):*

Why did you separate this from run()? this seems like an unnecessary abstraction
</blockquote></details>

it's laying groundwork for actual REMOVEPACKAGE processing step, which is my next task, for now reverted back
___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 158 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKdhOq9595p3mWh7dl:-NARBJWw-SEu6LIQ_6fR:brppg7w) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L158)):*

any particular reason for moving this into a separate function? It doesn't seem like we're calling it anywhere 
</blockquote></details>

it's laying groundwork for actual REMOVEPACKAGE processing step, which is my next task, for now reverted back
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 148 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALRFb65cqPHHaZSrFB:-NARBcy35TqBP3DUO_Mp:bakw87q) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L148)):*

We should probably add checks somewhere to make sure that allocation tokens of other types can't be made with the same token string as a static token.
</blockquote></details>

Now with a different naming convention and map in place it should be better
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 201 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALPn-Z5_6YOwsZt72-:-NARBnI60cYoImthDtUB:b-886gsx) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L201)):*

This would not throw anything in the case where a domain is in a package but no allocation token is present on the request. Renew requests should only succeed on package domains if they include the REMOVEPACKAGE token. 

This would also not throw anything in the case where a REMOVEPACKAGE token is included for a non-package domain. We probably should reject in that case as well.
</blockquote></details>

Fixed
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 115 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALSLbq-1hja3moe893:-NARBsVR6wnTuyoprd-C:b-b8u6hg) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L115)):*

Enum values should be listed in alphabetical order
</blockquote></details>

Value removed
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 117 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKfTGmEbY3XPG6AyBp:-NARBuyq-mstKp_2ah22:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L117)):*

Thoughts on using `__REMOVEPACKAGE__` instead of `REMOVEPACKAGE`? It removes any possibility of collision and makes it clear that they're passing in a non-standard token. 
</blockquote></details>

Done
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALT4TP8GZDx-j8r1AU:-NARBxI-3GC7qHxit_UP:b-zedsvg) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L131)):*

I'm not precisely sure what the best solution is, but having separate structures here, where we check enum values against the input in multiple places, seems not ideal. 

Maybe something like removing the STATIC TokenType, replacing StaticTokens with some sort of "TokenBehavior"-type of name, and then having a map from string (token) to singleton static instances of tokens with particular behavior? 

eg

```
private static final Map<String, AllocationToken> BEHAVIORAL_TOKENS = ImmutableMap.of(
    "__REMOVEPACKAGE__", new AllocationToken.Builder().setTokenBehavior(TokenBehavior.REMOVEPACKAGE).build()); // likely with other stuff in the token builder too

public enum TokenBehavior {
  DEFAULT,
  REMOVEPACKAGE
}

@Column(nullable = false)
TokenBehavior tokenBehavior = TokenBehavior.DEFAULT; // this defaults it
```

then when loading tokens, use the static singleton instance if they provide a matching name. Then everywhere we need, we can just check the TokenBehavior 

thoughts? 
</blockquote></details>

Great suggestion, done! It doesn't look like tokenBehavior should be stored in DB at the moment though
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 266 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKvsIn6YvmclGbgJ5X:-NARCGPtEXAXUcpmBt_w:b-886gsx) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L266)):*

If we go the static token route we might end up with several types of static tokens. One method per static token would likely end up too unwieldy, so it should probably just be the job of the caller to check the actual token 
</blockquote></details>

Fixed
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 292 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALMwTs47K1xniED-6S:-NARCJ57EjVnmNAMshXO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L292)):*

probably want to move this below the token-type-check, then you can just say 

`checkArgument(!getInstance().tokenType.equals(TokenType.STATIC) || Enums.getIfPresent...), "Static allocation...")`

possibly 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1192 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALSuIv6h022A7sR19c:-NARCLT10fwJj6FvGtHR:b-cut1te) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1192)):*

should also have test for no allocation token on a package domain, success with the REMOVEPACKAGE token, and use of the REMOVEPACKAGE domain on a non-package domain
</blockquote></details>

Added no token and non-package tests, however success will be added as a part of b/243147466 implemenation

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 0 of 8 files reviewed, 16 unresolved discussions (waiting on @CydeWeys, @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 317 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfB3fvDVtGXXR_AAT-:-NAfB3fvDVtGXXR_AAT0:b976f5q) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L317)):*
> ```Java
>     verifyOptionalAuthInfo(authInfo, existingDomain);
>     verifyNoDisallowedStatuses(existingDomain, RENEW_DISALLOWED_STATUSES);
>     AllocationTokenFlowUtils.verifyTokenAllowedOnDomain(existingDomain, allocationToken);
> ```

You can feel free to statically import stuff like this so the line is shorter and thus there's less to read.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 93 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfBHdX3RWZx5JgfTGw:-NAfBHdX3RWZx5JgfTGx:b-a9vyja) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L93)):*
> ```Java
>           .build();
> 
>   private static final Map<String, AllocationToken> BEHAVIORAL_TOKENS =
> ```

The type here should be `ImmutableMap`.  Generally speaking, using the immutable collection classes themselves as the type is preferred, when possible, as it makes it clear from all possible caller sites that it is immutable.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 95 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfC_kz39VgfeHY7_YW:-NAfC_kz39VgfeHY7_YX:b-qa6qyo) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L95)):*
> ```Java
>   private static final Map<String, AllocationToken> BEHAVIORAL_TOKENS =
>       ImmutableMap.of(
>           "__REMOVEPACKAGE__",
> ```

This should be a static final string in the `AllocationToken` class.  I'm sure we'll be referring to it from elsewhere.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 132 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfC0yCDoeqhsXb9c0p:-NAfC0yCDoeqhsXb9c0q:b-ytrkt7) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L132)):*
> ```Java
>   }
> 
>   public enum TokenBehavior {
> ```

What is TokenBehavior supposed to be exactly?  Add documentation.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 134 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfCJ4J4_eEbmJ6YYMR:-NAfCJ4J4_eEbmJ6YYMS:bpwalw4) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L134)):*
> ```Java
>   public enum TokenBehavior {
>     DEFAULT,
>     __REMOVEPACKAGE__
> ```

Why is this `__REMOVEPACKAGE__` instead of `REMOVE_PACKAGE`?  Just because the token string itself is `__REMOVEPACKAGE__` doesn't mean we have to duplicate that for the behavior as well.  As currently written this is a strange name for an enum value.

You also might consider not having behavior and the token text be separate at all, as this is kind of duplicating the same information in two separate places.  I.e. if the token name begins with an underscore, then it is necessarily a static token, and then you can just look up the value of it in a map which also tells you the behavior, rather than having to add the behavior as a separate field on the token?  Considering it's static data, it doesn't need to be persisted to the DB, yet adding it as a field will do precisely that.

So I'm thinking something like:

`ImmutableMap<String, TokenBehavior> STATIC_TOKEN_BEHAVIORS = ImmutableMap.of("__REMOVEPACKAGE__", TokenBehavior.REMOVE_PACKAGE)`

I think that might actually be best, with the least number of schema changes.  We don't have a use case for token behaviors on non-static tokens, right?
___
*[`core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml` line 15 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfBfqZ-K-qONGcZRLi:-NAfBfqZ-K-qONGcZRLj:blyr8a3) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml#L15)):*
> ```XML
>           xmlns:allocationToken=
>               "urn:ietf:params:xml:ns:allocationToken-1.0">
>         __REMOVEPACKAGE__
> ```

You can parameterize this as `%TOKEN%` for the XML testdata, and then you don't need both this file and domain_renew_remove_package_token_missing.xml.

They're both just passing a token on the renew command, and that token can be specified in the unit tests as a parameter.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 0 of 8 files reviewed, 15 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 317 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfB3fvDVtGXXR_AAT-:-NAfbs9A8cuYtUl8gPy8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L317)):*

You can feel free to statically import stuff like this so the line is shorter and thus there's less to read.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 93 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfBHdX3RWZx5JgfTGw:-NAfbwGb5x3JmI0l_aCd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L93)):*

The type here should be `ImmutableMap`.  Generally speaking, using the immutable collection classes themselves as the type is preferred, when possible, as it makes it clear from all possible caller sites that it is immutable.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 95 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfC_kz39VgfeHY7_YW:-NAfc-w13bNxIwVkTg4t:b-896fix) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L95)):*

This should be a static final string in the `AllocationToken` class.  I'm sure we'll be referring to it from elsewhere.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 132 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfC0yCDoeqhsXb9c0p:-NAfc14RB_VVwCiD2eU8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L132)):*

What is TokenBehavior supposed to be exactly?  Add documentation.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 134 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfCJ4J4_eEbmJ6YYMR:-NAfc1UF3599huMRBfOP:b-crbwb2) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L134)):*

Why is this `__REMOVEPACKAGE__` instead of `REMOVE_PACKAGE`?  Just because the token string itself is `__REMOVEPACKAGE__` doesn't mean we have to duplicate that for the behavior as well.  As currently written this is a strange name for an enum value.

You also might consider not having behavior and the token text be separate at all, as this is kind of duplicating the same information in two separate places.  I.e. if the token name begins with an underscore, then it is necessarily a static token, and then you can just look up the value of it in a map which also tells you the behavior, rather than having to add the behavior as a separate field on the token?  Considering it's static data, it doesn't need to be persisted to the DB, yet adding it as a field will do precisely that.

So I'm thinking something like:

`ImmutableMap<String, TokenBehavior> STATIC_TOKEN_BEHAVIORS = ImmutableMap.of("__REMOVEPACKAGE__", TokenBehavior.REMOVE_PACKAGE)`

I think that might actually be best, with the least number of schema changes.  We don't have a use case for token behaviors on non-static tokens, right?
</blockquote></details>

I've marked it as transient to minimize schema changes, but what you propose should minimize it even further. Done.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml` line 15 at r4](https://reviewable.io/reviews/google/nomulus/1758#-NAfBfqZ-K-qONGcZRLi:-NAfcavy1mqy5YW_9Ac8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/42b8fc06c9b904dfad88d1175155a322dfc0d4aa/core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml#L15)):*

You can parameterize this as `%TOKEN%` for the XML testdata, and then you don't need both this file and domain_renew_remove_package_token_missing.xml.

They're both just passing a token on the renew command, and that token can be specified in the unit tests as a parameter.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 0 of 8 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1219 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjm7G-3OH6yi_flWYu:-NAjm7G-3OH6yi_flWYv:b3kre) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1219)):*
> ```Java
> 
>   @Test
>   void testFaisMissingRemovePackageToken() throws Exception {
> ```

?



_Code quote:_
```Java
testFais
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 0 of 8 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1219 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjm7G-3OH6yi_flWYu:-NAjnutaC7ltiNuWxZFX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1219)):*

?
</blockquote></details>

Done.

CydeWeys (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 0 of 8 files reviewed, 16 unresolved discussions (waiting on @CydeWeys, @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjoQ3x6IAqOkwRAb1b:-NAjoQ3x6IAqOkwRAb1c:bej0rg4) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*
> ```Java
>       throws EppException {
> 
>     // Non default tokens do not require validation
> ```

Comment can be reworded for slightly better clarity.  Something like "Only tokens with default behavior require validation".
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 147 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjo7IA2gPJzP8hW_o6:-NAjo7IA2gPJzP8hW_o7:bbjho9i) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L147)):*
> ```Java
> 
>     Optional<AllocationToken> maybeTokenEntity =
>         AllocationToken.maybeGetStaticTokenInstance(token).isPresent()
> ```

You can rewrite this more tersely, without needing the ternary, by using `Optional`'s `.map()` and/or `.orElse()` functions.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 207 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjnL8QBoVYoyO-Bo_-:-NAjnL8QBoVYoyO-Bo_0:b8s2mkg) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L207)):*
> ```Java
>             && TokenBehavior.REMOVE_PACKAGE.equals(allocationToken.get().getTokenBehavior()))) {
>       throw domain.getCurrentPackageToken().isPresent()
>           // Only REMOVEPACKAGE token allowed when domain is in a package
> ```

Two different inline comments inside a ternary is a little unusual.  If you just place both comments together on a single line before the statement, then it's more legible.

Though honestly, the names of those exception classes are so complete that I don't think any comments are necessary here anyway.  It's quite self-explanatory.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 266 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjnAuc3u5psXN19xcc:-NAjnAuc3u5psXN19xcd:b-ej1ew1) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L266)):*
> ```Java
>       extends AssociationProhibitsOperationException {
>     MissingRemovePackageTokenOnPackageDomainException() {
>       super("Not allowed to renew package promo domain if REMOVEPACKAGE token is not present");
> ```

We'll probably want to coordinate here with Bruno on exact wording.  As a tentative guess, I would just call it a "package domain" (no "promo").

Also, the error message should be more along the lines of "Domains that are inside packages cannot be explicitly renewed".  Just because they've sent a renew command doesn't mean that what they really wanted to do was have the domain exit the package.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 134 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjn0ARBr1HqHtS4cu5:-NAjn0ARBr1HqHtS4cu6:b-eij755) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L134)):*
> ```Java
> 
>   /**
>    * DEFAULT tokens have standard behavior for Package promo tokens, they follow the same
> ```

This Javadoc should explain token behavior generally.  Then, there should be a separate javadoc individually on the tokens below explaining what they do.

sarahcaseybot (changes requested):
> Reviewed 1 of 2 files at r4, 6 of 7 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: all files reviewed, 14 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ptkach)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 318 at r6](https://reviewable.io/reviews/google/nomulus/1758#-NAjueoFEQBDphu1puPv:-NAjueoFEQBDphu1puPw:b-9q6gf5) ([raw file](https://github.com/google/nomulus/blob/0f3f7d8372777a29a43a9e503dfaaf20a139010c/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L318)):*
> ```Java
>     verifyOptionalAuthInfo(authInfo, existingDomain);
>     verifyNoDisallowedStatuses(existingDomain, RENEW_DISALLOWED_STATUSES);
>     verifyTokenAllowedOnDomain(existingDomain, allocationToken);
> ```

It may be useful to add a comment here noting that right now the only allocation token that may do something in this flow is the REMOVEPACKAGE token. All other tokens will either throw an exception or result in a no-op.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 148 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALRFb65cqPHHaZSrFB:-NAjxqJ39tf5a8ecd3UR:bgtshe5) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L148)):*

Now with a different naming convention and map in place it should be better
</blockquote></details>

I still think there needs t o be a check added in the AllocationToken builder that checks if a token string exists in the STATIC_TOKEN_BEHAVIORS map. The way the code is now, I still think that creating a normal default behavior token with the token string "__REMOVEPACKAGE__" would succeed. 
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1193 at r6](https://reviewable.io/reviews/google/nomulus/1758#-NAjzSpxE74sewB7k3Ff:-NAjzSpxE74sewB7k3Fg:b-8qv3qt) ([raw file](https://github.com/google/nomulus/blob/0f3f7d8372777a29a43a9e503dfaaf20a139010c/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1193)):*
> ```Java
> 
>   @Test
>   void testFailsPresentInvalidRemovePackageToken() throws Exception {
> ```

This is a confusing test title and took me a few minutes to understand what this was testing. I think a better name would be something like "testFailsPackageDomainInvalidAllocationToken"
___
*[`core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml` line 15 at r6](https://reviewable.io/reviews/google/nomulus/1758#-NAk-huL3y7M_FH2QMeA:-NAk-huL3y7M_FH2QMeB:buthtf6) ([raw file](https://github.com/google/nomulus/blob/0f3f7d8372777a29a43a9e503dfaaf20a139010c/core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml#L15)):*
> ```XML
>           xmlns:allocationToken=
>               "urn:ietf:params:xml:ns:allocationToken-1.0">
>         %TOKEN%
> ```

Since this file may have a token other than the REMOVEPACKAGE token, I think the existing domain_renew_allocationtoken.xml file should be used instead. Changing that token to be parameterized.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 3 of 9 files reviewed, 14 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 318 at r6](https://reviewable.io/reviews/google/nomulus/1758#-NAjueoFEQBDphu1puPv:-NAkDB2kB85ZTlHlBELY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f3f7d8372777a29a43a9e503dfaaf20a139010c/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L318)):*

It may be useful to add a comment here noting that right now the only allocation token that may do something in this flow is the REMOVEPACKAGE token. All other tokens will either throw an exception or result in a no-op.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 148 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALRFb65cqPHHaZSrFB:-NAkDE7ZA-ywrYwDn5hH:bws51vg) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L148)):*

I still think there needs t o be a check added in the AllocationToken builder that checks if a token string exists in the STATIC_TOKEN_BEHAVIORS map. The way the code is now, I still think that creating a normal default behavior token with the token string "__REMOVEPACKAGE__" would succeed. 
</blockquote></details>

Well  __REMOVEPACKAGE__  is considered static and complies with rules for static tokens. REMOVEPACKAGE token is no different from any other token from AllocationToken standpoint so it will get default behavior (if even it gets created) as any other token. I believe the point is to make sure that __REMOVEPACKAGE__ token behaves specially, which looks like it does
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjoQ3x6IAqOkwRAb1b:-NAkDDRI5aDNS0wAAsnE:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*

Comment can be reworded for slightly better clarity.  Something like "Only tokens with default behavior require validation".
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 147 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjo7IA2gPJzP8hW_o6:-NAkHg5o0-0OkOTOl9KG:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L147)):*

You can rewrite this more tersely, without needing the ternary, by using `Optional`'s `.map()` and/or `.orElse()` functions.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 207 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjnL8QBoVYoyO-Bo_-:-NAkHhj_-YIej30N3J3e:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L207)):*

Two different inline comments inside a ternary is a little unusual.  If you just place both comments together on a single line before the statement, then it's more legible.

Though honestly, the names of those exception classes are so complete that I don't think any comments are necessary here anyway.  It's quite self-explanatory.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 266 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjnAuc3u5psXN19xcc:-NAkHiyVAi1LkwYu7r5d:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L266)):*

We'll probably want to coordinate here with Bruno on exact wording.  As a tentative guess, I would just call it a "package domain" (no "promo").

Also, the error message should be more along the lines of "Domains that are inside packages cannot be explicitly renewed".  Just because they've sent a renew command doesn't mean that what they really wanted to do was have the domain exit the package.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 134 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjn0ARBr1HqHtS4cu5:-NAkHjiX4jORVxobLluY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L134)):*

This Javadoc should explain token behavior generally.  Then, there should be a separate javadoc individually on the tokens below explaining what they do.
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1193 at r6](https://reviewable.io/reviews/google/nomulus/1758#-NAjzSpxE74sewB7k3Ff:-NAkHljMEzqKSI5zDhuq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f3f7d8372777a29a43a9e503dfaaf20a139010c/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1193)):*

This is a confusing test title and took me a few minutes to understand what this was testing. I think a better name would be something like "testFailsPackageDomainInvalidAllocationToken"
</blockquote></details>

Done.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml` line 15 at r6](https://reviewable.io/reviews/google/nomulus/1758#-NAk-huL3y7M_FH2QMeA:-NAkHn2i25j_xFzSQTxi:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f3f7d8372777a29a43a9e503dfaaf20a139010c/core/src/test/resources/google/registry/flows/domain/domain_renew_remove_package_token.xml#L15)):*

Since this file may have a token other than the REMOVEPACKAGE token, I think the existing domain_renew_allocationtoken.xml file should be used instead. Changing that token to be parameterized. 
</blockquote></details>

Done.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 3 of 9 files reviewed, 14 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 148 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALRFb65cqPHHaZSrFB:-NAkIfuK9ckpmIUy7yHr:bzgj933) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L148)):*

Well  __REMOVEPACKAGE__  is considered static and complies with rules for static tokens. REMOVEPACKAGE token is no different from any other token from AllocationToken standpoint so it will get default behavior (if even it gets created) as any other token. I believe the point is to make sure that __REMOVEPACKAGE__ token behaves specially, which looks like it does
</blockquote></details>

Just now realized that markup has made your and then mine comment confusing. Anyhow getTokenBehavior https://github.com/google/nomulus/blob/0de29b8862c338ead70f2ce3000dac89da1debd3/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L292 should prevent it from reporting DEFAULT.

gbrodman (changes requested):
> Reviewed 1 of 2 files at r4, 2 of 7 files at r5, 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: all files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjoQ3x6IAqOkwRAb1b:-NAkLWzpB-wqB5Dh3hUB:b-byjq4s) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*

Done.
</blockquote></details>

i don't think this block is necessary at all given that tokens with particular behavior will never have a set of registrar IDs specified 
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 147 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjo7IA2gPJzP8hW_o6:-NAkTcZ_B6EYxn6ZtdDC:b8tdufx) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L147)):*

Done.
</blockquote></details>

i'm a bit confused, shouldn't this just be adding one step before the normal maybe-get-token check?
```
Optional<AllocationToken> maybeTokenEntity = AllocationToken.maybeGetStaticInstance(token);
if (maybeTokenEntity.isPresent()) {
  return maybeTokenEntity.get();
}
maybeTokenEntity = tm().transact(() -> tm().loadByKeyIfPresent(VKey.create(AllocationToken.class, token)));
```
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 207 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjnL8QBoVYoyO-Bo_-:-NAkUU6r4d-IXHIlEIIm:b-2ptj9p) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L207)):*

Done.
</blockquote></details>

I think the ternary and multi-part if statement is making this a bit harder to read, even if it reduces the number of lines 

could do something like 
```
boolean hasRemovePackageToken =....
boolean domainHasPackageToken = ...
if (hasRemovePackageToken && !domainHasPackageToken) {
  throw new ....;
else if (!hasRemovePackageToken && domainHasPackageToken) {
  throw new....
}

or something? 
``` 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALT4TP8GZDx-j8r1AU:-NAkW2d-E13jA8e-dfh9:bvjnbdb) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L131)):*

Great suggestion, done! It doesn't look like tokenBehavior should be stored in DB at the moment though
</blockquote></details>

We shouldn't store the token behavior in the DB but we can store it on the token as a `@Transient` field -- defaults to DEFAULT and can be set in the builder. That way, `getTokenBehavior()` can just return the field. 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 266 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKvsIn6YvmclGbgJ5X:-NAkVxxk-qTH2tyTflAL:b9hv3ht) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L266)):*

Fixed
</blockquote></details>

(referenced in other comment)
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1258 at r7](https://reviewable.io/reviews/google/nomulus/1758#-NAkWeVGDcLUM1tYMUD9:-NAkWeVGDcLUM1tYMUDA:b-ol8w27) ([raw file](https://github.com/google/nomulus/blob/0de29b8862c338ead70f2ce3000dac89da1debd3/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1258)):*
> ```Java
>         ImmutableMap.of("DOMAIN", "example.tld", "YEARS", "2", "TOKEN", "__REMOVEPACKAGE__"));
> 
>     EppException thrown =
> ```

no need to store the exception if we're not testing the message (which we don't need to for this type of exception)

same above

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: all files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALT4TP8GZDx-j8r1AU:-NAkZf3v7WJpswUfREAP:bq85er5) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L131)):*

We shouldn't store the token behavior in the DB but we can store it on the token as a `@Transient` field -- defaults to DEFAULT and can be set in the builder. That way, `getTokenBehavior()` can just return the field. 
</blockquote></details>

actually Pavlo points out that this is kinda conflicting with @cydeweys 's comment -- if we do it this way we add a final field and a private builder-setter, but we can remove the odd STATIC_TOKEN_BEHAVIORS map which I think might be worth it. Thoughts, Ben?

ptkach (commented):
> Reviewed 1 of 2 files at r4, 6 of 6 files at r7.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 7 of 9 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjoQ3x6IAqOkwRAb1b:-NAoFEXh51g6MDA_nYsV:b-iw3qzi) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*

i don't think this block is necessary at all given that tokens with particular behavior will never have a set of registrar IDs specified 
</blockquote></details>

This block covers all the conditions applicable to DEFAULT tokens, so we skip non-default tokens verifications and assume they are valid. 



_Code quote:_
```Java
Only tokens with default behavior require validation
```
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 147 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjo7IA2gPJzP8hW_o6:-NAoFmqG9Q5z510dqBrE:bx97zps) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L147)):*

i'm a bit confused, shouldn't this just be adding one step before the normal maybe-get-token check?
```
Optional<AllocationToken> maybeTokenEntity = AllocationToken.maybeGetStaticInstance(token);
if (maybeTokenEntity.isPresent()) {
  return maybeTokenEntity.get();
}
maybeTokenEntity = tm().transact(() -> tm().loadByKeyIfPresent(VKey.create(AllocationToken.class, token)));
```
</blockquote></details>

A little of context here - I used to have very similar written as ternary operator here (no early return for Static token)
```
    Optional<AllocationToken> maybeTokenEntity =         
            AllocationToken.maybeGetStaticTokenInstance(token).isPresent()
            ? AllocationToken.maybeGetStaticTokenInstance(token)
            : tm().transact(() -> tm().loadByKeyIfPresent(VKey.create(AllocationToken.class, token)));
```
Per @cydeweys  request I replaced above with orElseGet, which was the only option to avoid unnecessary evaluation of tm().transact I could think of. However I agree that if else with early return is more readable and the most efficient here. Updated.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 207 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjnL8QBoVYoyO-Bo_-:-NAoJOOZ4n2YeIPJ95GH:b8vah29) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L207)):*
> boolean domainHasPackageToken = ..
Sure, updated
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALT4TP8GZDx-j8r1AU:-NAoJv518Xp3_oZaVbnF:b74za2f) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L131)):*
> We shouldn't store the token behavior in the DB but we can store it on the token as a @Transient field -- defaults to DEFAULT and can be set in the builder. That way, getTokenBehavior() can just return the field.

That's exactly what I used to have there. This logic has been through multiple updates already, so I would probably keep it as it, since it doesn't drastically affect the performance or maintainability 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 266 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NAKvsIn6YvmclGbgJ5X:-NAoLlzyAyb4mmALEMNp:b-c8uuup) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L266)):*

(referenced in other comment)
</blockquote></details>

as mentioned in the question in context, for now it can stay as is. 



_Code quote (from core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java):_
```Java
RemovePackageTokenOnNonPackageDomainExc
```
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1258 at r7](https://reviewable.io/reviews/google/nomulus/1758#-NAkWeVGDcLUM1tYMUD9:-NAoMUHR1kIA2ZEZLZN4:b3lekcl) ([raw file](https://github.com/google/nomulus/blob/0de29b8862c338ead70f2ce3000dac89da1debd3/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1258)):*

no need to store the exception if we're not testing the message (which we don't need to for this type of exception)

same above
</blockquote></details>

sorry, I don't follow the request, Anyhow, I've added ```marshalsToXml()``` to my tests to match the rest of similar test suits in the file. Hope this should cover your concern.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 7 of 9 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 134 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjn0ARBr1HqHtS4cu5:-NAoPHlv8UDHmZsHPqGl:b-42y9vk) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L134)):*

Done.
</blockquote></details>

`DEFAULT` tokens are just normal tokens, right?  Not having anything to do with package promo tokens?  The description here can be simpler.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 144 at r7](https://reviewable.io/reviews/google/nomulus/1758#-NAoPXf70qkI6K-j7hPH:-NAoPXf70qkI6K-j7hPI:ba4vgf9) ([raw file](https://github.com/google/nomulus/blob/0de29b8862c338ead70f2ce3000dac89da1debd3/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L144)):*
> ```Java
>     DEFAULT,
>     /**
>      * REMOVE_PACKAGE is a special kind of token that behaves like a signal to the system to trigger
> ```

Nit, but, it's not _like_ a signal, it is _the_ signal.  Also I think we need to tighten up terminology here; I don't think "promotional package" is quite right.  These aren't promotional.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 6 of 9 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, @ptkach, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 134 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjn0ARBr1HqHtS4cu5:-NAoTozB1tPBCOI8nt97:b-aa4f3b) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L134)):*

`DEFAULT` tokens are just normal tokens, right?  Not having anything to do with package promo tokens?  The description here can be simpler.
</blockquote></details>

Updated similarly to other places in a codebase 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 144 at r7](https://reviewable.io/reviews/google/nomulus/1758#-NAoPXf70qkI6K-j7hPH:-NAoTspY2zuhBggs8msK:b5338qf) ([raw file](https://github.com/google/nomulus/blob/0de29b8862c338ead70f2ce3000dac89da1debd3/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L144)):*

Nit, but, it's not _like_ a signal, it is _the_ signal.  Also I think we need to tighten up terminology here; I don't think "promotional package" is quite right.  These aren't promotional.
</blockquote></details>

Promotional removed, text updated.

gbrodman (approved):
> Reviewed 2 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: all files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjoQ3x6IAqOkwRAb1b:-NAo_rls1B-F4qvgckyN:bfa1h9x) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*

This block covers all the conditions applicable to DEFAULT tokens, so we skip non-default tokens verifications and assume they are valid. 
</blockquote></details>

Sure, but I don't think any of these checks would ever trigger for non-DEFAULT tokens so I don't think the branching is necessary 
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 147 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjo7IA2gPJzP8hW_o6:-NAoxmyk6yfiyuZ3Am-y:b-j1iocg) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L147)):*

A little of context here - I used to have very similar written as ternary operator here (no early return for Static token)
```
    Optional<AllocationToken> maybeTokenEntity =         
            AllocationToken.maybeGetStaticTokenInstance(token).isPresent()
            ? AllocationToken.maybeGetStaticTokenInstance(token)
            : tm().transact(() -> tm().loadByKeyIfPresent(VKey.create(AllocationToken.class, token)));
```
Per @cydeweys  request I replaced above with orElseGet, which was the only option to avoid unnecessary evaluation of tm().transact I could think of. However I agree that if else with early return is more readable and the most efficient here. Updated.

</blockquote></details>

yeah, there's a famous Testing-on-the-Toilet about early return being better, https://engdoc.corp.google.com/eng/doc/tott/episodes/455.md?cl=head 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1758#-NALT4TP8GZDx-j8r1AU:-NAoxze64tV9HO2cye2Z:baom5ku) ([raw file](https://github.com/google/nomulus/blob/5386d760ae7efd7cfd4b5c68b41461b3efb37a48/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L131)):*

> We shouldn't store the token behavior in the DB but we can store it on the token as a @Transient field -- defaults to DEFAULT and can be set in the builder. That way, getTokenBehavior() can just return the field.

That's exactly what I used to have there. This logic has been through multiple updates already, so I would probably keep it as it, since it doesn't drastically affect the performance or maintainability 
</blockquote></details>

yeah if ben would prefer then that's fine, we can do this -- i wouldn't want to block you on a disagreement here 
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 1258 at r7](https://reviewable.io/reviews/google/nomulus/1758#-NAkWeVGDcLUM1tYMUD9:-NAoy3So9wmmQiIequEx:b-2vzmib) ([raw file](https://github.com/google/nomulus/blob/0de29b8862c338ead70f2ce3000dac89da1debd3/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L1258)):*

sorry, I don't follow the request, Anyhow, I've added ```marshalsToXml()``` to my tests to match the rest of similar test suits in the file. Hope this should cover your concern. 
</blockquote></details>

yeah, it's moot now since you're using it, but before you were storing `thrown` as an object which wasn't necessary since it wasn't being used

sarahcaseybot (changes requested):
> Reviewed 3 of 6 files at r7, 2 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @ptkach)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 205 at r9](https://reviewable.io/reviews/google/nomulus/1758#-NApJDtv6Kxn7OWlME10:-NApJDtw7jQlF-X09YBa:bhb0zo4) ([raw file](https://github.com/google/nomulus/blob/2685d238704efe4a22dac43525b9f92a8deb8926/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L205)):*
> ```Java
>       Domain domain, Optional<AllocationToken> allocationToken) throws EppException {
> 
>     boolean hasRemovePackageToken = domain.getCurrentPackageToken().isPresent();
> ```

I think these variable names are flipped. It's called hasRemovePackageToken but is checking if the domain has a package token. Opposite for the one below. This also makes the error messages being thrown below confusing if(hasRemovePackageToken) throw MissingRemovePackageToken...

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjoQ3x6IAqOkwRAb1b:-NApQEce4aL6Q2TCNFfO:b-vyxb3z) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*

Sure, but I don't think any of these checks would ever trigger for non-DEFAULT tokens so I don't think the branching is necessary 
</blockquote></details>

Oh, I see what you mean. Yes, you are correct, that non default won't trigger any of the checks (at the moment), however if in the future any of the checks changes it might affect non-default tokens in rather unexpected way. So I guess it'd make sense here to keep the block "if" to make intention explicit. 
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 147 at r5](https://reviewable.io/reviews/google/nomulus/1758#-NAjo7IA2gPJzP8hW_o6:-NApQoyp9_sRE1ABKzXA:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ca5ed5b95aeb0e6cc4a3e72323d82309d77b7ced/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L147)):*

yeah, there's a famous Testing-on-the-Toilet about early return being better, https://engdoc.corp.google.com/eng/doc/tott/episodes/455.md?cl=head 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 205 at r9](https://reviewable.io/reviews/google/nomulus/1758#-NApJDtv6Kxn7OWlME10:-NApR1I86IhqgKjsKzSp:be8tl6p) ([raw file](https://github.com/google/nomulus/blob/2685d238704efe4a22dac43525b9f92a8deb8926/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L205)):*

I think these variable names are flipped. It's called hasRemovePackageToken but is checking if the domain has a package token. Opposite for the one below. This also makes the error messages being thrown below confusing if(hasRemovePackageToken) throw MissingRemovePackageToken...
</blockquote></details>

Good point, still worked, but naming was off. Fixed

sarahcaseybot (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1758)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

PR #1766, opened 2022-08-30
Fix a few DB issues with the User class

gbrodman:
- Create a sequence to generate IDs for the user (this allows us to have
  Long ID types so that Hibernate can autogenerate IDs)
- Add an update timestamp column so we can extend BackupGroupRoot
- Add a restriction that there can't be multiple users with the same
  email address

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1766)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1766)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1769, opened 2022-09-01
Add mutating commands for PackagePromotion

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1769)


gbrodman (changes requested):
> Reviewed 3 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 3 of 7 files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 34 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzZCFy9M-kDT1Bb7vN:-NAzZCFy9M-kDT1Bb7vO:bnloq5l) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L34)):*
> ```Java
>       names = "--max_domains",
>       description = "Maximum concurrent active domains allowed in the package",
>       required = true)
> ```

these can't / shouldn't all be required if this is a generic "create or update" command, it doesn't make sense for updates 
___
*[`core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java` line 29 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAza4pCDW9D8zCdW_El:-NAza4pCDW9D8zCdW_Em:br70g6) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java#L29)):*
> ```Java
> 
> /** Command to create a PackagePromotion */
> @Parameters(separators = " =", commandDescription = "Create new package(s)")
> ```

should have a more descriptive description, what type of packages / what do they do? 
___
*[`core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java` line 31 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzaG9oDdGsAo0OtXVT:-NAzaG9oDdGsAo0OtXVU:bj6e4ww) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java#L31)):*
> ```Java
> @Parameters(separators = " =", commandDescription = "Create new package(s)")
> public final class CreatePackagePromotionCommand extends CreateOrUpdatePackagePromotionCommand
>     implements CommandWithRemoteApi {
> ```

just have the abstract class implement this, then the subclasses don't need it



_Code quote:_
```Java
CommandWithRemoteApi
```
___
*[`core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java` line 36 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzeKt02PFrNJmo4aB5:-NAzeKt02PFrNJmo4aB6:b-832ns0) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java#L36)):*
> ```Java
>   @Override
>   PackagePromotion getOldPackagePromotion(String tokenString) {
>     getAllocationToken(tokenString);
> ```

not doing anything with this line? 
___
*[`core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java` line 62 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzmR4ECj7PxQ8Tk9WJ:-NAzmR4FDpC9AV6Zvh2P:b6306cu) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java#L62)):*
> ```Java
>   @Nullable
>   @Override
>   DateTime getLastNotificationSent(String token) {
> ```

instead of doing this, we can do something like `CreateOrUpdateTldCommand#premiumListName` with an Optional
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 42 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzmccLEW37qHqK6Xjb:-NAzmccLEW37qHqK6Xjc:b-7fsh3j) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L42)):*
> ```Java
>   @Override
>   PackagePromotion getOldPackagePromotion(String tokenString) {
>     getAllocationToken(tokenString);
> ```

also not doing anything with this line?

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 4 of 7 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 27 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOfkm43XNXIP5b0de2:-NBOfkm43XNXIP5b0de3:b-xdmb1l) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L27)):*
> ```Java
> /** Shared base class for commands to create or update a PackagePromotion object. */
> abstract class CreateOrUpdatePackagePromotionCommand extends MutatingCommand
>     implements CommandWithRemoteApi {
> ```

MutatingCommand already implements this 
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 71 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOg7SF6vdl1AKI3MSg:-NBOg7SF6vdl1AKI3MSh:bhg5qif) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L71)):*
> ```Java
>               : oldPackage.asBuilder();
> 
>       builder.setMaxDomains(maxDomains);
> ```

all (?) of these might not be present, need to check for that before assigning them to the new object

Need tests for that as well -- this should be triggered fairly obviously by tests that test setting only some of the fields. 
___
*[`core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java` line 62 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzmR4ECj7PxQ8Tk9WJ:-NBOiRbXE4Vh_jOZx3uw:b-evmn6k) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java#L62)):*

I don't think that makes sense as I don't want the update command to be able to manually set the lastNotificationSent date to a specific date. I just want it to be able to clear it on the update command. 
</blockquote></details>

yeah, I see you have the clear_last_notification_sent flag below 

instead of doing this then, you can just (possibly) set the last-notification-sent flag to null in Update* if that's triggered. Then you don't need this other method. 
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 48 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOjg1S-dtBmtHOHjpG:-NBOjg1S-dtBmtHOHjpH:b-331jwy) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L48)):*
> ```Java
>   }
> 
>   @Override
> ```

1. this is the same method in both classes, shouldn't be abstract 
2. can inline the call to .get
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOjq_q6TQYJ1DMsJej:-NBOjq_q6TQYJ1DMsJek:bpxfr8e) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L54)):*
> ```Java
>   }
> 
>   private Optional<AllocationToken> checkAllocationToken(String tokenString) {
> ```

why is this method different between the two implementations, why does it need to be abstract?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 4 of 7 files reviewed, 9 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 33 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT3itkBONgMb_qZCQ-:-NBT3itkBONgMb_qZCQ0:b7j44v6) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L33)):*
> ```Java
> 
>   @Parameter(
>       names = "--max_domains",
> ```

You might consider short single-letter flags for all of these parameters as well.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 48 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT2aomC3IE5T6-sU5C:-NBT2aomC3IE5T6-sU5D:b-7ree4s) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L48)):*
> ```Java
>       names = "--next_billing_date",
>       description = "The next date that the package should be billed for its annual fee")
>   DateTime nextBillingDate;
> ```

It seems like this is a `DateTime`, not a `Date`, yet it is called a date in the command.  It should probably actually be a `Date`, so that you don't have to unnecessarily specify a time when setting it.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT2rhO5fMQShzHI5mA:-NBT2rhO5fMQShzHI5mB:b-6oypjg) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L50)):*
> ```Java
>   DateTime nextBillingDate;
> 
>   /** Returns the existing PackagePromotion (for update) or null, (for creates). */
> ```

I'm not seeing why this should be an abstract method given how it works differently across the two commands.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT32vgACL02Rv0MC2F:-NBT32vgACL02Rv0MC2G:b-tkgpdr) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L54)):*
> ```Java
>   abstract PackagePromotion getOldPackagePromotion(String token);
> 
>   abstract AllocationToken getAllocationToken(String token);
> ```

I don't understand why this is an abstract method (and maybe also the next one as well).  Seems like these should just be implemented normally?  It is possible that you're relying too much on inheritance in implementing the functionality here.  As Effective Java says, prefer composition over inheritance.  Inheritance is useful for sharing the common command-line parameters across the commands, but not much else.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 2 of 7 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 27 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOfkm43XNXIP5b0de2:-NBOmPKYENO4GbM1vNfG:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L27)):*

MutatingCommand already implements this 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 33 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT3itkBONgMb_qZCQ-:-NBTWm6e04mZv7WMgBD0:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L33)):*

You might consider short single-letter flags for all of these parameters as well.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 48 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT2aomC3IE5T6-sU5C:-NBTZNWTA1Qj3pT6rK-Q:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L48)):*

It seems like this is a `DateTime`, not a `Date`, yet it is called a date in the command.  It should probably actually be a `Date`, so that you don't have to unnecessarily specify a time when setting it.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT2rhO5fMQShzHI5mA:-NBTT3kWDz-YXA31rrQY:b-xz08ac) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L50)):*

I'm not seeing why this should be an abstract method given how it works differently across the two commands.
</blockquote></details>

Changed this to return the package promotion here in the abstract class, then just override it in CreatePackagePromotion to return null. 
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT32vgACL02Rv0MC2F:-NBTTYP74gw58CmC4oQm:bi04i26) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L54)):*

I don't understand why this is an abstract method (and maybe also the next one as well).  Seems like these should just be implemented normally?  It is possible that you're relying too much on inheritance in implementing the functionality here.  As Effective Java says, prefer composition over inheritance.  Inheritance is useful for sharing the common command-line parameters across the commands, but not much else.
</blockquote></details>

Changed this to getAndCheckAllocationToken(). Implemented it here with just the check that the token exists. But added an override in CreatePackagePromotionCommand to add the checks that it is a package token and the added error message to create the token first before creating the package, which would not make sense as an error message on an update. 
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 71 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOg7SF6vdl1AKI3MSg:-NBP2SVgBpOfU1LQrXPk:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L71)):*

all (?) of these might not be present, need to check for that before assigning them to the new object

Need tests for that as well -- this should be triggered fairly obviously by tests that test setting only some of the fields. 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java` line 62 at r1](https://reviewable.io/reviews/google/nomulus/1769#-NAzmR4ECj7PxQ8Tk9WJ:-NBTEXNk4B5rjuwbQXjc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/34064de1daa55532210c96df3ccd16caed32e19e/core/src/main/java/google/registry/tools/CreatePackagePromotionCommand.java#L62)):*

yeah, I see you have the clear_last_notification_sent flag below 

instead of doing this then, you can just (possibly) set the last-notification-sent flag to null in Update* if that's triggered. Then you don't need this other method. 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 48 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOjg1S-dtBmtHOHjpG:-NBTPPEP6O-J7QvJYsPo:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L48)):*

1. this is the same method in both classes, shouldn't be abstract 
2. can inline the call to .get
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOjq_q6TQYJ1DMsJej:-NBTW4j-6vNpAm-hv3tZ:b-p7yxsk) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L54)):*

why is this method different between the two implementations, why does it need to be abstract?
</blockquote></details>

changed these to normal implementations in the abstract class with overrides in the subclass

gbrodman (changes requested):
> Reviewed 2 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT2rhO5fMQShzHI5mA:-NBme1mN7mRIl4zGrFVP:b-hwatfv) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L50)):*

Changed this to return the package promotion here in the abstract class, then just override it in CreatePackagePromotion to return null. 
</blockquote></details>

I actually think that an abstract method is fine, similar to `CreateOrUpdateTldCommand::getOldRegistry` 

Then, the branching logic is only one line (whether we create a new PackagePromotion.Builder or use the existing one's `asBuilder()` 
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT32vgACL02Rv0MC2F:-NBmdiz32keBP4Mp_oyV:b-vqkgf6) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L54)):*

Changed this to getAndCheckAllocationToken(). Implemented it here with just the check that the token exists. But added an override in CreatePackagePromotionCommand to add the checks that it is a package token and the added error message to create the token first before creating the package, which would not make sense as an error message on an update. 
</blockquote></details>

I still don't think this really makes sense as an overrideable method, since the implementations are basically the same. There wouldn't be any harm in doing all the token checks in both places. 
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 81 at r3](https://reviewable.io/reviews/google/nomulus/1769#-NBmfYW4Bu8dRdfOQz0r:-NBmfYW4Bu8dRdfOQz0s:b2qjtlu) ([raw file](https://github.com/google/nomulus/blob/2479d8fb3e0940c614772a6d2682c9b9f6aa64c5/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L81)):*
> ```Java
> 
>   /** Does not clear the lastNotificationSent field. Subclasses can override this. */
>   boolean clearLastNotificationSent() {
> ```

this doesn't need to exist if you have an abstract `maybeGetExistingPackage()` method, then it can just be implemented only in UpdatePackage...
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOjq_q6TQYJ1DMsJej:-NBmg0Kb6YXGkBA-nUUy:b-zgxlrm) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L54)):*

changed these to normal implementations in the abstract class with overrides in the subclass
</blockquote></details>

this doesn't answer the root question though, which is "why does it need to be different?" -- as far as I can tell, it doesn't actually need to be different.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 1 of 7 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT2rhO5fMQShzHI5mA:-NBs1zbUDAVtCzhc8Gbf:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L50)):*

I actually think that an abstract method is fine, similar to `CreateOrUpdateTldCommand::getOldRegistry` 

Then, the branching logic is only one line (whether we create a new PackagePromotion.Builder or use the existing one's `asBuilder()` 


</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBT32vgACL02Rv0MC2F:-NBs5cKUC3JWjGhmq7A-:b-7tyai3) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L54)):*

I still don't think this really makes sense as an overrideable method, since the implementations are basically the same. There wouldn't be any harm in doing all the token checks in both places. 
</blockquote></details>

Ok, moved all the checks tot his class. The error message is a little strange now on updates, but I guess it's not really a big deal.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 81 at r3](https://reviewable.io/reviews/google/nomulus/1769#-NBmfYW4Bu8dRdfOQz0r:-NBs6-MJ1zNvwxHpEm4J:bmiy6ig) ([raw file](https://github.com/google/nomulus/blob/2479d8fb3e0940c614772a6d2682c9b9f6aa64c5/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L81)):*

this doesn't need to exist if you have an abstract `maybeGetExistingPackage()` method, then it can just be implemented only in UpdatePackage...
</blockquote></details>

unfortunately, that won't work because then the oldPackage variable for the stageEntityChange() will be incorrect which throws an error.
___
*[`core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java` line 54 at r2](https://reviewable.io/reviews/google/nomulus/1769#-NBOjq_q6TQYJ1DMsJej:-NBs6CVD4e6tUGMEyyyg:bgik34w) ([raw file](https://github.com/google/nomulus/blob/bfb752ab4ab67bcea9b1839fa1087b78819dad4f/core/src/main/java/google/registry/tools/UpdatePackagePromotionCommand.java#L54)):*

this doesn't answer the root question though, which is "why does it need to be different?" -- as far as I can tell, it doesn't actually need to be different. 
</blockquote></details>

removed the override

gbrodman (changes requested):
> Reviewed 3 of 6 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 81 at r3](https://reviewable.io/reviews/google/nomulus/1769#-NBmfYW4Bu8dRdfOQz0r:-NC0lZQ5AvEydjNBdFxI:bocr27h) ([raw file](https://github.com/google/nomulus/blob/2479d8fb3e0940c614772a6d2682c9b9f6aa64c5/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L81)):*

unfortunately, that won't work because then the oldPackage variable for the stageEntityChange() will be incorrect which throws an error.
</blockquote></details>

hmm right, that's unfortunate. 
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java` line 86 at r4](https://reviewable.io/reviews/google/nomulus/1769#-NC18qLQ8dYaFn3Fq4IU:-NC18qLQ8dYaFn3Fq4IV:b6p5gty) ([raw file](https://github.com/google/nomulus/blob/083418feb6839a78ecc21e0a2eb6be8180ddcc8a/core/src/main/java/google/registry/tools/CreateOrUpdatePackagePromotionCommand.java#L86)):*
> ```Java
>   }
> 
>   protected void initPackagePromotionCommand() {}
> ```

removable

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 100 at r7](https://reviewable.io/reviews/google/nomulus/1769#-NCNOYT72qfQFDJMn03_:-NCNOYT72qfQFDJMn03a:b-9k9jcc) ([raw file](https://github.com/google/nomulus/blob/9b37c13e5b2cd9e7cb41528f4b7e9f42ec69c69b/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L100)):*
> ```Java
>   /** Loads and returns a PackagePromotion entity by its token string directly from Cloud SQL. */
>   public static Optional<PackagePromotion> loadByTokenString(String tokenString) {
>     return tm().transact(
> ```

Starting a separate transaction in every little loader helper method is an antipattern.  Generally what you want is a single large transaction at a high level wrapping all the work you're doing (including both the load and any applicable saves).  Not doing this was the direct cause of our big ongoing issue with the re-save action having overwritten mutations.

So this method should instead merely have an assertion that we are already in a transaction, then a large wrapping transaction can be done somewhere in code at the callsite.
___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 103 at r7](https://reviewable.io/reviews/google/nomulus/1769#-NCNR2dqCwvnPhk2DDk-:-NCNR2dqCwvnPhk2DDk0:b-a04h1i) ([raw file](https://github.com/google/nomulus/blob/9b37c13e5b2cd9e7cb41528f4b7e9f42ec69c69b/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L103)):*
> ```Java
>             () ->
>                 jpaTm()
>                     .query("FROM PackagePromotion WHERE token = :token", PackagePromotion.class)
> ```

Why do you need to hand-write this query?  Can't it be handled through entity manager as `jpaTm().getEntityManager().find(...)` ?  Isn't this just a "load row from table by primary key" operation, i.e. the most basic load operation there is?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/model/domain/token/PackagePromotion.java` line 100 at r7](https://reviewable.io/reviews/google/nomulus/1769#-NCNOYT72qfQFDJMn03_:-NCR3m3b6Nyg1VPpYscb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9b37c13e5b2cd9e7cb41528f4b7e9f42ec69c69b/core/src/main/java/google/registry/model/domain/token/PackagePromotion.java#L100)):*

Starting a separate transaction in every little loader helper method is an antipattern.  Generally what you want is a single large transaction at a high level wrapping all the work you're doing (including both the load and any applicable saves).  Not doing this was the direct cause of our big ongoing issue with the re-save action having overwritten mutations.

So this method should instead merely have an assertion that we are already in a transaction, then a large wrapping transaction can be done somewhere in code at the callsite.
</blockquote></details>

Done.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1769)* status: 3 of 7 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1771, opened 2022-09-07
Remove ofy from Lock

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1771)

### DIFF
--- a/core/src/main/java/google/registry/model/EntityClasses.java
+++ b/core/src/main/java/google/registry/model/EntityClasses.java
@@ -28,7 +28,6 @@ import google.registry.model.index.EppResourceIndex;
 import google.registry.model.index.EppResourceIndexBucket;
 import google.registry.model.index.ForeignKeyIndex;
 import google.registry.model.reporting.HistoryEntry;
-import google.registry.model.server.Lock;
 import google.registry.model.server.ServerSecret;
 
 /** Sets of classes of the Objectify-registered entities in use throughout the model. */
@@ -52,7 +51,6 @@ public final class EntityClasses {
           HistoryEntry.class,
           Host.class,
           HostHistory.class,
-          Lock.class,
           ServerSecret.class);
 
   private EntityClasses() {}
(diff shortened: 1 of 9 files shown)


jianglai (commented):
> Reviewed 9 of 9 files at r1, 2 of 2 files at r2, 3 of 3 files at r3, 3 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1771)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/server/Lock.java` line 213 at r4](https://reviewable.io/reviews/google/nomulus/1771#-NBObaeW0vxiOrPlBF2f:-NBOdgEJEt9ENwCg8W83:blhehke) ([raw file](https://github.com/google/nomulus/blob/11dc04312c031d158a39b560b5353c1498d28657/core/src/main/java/google/registry/model/server/Lock.java#L213)):*

We shouldn't need to do this. This was so that we could use the SQL locks even in ofy, and that won't be necessary any more. We can just use jpaTm() everywhere
</blockquote></details>

You are right. There's no need to introduce a local variable anymore.
___
*[`core/src/main/java/google/registry/model/server/Lock.java` line 260 at r4](https://reviewable.io/reviews/google/nomulus/1771#-NBOcBoBE8lDYg7twDC6:-NBOdkAdAFMlYvElJrcq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/11dc04312c031d158a39b560b5353c1498d28657/core/src/main/java/google/registry/model/server/Lock.java#L260)):*

same comment here as above re: jpaTm
</blockquote></details>

Done.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1771)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1777, opened 2022-09-07
Remove ofy support from ForeignKeyIndex

jianglai:
FKI used to be persisted in datastore to help speed up loading by foreign key.
Now it is just a helper class to do the same thing in SQL because
indexing is natively supported in SQL.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1777)

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -61,12 +61,6 @@ by Joshua Bloch in his book Effective Java -->
     <property name="message" value="Use assertThrows and expectThrows from JUnitBackports instead of the deprecated methods on ExpectedException."/>
   </module>
 
-  <!-- Checks that the deprecated MockitoJUnitRunner is not used. -->
-  <module name="RegexpSingleline">
-    <property name="format" value="MockitoJUnitRunner"/>
-    <property name="message" value="MockitoJUnitRunner is deprecated. Use @RunWith(JUnit4.class) and MockitoRule instead."/>
-  </module>
-
   <module name="LineLength">
     <!-- Checks if a line is too long. -->
     <property name="max" value="${com.puppycrawl.tools.checkstyle.checks.sizes.LineLength.max}" default="100"/>
--- a/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
@@ -19,7 +19,6 @@ import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.batch.BatchModule.PARAM_DRY_RUN;
 import static google.registry.config.RegistryEnvironment.PRODUCTION;
-import static google.registry.model.ResourceTransferUtils.updateForeignKeyIndexDeletionTime;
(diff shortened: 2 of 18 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1777)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/model/index/ForeignKeyIndex.java` line 70 at r2](https://reviewable.io/reviews/google/nomulus/1777#-NBT4GJh8Er9BooO7cXm:-NBT4GJh8Er9BooO7cXn:b1i821i) ([raw file](https://github.com/google/nomulus/blob/49ade014ab4a9fad3109a270bb7472881e928f9b/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L70)):*
> ```Java
>  * becomes the active entity.
>  */
> @DeleteAfterMigration
> ```

I think it's fair to leave `@DeleteAfterMigration` in place, as there's still a lot more work to be done to entirely get rid of the concept of a `ForeignKeyIndex` before we can say we're totally done with this refactoring.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1777#-:-NBT4Y0tBW4XUbS1sq1a:bnfp4nl)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

jianglai:
> 
> 
> > *[Reviewable](https://reviewable.io/reviews/google/nomulus/1777)* status: 0 of 18 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)
> 
> ___
> *[`core/src/main/java/google/registry/model/index/ForeignKeyIndex.java` line 70 at r2](https://reviewable.io/reviews/google/nomulus/1777#-NBT4GJh8Er9BooO7cXm:-NBT4GJh8Er9BooO7cXn:b1i821i) ([raw file](https://github.com/google/nomulus/blob/49ade014ab4a9fad3109a270bb7472881e928f9b/core/src/main/java/google/registry/model/index/ForeignKeyIndex.java#L70)):*
> > ```Java
> >  * becomes the active entity.
> >  */
> > @DeleteAfterMigration
> > ```
> 
> I think it's fair to leave `@DeleteAfterMigration` in place, as there's still a lot more work to be done to entirely get rid of the concept of a `ForeignKeyIndex` before we can say we're totally done with this refactoring.
> 
> 
> 
> 

I think the class name can probably be changed, but most of its functionalities should survive, like caching and determining the Vkey based on the current time. It's more like a utility class than an index that's saved on the database now. 

Are you suggesting that we get rid of this class altogether? Its functionalities will have to live somewhere...

CydeWeys:
The functionality can continue to live in some kind of static final `Utils` class with a private constructor (and the per-EPP-resource descendant types deleted).  And the cache can be rewritten to not refer to this entity anymore, as it still contains lots of unnecessary junk.  Even though the `@Entity` annotation is gone as of this PR, there's still lots of Objectify-related cruft in it; if we were rewriting this from scratch for SQL only it would not look remotely like this.

jianglai:
Fair point. Created b/245804460 to track this.

PR #1783, opened 2022-09-13
Refactor ForeignKeyIndex into ForeignKeyUtils

jianglai:
The old class is modeled after datastore with some logic jammed in for it to work with SQL as well. As of #1777, the ofy related logic is deleted, however the general structure of the class remained datastore oriented.

This PR refactors the existing class into a `ForeignKeyUtils` helper class that does away wit the index subclasses and provides static helper methods to do the same, in a SQL-idiomatic fashion. 

Some minor changes are made to the EPP resource classes to make it possible to create them in a SQL only environment in tests.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1783)


jianglai (commented):
> Reviewed 39 of 49 files at r1, 10 of 10 files at r2, 1 of 3 files at r3, 29 of 30 files at r4, 76 of 76 files at r5, 1 of 2 files at r6, 1 of 1 files at r7, 13 of 13 files at r8, 2 of 2 files at r9, 2 of 2 files at r11, 2 of 2 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1783)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/ForeignKeyUtils.java` line 61 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6lQco-NoaPj6Eo_as:-NC72Eoh1XYF79suiU6R:bfyzaws) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/ForeignKeyUtils.java#L61)):*

probably out of the scope of this PR but we should change the Java field names of the host and domain names soon 
</blockquote></details>

Fun fact, I did change them, but that part of the change becomes too big so I had to branch them off to a different PR (#1784), still WIP.
___
*[`core/src/main/java/google/registry/model/ForeignKeyUtils.java` line 112 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6vtAwDLCHKe--xXxn:-NCL6XF21iG7XNeqhVhu:b-ll5apa) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/ForeignKeyUtils.java#L112)):*

Is there a way to do this query more simply with fewer selects? I don't know what HQL supports, but it's definitely possible in Postgres to do something like 

`SELECT DISTINCT ON (domain_name) domain_name, repo_id, deletion_time FROM "Domain" WHERE domain_name IN (:fks) ORDER BY domain_name, deletion_time DESC`

HQL might not support DISTINCT ON though...

if it were just one foreign key then we could just do SELECT * FROM "Domain" WHERE fk = fk ORDER BY deletionTime DESC LIMIT 1
</blockquote></details>

Hmm, I guess the question is do we want to error out here when more than one row have the same max `deletionTime`. The way it is currently written, we get a duplicate key exception when trying to insert the key into the immutable map. If we just use the first value ordered by desc, then there won't be an error.

This shouldn't be an issue in production as we simply cannot have two epp resources with the same foreign key have the same deletion time, one must be deleted before the next is created. In test, however, we are pretty liberal in recreating epp resources with the same foreign key, and they need to be fixed, as is done in this PR.

If those tests have not been fixed, I'd lean toward your suggestion, to save some time. But given that they are done, I think throwing an exception might be better because it alerts us of a clear invariance violation, what do you think?
___
*[`core/src/main/java/google/registry/model/contact/Contact.java` line 48 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6yZ-iAWMmmtBnAQBb:-NCLN7Ka2LAzNPRavzAY:b-gjnzq5) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/contact/Contact.java#L48)):*

q: how will this affect instances where we already store VKey<Contact> in the database, e.g. Domain's contacts?  (same question for the other places where we do this, I suppose) 
</blockquote></details>

From what I understand, all this does is to instruct the VKey converter to not create a symmetric VKey (the one that has both a sql and a ofy key), when loading the data from SQL:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/persistence/converter/VKeyConverter.java

I had to make this change because otherwise some of the tests are failing, probably because it was trying to recover VKeys directly instead of going through the FKI intermediary. Don't really the exact detail at this point. But regardless that should have no effect on how the data saved in SQL is used.

In fact, I think we can probably remove `WithStringVKey` and `WithLongVKey` now, as all they do is to allow you to specific if the key is composite, and if so, you'd have to write your own function to recover the full Ofy key with the ancestor chain, which is not needed now.
___
*[`core/src/main/java/google/registry/model/host/Host.java` line 54 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6zE321ePwoxYkbmZv:-NCLOmPb6vXU_cpxtuXj:byn65ru) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/host/Host.java#L54)):*

should this also be composite? 
</blockquote></details>

It probably should, but does not seem to impact test. I think adding it would not hurt, in case there are some edge cases not covered by tests.
___
*[`core/src/test/resources/google/registry/webdriver/goldens/chrome-linux/RegistrarConsoleScreenshotTest_registryLockVerify_success_page.png` line  at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC7-Ldw5RT5S6ify3PZ:-NCLOz3G6dmPDdczBaor:bpe40k1) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/test/resources/google/registry/webdriver/goldens/chrome-linux/RegistrarConsoleScreenshotTest_registryLockVerify_success_page.png#L)):*

wat
</blockquote></details>

This is one of the incidents where example.tld were persisted too many times, each time creating a new EPP resource. This caused the FKUtils to fail due to duplicate rows returned by the query. I updated the domain name to example-lock.tld, which caused the returned message to change as well.

gbrodman (approved):
> Reviewed 2 of 2 files at r14, 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1783)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/ForeignKeyUtils.java` line 61 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6lQco-NoaPj6Eo_as:-NCLp9uk0RJ9augTsgO9:bp7vbfx) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/ForeignKeyUtils.java#L61)):*

Fun fact, I did change them, but that part of the change becomes too big so I had to branch them off to a different PR (#1784), still WIP.
</blockquote></details>

sounds good!
___
*[`core/src/main/java/google/registry/model/ForeignKeyUtils.java` line 112 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6vtAwDLCHKe--xXxn:-NCLvnq76Mbcur1uBEdF:b-h91mmp) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/ForeignKeyUtils.java#L112)):*

Hmm, I guess the question is do we want to error out here when more than one row have the same max `deletionTime`. The way it is currently written, we get a duplicate key exception when trying to insert the key into the immutable map. If we just use the first value ordered by desc, then there won't be an error.

This shouldn't be an issue in production as we simply cannot have two epp resources with the same foreign key have the same deletion time, one must be deleted before the next is created. In test, however, we are pretty liberal in recreating epp resources with the same foreign key, and they need to be fixed, as is done in this PR.

If those tests have not been fixed, I'd lean toward your suggestion, to save some time. But given that they are done, I think throwing an exception might be better because it alerts us of a clear invariance violation, what do you think?
</blockquote></details>

Like you said I don't think it matters so much with regards to exceptions. My suggestion was mainly to simplify the query. Unfortunately, it doesn't seem like this simplified query is possible in HQL. 
___
*[`core/src/main/java/google/registry/model/contact/Contact.java` line 48 at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC6yZ-iAWMmmtBnAQBb:-NCLpGWI48rFVznxm-5Q:b-bzzmck) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/main/java/google/registry/model/contact/Contact.java#L48)):*

From what I understand, all this does is to instruct the VKey converter to not create a symmetric VKey (the one that has both a sql and a ofy key), when loading the data from SQL:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/persistence/converter/VKeyConverter.java

I had to make this change because otherwise some of the tests are failing, probably because it was trying to recover VKeys directly instead of going through the FKI intermediary. Don't really the exact detail at this point. But regardless that should have no effect on how the data saved in SQL is used.

In fact, I think we can probably remove `WithStringVKey` and `WithLongVKey` now, as all they do is to allow you to specific if the key is composite, and if so, you'd have to write your own function to recover the full Ofy key with the ancestor chain, which is not needed now.
</blockquote></details>

yeah, that makes sense cool. We can punt on removing those annotations for now. 
___
*[`core/src/test/resources/google/registry/webdriver/goldens/chrome-linux/RegistrarConsoleScreenshotTest_registryLockVerify_success_page.png` line  at r13](https://reviewable.io/reviews/google/nomulus/1783#-NC7-Ldw5RT5S6ify3PZ:-NCLytbl5yAYLprKpHc6:bbyy56v) ([raw file](https://github.com/google/nomulus/blob/00c8ab927fd249f26721a73e4afad28d42bde5fb/core/src/test/resources/google/registry/webdriver/goldens/chrome-linux/RegistrarConsoleScreenshotTest_registryLockVerify_success_page.png#L)):*

This is one of the incidents where example.tld were persisted too many times, each time creating a new EPP resource. This caused the FKUtils to fail due to duplicate rows returned by the query. I updated the domain name to example-lock.tld, which caused the returned message to change as well.
</blockquote></details>

aha, it wasn't obvious in the images

jianglai (commented):
> Reviewed 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1783)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1785, opened 2022-09-13
Do not enqueue DNS updates when flow doesn't affect nameservers

ptkach:
I believe this should safely fix the issue with triggering dns publish tasks for when they are not required. As far as I understand this process only **nameservers** and **ds data** change should trigger dns publish task.

I removed ```assertDnsTasksEnqueued``` from ```doSuccessfulTest``` as evidentaly it's not expected. Added the test for the this particular use case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1785)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
@@ -181,7 +181,10 @@ public final class DomainUpdateFlow implements TransactionalFlow {
     DomainHistory domainHistory =
         historyBuilder.setType(DOMAIN_UPDATE).setDomain(newDomain).build();
     validateNewState(newDomain);
-    dnsQueue.addDomainRefreshTask(targetId);
+    if (newDomain.getDsData() != existingDomain.getDsData()
+        || newDomain.getNsHosts() != existingDomain.getNsHosts()) {
+      dnsQueue.addDomainRefreshTask(targetId);
+    }
     ImmutableSet.Builder<ImmutableObject> entitiesToSave = new ImmutableSet.Builder<>();
     entitiesToSave.add(newDomain, domainHistory);
     Optional<BillingEvent.OneTime> statusUpdateBillingEvent =
@@ -268,12 +271,18 @@ public final class DomainUpdateFlow implements TransactionalFlow {
             .setLastEppUpdateRegistrarId(registrarId)
             .addStatusValues(add.getStatusValues())
             .removeStatusValues(remove.getStatusValues())
-            .addNameservers(add.getNameservers().stream().collect(toImmutableSet()))
-            .removeNameservers(remove.getNameservers().stream().collect(toImmutableSet()))
             .removeContacts(remove.getContacts())
             .addContacts(add.getContacts())
             .setRegistrant(firstNonNull(change.getRegistrant(), domain.getRegistrant()))
             .setAuthInfo(firstNonNull(change.getAuthInfo(), domain.getAuthInfo()));
+
+    if (!add.getNameservers().isEmpty()) {
+      domainBuilder.addNameservers(add.getNameservers().stream().collect(toImmutableSet()));
+    }
+    if (!remove.getNameservers().isEmpty()) {
+      domainBuilder.removeNameservers(remove.getNameservers().stream().collect(toImmutableSet()));
+    }
+
     Optional<DomainUpdateSuperuserExtension> superuserExt =
         eppInput.getSingleExtension(DomainUpdateSuperuserExtension.class);
     if (superuserExt.isPresent()) {
--- a/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java
+++ b/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java
@@ -47,6 +47,7 @@ import static google.registry.testing.DomainSubject.assertAboutDomains;
 import static google.registry.testing.EppExceptionSubject.assertAboutEppExceptions;
 import static google.registry.testing.HistoryEntrySubject.assertAboutHistoryEntries;
 import static google.registry.testing.TaskQueueHelper.assertDnsTasksEnqueued;
+import static google.registry.testing.TaskQueueHelper.assertNoDnsTasksEnqueued;
 import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static org.joda.money.CurrencyUnit.USD;
 import static org.junit.jupiter.api.Assertions.assertThrows;
@@ -1726,5 +1727,11 @@ class DomainUpdateFlowTest extends ResourceFlowTestCase<DomainUpdateFlow, Domain
     clock.advanceOneMilli();
     runFlowAsSuperuser();
     assertAboutDomains().that(reloadResourceByForeignKey()).hasNoAutorenewEndTime();
+  }
+
+  @Test
+  void testDnsTaskIsNotTriggeredWhenNoDSChangeSubmitted() throws Exception {
+    setEppInput("domain_update_no_ds_change.xml");
+    assertNoDnsTasksEnqueued();
   }
 }
--- a/core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml
+++ b/core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml
@@ -1,0 +1,18 @@
+<epp xmlns="urn:ietf:params:xml:ns:epp-1.0">
+  <command>
+    <update>
+      <domain:update
+       xmlns:domain="urn:ietf:params:xml:ns:domain-1.0">
+        <domain:name>example.tld</domain:name>
+        <domain:add>
+          <domain:contact type="admin">mak21</domain:contact>
+          <domain:contact type="billing">mak21</domain:contact>
+          <domain:contact type="tech">mak21</domain:contact>
+          <domain:status s="serverHold"
+              lang="en">Server hold.</domain:status>
+        </domain:add>
+      </domain:update>
+    </update>
+    <clTRID>ABC-12345</clTRID>
+  </command>
+</epp>


core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
```diff
@@ -268,12 +271,18 @@ private Domain performUpdate(Update command, Domain domain, DateTime now) throws
...
             .addStatusValues(add.getStatusValues())
             .removeStatusValues(remove.getStatusValues())
-            .addNameservers(add.getNameservers().stream().collect(toImmutableSet()))
-            .removeNameservers(remove.getNameservers().stream().collect(toImmutableSet()))
             .removeContacts(remove.getContacts())
             .addContacts(add.getContacts())
             .setRegistrant(firstNonNull(change.getRegistrant(), domain.getRegistrant()))
             .setAuthInfo(firstNonNull(change.getAuthInfo(), domain.getAuthInfo()));
+
+    if (add.getNameservers().size() != 0) {
```
CydeWeys, line 279: It doesn't seem like these changes are necessary?

core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java
```diff
@@ -218,7 +218,6 @@ private void doSuccessfulTest(String expectedXmlFilename) throws Exception {
         .and()
         .hasNoAutorenewEndTime();
     assertNoBillingEvents();
-    assertDnsTasksEnqueued("example.tld");
```
CydeWeys, line 221: This assert needs to be present on at least some tests, no?

core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
CydeWeys: Or are these necessary for the conditional statement above to work?

ptkach (commented):
Fixed.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1785#-:-NBxAi7i94xwloWWaVY_:bt2vr1b)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 279 at r1](https://reviewable.io/reviews/google/nomulus/1785#-NBs6CTCDcUed7c-bcu5-r1-279:-NBxAktV4Gue23i5Q4MM:b-3pn9zn) ([raw file](https://github.com/google/nomulus/blob/3939a2f57dc13ab0efecea8af88f6a0ec0ad6026/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L279)):*

Or are these necessary for the conditional statement above to work?
</blockquote></details>

Exactly, it's necessary for reference check above. If addNameservers is called with any value - reference gets updated and we no longer can easily tell if dnsPublish task is required.

ptkach:
Also tested it in alpha and confirmed that without ns data in epp command dns publish task no longer gets added to the queue.

jianglai (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1785)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @ptkach)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 279 at r1](https://reviewable.io/reviews/google/nomulus/1785#-NBs6CTCDcUed7c-bcu5-r1-279:-NC0v0neE7W8kWEnNfuD:bqltta3) ([raw file](https://github.com/google/nomulus/blob/3939a2f57dc13ab0efecea8af88f6a0ec0ad6026/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L279)):*

Exactly, it's necessary for reference check above. If addNameservers is called with any value - reference gets updated and we no longer can easily tell if dnsPublish task is required. 
</blockquote></details>

You can use `.isEmpty()` to check for emptiness. Also, I'm still not understanding this entirely, if there's nothing to add or remove, why would calling `addNameservers` and `removeNameservers`, as it is currently done, produce a new domain whose dsData is different than the old one, therefore triggering the DNS update? One would expect that adding/removing an empty set is a no-op.
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1734 at r2](https://reviewable.io/reviews/google/nomulus/1785#-NC0vqFy5Z6T2zHDRa_j:-NC0vqFz8K6kai7wPYfi:b-2voyc8) ([raw file](https://github.com/google/nomulus/blob/54d30cfc3a5379966d004da5b606898923040b53/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1734)):*
> ```Java
>   void testDnsTaskIsNotTriggeredWhenNoDSChangeSubmitted() throws Exception {
>     setEppInput("domain_update_no_ds_change.xml");
>     assertDnsTasksEnqueued();
> ```

You can use `assertNoDnsTasksEnqueued()`.



_Code quote:_
```Java
assertDnsTasksEnqueued();
```
___
*[`core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml` line 1 at r2](https://reviewable.io/reviews/google/nomulus/1785#-NC0vwhX42zPVL31UXVD:-NC0vwhX42zPVL31UXVE:b-wkgusl) ([raw file](https://github.com/google/nomulus/blob/54d30cfc3a5379966d004da5b606898923040b53/core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml#L1)):*
> ```XML
> <epp xmlns="urn:ietf:params:xml:ns:epp-1.0">
> ```

Are there no existing xml files that update a domain without updating it's NS or DS records?



_Code quote:_
```XML
<epp xmlns="urn:ietf:params:xml:ns:epp-1.0">
  <command>
    <update>
      <domain:update
       xmlns:domain="urn:ietf:params:xml:ns:domain-1.0">
        <domain:name>example.tld</domain:name>
        <domain:add>
          <domain:contact type="admin">mak21</domain:contact>
          <domain:contact type="billing">mak21</domain:contact>
          <domain:contact type="tech">mak21</domain:contact>
          <domain:status s="serverHold"
              lang="en">Server hold.</domain:status>
        </domain:add>
      </domain:update>
    </update>
    <clTRID>ABC-12345</clTRID>
  </command>
</epp>
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1785)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 279 at r1](https://reviewable.io/reviews/google/nomulus/1785#-NBs6CTCDcUed7c-bcu5-r1-279:-NC5NBN61vXgc9zxv8wK:bpuedbn) ([raw file](https://github.com/google/nomulus/blob/3939a2f57dc13ab0efecea8af88f6a0ec0ad6026/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L279)):*

You can use `.isEmpty()` to check for emptiness. Also, I'm still not understanding this entirely, if there's nothing to add or remove, why would calling `addNameservers` and `removeNameservers`, as it is currently done, produce a new domain whose dsData is different than the old one, therefore triggering the DNS update? One would expect that adding/removing an empty set is a no-op.
</blockquote></details>

Updated isEmty().
There's no checks for whether or not domain DS data has actually changed neither in ReadDnsQueueAction nor in PublishDnsUpdatesAction. This is a short circuit to not even waste time on those jobs if it's not required.  
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1734 at r2](https://reviewable.io/reviews/google/nomulus/1785#-NC0vqFy5Z6T2zHDRa_j:-NC5PwTJ0_M39ha2DfL_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/54d30cfc3a5379966d004da5b606898923040b53/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1734)):*

You can use `assertNoDnsTasksEnqueued()`.
</blockquote></details>

Done.
___
*[`core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml` line 1 at r2](https://reviewable.io/reviews/google/nomulus/1785#-NC0vwhX42zPVL31UXVD:-NC5PLL28znC9czPv4_1:bx3unm3) ([raw file](https://github.com/google/nomulus/blob/54d30cfc3a5379966d004da5b606898923040b53/core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml#L1)):*

Are there no existing xml files that update a domain without updating it's NS or DS records?
</blockquote></details>

There's, but they serve different tests with different purposes (like domain_update_superuser_extension). Me reusing them can cause unexpected side effects down the road if they are update for the purpose of the original test, unintentionally breaking my test. Does it make sense?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1785)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 279 at r1](https://reviewable.io/reviews/google/nomulus/1785#-NBs6CTCDcUed7c-bcu5-r1-279:-NC6I_7BAHXqqOlOxLA7:b8t6mrx) ([raw file](https://github.com/google/nomulus/blob/3939a2f57dc13ab0efecea8af88f6a0ec0ad6026/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L279)):*

Those two actions cannot check for the change because they do not know what the DS/NS data was before. That information is lost by the time they read the tasks.

I guess I didn't make my point clear. I'm not asking why we need the checks here to save the DNS publish jobs from having to work on a no-op job. Rather, I'm saying it doesn't seem necessary to have this check, in conjunction with the `newDomain.getDsData() != existingDomain.getDsData()` check you added above. i.e. if you don't do the checks here, we should will NOT enqueue unnecessary DNS update tasks as long as the check before enqueuing that you added are there, right?

So it's fine to check for the emptiness here, but what you said about it being necessary for the reference check above seems to be incorrect.

Not asking you to remove the checks, just wanting to make sure that I understand the logic here.
</blockquote></details>

Oh that, the reason I added originally size() check was me trying to be extra safe. Let me explain - previously all changes to domain caused DNS publish task, which was not correct, so the point was to make it so only certain (necessary) changes will set off DNS publish. It's clear that DsData should trigger it, but what about nameservers chage? As if for example nameservers change is used in DNS publish task and I were to miss adding it to the check we'd fail to schedule DNS publish tasks when it's required.
It's not absolutely straightforward from codebase and I also asked around and didn't get a definite answer, for this question, so based on a "better be safe than sorry" rule I've decided to add this check. It's not going to get anything worse than it's been, while still serving the purpose of the task. 
If you can confirm that nameservers change should not trigger the task I will gladly remove it. 
Thanks!
___
*[`core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml` line 1 at r2](https://reviewable.io/reviews/google/nomulus/1785#-NC0vwhX42zPVL31UXVD:-NC6MGNBBYO0PDxqfM40:b-dvoa1t) ([raw file](https://github.com/google/nomulus/blob/54d30cfc3a5379966d004da5b606898923040b53/core/src/test/resources/google/registry/flows/domain/domain_update_no_ds_change.xml#L1)):*

Personally, I'd make the test file more generic and reuse them if possible, and fork if necessary. But that's just my personal preference.
</blockquote></details>

I've used generic approach elsewhere, but here it seems to be more appropriate and inline with rest of the tests, when it's using specific files to cover the test.

jianglai (approved):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1785)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 279 at r1](https://reviewable.io/reviews/google/nomulus/1785#-NBs6CTCDcUed7c-bcu5-r1-279:-NC6zIjZ4pxiMp573Fu5:b9c15ky) ([raw file](https://github.com/google/nomulus/blob/3939a2f57dc13ab0efecea8af88f6a0ec0ad6026/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L279)):*
> Action
Hmm, are we sure we are talking about the same thing here...? All I'm saying here is that, if you don't check for emptiness here and just add/remove the nameservers, the end result should still be the same thing, right? You added an empty set to and then removed an empty set from the original set, therefore it remain unchanged.

"As if for example nameservers change is used in DNS publish task and I were to miss adding it to the check we'd fail to schedule DNS publish tasks when it's required." I'm not sure I completely understand what you meant here. What "nameserver flow change used in DNS publish task" do you mean? Isn't nameserver change the add/remove in discussion here? You cannot "change" a nameserver in a domain update command, you can only add new ones and remove old ones. See [here](https://www.rfc-editor.org/rfc/rfc5731.html#page-25) for details.

In any case, like I said in the previous comment, this PR itself looks correct. I'll approve it so it is ready for next week's deployment. We can keep discussing in this thread.

weiminyu:
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 186 at r3](https://reviewable.io/reviews/google/nomulus/1785#-NDZpTmi57nDVfe_kJ2a:-NDZpTmi57nDVfe_kJ2b:bjvzmoc) ([raw file](https://github.com/google/nomulus/blob/07a0a3d96224ef00439413990c0d09b5a18d1370/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L186)):*
> ```Java
>     if (newDomain.getDsData() != existingDomain.getDsData()
>         || newDomain.getNsHosts() != existingDomain.getNsHosts()) {
>       dnsQueue.addDomainRefreshTask(targetId);
> ```

This is not correct. Status change (e.g., a client hold, should trigger a refresh too).

This change is the cause of b/250731430



_Code quote:_
```Java
    if (newDomain.getDsData() != existingDomain.getDsData()
        || newDomain.getNsHosts() != existingDomain.getNsHosts()) {
      dnsQueue.addDomainRefreshTask(targetId);
    }
```

PR #1790, opened 2022-09-15
Create a scrap command to cancel OneTime billing events by ID

gbrodman:
This allows us to correct situations where we have erroneously charged registrars for an action, without explicitly issuing a refund.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1790)

### DIFF
--- a/core/src/main/java/google/registry/tools/RegistryTool.java
+++ b/core/src/main/java/google/registry/tools/RegistryTool.java
@@ -16,6 +16,7 @@ package google.registry.tools;
 
 import com.google.common.collect.ImmutableMap;
 import google.registry.tools.javascrap.CompareEscrowDepositsCommand;
+import google.registry.tools.javascrap.CreateCancellationsForOneTimesCommand;
 
 /** Container class to create and run remote commands against a Datastore instance. */
 public final class RegistryTool {
@@ -36,6 +37,7 @@ public final class RegistryTool {
           .put("convert_idn", ConvertIdnCommand.class)
           .put("count_domains", CountDomainsCommand.class)
           .put("create_anchor_tenant", CreateAnchorTenantCommand.class)
+          .put("create_cancellations_for_one_times", CreateCancellationsForOneTimesCommand.class)
           .put("create_cdns_tld", CreateCdnsTld.class)
           .put("create_contact", CreateContactCommand.class)
           .put("create_domain", CreateDomainCommand.class)
--- a/core/src/main/java/google/registry/tools/RegistryToolComponent.java
+++ b/core/src/main/java/google/registry/tools/RegistryToolComponent.java
@@ -42,6 +42,7 @@ import google.registry.request.Modules.UrlFetchServiceModule;
 import google.registry.request.Modules.UserServiceModule;
 import google.registry.tools.AuthModule.LocalCredentialModule;
 import google.registry.tools.javascrap.CompareEscrowDepositsCommand;
+import google.registry.tools.javascrap.CreateCancellationsForOneTimesCommand;
 import google.registry.util.UtilsModule;
 import google.registry.whois.NonCachingWhoisModule;
 import javax.annotation.Nullable;
@@ -94,6 +95,8 @@ interface RegistryToolComponent {
   void inject(CountDomainsCommand command);
 
   void inject(CreateAnchorTenantCommand command);
+
+  void inject(CreateCancellationsForOneTimesCommand command);
 
   void inject(CreateCdnsTld command);
 
--- a/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java
+++ b/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java
@@ -1,0 +1,126 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools.javascrap;
+
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+
+import com.beust.jcommander.Parameter;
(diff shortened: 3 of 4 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1790)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java` line 60 at r1](https://reviewable.io/reviews/google/nomulus/1790#-NC5ezF62VTS1s3iAdeH:-NC5ezF62VTS1s3iAdeI:bbt0879) ([raw file](https://github.com/google/nomulus/blob/b2369cbeb9763ba442a07bd8b963563ce7d48055/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java#L60)):*
> ```Java
>                 VKey<OneTime> key = VKey.createSql(OneTime.class, billingEventId);
>                 if (tm().exists(key)) {
>                   OneTime oneTime = tm().loadByKey(key);
> ```

Might be more efficient to use `loadByKeysIfPresent` and do the processing outside the transaction? Since you are creating a cancellation in a different transaction in the `execute`, it's not like transactionality is preserved anyway. I don't think cancellations can be created in other ways so we should be safe to do it in two transactions.



_Code quote:_
```Java
OneTime oneTime = tm().loadByKey(key);
```
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java` line 96 at r1](https://reviewable.io/reviews/google/nomulus/1790#-NC5ho5V-LgVCASLM6Zp:-NC5ho5V-LgVCASLM6Zq:b-wxrig5) ([raw file](https://github.com/google/nomulus/blob/b2369cbeb9763ba442a07bd8b963563ce7d48055/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java#L96)):*
> ```Java
>           tm().transact(
>                   () -> {
>                     if (alreadyCancelled(oneTime)) {
> ```

It seems a bit redundant to check for this twice, here and above when you create the `oneTimesToCancel`, given that the odds of some `OneTime` becomes cancelled between now and then is practically nonexistent.

A better approach might be to just load all existing `OneTime`s from the list and filter out the ones that are referenced in `Cancellation` in one query.

But I do recognize that we may be over engineering for a scrap tool that only have to deal with a couple thousand entities at most...



_Code quote:_
```Java
alreadyCancelled
```
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java` line 102 at r1](https://reviewable.io/reviews/google/nomulus/1790#-NC5jexP9ZjKI7-0JpmO:-NC5jexP9ZjKI7-0JpmP:b-r9q4fn) ([raw file](https://github.com/google/nomulus/blob/b2369cbeb9763ba442a07bd8b963563ce7d48055/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java#L102)):*
> ```Java
>                     }
>                     tm().put(
>                             new Cancellation.Builder()
> ```

Technically these values should be for the "thing" that created the cancellation, not just values copied over from the onetime. I understand for a lot of them we simply cannot find a better proxy than the value in onetime because this is a manual operation. But maybe there should be a better `Reason` at least? How about the deprecated `ERROR` enum value, which according to b/31676071 is created for similar reasons. In fact, the bug mentions a scrap tool to manually create cancellations to correct errors...

Also, what happened to the history entries of the change that were reverted during the resave?



_Code quote:_
```Java
                            new Cancellation.Builder()
                                .setOneTimeEventKey(oneTime.createVKey())
                                .setBillingTime(oneTime.getBillingTime())
                                .setDomainHistoryId(oneTime.getDomainHistoryId())
                                .setRegistrarId(oneTime.getRegistrarId())
                                .setEventTime(oneTime.getEventTime())
                                .setFlags(oneTime.getFlags())
                                .setReason(oneTime.getReason())
                                .setTargetId(oneTime.getTargetId())
                                .build());
```

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1790)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java` line 102 at r1](https://reviewable.io/reviews/google/nomulus/1790#-NC5jexP9ZjKI7-0JpmO:-NC6n91j3wZs6Bigo_Km:b953ed0) ([raw file](https://github.com/google/nomulus/blob/b2369cbeb9763ba442a07bd8b963563ce7d48055/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java#L102)):*

The history entries are still there but those are harmless, no issues with those. 

I noticed for existing cancellation BillingEvents they usually do tend to map to the BillingEvent that they're cancelling.  My thinking is that effectively we should treat these cancellations as "occurring" at the same time as the OneTime event since they should never have happened.  

Good to know about ERROR though, changed. 
</blockquote></details>

Yeah, if you read b/31676071 and it's associated CL, Nick mentioned that the entries in cancellations map to onetime because, like you said earlier, it's caused by cancellation during grace period, so it was an intentional decision to map them.

Any reason to not copy the flag? I guess it shouldn't really matter.

sarahcaseybot (approved):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1790)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1790)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java` line 102 at r1](https://reviewable.io/reviews/google/nomulus/1790#-NC5jexP9ZjKI7-0JpmO:-NC6uQCYEo2lZaqYYKTW:b-yqttcw) ([raw file](https://github.com/google/nomulus/blob/b2369cbeb9763ba442a07bd8b963563ce7d48055/core/src/main/java/google/registry/tools/javascrap/CreateCancellationsForOneTimesCommand.java#L102)):*

Yeah, if you read b/31676071 and it's associated CL, Nick mentioned that the entries in cancellations map to onetime because, like you said earlier, it's caused by cancellation during grace period, so it was an intentional decision to map them.

Any reason to not copy the flag? I guess it shouldn't really matter.
</blockquote></details>

yeah my thinking was that the flags are unnecessary for cancellations and none of them really apply here

PR #1791, opened 2022-09-15
Fail generate invoices job when billing events not finished expanding

ptkach:
In case expand billing events job overlaps with generate invoices job, there's a chance that some auto-renewals might not get properly billed. This PR ensures ```/_dr/task/generateInvoices``` fails if last successful ```/_dr/task/expandRecurringBillingEvents``` is older than 1 day.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1791)

### DIFF
--- a/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml
+++ b/core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml
@@ -267,7 +267,7 @@
     about 2 hours to complete, so we give 11 hours to be safe. Normally, we give 24+ hours (see
     icannReportingStaging), but the invoicing team prefers receiving the e-mail on the first of
     each month. -->
-    <schedule>1 of month 17:00</schedule>
+    <schedule>1 of month 19:00</schedule>
     <target>backend</target>
   </cron>
 
--- a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
+++ b/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
@@ -15,7 +15,10 @@
 package google.registry.reporting.billing;
 
 import static google.registry.beam.BeamUtils.createJobName;
+import static google.registry.model.common.Cursor.CursorType.RECURRING_BILLING;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.Action.Method.POST;
+import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_OK;
 
@@ -29,6 +32,7 @@ import com.google.common.flogger.FluentLogger;
 import com.google.common.net.MediaType;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.config.RegistryEnvironment;
+import google.registry.model.common.Cursor;
 import google.registry.persistence.PersistenceModule;
 import google.registry.reporting.ReportingModule;
 import google.registry.request.Action;
@@ -40,6 +44,7 @@ import google.registry.util.Clock;
 import google.registry.util.CloudTasksUtils;
 import java.io.IOException;
 import javax.inject.Inject;
+import org.joda.time.DateTime;
 import org.joda.time.Duration;
 import org.joda.time.YearMonth;
 
@@ -108,6 +113,19 @@ public class GenerateInvoicesAction implements Runnable {
     response.setContentType(MediaType.PLAIN_TEXT_UTF_8);
     logger.atInfo().log("Launching invoicing pipeline for %s.", yearMonth);
     try {
+      DateTime currentCursorTime =
+          tm().transact(
+                  () ->
+                      tm().loadByKeyIfPresent(Cursor.createGlobalVKey(RECURRING_BILLING))
+                          .orElse(Cursor.createGlobal(RECURRING_BILLING, START_OF_TIME))
+                          .getCursorTime());
+
+      if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()) {
+        throw new IllegalStateException(
+            "Latest billing events expansion cycle hasn't finished yet, terminating invoicing"
+                + " pipeline");
+      }
+
       LaunchFlexTemplateParameter parameter =
           new LaunchFlexTemplateParameter()
               .setJobName(createJobName("invoicing", clock))
@@ -150,7 +168,7 @@ public class GenerateInvoicesAction implements Runnable {
       }
       response.setStatus(SC_OK);
       response.setPayload(String.format("Launched invoicing pipeline: %s", jobId));
-    } catch (IOException e) {
+    } catch (IOException | IllegalStateException e) {
       logger.atWarning().withCause(e).log("Template Launch failed.");
       emailUtils.sendAlertEmail(String.format("Pipeline Launch failed due to %s", e.getMessage()));
       response.setStatus(SC_INTERNAL_SERVER_ERROR);
--- a/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java
+++ b/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java
@@ -15,6 +15,8 @@
 package google.registry.reporting.billing;
 
 import static com.google.common.truth.Truth.assertThat;
+import static google.registry.model.common.Cursor.CursorType.RECURRING_BILLING;
+import static google.registry.testing.DatabaseHelper.persistResource;
 import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_OK;
 import static org.mockito.Mockito.mock;
@@ -24,6 +26,7 @@ import static org.mockito.Mockito.when;
 import com.google.cloud.tasks.v2.HttpMethod;
 import com.google.common.net.MediaType;
 import google.registry.beam.BeamActionTestBase;
+import google.registry.model.common.Cursor;
 import google.registry.reporting.ReportingModule;
 import google.registry.testing.AppEngineExtension;
 import google.registry.testing.CloudTasksHelper;
@@ -31,8 +34,10 @@ import google.registry.testing.CloudTasksHelper.TaskMatcher;
 import google.registry.testing.FakeClock;
 import google.registry.util.CloudTasksUtils;
 import java.io.IOException;
+import org.joda.time.DateTime;
 import org.joda.time.Duration;
 import org.joda.time.YearMonth;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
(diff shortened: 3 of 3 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1791)* status: all files reviewed, 4 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 74 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCIt3vS0clbHUPT7YCW:-NCIt3vS0clbHUPT7YCX:b4gpaa5) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L74)):*
> ```Java
> public class InvoicingPipeline implements Serializable {
> 
>   @Inject Clock clock;
> ```

This will not work. You need Dagger to actually inject the field. The class right now is not managed or created by dagger.



_Code quote:_
```Java
@Inject Clock clock;
```
___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCIty3O-kKlJQfGXZWD:-NCIty3PCLHPShD8lNI5:bn1qe2v) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L97)):*
> ```Java
>                         .getCursorTime());
> 
>     if (currentCursorTime.isBefore(clock.nowUtc().minusDays(1))) {
> ```

`RECURRING_BILLING` is the inclusive upper limit the time for which the billing event has been expanded to:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/common/Cursor.java;l=87?q=recurring_bill&sq=&ss=nomulus%2Fnomulus

What you want to make sure is that when the billing pipeline runs, all the recurring billing events that need to be expanded for the time period you care about are indeed expanded. That means you want to make sure this cursor is after the end of the month that the invoice is generated for. The current time that the pipeline runs doesn't really matter. The `yearMonth` value in `options` is what matters, for it designates the time period the invoice is for.

Also, if all you want to do is to not run the pipeline given this condition, you can do this check in the main function below, or even better, in the action that actually kicks off the pipeline: [`GenerateInvoicesAction`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java;l=107?q=generateinvoic&ss=nomulus%2Fnomulus).



_Code quote:_
```Java
(currentCursorTime.isBefore(clock.nowUtc().minusDays(1)))
```
___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCIzQE2AlGyfvu3vmNo:-NCIzQE2AlGyfvu3vmNp:b-djvjig) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L98)):*
> ```Java
> 
>     if (currentCursorTime.isBefore(clock.nowUtc().minusDays(1))) {
>       throw new RuntimeException(
> ```

This is not the most elegant solution as throwing an exception will fail the pipeline outright, and we will have to fix it (most likely by waiting a bit and rerunning it later) manually.

Granted, I don't think this should happen very often, as we kick off the pipeline at [17:00 on the first day of a month](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml;l=270;drc=775f672f2a1235832996ee8c2c98da7edd99a043;bpv=0;bpt=0), and the recurring expansion job runs at [03:00 each day](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml;l=139;drc=775f672f2a1235832996ee8c2c98da7edd99a043;bpv=0;bpt=0). So we have 14 hours for the expansion (of just one day's worth of new billing events) to finish.

But still, a better way to handle it is to use Cloud Tasks so we can have retries with exponential back off. In that case, you'd have a cron job that runs an action, which enqueues a task for `GenerateInvoicesAction`. When that task runs, it compares `yearMonth` and the cursor to determine if it should launch this pipeline. If not, it will return a [non-2XX](https://cloud.google.com/appengine/docs/legacy/standard/java/taskqueue/push/retrying-tasks) response, indicating to Cloud Tasks that it should be retried latter.



_Code quote:_
```Java
      throw new RuntimeException(
          "Latest billing events expansion cycle hasn't finished yet, terminating invoicing"
              + " pipeline");
```
___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 245 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCItM6Y12LLpjv7ew4j:-NCItM6Y12LLpjv7ew4k:b-k1jaaz) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L245)):*
> ```Java
>     InvoicingPipelineOptions options =
>         PipelineOptionsFactory.fromArgs(args).withValidation().as(InvoicingPipelineOptions.class);
>     new InvoicingPipeline(options).run();
> ```

If you do want to provide a `Clock`, use `SystemClock` as a constructor argument here, no need to use Dagger. But as I mentioned in the comment above, there's no need to use a `Clock` at all. What you should check is the `yearMonth` against the cursor, here, or more preferably in `GenerateInvoicesAction`.



_Code quote:_
```Java
InvoicingPipeline(options).run();
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1791)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 74 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCIt3vS0clbHUPT7YCW:-NCMUfSU1QZ_yJDNVFa5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L74)):*

This will not work. You need Dagger to actually inject the field. The class right now is not managed or created by dagger.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCIty3O-kKlJQfGXZWD:-NCMUhA92WwfQ9nQ8E6T:b-q8d5ex) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L97)):*

`RECURRING_BILLING` is the inclusive upper limit the time for which the billing event has been expanded to:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/common/Cursor.java;l=87?q=recurring_bill&sq=&ss=nomulus%2Fnomulus

What you want to make sure is that when the billing pipeline runs, all the recurring billing events that need to be expanded for the time period you care about are indeed expanded. That means you want to make sure this cursor is after the end of the month that the invoice is generated for. The current time that the pipeline runs doesn't really matter. The `yearMonth` value in `options` is what matters, for it designates the time period the invoice is for.

Also, if all you want to do is to not run the pipeline given this condition, you can do this check in the main function below, or even better, in the action that actually kicks off the pipeline: [`GenerateInvoicesAction`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java;l=107?q=generateinvoic&ss=nomulus%2Fnomulus).
</blockquote></details>

Thanks, that was very helpful. Updated accordingly! 
___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCIzQE2AlGyfvu3vmNo:-NCMUnjN5RGBjAmNAFwV:b-qeg7qb) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L98)):*

This is not the most elegant solution as throwing an exception will fail the pipeline outright, and we will have to fix it (most likely by waiting a bit and rerunning it later) manually.

Granted, I don't think this should happen very often, as we kick off the pipeline at [17:00 on the first day of a month](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml;l=270;drc=775f672f2a1235832996ee8c2c98da7edd99a043;bpv=0;bpt=0), and the recurring expansion job runs at [03:00 each day](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/env/production/default/WEB-INF/cron.xml;l=139;drc=775f672f2a1235832996ee8c2c98da7edd99a043;bpv=0;bpt=0). So we have 14 hours for the expansion (of just one day's worth of new billing events) to finish.

But still, a better way to handle it is to use Cloud Tasks so we can have retries with exponential back off. In that case, you'd have a cron job that runs an action, which enqueues a task for `GenerateInvoicesAction`. When that task runs, it compares `yearMonth` and the cursor to determine if it should launch this pipeline. If not, it will return a [non-2XX](https://cloud.google.com/appengine/docs/legacy/standard/java/taskqueue/push/retrying-tasks) response, indicating to Cloud Tasks that it should be retried latter.
</blockquote></details>

Updated. It looks like it's run under retryable-cron-tasks and should repeat 3 times if response if non-2XX, which should now be the case with my change. 
___
*[`core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java` line 245 at r1](https://reviewable.io/reviews/google/nomulus/1791#-NCItM6Y12LLpjv7ew4j:-NCMVOdB3oNUhBpW56U2:br6k48n) ([raw file](https://github.com/google/nomulus/blob/7209fc90b46d2db392310e1bc11f26d2ec341b2a/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L245)):*

If you do want to provide a `Clock`, use `SystemClock` as a constructor argument here, no need to use Dagger. But as I mentioned in the comment above, there's no need to use a `Clock` at all. What you should check is the `yearMonth` against the cursor, here, or more preferably in `GenerateInvoicesAction`.
</blockquote></details>

I removed, thanks for your suggestions!

jianglai:
I thought we are not going to rely on the cron queue retry logic? We should return a 2XX message to stop the retry and send an email eng when the cursor is not up to date.

ptkach:
> I thought we are not going to rely on the cron queue retry logic? We should return a 2XX message to stop the retry and send an email eng when the cursor is not up to date.

We agreed upon email and shifting cron job 17.00 -> 19.00, but I don't think we settled on 2XX. 
I believe non-2XX response seems more natural here as it's not a successful scenario, plus with retries we still increase (by tiny little) our chances to success. The only downside is 4 emails that will be sent out instead of one, but given that it's a rare event and we actually want to draw attention maybe it's not bad. Anyway both options are good and should suffice, so if you strongly feel that 2XX is required I can update. Let me know.

PR #1795, opened 2022-09-22
Create pipeline to save synthetic DomainHistory objects

gbrodman:
This runs over all domains that weren't deleted as of September 5. This will fix most of b/248112997, which is itself caused by b/245940594 -- creating synthetic history objects means that the RDE pipeline will look at those instead of the potentially-no-longer-valid data in the old history objects.


jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1795)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesPipeline.java` line 75 at r1](https://reviewable.io/reviews/google/nomulus/1795#-NC_w-tz6-88ZGdrFMUm:-NC_w-u-8-XQ-fywVePI:bem86wn) ([raw file](https://github.com/google/nomulus/blob/1fef5e789bf3ba0584b0e0f149dbe7cfd7208b3f/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesPipeline.java#L75)):*
> ```Java
>         .apply(
>             "Save a synthetic DomainHistory for each domain",
>             MapElements.into(TypeDescriptor.of(Void.class))
> ```

You can probably just use a [ParDo](https://beam.apache.org/documentation/transforms/java/elementwise/pardo/) instead of mapping into Void. But given that this is a scrap tool, I don't feel strongly about this.



_Code quote:_
```Java
MapElements.into(TypeDescriptor.of(Void.class))
```

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1795)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesPipeline.java` line 104 at r2](https://reviewable.io/reviews/google/nomulus/1795#-NCaYzqTAAZT8eN_XdAM:-NCaYzqU90MEbPrL4-sf:b-5p0mpc) ([raw file](https://github.com/google/nomulus/blob/b09f0f483d8b4287a8d53bfe1076194bc3e2899d/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesPipeline.java#L104)):*
> ```Java
>                             .setType(HistoryEntry.Type.SYNTHETIC)
>                             .build());
>                 outputReceiver.output(null);
> ```

Do you even need this? I think you can just not call the receiver if you do not want to produce downstream data.



_Code quote:_
```Java
outputReceiver.output(null);
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1795)* status: all files reviewed, 1 unresolved discussion

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesPipeline.java` line 104 at r2](https://reviewable.io/reviews/google/nomulus/1795#-NCaYzqTAAZT8eN_XdAM:-NCaZJt176kgOj83mubZ:b-560069) ([raw file](https://github.com/google/nomulus/blob/b09f0f483d8b4287a8d53bfe1076194bc3e2899d/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesPipeline.java#L104)):*

Do you even need this? I think you can just not call the receiver if you do not want to produce downstream data.
</blockquote></details>

hm yeah i don't actually know, I wasn't sure if anything bad would happen if it didn't receive any output

PR #1797, opened 2022-09-28
Create pipeline to reset domains in QA prior to bad pipeline

gbrodman:
This can only be run against the QA instance. If a domain was created prior to the bad pipeline run (2022-09-05T09:00:00.000Z) then we will reset to the last point before the pipeline run. If a domain was created after that point, we reset to the domain creation point. That way, we can replay all DOMAIN_UPDATE objects at or after the pipeline start.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1797)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -711,6 +711,10 @@ createToolTask(
     'createSyntheticDomainHistories',
     'google.registry.tools.javascrap.CreateSyntheticDomainHistoriesPipeline')
 
+createToolTask(
+    'resetDomainsToBeforeResave',
+    'google.registry.tools.javascrap.ResetDomainsToBeforeResavePipeline')
+
 project.tasks.create('generateSqlSchema', JavaExec) {
   classpath = sourceSets.nonprod.runtimeClasspath
   main = 'google.registry.tools.DevTool'
--- a/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java
+++ b/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java
@@ -1,0 +1,124 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools.javascrap;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
+import static google.registry.util.DateTimeUtils.isBeforeOrAt;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Streams;
+import google.registry.beam.common.RegistryJpaIO;
+import google.registry.beam.common.RegistryPipelineOptions;
+import google.registry.config.RegistryEnvironment;
+import google.registry.model.domain.Domain;
+import google.registry.model.domain.DomainBase;
+import google.registry.model.domain.DomainHistory;
+import google.registry.model.reporting.HistoryEntryDao;
+import google.registry.persistence.PersistenceModule.TransactionIsolationLevel;
+import google.registry.persistence.VKey;
+import java.io.Serializable;
+import java.util.Optional;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.joda.time.DateTime;
+
+/**
+ * Pipeline that resets domains to their state prior to the buggy pipeline that caused b/245940594.
+ *
+ * <p>The pipeline started at 2022-09-05T09:00:00.000Z. If the domain has any history objects before
+ * that time, we reset the domain to the last one before that time. If not, we reset the domain to
+ * the first time (the creation time). That way, it will be valid to replay all (and only)
+ * DOMAIN_UPDATES after the pipeline start time.
+ *
+ * <p>We will not alter contacts or hosts -- this allows domains created post-2022-09-05 to maintain
+ * their foreign key constraints, and it means that we won't need to replay any contact or host EPP
+ * actions.
+ *
+ * <p>After resetting domains to 2022-09-05, we can replay only the DOMAIN_UPDATE EPP actions that
+ * took place after that point, up until the first DomainHistory post-2022-09-10T12:00:00Z (the end
+ * of the pipeline, roughly). At that point we can compare the fields of our domain object to the
+ * domain object stored in that DomainHistory. If the fields are different, then that means that the
+ * domain was likely affected by the overwrite bug.
+ *
+ * <p>To run the pipeline:
+ *
+ * <p><code>
+ * $ ./nom_build :core:resetDomainsToBeforeResave --args="--region=us-central1
+ *   --runner=DataflowRunner
+ *   --registryEnvironment=QA
+ *   --project={project-id}
+ *   --workerMachineType=n2-standard-4"
+ * </code>
+ */
+public class ResetDomainsToBeforeResavePipeline implements Serializable {
+
+  private static final DateTime BAD_PIPELINE_START_TIME =
+      DateTime.parse("2022-09-05T09:00:00.000Z");
+
+  static void setup(Pipeline pipeline) {
+    pipeline
+        .apply(
+            "Select all domain repo IDs",
+            RegistryJpaIO.read(
+                "SELECT d.repoId FROM Domain d", String.class, r -> VKey.create(Domain.class, r)))
+        .apply("Reset each domain", ParDo.of(new DomainResetFunction()));
+  }
+
+  private static class DomainResetFunction extends DoFn<VKey<Domain>, Void> {
+
+    @ProcessElement
+    public void processElement(
+        @Element VKey<Domain> key, PipelineOptions options, OutputReceiver<Void> outputReceiver) {
+      jpaTm()
+          .transact(
+              () -> {
+                // note: these are already sorted from earliest to latest
+                ImmutableList<DomainHistory> allHistories =
+                    HistoryEntryDao.loadHistoryObjectsForResource(key, DomainHistory.class);
+                // If the domain was created before the pipeline, use the last history before it
+                Optional<DomainBase> possibleDomainBeforePipeline =
+                    Streams.findLast(
+                        allHistories.stream()
+                            .filter(
+                                dh ->
+                                    isBeforeOrAt(dh.getModificationTime(), BAD_PIPELINE_START_TIME))
+                            .map(dh -> dh.getDomainBase().get()));
+                // Otherwise (created after the pipeline), use the first (CREATE) history
+                DomainBase domainToPersist =
+                    possibleDomainBeforePipeline.orElseGet(
+                        () -> allHistories.get(0).getDomainBase().get());
+                jpaTm().put(new Domain.Builder().copyFrom(domainToPersist).build());
+              });
+    }
+  }
+
+  public static void main(String[] args) {
+    RegistryPipelineOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(RegistryPipelineOptions.class);
+    checkArgument(
+        options.getRegistryEnvironment().equals(RegistryEnvironment.QA),
+        "This command is only meant to be run against the QA instance");
+    RegistryPipelineOptions.validateRegistryPipelineOptions(options);
+    options.setIsolationOverride(TransactionIsolationLevel.TRANSACTION_REPEATABLE_READ);
+
+    Pipeline pipeline = Pipeline.create(options);
+    setup(pipeline);
+    pipeline.run();
+  }
+}
--- a/core/src/test/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipelineTest.java
+++ b/core/src/test/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipelineTest.java
@@ -1,0 +1,203 @@
+// Copyright 2022 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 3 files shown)


jianglai (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1797)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8bSTs3jfp3ynUl0WL:-ND8bSTs3jfp3ynUl0WM:blz3g6k) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L45)):*
> ```Java
>  *
>  * <p>The pipeline started at 2022-09-05T09:00:00.000Z. If the domain has any history objects before
>  * that time, we reset the domain to the last one before that time. If not, we reset the domain to
> ```

I don't think we need to be concerned about domains created during the pipeline run. The domains to be re-saved are determined by the query that ran at the start of the pipeline, which means if a domain is created afterward, it is never included, so never re-saved. We do not need to worry about subsequent updates to these domains as a result because they are never affected.

Note that the pipeline runs at [isolation level read committed](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java;l=94?q=resaveall&ss=nomulus%2Fnomulus), which [only sees data committed before the query starts](https://www.postgresql.org/docs/7.2/xact-read-committed.html).



_Code quote:_
```Java
 * that time, we reset the domain to the last one before that time. If not, we reset the domain to
 * the first time (the creation time). That way, it will be valid to replay all (and only)
```
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 79 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8mXkf74fALOWilY2w:-ND8mXkf74fALOWilY2x:b8moj4c) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L79)):*
> ```Java
>             "Select all domain repo IDs",
>             RegistryJpaIO.read(
>                 "SELECT d.repoId FROM Domain d", String.class, r -> VKey.create(Domain.class, r)))
> ```

Not sure if it matters, but it might be more efficient to load the history entries for each domain with max revision id before bad pipeline start time, and then just take the `domainBase`s inside them and persist them. That should do all the resets that we need?



_Code quote:_
```Java
            RegistryJpaIO.read(
                "SELECT d.repoId FROM Domain d", String.class, r -> VKey.create(Domain.class, r)))
```
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 115 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8lmW_EnwAsLzX21ur:-ND8lmW_EnwAsLzX21us:b-46rxxj) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L115)):*
> ```Java
>         PipelineOptionsFactory.fromArgs(args).withValidation().as(RegistryPipelineOptions.class);
>     checkArgument(
>         options.getRegistryEnvironment().equals(RegistryEnvironment.QA),
> ```

Since this is only for QA, maybe we can just hard code it instead of having to pass in a parameter with a fixed value?



_Code quote:_
```Java
        options.getRegistryEnvironment().equals(RegistryEnvironment.QA),
        "This command is only meant to be run against the QA instance");
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1797)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8bSTs3jfp3ynUl0WL:-ND8nRqB0anyCDaTsRmF:b-cxjinn) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L45)):*

I don't think we need to be concerned about domains created during the pipeline run. The domains to be re-saved are determined by the query that ran at the start of the pipeline, which means if a domain is created afterward, it is never included, so never re-saved. We do not need to worry about subsequent updates to these domains as a result because they are never affected.

Note that the pipeline runs at [isolation level read committed](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/beam/resave/ResaveAllEppResourcesPipeline.java;l=94?q=resaveall&ss=nomulus%2Fnomulus), which [only sees data committed before the query starts](https://www.postgresql.org/docs/7.2/xact-read-committed.html).
</blockquote></details>

In order to make future actions simpler, we do want to handle domains created during or after the bad pipeline run. If not, we could see the following situation:

1. Domain created during pipeline run
2. Domain updated during/after pipeline run
3. If we don't go back to the domain as it was in step 1, if we try to reapply the update from 2, that application could fail. 

Basically, if we reset all domains to "creation time" or "last update before pipeline" (whichever comes last), then we can seamlessly apply all DOMAIN_UPDATEs starting at the pipeline start without worrying about if they've been applied already. 
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 79 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8mXkf74fALOWilY2w:-ND8o7Tp89FfPc-KmhbK:b-5wyw00) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L79)):*

Not sure if it matters, but it might be more efficient to load the history entries for each domain with max revision id before bad pipeline start time, and then just take the `domainBase`s inside them and persist them. That should do all the resets that we need?
</blockquote></details>

See the other comment -- I think we may want to be a bit more clever so that we can apply all updates en masse 
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 115 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8lmW_EnwAsLzX21ur:-ND8oD0D3zYFDxQMaDpt:b-cjolbw) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L115)):*

Since this is only for QA, maybe we can just hard code it instead of having to pass in a parameter with a fixed value?
</blockquote></details>

i think because we're using the standard one-size-fits-all registry options, we want to have this check here instead of hard-coding QA. My primary reason is that I'd want to avoid the situation where the user passes in a registry environment other than QA which we ignore, doing it on QA anyway.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1797)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8bSTs3jfp3ynUl0WL:-NDGZVvC8P9IDUK1mVZn:bvjeqpv) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L45)):*

In order to make future actions simpler, we do want to handle domains created during or after the bad pipeline run. If not, we could see the following situation:

1. Domain created during pipeline run
2. Domain updated during/after pipeline run
3. If we don't go back to the domain as it was in step 1, if we try to reapply the update from 2, that application could fail. 

Basically, if we reset all domains to "creation time" or "last update before pipeline" (whichever comes last), then we can seamlessly apply all DOMAIN_UPDATEs starting at the pipeline start without worrying about if they've been applied already. 
</blockquote></details>

Fair point. I however was thinking about identifying the pool of domains from domain histories, not domains themselves, which should shrink the pool and get around the concerns you have.

Basically, we know that only domains with a history entry in the impacted window would potentially be in a bad state, so it makes sense to only restore those domains to a prior state and replay the history on them. We can then query all the history in the window, including the one immediately before and after the window, and group them by domain repo ID, to find the oldest state to revert to. If a domain's first history entry in this list is after the start time, we know that it didn't exist before the pipeline starts, so we can safely ignore the whole domain.

The same logic can apply to when we compare the domains after the replay, we find the domains to compare from the above history entry list that was used to reset domains and to replay, instead of the other way around, namely finding all domains that exist in the domain table and matching them to history entries.

I believe by limiting us to operate on only domains that has history entries during the window, we can reduce unnecessary corner cases that we had to deal with during comparison. For example, right now I'm seeing some discrepancies for domains that have not been updated in years, which should not been included in the comparison at all, but are because I'm comparing all embedded domains within the most recent history entries. This suggests that a more limited pool to perform the comparison would not only save computer time, but human time in identifying false positives.

In the end, it might not matter much if the false positive rate is low enough. Right now I'm seeing 22 false positives in the first 1000 domains (alphabetically) that I compared. Most of them due to missing ds data in the embedded resources. 
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 79 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8mXkf74fALOWilY2w:-NDIrmrwEP6WtKo69zc4:b-q64pkt) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L79)):*

See the other comment -- I think we may want to be a bit more clever so that we can apply all updates en masse 
</blockquote></details>

See above.
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 115 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8lmW_EnwAsLzX21ur:-NDIrtHH93zHDNvRhrQK:b-yfvt4z) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L115)):*

i think because we're using the standard one-size-fits-all registry options, we want to have this check here instead of hard-coding QA. My primary reason is that I'd want to avoid the situation where the user passes in a registry environment other than QA which we ignore, doing it on QA anyway. 
</blockquote></details>

lol, I think since you are the only one running this pipeline, the chance of that happening is pretty slim. But sure, I guess it's better to give explicit error than silently correcting it.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1797)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8bSTs3jfp3ynUl0WL:-NDTsuyh9sK-zWcnMRq9:b-3xz365) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L45)):*
Yeah if we're replaying everything in bulk and comparing the pure resources we're going to get a TON of false positives, for that reason. That's why i was suggesting that we only compare relevant fields, specifically 

```
dsData
status values
contacts
registrant
auth info
nameservers
```

do you happen to know offhand how many of those false positives would still be false positives if we only compare those fields? 

> If a domain's first history entry in this list is after the start time, we know that it didn't exist before the pipeline starts, so we can safely ignore the whole domain.

This is not quite true, a domain could have been both created and updated during the pipeline. 
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 79 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8mXkf74fALOWilY2w:-NDTsT0k9zIvtEA1tfHu:b-896fix) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L79)):*

See above.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 115 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8lmW_EnwAsLzX21ur:-NDTsTzG4e0nJT3-9ivA:b-896fix) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L115)):*

lol, I think since you are the only one running this pipeline, the chance of that happening is pretty slim. But sure, I guess it's better to give explicit error than silently correcting it.
</blockquote></details>

Done.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1797)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1797#-ND8bSTs3jfp3ynUl0WL:-NDTxiAX4FLDm7ER50_0:b1spdae) ([raw file](https://github.com/google/nomulus/blob/71a188ac7af5a4e4569acaec929fd8b701189ad3/core/src/main/java/google/registry/tools/javascrap/ResetDomainsToBeforeResavePipeline.java#L45)):*

Yeah if we're replaying everything in bulk and comparing the pure resources we're going to get a TON of false positives, for that reason. That's why i was suggesting that we only compare relevant fields, specifically 

```
dsData
status values
contacts
registrant
auth info
nameservers
```

do you happen to know offhand how many of those false positives would still be false positives if we only compare those fields? 

> If a domain's first history entry in this list is after the start time, we know that it didn't exist before the pipeline starts, so we can safely ignore the whole domain.

This is not quite true, a domain could have been both created and updated during the pipeline. 


</blockquote></details>

From what I saw, only dsData might be problematic, which hopefully can be solved by rerunning the pipeline and recreating the synthetic entries. The exact cause of missing da data histories is still unclear.

A domain could have been created and updated during the pipeline, but since we are running with "read committed" isolation level, any domain that was not present by the time the initial query ran would not be included the subsequent processing stages.

https://www.postgresql.org/docs/7.2/xact-read-committed.html

So the domains that were created during the pipeline run would never have been picked up by the pipeline to be resaved.

gbrodman:
this is moot

PR #1802, opened 2022-09-30
Change some READ_COMMITTED levels to REPEATABLE_READ

gbrodman:
Basically, any time we're loading a bunch of linked objects that might change, we want to have REPEATABLE_READ so that another transaction doesn't come along and smush whatever we think we're loading.

The following instances of READ_COMMITTED haven't changed:
- RdePipeline (it only loads immutable objects like histories)
- Invoicing pipeline (only immutable objects like BillingEvents)
- Spec11 (doesn't use any linked info from Domain)

This also changes the PersistenceModule to use REPEATABLE_READ by default on the replica JPA TM, for the standard reasoning.


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1802)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1803, opened 2022-09-30
Modify the CreateSynthetic pipeline to run over all non-deleted domains

gbrodman:
We're seeing some discrepancies with the history objects created with the previous READ_COMMITTED pipeline so we can make this change to create a new set of history objects in QA now that it's set to REPEATABLE_READ

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1803)


jianglai (approved):
I think we only need to create entries for domains whose latest entry is a synthetic entry, if all we want is to correct potential synthetic entry discrepancies?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1803#-:-NDIt4n6Dx-Y0cqtbKc4:bgrszla)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (commented):
Yeah that's definitely true, but it shouldn't matter since this only takes like 15 minutes or so to run anyway. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1803#-:-NDUGUV19bux3oSrERRM:by1scmv)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jianglai)

PR #1805, opened 2022-10-03
Use injected times in URSC and CommandTestCase

gbrodman:
We started getting failures because some of the tests used October. In general we should freeze the clock for testing as much as possible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1805)

### DIFF
--- a/core/src/main/java/google/registry/tools/GetEppResourceCommand.java
+++ b/core/src/main/java/google/registry/tools/GetEppResourceCommand.java
@@ -15,29 +15,30 @@
 package google.registry.tools;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static org.joda.time.DateTimeZone.UTC;
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
 import google.registry.model.EppResource;
+import google.registry.util.Clock;
 import java.util.Optional;
+import javax.inject.Inject;
 import org.joda.time.DateTime;
 
 /** Abstract command to print one or more resources to stdout. */
 @Parameters(separators = " =")
 abstract class GetEppResourceCommand implements CommandWithRemoteApi {
 
-  private final DateTime now = DateTime.now(UTC);
-
   @Parameter(
       names = "--read_timestamp",
       description = "Timestamp to use when reading. May not be in the past.")
-  protected DateTime readTimestamp = now;
+  protected DateTime readTimestamp;
 
   @Parameter(
       names = "--expand",
       description = "Fully expand the requested resource. NOTE: Output may be lengthy.")
   boolean expand;
+
+  @Inject Clock clock;
 
   /** Runs the command's own logic that calls {@link #printResource}. */
   abstract void runAndPrint();
@@ -59,7 +60,11 @@ abstract class GetEppResourceCommand implements CommandWithRemoteApi {
 
   @Override
   public void run() {
-    checkArgument(!readTimestamp.isBefore(now), "--read_timestamp may not be in the past");
+    if (readTimestamp == null) {
+      readTimestamp = clock.nowUtc();
+    }
+    checkArgument(
+        !readTimestamp.isBefore(clock.nowUtc()), "--read_timestamp may not be in the past");
     runAndPrint();
   }
 }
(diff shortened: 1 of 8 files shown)


gbrodman:
github / kokoro integration is being wonky for now, but this should pass tests

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1805)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/tools/GetEppResourceCommand.java` line 60 at r1](https://reviewable.io/reviews/google/nomulus/1805#-NDTz4a01krFgNvmy_Qf:-NDTz4a01krFgNvmy_Qg:b-bm90tn) ([raw file](https://github.com/google/nomulus/blob/0606694bb69d95accab77444c78a338a5f0143b2/core/src/main/java/google/registry/tools/GetEppResourceCommand.java#L60)):*
> ```Java
>   @Override
>   public void run() {
>     DateTime now = tm().transact(() -> tm().getTransactionTime());
> ```

Is there a reason why we want to get the time from a transaction? The `run()` method doesn't run in a transaction, does it? The old way seems to be working just fine. Same question for the change in the next file.



_Code quote:_
```Java
    DateTime now = tm().transact(() -> tm().getTransactionTime());
    if (readTimestamp == null) {
      readTimestamp = now;
    }
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1805)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/tools/GetEppResourceCommand.java` line 60 at r1](https://reviewable.io/reviews/google/nomulus/1805#-NDTz4a01krFgNvmy_Qf:-NDULcrU33DuiRK261ae:b-w9sqgv) ([raw file](https://github.com/google/nomulus/blob/0606694bb69d95accab77444c78a338a5f0143b2/core/src/main/java/google/registry/tools/GetEppResourceCommand.java#L60)):*

Is there a reason why we want to get the time from a transaction? The `run()` method doesn't run in a transaction, does it? The old way seems to be working just fine. Same question for the change in the next file.
</blockquote></details>

It does not run in a transaction. You are correct that the tests for these classes, as they stand, work fine. However changing the FakeClock in the CommandTestCase causes some of them to break. 

Even if they work now, we can't guarantee that they'll work in the future because any usage of DateTime.now() is potentially dangerous to tests.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1805)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/tools/GetEppResourceCommand.java` line 60 at r1](https://reviewable.io/reviews/google/nomulus/1805#-NDTz4a01krFgNvmy_Qf:-NDVvEFvB7pt_QAhiJIZ:bb89joa) ([raw file](https://github.com/google/nomulus/blob/0606694bb69d95accab77444c78a338a5f0143b2/core/src/main/java/google/registry/tools/GetEppResourceCommand.java#L60)):*

It does not run in a transaction. You are correct that the tests for these classes, as they stand, work fine. However changing the FakeClock in the CommandTestCase causes some of them to break. 

Even if they work now, we can't guarantee that they'll work in the future because any usage of DateTime.now() is potentially dangerous to tests. 
</blockquote></details>

I am still a bit confused as to why the change here is necessary. I understand that the URS command test used a fake clock that was initialized to `DateTime.now()`, but some resources were constructed with expiration date as of 2022-10-01, thus leading to the failure.

But that seems to be fixed by just setting the clock to a fixed value in `CommandTestCase`? You also changed a few other tests which depends on `DateTime.now()` to depend on the fake clock instead, which makes sense.

But why does the production code needs to change to use the transaction time as well? Unless the return value of `DateTime.now()` is somehow affected by the fake clock in tests, calling `tm().transact(() -> tm().getTransactionTime())` should effectively give you the same time as calling `DateTime.now()`, I think.

I feel like I'm missing something (maybe obvious) here... :O

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1805)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1784, opened 2022-09-13
Rename various fields and classes after migration

jianglai:
Also fixed a bug introduced in #1785 where identity checked were performed instead of equality. This resulted in two sets containing the same elements not being regarded as equal and subsequent DNS updated being unnecessarily enqueued.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1784)

PR #1806, opened 2022-10-04
Add fallback for Spec11 ThreatMatch parser

ptkach:
The issue we encountered with Spec11 failing to publish reports on 09/29/2022 is a side effect of https://github.com/google/nomulus/pull/1784 change. 
In nutshell the root cause is that Spec11 parser processes data with 1 day lag, so at the time release went through it went on processing data saved with ```fullyQualifiedDomainName```, the next day it autocorrected itself due to saved files usage of ```domainName``` as a field name. 

However we still want to be able to parse older files stored in a cloud, so this change adds a support for both older field and newer field.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1806)

### DIFF
--- a/core/src/main/java/google/registry/beam/spec11/ThreatMatch.java
+++ b/core/src/main/java/google/registry/beam/spec11/ThreatMatch.java
@@ -26,6 +26,7 @@ public abstract class ThreatMatch implements Serializable {
 
   private static final String THREAT_TYPE_FIELD = "threatType";
   private static final String DOMAIN_NAME_FIELD = "domainName";
+  private static final String OUTDATED_NAME_FIELD = "fullyQualifiedDomainName";
 
   /** Returns what kind of threat it is (malware, phishing etc.) */
   public abstract String threatType();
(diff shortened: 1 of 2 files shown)


jianglai (approved):
Good catch! I think we only need to keep the old field name for a month or so. We send daily diffs and a monthly snapshot. So after a month we should never need to parse the old files anymore. Add a `TODO` to delete the fallback later, maybe in 2023. That way even if we don't delete it right now, next time someone sees it they'll know it's safe to delete.

In truth, the monthly snapshot only looks at the current output, so I think it's only that one daily diff that would error out, but let's err on the side of safety.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1806#-:-ND_w26a0cMr-5PKmG31:btbdl4u)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`core/src/test/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParserTest.java` line 111 at r1](https://reviewable.io/reviews/google/nomulus/1806#-ND_w-Qj-z1EOpUf42nM:-ND_w-Qk-QXMxMZTMsqU:b-43bgnu) ([raw file](https://github.com/google/nomulus/blob/4a8ff150e4ba1939a1e80fdd305ea8b06e370503/core/src/test/java/google/registry/reporting/spec11/Spec11RegistrarThreatMatchesParserTest.java#L111)):*
> ```Java
>                     "threatType", "MALWARE",
>                     "fullyQualifiedDomainName", "c.com")));
>     ThreatMatch objectWithoutOutdatedFields =
> ```

`WithNewFields`.



_Code quote:_
```Java
WithoutOutdatedFields
```

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1806)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

PR #1807, opened 2022-10-04
Limit environments allowed to send emails out

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1807)

### DIFF
--- a/util/src/main/java/google/registry/util/SendEmailService.java
+++ b/util/src/main/java/google/registry/util/SendEmailService.java
@@ -14,8 +14,12 @@
 
 package google.registry.util;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.Iterables.toArray;
 
+import com.google.common.base.Ascii;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.flogger.FluentLogger;
 import com.google.common.net.MediaType;
 import google.registry.util.EmailMessage.Attachment;
 import java.io.IOException;
@@ -44,6 +48,10 @@ public class SendEmailService {
   private final Retrier retrier;
   private final TransportEmailSender transportEmailSender;
 
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+  private static final ImmutableSet<String> ALLOWED_ENVS =
+      ImmutableSet.of("PRODUCTION", "UNITTEST");
+
   @Inject
   SendEmailService(Retrier retrier, TransportEmailSender transportEmailSender) {
     this.retrier = retrier;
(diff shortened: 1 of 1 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1807)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 62 at r1](https://reviewable.io/reviews/google/nomulus/1807#-ND_xGFUCUmR0g0JkrLV:-ND_xGFUCUmR0g0JkrLW:bl6osmq) ([raw file](https://github.com/google/nomulus/blob/b120924f2abfef1aa72e2f84b8d60395683a2d1f/util/src/main/java/google/registry/util/SendEmailService.java#L62)):*
> ```Java
>    */
>   public void sendEmail(EmailMessage emailMessage) {
>     if (ALLOWED_ENVS.contains(
> ```

Let's make it configurable through the config file, rather then hard-coding it. That way it's easier (and cleaner) if we want to enable it for any non-prod environment for any reason. Your can then inject this boolean, which makes testing possible (i. e. spying on `transportEmailSender` to make sure that it does not send emails when the boolean is false).

Also, I think it might be a good idea to log the email that would have been sent, that way we probably won't need to enable emailing for any environment for testing purpose either.

So basically, based on the config value, you call `transportEmailSender.sendMessage(msg)`, or `log.atInfo().log(msg)`, assuming `msg` has a nice string representation.



_Code quote:_
```Java
    if (ALLOWED_ENVS.contains(
        Ascii.toUpperCase(System.getProperty("google.registry.environment", "UNITTEST")))) {
```

ptkach (commented):
As agreed I'm not logging an entire message, only substring of a subject. I think that should be enough for the purpose.

I've tested the change in alpha by adding artificial http listener that triggered an attempt to send an email, verification was successful and without "ALPHA" in ALLOWED_ENVS I could see the log and no email and vice-versa with "ALPHA".

On being extra cautious side - the only thing that worries is me is that I'm not 100% certain that ```google.registry.environment``` is configured for every environment properly. As far as I can tell from source code it should be, but please confirm. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1807#-:-NDigNsKCSMp3TbX1XOa:biwhu5r)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @ptkach)

jianglai (approved):
Yes, that property is essential for each environment to work property, so it is definitely configured.

As a side note (and a procedural matter), you'll want to resolve each comment that the reviewer had before publishing your round of comment. This is in addition to making the changes to the code, of course.

Basically, you'll want to see the red "to reply" count in the top bar to be grey zero instead.

Also, a reviewer might approve your PR with some comments, in that case the PR is basically good to go without needing further reviews. As along as you resolve those lingering comments (you can choose to do what's suggested, or not if you can give a good reason), there's no need to wait for further approval before merging.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1807#-:-NDkEvXh4zHrXo67M7Xh:bi183m8)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 62 at r1](https://reviewable.io/reviews/google/nomulus/1807#-ND_xGFUCUmR0g0JkrLV:-NDiopeb42sKYP_gR7A1:b34n1gs) ([raw file](https://github.com/google/nomulus/blob/b120924f2abfef1aa72e2f84b8d60395683a2d1f/util/src/main/java/google/registry/util/SendEmailService.java#L62)):*

Let's make it configurable through the config file, rather then hard-coding it. That way it's easier (and cleaner) if we want to enable it for any non-prod environment for any reason. Your can then inject this boolean, which makes testing possible (i. e. spying on `transportEmailSender` to make sure that it does not send emails when the boolean is false).

Also, I think it might be a good idea to log the email that would have been sent, that way we probably won't need to enable emailing for any environment for testing purpose either.

So basically, based on the config value, you call `transportEmailSender.sendMessage(msg)`, or `log.atInfo().log(msg)`, assuming `msg` has a nice string representation.
</blockquote></details>

I guess we cannot call `RegistryEnvironment.get()` because of the circular dependency too?
___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 51 at r2](https://reviewable.io/reviews/google/nomulus/1807#-NDioS_i2zR7F1Eodz5T:-NDioS_i2zR7F1Eodz5U:b-q9uzsf) ([raw file](https://github.com/google/nomulus/blob/4d4fd8a4f609e80a9740f44473a5da73ee493ded/util/src/main/java/google/registry/util/SendEmailService.java#L51)):*
> ```Java
> 
>   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
>   private static final ImmutableSet<String> ALLOWED_ENVS =
> ```

You may want to add ALPHA to the list for ease of testing.



_Code quote:_
```Java
ALLOWED_ENVS
```
___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 69 at r2](https://reviewable.io/reviews/google/nomulus/1807#-NDip0kf1RBlhSNgYpfe:-NDip0kgC-6YdBie7Ev1:b7cr22j) ([raw file](https://github.com/google/nomulus/blob/4d4fd8a4f609e80a9740f44473a5da73ee493ded/util/src/main/java/google/registry/util/SendEmailService.java#L69)):*
> ```Java
>       logger.atInfo().log(
>           String.format(
>               "Email with subject %s... was reverted from being sent out in"
> ```

A better expression:

`"Email with subject %s would have been sent to recipient %s".`



_Code quote:_
```Java
              "Email with subject %s... was reverted from being sent out in"
                  + " incompatible environment",
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1807)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 62 at r1](https://reviewable.io/reviews/google/nomulus/1807#-ND_xGFUCUmR0g0JkrLV:-NDn3NeUAdZOoVKzFyNx:bxmtbwy) ([raw file](https://github.com/google/nomulus/blob/b120924f2abfef1aa72e2f84b8d60395683a2d1f/util/src/main/java/google/registry/util/SendEmailService.java#L62)):*

I guess we cannot call `RegistryEnvironment.get()` because of the circular dependency too?
</blockquote></details>

Yep, that was the first think I tried actually
___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 51 at r2](https://reviewable.io/reviews/google/nomulus/1807#-NDioS_i2zR7F1Eodz5T:-NDn2ULD0fmtJvo6R4I4:b-f2v52w) ([raw file](https://github.com/google/nomulus/blob/4d4fd8a4f609e80a9740f44473a5da73ee493ded/util/src/main/java/google/registry/util/SendEmailService.java#L51)):*

You may want to add ALPHA to the list for ease of testing.
</blockquote></details>

Both dns pull and publish queues seem to be enabled in alpha, so in order to lower the possibility of outgoing email I think it'd be better not to add it
___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 69 at r2](https://reviewable.io/reviews/google/nomulus/1807#-NDip0kf1RBlhSNgYpfe:-NDn3Qmt-f091uLzPSTe:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d4fd8a4f609e80a9740f44473a5da73ee493ded/util/src/main/java/google/registry/util/SendEmailService.java#L69)):*

A better expression:

`"Email with subject %s would have been sent to recipient %s".`
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1807)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`util/src/main/java/google/registry/util/SendEmailService.java` line 51 at r2](https://reviewable.io/reviews/google/nomulus/1807#-NDioS_i2zR7F1Eodz5T:-NDn7hkR5zyPp3RNX9A4:bx6kjaj) ([raw file](https://github.com/google/nomulus/blob/4d4fd8a4f609e80a9740f44473a5da73ee493ded/util/src/main/java/google/registry/util/SendEmailService.java#L51)):*

Both dns pull and publish queues seem to be enabled in alpha, so in order to lower the possibility of outgoing email I think it'd be better not to add it
</blockquote></details>

That's not really a risk because alpha does not have real email addresses configured. But if you feel strongly about it...

PR #1809, opened 2022-10-05
Donot self-allocate IDs in Beam by default.

jianglai:
Per b/250948425, it is dangerous to implicitly allow all Beam pipelines
to create buildables by self allocating the IDs. This change makes it so
that one has to explicitly use self allocated IDs in Beam.

A boolean is added to the pipeline option so that it can be passed to
the beam worker initializer that controls the behavior of the JVM on
each worker. Note that we did not add the option in the metadata.json file
because we did not want people to use the override at run time when launching
a pipeline, due to the risk. As shown in RdePipeline.java, we instead
explicitly hard-code the option in the pipeline. There is nothing that
stops one to supply that option when launching the pipeline, but it's
not advised.

Tested=deployed the pipeline alpha and ran it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1809)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 78 at r2](https://reviewable.io/reviews/google/nomulus/1809#-NDhva4G3FPiV2Zgqox_:-NDhxM0o0ddVmzI_rOl5:bniurfb) ([raw file](https://github.com/google/nomulus/blob/a866ab8849ad715d3605101d2d0771960252c3ce/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L78)):*

Instead of using a flag, may be we can provide a bring-your-own-id-service api.

Open up IdService to accept user provided Supplier<Long> if the environment is BEAM,
and do not provide a default implementation (throw is on beam and supplier is not provided.).


</blockquote></details>

This way we can plug in a raw datastore api to allocate ids.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[`core/src/main/java/google/registry/model/IdService.java` line 67 at r2](https://reviewable.io/reviews/google/nomulus/1809#-NDilHmqC2lV9RN0kVIR:-NDilHmqC2lV9RN0kVIS:b-qtgqik) ([raw file](https://github.com/google/nomulus/blob/a866ab8849ad715d3605101d2d0771960252c3ce/core/src/main/java/google/registry/model/IdService.java#L67)):*
> ```Java
>    * @see #isSelfAllocated
>    */
>   public static void useSelfAllocatedId() {
> ```

Can we use different verbiage for this?  E.g. "fake id", "testing id", "fake testing id" or similar?  I don't think "self-allocated ID" conveys the right meaning, namely, that this is horribly unsafe for anything in production.

jianglai (commented):
PTAL.

> Reviewed 4 of 4 files at r1, 2 of 2 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809#-:-NEb2igP0fuxc-LWewVJ:b-83330m)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 78 at r2](https://reviewable.io/reviews/google/nomulus/1809#-NDhva4G3FPiV2Zgqox_:-NEab5lhAMYmOQQnhC54:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a866ab8849ad715d3605101d2d0771960252c3ce/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L78)):*

This way we can plug in a raw datastore api to allocate ids.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/IdService.java` line 67 at r2](https://reviewable.io/reviews/google/nomulus/1809#-NDilHmqC2lV9RN0kVIR:-NEb1xj6BozhO2i2HPqv:bg3zmnq) ([raw file](https://github.com/google/nomulus/blob/a866ab8849ad715d3605101d2d0771960252c3ce/core/src/main/java/google/registry/model/IdService.java#L67)):*

Can we use different verbiage for this?  E.g. "fake id", "testing id", "fake testing id" or similar?  I don't think "self-allocated ID" conveys the right meaning, namely, that this is horribly unsafe for anything in production.
</blockquote></details>

Reworked the class to take a ID supplier per Weimin's suggestion. I think the supplier used in test can still be called `SelfAllocatedIdSupplier` as it is an accurate description of what it does, whereas a `FakeIdSupplier` does not convey exactly what kind of "fake" it is, e.g. is it a constant (which should probably better named `ConstantIdSupplier`), or random (`RandomIdSupplier`).

weiminyu:
*[`core/src/main/java/google/registry/model/IdService.java` line 48 at r3](https://reviewable.io/reviews/google/nomulus/1809#-NEbM3Ds3-XD8QAxXLjG:-NEbM3Ds3-XD8QAxXLjH:ba4lahu) ([raw file](https://github.com/google/nomulus/blob/733fc8a017b276ec74e721b84af5bc9abb60b0dd/core/src/main/java/google/registry/model/IdService.java#L48)):*
> ```Java
>   private static Supplier<Long> idSupplier =
>       RegistryEnvironment.UNITTEST.equals(RegistryEnvironment.get())
>           ? SelfAllocatedIdSupplier.getInstance()
> ```

Building on Ben's comment, maybe we should call this class NonUniqueIdSupplier, and
make the useSelfAllocatedId BEAM flag a Enum with NonUniqueId, OfyId, and maybe RawDatastoreId if it comes to that.



_Code quote:_
```Java
SelfAllocatedIdSupplier
```

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/model/IdService.java` line 48 at r3](https://reviewable.io/reviews/google/nomulus/1809#-NEbM3Ds3-XD8QAxXLjG:-NEbfH8N9ooGCV0DVBb1:b-2s2foo) ([raw file](https://github.com/google/nomulus/blob/733fc8a017b276ec74e721b84af5bc9abb60b0dd/core/src/main/java/google/registry/model/IdService.java#L48)):*

Building on Ben's comment, maybe we should call this class NonUniqueIdSupplier, and
make the useSelfAllocatedId BEAM flag a Enum with NonUniqueId, OfyId, and maybe RawDatastoreId if it comes to that.

</blockquote></details>

Just to clarify, do you mean to have a lookup map from the Beam enums to ID suppliers that one use to determine the override?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/model/IdService.java` line 48 at r3](https://reviewable.io/reviews/google/nomulus/1809#-NEbM3Ds3-XD8QAxXLjG:-NEkoyTU5f_0AAXF7fR_:blun7qw) ([raw file](https://github.com/google/nomulus/blob/733fc8a017b276ec74e721b84af5bc9abb60b0dd/core/src/main/java/google/registry/model/IdService.java#L48)):*

Just to clarify, do you mean to have a lookup map from the Beam enums to ID suppliers that one use to determine the override?
</blockquote></details>

Also, I think the current iteration is explicit enough to discourage the use of ID supplier overrides. It also allows for the further migration to a SQL-provided ID (as another ID provider), by which time we can simply remove the setter, as Beam has access to SQL. It doesn't seem worth it to make the Beam flags more flexible as it would just be throwaway work.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

jianglai (commented):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1809)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

PR #1810, opened 2022-10-07
Prevent saving duplicate rows in spec11 pipeline

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1810)

### DIFF
--- a/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java
+++ b/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java
@@ -26,6 +26,7 @@ import google.registry.beam.common.RegistryJpaIO;
 import google.registry.beam.common.RegistryJpaIO.Read;
 import google.registry.beam.spec11.SafeBrowsingTransforms.EvaluateSafeBrowsingFn;
 import google.registry.config.RegistryConfig.ConfigModule;
+import google.registry.model.IdService;
 import google.registry.model.domain.Domain;
 import google.registry.model.reporting.Spec11ThreatMatch;
 import google.registry.model.reporting.Spec11ThreatMatch.ThreatType;
@@ -45,6 +46,7 @@ import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.GroupByKey;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.Reshuffle;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.TypeDescriptor;
@@ -154,25 +156,36 @@ public class Spec11Pipeline implements Serializable {
 
   static void saveToSql(
       PCollection<KV<DomainNameInfo, ThreatMatch>> threatMatches, Spec11PipelineOptions options) {
-    String transformId = "Spec11 Threat Matches";
     LocalDate date = LocalDate.parse(options.getDate(), ISODateTimeFormat.date());
-    threatMatches.apply(
-        "Write to Sql: " + transformId,
-        RegistryJpaIO.<KV<DomainNameInfo, ThreatMatch>>write()
-            .withName(transformId)
-            .withBatchSize(options.getSqlWriteBatchSize())
-            .withJpaConverter(
-                (kv) -> {
-                  DomainNameInfo domainNameInfo = kv.getKey();
-                  return new Spec11ThreatMatch.Builder()
-                      .setThreatTypes(
-                          ImmutableSet.of(ThreatType.valueOf(kv.getValue().threatType())))
-                      .setCheckDate(date)
-                      .setDomainName(domainNameInfo.domainName())
-                      .setDomainRepoId(domainNameInfo.domainRepoId())
-                      .setRegistrarId(domainNameInfo.registrarId())
-                      .build();
-                }));
+    String transformId = "Spec11 Threat Matches";
+    threatMatches
+        .apply(
+            "Construct objects",
+            ParDo.of(
+                new DoFn<KV<DomainNameInfo, ThreatMatch>, Spec11ThreatMatch>() {
+                  @ProcessElement
+                  public void processElement(
+                      @Element KV<DomainNameInfo, ThreatMatch> input,
+                      OutputReceiver<Spec11ThreatMatch> output) {
+                    Spec11ThreatMatch spec11ThreatMatch =
+                        new Spec11ThreatMatch.Builder()
+                            .setThreatTypes(
+                                ImmutableSet.of(ThreatType.valueOf(input.getValue().threatType())))
+                            .setCheckDate(date)
+                            .setDomainName(input.getKey().domainName())
+                            .setDomainRepoId(input.getKey().domainRepoId())
+                            .setRegistrarId(input.getKey().registrarId())
+                            .setId(IdService.allocateId())
+                            .build();
+                    output.output(spec11ThreatMatch);
+                  }
+                }))
+        .apply("Prevent Fusing", Reshuffle.viaRandomKey())
+        .apply(
+            "Write to Sql: " + transformId,
+            RegistryJpaIO.<Spec11ThreatMatch>write()
+                .withName(transformId)
+                .withBatchSize(options.getSqlWriteBatchSize()));
   }
 
   static void saveToGcs(
--- a/core/src/main/java/google/registry/model/reporting/Spec11ThreatMatch.java
+++ b/core/src/main/java/google/registry/model/reporting/Spec11ThreatMatch.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableSet;
 import google.registry.model.Buildable;
 import google.registry.model.ImmutableObject;
 import google.registry.util.DomainNameUtils;
+import java.io.Serializable;
 import java.util.Set;
 import javax.persistence.Column;
 import javax.persistence.Entity;
@@ -39,7 +40,7 @@ import org.joda.time.LocalDate;
       @Index(name = "spec11threatmatch_tld_idx", columnList = "tld"),
       @Index(name = "spec11threatmatch_check_date_idx", columnList = "checkDate")
     })
-public class Spec11ThreatMatch extends ImmutableObject implements Buildable {
+public class Spec11ThreatMatch extends ImmutableObject implements Buildable, Serializable {
 
   /** The type of threat detected. */
   public enum ThreatType {
--- a/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
+++ b/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
@@ -280,10 +280,10 @@ class Spec11PipelineTest {
   private void verifySaveToCloudSql() {
     tm().transact(
             () -> {
-              ImmutableList<Spec11ThreatMatch> sqlThreatMatches =
+              ImmutableList<Spec11ThreatMatch> spec11ThreatMatches =
                   Spec11ThreatMatchDao.loadEntriesByDate(tm(), new LocalDate(2020, 1, 27));
-              assertThat(sqlThreatMatches)
-                  .comparingElementsUsing(immutableObjectCorrespondence())
+              assertThat(spec11ThreatMatches)
+                  .comparingElementsUsing(immutableObjectCorrespondence("id"))
                   .containsExactlyElementsIn(sqlThreatMatches);
             });
   }
--- a/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
+++ b/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java
@@ -283,7 +283,7 @@ class Spec11PipelineTest {
               ImmutableList<Spec11ThreatMatch> sqlThreatMatches =
                   Spec11ThreatMatchDao.loadEntriesByDate(tm(), new LocalDate(2020, 1, 27));
               assertThat(sqlThreatMatches)
-                  .comparingElementsUsing(immutableObjectCorrespondence("id"))
+                  .comparingElementsUsing(immutableObjectCorrespondence())
                   .containsExactlyElementsIn(sqlThreatMatches);
             });
   }


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NE6NzbA4kLGG-cL8VkP:babkgxb) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*
> ```Java
>                             .setDomainRepoId(input.getKey().domainRepoId())
>                             .setRegistrarId(input.getKey().registrarId())
>                             .setId(IdService.allocateId())
> ```

IdService is not working correctly on BEAM right now. Each worker has their own sequence generator so there will be collisions.

If we do this, we need to wait until IdService is fixed.  



_Code quote:_
```Java
.setId(IdService.allocateId()
```
___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 184 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6OxRDE23sz8iM-fKH:-NE6OxRDE23sz8iM-fKI:b-q812ej) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L184)):*
> ```Java
>                 }));
> 
>     spec11ThreatMatches.apply("Prevent Fusing", Reshuffle.viaRandomKey());
> ```

spec11ThreatMatches = spec11ThreatMatches.apply(...)

The apply method returns a  new instance. 

It may read better to switch order with the transformId declaration and chain the remaining transformations here.
This way we won't need this assignment.



_Code quote:_
```Java
spec11Threat
```
___
*[`core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Qcne9aLxLPs08Oj3:-NE6Qcne9aLxLPs08Oj4:b-j2v84) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L291)):*
> ```Java
>         .transact(
>             () -> {
>               ImmutableList<Spec11ThreatMatch> sqlThreatMatches =
> ```

This name shadows the instance variable. Better rename it.



_Code quote:_
```Java
sqlThreatMatches
```

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEMnVSW6JUh-heKyoge:b-ruca1g) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

IdService is not working correctly on BEAM right now. Each worker has their own sequence generator so there will be collisions.

If we do this, we need to wait until IdService is fixed.  
</blockquote></details>

Ok, I will wait for b/201547855 to be fixed before merging this. Or do you recommend I do this another way?
___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 184 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6OxRDE23sz8iM-fKH:-NE7f-q62MECT4KvteVO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L184)):*

spec11ThreatMatches = spec11ThreatMatches.apply(...)

The apply method returns a  new instance. 

It may read better to switch order with the transformId declaration and chain the remaining transformations here.
This way we won't need this assignment.




</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java` line 291 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Qcne9aLxLPs08Oj3:-NEMnder9sXAY63MLnkW:b-896fix) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/test/java/google/registry/beam/spec11/Spec11PipelineTest.java#L291)):*

This name shadows the instance variable. Better rename it.
</blockquote></details>

Done.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEMna7f3jyR6d4u-G3z:bkc0frr) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

Ok, I will wait for b/201547855 to be fixed before merging this. Or do you recommend I do this another way?
</blockquote></details>

I'm working on making it possible to bring your own ID allocation provider in #1809.

Given that it probably will be a while before we can start using SQL to allocate the ID, I wonder if we should look into calling the non-GAE datastore API [here](https://cloud.google.com/datastore/docs/reference/libraries) to allocate the ID in Beam. If that works, we might just migrate to that, for the Beam pipeline, and for the main application, which will make it less important (or urgent) to use SQL to allocate IDs.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEb2lerBhuElboAjPbd:bji3131) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

I'm working on making it possible to bring your own ID allocation provider in #1809.

Given that it probably will be a while before we can start using SQL to allocate the ID, I wonder if we should look into calling the non-GAE datastore API [here](https://cloud.google.com/datastore/docs/reference/libraries) to allocate the ID in Beam. If that works, we might just migrate to that, for the Beam pipeline, and for the main application, which will make it less important (or urgent) to use SQL to allocate IDs.
</blockquote></details>

@ptkach  FYI. If switching to SQL sequence is going to take some time, we should look into if we can use the non-GAE datastore API to allocate IDs to unblock this use case.

ptkach:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)_ status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @ptkach and @weiminyu)
> 
> _[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEb2lerBhuElboAjPbd:bji3131) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):_
> 
> @ptkach FYI. If switching to SQL sequence is going to take some time, we should look into if we can use the non-GAE datastore API to allocate IDs to unblock this use case.

@jianglai @CydeWeys  I'm looking at this whole allocate id thing, however there seems to be a lot of history and complexity to it, so it's gonna take time for me to fully assess the problem. However, maybe a dumb question, but still - is there any reason we don't want to update data type and start using UUID generator? This is how it commonly done in similar circumstances?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEh7HJtA4FLRFzB96af:ba7gyo) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

@jianglai @CydeWeys  I'm looking at this whole allocate id thing, however there seems to be a lot of history and complexity to it, so it's gonna take time for me to fully assess the problem. However, maybe a dumb question, but still - is there any reason we don't want to update data type and start using UUID generator? This is how it commonly done in similar circumstances? 
</blockquote></details>

Isn't UUID a combination of letters and numbers? We need a `long` allocated in this case.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEkdc8W061V9-A_O5sV:bnmzj52) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

Isn't UUID a combination of letters and numbers? We need a `long` allocated in this case.
</blockquote></details>

Also, perhaps more importantly, the ID must be generated from the server side to prevent collision.

ptkach:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)_ status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @ptkach, and @weiminyu)
> 
> _[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEkdc8W061V9-A_O5sV:bnmzj52) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):_
> 
> Also, perhaps more importantly, the ID must be generated from the server side to prevent collision.

I realize it's currently long and UUID is a string 😀, which is why I mentioned "update data type". Sorry I wasn't clear enough. I meant update all the places where the data type is expected as long to actually accept a string UUID, then update schemas and that should probably do the trick.

By random UUID generator I suppose we can use ```java.util.UUID. randomUUID``` which is immutable and provides enough entropy for it to be thread-safe. So that should cover beam case. 

Am I missing something?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEksusNA5898enfQM4r:bd2ol5u) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

I realize it's currently long and UUID is a string 😀, which is why I mentioned "update data type". Sorry I wasn't clear enough. I meant update all the places where the data type is expected as long to actually accept a string UUID, then update schemas and that should probably do the trick.

By random UUID generator I suppose we can use ```java.util.UUID. randomUUID``` which is immutable and provides enough entropy for it to be thread-safe. So that should cover beam case. 

Am I missing something? 

</blockquote></details>

Hmmm, I'm not sure if we would want to entertain the idea of changing the primary ID type for most of the entities... It seems like a risky migration that would undoubtedly require some downtime to perform. It certainly warrants some discussion to weight the pros and cons.

But still, as I mentioned earlier, I doubt an client-side allocation scheme would ever work, due to the chance of collision between multiple instances. We are not just talking about thread safety here, but rather distributed workers that all try to allocate globally unique IDs at the same time. This applies to Beam works, or just multiple Nomulus instances serving different clients concurrently. Is this achievable without some sort of client-server coordination (which is what Datastore gives us)?

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @ptkach)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEqm1shCjpnnX6TXjOn:bupv4yz) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

Hmmm, I'm not sure if we would want to entertain the idea of changing the primary ID type for most of the entities... It seems like a risky migration that would undoubtedly require some downtime to perform. It certainly warrants some discussion to weight the pros and cons.

But still, as I mentioned earlier, I doubt an client-side allocation scheme would ever work, due to the chance of collision between multiple instances. We are not just talking about thread safety here, but rather distributed workers that all try to allocate globally unique IDs at the same time. This applies to Beam works, or just multiple Nomulus instances serving different clients concurrently. Is this achievable without some sort of client-server coordination (which is what Datastore gives us)?
</blockquote></details>

UUID in practice can run into bad states with lots of collisions. See https://github.com/ramsey/uuid/issues/80, at some point someone saw hundreds of collisions out of one million calls. And someone reported this as late as 09/2020.

ptkach:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)_ status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @ptkach)
> 
> _[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NEqm1shCjpnnX6TXjOn:bupv4yz) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):_
> 
> UUID in practice can run into bad states with lots of collisions. See [ramsey/uuid#80](https://github.com/ramsey/uuid/issues/80), at some point someone saw hundreds of collisions out of one million calls. And someone reported this as late as 09/2020.

UUID generator can not run into a collision in our case, it's just not possible. According to wiki - `50% probability of at least one collision can be achieved (after) ... generating 1 billion UUIDs per second for about 85 years.` The issue [mentioned](https://github.com/ramsey/uuid/issues/80) is just implementational bug for this particular library. Same issue can happen with datastore's allocateId implementation, postgres sequence nextVal or any other implementation

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java` line 178 at r1](https://reviewable.io/reviews/google/nomulus/1810#-NE6Nzb9A5RQifcU8dVw:-NFUq-xJ9HxLZF6_47p_:b-7lukw1) ([raw file](https://github.com/google/nomulus/blob/99abe3479c8d7e7a118b0c75888f8ce4a2e3d99d/core/src/main/java/google/registry/beam/spec11/Spec11Pipeline.java#L178)):*

UUID generator can not run into a collision in our case, it's just not possible. According to wiki - `50% probability of at least one collision can be achieved (after) ... generating 1 billion UUIDs per second for about 85 years.` The issue [mentioned](https://github.com/ramsey/uuid/issues/80) is just implementational bug for this particular library. Same issue can happen with datastore's allocateId implementation, postgres sequence nextVal or any other implementation 
</blockquote></details>

For the record, changing all IDs across our codebase from longs to Strings would be a lot of work that we simply don't want to take on right now.

weiminyu (approved):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1810)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach and @sarahcaseybot)

PR #1811, opened 2022-10-07
Update conditions when domain update flow triggers dns publish task

ptkach:
Addresses b/246375161

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1811)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
@@ -87,6 +87,7 @@ import google.registry.model.poll.PendingActionNotificationResponse.DomainPendin
 import google.registry.model.poll.PollMessage;
 import google.registry.model.reporting.IcannReportingTypes.ActivityReportField;
 import google.registry.model.tld.Registry;
+import java.util.Objects;
 import java.util.Optional;
 import javax.inject.Inject;
 import org.joda.time.DateTime;
@@ -181,7 +182,9 @@ public final class DomainUpdateFlow implements TransactionalFlow {
     DomainHistory domainHistory =
         historyBuilder.setType(DOMAIN_UPDATE).setDomain(newDomain).build();
     validateNewState(newDomain);
-    dnsQueue.addDomainRefreshTask(targetId);
+    if (requiresDnsUpdate(existingDomain, newDomain)) {
+      dnsQueue.addDomainRefreshTask(targetId);
+    }
     ImmutableSet.Builder<ImmutableObject> entitiesToSave = new ImmutableSet.Builder<>();
     entitiesToSave.add(newDomain, domainHistory);
     Optional<BillingEvent.OneTime> statusUpdateBillingEvent =
@@ -201,6 +204,16 @@ public final class DomainUpdateFlow implements TransactionalFlow {
                 .build());
     persistEntityChanges(entityChanges);
     return responseBuilder.build();
+  }
+
+  /** Determines if any of the changes to new domain should trigger DNS update. */
+  private boolean requiresDnsUpdate(Domain existingDomain, Domain newDomain) {
+    if (existingDomain.shouldPublishToDns() != newDomain.shouldPublishToDns()
+        || !Objects.equals(newDomain.getDsData(), existingDomain.getDsData())
+        || !Objects.equals(newDomain.getNsHosts(), existingDomain.getNsHosts())) {
+      return true;
+    }
+    return false;
   }
 
   /** Fail if the object doesn't exist or was deleted. */
@@ -267,12 +280,18 @@ public final class DomainUpdateFlow implements TransactionalFlow {
             .setLastEppUpdateRegistrarId(registrarId)
             .addStatusValues(add.getStatusValues())
             .removeStatusValues(remove.getStatusValues())
-            .addNameservers(add.getNameservers().stream().collect(toImmutableSet()))
-            .removeNameservers(remove.getNameservers().stream().collect(toImmutableSet()))
             .removeContacts(remove.getContacts())
             .addContacts(add.getContacts())
             .setRegistrant(firstNonNull(change.getRegistrant(), domain.getRegistrant()))
             .setAuthInfo(firstNonNull(change.getAuthInfo(), domain.getAuthInfo()));
+
+    if (!add.getNameservers().isEmpty()) {
+      domainBuilder.addNameservers(add.getNameservers().stream().collect(toImmutableSet()));
+    }
+    if (!remove.getNameservers().isEmpty()) {
+      domainBuilder.removeNameservers(remove.getNameservers().stream().collect(toImmutableSet()));
+    }
+
     Optional<DomainUpdateSuperuserExtension> superuserExt =
         eppInput.getSingleExtension(DomainUpdateSuperuserExtension.class);
     if (superuserExt.isPresent()) {
--- a/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java
+++ b/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java
@@ -47,6 +47,7 @@ import static google.registry.testing.DomainSubject.assertAboutDomains;
 import static google.registry.testing.EppExceptionSubject.assertAboutEppExceptions;
 import static google.registry.testing.HistoryEntrySubject.assertAboutHistoryEntries;
 import static google.registry.testing.TaskQueueHelper.assertDnsTasksEnqueued;
+import static google.registry.testing.TaskQueueHelper.assertNoDnsTasksEnqueued;
 import static google.registry.util.DateTimeUtils.START_OF_TIME;
 import static org.joda.money.CurrencyUnit.USD;
 import static org.junit.jupiter.api.Assertions.assertThrows;
@@ -496,13 +497,11 @@ class DomainUpdateFlowTest extends ResourceFlowTestCase<DomainUpdateFlow, Domain
             expectedDsData.stream()
                 .map(ds -> ds.cloneWithDomainRepoId(resource.getRepoId()))
                 .collect(toImmutableSet()));
-
-    // TODO: REENABLE AFTER PROPER FIX FOR DNS PUBLISHING TASKS IS FOUND
-    // if (dnsTaskEnqueued) {
-    //   assertDnsTasksEnqueued("example.tld");
-    // } else {
-    //   assertNoDnsTasksEnqueued();
-    // }
+    if (dnsTaskEnqueued) {
+      assertDnsTasksEnqueued("example.tld");
+    } else {
+      assertNoDnsTasksEnqueued();
+    }
   }
 
   @Test
@@ -1747,4 +1746,51 @@ class DomainUpdateFlowTest extends ResourceFlowTestCase<DomainUpdateFlow, Domain
     assertAboutDomains().that(reloadResourceByForeignKey()).hasNoAutorenewEndTime();
   }
 
+  @Test
+  void testAddDnsPublishStatus_enqueueDnsTask() throws Exception {
+    setEppInput(
+        "domain_update_status_change.xml",
+        ImmutableMap.of("STATUS_ADD", "clientHold", "STATUS_REM", "clientTransferProhibited"));
+    persistReferencedEntities();
+    persistResource(
+        persistDomain()
+            .asBuilder()
+            .setDomainName("example.tld")
+            .setStatusValues(ImmutableSet.of(StatusValue.CLIENT_TRANSFER_PROHIBITED))
+            .build());
+    runFlowAsSuperuser();
+    assertDnsTasksEnqueued("example.tld");
+  }
+
+  @Test
+  void testRemoveEveryDnsPublishStatus_enqueueDnsTask() throws Exception {
+    setEppInput(
+        "domain_update_status_change.xml",
+        ImmutableMap.of("STATUS_REM", "serverHold", "STATUS_ADD", "clientTransferProhibited"));
+    persistReferencedEntities();
+    persistResource(
+        persistDomain()
+            .asBuilder()
+            .setDomainName("example.tld")
+            .setStatusValues(ImmutableSet.of(StatusValue.SERVER_HOLD))
+            .build());
+    runFlowAsSuperuser();
+    assertDnsTasksEnqueued("example.tld");
+  }
+
+  @Test
+  void testChangeSomeOrNoChangeDnsPublishStatus_doNotEnqueueDnsTask() throws Exception {
+    setEppInput(
+        "domain_update_status_change.xml",
+        ImmutableMap.of("STATUS_ADD", "clientUpdateProhibited", "STATUS_REM", "pendingDelete"));
+    persistReferencedEntities();
+    persistResource(
+        persistDomain()
+            .asBuilder()
+            .setDomainName("example.tld")
+            .setStatusValues(ImmutableSet.of(StatusValue.PENDING_DELETE, StatusValue.SERVER_HOLD))
+            .build());
+    runFlowAsSuperuser();
+    assertNoDnsTasksEnqueued();
+  }
 }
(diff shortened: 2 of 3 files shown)


ptkach:
@CydeWeys 
I restored reverted part and added 2 tests to cover status change affect on dns publish task. 
DS data change affect on DNS publish task has already been covered under multiple tests written by @jianglai ,for them I've reenabled the dns publish task assertion. 
I've researched the code and I don't see any direct affect of ```OK``` or ```INACTIVE``` status on DNS writer, so per your recommendation I've excluded them from ```affectsDns``` group.

jianglai (changes requested):
Isn't it easier to just have a static field `DNS_PUBLISH_AFFECTING_STATUS` that is a set containing `CLIENT_HOLD` and `SERVER_HOLD`, comparing to adding a field to the enum itself? I think it is also more readable, when you check the presence of a status within a set that clearing means "any status in this set should affect if DNS is published".

Also, we should include the bug that this PR tries to address. The bug numbers themselves are OK to include on GitHub as buganizer is only accessible internally.

> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1811#-:-NE18XVtD2C2SLc5g_0m:betxd8s)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @ptkach)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 187 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1w8rS4xh9lZSwZ6Ph:-NE1w8rS4xh9lZSwZ6Pi:bein87j) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L187)):*
> ```Java
>         historyBuilder.setType(DOMAIN_UPDATE).setDomain(newDomain).build();
>     validateNewState(newDomain);
>     if (difference(newDomain.getStatusValues(), existingDomain.getStatusValues()).stream()
> ```

This logic is complicated enough to warrant its own function with proper javadoc explaining why we are checking each of the conditions. Or at least some in-place comments to the same effect.



_Code quote:_
```Java
    if (difference(newDomain.getStatusValues(), existingDomain.getStatusValues()).stream()
            .anyMatch(sv -> sv.affectsDns)
        || difference(existingDomain.getStatusValues(), newDomain.getStatusValues()).stream()
            .anyMatch(sv -> sv.affectsDns)
        || !Objects.equals(newDomain.getDsData(), existingDomain.getDsData())
        || !Objects.equals(newDomain.getNsHosts(), existingDomain.getNsHosts())) {
      dnsQueue.addDomainRefreshTask(targetId);
    }
```
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1749 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1wqRqB9fYav2Krk1a:-NE1wqRqB9fYav2Krk1b:b-8ocvab) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1749)):*
> ```Java
>   }
> 
>   @Test
> ```

Also needs test to confirm that if a domain is in `CLIENT_HOLD` or `SERVER_HOLD`, removing the status through a domain update would enqueue a DNS refresh task.



_Code quote:_
```Java
  @Test
  void testDnsTaskTriggeredByChangeInStatusAffectingDns() throws Exception {
    setEppInput("domain_update_status_change.xml", ImmutableMap.of("STATUS", "clientHold"));
    persistReferencedEntities();
    persistResource(
        persistDomain()
            .asBuilder()
            .setDomainName("example.tld")
            .setStatusValues(ImmutableSet.of(StatusValue.PENDING_DELETE, StatusValue.SERVER_HOLD))
            .build());
    runFlowAsSuperuser();
    assertDnsTasksEnqueued("example.tld");
  }
```
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1751 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1x8BW08P-urwvMx0Z:-NE1x8BW08P-urwvMx0_:b-p7enhv) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1751)):*
> ```Java
>   @Test
>   void testDnsTaskTriggeredByChangeInStatusAffectingDns() throws Exception {
>     setEppInput("domain_update_status_change.xml", ImmutableMap.of("STATUS", "clientHold"));
> ```

This test is correct, but isn't it redundant? If there is already a `SERVER_HOLD`, adding a `CLIENT_HOLD` shouldn't trigger a DNS refresh, right? As it stands now, it a task will be enqueued.

I think this is another case where using a set to represent the statuses that would affect DNS publishing makes the logic better. You can check if the existing domain contains such a status, and then check for the same thing on the new domain. If the two booleans differ, you know you need to enqueue a task.



_Code quote:_
```Java
    setEppInput("domain_update_status_change.xml", ImmutableMap.of("STATUS", "clientHold"));
    persistReferencedEntities();
    persistResource(
        persistDomain()
            .asBuilder()
            .setDomainName("example.tld")
            .setStatusValues(ImmutableSet.of(StatusValue.PENDING_DELETE, StatusValue.SERVER_HOLD))
            .build());
    runFlowAsSuperuser();
    assertDnsTasksEnqueued("example.tld");
```

ptkach (commented):
RE "static field...that is a set" - I updated it to a static method instead. Method seems to be more preferable to status field, as it encapsulates logic within StatusValue in case we ever decide to update it

RE "include bug" - Updated commit message

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1811#-:-NE2b3Uj0aL4QmmiLB6V:bn80qvp)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 187 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1w8rS4xh9lZSwZ6Ph:-NE2c0on67j9j02BKIxI:bnc3qcy) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L187)):*

This logic is complicated enough to warrant its own function with proper javadoc explaining why we are checking each of the conditions. Or at least some in-place comments to the same effect.
</blockquote></details>

Good point. Extracted as method and added a comment to it. Conditions seem to be self-explanatory. 
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1749 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1wqRqB9fYav2Krk1a:-NE2cHCg-zMVpZUksEFA:b-bw8e14) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1749)):*

Also needs test to confirm that if a domain is in `CLIENT_HOLD` or `SERVER_HOLD`, removing the status through a domain update would enqueue a DNS refresh task.
</blockquote></details>

Updated.
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1751 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1x8BW08P-urwvMx0Z:-NE2cM-97wOLFtpAMBVb:bhbjfzw) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1751)):*

This test is correct, but isn't it redundant? If there is already a `SERVER_HOLD`, adding a `CLIENT_HOLD` shouldn't trigger a DNS refresh, right? As it stands now, it a task will be enqueued.

I think this is another case where using a set to represent the statuses that would affect DNS publishing makes the logic better. You can check if the existing domain contains such a status, and then check for the same thing on the new domain. If the two booleans differ, you know you need to enqueue a task.
</blockquote></details>

I don't believe it's redundant, as per @cydeweys comment #9 in b/250731430 any change to dns affecting status should trigger an update.

jianglai (changes requested):
Remember to update the PR description as well, that's what's recorded once the PR is merged.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1811#-:-NE3zB7a0zEg5njn0YS3:b-v62i50)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @ptkach)

___
*[`core/src/main/java/google/registry/model/eppcommon/StatusValue.java` line 181 at r3](https://reviewable.io/reviews/google/nomulus/1811#-NE3zJk26tjY639FzH_k:-NE3zJk26tjY639FzH_l:b-lepdk8) ([raw file](https://github.com/google/nomulus/blob/0bb7702828d1e51d157d6520aca9b99dc396fbdd/core/src/main/java/google/registry/model/eppcommon/StatusValue.java#L181)):*
> ```Java
>   }
> 
>   public static boolean affectsDns(StatusValue sv) {
> ```

I think using a `static final` field of `ImmutableSet<StatusValue>` works fine there (and makes its intention clearer than the method does at first glance). I don't see how a method would make it more flexible than a field. If things every change (I doubt it will), we can just update the field.

Plus it has the added benefit of making the check more readable in my opinion, see my comment below in the test.



_Code quote:_
```Java
  public static boolean affectsDns(StatusValue sv) {
    return CLIENT_HOLD.equals(sv) || SERVER_HOLD.equals(sv);
  }
```
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1751 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1x8BW08P-urwvMx0Z:-NE3w3W1DuXHabSWnfp3:b-1lqsyo) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1751)):*

I don't believe it's redundant, as per @cydeweys comment #9 in b/250731430 any change to dns affecting status should trigger an update. 
</blockquote></details>

Looking at `CloudDnsWriter` [implementation](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/dns/writer/clouddns/CloudDnsWriter.java;l=129;drc=9e31047c3aa964aaea112b7f0e49f97e8f7bf119;bpv=0;bpt=1), as long as the domain is on one of the states that prohibit publication of DNS, the DNS writer would short-circuit and wipe out any records.

So what matters is the change from should-publish to should-not-publish, or vice versa, but not the change of individual status. In this specific example, if the domain is already in `CLIENT_HOLD`, we are already not publishing DNS records for it, adding a `SERVER_HOLD` should not make any difference.

That's why I suggested checking something like `Sets.intersection(DNS_PUBLISH_AFFECTING_STATUSES, newDomain.getStatusValues()) ^ Sets.intersection(DNS_PUBLISH_AFFECTING_STATUSES, existingDomain.getStatusValues())`
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1791 at r3](https://reviewable.io/reviews/google/nomulus/1811#-NE41BOs4MriKkD7rXTR:-NE41BOt7egnSOz2Mbs4:bww0r8o) ([raw file](https://github.com/google/nomulus/blob/0bb7702828d1e51d157d6520aca9b99dc396fbdd/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1791)):*
> ```Java
>             .asBuilder()
>             .setDomainName("example.tld")
>             .setStatusValues(ImmutableSet.of(StatusValue.PENDING_DELETE, StatusValue.SERVER_HOLD))
> ```

I believe we actually should enqueue a DNS update task if `PENDING_DELETE` is changed. It is one of the [statuses](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/domain/DomainBase.java;l=104;drc=9e31047c3aa964aaea112b7f0e49f97e8f7bf119;bpv=0;bpt=1?q=DomainBase&ss=nomulus%2Fnomulus) that prohibit publication of DNS, and it can only be added/removed in a domain update flow [by a super user](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java;l=146;drc=9e31047c3aa964aaea112b7f0e49f97e8f7bf119;bpv=0;bpt=1?q=domainupdateflow&ss=nomulus%2Fnomulus). So if for whatever reason a super user needs to go in and remove a pending delete on a domain, we should resume publication of the DNS. This means this status should be part of the `DNS_PUBLISH_AFFECTING_STATUSES`.

@cydeweys thoughts?



_Code quote:_
```Java
PENDING_DELETE
```
___
*[`docs/flows.md` line 785 at r3](https://reviewable.io/reviews/google/nomulus/1811#-NE3z4390YBh3BevlzN1:-NE3z4390YBh3BevlzN2:b-hviqet) ([raw file](https://github.com/google/nomulus/blob/9e31047c3aa964aaea112b7f0e49f97e8f7bf119/docs/flows.md#L785)):*
> ```Markdown
> *   2003
>     *   At least one of 'add' or 'rem' is required on a secDNS update.
>     *   Fees must be explicitly acknowledged when performing an operation which
> ```

What is going on with this change?



_Code quote:_
```Markdown
    *   Fees must be explicitly acknowledged when performing an operation which
        is not free.
```

jianglai (approved):
:laughing: I meant the description [here](https://github.com/google/nomulus/pull/1811), above the "This change is reviewable" button. Right now it's blank. Sorry if my phrasing was ambiguous.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1811#-:-NE8dGsDDSOeqyb4cnGi:b-oyl6fh)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @ptkach)

___
*[`core/src/main/java/google/registry/model/eppcommon/StatusValue.java` line 181 at r3](https://reviewable.io/reviews/google/nomulus/1811#-NE3zJk26tjY639FzH_k:-NE8_rC64Z8g8IZHKrVw:b-nshjvy) ([raw file](https://github.com/google/nomulus/blob/0bb7702828d1e51d157d6520aca9b99dc396fbdd/core/src/main/java/google/registry/model/eppcommon/StatusValue.java#L181)):*

Simplified, see next comment for details
</blockquote></details>

Seems like this file doesn't need change at all. You added an extra newline that should be removed.
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1751 at r1](https://reviewable.io/reviews/google/nomulus/1811#-NE1x8BW08P-urwvMx0Z:-NE8_z0R-4dDHKpqyh7m:brksjcy) ([raw file](https://github.com/google/nomulus/blob/92c2ca7b2922cd2737ff18b3ea552a59391ed893/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1751)):*

Based on your suggestion here and PENDING_DELETE add request in next comment, I believe we can simplify and get rid of the ```AFFECTING_STATUSES``` field at all, by relying solely on shouldPublishToDns. No code is always better :) 
</blockquote></details>

lol yes you are very much correct!
___
*[`core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java` line 1750 at r4](https://reviewable.io/reviews/google/nomulus/1811#-NE8cePyE9CbGqS1DUGi:-NE8cePyE9CbGqS1DUGj:b-z36rqo) ([raw file](https://github.com/google/nomulus/blob/867d618ce66ceb4f0a1b72e48e9ba82e9bcafbbc/core/src/test/java/google/registry/flows/domain/DomainUpdateFlowTest.java#L1750)):*
> ```Java
> 
>   @Test
>   void testAnyPublishProhibitedStatusAddedTriggersDnsPubTask() throws Exception {
> ```

Nit: These`CamelCase` names are getting a little to long to be unwieldy.

I'd suggest changing them to `testDoSomething_resultsInSomething`, like `testAddDnsPublishStatus_enqueueDnsTask`. Same for the ones below. 



_Code quote:_
```Java
AnyPublishProhibitedStatusAddedTriggersDnsPubTask
```

PR #1813, opened 2022-10-10
Refactor createSynthetic to be a command instead of a pipeline

gbrodman:
This lifts some code from the RegistryCli class to install the remote API on worker threads so they can use Datastore for ID allocation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1813)


jianglai (approved):
Given that the pipeline ran successfully on QA and I verified that the generated history entries agree with the latest state of domain resources (save for a few fields that we know should not be compared), I think we have pretty high confidence that it is correct and should work in production.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1813#-:-NE2X_RgA_fXJbDbAdeq:b-2i0xto)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 57 at r1](https://reviewable.io/reviews/google/nomulus/1813#-NE2T0KS5MRyagRsIQyh:-NE2T0KS5MRyagRsIQyi:b-9bxw3u) ([raw file](https://github.com/google/nomulus/blob/3a2887d3007d1d84c90a07c4e80968805829b2e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L57)):*
> ```Java
>  * domains where the most recent history object did not represent the state of the domain as it
>  * exists in the world. Because RDE loads only from DomainHistory objects, this means that RDE was
>  * producing wrong data. This pipeline mitigates that issue by creating synthetic history events for
> ```

I think you meant "that pipeline". It might be helpful to link to the code directly.



_Code quote:_
```Java
This pipeline
```
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/1813#-NE2TM1I8SrazloK8S0x:-NE2TM1I8SrazloK8S0y:bx9tcy2) ([raw file](https://github.com/google/nomulus/blob/3a2887d3007d1d84c90a07c4e80968805829b2e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L63)):*
> ```Java
>  * <p>This command only operates over domains where the most recent history object is synthetic,
>  * occurred after the bad pipeline's end time, and occurred before we've begun running this
>  * pipeline. That way, this should fix any bad dsData entries created before while remaining
> ```

Again, I think it should be "before we've begun running that pipeline".

It's a bit confusing because there are two "bad" pipelines. The first is the re-save pipeline, whereas the other is the pipeline that creates synthetic entries with self-allocated IDs.

Given that we are checking in this piece of code, we might as well write down, in javadoc, why we are doing this in a command instead of a pipeline. Otherwise it might not be obvious to future readers (ourselves involved) what we are fixing in the first place.

Or do we plan to eventually check in this PR? I'm not sure what the plan is. This seems to be throwaway. Maybe we check it in and remove it later once all the clean up is done? That way it will be part of the git history in case we ever need to revisit it in the future.



_Code quote:_
```Java
 * occurred after the bad pipeline's end time, and occurred before we've begun running this
 * pipeline. That way, this should fix any bad dsData entries created before while remaining
```
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 114 at r1](https://reviewable.io/reviews/google/nomulus/1813#-NE2VCjM3gElrCrHWkZo:-NE2VCjN2C0borXduvwh:b-tbeimk) ([raw file](https://github.com/google/nomulus/blob/3a2887d3007d1d84c90a07c4e80968805829b2e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L114)):*
> ```Java
>                       .setParameter("newSyntheticRoundStart", NEW_SYNTHETIC_ROUND_START)
>                       .getResultStream()
>                       .map(r -> VKey.create(Domain.class, r))
> ```

Might be more efficient if we kept the Repo Id as return value and create the key (`VKey.createSql(Domain.class, r)` should suffice) in the threads later. Creating an Sql only key also saves some extra calls to ofy to create the pesky ofy keys. Could save some time in creating and passing the `VKey` entities in a transaction.



_Code quote:_
```Java
r -> VKey.create(Domain.class, r)
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1813)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 57 at r1](https://reviewable.io/reviews/google/nomulus/1813#-NE2T0KS5MRyagRsIQyh:-NE2Zuo7BQA02YdGpOzi:b-hfvxkg) ([raw file](https://github.com/google/nomulus/blob/3a2887d3007d1d84c90a07c4e80968805829b2e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L57)):*

I think you meant "that pipeline". It might be helpful to link to the code directly.
</blockquote></details>

No, I meant this one. This one is the pipeline that creates synthetic history events to fix RDE. 
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/1813#-NE2TM1I8SrazloK8S0x:-NE2aNMX4v-JdRmc0swW:b-m7lmgk) ([raw file](https://github.com/google/nomulus/blob/3a2887d3007d1d84c90a07c4e80968805829b2e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L63)):*

Again, I think it should be "before we've begun running that pipeline".

It's a bit confusing because there are two "bad" pipelines. The first is the re-save pipeline, whereas the other is the pipeline that creates synthetic entries with self-allocated IDs.

Given that we are checking in this piece of code, we might as well write down, in javadoc, why we are doing this in a command instead of a pipeline. Otherwise it might not be obvious to future readers (ourselves involved) what we are fixing in the first place.

Or do we plan to eventually check in this PR? I'm not sure what the plan is. This seems to be throwaway. Maybe we check it in and remove it later once all the clean up is done? That way it will be part of the git history in case we ever need to revisit it in the future.
</blockquote></details>

Yeah I think we should check it in just in case. It's good to have it in the history. 

I rewrote the Javadoc to add more detail, hopefully this looks good. 
___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 114 at r1](https://reviewable.io/reviews/google/nomulus/1813#-NE2VCjM3gElrCrHWkZo:-NE2ZW4kDLxUPanpNeyq:b9ysczg) ([raw file](https://github.com/google/nomulus/blob/3a2887d3007d1d84c90a07c4e80968805829b2e0/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L114)):*

Might be more efficient if we kept the Repo Id as return value and create the key (`VKey.createSql(Domain.class, r)` should suffice) in the threads later. Creating an Sql only key also saves some extra calls to ofy to create the pesky ofy keys. Could save some time in creating and passing the `VKey` entities in a transaction.
</blockquote></details>

yep that should be all good

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1813)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 71 at r2](https://reviewable.io/reviews/google/nomulus/1813#-NE3tYIV4RBvUzDqMuVW:-NE3tYIV4RBvUzDqMuVX:bkgworh) ([raw file](https://github.com/google/nomulus/blob/7a52b887b6961f7267de6171489df89b4eca7a6e/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L71)):*
> ```Java
>  * <p>We further restrict the domains to domains whose latest history object is before October 4.
>  * This is an arbitrary date that is suitably far after the previous incorrect run of this synthetic
>  * history pipeline, with the purpose of making future runs of this command idempotent (in case the
> ```

It wouldn't be strictly idempotent, is it? The domain repo IDs returned from the query would state the same but every time you run the command it would still create a bunch of new history entries.



_Code quote:_
```Java
idempotent
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1813)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 71 at r2](https://reviewable.io/reviews/google/nomulus/1813#-NE3tYIV4RBvUzDqMuVW:-NE3u7VGDV0RuQ81gqJz:b-rog44h) ([raw file](https://github.com/google/nomulus/blob/7a52b887b6961f7267de6171489df89b4eca7a6e/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L71)):*

It wouldn't be strictly idempotent, is it? The domain repo IDs returned from the query would state the same but every time you run the command it would still create a bunch of new history entries.
</blockquote></details>

Nah it runs over fewer domains next time if you stop it partway through, because we only count domains where the most recent synthetic history was before October 4 (arbitrary date)

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1813)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java` line 71 at r2](https://reviewable.io/reviews/google/nomulus/1813#-NE3tYIV4RBvUzDqMuVW:-NE42XpIEv97AaGwl6Yb:b-a0ahs6) ([raw file](https://github.com/google/nomulus/blob/7a52b887b6961f7267de6171489df89b4eca7a6e/core/src/main/java/google/registry/tools/javascrap/CreateSyntheticDomainHistoriesCommand.java#L71)):*

Nah it runs over fewer domains next time if you stop it partway through, because we only count domains where the most recent synthetic history was before October 4 (arbitrary date) 
</blockquote></details>

Oh yes you are right.

PR #1714, opened 2022-07-26
Upgrade App Engine Standard to Java 17 w/ bundled APIs

CydeWeys:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1714)

PR #1816, opened 2022-10-11
Upgrade App Engine Standard to Java 17 w/ bundled APIs

CydeWeys:
Note that this doesn't yet upgrade our actual Gradle scripts to use a more recent of Java (that will happen separately); this solely affects the GAE instances.

I followed the instructions here:
https://cloud.google.com/appengine/docs/standard/java-gen2/services/access

And note that I removed threadsafe true from appengine's XML config because that doesn't do anything anymore and was just throwing errors (the new instances handle multiple requests in parallel by default, no configuration necessary).

This is our second try (see #1714 for the first try). Spinnaker wasn't ready for Java 17 yet, but now it is.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1816)


jianglai (approved):
Note that we still don't know for sure if Spinnaker now supports the `app-engine-apis` tag. The currently deployed Spinnaker version contains clouddrive [v5.78.0](https://github.com/spinnaker/clouddriver/releases/tag/v5.78.0), which includes https://github.com/spinnaker/clouddriver/pull/5761. We are hopeful that it would fix the issue with the bundled `gcloud` being too old to support the said flag.

We will try to verify after this PR is merged.

> Reviewed 35 of 35 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1816#-:-NE8RICVEpDg8PZGystw:b-1m0flv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1824, opened 2022-10-20
Check for entity nonexistence in SqlBatchWriter

gbrodman:
Passing in an already-existing instance is an antipattern because it can lead to race conditions where something else modified the object in between when the pipeline loaded it and when you're saving it. The Write action should only be writing new entities.

We cannot check IDs for the objects (some IDs are not autogenerated so they might exist already). We also cannot call `insert` on the objects because the underlying JPA `persist` call adds the input object to the persistence context, meaning that any modifications (e.g. updateTimestamp) are reflected in the input object. Beam doesn't allow modification of input objects, so this throws an exception

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1824)

### DIFF
--- a/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java
+++ b/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java
@@ -14,6 +14,7 @@
 
 package google.registry.beam.common;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.persistence.transaction.TransactionManagerFactory.jpaTm;
 import static org.apache.beam.sdk.values.TypeDescriptors.integers;
 
@@ -405,7 +406,13 @@ public final class RegistryJpaIO {
               .filter(Objects::nonNull)
               .collect(ImmutableList.toImmutableList());
       try {
-        jpaTm().transact(() -> jpaTm().putAll(entities));
+        jpaTm()
+            .transact(
+                () -> {
+                  // Don't modify existing objects as it could lead to race conditions
+                  entities.forEach(this::verifyObjectNonexistence);
+                  jpaTm().putAll(entities);
+                });
         counter.inc(entities.size());
       } catch (RuntimeException e) {
         processSingly(entities);
@@ -419,7 +426,13 @@ public final class RegistryJpaIO {
     private void processSingly(ImmutableList<Object> entities) {
       for (Object entity : entities) {
         try {
-          jpaTm().transact(() -> jpaTm().put(entity));
+          jpaTm()
+              .transact(
+                  () -> {
+                    // Don't modify existing objects as it could lead to race conditions
+                    verifyObjectNonexistence(entity);
+                    jpaTm().put(entity);
+                  });
           counter.inc();
         } catch (RuntimeException e) {
           throw new RuntimeException(toEntityKeyString(entity), e);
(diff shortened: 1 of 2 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1824)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/google/nomulus/1824#-NFAEE8SCXBJakPzReeq):*
Actually I think we should require that all IDs are already filled in, for reasons stated in https://b.corp.google.com/issues/247806191#comment4 



_Code quote:_
```
  We cannot check IDs for the objects (some IDs are not autogenerated so
  they might exist already). We also cannot call `insert` on the objects
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1824)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/google/nomulus/1824#-NFAEE8SCXBJakPzReeq):*

Actually I think we should require that all IDs are already filled in, for reasons stated in https://b.corp.google.com/issues/247806191#comment4 
</blockquote></details>

Fair point, that would match up with https://docs.google.com/document/d/1mFWC9zshHnAL3Cy-IoD3MKKylvWKN0bizZoE5i08rCg/edit# well. 

My initial thought is that that should wait for a separate PR because it'll require the implementation of the SQL-based ID allocation, right?

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1824)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/google/nomulus/1824#-NFAEE8SCXBJakPzReeq):*

Fair point, that would match up with https://docs.google.com/document/d/1mFWC9zshHnAL3Cy-IoD3MKKylvWKN0bizZoE5i08rCg/edit# well. 

My initial thought is that that should wait for a separate PR because it'll require the implementation of the SQL-based ID allocation, right? 
</blockquote></details>

If you want to check in a version that just checks for updates now, please add a todo for the id check. 
___
*[`core/src/main/java/google/registry/beam/common/RegistryJpaIO.java` line 411 at r1](https://reviewable.io/reviews/google/nomulus/1824#-NFJyRDm9fIBVgooDZHP:-NFJyRDm9fIBVgooDZHQ:b-f4d4hm) ([raw file](https://github.com/google/nomulus/blob/3579639ffbe3c1ebd30433b082b5840d72536b19/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L411)):*
Could you inline the relevant commit message? It's easier to find in the code. 

> We also cannot call insert on the objects because the underlying JPA persist call adds the input object to the persistence context, meaning that any modifications (e.g. updateTimestamp) are reflected in the input object. Beam doesn't allow modification of input objects, so this throws an exception

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1824)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`-- commits` line 10 at r1](https://reviewable.io/reviews/google/nomulus/1824#-NFAEE8SCXBJakPzReeq):*

If you want to check in a version that just checks for updates now, please add a todo for the id check. 
</blockquote></details>

Yeah I see the ID check as completely separate, because once we implement that design doc, every object will always have an ID anyway regardless of whether it's an existing object or a new one. 
___
*[`core/src/main/java/google/registry/beam/common/RegistryJpaIO.java` line 411 at r1](https://reviewable.io/reviews/google/nomulus/1824#-NFJyRDm9fIBVgooDZHP:-NFKHeDS4lHOQ27-Ad6-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3579639ffbe3c1ebd30433b082b5840d72536b19/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L411)):*

Could you inline the relevant commit message? It's easier to find in the code. 

> We also cannot call insert on the objects because the underlying JPA persist call adds the input object to the persistence context, meaning that any modifications (e.g. updateTimestamp) are reflected in the input object. Beam doesn't allow modification of input objects, so this throws an exception
</blockquote></details>

Done.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1824)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

PR #1825, opened 2022-10-21
Check token type of currentPackageToken

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1825)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -920,13 +920,17 @@ public class DomainBase extends EppResource
     }
 
     public B setCurrentPackageToken(@Nullable VKey<AllocationToken> currentPackageToken) {
-      if(currentPackageToken == null) {
+      if (currentPackageToken == null) {
+        getInstance().currentPackageToken = currentPackageToken;
         return thisCastToDerived();
       }
-      Optional<AllocationToken> token = tm().transact(() -> tm().loadByKeyIfPresent(currentPackageToken));
+      Optional<AllocationToken> token =
+          tm().transact(() -> tm().loadByKeyIfPresent(currentPackageToken));
       token.ifPresent(
-          allocationToken -> checkArgument(allocationToken.getTokenType().equals(TokenType.PACKAGE),
-              "The currentPackageToken must have a PACKAGE TokenType"));
+          allocationToken ->
+              checkArgument(
+                  allocationToken.getTokenType().equals(TokenType.PACKAGE),
+                  "The currentPackageToken must have a PACKAGE TokenType"));
       getInstance().currentPackageToken = currentPackageToken;
       return thisCastToDerived();
     }
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -55,6 +55,7 @@ import google.registry.model.domain.launch.LaunchNotice;
 import google.registry.model.domain.rgp.GracePeriodStatus;
 import google.registry.model.domain.secdns.DomainDsData;
 import google.registry.model.domain.token.AllocationToken;
+import google.registry.model.domain.token.AllocationToken.TokenType;
 import google.registry.model.eppcommon.StatusValue;
 import google.registry.model.host.Host;
 import google.registry.model.poll.PollMessage;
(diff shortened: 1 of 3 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1825)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/DomainBase.java` line 923 at r1](https://reviewable.io/reviews/google/nomulus/1825#-NF9PCR9CrSUM8la0V1z:-NF9PCR9CrSUM8la0V2-:b-nfiuh) ([raw file](https://github.com/google/nomulus/blob/3a32fdc188f04938ae7869e97931ff3001372834/core/src/main/java/google/registry/model/domain/DomainBase.java#L923)):*
> ```Java
> 
>     public B setCurrentPackageToken(@Nullable VKey<AllocationToken> currentPackageToken) {
>       if (currentPackageToken == null) {
> ```

do we have any tests on removal of package tokens? we might want to add one 
___
*[`core/src/main/java/google/registry/model/domain/DomainBase.java` line 928 at r1](https://reviewable.io/reviews/google/nomulus/1825#-NF9OxFF0gA6971mcr1i:-NF9OxFF0gA6971mcr1j:bhre0g2) ([raw file](https://github.com/google/nomulus/blob/3a32fdc188f04938ae7869e97931ff3001372834/core/src/main/java/google/registry/model/domain/DomainBase.java#L928)):*
> ```Java
>       }
>       Optional<AllocationToken> token =
>           tm().transact(() -> tm().loadByKeyIfPresent(currentPackageToken));
> ```

1. It's likely a good idea (I think) to not bother creating a new transaction here, since hopefully all callers of this will already be in a transaction 
2. Is there any situation in which the token should not be present in the DB? If so, we should probably test it.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1825)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/domain/DomainBase.java` line 923 at r1](https://reviewable.io/reviews/google/nomulus/1825#-NF9PCR9CrSUM8la0V1z:-NFANmK7BjRRw1KjqH-x:b-6977rz) ([raw file](https://github.com/google/nomulus/blob/3a32fdc188f04938ae7869e97931ff3001372834/core/src/main/java/google/registry/model/domain/DomainBase.java#L923)):*

do we have any tests on removal of package tokens? we might want to add one 
</blockquote></details>

added one
___
*[`core/src/main/java/google/registry/model/domain/DomainBase.java` line 928 at r1](https://reviewable.io/reviews/google/nomulus/1825#-NF9OxFF0gA6971mcr1i:-NFAOVjSE9CrgyBfGuT7:b-1ji7fg) ([raw file](https://github.com/google/nomulus/blob/3a32fdc188f04938ae7869e97931ff3001372834/core/src/main/java/google/registry/model/domain/DomainBase.java#L928)):*

1. It's likely a good idea (I think) to not bother creating a new transaction here, since hopefully all callers of this will already be in a transaction 
2. Is there any situation in which the token should not be present in the DB? If so, we should probably test it. 
</blockquote></details>

1.Wouldn't it be safer to include the transaction here? If the call site is already in a transaction, wouldn't this transaction be subsumed by that transaction? I don't think we can be certain that all callers will be in a transaction. 

2. The token should be in the database, fixed the code to throw if the token is not and added a test

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1825)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/DomainBase.java` line 928 at r1](https://reviewable.io/reviews/google/nomulus/1825#-NF9OxFF0gA6971mcr1i:-NFAgnX5-K7KgWX2mkmM:b-omwgk8) ([raw file](https://github.com/google/nomulus/blob/3a32fdc188f04938ae7869e97931ff3001372834/core/src/main/java/google/registry/model/domain/DomainBase.java#L928)):*

1.Wouldn't it be safer to include the transaction here? If the call site is already in a transaction, wouldn't this transaction be subsumed by that transaction? I don't think we can be certain that all callers will be in a transaction. 

2. The token should be in the database, fixed the code to throw if the token is not and added a test

</blockquote></details>

Yeah I guess it's probably safe since this is a check and we wouldn't anticipate tokens being deleted out from under us at any point where this might be called (can you tell i'm a bit paranoid about race conditions now?) 

2. cool thx

PR #1830, opened 2022-10-26
Add flyway files for allocateId sequence alternative

ptkach:
This PR implement DB part for allocateId SQL sequence alternative. Java part - https://github.com/google/nomulus/pull/1831

Tested in alpha

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1830)

PR #1831, opened 2022-10-26
Switch to SQL sequence based allocateId

ptkach:
This is a Java part of SQL sequence based allocateId swith. DB part - https://github.com/google/nomulus/pull/1830.

Tested in alpha with Contact update, Domain create EPP commands - observed entities created with the key generated with SQL Sequence

Tested locally by mapping all unit tests to SQL Sequence based id allocation - all tests succeeded 

Note: this PR is expected to fail as long as https://github.com/google/nomulus/pull/1830 not merged

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1831)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/model/IdService.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1831#-NFLJ8-XDPq9pzP8dQpM:-NFLJ8-YCUPuKR2mjcAC:b7a1idu) ([raw file](https://github.com/google/nomulus/blob/e8c60a5d5661ddaea320b2b3daddcf0277cd8db9/core/src/main/java/google/registry/model/IdService.java#L97)):*
> ```Java
>     @Override
>     public Long get() {
>       return tm().transact(
> ```

Since jpaTm is used in the lambda, we may as well use it here.



_Code quote:_
```Java
tm
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/model/IdService.java` line 97 at r1](https://reviewable.io/reviews/google/nomulus/1831#-NFLJ8-XDPq9pzP8dQpM:-NFP1VR32fjUcEvWpdqY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e8c60a5d5661ddaea320b2b3daddcf0277cd8db9/core/src/main/java/google/registry/model/IdService.java#L97)):*

Since jpaTm is used in the lambda, we may as well use it here.
</blockquote></details>

Done.

jianglai (commented):
I have not had time to review this, but just as a general comment, I still think we need a migration schedule with a read-only period, instead of simply switching to using a new ID allocation scheme like this.

The deployment of a new version does not happen atomically across all servers, so there would be a period in which both the old and the new server co-exist during deployment, during which both allocation scheme would work concurrently, based on which server is serving the request. That would lead to conflict (though unlikely due to the high floor set in the SQL sequence allocation scheme), but also invalidate the monotonicity of IDs allocated.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFTebQD3NPKqOw1cL6M:b-ngjsfq)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

ptkach (commented):
@jianglai Added a new migration state, not exactly sure why does it need to be readonly as you mentioned, so made it not readonly for now, because we're mainly concerned about consistency between instances, for which flag should suffice. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFj8hHf0KQdyMp7ZcmU:b-ge3ky0)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

jianglai (commented):
Note that just adding a migration state does not actually do anything. You will have to actually check the value and do something with it.

Reference PR: https://cs.opensource.google/nomulus/nomulus/+/bb5d2dcf0af1e5efbcee05b70b14e2cb854850a4

During a gradual rollout, the old version will use Ofy while the new version will use SQL to allocate IDs. Is there a way to not cause conflict without a read-only phase? For example, imagine ofy will allocate 100, 101, 102, ... whereas SQL will allocate 1000, 1001, 1002, .... The first request goes to an old instance, which gets allocated 100, the second goes to a new instance, which gets allocated 1000, then the third request goes to another old instance, which gets allocated 101. You then have 100, 1000, 102, which are not monotonically increasing.

What kind of flags do you have in mind? Flags usually roll out together with new instance, but we'd need something that is applied instantly to all running instances.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFjyEKH9j8B_BPXTAfT:b-2xmda)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

ptkach:
> Note that just adding a migration state does not actually do anything. You will have to actually check the value and do something with it.
> 
> Reference PR: https://cs.opensource.google/nomulus/nomulus/+/bb5d2dcf0af1e5efbcee05b70b14e2cb854850a4
> 
> During a gradual rollout, the old version will use Ofy while the new version will use SQL to allocate IDs. Is there a way to not cause conflict without a read-only phase? For example, imagine ofy will allocate 100, 101, 102, ... whereas SQL will allocate 1000, 1001, 1002, .... The first request goes to an old instance, which gets allocated 100, the second goes to a new instance, which gets allocated 1000, then the third request goes to another old instance, which gets allocated 101. You then have 100, 1000, 102, which are not monotonically increasing.
> 
> What kind of flags do you have in mind? Flags usually roll out together with new instance, but we'd need something that is applied instantly to all running instances.
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFjyEKH9j8B_BPXTAfT:b-2xmda)_ status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

@jianglai  I'm confused - I do actually check and based on that I switch between old vs way to allocateId. Can you please take another look? My change:

`DatabaseMigrationStateSchedule.getValueAtTime(jpaTm().getTransactionTime()).equals(MigrationState.SEQUENCE_BASED_ALLOCATE_ID)`

This new value I added to Enum is what I called a flag. I believe it can be changed from command line command after release is complete and that should switch all server instances at once. This is my understanding per the doc I read on this functionality. The reason for us having it configured with command line command is to allow manually switch between different states in case roll back required, etc.

jianglai (commented):
My bad, I didn't check the other file. You can even set the schedule ahead of deployment (because it is a schedule, it just says at the given time switch to the new phase).

Have you tested the new allocation mechanism in unit tests and beam? We don't have to use it in unit tests, but seeing that unit tests have access to test SQL database anyway, it is better to use the same allocation scheme than continuing to rely on self allocation.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFo-axE4fK94oNkLglW:b-dulwvb)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

ptkach:
> My bad, I didn't check the other file. You can even set the schedule ahead of deployment (because it is a schedule, it just says at the given time switch to the new phase).
> 
> Have you tested the new allocation mechanism in unit tests and beam? We don't have to use it in unit tests, but seeing that unit tests have access to test SQL database anyway, it is better to use the same allocation scheme than continuing to rely on self allocation.
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFo-axE4fK94oNkLglW:b-dulwvb)_ status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

@jianglai  I had tested it before by:

- running unit tests as I switched my local version to always provide allocateId based on sequence. As a result I saw that majority of them succeed, however a few fail as they in fact in one way or another have hardcoded expectation of a value provided with SelfIdAllocation. I don't think we should be fixing them now assuming it's blocking another PR
- running RdePipelineTest which succeeded as well
- tested on alpha for multiple e2e operations, like create domain, update contact and then checked ids stored in db. They matched sequence generated ids

Let me know if there's anything I need check as well. 

On a separate note, I've just now reverted the change made per your previous request to rename sequence to PascalCase. PascalCase doesn't work with [nativeQuery select](https://github.com/google/nomulus/blob/9db09c01ef36adffec03d00765f4c809c6d57803/core/src/main/java/google/registry/model/IdService.java#L108) as it transforms the input to be lowercase and then it fails to find corresponding sequence. I've updated both PR correspondingly.

jianglai (commented):
The thing with native query is weird. Have you tried to use double quotes around the table name, like [here](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/persistence/converter/TimedTransitionPropertyConverterBaseTest.java;l=93?q=nativequery&ss=nomulus%2Fnomulus)?

Regarding #1810, I don't think we are currently using the entries stored in SQL for anything. The spec11 report should still be generated from the json file stored in GCS. @sarahcaseybot what is the status of spec11 now, are we currently writing entries with duplicate IDs into SQL? If so we probably want to disable writing to SQL for now before the fix is rolled out.

I understand that you want to get this out of door sooner than later to unblock #1810. But realistically, it will be at least two weeks before that happens (see the [README](https://cs.opensource.google/nomulus/nomulus/+/master:db/) on schema changes). If #1830 gets merged this week, the schema change will be live after next week's release, and you can merge this PR to actually use the newly created sequence. This PR will then be deployed two weeks from now.

So I think there's enough time to get rid of self allocated IDs and fix the unit tests.

Lastly, when you ran `RdePipelineTest`, you turned off self-allocated ID, right? Otherwise it is not testing the new allocation mechanism.

It's probably worth it documenting what was tested in your PR description, using line like this:

TESTED=did this and that, verified something something.

This better surfaces the testing done than having to go though PR discussions.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NFprWvkCzn_TlReSC3n:b-1s1c1z)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/model/IdService.java` line 56 at r4](https://reviewable.io/reviews/google/nomulus/1831#-NFqOXNfEkobd7vSMos0:-NFqOXNfEkobd7vSMos1:b-ym05yo) ([raw file](https://github.com/google/nomulus/blob/9db09c01ef36adffec03d00765f4c809c6d57803/core/src/main/java/google/registry/model/IdService.java#L56)):*
> ```Java
>                   .equals(MigrationState.SEQUENCE_BASED_ALLOCATE_ID)
>               ? SequenceBasedIdSupplier.getInstance()
>               : DatastoreIdSupplier.getInstance());
> ```

The way this is written, the choice between DatastoreId and SqlBased Id is hardened at start time.

We want to be able to dynamically switch between these two by modifying the DatabaseMigrationSchedule.
This entire block should be the body of a supplier.



_Code quote:_
```Java
          : (DatabaseMigrationStateSchedule.getValueAtTime(jpaTm().getTransactionTime())
                  .equals(MigrationState.SEQUENCE_BASED_ALLOCATE_ID)
              ? SequenceBasedIdSupplier.getInstance()
              : DatastoreIdSupplier.getInstance());
```
___
*[`core/src/main/java/google/registry/model/IdService.java` line 92 at r4](https://reviewable.io/reviews/google/nomulus/1831#-NFqPS3z4Qzu2_P1bP7C:-NFqPS3z4Qzu2_P1bP7D:bjahclk) ([raw file](https://github.com/google/nomulus/blob/9db09c01ef36adffec03d00765f4c809c6d57803/core/src/main/java/google/registry/model/IdService.java#L92)):*
> ```Java
>    * <p>The generated IDs are project-wide unique
>    */
>   private static class SequenceBasedIdSupplier implements Supplier<Long> {
> ```

A class seems to be an overkill, especially when it is private.

Consider change the instance method get to a static method.



_Code quote:_
```Java
private static class SequenceBasedIdSupplier implements Supplier<Long> {
```

jianglai (changes requested):
You can also remove the `@OfyIdAllocation` annotation and related code:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/Buildable.java;l=66?q=ofyid&ss=nomulus%2Fnomulus

Lastly, I don't see the PR description mention testing the beam pipeline. If tested, please mention it.

> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NGNbvofDfEv1sUgtbXi:boym2ba)* status: all files reviewed, 5 unresolved discussions (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 79 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGMtEV6ChX90tE26SRX:-NGMtEV6ChX90tE26SRY:b-w1yt1d) ([raw file](https://github.com/google/nomulus/blob/e273a18b4a02d44cc6058b7172db8ffdc4ac3f24/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L79)):*
> ```Java
>     // allocated, and you don't care about the generated IDs as long
>     // as you can build the entities.
>     if (registryOptions.getUseSelfAllocatedId()) {
> ```

You can also delete the getter and setters in `RegistryPipelineOptions`.



_Code quote:_
```Java
if (registryOptions.getUseSelfAllocatedId()) {
```
___
*[`core/src/main/java/google/registry/model/IdService.java` line 92 at r4](https://reviewable.io/reviews/google/nomulus/1831#-NFqPS3z4Qzu2_P1bP7C:-NGMtptkERl-RLKwPH0w:b-jo24f4) ([raw file](https://github.com/google/nomulus/blob/9db09c01ef36adffec03d00765f4c809c6d57803/core/src/main/java/google/registry/model/IdService.java#L92)):*

nit: This falls into the unnecessary lambda pattern: go/bugpattern/UnnecessaryLambda
</blockquote></details>

It was a class to accommodate additional features like resetting the counter in tests. But I agree that now it can just be a supplier.

Better still, we should add a TODO to remove the supplier after we migrated. It can just be inlined in the `allocateId` function.

Also, the field name should be a camelCase.
___
*[`core/src/main/java/google/registry/model/IdService.java` line 59 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGNauxfD3Tl_Zkd3ncS:-NGNauxfD3Tl_Zkd3ncT:bawvj59) ([raw file](https://github.com/google/nomulus/blob/20f650855616bdd9e68df0854f08d684549644ac/core/src/main/java/google/registry/model/IdService.java#L59)):*
> ```Java
>    * <p>The generated IDs are project-wide unique
>    */
>   private static final Supplier<Long> DatastoreIdSupplier =
> ```

camelCase.



_Code quote:_
```Java
DatastoreIdSupplier
```
___
*[`core/src/main/java/google/registry/model/IdService.java` line 70 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGNahXaChwZM7dbz7CI:-NGNahXaChwZM7dbz7CJ:b-tpjzxb) ([raw file](https://github.com/google/nomulus/blob/20f650855616bdd9e68df0854f08d684549644ac/core/src/main/java/google/registry/model/IdService.java#L70)):*
> ```Java
> 
>   public static long allocateId() {
>     Supplier<Long> idSupplier = DatastoreIdSupplier;
> ```

Cloud be a ternary. 



_Code quote:_
```Java
    Supplier<Long> idSupplier = DatastoreIdSupplier;
    if (DatabaseMigrationStateSchedule.getValueAtTime(DateTime.now(UTC))
            .equals(MigrationState.SEQUENCE_BASED_ALLOCATE_ID)
        || RegistryEnvironment.UNITTEST.equals(RegistryEnvironment.get())) {
      idSupplier = SequenceBasedIdSupplier;
    }
```
___
*[`core/src/test/java/google/registry/model/ofy/OfyFilterTest.java` line 67 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGNbJDs4pBfg1y8kYN2:-NGNbJDs4pBfg1y8kYN3:b33glj7) ([raw file](https://github.com/google/nomulus/blob/20f650855616bdd9e68df0854f08d684549644ac/core/src/test/java/google/registry/model/ofy/OfyFilterTest.java#L67)):*
> ```Java
>   @RegisterExtension
>   final JpaIntegrationTestExtension database =
>       new JpaTestExtensions.Builder().withClock(new FakeClock()).buildIntegrationTestExtension();
> ```

Why is this extension needed here?

Also, if you don't plan to do anything with the lock, you can omit the `withClock()` part.

ptkach (commented):
> OfyIdAllocation
I don't think I can remove it. I mean the functionality should remain the same - if entity marked with annotation got no id, it's necessary to get one allocated. Me removing it will cause no allocation for such entities. Currently functionality is preserved while allocation is based on sql sequence.

> Lastly, I don't see the PR description mention testing the beam pipeline. If tested, please mention it.
I've mentioned that all tests succeeded, including the beam pipeline related tests obviously. Is there something else I needed to check? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NGSIB7QA3jBLLaYwz8o:be8a5ai)* status: 3 of 7 files reviewed, 4 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 79 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGMtEV6ChX90tE26SRX:-NGSX8hZDdsQNcYmrepa:b5fx966) ([raw file](https://github.com/google/nomulus/blob/e273a18b4a02d44cc6058b7172db8ffdc4ac3f24/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L79)):*
> getUseSelfAllocatedId
Done
___
*[`core/src/main/java/google/registry/model/IdService.java` line 92 at r4](https://reviewable.io/reviews/google/nomulus/1831#-NFqPS3z4Qzu2_P1bP7C:-NGSf7il3RP4eZk-LxoL:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/9db09c01ef36adffec03d00765f4c809c6d57803/core/src/main/java/google/registry/model/IdService.java#L92)):*

It was a class to accommodate additional features like resetting the counter in tests. But I agree that now it can just be a supplier.

Better still, we should add a TODO to remove the supplier after we migrated. It can just be inlined in the `allocateId` function.

Also, the field name should be a camelCase.
</blockquote></details>

Done
___
*[`core/src/main/java/google/registry/model/IdService.java` line 59 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGNauxfD3Tl_Zkd3ncS:-NGSf947CNvm-YFNifEJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/20f650855616bdd9e68df0854f08d684549644ac/core/src/main/java/google/registry/model/IdService.java#L59)):*

camelCase.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/IdService.java` line 70 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGNahXaChwZM7dbz7CI:-NGSi7U4D4rdJe3_B9Fl:b-896fix) ([raw file](https://github.com/google/nomulus/blob/20f650855616bdd9e68df0854f08d684549644ac/core/src/main/java/google/registry/model/IdService.java#L70)):*

Cloud be a ternary. 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/ofy/OfyFilterTest.java` line 67 at r5](https://reviewable.io/reviews/google/nomulus/1831#-NGNbJDs4pBfg1y8kYN2:-NGSis6L4Hyj2JMFXqLL:bpre9oy) ([raw file](https://github.com/google/nomulus/blob/20f650855616bdd9e68df0854f08d684549644ac/core/src/test/java/google/registry/model/ofy/OfyFilterTest.java#L67)):*

Why is this extension needed here?

Also, if you don't plan to do anything with the lock, you can omit the `withClock()` part.
</blockquote></details>

This test suite triggers id allocation, which is now transaction based, thus doesn't work without jpaextension (I tried to run without)

ptkach (commented):
Updated annotation name

Tested RDE pipeline in alpha and I can see succeeding job, one is failed, but I don't think that's related to my change.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831#-:-NGTm3v15H0LtfosXlnr:b-6qwi7h)* status: 7 of 13 files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai (approved):
> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1831)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/model/annotations/IdAllocation.java` line 24 at r7](https://reviewable.io/reviews/google/nomulus/1831#-NGYQfiJ4XfM0KfEKdT-:-NGYQfiJ4XfM0KfEKdT0:b-7pcpjx) ([raw file](https://github.com/google/nomulus/blob/04083e5669812d89cdaea78ca52fcb6915833a80/core/src/main/java/google/registry/model/annotations/IdAllocation.java#L24)):*
> ```Java
> 
> /**
>  * This annotation is only used for the field of a {@link google.registry.model.Buildable} class
> ```

The javadoc needs to be updated. We can say that this is needed for any ID field that needs to be allocated by `IdService`.



_Code quote:_
```Java
 * This annotation is only used for the field of a {@link google.registry.model.Buildable} class


 * that was previously annotated by both Ofy's and JPA's {@code @Id} annotations, of which the Ofy
 * annotation has been removed. The field still needs to be allocated automatically by the builder,
 * via the {@link IdService#allocateId()}.
 *
 * <p>It should be removed after we switch to using SQL to directly allocate IDs.
```

jianglai:
FYI, I just realized that we need to turn this on (i. e. make the switch to actually using the SQL-based allocation scheme) after the deployment, because the RDE pipeline runs daily, and it would no longer work without the self allocated ID provider (which is deleted in this PR).

CydeWeys:
If that's the case then we need to cancel tomorrow's release and rollback (or rewrite) this PR, as we aren't remotely ready to switch over to the new ID allocation scheme tomorrow, and we obviously can't afford to have RDE broken either.  Switching over ID allocation is a very high risk operation that we need to be incredibly careful about (especially seeing as how many other production issues we've had lately).  We're definitely not doing it on the 2nd day of GA for two newly launched TLDs.  The plan was to enable it and test it in sandbox for at least a week, and then switch over in production no sooner than after Thanksgiving.

This PR should not have included the removal of the existing ID allocation method used by RDE.  That change should only be committed once we have already switched over from using it in production, which definitely hasn't happened yet.

ptkach:
@CydeWeys  It was specifically requested here to remove SelfAllocate, so I've removed it. I guess the plan was to proceed with enabling it in prod immediately after release as we initially discussed MIgration State as only a precaution to keep all instances in sync. 
Anyhow, I restored SelfAllocateId which can be used for RDE here - https://github.com/google/nomulus/pull/1853

PR #1834, opened 2022-10-28
Remove bulk query entities

jianglai:
These alternative ORMs are introduced as a way to make querying large number of
domains and domain histories more efficient through bulk loading from several
to-be-joined tables separately, then in-memory re-assembly of the final entity,
bypassing the need to query multiple tables each time an entity is queried.

Their primary use case is loading these entities for comparison between
datastore and SQL during the migration, which has been completed. The
code remain unused as of now and their existence makes refactoring and
general maintenance more complicated than necessary due to the need to keep
them up to date.

Therefore we remove the related code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1834)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1834)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1834)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1838, opened 2022-11-03
Remove ofy from EppResource

jianglai:
The only remaining Ofy entity is `GaeUserConverter` after this PR. Therefore this PR also removed several classes and related tests that support the setup and verification of Ofy entities.

In addition, support for creating a `VKey` from a string is limited to `VKey<? extends EppResource>` only because it is the only use case (to pass a key to an EPP resource in a web safe way to facilitate resave), and we do not want to keep an extra simple name to class mapping, in addition to what `persistence.xml` contains. I looked into using `PersistenceXmlUtility` to obtain the mapping, but the xml file contains classes with the same simple name (namely `OneTime` from both `PollMessage` and `BillingEvent`). It doesn't seem like a worthwhile investment to write more code to deal with that, when the fact is that we only need to consider `EppResource`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1838)


jianglai (commented):
> Reviewed 105 of 105 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1838)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman and @weiminyu)

weiminyu (approved):
> Reviewed 66 of 105 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1838)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1661, opened 2022-06-09
Delete code relating to SQL init and scheduling

gbrodman:
One of the more significant changes introduced in this PR is that we use
SQL as the backing database in all tests unless otherwise specified,
e.g. by using the TmOverrideExtension. We change various ofy-related
tests to use this.

This includes various changes:
- Deletion of SqlEntity/DatastoreEntity and related classes. Includes
  any necessary changes because of that (e.g. getting a nice SQL key on
  error in RegistryJpaIO).
- Deletion of classes that used libraries from the init-sql code
  (RefreshDnsOnHostRenameAction)
- Removal of the JpaTransactionManager's backup implementation
- Modification of RegistryJpaWriteTest to not use init-sql code
- Removal of the Transaction class and related classes, however it does
  not remove the TransactionEntity class as that would require DB
  changes
- Removal of anything related to the actual usage of the database
  migration schedule or read-only phases
- Various test changes and fixes to account for the differences in SQL
  (like how foreign keys need to exist)

This deliberately doesn't do anything to alter the objects actually
stored in the DB yet, just how we use them

---
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1661)

PR #1845, opened 2022-11-08
Re-add RefreshDnsOnHostRenameAction

jianglai:
This class was accidentally deleted in #1661. This PR recreates it by
mostly re-adding its SQL-based code flow:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/test/java/google/registry/batch/RefreshDnsOnHostRenameActionTest.java;drc=9912e35ea297e969a428efdb1f8f01c86d794305;bpv=0;bpt=0

It does away with a pull queue due to incompatibility with Cloud Tasks.
Given what we have seen (about 700 tasks enqueued since May 2022), it
does not add much value in batching this operation anyway.

Also deleted AsyncTaskMetrics, which is not used any more. I don't think
we need to re-add metrics for this class either.


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1845)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/dns/DnsModule.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOAt5z1XRYDh5gUxbV:-NGOAt5z1XRYDh5gUxbW:bktvtil) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/DnsModule.java#L65)):*
> ```Java
>   @Provides
>   static HashFunction provideHashFunction() {
>     return Hashing.murmur3_32_fixed();
> ```

what's up with this change? I don't see this method in https://guava.dev/releases/21.0/api/docs/com/google/common/hash/Hashing.html, is it deprecated? 
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOFe7f9nm9JgrHt7VM:-NGOFe7f9nm9JgrHt7VN:b3y14ll) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L45)):*
> ```Java
>   public static final String QUEUE_HOST_RENAME = "async-host-rename";
>   public static final String PARAM_HOST_KEY = "hostKey";
>   public static final String PATH = "/_dr/task/refreshDnsOnHostRename";
> ```

the web.xml file refers to a different path, we need to either update that or update this to match so that requests get routed correctly 
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 67 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGODf1Y2Mxw2mNWf55x:-NGODf1Y2Mxw2mNWf55y:b-qf3cxd) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L67)):*
> ```Java
>   public void run() {
>     DateTime now = clock.nowUtc();
>     Host host = tm().transact(() -> tm().loadByKeyIfPresent(hostKey)).orElse(null);
> ```

we should just have one overarching transaction across everything rather than doing multiple transactions, just to reduce the chance of any race conditions (the previous action used a lock for the same effect) 

it also avoids any weirdness with the call to EppResourceUtils.isDeleted since we can just use the transaction time.

ditto EppResourceUtils.getLinkedDomainKeys
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 91 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOEr5Q8BqPYaIJA-l9:-NGOEr5Q8BqPYaIJA-lA:b7v0uba) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L91)):*
> ```Java
>   private void enqueueDomainsForRefresh(Host host) {
>     getLinkedDomainKeys(host.createVKey(), host.getUpdateTimestamp().getTimestamp(), null).stream()
>         .map(domainKey -> tm().transact(() -> tm().loadByKeyIfPresent(domainKey)).orElse(null))
> ```

when might this ever not be present?

jianglai (commented):
> Reviewed 15 of 15 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1845)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/dns/DnsModule.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOAt5z1XRYDh5gUxbV:-NGONsYw3iirML4bDtzQ:bwuv20u) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/DnsModule.java#L65)):*

what's up with this change? I don't see this method in https://guava.dev/releases/21.0/api/docs/com/google/common/hash/Hashing.html, is it deprecated? 
</blockquote></details>

It's deprecated in Guava 31 (which is the version IntelliJ resolved for me). The javadoc says:

```
   * @deprecated This implementation produces incorrect hash values from the {@link
   *     HashFunction#hashString} method if the string contains non-BMP characters. Use {@link
   *     #murmur3_32_fixed(int)} instead.
```
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOFe7f9nm9JgrHt7VM:-NGOOD2j-Vw_QIJzISME:bf07wjz) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L45)):*

the web.xml file refers to a different path, we need to either update that or update this to match so that requests get routed correctly 
</blockquote></details>

Does it? It seems to be the same:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/env/common/backend/WEB-INF/web.xml;l=256?q=web.xml&ss=nomulus%2Fnomulus

(The web.xml entry should actually be removed when the action was removed).



_Code quote:_
```Java
  public static final String PATH = "/_dr/task/refreshDnsOnHostRename";
```
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 67 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGODf1Y2Mxw2mNWf55x:-NGOQbOF6rvffJtiqir5:b-gjnwpa) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L67)):*

we should just have one overarching transaction across everything rather than doing multiple transactions, just to reduce the chance of any race conditions (the previous action used a lock for the same effect) 

it also avoids any weirdness with the call to EppResourceUtils.isDeleted since we can just use the transaction time.

ditto EppResourceUtils.getLinkedDomainKeys
</blockquote></details>

lol I get where you are coming from.

Though this is just to schedule a idempotent operation, for which transactionality is not crucial, but I agree that doing everything in one transaction is a good idea regardless. 
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 91 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOEr5Q8BqPYaIJA-l9:-NGOR5OE1dP_6DjeUJ3l:b-tweq5m) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L91)):*

when might this ever not be present? 
</blockquote></details>

There shouldn't be any cases. I guess I just mechanically transplanted this part of the old code without much thinking.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1845)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/dns/DnsModule.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOAt5z1XRYDh5gUxbV:-NGSC-6v-AptXgBFlTpA:b-vhysw6) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/DnsModule.java#L65)):*

It's deprecated in Guava 31 (which is the version IntelliJ resolved for me). The javadoc says:

```
   * @deprecated This implementation produces incorrect hash values from the {@link
   *     HashFunction#hashString} method if the string contains non-BMP characters. Use {@link
   *     #murmur3_32_fixed(int)} instead.
```
</blockquote></details>

ugh yeah this isthe worst, when I Googled "guava hashing javadoc" the first result was actually Guava 21.0 which doesn't have this new method at all, hence the confusion. Should be all good! 



_Code quote:_
```Java
Hashing.murmur3_32_fixed();
```
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOFe7f9nm9JgrHt7VM:-NGP0LB2D7LjbIXToOgg:b-jrw7mo) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L45)):*

Does it? It seems to be the same:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/env/common/backend/WEB-INF/web.xml;l=256?q=web.xml&ss=nomulus%2Fnomulus

(The web.xml entry should actually be removed when the action was removed).
</blockquote></details>

haha I misread it -- there are entries for both `dnsRefreshForHostRename` and `refreshDnsOnHostRename` somehow?!?!?

Can we remove the unused one?
___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 91 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOEr5Q8BqPYaIJA-l9:-NGP7xBb0Yc3VGLIv7Yv:beie83w) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L91)):*

There shouldn't be any cases. I guess I just mechanically transplanted this part of the old code without much thinking. 
</blockquote></details>

I think that makes sense yeah -- from what I see the old SQL code also used `loadByKey` so we should be good to go

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1845)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1845#-NGOFe7f9nm9JgrHt7VM:-NGTHf9T8mBy_7oU3BFa:b-j4au32) ([raw file](https://github.com/google/nomulus/blob/1beabded186ab304919da972628736b110418af4/core/src/main/java/google/registry/dns/RefreshDnsOnHostRenameAction.java#L45)):*

haha I misread it -- there are entries for both `dnsRefreshForHostRename` and `refreshDnsOnHostRename` somehow?!?!?

Can we remove the unused one?
</blockquote></details>

The other one was probably orphaned a looooong time ago...

Removed.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1845)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1845)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1850, opened 2022-11-11
Add defaultPromoTokens to Registry

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1850)


jianglai (changes requested):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 460 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGqrxZ96NO5rTTbwrYq:-NGqrxZ96NO5rTTbwrYr:b-59zjua) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L460)):*
> ```Java
>    * domain create.
>    */
>   @EmptySetToNull Set<VKey<AllocationToken>> defaultPromoTokens;
> ```

I wonder if the `@EmptySetToNull` annotation is still necessary? I was under the impression that we need this because some datastore-specific restrictions regarding saving empty sets (I could be wrong). But it would be nice if we don't have to consider this any more.

Also, similar to Gus's question in the other PR, how exactly do we plan to use the tokens when none is supplied during a creation? Do we apply all the default ones regardless of which registrar is creating the domain?



_Code quote:_
```Java
@EmptySetToNull
```
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 923 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGqtrRu9VomTuMn0xIJ:-NGqtrRu9VomTuMn0xIK:b-jr04r) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L923)):*
> ```Java
>                       token.getTokenType().equals(TokenType.DEFAULT_PROMO),
>                       String.format(
>                           "Token %s has an invalid token type. DefaultPromoTokens must be of the"
> ```

Why not show what the incompatible type is in the error message?



_Code quote:_
```Java
                          "Token %s has an invalid token type. DefaultPromoTokens must be of the"
                              + " type DEFAULT_PROMO",
```
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 928 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGquJ2bBiLu9QSWWAwH:-NGquJ2bBiLu9QSWWAwI:b-51z0d7) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L928)):*
> ```Java
>                   checkArgument(
>                       token.getAllowedTlds().contains(getInstance().tldStr),
>                       String.format("The token %s is not valid for this TLD", token.getToken()));
> ```

Same here, we can show what TLDs are allowed for this token.



_Code quote:_
```Java
String.format("The token %s is not valid for this TLD", token.getToken()))
```
___
*[`db/src/main/resources/sql/flyway.txt` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGqt0eKBtAhFX-jfKKc:-NGqt0eKBtAhFX-jfKKd:bbvh6qv) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/db/src/main/resources/sql/flyway.txt#L131)):*
> ```help
> V129__create_allocateId_sequence_alternative.sql
> V130__add_login_email_column_to_registrarpoc.sql
> V131__add_default_promo_tokens_to_registry.sql
> ```

These files are already in #1851. Should remove them from there.



_Code quote:_
```help
V131__add_default_promo_tokens_to_registry.sql
```

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: 6 of 11 files reviewed, 6 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 460 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGciGmn2LN4cDlChJs3:-NGciGmn2LN4cDlChJs4:b-n74n0e) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L460)):*
> ```Java
>    * domain create.
>    */
>   @EmptySetToNull Set<VKey<AllocationToken>> defaultPromoTokens;
> ```

I don't think we want this annotation here, since it's not actually true (empty set remains empty set, like with all the other StringConverters that we have) 
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 930 at r2](https://reviewable.io/reviews/google/nomulus/1850#-NHVay96EFbg4f13D2Pe:-NHVay9794W3GiVtmRiQ:b7gd08o) ([raw file](https://github.com/google/nomulus/blob/47e64e01c9627a58e294a1de9b3a9a6d9591144d/core/src/main/java/google/registry/model/tld/Registry.java#L930)):*
> ```Java
>                       String.format("The token %s is not valid for this TLD", token.getToken()));
>                 }
>                 getInstance().defaultPromoTokens = promoTokens;
> ```

As mentioned in the other PR, we'll want some verification that there aren't any overlapping allowed registrars (since we don't know what would happen if a registrar has two default tokens)

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: 1 of 11 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 460 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGciGmn2LN4cDlChJs3:-NH_dXQzDkovwgi9uAWf:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L460)):*

I don't think we want this annotation here, since it's not actually true (empty set remains empty set, like with all the other StringConverters that we have) 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 928 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGquJ2bBiLu9QSWWAwH:-NH_eznPB9q9KhFN9Sws:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L928)):*

Same here, we can show what TLDs are allowed for this token.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 930 at r2](https://reviewable.io/reviews/google/nomulus/1850#-NHVay96EFbg4f13D2Pe:-NI-4UAPCh9m_h86wohE:bhhbln7) ([raw file](https://github.com/google/nomulus/blob/47e64e01c9627a58e294a1de9b3a9a6d9591144d/core/src/main/java/google/registry/model/tld/Registry.java#L930)):*

As mentioned in the other PR, we'll want some verification that there aren't any overlapping allowed registrars (since we don't know what would happen if a registrar has two default tokens) 
</blockquote></details>

Instead decided to change this to a list since that will preserve an ordering of the tokens. That way, we can iterate through the default tokens on a TLD and apply the first one (if any) that is valid for a domain create. This is better since adding checks here would not have prevented modifications to the tokens after they are already saved to the TLD. 
___
*[`db/src/main/resources/sql/flyway.txt` line 131 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGqt0eKBtAhFX-jfKKc:-NHQmdygAGJQGHeQ1XbL:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/db/src/main/resources/sql/flyway.txt#L131)):*

These files are already in #1851. Should remove them from there.
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 2 of 11 files at r1, 5 of 7 files at r3, 3 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: 9 of 11 files reviewed, 5 unresolved discussions (waiting on @jianglai and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 930 at r2](https://reviewable.io/reviews/google/nomulus/1850#-NHVay96EFbg4f13D2Pe:-NI-G9G5DXJwWfYsJUsG:bgkjdap) ([raw file](https://github.com/google/nomulus/blob/47e64e01c9627a58e294a1de9b3a9a6d9591144d/core/src/main/java/google/registry/model/tld/Registry.java#L930)):*

Instead decided to change this to a list since that will preserve an ordering of the tokens. That way, we can iterate through the default tokens on a TLD and apply the first one (if any) that is valid for a domain create. This is better since adding checks here would not have prevented modifications to the tokens after they are already saved to the TLD. 
</blockquote></details>

I don't think it would have prevented modifications? Just using the first one doesn't seem like a clean / robust / clear solution to me. It might also not be possible or easy to modify the order if you wanted one or the other to have priority. 

My goal / thinking is to avoid any possibility of unexpected behavior, and having multiple possible default tokens for a given registrar+TLD combo seems like it could very easily lead to unexpected behavior. 

We'd just want to make sure there's never any possibility of registrar+tld combo overlap

jianglai (changes requested):
> Reviewed 1 of 7 files at r3, 3 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 460 at r1](https://reviewable.io/reviews/google/nomulus/1850#-NGqrxZ96NO5rTTbwrYq:-NI3l0oR4YbJEGEtJQ7e:b-m8cs3u) ([raw file](https://github.com/google/nomulus/blob/f3c15c1377b1c1aea30086b6cccf2b38f839f9b2/core/src/main/java/google/registry/model/tld/Registry.java#L460)):*

removed
</blockquote></details>

Have you tested if it impacts how null value/empty set is handled? i. e. how it impacts entity comparison and/or entity persistence/loading? If nothing is unexpected, we should make a backlog bug to remove this annotation.
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 930 at r2](https://reviewable.io/reviews/google/nomulus/1850#-NHVay96EFbg4f13D2Pe:-NI40YVG1asBjM-OulFQ:b-90sg7a) ([raw file](https://github.com/google/nomulus/blob/47e64e01c9627a58e294a1de9b3a9a6d9591144d/core/src/main/java/google/registry/model/tld/Registry.java#L930)):*

Since it is an ImmutableList, to change the ordering that we want the default tokens to be looked at in, we can save a new list to the TLD. There is no way to prevent registrar+TLD combo overlap, since if we add a check here, that won't prevent modifications to the actual token objects after they are saved to the TLD. By using a list, we can just use the first token in the list that works for that registrar. 
</blockquote></details>

I agree with Gus here, we should be very clear about what the behavior is regarding registrar/tld pair to default token mapping. My understanding is that only one default token should be applied for any domain creation. So it seems to make no sense for a situation where a registrar/tld pair could be mapped to multiple default tokens to exist.

Therefore, saying "always return the first one from the list" compared to "always return one from the set" is not solving the potential issue completely. Yes, you get reproducible results, which is better than what you can expect from a set, but it then depends on the order in which the list is constructed, which still leaves room for arbitrariness.

If the presumption that one registrar/tld should only map to a single default token is correct, we should enforce that in the builder, which is the only place to mutate the value. So what we should check in the setter is, for the set of tokens given, do their "allowedRegistrars" set intersect? 
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 931 at r5](https://reviewable.io/reviews/google/nomulus/1850#-NI3RHkU84oJ7sF8IlkP:-NI3RHkU84oJ7sF8IlkQ:bajdar8) ([raw file](https://github.com/google/nomulus/blob/bab82fe5276bb4456637f9e72126dbfaedf8a898/core/src/main/java/google/registry/model/tld/Registry.java#L931)):*
> ```Java
>                       token.getAllowedTlds().contains(getInstance().tldStr),
>                       String.format(
>                           "The token %s is not valid for this TLD. The valid TLDs for %s are %s",
> ```

No need to display the token twice. It could just be "The valid TLDs for it are %s".

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 931 at r5](https://reviewable.io/reviews/google/nomulus/1850#-NI3RHkU84oJ7sF8IlkP:-NI4IA2X2xsNE6PXpwol:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bab82fe5276bb4456637f9e72126dbfaedf8a898/core/src/main/java/google/registry/model/tld/Registry.java#L931)):*

No need to display the token twice. It could just be "The valid TLDs for it are %s".
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: 9 of 11 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 930 at r2](https://reviewable.io/reviews/google/nomulus/1850#-NHVay96EFbg4f13D2Pe:-NI4KGtS-xauGHHHGUpR:bz128lr) ([raw file](https://github.com/google/nomulus/blob/47e64e01c9627a58e294a1de9b3a9a6d9591144d/core/src/main/java/google/registry/model/tld/Registry.java#L930)):*

Checking in this setter that the allowedRegistrars set don't intersect will still not prevent changes being made to the allowed registrars on one of those tokens after these default tokens are saved to the TLD. Either way, I'm not sure we want to enforce that there can only be one tld/registrar pairing. We may potentially in the future want some sort of list such as {30% off registrar1, 20% off registrar2, 10% off all registrars} where the final token in the list would apply to all creations if no token in the list before that one can be applied.
</blockquote></details>

I think the point that Gus and Lai are missing is that there are a large number of different validity checks that go into whether a particular token is applicable to a particular create.  Some examples include: Whether the domain is premium, how long the initial registration term is for, if the registrar has access to that token, and if the token's validity schedule says that it is active at the current time.  Given all these different possible factors, it is not possible to somehow limit the potentially active tokens to a single one per registrar.  The best we can do is provide a prioritization through ordering, such that the listed tokens are tried in order and the first applicable one (if any) is used.

There are plenty of use cases for having multiple tokens for a single registrar, such as: There is a global ongoing promotion that dozens of registrars are enrolled in and then also a specific promotion for a certain event for a certain registrar (in which case the more specific one would be listed first and take priority), or there are multiple promotions scheduled ahead of time with different validity periods, or there are two tokens that are live simultaneously which provide a different discount on premium vs non-premium domains, etc.

I think what we need to be careful of is, in the `nomulus [create|update]_tld` commands, ensure that the parameter for setting this list of default allocation tokens makes it clear that ordering is important, and only allow the list to be set by providing its entire desired contents in order, e.g. it doesn't make sense to have an `--add_default_tokens` parameter (whereas the `--add_reserved_lists` parameter does make sense).

gbrodman (approved):
> Reviewed 1 of 1 files at r5, 1 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1850)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 930 at r2](https://reviewable.io/reviews/google/nomulus/1850#-NHVay96EFbg4f13D2Pe:-NIJoEkf1G3_f0a1WMNh:b-96173) ([raw file](https://github.com/google/nomulus/blob/47e64e01c9627a58e294a1de9b3a9a6d9591144d/core/src/main/java/google/registry/model/tld/Registry.java#L930)):*

Added a note to the field declaration. As far as tracking the history of this field, maybe we could implement something like a change notification email that is sent by the update TLD command. 

Either way, I believe it is a current task on our backlog to implement some sort of source control for TLD modifications, this is another reason why that would be useful.
</blockquote></details>

Yeah, source control / history is definitely out of the scope of this PR. This makes sense to me, lgtm

PR #1851, opened 2022-11-11
Flyway files for defaultPromoTokens

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1851)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1851)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (changes requested):
actually hold up

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1851#-:-NGcib1j04DC463xVLHV:bdf5hy4)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (changes requested):
so if the structure is "when a registrar tries to register a domain D on some TLD T, we check to see if T has a default token to apply for that registrar". This seems to me to imply that there should be precisely zero or one default tokens for a given registrar+tld pair 

in which case we'd probably want like, a Map<Registrar, token key> rather than a simple set of token keys, right? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1851#-:-NGcid2d4Qx9ZrEKpMa4:b768901)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

CydeWeys (commented):
The tokens already have a list of allowed registrars on them, so why would we want to duplicate that here too?  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/domain/token/AllocationToken.java;l=166-169;drc=a0f177b71f9e33f268ff84873eda9efff2eb3a08

It's hard to even say what the correct behavior would be if the two were to come into conflict.

And we're also likely to have a lot of registrars participating in the same promo, so adding the same value to the map for potentially dozens of keys seems annoying.  Better to keep that information centralized on the token itself, rather than needing to have two separate places to update.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1851#-:-NH6O5TV4v_4VSzEK5JA:bh2jk07)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (approved):
yeah fair enough, I see y'all's point. 

The annoying bit, which doesn't have to be in this PR, is that we'll need to make sure that there are never any overlapping registrars on default tokens for a given TLD. Like, when adding tokens or changing tokens' allowed registrar IDs, we'll need to check that. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1851#-:-NHVIz7C4YoQ1Pn7joHY:b9fj1fk)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1852, opened 2022-11-14
Use login email instead of GAE user ID for RegistrarPoc

jianglai:
Switch to using the login email address instead of GAE user ID to
  identify console users. The primary use cases are:

  1) When the user logged in the registrar console, need to figure out
     which registrars they have access to (in
     AuthenticatedReigstrarAccess).

  2) When a user tries to apply a registry lock, needs to know if they
     can (in RegistryLockGetAction).

  Both cases are tested in alpha with a personal email address to ensure
  it does not get the permission due to being a GAE admin account.

  Also verified that the soy templates includes the hidden login email
  form field instead of GAE user ID when registrars are displayed on the
  console; and consequently when a contact update is posted to the server,
  the login email is part of the JSON payload. Even though it does not
  look like it is used in any way by RegistrarSettingsAction, which
  receives the POST request. Like GAE user ID, the field is hidden, so
  cannot be changed by the user from the console, it is also not used to
  identify the RegistryPoc entity, whose composite keys are the contact
  email and the registrar ID associated with it.

  The login email address is backfilled for all RegistrarPocs that have a
  non-null GAE user ID. The backfilled addresses converted to the same ID
  as stored in the database.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1852)


jianglai (commented):
> Reviewed 33 of 33 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

ptkach (changes requested):
> Reviewed 28 of 33 files at r1, 3 of 4 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV1Ymv4aNckoRQBdXU:-NHV1Ymv4aNckoRQBdXV:b-bcrz2u) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L320)):*
> ```Java
>                       .query(
>                           "SELECT r FROM Registrar r INNER JOIN RegistrarPoc rp ON r.registrarId ="
>                               + " rp.registrarId WHERE lower(rp.loginEmailAddress) = :email AND"
> ```

what about previously created records, before this change loginEmailAddress introduces a new column? I suppose they will be missed
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 127 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV7FkgBm5WjFBaZoWv:-NHV7FkgBm5WjFBaZoWw:b-4gnlfe) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L127)):*
> ```Java
>     ImmutableList<RegistrarPoc> matchingContacts =
>         registrar.getContacts().stream()
>             .filter(contact -> contact.getLoginEmailAddress() != null)
> ```

Since we're not backfilling this is going to filter out all records created before this change, won't it? 



_Code quote:_
```Java
.filter(contact -> contact.getLoginEmailAddress() != null)
```
___
*[`core/src/main/resources/google/registry/ui/soy/registrar/ContactSettings.soy` line 186 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV3UkeDRsinaYIq3-J:-NHV3UkeDRsinaYIq3-K:b-7djp5q) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/resources/google/registry/ui/soy/registrar/ContactSettings.soy#L186)):*
> ```soy
>       {/if}
>     </table>
>     {if isNonnull($item['loginEmailAddress'])}
> ```

I presume it should work as it's just a rename, but I'm not sure how - for input hidden to be submitted with an action it should be a part of a form and I don't see any enclosing form tag for it, probably it's located somewhere else.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV1Ymv4aNckoRQBdXU:-NHVDU8-0nw2753KH5h3:b4ovjxv) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L320)):*

what about previously created records, before this change loginEmailAddress introduces a new column? I suppose they will be missed
</blockquote></details>

I'm in the process of backfilling the `login_email_address` column. The schema just went in this Monday on production.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java` line 127 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV7FkgBm5WjFBaZoWv:-NHVDfev21nn0BUyL6vh:b2w9zl3) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/ui/server/registrar/RegistryLockGetAction.java#L127)):*

Since we're not backfilling this is going to filter out all records created before this change, won't it? 
</blockquote></details>

Same as above, we will be backfilling this column.
___
*[`core/src/main/resources/google/registry/ui/soy/registrar/ContactSettings.soy` line 186 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV3UkeDRsinaYIq3-J:-NHVElal23sJt7ExK4kj:b-y8570b) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/resources/google/registry/ui/soy/registrar/ContactSettings.soy#L186)):*

I presume it should work as it's just a rename, but I'm not sure how - for input hidden to be submitted with an action it should be a part of a form and I don't see any enclosing form tag for it, probably it's located somewhere else.
</blockquote></details>

From what I understand, when you go to this fragment (`/registrar#contact-settings/someone@somedomain.com`), an XHR request is POSTed to `RegistrarSettingsActions` (with argument `op: "read"`), which returns a JSON presentation of the registrar, which contains the list of registrar contacts (`RegistrarPoc`).

The JSON response is then used by soy to generate the HTML, with the form pre-populated with the relevant registrar contact's current info when you click "Edit". This is there the hidden form field goes. When you click on "Save", another XHR is POSTed to the same action (with argument `op: "update"`), including whatever information is in the form.

I verified that hidden tag does exist on that page with developer tools. I think it's hidden because we do not want it changed by the user. We use the JSON payload in an update request to construct `RegistrarPoc` objects, which is then saved to the database, so the JSON request needs to contain all the information required to construct a `RegistrarPoc`, including the login email address, or the gae user id previously:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java;l=274;drc=b1ee1e84418cd0c1da6e63eb123be2ae094da121;bpv=0;bpt=1

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV1Ymv4aNckoRQBdXU:-NHVNnOhCa8kOM2HCL0I:b-beruh9) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L320)):*

I'm in the process of backfilling the `login_email_address` column. The schema just went in this Monday on production.
</blockquote></details>

how are we backfilling this? I don't see a PR to that effect yet, but I might just be missing it.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV1Ymv4aNckoRQBdXU:-NHVnl8qCDZXGKOTM2Jt:brbqunz) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L320)):*

how are we backfilling this? I don't see a PR to that effect yet, but I might just be missing it. 
</blockquote></details>

Mostly just picking the contact email address, trying to convert it to a GAE user Id, and check if it matches the one stored in the database. If yes, copy this email over to the login email address field. If not, the best guess is registry-id@registry.google, which we then test against the stored GAE user id.

If both failed, we'll have to log into the dasher domain admin console and look for the user account, but from what I can see on alpha and sandbox, those two guesses covered all cases other than one, which seems is for an registrar that genuinely doesn't have a dasher account set up in OT&E environment...

The code that I am using is here:

https://github.com/jianglai/nomulus/blob/b2af6f650a941229e02c690ce5ff686718ae53f6/core/src/main/java/google/registry/tools/javascrap/BackfillRegistrarPocLoginEmailCommand.java

I don't think it is necessary to check in this throwaway code, but if you think it is, I can certainly open a PR.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV1Ymv4aNckoRQBdXU:-NHVwVnL7yizG0UQkvy4:bnhb2c7) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L320)):*

Mostly just picking the contact email address, trying to convert it to a GAE user Id, and check if it matches the one stored in the database. If yes, copy this email over to the login email address field. If not, the best guess is registry-id@registry.google, which we then test against the stored GAE user id.

If both failed, we'll have to log into the dasher domain admin console and look for the user account, but from what I can see on alpha and sandbox, those two guesses covered all cases other than one, which seems is for an registrar that genuinely doesn't have a dasher account set up in OT&E environment...

The code that I am using is here:

https://github.com/jianglai/nomulus/blob/b2af6f650a941229e02c690ce5ff686718ae53f6/core/src/main/java/google/registry/tools/javascrap/BackfillRegistrarPocLoginEmailCommand.java

I don't think it is necessary to check in this throwaway code, but if you think it is, I can certainly open a PR.
</blockquote></details>

I don't know necessarily that we need to check it in, but we should definitely be documenting somewhere what email addresses we assign / guess. Just in case things go wrong in the future I suppose?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV1Ymv4aNckoRQBdXU:-NHVx1vv3UpWpONXT6GS:btr9ywm) ([raw file](https://github.com/google/nomulus/blob/7fda77f6f3d99434e96df9c9c7dd05fbae35799b/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L320)):*

I don't know necessarily that we need to check it in, but we should definitely be documenting somewhere what email addresses we assign / guess. Just in case things go wrong in the future I suppose? 
</blockquote></details>

Yes, once production backfill is done, I think we will have a better understanding of what caveats we want to document.

jianglai (commented):
Backfill is complete.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852#-:-NI4GzQ59jfdULgaWNvk:bg0j6kw)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java` line 323 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NI3Gta16aoigl552yR5:-NI4BQY21JRkCaSYq1A7:b9mazoj) ([raw file](https://github.com/google/nomulus/blob/5114fb3b1d817dea75287abc2a07c01cef13cf58/core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java#L323)):*

out of curiosity, why use Ascii.toLowerCase rather than String's built-in lowercase method? 
</blockquote></details>

To be honest, I just searched for "toLowerCase" in our code base and used the first one I found... I don't think there's a difference and we do use both in the code base. I believe with String's built-in method you can supply an encoding, but ASCII should suffice for email addresses. 
___
*[`core/src/main/java/google/registry/tools/RegistrarPocCommand.java` line 183 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NI3p9Hm1Rz2pLyQx6El:-NI4CDY14ABef8mV83TM:b-oxgb06) ([raw file](https://github.com/google/nomulus/blob/5114fb3b1d817dea75287abc2a07c01cef13cf58/core/src/main/java/google/registry/tools/RegistrarPocCommand.java#L183)):*

nit, but the formatting on this comment got messed up 
</blockquote></details>

Good find!
___
*[`core/src/main/resources/google/registry/ui/soy/registrar/ContactSettings.soy` line 186 at r3](https://reviewable.io/reviews/google/nomulus/1852#-NHV3UkeDRsinaYIq3-J:-NI4GLtOD3apXWPubF87:bzbpuo3) ([raw file](https://github.com/google/nomulus/blob/5114fb3b1d817dea75287abc2a07c01cef13cf58/core/src/main/resources/google/registry/ui/soy/registrar/ContactSettings.soy#L186)):*

yeah this is correct, basically the existing console bases everything around the "Registrar" object which contains pretty much everything related to the registrar, including contacts. 
</blockquote></details>

Thanks for confirming.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1852)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

PR #1856, opened 2022-11-22
Remove sharding parameter from RegistryJpaIO

jianglai:
This parameter is misleading and does not do what it purports to do.
Namely, it does not impact the level of parallelism. Given the input n for this
parameter, and m for the batch size, the elements are divided (keyed) into n
groups, each of which are then spread evenly across all threads, which
are eventually in turn batched into batches with size m:

https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L227

This is also evident in the implementation itself, where the ShardedKey
is determined by the unique number for a worker/thread combo and the
original key:

https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L268

Using a more concrete example, suppose we have 100 elements and 10
worker threads, with a target batch size of 5. If the "shard" number is set to
1, we first spread the 100 elements across 10 threads, resulting in 10
elements per thread, each thread then batches the elements into 2
batches of size 5.

If the "shard" number is set to 2, the 100 elements are first divided into 2
"shards" of 50 each. Each "shard" is then distributed within the 10
threads, resulting in 5 elements per "shard" per thread. They then get
turned into 1 batch per "shard" per thread. In the end, each thread
still processes 2 batches, even though they are from 2 different "shards".

Therefore this "shard" number does not perform horizontal partitioning
that one normally associates with sharding, and provides no
performance benefits but rather confuses the user.

It is also suggested that using withShardedKey() alone is already
sufficient to achieve auto-sharding within the keyed group. There is no
need to manually divide the input by keying them differently based on
the "shard" number specified:

https://youtu.be/jses0W4Zalc?t=967


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1856)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1856)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/beam/common/RegistryJpaIO.java` line 293 at r1](https://reviewable.io/reviews/google/nomulus/1856#-NHzHIE-3Swvq7Xp9Eur:-NHzHIE04ZFqshHoJjn5:b-zabtmd) ([raw file](https://github.com/google/nomulus/blob/279181ac4b7dba9d9631624a7915ac6128b0c22d/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L293)):*
> ```Java
>     public PCollection<Void> expand(PCollection<T> input) {
>       return input
>           .apply("Add key to data " + name(), WithKeys.<Integer, T>of(0).withKeyType(integers()))
> ```

Have you measured the performance using a single key?

I'm under the impression that the number of unique keys affects the parallelism in grouping-related transforms.  Using your example of 100 items, 2 shards, and 10 workers, my understanding is that the input may not spread to 10 workers.



_Code quote:_
```Java
0
```

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1856)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/beam/common/RegistryJpaIO.java` line 293 at r1](https://reviewable.io/reviews/google/nomulus/1856#-NHzHIE-3Swvq7Xp9Eur:-NI80yF47V07YJy49wFW:ba188mx) ([raw file](https://github.com/google/nomulus/blob/279181ac4b7dba9d9631624a7915ac6128b0c22d/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L293)):*

Have you measured the performance using a single key?

I'm under the impression that the number of unique keys affects the parallelism in grouping-related transforms.  Using your example of 100 items, 2 shards, and 10 workers, my understanding is that the input may not spread to 10 workers.

</blockquote></details>

I can do some comparison as part of testing the expand billing events pipeline.

When you say the input may not spread to 10 workers, do you mean using more shards will make it spread better? Like using 10 shards instead of 2? I would like to emphasize again that the "shards" we are talking about are not actually shards in beam parlance. When we set this value to n, we are just evenly distributing the input into different groups, keyed by integers between 1 to n. The KV is then sharded by beam when you use `withShardedKey()`, which than computes a new KV, whose key is based on the old key and the total number of worker thread. The Javadoc clearly stated that all elements that share the same old key will be evenly distributed across all threads:

https://github.com/apache/beam/blob/4ab557bf98784e51b6f8dca7d89ba5bcea9f04ef/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L227

Further more, if you look at how the new key is computed, it takes the old key, and a value determined by the worker ID and local thread ID (therefore making it globally unique among all worker threads). That's how it is able to distribute elements with the same (old) key evenly across all threads, which is then batched into groups:

https://github.com/apache/beam/blob/4ab557bf98784e51b6f8dca7d89ba5bcea9f04ef/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L264

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1856)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/beam/common/RegistryJpaIO.java` line 293 at r1](https://reviewable.io/reviews/google/nomulus/1856#-NHzHIE-3Swvq7Xp9Eur:-NJ2nLUG3tJgIim3wDLD:b-3o1exe) ([raw file](https://github.com/google/nomulus/blob/279181ac4b7dba9d9631624a7915ac6128b0c22d/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L293)):*

There is some 'magic' in BEAM that I don't fully understand, but I believe the pipeline scheduler treats PCollection<KV> differently than PCollection of non-KV entities, doing hidden logic according to the key values. Flume seems to work this way too.

E.g., if we look at the code of GroupByKey, there is almost nothing, but eventually all elements with the same key are sent to the same worker.
The GroupIntoBatches also groups entities by key, we should verify if it is affected by the same logic, which could limit parallelism.

Coming back to `GroupIntoBatches.withSharededKey()`, and with all input keys being the same,  the number of unique output keys is limited to the number of threads employed by the phase that runs the `Map.via` transformation. 

I think we should get some measurements first. 







</blockquote></details>

So I did some testing in #1881. This is done within unit tests so it employs only the direct runner, but I believe we should expect the axiom to hold that the parallelism behavior is not determined by the type of runner used.

I started with two elements `T`, which are then keyed to the same dummy key `Integer`. The resulting `KV<Integer, T>` are then grouped into batches of size 100, with sharding.

When there is only one worker thread, as expected, they are processed by the sole worker and put into the same `KV<Integer, Iterable<T>>`, which is then processed by the same worker.

I then set the worker number to 2. I did not check if the 2 `KV`s are processed by different workers, but they are indeed (consistently) grouped into two batches of `KV<Integer, Iterable<T>>`, each containing one `T`.

So my understanding is that, `GroupIntoBatches.withSharededKey()` processes all the `KV` with the same key serially, but not necessarily on the same worker (otherwise it would not have been possible to get different `ShardedKey` for these `KV`s, as the `ShardedKey` is determined by the worker and thread ID). Otherwise it would contradict with the [claim](https://github.com/apache/beam/blob/4ab557bf98784e51b6f8dca7d89ba5bcea9f04ef/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L227) that "by default, keys are sharded to such that the input elements with the same key are spread to all available threads executing the transform". There must be some communication between workers working on this transform to coordinate the emission of batches in a serial manner, which makes total sense as you have to have processed `n` number of KVs *in total* in order to emit a batch of size `n`. It is also implied in the [javadoc](https://github.com/apache/beam/blob/4ab557bf98784e51b6f8dca7d89ba5bcea9f04ef/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L57) where it says "Elements are buffered until there are enough elements for a batch, at which point they are emitted to the output {@link PCollection}". This is likely achieved through [stateful processing](https://beam.apache.org/blog/stateful-processing/).

Given that the "sharding" function in our code base only assigns a random key to the input date, this "map" transform would not alter the distribution of the input data among workers, regardless of the range of random key assigned. The resulting `KV` should, statistically, have exactly the same distribution of elements within the same key. Therefore the net result of this "sharding" should be negligible. In fact, because each set with the same key has fewer element in them, and a batch is only emitted when a batch is full or when the elements are exhausted, if we assume the number of total number of elements is much larger than the batch size, and the "sharding" factor x is reasonably small so that each "shard" still contains much more elements than the batch size, it appears that having more "shards" this way would increase the latency of this transform as we have the same number of workers, but the buffer size is x times larger (as each key has its own buffer), which takes x times to fill before any worker starts emitting batches.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1856)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/beam/common/RegistryJpaIO.java` line 293 at r1](https://reviewable.io/reviews/google/nomulus/1856#-NHzHIE-3Swvq7Xp9Eur:-NJ3nDg07s5nwTo74ZJt:bbrhij) ([raw file](https://github.com/google/nomulus/blob/279181ac4b7dba9d9631624a7915ac6128b0c22d/core/src/main/java/google/registry/beam/common/RegistryJpaIO.java#L293)):*

So I did some testing in #1881. This is done within unit tests so it employs only the direct runner, but I believe we should expect the axiom to hold that the parallelism behavior is not determined by the type of runner used.

I started with two elements `T`, which are then keyed to the same dummy key `Integer`. The resulting `KV<Integer, T>` are then grouped into batches of size 100, with sharding.

When there is only one worker thread, as expected, they are processed by the sole worker and put into the same `KV<Integer, Iterable<T>>`, which is then processed by the same worker.

I then set the worker number to 2. I did not check if the 2 `KV`s are processed by different workers, but they are indeed (consistently) grouped into two batches of `KV<Integer, Iterable<T>>`, each containing one `T`.

So my understanding is that, `GroupIntoBatches.withSharededKey()` processes all the `KV` with the same key serially, but not necessarily on the same worker (otherwise it would not have been possible to get different `ShardedKey` for these `KV`s, as the `ShardedKey` is determined by the worker and thread ID). Otherwise it would contradict with the [claim](https://github.com/apache/beam/blob/4ab557bf98784e51b6f8dca7d89ba5bcea9f04ef/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L227) that "by default, keys are sharded to such that the input elements with the same key are spread to all available threads executing the transform". There must be some communication between workers working on this transform to coordinate the emission of batches in a serial manner, which makes total sense as you have to have processed `n` number of KVs *in total* in order to emit a batch of size `n`. It is also implied in the [javadoc](https://github.com/apache/beam/blob/4ab557bf98784e51b6f8dca7d89ba5bcea9f04ef/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L57) where it says "Elements are buffered until there are enough elements for a batch, at which point they are emitted to the output {@link PCollection}". This is likely achieved through [stateful processing](https://beam.apache.org/blog/stateful-processing/).

Given that the "sharding" function in our code base only assigns a random key to the input date, this "map" transform would not alter the distribution of the input data among workers, regardless of the range of random key assigned. The resulting `KV` should, statistically, have exactly the same distribution of elements within the same key. Therefore the net result of this "sharding" should be negligible. In fact, because each set with the same key has fewer element in them, and a batch is only emitted when a batch is full or when the elements are exhausted, if we assume the number of total number of elements is much larger than the batch size, and the "sharding" factor x is reasonably small so that each "shard" still contains much more elements than the batch size, it appears that having more "shards" this way would increase the latency of this transform as we have the same number of workers, but the buffer size is x times larger (as each key has its own buffer), which takes x times to fill before any worker starts emitting batches.
</blockquote></details>

FYI I will do more testing by running `ExpandRecurringBillingEventsPipeline` on dataflow with different "sharding" factors.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1856)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1858, opened 2022-11-23
Remove usage of the AppEngine remote API

gbrodman:
This is only used for contacting Datastore. With the removal of:
1. All standard usages of Datastore
2. Usage of Datastore for allocation of object IDs
3. Usage of Datastore for GAE user IDs

we can remove the remote API without affecting functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1858)


jianglai (changes requested):
I believe the connection also allows you to call GAE SDKs APIs locally, namely any API in the following packages:

com.google.appengine.*
com.google.apphosting.*

I suspect it's not the reason that we use the connection, but we should still check if there's any import statements for packages under those namespaces.

Plus, I know there's a lot of commands here, but it might be worth it to do the following:

1) spot checking some of them to make sure they still work.
2) find if any of them actually need to implement `CommandWithSql`. Anything whose name that starts with `List|Create|Update` probably can be safely assumed to need SQL access. But what about, for example, `GhostrydeCommand`? It doesn't seem to call `tm()` directly. Does it actually need to set up SQL?
3) alternatively, do we know if it's worth it to make such a distinction anymore? Previously (and supposedly), setting up a connection is an expensive operation that you only want to do if you have to. But is setting up the SQL credential itself expensive? I don't think we are opening connections by setting the transaction manager, right? Maybe we can just set it regardless of what kind of command it is?

> Reviewed 50 of 50 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1858#-:-NI4Iqt13UZx5QAQ0ftr:b-w9hf6e)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/RegistryCli.java` line 205 at r1](https://reviewable.io/reviews/google/nomulus/1858#-NI4I2vo9ViddrVLVAaJ:-NI4I2vo9ViddrVLVAaK:b2tlf97) ([raw file](https://github.com/google/nomulus/blob/0f2c751b9df7b6068ef641cb4f86495bbe17abfa/core/src/main/java/google/registry/tools/RegistryCli.java#L205)):*
> ```Java
> 
>   @Override
>   public void close() {
> ```

This is no longer necessary. In fact I think the class does not need to implement `AutoCloseable` any more.



_Code quote:_
```Java
public void close() {
```
___
*[`core/src/main/java/google/registry/tools/RegistryCli.java` line 254 at r1](https://reviewable.io/reviews/google/nomulus/1858#-NI4ILYCCE8OmJEXZ_0B:-NI4ILYCCE8OmJEXZ_0C:bwm1asz) ([raw file](https://github.com/google/nomulus/blob/292bc788fbc3ce539b183582905a347ab3f29bb5/core/src/main/java/google/registry/tools/RegistryCli.java#L254)):*
> ```Java
>           options.useDevelopmentServerCredential();
>         } else {
>           RemoteApiOptionsUtil.useGoogleCredentialStream(
> ```

We can finally remove this class now!



_Code quote:_
```Java
 RemoteApiOptionsUtil.
```

gbrodman (commented):
1. I spot-checked the List*Commands but yeah there's no harm in trying more. Will do that. 
2. Actually, the SQL setup seems to be lazy (it sets a Supplier rather than the entity itself). So I would expect that it is actually pretty quick! 
3. Yes, we can just remove that concept entirely I think. It simplifies things a lot and removes a possible footgun (that we've run into multiple times) where tests for a command work because they have SQL set up but the command fails remotely because we forgot to set up SQL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1858#-:-NI8O0eN0rMzzeRlqi11:b18oku1)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/RegistryCli.java` line 205 at r1](https://reviewable.io/reviews/google/nomulus/1858#-NI4I2vo9ViddrVLVAaJ:-NI8MM2O1J3sSSeWGNgy:bgsb48g) ([raw file](https://github.com/google/nomulus/blob/0f2c751b9df7b6068ef641cb4f86495bbe17abfa/core/src/main/java/google/registry/tools/RegistryCli.java#L205)):*

This is no longer necessary. In fact I think the class does not need to implement `AutoCloseable` any more.
</blockquote></details>

yeah, I think so -- we can remove the try-with from all the callsites too 
___
*[`core/src/main/java/google/registry/tools/RegistryCli.java` line 254 at r1](https://reviewable.io/reviews/google/nomulus/1858#-NI4ILYCCE8OmJEXZ_0B:-NI8LfF-7UoCwZak15H4:b-3rfhu1) ([raw file](https://github.com/google/nomulus/blob/292bc788fbc3ce539b183582905a347ab3f29bb5/core/src/main/java/google/registry/tools/RegistryCli.java#L254)):*

We can finally remove this class now!
</blockquote></details>

!!! yeah

jianglai (approved):
LGTM. Can be submitted after sufficient testing.

> Reviewed 55 of 55 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1858#-:-NI8ZhK6BPwAFeuU2rxr:b-6hhznv)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/RegistryCli.java` line 222 at r2](https://reviewable.io/reviews/google/nomulus/1858#-NI8ZVmn-sTyHjceMUnb:-NI8ZVmo3uXw1Y8bm7GS:bbdb80t) ([raw file](https://github.com/google/nomulus/blob/4a4254362914143294d26e30a5ce62e6fd1551ed/core/src/main/java/google/registry/tools/RegistryCli.java#L222)):*
> ```Java
>     if (!databaseInstalled) {
>         // Ensure that all entity classes are loaded before command code runs.
>         ObjectifyService.initOfy();
> ```

You probably can remove this as well. Even if it initializes ofy locally, it's not going to be able to use it anyway, right?



_Code quote:_
```Java
ObjectifyService.initOfy();
```

gbrodman (commented):
Testing:
- list TLDs / domains  / get /modify domain
- listing / getting / updating premium list 
- generate / get allocation tokens
- get claims list
- list / get / update registrars
- shell

maybe also double-check the changes in RegistryCli (tests were failing) 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1858#-:-NI8xfGx2rd1Y8ChSvG4:bgrmgxr)* status: 53 of 55 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/RegistryCli.java` line 222 at r2](https://reviewable.io/reviews/google/nomulus/1858#-NI8ZVmn-sTyHjceMUnb:-NI8xcMfBBUgPlnCVJBL:bls4lf) ([raw file](https://github.com/google/nomulus/blob/4a4254362914143294d26e30a5ce62e6fd1551ed/core/src/main/java/google/registry/tools/RegistryCli.java#L222)):*

You probably can remove this as well. Even if it initializes ofy locally, it's not going to be able to use it anyway, right?
</blockquote></details>

yeah, I think because we don't need it for ID allocation any more

gbrodman:
just double-checking, did you want to take a look at the changes in RegistryCli.java or do you think it's all good?

PR #1869, opened 2022-12-03
Add a command to create synthetic contact history

weiminyu:
Create synthetic ContactHistory to fix b/261192144.

Caller provides a file that enumerates the contact repo_ids to operate on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1869)

### DIFF
--- a/core/src/main/java/google/registry/tools/RegistryTool.java
+++ b/core/src/main/java/google/registry/tools/RegistryTool.java
@@ -17,6 +17,7 @@ package google.registry.tools;
 import com.google.common.collect.ImmutableMap;
 import google.registry.tools.javascrap.CompareEscrowDepositsCommand;
 import google.registry.tools.javascrap.CreateCancellationsForOneTimesCommand;
+import google.registry.tools.javascrap.CreateSyntheticContactHistoriesCommand;
 import google.registry.tools.javascrap.CreateSyntheticDomainHistoriesCommand;
 
 /** Container class to create and run remote commands against a Datastore instance. */
@@ -48,6 +49,7 @@ public final class RegistryTool {
           .put("create_registrar", CreateRegistrarCommand.class)
           .put("create_registrar_groups", CreateRegistrarGroupsCommand.class)
           .put("create_reserved_list", CreateReservedListCommand.class)
+          .put("create_synthetic_contact_histories", CreateSyntheticContactHistoriesCommand.class)
           .put("create_synthetic_domain_histories", CreateSyntheticDomainHistoriesCommand.class)
           .put("create_tld", CreateTldCommand.class)
           .put("curl", CurlCommand.class)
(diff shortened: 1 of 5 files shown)


gbrodman (commented):
According to the linked bug, it seems like you've already run this against production. Is that correct? Just want to double check, if so we probably don't absolutely `need` to merge this. 

> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1869#-:-NIY9TZ44QkulrSoIiE-:b-9thrnw)* status: 1 of 5 files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (commented):
Yes, they've been applied.

We don't have to merge it but should probably keep this PR around for a while.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1869#-:-NIYGeeJ221bfR3yaiB_:bygesux)* status: 1 of 5 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (changes requested):
I somewhat feel conflicted about putting these commands in `javascrap`. On one hand, I think `javascrap` should only host tools that are temporarily needed for a while and to be deleted later. On the other hand, these commands seem genuinely useful, especially for troubleshooting and fixing data issues and can be used in the future. Should we consider moving them outside `javascrap` to `tools` itself and add proper tests for them?

Or maybe alternatively, we should consider adding a subpackage like `troubleshooting` or `debugging` that hosts these commands that are not part of the normal operation but nevertheless is useful to keep permanently.

> Reviewed 2 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1869#-:-NIZZgWn-HGpy6Ci24fA:b-xoh9e4)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/javascrap/DeleteContactByRepoIdCommand.java` line 61 at r1](https://reviewable.io/reviews/google/nomulus/1869#-NIXva-x4PwX8h0fD7nK:-NIXva-x4PwX8h0fD7nL:b-nsvlb) ([raw file](https://github.com/google/nomulus/blob/5fa45809c4fdfec4b3caf2c7a24de7edfce1b16b/core/src/main/java/google/registry/tools/javascrap/DeleteContactByRepoIdCommand.java#L61)):*
> ```Java
>  * generates a history event.
>  *
>  * <p>The {@code ContactDeleteFlow} itself cannot be used because deletion by EPP command is based
> ```

I wonder what happens when an EPP contact delete request is sent with a contact ID that has multiple entities? I suspect `loadByForeignKey` would throw an error. We should mention that behavior here.



_Code quote:_
```Java
 * <p>The {@code ContactDeleteFlow} itself cannot be used because deletion by EPP command is based
 * on contact_id, and there is no control on which one of a pair is deleted.
```

PR #1881, opened 2022-12-11
Add a beam pipeline to expand recurring billing event

jianglai:
This will replace the ExpandRecurringBillingEventsAction, which has a
couple of issues:

1) The action starts with too many Recurrings that are later filtered out
   because their expanded OneTimes are not actually in scope. This is due
   to the Recurrings not recording its latest expanded event time, and
   therefore many Recurrings that are not yet due for renewal get included
   in the initial query.

2) The action works in sequence, which exacerbated the issue in 1) and
   makes it very slow to run if the window of operation is wider than
   one day, which in turn makes it impossible to run any catch-up
   expansions with any significant gap to fill.

3) The action only expands the recurrence when the billing times because
   due, but most of its logic works on event time, which is 45 days
   before billing time, making the code hard to reason about and
   error-prone.  This has led to b/258822640 where a premature
   optimization intended to fix 1) caused some autorenwals to not be
   expanded correctly when subsequent manual renews within the autorenew
   grace period closed the original recurrece.

As a result, the new pipeline addresses the above issues in the
following way:

1) Update the recurrenceLastExpansion field on the Recurring when a new
   expansion occurs, and narrow down the Recurrings in scope for
   expansion by only looking for the ones that have not been expanded for
   more than a year.

2) Make it a Beam pipeline so expansions can happen in parallel. The
   Recurrings are grouped into batches in order to not overwhelm the
   database with writes for each expansion.

3) Create new expansions when the event time, as opposed to billing
   time, is within the operation window. This streamlines the logic and
   makes it clearer and easier to reason about. This also aligns with
   how other (cancelllable) operations for which there are accompanying
   grace periods are handled, when the corresponding data is always
   speculatively created at event time. Lastly, doing this negates the
   need to check if the expansion has finished running before generating
   the monthly invoices, because the billing events are now created not
   just-in-time, but 45 days in advance.

Note that this PR only adds the pipeline. It does not switch the default
behavior to using the pipeline, which is still done by
ExpandRecurringBillingEventsAction. We will first use this pipeline to
generate missing billing events and domain histories caused by
b/258822640. This also allows us to test it in production, as it
backfills data that will not affect ongoing invoice generation. If
anything goes wrong, we can always delete the generated billing events
and domain histories, based on the unique "reason" in them.

This pipeline can only run after we switch to use SQL sequence based ID
allocation, introduced in https://github.com/google/nomulus/pull/1831.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1881)


jianglai (commented):
> Reviewed 20 of 24 files at r1, 10 of 10 files at r2, 1 of 1 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys and @weiminyu)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys and @weiminyu)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 244 at r5](https://reviewable.io/reviews/google/nomulus/1881#-NJ6IjTR-4bZefkEmvpI:-NJ6UEUt4TwYPoEmmRYS:bz6xhuo) ([raw file](https://github.com/google/nomulus/blob/551e3accfccc9017e910f670250b4211a7a447cf/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L244)):*

I realized after commenting that expandOneRecurring makes a fresh query for OneTime entities each time.

It is still possible for a write-after-read race, and we should use SERIALIZABLE isolation level.


</blockquote></details>

Scenario: one batch is sent but got delayed. Scheduler decides to retry. Now there is a (slim) chance that two transactions expanding the
same recurring entity are concurrent.
___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 390 at r5](https://reviewable.io/reviews/google/nomulus/1881#-NJ6U70TA5NTZn7a-nb8:-NJ6U70UECJoLCVNNuxY:b-fadlrq) ([raw file](https://github.com/google/nomulus/blob/551e3accfccc9017e910f670250b4211a7a447cf/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L390)):*
> ```Java
>     //
>     // See: https://www.postgresql.org/docs/current/transaction-iso.html
>     options.setIsolationOverride(TransactionIsolationLevel.TRANSACTION_REPEATABLE_READ);
> ```

I think SERIALIZABLE is required. Please see comments above.



_Code quote:_
```Java
TRANSACTION_REPEATABLE_READ
```

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 244 at r5](https://reviewable.io/reviews/google/nomulus/1881#-NJ6IjTR-4bZefkEmvpI:-NJ7FLesEAgbW_jxDXFG:b-2upwvp) ([raw file](https://github.com/google/nomulus/blob/551e3accfccc9017e910f670250b4211a7a447cf/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L244)):*

I considered using `SERIALIZABLE`, but that would result in many retries. From what I saw in unit test, even if I just had two different recurrings for two different domains, running the pipeline on two workers would result in serialization errors and retries, i.e. the following:

```ERROR:  could not serialize access due to read/write dependencies among transactions```

That would imply that all `expandOneRecurring` would effectively have to run serially, defeating the purpose of the pipeline. I don't quite understand why this would have caused an serialization error in practice as the expansions of the two recurrings should not conflict one another. Maybe PostgreSQL just sees that both queries tried to write to the same table so it thinks serialization error *could* happen? From what I read in their doc, they seem to imply that the database is actually monitoring the rows affected by each query and determining if actual serialization anomaly would occur based on the query condition and the data in the database, i. e. it should not lock the entire table just to proactively prevent serialization anomaly, which is what I observed in reality.

Not that I read again on [transaction isolation level for PostgreSQL](https://www.postgresql.org/docs/current/transaction-iso.html), it does seem that repeatable read is not sufficient as the query to find existing OneTimes prevents phantom read by already reading from a snapshot taken at the beginning of a transaction, regardless of what other concurrent transactions might have done to alter the result of the query. I had believe that it prevents phantom read by not allowing concurrent transactions to do so, but it looks like it simply ignores the effect of the current transactions...
</blockquote></details>

The reason the `could not serialize access` error happens in unit test is that the query for `OneTime` uses an index, and in addition to placing row-locks on the
rows loaded, the query also places a page-level predicate lock on the index. With unit tests, the index is guaranteed to be on the same page. In production, this should happen
less frequently.

Have you tried to run a job at SERIALIZABLE and see how much slower it is compared to with REPEATABLE_READ?

If there is a big slow down, an alternative may be to:
1. Use a lock to ensure one pipeline running at a time.
2.  Stick to REPEATABLE_READ
3. Separate the creation of new OneTimes and their persistence. Basically like what I outlined in my original comment.

**Review by jianglai — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 244 at r5](https://reviewable.io/reviews/google/nomulus/1881#-NJ6IjTR-4bZefkEmvpI:-NJ7Wkpv2P_eFotryCeO:b-g8orq1) ([raw file](https://github.com/google/nomulus/blob/551e3accfccc9017e910f670250b4211a7a447cf/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L244)):*

The reason the `could not serialize access` error happens in unit test is that the query for `OneTime` uses an index, and in addition to placing row-locks on the
rows loaded, the query also places a page-level predicate lock on the index. With unit tests, the index is guaranteed to be on the same page. In production, this should happen
less frequently.

Have you tried to run a job at SERIALIZABLE and see how much slower it is compared to with REPEATABLE_READ?

If there is a big slow down, an alternative may be to:
1. Use a lock to ensure one pipeline running at a time.
2.  Stick to REPEATABLE_READ
3. Separate the creation of new OneTimes and their persistence. Basically like what I outlined in my original comment.

</blockquote></details>

That's interesting, where can I read about this, regarding which locks are used at different isolation level? Can you share a link?

I will test the performance at `SERIALIZABLE`, though it probably will need to be done on production in dryrun mode to be meaningful.

Regarding the alternatives:

1) It looks like we have to use a `Lock` directly, and keep track of the dataflow job status to unlock it, instead of using a `LockHandler`, if we really want to ensure only one job is running. It also doesn't solve the problem with retries attempted within the same pipeline by the scheduler.

2) That means accepting the risk of potential duplicates? That seems unacceptable given that this concerns billing...

3) I originally had what you suggested, where the creation and persistence are separated. I had to make some changes to the `SqlWrite` as it needs to support writing several entities in a single transaction (OneTime, DomainHistory, and the updated Recurring). This should prevent duplicate entities from being created because the creation of OneTime and DomainHistory is now a pure function with no side effect (but there is still complication with Recurring, as we currently prevent updating existing entities in the writer, see #1824). This also does not prevent concurrent running pipelines to create duplicate entities.

So it looks like there's no perfect solution other than using `SERIALIZABLE`, which is why I'm keen on learning more about its behavior.

**Review by weiminyu — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: 28 of 30 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 244 at r5](https://reviewable.io/reviews/google/nomulus/1881#-NJ6IjTR-4bZefkEmvpI:-NJ956hKC9jQohY5jYEJ:b7uahav) ([raw file](https://github.com/google/nomulus/blob/551e3accfccc9017e910f670250b4211a7a447cf/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L244)):*

That's interesting, where can I read about this, regarding which locks are used at different isolation level? Can you share a link?

I will test the performance at `SERIALIZABLE`, though it probably will need to be done on production in dryrun mode to be meaningful.

Regarding the alternatives:

1) It looks like we have to use a `Lock` directly, and keep track of the dataflow job status to unlock it, instead of using a `LockHandler`, if we really want to ensure only one job is running. It also doesn't solve the problem with retries attempted within the same pipeline by the scheduler.

2) That means accepting the risk of potential duplicates? That seems unacceptable given that this concerns billing...

3) I originally had what you suggested, where the creation and persistence are separated. I had to make some changes to the `SqlWrite` as it needs to support writing several entities in a single transaction (OneTime, DomainHistory, and the updated Recurring). This should prevent duplicate entities from being created because the creation of OneTime and DomainHistory is now a pure function with no side effect (but there is still complication with Recurring, as we currently prevent updating existing entities in the writer, see #1824). This also does not prevent concurrent running pipelines to create duplicate entities.

So it looks like there's no perfect solution other than using `SERIALIZABLE`, which is why I'm keen on learning more about its behavior.
</blockquote></details>

If we separate creation and persistence, AND pre-assign IDs to created entities, then we are robust against pipeline retries, although we may waste some ids if retry happens.
It would work like this:
1. Stage one, we expand recurrings and generate onetimes with ids. Retries by beam are fine because nothing is written, so every recurring always gets the same set of onetimes.
2. anti-fusing filler like a groupby, this ensures that writes in the next stage do not happen until every recurring has been successfully expanded.
3. Stage two, write those Onetimes. Retries by beam are again fine, since duplicate writes will fail (if we use insert) or write idempotently (if we use upsert).

If we want to save ids, we can add another stage that only generates ids for OneTimes created by an earlier stage. That is, making it simple so that retries are unlikely to happen. 

Let's see how a run with SERIALIZABLE is like first.


As for predicate locking, the link below passingly mentions that the B+-tree based index uses page-locks. The document itself is a proposal to implement page lock for
other types of indexes, which by the time of writing, only had relation-level locks.

https://wiki.postgresql.org/wiki/Predicate_locks_in_index_GSoC_2017

**Review by jianglai — COMMENTED** [review]
> Reviewed 1 of 6 files at r7, 5 of 8 files at r9.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: 26 of 31 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

**Review by weiminyu — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 430 at r12](https://reviewable.io/reviews/google/nomulus/1881#-NLN94uJ-31dGVKpE-2H:-NLN94uJ-31dGVKpE-2I:b-tvw5nf) ([raw file](https://github.com/google/nomulus/blob/368469f97b51e47ee984012b058d9290d4de386a/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L430)):*
> ```Java
>     // Hardcode the transaction level to be at repeatable read because we do not want concurrent
>     // runs of the pipeline for the same window to create duplicate OneTimes. This ensures that the
>     // set of existing OneTimes do not change by the time new OneTimes are inserted within a
> ```

The comment is outdated.



_Code quote:_
```Java
    // Hardcode the transaction level to be at repeatable read because we do not want concurrent
    // runs of the pipeline for the same window to create duplicate OneTimes. This ensures that the
    // set of existing OneTimes do not change by the time new OneTimes are inserted within a
    // transaction.
    //
    // Note that the SQL spec itself allows for phantom read in a transaction, but PostgreSQL's
    // implementation does not, which means the set of OneTimes that satisfy the condition of being
    // matched to the same recurring events does not change during the transaction, preventing
    // the aforementioned duplication.
    //
    // We could have used the default SERIALIZABLE level, but that would have locked the entire
    // BillingEvent table and force every write to it to be done sequentially, effectively reducing
    // the parallelism of the pipeline to 1.
    //
    // See: https://www.postgresql.org/docs/current/transaction-iso.html
```

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 430 at r12](https://reviewable.io/reviews/google/nomulus/1881#-NLN94uJ-31dGVKpE-2H:-NLNWcwI3tbjk286q7se:bp0n5gd) ([raw file](https://github.com/google/nomulus/blob/368469f97b51e47ee984012b058d9290d4de386a/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L430)):*

The comment is outdated.
</blockquote></details>

Good catch. done.

jianglai (commented):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1881)* status: 30 of 31 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

PR #1887, opened 2022-12-20
Add an optional IAP-enabled ID token when using the Nomulus tool

gbrodman:
We can use the saved refresh token associated with the nomulus tool to request an ID token with an audience of the IAP client in order to satisfy IAP with with the Nomulus tool.

Note: this requires that the user of the Nomulus tool, e.g. "gbrodman@google.com" has a User object stored in SQL if using IAP.

Tested on QA

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1887)

### DIFF
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1153,6 +1153,12 @@ public final class RegistryConfig {
       return ImmutableSet.copyOf(config.oAuth.allowedOauthClientIds);
     }
 
+    @Provides
+    @Config("iapClientId")
+    public static Optional<String> provideIapClientId(RegistryConfigSettings config) {
+      return Optional.ofNullable(config.oAuth.iapClientId);
+    }
+
     /**
      * Provides the OAuth scopes required for accessing Google APIs using the default credential.
      */
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -61,6 +61,7 @@ public class RegistryConfigSettings {
     public List<String> availableOauthScopes;
     public List<String> requiredOauthScopes;
     public List<String> allowedOauthClientIds;
+    public String iapClientId;
   }
 
   /** Configuration options for accessing Google APIs. */
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -306,6 +306,9 @@ oAuth:
   # in this list. Client IDs are typically of the format
   # numbers-alphanumerics.apps.googleusercontent.com
   allowedOauthClientIds: []
+  # GCP Identity-Aware Proxy client ID, if set up (note: this requires manual setup
+  # of User objects in the database for Nomulus tool users)
+  iapClientId: null
 
 credentialOAuth:
   # OAuth scopes required for accessing Google APIs using the default
--- a/core/src/main/java/google/registry/tools/RequestFactoryModule.java
+++ b/core/src/main/java/google/registry/tools/RequestFactoryModule.java
@@ -14,13 +14,23 @@
 
 package google.registry.tools;
 
+import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
+import com.google.api.client.http.HttpResponse;
+import com.google.api.client.http.UrlEncodedContent;
 import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.util.GenericData;
+import com.google.auth.oauth2.UserCredentials;
 import dagger.Module;
 import dagger.Provides;
 import google.registry.config.CredentialModule.DefaultCredential;
 import google.registry.config.RegistryConfig;
+import google.registry.config.RegistryConfig.Config;
 import google.registry.util.GoogleCredentialsBundle;
+import java.io.IOException;
+import java.net.URI;
+import java.util.Optional;
 
 /**
  * Module for providing the HttpRequestFactory.
@@ -33,9 +43,21 @@ class RequestFactoryModule {
 
   static final int REQUEST_TIMEOUT_MS = 10 * 60 * 1000;
 
+  /**
+   * Server to use if we want to manually request an IAP ID token
+   *
+   * <p>If we need to have an IAP-enabled audience, we can use the existing refresh token and the
+   * IAP client ID audience to request an IAP-enabled ID token. This token is read and used by
+   * {@link google.registry.request.auth.IapHeaderAuthenticationMechanism}, and it requires that the
+   * user have a {@link google.registry.model.console.User} object present in the database.
+   */
+  private static final GenericUrl TOKEN_SERVER_URL =
+      new GenericUrl(URI.create("https://oauth2.googleapis.com/token"));
+
   @Provides
   static HttpRequestFactory provideHttpRequestFactory(
-      @DefaultCredential GoogleCredentialsBundle credentialsBundle) {
+      @DefaultCredential GoogleCredentialsBundle credentialsBundle,
+      @Config("iapClientId") Optional<String> iapClientId) {
     if (RegistryConfig.areServersLocal()) {
       return new NetHttpTransport()
           .createRequestFactory(
@@ -47,7 +69,15 @@ class RequestFactoryModule {
       return new NetHttpTransport()
           .createRequestFactory(
               request -> {
-                credentialsBundle.getHttpRequestInitializer().initialize(request);
+                // If using IAP, use the refresh token to acquire an IAP-enabled ID token and use
+                // that for authentication.
+                if (iapClientId.isPresent()) {
+                  String idToken = getIdToken(credentialsBundle, iapClientId.get());
+                  request.getHeaders().setAuthorization("Bearer " + idToken);
+                } else {
+                  // Otherwise, use the standard credential HTTP initializer
+                  credentialsBundle.getHttpRequestInitializer().initialize(request);
+                }
                 // GAE request times out after 10 min, so here we set the timeout to 10 min. This is
                 // needed to support some nomulus commands like updating premium lists that take
                 // a lot of time to complete.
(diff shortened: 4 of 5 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1887)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 308 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJ9J-V3JKZ_AMH30qU:-NKJ9J-V3JKZ_AMH30qV:bjqp556) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/config/files/default-config.yaml#L308)):*
> ```YAML
>   # in this list. Client IDs are typically of the format
>   # numbers-alphanumerics.apps.googleusercontent.com
>   allowedOauthClientIds: []
> ```

Is my understanding correct that once we put every endpoint behind IAP, we will only need the `iapClientId`, but not this list, as client id checks would have been done by IAP before the request reaches our app?



_Code quote:_
```YAML
  allowedOauthClientIds: []
```
___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJCQ6LEkjOsy5eIyOo:-NKJCQ6LEkjOsy5eIyOp:b-bfpsj3) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L64)):*
> ```Java
>           .createRequestFactory(
>               request -> {
>                 credentialsBundle.getHttpRequestInitializer().initialize(request);
> ```

I think what this does is to set the `Credential` as an interceptor for the request, which in turn refreshes (if necessary) and add the token to the request. So It should be part of the `else` clause below, as we only need this behavior (i.e. using the access token instead of ID token) if the IAP client ID is not set. In another word, it should not be necessary to do this, if we know we are going to overwrite the header with an ID token later.



_Code quote:_
```Java
credentialsBundle.getHttpRequestInitializer().initialize(request);
```
___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJE3ix5_kCffOIO9JM:-NKJE3iyC4949yLkRwVJ:b-o80di1) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L68)):*
> ```Java
>                 // that instead for authentication
>                 if (iapClientId.isPresent()) {
>                   String idToken = getIdToken(credentialsBundle, iapClientId.get());
> ```

It probably doesn't matter much for the tool, where performance is not critical. But it would be better to cache the ID token instead of having to request one for each HTTP request, which involves an extra round trip to the token server for each HTTP request we send to nomulus.

It would be a must-have for the proxy, which currently does this:

https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/ProxyModule.java;l=230;bpv=0;bpt=1

But my understanding is that we only have to manually create the ID token for local credentials, is that right? For the proxy which uses default service account credential derived from GKE cluster, the ID token can be obtained directly using existing libraries, is that correct? In that case the refresh-if-necessary behavior should be taken care of by the library.



_Code quote:_
```Java
String idToken = getIdToken(credentialsBundle, iapClientId.get());
```

jianglai (commented):
> this requires that the user of the Nomulus tool, e.g. "gbrodman@google.com" has a User object stored in SQL if using IAP.
We should put this note in the code somewhere, maybe in the config file where the `iapClientId` field is defined?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1887#-:-NKJJF87-sgoUETWAZr1:bctq84t)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @ptkach)

gbrodman (commented):
good call

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1887#-:-NKOpR9G1poT6_396oK_:b7anmnw)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 308 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJ9J-V3JKZ_AMH30qU:-NKOpzsZ-f4m1pm-CtYL:bhg9l4s) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/config/files/default-config.yaml#L308)):*

Is my understanding correct that once we put every endpoint behind IAP, we will only need the `iapClientId`, but not this list, as client id checks would have been done by IAP before the request reaches our app?
</blockquote></details>

Possibly, hopefully! It seems like the other users of this are the "domain-watcher" and the proxy. 
___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJCQ6LEkjOsy5eIyOo:-NKOrJPv-xCpujwWc8So:b-6fta94) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L64)):*

I think what this does is to set the `Credential` as an interceptor for the request, which in turn refreshes (if necessary) and add the token to the request. So It should be part of the `else` clause below, as we only need this behavior (i.e. using the access token instead of ID token) if the IAP client ID is not set. In another word, it should not be necessary to do this, if we know we are going to overwrite the header with an ID token later.
</blockquote></details>

Good call (we still want to keep the timeouts) 
___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJE3ix5_kCffOIO9JM:-NKOsJwU1S37lCs_k1fd:b-l4q20e) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L68)):*

It probably doesn't matter much for the tool, where performance is not critical. But it would be better to cache the ID token instead of having to request one for each HTTP request, which involves an extra round trip to the token server for each HTTP request we send to nomulus.

It would be a must-have for the proxy, which currently does this:

https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/ProxyModule.java;l=230;bpv=0;bpt=1

But my understanding is that we only have to manually create the ID token for local credentials, is that right? For the proxy which uses default service account credential derived from GKE cluster, the ID token can be obtained directly using existing libraries, is that correct? In that case the refresh-if-necessary behavior should be taken care of by the library.
</blockquote></details>

Yeah, if the proxy uses the default service account then we should just be able to do https://cloud.google.com/iap/docs/authentication-howto#obtaining_an_oidc_token_for_the_default_service_account using the standard libraries and that should hopefully take care of everything under the hood for us. 

I'm a bit leery of ID token caching because it'd have to use the disk and (like you said) it's not a huge deal for one more HTTP round trip with the tool.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1887)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 308 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJ9J-V3JKZ_AMH30qU:-NKTBqpo5KEn0gkgyzOO:b-r445f0) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/config/files/default-config.yaml#L308)):*

Possibly, hopefully! It seems like the other users of this are the "domain-watcher" and the proxy. 
</blockquote></details>

Yeah, we might to investigate how to let domain-watcher send requests that can pass IAP if we want to put everything behind IAP. But that's for another day.
___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJE3ix5_kCffOIO9JM:-NKTCR935psrXfNdGYMU:blbdz1g) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L68)):*

Yeah, if the proxy uses the default service account then we should just be able to do https://cloud.google.com/iap/docs/authentication-howto#obtaining_an_oidc_token_for_the_default_service_account using the standard libraries and that should hopefully take care of everything under the hood for us. 

I'm a bit leery of ID token caching because it'd have to use the disk and (like you said) it's not a huge deal for one more HTTP round trip with the tool. 
</blockquote></details>

Not sure what you meant by "having to use the disk"? The refresh token is already saved in the disk, isn't it? Why is it a concern to use the disk? Plus, the caching itself doesn't need disk, per se. We can just use an in-memory memoizer with expiry that Guava provides, right?

I guess another concern (aside from performance) is that whether there's any adverse effect to refresh a token too frequently? Would the token server have any sort of rate limit? That's why I think if implementing caching is relatively trivial, we might as well do it.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1887)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJE3ix5_kCffOIO9JM:-NKTmlbKCBISzKhAdANt:b-78fqy4) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L68)):*

Not sure what you meant by "having to use the disk"? The refresh token is already saved in the disk, isn't it? Why is it a concern to use the disk? Plus, the caching itself doesn't need disk, per se. We can just use an in-memory memoizer with expiry that Guava provides, right?

I guess another concern (aside from performance) is that whether there's any adverse effect to refresh a token too frequently? Would the token server have any sort of rate limit? That's why I think if implementing caching is relatively trivial, we might as well do it.
</blockquote></details>

There's no point in using an in-memory cache because this is client-side so the memory only lives as long as the nomulus command (unless, possibly, we're running the shell but I believe that is infrequent). We'd have to write the ID token to disk and check its expiration every time. Much simpler to just use the already-existing refresh token.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1887)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 68 at r1](https://reviewable.io/reviews/google/nomulus/1887#-NKJE3ix5_kCffOIO9JM:-NKxQeXxDjWvza1wlDMd:bx3z4xe) ([raw file](https://github.com/google/nomulus/blob/e7b035b9e2a2516d2c280eb6d6bd6eb5b29a1c64/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L68)):*

There's no point in using an in-memory cache because this is client-side so the memory only lives as long as the nomulus command (unless, possibly, we're running the shell but I believe that is infrequent). We'd have to write the ID token to disk and check its expiration every time. Much simpler to just use the already-existing refresh token. 
</blockquote></details>

I was thinking about just a fixed-timed cache, which would not require checking the expiration every time (but may result in some unnecessary refreshing). But yeah, I guess other then using the shell mode, an in-memory would not necessarily help much. I however think the library might be doing exactly what you described with the access token, i.e. persisting it to the disk and reusing it if possible.

Anyway, it doesn't seem worth the effort for us to implement the same. Let's do without caching for now.

PR #1890, opened 2022-12-28
Domain update flow - disable DS data validation if no dns update required

ptkach:
This should address b/257080205

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1890)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java
@@ -181,7 +181,7 @@ public final class DomainUpdateFlow implements TransactionalFlow {
     Domain newDomain = performUpdate(command, existingDomain, now);
     DomainHistory domainHistory =
         historyBuilder.setType(DOMAIN_UPDATE).setDomain(newDomain).build();
-    validateNewState(newDomain);
+    validateNewState(newDomain, existingDomain);
     if (requiresDnsUpdate(existingDomain, newDomain)) {
       dnsQueue.addDomainRefreshTask(targetId);
     }
@@ -307,9 +307,11 @@ public final class DomainUpdateFlow implements TransactionalFlow {
     }
   }
 
-  private void validateNewState(Domain newDomain) throws EppException {
+  private void validateNewState(Domain newDomain, Domain existingDomain) throws EppException {
     validateRequiredContactsPresent(newDomain.getRegistrant(), newDomain.getContacts());
-    validateDsData(newDomain.getDsData());
+    if (requiresDnsUpdate(existingDomain, newDomain)) {
+      validateDsData(newDomain.getDsData());
+    }
     validateNameserversCountForTld(
         newDomain.getTld(),
         InternetDomainName.from(newDomain.getDomainName()),


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1890)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 312 at r1](https://reviewable.io/reviews/google/nomulus/1890#-NKsVmgw1EcUrEKYLyNb:-NKsVmgw1EcUrEKYLyNc:bp911eb) ([raw file](https://github.com/google/nomulus/blob/ca1ed9adfef0f3a00bf20932ff08ef4a527be08f/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L312)):*
> ```Java
>   private void validateNewState(Domain newDomain, Domain existingDomain) throws EppException {
>     validateRequiredContactsPresent(newDomain.getRegistrant(), newDomain.getContacts());
>     if (requiresDnsUpdate(existingDomain, newDomain)) {
> ```

Could you add more details to the bug and explain how this check fixes the problem?

Is the dsdata-length check our requirement or CloudDns' requirement?

The requiresDnsUpdate method may return true on domain status changes. Should we disallow such changes? 



_Code quote:_
```Java
requiresDnsUpdate
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1890)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 312 at r1](https://reviewable.io/reviews/google/nomulus/1890#-NKsVmgw1EcUrEKYLyNb:-NKsa67M-6pD5yvHdiHY:bpfoak) ([raw file](https://github.com/google/nomulus/blob/ca1ed9adfef0f3a00bf20932ff08ef4a527be08f/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L312)):*
Sure, see the answers below:

> Could you add more details to the bug and explain how this check fixes the problem?
The idea is that unless there's any change in DS data, status (which affects DNS) or NS data no verification is needed. This follows the logic we used to have for scheduling cloudDnsTask and I figured that'd be a safe change this way.

> Is the dsdata-length check our requirement or CloudDns' requirement?
As per referenced GoDaddy issue description in b/257080205 the error thrown is ours.

> The requiresDnsUpdate method may return true on domain status changes. Should we disallow such changes?
Correct, even though this doesn't affect this particular issue, which is related to DS only data, I still think it's safer to go with status verification as well for those cases which then end up in DNS publish task.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1890)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 312 at r1](https://reviewable.io/reviews/google/nomulus/1890#-NKsVmgw1EcUrEKYLyNb:-NKxZObs3ItooUx30lDp:buv7w8k) ([raw file](https://github.com/google/nomulus/blob/ae7b613f729160f51ad5aad830b950c913b4bfe6/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L312)):*

Sure, see the answers below:

> Could you add more details to the bug and explain how this check fixes the problem?
The idea is that unless there's any change in DS data, status (which affects DNS) or NS data no verification is needed. This follows the logic we used to have for scheduling cloudDnsTask and I figured that'd be a safe change this way.

> Is the dsdata-length check our requirement or CloudDns' requirement?
As per referenced GoDaddy issue description in b/257080205 the error thrown is ours.

> The requiresDnsUpdate method may return true on domain status changes. Should we disallow such changes?
Correct, even though this doesn't affect this particular issue, which is related to DS only data, I still think it's safer to go with status verification as well for those cases which then end up in DNS publish task.
</blockquote></details>

Strictly speaking, we added the DS check because CloudDNS would reject bad DS updates (b/209037111). So yes, the error is thrown in our code, but only because we do not want to wait for CloudDNS to reject the update, which happens asynchronously. Rather we want to stop the client from being able to set a bad DS record when it updates the domain.

A corner case of using `requiresDnsUpdate` is that if a domain is changing from publishing to not publishing (e.g. being put into `CLIENT_HOLD`), the DS validation would fail if the existing DS record is bad, even though it would be fine in this case, because the update to DNS would be to wipe out anything already in the CloudDNS for that domain. In theory, the DS records already in CloudDNS should match what the domain has, so they be conforming in the first place, unless they were somehow put in CloudDNS before CloudDNS started enforcing DS validation. I don't think we need to alter the proposed behavior for such a corner case, but it's better to be aware of it.

A test is needed to verify the modified behavior.

ptkach:
It was agreed that this PR introduces inconsistency while not fixing the root issue - invalid DS data. It's not an ideal fix and rather just a symptom cover up, so we will update the error message instead to make it more straightforward.

PR #1891, opened 2023-01-04
Fix billing pipeline first month scheduling

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1891)

### DIFF
--- a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
+++ b/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
@@ -120,8 +120,8 @@ public class GenerateInvoicesAction implements Runnable {
                           .orElse(Cursor.createGlobal(RECURRING_BILLING, START_OF_TIME))
                           .getCursorTime());
 
-      if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()
-          && yearMonth.getYear() == currentCursorTime.getYear()) {
+      YearMonth expansionPeriodNextMonth = yearMonth.plusMonths(1);
+      if (expansionPeriodNextMonth.getMonthOfYear() != currentCursorTime.getMonthOfYear()) {
         throw new IllegalStateException(
             "Latest billing events expansion cycle hasn't finished yet, terminating invoicing"
                 + " pipeline");
--- a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
+++ b/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
@@ -120,8 +120,7 @@ public class GenerateInvoicesAction implements Runnable {
                           .orElse(Cursor.createGlobal(RECURRING_BILLING, START_OF_TIME))
                           .getCursorTime());
 
-      YearMonth expansionPeriodNextMonth = yearMonth.plusMonths(1);
-      if (expansionPeriodNextMonth.getMonthOfYear() != currentCursorTime.getMonthOfYear()) {
+      if (!YearMonth.fromDateFields(currentCursorTime.toDate()).isAfter(yearMonth)) {
         throw new IllegalStateException(
             "Latest billing events expansion cycle hasn't finished yet, terminating invoicing"
                 + " pipeline");
--- a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
+++ b/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java
@@ -120,7 +120,8 @@ public class GenerateInvoicesAction implements Runnable {
                           .orElse(Cursor.createGlobal(RECURRING_BILLING, START_OF_TIME))
                           .getCursorTime());
 
-      if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()) {
+      if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()
+          && yearMonth.getYear() == currentCursorTime.getYear()) {
         throw new IllegalStateException(
             "Latest billing events expansion cycle hasn't finished yet, terminating invoicing"
                 + " pipeline");
--- a/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java
+++ b/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java
@@ -168,4 +168,28 @@ class GenerateInvoicesActionTest extends BeamActionTestBase {
                 + " terminating invoicing pipeline");
     cloudTasksHelper.assertNoTasksEnqueued("beam-reporting");
   }
+
+  @Test
+  void testSucceedsToGenerateInvoicesFirstDayOfTheYear() throws Exception {
+    persistResource(Cursor.createGlobal(RECURRING_BILLING, DateTime.parse("2017-01-01TZ")));
+    action =
+        new GenerateInvoicesAction(
+            "test-project",
+            "test-region",
+            "staging_bucket",
+            "billing_bucket",
+            "REG-INV",
+            false,
+            new YearMonth(2016, 12),
+            emailUtils,
+            cloudTasksUtils,
+            clock,
+            response,
+            dataflow);
+    action.run();
+    assertThat(response.getContentType()).isEqualTo(MediaType.PLAIN_TEXT_UTF_8);
+    assertThat(response.getStatus()).isEqualTo(SC_OK);
+    assertThat(response.getPayload()).isEqualTo("Launched invoicing pipeline: jobid");
+    cloudTasksHelper.assertNoTasksEnqueued("beam-reporting");
+  }
 }
--- a/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java
+++ b/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java
@@ -171,7 +171,7 @@ class GenerateInvoicesActionTest extends BeamActionTestBase {
 
   @Test
   void testSucceedsToGenerateInvoicesFirstDayOfTheYear() throws Exception {
-    persistResource(Cursor.createGlobal(RECURRING_BILLING, DateTime.parse("2017-01-01TZ")));
+    persistResource(Cursor.createGlobal(RECURRING_BILLING, DateTime.parse("2017-01-01T13:15:00Z")));
     action =
         new GenerateInvoicesAction(
             "test-project",


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 123 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKxMt8X0VKYglwBSBgv:-NKxMt8X0VKYglwBSBgw:b-c1n9sb) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L123)):*
> ```Java
>                           .getCursorTime());
> 
>       if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()
> ```

Instead of using `currentCursorTime` (which to a naive reader doesn't mean anything), we should have a `DateTime lastRecurringExpansionTime = currentCursorTime.minusDays(1)` and just compare it the `yearMonth` (some conversion necessary between `DateTime` and `YearMonth`) that was provided. This way we can avoid any potential pitfalls like this, which should all be handled by Joda natively.

Also, a test should be added.



_Code quote:_
```Java
      if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()
          && yearMonth.getYear() == currentCursorTime.getYear()) {
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 123 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKxMt8X0VKYglwBSBgv:-NKxat7oA4XlSLk8QaMi:b-7h6bwi) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L123)):*

Instead of using `currentCursorTime` (which to a naive reader doesn't mean anything), we should have a `DateTime lastRecurringExpansionTime = currentCursorTime.minusDays(1)` and just compare it the `yearMonth` (some conversion necessary between `DateTime` and `YearMonth`) that was provided. This way we can avoid any potential pitfalls like this, which should all be handled by Joda natively.

Also, a test should be added.
</blockquote></details>

Can you please elaborate on your request to introduce `DateTime lastRecurringExpansionTime = currentCursorTime.minusDays(1)` ? Yes it will make naming more straightforward, however it will also make comparison more difficult because the expansion pipeline running successfully on a first day will produce `lastRecurringExpansionTime` in previous month, while the pipeline failing on first day and succeeding on a second day will produce the `lastRecurringExpansionTime` in current month, thus making yearMonth comparison more difficult. 

I will update an existing test

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 123 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKxMt8X0VKYglwBSBgv:-NKxdumVAsRRgaIE9KUg:b5o2h8k) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L123)):*

Can you please elaborate on your request to introduce `DateTime lastRecurringExpansionTime = currentCursorTime.minusDays(1)` ? Yes it will make naming more straightforward, however it will also make comparison more difficult because the expansion pipeline running successfully on a first day will produce `lastRecurringExpansionTime` in previous month, while the pipeline failing on first day and succeeding on a second day will produce the `lastRecurringExpansionTime` in current month, thus making yearMonth comparison more difficult. 

I will update an existing test
</blockquote></details>

The cursor indicates the next `DateTime` the expansion action should run for. It's not possible for the pipeline to fail on the first day and succeed on the second day because the cursor would be stuck on the first day. It would need to finish the first day, before moving on to processing the second day. In another word, the expansion would only run up to the cursor time, which is only updated to the next day after the expansion succeeds for the current cursor time.

So I think a more accurate comparison is to take the `yearTime`, plus 1 month, and get the start `DateTime` of that month, which would be the exclusive upper bound of the period for which the invoicing pipeline runs for, and compare it to the cursor, which is the exclusive upper bound of the period for which recurring billing events has expanded to. As long as the former is not after the latter, we can generate invoices.

Note that in #1881, we will change the behavior of recurrence expansion so that it happens when the event time becomes in scope, instead of when the billing time is in scope. In practice, it means the renewals would be generated 45 days before the billing time. That should make this check moot because we'd be 45 days behind in generating renewals, which hopefully would have been noticed long before the invoicing pipeline runs.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 124 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKy3rL44QMpJgqa2dMa:-NKy3rL44QMpJgqa2dMb:buruvqq) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L124)):*
> ```Java
> 
>       if (yearMonth.getMonthOfYear() >= currentCursorTime.getMonthOfYear()
>           && yearMonth.getYear() == currentCursorTime.getYear()) {
> ```

This seems to be the wrong way to do it?  Why not just use `isBefore`/`isAfter`?  https://docs.oracle.com/javase/8/docs/api/java/time/YearMonth.html#isBefore-java.time.YearMonth-

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 124 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKy3rL44QMpJgqa2dMa:-NKy3yb--B7T5Wbz50TM:b6york1) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L124)):*

This seems to be the wrong way to do it?  Why not just use `isBefore`/`isAfter`?  https://docs.oracle.com/javase/8/docs/api/java/time/YearMonth.html#isBefore-java.time.YearMonth-
</blockquote></details>

In general, any kind of time-based comparison code should always already be written and provided by the library; if you're handrolling these methods yourself that means you're probably missing something.

Also that link above is to the wrong library (Java's built-in instead of Joda Time), but Joda has the same methods that work the same way.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 123 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKxMt8X0VKYglwBSBgv:-NKy8SE-91NdixsFjhA3:b-18ehip) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L123)):*

The cursor indicates the next `DateTime` the expansion action should run for. It's not possible for the pipeline to fail on the first day and succeed on the second day because the cursor would be stuck on the first day. It would need to finish the first day, before moving on to processing the second day. In another word, the expansion would only run up to the cursor time, which is only updated to the next day after the expansion succeeds for the current cursor time.

So I think a more accurate comparison is to take the `yearTime`, plus 1 month, and get the start `DateTime` of that month, which would be the exclusive upper bound of the period for which the invoicing pipeline runs for, and compare it to the cursor, which is the exclusive upper bound of the period for which recurring billing events has expanded to. As long as the former is not after the latter, we can generate invoices.

Note that in #1881, we will change the behavior of recurrence expansion so that it happens when the event time becomes in scope, instead of when the billing time is in scope. In practice, it means the renewals would be generated 45 days before the billing time. That should make this check moot because we'd be 45 days behind in generating renewals, which hopefully would have been noticed long before the invoicing pipeline runs.
</blockquote></details>

"get the start DateTime of that month" seems unnecessary, JODA provides `plusMonths()` which takes care of convertion and makes it easy to compare
___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 124 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKy3rL44QMpJgqa2dMa:-NKy8leE0spm5rQazook:bjlz2ys) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L124)):*

In general, any kind of time-based comparison code should always already be written and provided by the library; if you're handrolling these methods yourself that means you're probably missing something.

Also that link above is to the wrong library (Java's built-in instead of Joda Time), but Joda has the same methods that work the same way.
</blockquote></details>

YearMonth has to be converted to ReadablePartial before compering, it's not efficient and bulky. I used JODA's plusMonth which takes care of corner cases like this.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 123 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKxMt8X0VKYglwBSBgv:-NL-Mi_SDork1R_YEtWJ:b-l93lg7) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L123)):*

"get the start DateTime of that month" seems unnecessary, JODA provides `plusMonths()` which takes care of convertion and makes it easy to compare
</blockquote></details>

By changing '>=' to '!=' you lose the ability to run the action retroactively to backfill previous invoices. Further more, it's a clever trick to make things work, but it obfuscates the code even more. You never know what other edge cases might bite you next time. Ben's comment also mentioned using Joda for comparison instead of hand-rolling it, because the library is battle tested and much more likely to handle every conceivable corner cases correctly. Some conversion is necessary but it's worth doing.

You are right that we can just compare the `YearMonth` and not care about the day/time part, as the invoice period covers up to the midnight of the 1st day of next month (exclusive), and any cursor time that is in the next month would be after that instant. So you just need to convert the cursor time from `DateTime` to `YearMonth` and call `DateTime.isAfter`.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys, @jianglai, and @weiminyu)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 123 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKxMt8X0VKYglwBSBgv:-NL1WuCFBm8WWqAOjNI-:b-bu91xm) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L123)):*

By changing '>=' to '!=' you lose the ability to run the action retroactively to backfill previous invoices. Further more, it's a clever trick to make things work, but it obfuscates the code even more. You never know what other edge cases might bite you next time. Ben's comment also mentioned using Joda for comparison instead of hand-rolling it, because the library is battle tested and much more likely to handle every conceivable corner cases correctly. Some conversion is necessary but it's worth doing.

You are right that we can just compare the `YearMonth` and not care about the day/time part, as the invoice period covers up to the midnight of the 1st day of next month (exclusive), and any cursor time that is in the next month would be after that instant. So you just need to convert the cursor time from `DateTime` to `YearMonth` and call `DateTime.isAfter`.
</blockquote></details>

Updated

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 124 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKy3rL44QMpJgqa2dMa:-NL1gc1W-cYi2fk4Cye-:b-hh37rm) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L124)):*

YearMonth has to be converted to ReadablePartial before compering, it's not efficient and bulky. I used JODA's plusMonth which takes care of corner cases like this.
</blockquote></details>

YearMonth implements ReadableParitial. The following can compile and doesn't seem too bad:

`if (!yearMonth.isBefore(new YearMonth(currentCursorTime.getYear(), currentCursorTime.getMonthOfYear())))`

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java` line 124 at r1](https://reviewable.io/reviews/google/nomulus/1891#-NKy3rL44QMpJgqa2dMa:-NL1qtrv21z9lW1ag61y:b-gmsj95) ([raw file](https://github.com/google/nomulus/blob/0fa06e60cea2b3d2ac614f24e4c1c1ef9ee9204a/core/src/main/java/google/registry/reporting/billing/GenerateInvoicesAction.java#L124)):*

YearMonth implements ReadableParitial. The following can compile and doesn't seem too bad:

`if (!yearMonth.isBefore(new YearMonth(currentCursorTime.getYear(), currentCursorTime.getMonthOfYear())))`
</blockquote></details>

Thanks @weiminyu . I have already updated it this morning to rely on Joda comparison

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

jianglai (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java` line 174 at r4](https://reviewable.io/reviews/google/nomulus/1891#-NL2okc3D3MOOR0MYkAC:-NL2okc3D3MOOR0MYkAD:blr0cpq) ([raw file](https://github.com/google/nomulus/blob/4983289cbf06d0cf7c6856b59dc3e3bb9a32b22b/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java#L174)):*
> ```Java
>   @Test
>   void testSucceedsToGenerateInvoicesFirstDayOfTheYear() throws Exception {
>     persistResource(Cursor.createGlobal(RECURRING_BILLING, DateTime.parse("2017-01-01TZ")));
> ```

Let's not use midnight here, use something after midnight to make sure we are not hitting some edge case that only works for midnight.



_Code quote:_
```Java
2017-01-01TZ
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1891)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java` line 174 at r4](https://reviewable.io/reviews/google/nomulus/1891#-NL2okc3D3MOOR0MYkAC:-NL329At8POdMwIrpSbh:bvnrpcm) ([raw file](https://github.com/google/nomulus/blob/4983289cbf06d0cf7c6856b59dc3e3bb9a32b22b/core/src/test/java/google/registry/reporting/billing/GenerateInvoicesActionTest.java#L174)):*

Let's not use midnight here, use something after midnight to make sure we are not hitting some edge case that only works for midnight.
</blockquote></details>

updated

PR #1904, opened 2023-01-11
Modify DomainCreateFlow to check for an applicable defaultPromoToken

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1904)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -15,6 +15,7 @@
 package google.registry.flows.domain;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.flows.FlowUtils.persistEntityChanges;
 import static google.registry.flows.FlowUtils.validateRegistrarIsLoggedIn;
@@ -45,9 +46,9 @@ import static google.registry.flows.domain.DomainFlowUtils.verifyRegistrarIsActi
 import static google.registry.flows.domain.DomainFlowUtils.verifyUnitIsYears;
 import static google.registry.model.EppResourceUtils.createDomainRepoId;
 import static google.registry.model.IdService.allocateId;
+import static google.registry.model.domain.token.AllocationToken.ALLOCATION_TOKENS_CACHE;
 import static google.registry.model.eppcommon.StatusValue.SERVER_HOLD;
 import static google.registry.model.reporting.HistoryEntry.Type.DOMAIN_CREATE;
-import static google.registry.model.tld.Registry.DEFAULT_PROMO_TOKENS_CACHE;
 import static google.registry.model.tld.Registry.TldState.GENERAL_AVAILABILITY;
 import static google.registry.model.tld.Registry.TldState.QUIET_PERIOD;
 import static google.registry.model.tld.Registry.TldState.START_DATE_SUNRISE;
@@ -120,7 +121,9 @@ import google.registry.model.tld.Registry.TldType;
 import google.registry.model.tld.label.ReservationType;
 import google.registry.model.tmch.ClaimsList;
 import google.registry.model.tmch.ClaimsListDao;
+import google.registry.persistence.VKey;
 import google.registry.tmch.LordnTaskUtils;
+import java.util.Map;
 import java.util.Optional;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -274,23 +277,7 @@ public final class DomainCreateFlow implements TransactionalFlow {
             now,
             eppInput.getSingleExtension(AllocationTokenExtension.class));
     if (!allocationToken.isPresent() && !registry.getDefaultPromoTokens().isEmpty()) {
-      ImmutableList<Optional<AllocationToken>> tokenList = DEFAULT_PROMO_TOKENS_CACHE.get(registry);
-      checkArgument(!isNullOrEmpty(tokenList), "Failed loading defaultPromoTokens from database");
-      // Check if any of the tokens are valid for this domain registration
-      for (Optional<AllocationToken> token : tokenList) {
-        if (!token.isPresent()) {
-          continue;
-        }
-        try {
-          AllocationTokenFlowUtils.validateToken(
-              InternetDomainName.from(command.getDomainName()), token.get(), registrarId, now);
-        } catch (AssociationProhibitsOperationException e) {
-          continue;
-        }
-        // Only use the first valid token in the list
-        allocationToken = token;
-        break;
-      }
+      allocationToken = checkForDefaultToken(registry, command);
     }
     boolean isAnchorTenant =
         isAnchorTenant(
@@ -456,6 +443,31 @@ public final class DomainCreateFlow implements TransactionalFlow {
         .build();
   }
 
+  private Optional<AllocationToken> checkForDefaultToken(
+      Registry registry, DomainCommand.Create command) throws EppException {
+    Map<VKey<AllocationToken>, Optional<AllocationToken>> tokenMap =
+        ALLOCATION_TOKENS_CACHE.getAll(registry.getDefaultPromoTokens());
+    checkState(
+        !isNullOrEmpty(tokenMap), "Failure while loading default TLD promotions from the database");
+    // Check if any of the tokens are valid for this domain registration
+    for (Optional<AllocationToken> token : tokenMap.values()) {
+      if (!token.isPresent()) {
+        continue;
+      }
+      try {
+        AllocationTokenFlowUtils.validateToken(
+            InternetDomainName.from(command.getDomainName()),
+            token.get(),
+            registrarId,
+            tm().getTransactionTime());
+      } catch (AssociationProhibitsOperationException e) {
+        continue;
+      }
+      // Only use the first valid token in the list
+      return token;
+    }
+    return Optional.empty();
+  }
   /**
    * Verifies that signed marks are only sent during sunrise.
    *
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -121,7 +121,9 @@ import google.registry.model.tld.Registry.TldType;
 import google.registry.model.tld.label.ReservationType;
 import google.registry.model.tmch.ClaimsList;
 import google.registry.model.tmch.ClaimsListDao;
+import google.registry.persistence.VKey;
 import google.registry.tmch.LordnTaskUtils;
+import java.util.Map;
 import java.util.Optional;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -443,9 +445,11 @@ public final class DomainCreateFlow implements TransactionalFlow {
 
   private Optional<AllocationToken> checkForDefaultToken(
       Registry registry, DomainCommand.Create command) throws EppException {
+    Map<VKey<AllocationToken>, Optional<AllocationToken>> tokens =
+        AllocationToken.getAll(registry.getDefaultPromoTokens());
     ImmutableList<Optional<AllocationToken>> tokenList =
         registry.getDefaultPromoTokens().stream()
-            .map(key -> AllocationToken.get(key))
+            .map(key -> tokens.get(key))
             .collect(toImmutableList());
     checkState(
         !isNullOrEmpty(tokenList),
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -274,18 +274,21 @@ public final class DomainCreateFlow implements TransactionalFlow {
             now,
             eppInput.getSingleExtension(AllocationTokenExtension.class));
     if (!allocationToken.isPresent() && !registry.getDefaultPromoTokens().isEmpty()) {
-      ImmutableList<AllocationToken> tokenList = DEFAULT_PROMO_TOKENS_CACHE.get(registry);
+      ImmutableList<Optional<AllocationToken>> tokenList = DEFAULT_PROMO_TOKENS_CACHE.get(registry);
       checkArgument(!isNullOrEmpty(tokenList), "Failed loading defaultPromoTokens from database");
       // Check if any of the tokens are valid for this domain registration
-      for (AllocationToken token : tokenList) {
+      for (Optional<AllocationToken> token : tokenList) {
+        if (!token.isPresent()) {
+          continue;
+        }
         try {
           AllocationTokenFlowUtils.validateToken(
-              InternetDomainName.from(command.getDomainName()), token, registrarId, now);
+              InternetDomainName.from(command.getDomainName()), token.get(), registrarId, now);
         } catch (AssociationProhibitsOperationException e) {
           continue;
         }
         // Only use the first valid token in the list
-        allocationToken = Optional.of(token);
+        allocationToken = token;
         break;
       }
     }
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -449,16 +449,14 @@ public final class DomainCreateFlow implements TransactionalFlow {
         AllocationToken.getAll(registry.getDefaultPromoTokens());
     ImmutableList<Optional<AllocationToken>> tokenList =
         registry.getDefaultPromoTokens().stream()
-            .map(key -> tokens.get(key))
+            .map(tokens::get)
+            .filter(Optional::isPresent)
             .collect(toImmutableList());
     checkState(
         !isNullOrEmpty(tokenList),
         "Failure while loading default TLD promotions from the database");
     // Check if any of the tokens are valid for this domain registration
     for (Optional<AllocationToken> token : tokenList) {
-      if (!token.isPresent()) {
-        continue;
-      }
       try {
         AllocationTokenFlowUtils.validateToken(
             InternetDomainName.from(command.getDomainName()),
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -16,6 +16,7 @@ package google.registry.flows.domain;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.flows.FlowUtils.persistEntityChanges;
 import static google.registry.flows.FlowUtils.validateRegistrarIsLoggedIn;
@@ -46,7 +47,6 @@ import static google.registry.flows.domain.DomainFlowUtils.verifyRegistrarIsActi
 import static google.registry.flows.domain.DomainFlowUtils.verifyUnitIsYears;
 import static google.registry.model.EppResourceUtils.createDomainRepoId;
 import static google.registry.model.IdService.allocateId;
-import static google.registry.model.domain.token.AllocationToken.ALLOCATION_TOKENS_CACHE;
 import static google.registry.model.eppcommon.StatusValue.SERVER_HOLD;
 import static google.registry.model.reporting.HistoryEntry.Type.DOMAIN_CREATE;
 import static google.registry.model.tld.Registry.TldState.GENERAL_AVAILABILITY;
@@ -121,9 +121,7 @@ import google.registry.model.tld.Registry.TldType;
 import google.registry.model.tld.label.ReservationType;
 import google.registry.model.tmch.ClaimsList;
 import google.registry.model.tmch.ClaimsListDao;
-import google.registry.persistence.VKey;
 import google.registry.tmch.LordnTaskUtils;
-import java.util.Map;
 import java.util.Optional;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -445,12 +443,15 @@ public final class DomainCreateFlow implements TransactionalFlow {
 
   private Optional<AllocationToken> checkForDefaultToken(
       Registry registry, DomainCommand.Create command) throws EppException {
-    Map<VKey<AllocationToken>, Optional<AllocationToken>> tokenMap =
-        ALLOCATION_TOKENS_CACHE.getAll(registry.getDefaultPromoTokens());
+    ImmutableList<Optional<AllocationToken>> tokenList =
+        registry.getDefaultPromoTokens().stream()
+            .map(key -> AllocationToken.get(key))
+            .collect(toImmutableList());
     checkState(
-        !isNullOrEmpty(tokenMap), "Failure while loading default TLD promotions from the database");
+        !isNullOrEmpty(tokenList),
+        "Failure while loading default TLD promotions from the database");
     // Check if any of the tokens are valid for this domain registration
-    for (Optional<AllocationToken> token : tokenMap.values()) {
+    for (Optional<AllocationToken> token : tokenList) {
       if (!token.isPresent()) {
         continue;
       }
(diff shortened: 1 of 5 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 276 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAMfgS-5a5myUihqfs:-NMAMfgS-5a5myUihqft:bccmqx6) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L276)):*
> ```Java
>             now,
>             eppInput.getSingleExtension(AllocationTokenExtension.class));
>     if (!allocationToken.isPresent() && !registry.getDefaultPromoTokens().isEmpty()) {
> ```

Probably worth splitting this all off into a helper method.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 278 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAN3gQA-XUmlKidc1L:-NMAN3gQA-XUmlKidc1M:buki6hl) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L278)):*
> ```Java
>     if (!allocationToken.isPresent() && !registry.getDefaultPromoTokens().isEmpty()) {
>       ImmutableList<Optional<AllocationToken>> tokenList = DEFAULT_PROMO_TOKENS_CACHE.get(registry);
>       checkArgument(!isNullOrEmpty(tokenList), "Failed loading defaultPromoTokens from database");
> ```

`checkArgument()` isn't exactly right here; maybe something like `checkState()`?  Also, this error message might be registrar-visible because it's being thrown in an EPP flow, so the wording might need to be adjusted to be more externally understandable.
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 218 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAM9eH6jsVOS_sl_pB:-NMAM9eH6jsVOS_sl_pC:b-mtzmqk) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/model/tld/Registry.java#L218)):*
> ```Java
> 
>   /** A cache that loads the default promo tokens for a given {@link Registry}. */
>   public static final LoadingCache<Registry, ImmutableList<Optional<AllocationToken>>>
> ```

The same tokens are likely to be used across multiple TLDs, so it's better to cache just the tokens.

Something like:

```
LoadingCache<VKey<AllocationToken>,  Optional<AllocationToken>>
```

or

```
LoadingCache<String,  Optional<AllocationToken>>
```

where the String is the allocation token ID.

You'll also then need to implement the `loadAll()` method so that multiple tokens can be loaded efficiently in a batch, which would look similar to this existing `load()` method.  (We have examples of this in our codebase.)
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 1629 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAOEuyD_inIgNihVAV:-NMAOEuz83_-hLv6Vrk5:b-lqptiv) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L1629)):*
> ```Java
>     AllocationToken defaultToken2 =
>         persistResource(
>             new AllocationToken.Builder()
> ```

We'll also need a test that the token's discount is actually applied (and shows the correct price through the fee extension).

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 218 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAM9eH6jsVOS_sl_pB:-NMAalEb4eNaC5fICbpA:bk4tujj) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/model/tld/Registry.java#L218)):*

The same tokens are likely to be used across multiple TLDs, so it's better to cache just the tokens.

Something like:

```
LoadingCache<VKey<AllocationToken>,  Optional<AllocationToken>>
```

or

```
LoadingCache<String,  Optional<AllocationToken>>
```

where the String is the allocation token ID.

You'll also then need to implement the `loadAll()` method so that multiple tokens can be loaded efficiently in a batch, which would look similar to this existing `load()` method.  (We have examples of this in our codebase.)
</blockquote></details>

Yeah I think this also would solve the issue that I'm having where we have that small disconnect between "checking the actual registry object as it currently exists for the tokens" vs "checking a cache that may or may not be up to date". I wasn't sure what the best way of reconciling that would be, but switching the cache key would definitely help. 
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 228 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NM9q8mU2mWilwLDUk_N:-NM9q8mU2mWilwLDUk_O:b-l8fyn) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/model/tld/Registry.java#L228)):*
> ```Java
>                           registry.getDefaultPromoTokens();
>                       return tokenKeys.stream()
>                           .map(key -> tm().transact(() -> tm().loadByKeyIfPresent(key)))
> ```

better to do this all in one transaction

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 276 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAMfgS-5a5myUihqfs:-NMG5jvg3x6KWcWkTPUX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L276)):*

Probably worth splitting this all off into a helper method.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 278 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAN3gQA-XUmlKidc1L:-NMFsBke4qfWfvNRebp-:b-xs40lo) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L278)):*

`checkArgument()` isn't exactly right here; maybe something like `checkState()`?  Also, this error message might be registrar-visible because it's being thrown in an EPP flow, so the wording might need to be adjusted to be more externally understandable.
</blockquote></details>

Changed to "Failure while loading default TLD promotions from the database". Is this alright?
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 1629 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAOEuyD_inIgNihVAV:-NMGBmax5yaWZZUUmGTp:b5glwx5) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L1629)):*

We'll also need a test that the token's discount is actually applied (and shows the correct price through the fee extension).
</blockquote></details>

Added a test to make sure the correct discount is applied. Still waiting to hear back from Bruno on how to handle mismatch on fee extension in the create request and will handle that in a follow up PR. 
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 218 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAM9eH6jsVOS_sl_pB:-NMVaxS3-vbyJBkwlbsi:bevxf7s) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/model/tld/Registry.java#L218)):*

Yeah I think this also would solve the issue that I'm having where we have that small disconnect between "checking the actual registry object as it currently exists for the tokens" vs "checking a cache that may or may not be up to date". I wasn't sure what the best way of reconciling that would be, but switching the cache key would definitely help. 
</blockquote></details>

Done. Moved the cahce definition to the Allocation Token class since it makes more sense there now
___
*[`core/src/main/java/google/registry/model/tld/Registry.java` line 228 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NM9q8mU2mWilwLDUk_N:-NMVb6io-ghbk6aUXo5J:b-896fix) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/model/tld/Registry.java#L228)):*

better to do this all in one transaction 
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 1 of 4 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 4 of 5 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 448 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZsPvnAESR1JV2aWQ4:-NMZsPvnAESR1JV2aWQ5:b2wif6t) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L448)):*
> ```Java
>   private Optional<AllocationToken> checkForDefaultToken(
>       Registry registry, DomainCommand.Create command) throws EppException {
>     Map<VKey<AllocationToken>, Optional<AllocationToken>> tokenMap =
> ```

shouldn't the tokens all exist? Why might they be Optional.empty? 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 449 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZr6Ji8VMgwK3Kf7Rz:-NMZr6Jj5esFLl3R-V4T:b-ahltis) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L449)):*
> ```Java
>       Registry registry, DomainCommand.Create command) throws EppException {
>     Map<VKey<AllocationToken>, Optional<AllocationToken>> tokenMap =
>         ALLOCATION_TOKENS_CACHE.getAll(registry.getDefaultPromoTokens());
> ```

instead of accessing the cache directly here, let's make a method in `AllocationToken` that loads from the cache, similar to how we have `Registry::get` to abstract it away 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 453 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZs-8iBFVDtQ0dYvNK:-NMZs-8iBFVDtQ0dYvNL:b-zg23mf) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L453)):*
> ```Java
>         !isNullOrEmpty(tokenMap), "Failure while loading default TLD promotions from the database");
>     // Check if any of the tokens are valid for this domain registration
>     for (Optional<AllocationToken> token : tokenMap.values()) {
> ```

q: didn't we discuss something about how "if multiple tokens are valid for a given domain create request, we will use the first one"?

My concern is that by converting into a `Map<VKey< >, Optional<AT>>` we can lose that ordering. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 463 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZrpwUED4AvRSWfAgy:-NMZrpwUED4AvRSWfAgz:b-ibwlyw) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L463)):*
> ```Java
>             registrarId,
>             tm().getTransactionTime());
>       } catch (AssociationProhibitsOperationException e) {
> ```

should add a comment explaining why it's ok to just swallow this exception 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 290 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZs_9W9CA5NHXyKXd6:-NMZs_9W9CA5NHXyKXd7:brr1the) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L290)):*
> ```Java
>                     @Override
>                     public Map<VKey<AllocationToken>, Optional<AllocationToken>> loadAll(
>                         Iterable<? extends VKey<AllocationToken>> keys) {
> ```

not `loadByKeysIfPresent`?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 2 of 5 files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 448 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZsPvnAESR1JV2aWQ4:-NM_jfMC-0dsvg5XPMC-:bzazog8) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L448)):*

shouldn't the tokens all exist? Why might they be Optional.empty? 
</blockquote></details>

I don't think it's safe to assume that all VKeys will exist in the database, there could be a case where an allocation token was deleted from the database but the VKey for it is still saved on a TLD. Also, loadByKeyIfPresent returns an optional.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 449 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZr6Ji8VMgwK3Kf7Rz:-NM_lMDU4xDemrg3DwRX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L449)):*

instead of accessing the cache directly here, let's make a method in `AllocationToken` that loads from the cache, similar to how we have `Registry::get` to abstract it away 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 453 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZs-8iBFVDtQ0dYvNK:-NM_jKp90v0ZdRY6JSJ_:bizws49) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L453)):*

q: didn't we discuss something about how "if multiple tokens are valid for a given domain create request, we will use the first one"?

My concern is that by converting into a `Map<VKey< >, Optional<AT>>` we can lose that ordering. 
</blockquote></details>

oh, good point. Changed this to just use the get() and not the getAll() so we can save to a list. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 463 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZrpwUED4AvRSWfAgy:-NMeA59_AoD1AUph-ZjV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L463)):*

should add a comment explaining why it's ok to just swallow this exception 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 290 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZs_9W9CA5NHXyKXd6:-NMeArYKCfXBKGLdOXGi:b-o4aa4w) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L290)):*

not `loadByKeysIfPresent`? 
</blockquote></details>

nope, this does not work because for the loading cache we need it to return a Map of VKey to Optional<AllocationToken>,  but `loadByKeysIfPresent` only return a Map of Vkey to AllocationToken of the tokens that are present

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 4 of 5 files reviewed, 7 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 448 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZsPvnAESR1JV2aWQ4:-NMf-zpoDrtEa90DUAJd:b-cqoa7i) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L448)):*

I don't think it's safe to assume that all VKeys will exist in the database, there could be a case where an allocation token was deleted from the database but the VKey for it is still saved on a TLD. Also, loadByKeyIfPresent returns an optional.
</blockquote></details>

right yeah, i mean I feel like if the tokens aren't present that should be a Very Bad Thing. 

Also, are the tokens on the TLD object foreign-keyed to the AllocationToken table? If not, they probably should be, and if so, that wouldn't be possible anyway 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 448 at r3](https://reviewable.io/reviews/google/nomulus/1904#-NMf0Vrz5_0fgv0O8f1J:-NMf0Vrz5_0fgv0O8f1K:b-sb0ti9) ([raw file](https://github.com/google/nomulus/blob/2f71d4df89a2571ec3f08330c1d9622b07871e31/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L448)):*
> ```Java
>     ImmutableList<Optional<AllocationToken>> tokenList =
>         registry.getDefaultPromoTokens().stream()
>             .map(key -> AllocationToken.get(key))
> ```

we can probably use AllocationToken::getAll here if we get the map then iterate over the VKeys, mapping each to the token object in order 
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 290 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZs_9W9CA5NHXyKXd6:-NMf1DGLAgkRUYf-Uilu:bvhkjkf) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L290)):*

nope, this does not work because for the loading cache we need it to return a Map of VKey to Optional<AllocationToken>,  but `loadByKeysIfPresent` only return a Map of Vkey to AllocationToken of the tokens that are present
</blockquote></details>

see other comments -- we probably should be assuming that all of the keys exist

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 278 at r1](https://reviewable.io/reviews/google/nomulus/1904#-NMAN3gQA-XUmlKidc1L:-NMjcFz-9lqPxfw45dUm:bk0433) ([raw file](https://github.com/google/nomulus/blob/44e5b37630a593280850cf34e8b7801b647720be/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L278)):*

Changed to "Failure while loading default TLD promotions from the database". Is this alright?
</blockquote></details>

Yeah that's fine.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 448 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZsPvnAESR1JV2aWQ4:-NMk9lgn5ZsmOlgWKSyT:b-1dxsz7) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L448)):*

right yeah, i mean I feel like if the tokens aren't present that should be a Very Bad Thing. 

Also, are the tokens on the TLD object foreign-keyed to the AllocationToken table? If not, they probably should be, and if so, that wouldn't be possible anyway 
</blockquote></details>

They aren't foreign keyed since it's just a list of VKeys that's stored on the TLD
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 448 at r3](https://reviewable.io/reviews/google/nomulus/1904#-NMf0Vrz5_0fgv0O8f1J:-NMkGX6G3-WOLsOUzw2R:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2f71d4df89a2571ec3f08330c1d9622b07871e31/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L448)):*

we can probably use AllocationToken::getAll here if we get the map then iterate over the VKeys, mapping each to the token object in order 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 290 at r2](https://reviewable.io/reviews/google/nomulus/1904#-NMZs_9W9CA5NHXyKXd6:-NMkGpd93X5B4FVF4Eh2:b-ikqpfu) ([raw file](https://github.com/google/nomulus/blob/f372037de1493e686e53d9d70a91752a46de843d/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L290)):*

see other comments -- we probably should be assuming that all of the keys exist 
</blockquote></details>

I disagree, we cannot assume that all of the tokens exist and we would not want to fail a domain create because checking for a default token on the TLD fails. Especially since if somehow a token were to get deleted that were saved in the VKey list on a TLD, that would cause all creates on that TLD to fail, not just the ones that would have used that default token.

gbrodman:
reviewable isn't working so:
- yeah I suppose it's not possible in Postgres to have an array value where each member of the array is foreign-keyed on to another table (in this case we'd want the array of VKeys to be foreign-keyed onto the allocation token table). 

- I don't know that it should necessarily be in this PR, but can we have a check in DeleteAllocationTokenCommand to make sure that we're not deleting any TLD-default tokens when we delete tokens? Like, manually perform the check(s) that a foreign key would otherwise perform? 

I agree that it would be odd to fail all creates if we can't load the tokens. It does seem odd that we'd silently skip them, but given that we very rarely delete tokens anyway it should probably be fine.

gbrodman (commented):
> Reviewed 1 of 4 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

sarahcaseybot (commented):
DeleteAllocationTokenCommand only deletes SINGLE_USE tokens, so this should not be an issue.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904#-:-NN3RWVz0IOQAGwF5s50:bpuj2pj)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 458 at r4](https://reviewable.io/reviews/google/nomulus/1904#-NLY52BBPihcK65RjHx7-r4-458:-NN3QOou24yCVwCihtlv:b-896fix) ([raw file](https://github.com/google/nomulus/blob/26f6d9aea7df5a697886b4ef2ffd52fa97d94ce7/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L458)):*

if we're filtering out empties anyway, we can just do a "filter" on Optional::isPresent in the stream above on line ~451 
</blockquote></details>

Done.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1904)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #1907, opened 2023-01-17
Update ExpandRecurringBillingEventsAction to use the beam pipeline

jianglai:
Due to the way the beam pipeline is designed, it will expand an
recurring billing event when its event time is in scope for expansion,
instead of billing time. This means that the one time will be generated
45 days earlier. This would negate the need to check if the expansion is
finished when generating monthly invoices.

We will need to backfill the past 45 days of onetimes before the new
code is deployed. As an illustration, with the old code, a cursor time
of 2023-01-17 means that all auto-renewals whose billing time is before
2023-01-17 were created, which corresponds to an effective cursor time
of 2022-12-03 (45 days before 2023-01-17) for event time. This cursor
will need to be brought to 2023-01-17 to ensure that there is no gap in
generated event times when switching to use the new code.

TESTED=deployed to alpha and verified that the beam job is launched.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1907)

### DIFF
--- a/core/src/main/java/google/registry/batch/BatchModule.java
+++ b/core/src/main/java/google/registry/batch/BatchModule.java
@@ -105,10 +105,22 @@ public class BatchModule {
   }
 
   @Provides
-  @Parameter(ExpandRecurringBillingEventsAction.PARAM_CURSOR_TIME)
-  static Optional<DateTime> provideCursorTime(HttpServletRequest req) {
+  @Parameter(ExpandRecurringBillingEventsAction.PARAM_START_TIME)
+  static Optional<DateTime> provideStartTime(HttpServletRequest req) {
     return extractOptionalDatetimeParameter(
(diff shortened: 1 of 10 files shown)


jianglai (commented):
> Reviewed 10 of 11 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1907)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (approved):
> Reviewed 8 of 11 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1907)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1910, opened 2023-01-18
Remove app engine deps from Lock

ptkach:
Flyway change - https://github.com/google/nomulus/pull/1911

The reasoning behind RequestStatusChecker removal is in it's heavy dependency on AppEngine RequestLogs to verify request `isFinish` indicator. Unfortunately with our GKE migration RequestLogs won't be available anyway. I've not discovered an alternative to track request abrupt exit that would work on GAE and GKE and after a consultation with @jianglai  it was agreed that we can just remove this request indicator and let the lock expire. 

Tested on alpha all actions which invoke [executeWithLocks](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/lock/LockHandlerImpl.java;l=71;drc=0d3c0f7b76937a7083be94e9449c93947c1673f7;bpv=1;bpt=1) - `/_dr/task/icannReportingUpload`, `/_dr/task/syncRegistrarsSheet`, `/_dr/task/publishDnsUpdates`, `/_dr/task/deleteExpiredDomains`, `/_dr/task/rdeUpload` - no errors related to my change

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1910)


ptkach (commented):
done

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910#-:-NMAuZ7RDkjp3awZs4l8:bb7f2mf)* status: 7 of 10 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/model/server/Lock.java` line 159 at r1](https://reviewable.io/reviews/google/nomulus/1910#-NM9jmEy8RPmZcQf2R5N:-NMApgzQ83QNziALcR7M:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f41b31a01dd5bae92e90643d4105375feac6c1f5/core/src/main/java/google/registry/model/server/Lock.java#L159)):*

Need to remove the fist `%s`. Same below.
</blockquote></details>

Done.

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 46 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZer-T66ujztXqpHLy:-NMZer-T66ujztXqpHLz:bcb7dvy) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L46)):*
> ```Java
>    * <p>Since there are no cached resources, this class doesn't have to support being shutdown.
>    */
>   private static class NewRequestThreadExecutorService extends AbstractExecutorService {
> ```

The name is outdated if you don't use currentRequestThreadFactory.



_Code quote:_
```Java
NewRequestThreadExecutorService
```
___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NMZeCuN8W6Y5tEhBS4g:bho8387) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*
> ```Java
>     @Override
>     public void execute(Runnable command) {
>       Executor executor = Executors.newSingleThreadExecutor();
> ```

If you don't use currentRequestThreadFactory, you need to shut down your executor explicitly. 



_Code quote:_
```Java
xecutors.newSingleThreadExecutor()
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NM_fgWrABmn0F1LT2jU:b7o52sx) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

If you don't use currentRequestThreadFactory, you need to shut down your executor explicitly. 
</blockquote></details>

Are you sure? I was checking the docs https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#shutdown-- and SO https://stackoverflow.com/questions/16122987/reason-for-calling-shutdown-on-executorservice and as they say the only reason is that **"prevents clients to send more work to the executor service"**, however it also might affect current tasks under execution and interrupt them. Based on that I've decided not to do it.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NM_hNk6APb2AKjFKXyf:bgmiezs) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

Are you sure? I was checking the docs https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#shutdown-- and SO https://stackoverflow.com/questions/16122987/reason-for-calling-shutdown-on-executorservice and as they say the only reason is that **"prevents clients to send more work to the executor service"**, however it also might affect current tasks under execution and interrupt them. Based on that I've decided not to do it. 
</blockquote></details>

And to be clear - we do not assign any more tasks to executors anyway, so that's why preventing clients to send more tasks is not really applicable.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NMe2ANi0BTNiJmGhTov:b-3mfra9) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

And to be clear - we do not assign any more tasks to executors anyway, so that's why preventing clients to send more tasks is not really applicable. 
</blockquote></details>

Every time you call newSingleThreadExecutor() a new thread is created which never dies on its own (note the coreSize=1 parameter it passes to the delegate). This ties up resources and you'll run out of threads sooner or later.

Previously, threads created by Appengine's currentRequestThreadFactory are requested scoped, and the factory takes care of killing them on request completion.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 46 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZer-T66ujztXqpHLy:-NND5oJnAbZuFBgp-AjX:be0fyvw) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L46)):*

The name is outdated if you don't use currentRequestThreadFactory.
</blockquote></details>

After latest update which now creates ExecutorService the name fits right in
___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NND5wuN8AE__vzinahD:b-bu91xm) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

Every time you call newSingleThreadExecutor() a new thread is created which never dies on its own (note the coreSize=1 parameter it passes to the delegate). This ties up resources and you'll run out of threads sooner or later.

Previously, threads created by Appengine's currentRequestThreadFactory are requested scoped, and the factory takes care of killing them on request completion. 
</blockquote></details>

Updated

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @jianglai and @ptkach)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NNDX6Q89caQGeHcTCAg:b-v45cj1) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

Updated
</blockquote></details>

Three issues to pay attention here.

First, if the `command` Runnable may call appengine API (com.google.appengine.api.*), then you cannot phase out the currentRequestThreadFactory here. See https://cloud.google.com/appengine/docs/legacy/standard/java/runtime#threads

It is safer to wait until all other classes no longer import from the appengine api package.

For seamless migration to GKE, change the original line to the following:

MoreExecutors.platformThreadFactory().newThread(command).start();

This way, the code will run correctly both on and off AppEngine. MoreExecutors is a Guava class and checks if the app is on AppEngine when returning the thrad factory.

Second, when we move off AppEngine, we should start using a shared ExecutorService that only gets shutdown when the system goes down.
For now, leave a TODO here to change the code and remove the class javadoc above.


Lastly, we must test this change on the AppEngine.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NNHXmM7338HQ6VvG9NA:bdj7w1p) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

Three issues to pay attention here.

First, if the `command` Runnable may call appengine API (com.google.appengine.api.*), then you cannot phase out the currentRequestThreadFactory here. See https://cloud.google.com/appengine/docs/legacy/standard/java/runtime#threads

It is safer to wait until all other classes no longer import from the appengine api package.

For seamless migration to GKE, change the original line to the following:

MoreExecutors.platformThreadFactory().newThread(command).start();

This way, the code will run correctly both on and off AppEngine. MoreExecutors is a Guava class and checks if the app is on AppEngine when returning the thrad factory.

Second, when we move off AppEngine, we should start using a shared ExecutorService that only gets shutdown when the system goes down.
For now, leave a TODO here to change the code and remove the class javadoc above.


Lastly, we must test this change on the AppEngine. 
</blockquote></details>

Using `MoreExecutors` to be platform neutral is a good idea in general. FWIW, I believe this class is only called by the lock handler, which we removed the app engine API call from. This should have been explicated tested on GAE already.

As to what happens after we move to GKE, would it make sense to use a daemon thread for one-off asynchronous tasks? Using an `ExecutorService` seems to overkill.

weiminyu (approved):
> Reviewed 6 of 9 files at r1, 2 of 3 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach)

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NPTtRAL3gtdAS4PQBHF:b-eoistg) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

Using `MoreExecutors` to be platform neutral is a good idea in general. FWIW, I believe this class is only called by the lock handler, which we removed the app engine API call from. This should have been explicated tested on GAE already.

As to what happens after we move to GKE, would it make sense to use a daemon thread for one-off asynchronous tasks? Using an `ExecutorService` seems to overkill.
</blockquote></details>

@weiminyu  @jianglai  There seems to be 5 actions which invoke ```executeWithLocks```.  
I've checked the files and non of them appears to be relying on `com.google.appengine.api.*`
I've made the calls to all of them on **alpa** either from cloud scheduler or by making a POST requests via ```nomulus curl``` . All of them responded with *200 OK*. 

Is there anything else you can suggest to test to make sure it's safe to merge?

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1910)* status: all files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[`util/src/main/java/google/registry/util/AppEngineTimeLimiter.java` line 50 at r2](https://reviewable.io/reviews/google/nomulus/1910#-NMZeCuN8W6Y5tEhBS4f:-NPYvVw8DNtexxf7R7BL:b-ogfoum) ([raw file](https://github.com/google/nomulus/blob/6d0207d0384993f54e3837a2260a44abe33e15b1/util/src/main/java/google/registry/util/AppEngineTimeLimiter.java#L50)):*

@weiminyu  @jianglai  There seems to be 5 actions which invoke ```executeWithLocks```.  
I've checked the files and non of them appears to be relying on `com.google.appengine.api.*`
I've made the calls to all of them on **alpa** either from cloud scheduler or by making a POST requests via ```nomulus curl``` . All of them responded with *200 OK*. 

Is there anything else you can suggest to test to make sure it's safe to merge?
</blockquote></details>

If you already verified on alpha that none of the tasks errors out, I think that's pretty safe. Pay attention during deployment and check sandbox/production canary logs for errors.

PR #1912, opened 2023-01-18
Use a fake instance id in metric

jianglai:
Currently we synthesize a instance id which requires the use of App
Engine Module API. Given that we only have one version of code running
at one time, and HTTP is stateless, there is no point tracking exactly
which GAE "instance" is. We do lose information on which service (default,
backend, etc) is writing the metric, but that does not seem very
important.

Using a constant fake instance ID allows us to get rid of another GAE
dependency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1912)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1912)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1920, opened 2023-01-24
Add a condition update precaution to validateNewState

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1920)


jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1920)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java` line 312 at r1](https://reviewable.io/reviews/google/nomulus/1920#-NMdNaUJ26RoIQWF4HTS:-NMdNaUJ26RoIQWF4HTT:b-sy7rtw) ([raw file](https://github.com/google/nomulus/blob/6bf6ee40b8c44cdc6b37944ed5b74c41bf930a86/core/src/main/java/google/registry/flows/domain/DomainUpdateFlow.java#L312)):*
> ```Java
>   /**
>    * Before adding or amending conditions, existing data has to be verified for being compliant with
>    * an addition or amendment, otherwise existing data can become invalid and cause Domain update
> ```

"the additions or amendments"

Also, if we make this a javadoc instead of a comment, the first line should explain what the function does, in imperative mood. So it should be something like:

```
/** Checks whether the new state of the domain is valid.
 *
 * <p>Note that before adding or amending conditions...
 */
```



_Code quote:_
```Java
an addition or amendment
```

PR #1921, opened 2023-01-24
Use the Proxy-Authorization header when using nomulus + IAP

gbrodman:
https://cloud.google.com/iap/docs/authentication-howto#authenticating_from_proxy-authorization_header discusses how we can use Proxy-Authorization instead of the simple Authorization header to authenticate with IAP. This is helpful because it will allow us to seamlessly transition from non-IAP to IAP. That is, we can have a period where clients (e.g. Nomulus tool, cron jobs) send both the old Authorization and the new IAP Proxy-Authorization headers and the app will use either the old mechanism or IAP to authenticate the client based on whether or not IAP is enabled in Pantheon.

tested in QA that setting Proxy-Authorization works

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1921)

### DIFF
--- a/core/src/main/java/google/registry/tools/RequestFactoryModule.java
+++ b/core/src/main/java/google/registry/tools/RequestFactoryModule.java
@@ -69,14 +69,15 @@ class RequestFactoryModule {
       return new NetHttpTransport()
           .createRequestFactory(
               request -> {
+                // Use the standard credential initializer to set the Authorization header
+                credentialsBundle.getHttpRequestInitializer().initialize(request);
                 // If using IAP, use the refresh token to acquire an IAP-enabled ID token and use
                 // that for authentication.
                 if (iapClientId.isPresent()) {
                   String idToken = getIdToken(credentialsBundle, iapClientId.get());
-                  request.getHeaders().setAuthorization("Bearer " + idToken);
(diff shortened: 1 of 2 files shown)


jianglai (approved):
One thing to keep in mind is that while we do have control over the headers from the nomulus too and from the proxy, we should look into if Cloud Scheduler and Cloud Tasks allows setting the `Proxy-Authorization` header instead of the `Authorization` header (which is the default I presume).

See b/239727032 where it is mentioned. @ptkach will do some testing to see what kind of flexibility we have in terms of header names.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1921#-:-NMfegeu1-no_ViqkVks:b-7kv0by)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (commented):
yep, this is only one of the four clients where we'll need to make sure that we're setting the right header(s) 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1921#-:-NMjguLl9aWPM8vaWtPp:b-t7t9p2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1927, opened 2023-02-01
Allow incorrect fee extensions on domain creates with default tokens

sarahcaseybot:
Allow Domain creations to succeed if the fee in the fee extension is equal to or higher than the actual cost of the registration with the default token applied. Still require fees to match if there are no valid default tokens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1927)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -337,7 +337,7 @@ public final class DomainCreateFlow implements TransactionalFlow {
     FeesAndCredits feesAndCredits =
         pricingLogic.getCreatePrice(
             registry, targetId, now, years, isAnchorTenant, allocationToken);
-    validateFeeChallenge(feeCreate, feesAndCredits);
+    validateFeeChallenge(feeCreate, feesAndCredits, false);
     Optional<SecDnsCreateExtension> secDnsCreate =
         validateSecDnsExtension(eppInput.getSingleExtension(SecDnsCreateExtension.class));
     DateTime registrationExpirationTime = leapSafeAddYears(now, years);
--- a/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
@@ -277,8 +277,12 @@ public final class DomainCreateFlow implements TransactionalFlow {
             registrarId,
             now,
             eppInput.getSingleExtension(AllocationTokenExtension.class));
+    boolean defaultTokenUsed = false;
     if (!allocationToken.isPresent() && !registry.getDefaultPromoTokens().isEmpty()) {
       allocationToken = checkForDefaultToken(registry, command);
+      if (allocationToken.isPresent()) {
+        defaultTokenUsed = true;
+      }
     }
     boolean isAnchorTenant =
         isAnchorTenant(
@@ -337,7 +341,7 @@ public final class DomainCreateFlow implements TransactionalFlow {
     FeesAndCredits feesAndCredits =
         pricingLogic.getCreatePrice(
             registry, targetId, now, years, isAnchorTenant, allocationToken);
-    validateFeeChallenge(feeCreate, feesAndCredits, false);
+    validateFeeChallenge(feeCreate, feesAndCredits, defaultTokenUsed);
     Optional<SecDnsCreateExtension> secDnsCreate =
         validateSecDnsExtension(eppInput.getSingleExtension(SecDnsCreateExtension.class));
     DateTime registrationExpirationTime = leapSafeAddYears(now, years);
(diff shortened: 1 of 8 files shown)


gbrodman (changes requested):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1927)* status: 1 of 8 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 202 at r1](https://reviewable.io/reviews/google/nomulus/1927#-NNIXkdgC4L7_02RuM-B:-NNIXkdgC4L7_02RuM-C:biojy5z) ([raw file](https://github.com/google/nomulus/blob/3f344ab6c1000aea08cca0651093f011ad5325c2/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L202)):*
> ```Java
>             years,
>             existingRecurringBillingEvent);
>     validateFeeChallenge(feeRenew, feesAndCredits, false);
> ```

this seems a bit odd to me, shouldn't renewal for default-token-TLDs also be discounted? 

similar question for other operations like restores

gbrodman (changes requested):
> Reviewed 7 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1927)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 861 at r1](https://reviewable.io/reviews/google/nomulus/1927#-NOzijBT3v-bGnbRkOC-:-NOzijBT3v-bGnbRkOC0:bmtu5vd) ([raw file](https://github.com/google/nomulus/blob/3f344ab6c1000aea08cca0651093f011ad5325c2/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L861)):*
> ```Java
>     // Or if there is only a single fee type expected.
>     if (!feeTotal.equals(feesAndCredits.getTotalCost())) {
>       if (!defaultTokenUsed || feeTotal.isLessThan(feesAndCredits.getTotalCost())) {
> ```

this seems like a behavioral change (allowing situations without default tokens where the fee we pass in is more than what we expect). Is that expected? 

Unless I'm reading it incorrectly, this also means that we wouldn't throw an exception when we are using a default token and the expected fee total is less than the actual fee total, which seems less than ideal. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 202 at r1](https://reviewable.io/reviews/google/nomulus/1927#-NNIXkdgC4L7_02RuM-B:-NOzk4nC9M8ecOZA6a2V:bfe0uwd) ([raw file](https://github.com/google/nomulus/blob/3f344ab6c1000aea08cca0651093f011ad5325c2/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L202)):*

No, right now default tokens only affect create flows. However, we are considering in the future adding support for default token on renewals. 
</blockquote></details>

go/renewal-default-tokens

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1927)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 861 at r1](https://reviewable.io/reviews/google/nomulus/1927#-NOzijBT3v-bGnbRkOC-:-NPJ6fsDEWTPi4KN8Zrv:b-kgyg3v) ([raw file](https://github.com/google/nomulus/blob/3f344ab6c1000aea08cca0651093f011ad5325c2/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L861)):*

this seems like a behavioral change (allowing situations without default tokens where the fee we pass in is more than what we expect). Is that expected? 

Unless I'm reading it incorrectly, this also means that we wouldn't throw an exception when we are using a default token and the expected fee total is less than the actual fee total, which seems less than ideal. 
</blockquote></details>

I think you may be reading this incorrectly. If we are using a default token, then the only case where we would throw an exception is if the expected fee (feeTotal) is less than the actual fee (feesAndCredits.getTotalCost()). I added some clarifying comments in the tests in DomainCreateFlowTest to show where this is tested.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1927)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 861 at r1](https://reviewable.io/reviews/google/nomulus/1927#-NOzijBT3v-bGnbRkOC-:-NPJzwbXArDM88bZYMr9:bljappm) ([raw file](https://github.com/google/nomulus/blob/3f344ab6c1000aea08cca0651093f011ad5325c2/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L861)):*

I think you may be reading this incorrectly. If we are using a default token, then the only case where we would throw an exception is if the expected fee (feeTotal) is less than the actual fee (feesAndCredits.getTotalCost()). I added some clarifying comments in the tests in DomainCreateFlowTest to show where this is tested. 
</blockquote></details>

Right, so the only behavioral changes are that we allow lower-than-expected fees when using a default token, cool thanks.

PR #1933, opened 2023-02-07
Allow null GAIA IDs for User objects

gbrodman:
We were under the mistaken impression before that there was a reliable way to, out-of-band, get a GAIA ID for a particular email address. Unfortunately, that isn't the case (at least, not in a scalable way or one that support agents could use). As a result, we have to allow null GAIA IDs in the database.

When we (or the support team) create new users, we will only specify the email address and not the GAIA ID. Then, when the user logs in for the first time, we will have the GAIA ID from the provided ID token, and we can populate it then.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1933)


jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1933)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #1937, opened 2023-02-09
Change default beam job region

jianglai:
For reasons that I cannot explain, the same expand recurring billing
event pipeline would fail in us-east1 but succeed in us-central1.

See:

https://pantheon.corp.google.com/dataflow/jobs/us-central1/2023-02-09_14_52_24-162498476138221714;graphView=0?project=domain-registry

https://pantheon.corp.google.com/dataflow/jobs/us-east1/2023-02-09_14_26_07-4564782062878841960;graphView=1?project=domain-registry

Also improved how the accuracy of the metrics:

It is observed that both counters are consistently higher for the same
start and end times when running in dry run mode. There is no way to
test for consistency when not running in dry run, for obviously reasons.

I can make the recurrings in scope counter consistent by not updating it
in a side-effect-causing transaction, but there is no way around the
other counter. It can only be trusted when running in dry run mode,
unfortunately.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1937)

### DIFF
--- a/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java
+++ b/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java
@@ -138,9 +138,15 @@ public class ExpandRecurringBillingEventsPipeline implements Serializable {
   private final boolean isDryRun;
   private final boolean advanceCursor;
   private final Counter recurringsInScopeCounter =
-      Metrics.counter("ExpandBilling", "RecurringsInScope");
-  private final Counter expandedOneTimeCounter =
-      Metrics.counter("ExpandBilling", "ExpandedOneTime");
+      Metrics.counter("ExpandBilling", "Recurrings in scope for expansion");
+  // Note that this counter is only accurate when running in dry run mode. Because SQL persistence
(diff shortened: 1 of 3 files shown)


gbrodman (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1937)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ptkach and @weiminyu)

___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 430 at r3](https://reviewable.io/reviews/google/nomulus/1937#-NOBWdUg9KNnjpCRxIoJ:-NOBWdUh5g0-W3G89_xA:b-6xwlz1) ([raw file](https://github.com/google/nomulus/blob/61abec27b4023c8c0a178b6709b89089fcd1c66f/core/src/main/java/google/registry/config/files/default-config.yaml#L430)):*
> ```YAML
> beam:
>   # The default region to run Apache Beam (Cloud Dataflow) jobs in.
>   defaultJobRegion: us-central1
> ```

super weird

PR #1940, opened 2023-02-17
Add a table to record EPP resources needing DNS refresh

jianglai:
The corresponding entity is created in #1941.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1940)

PR #1941, opened 2023-02-22
Create a DnsRefreshRequest entity backed by the corresponding table

jianglai:
The table is created in #1940.

Also adds a DnsUtils class to deal with adding, polling, and removing
DNS refresh requests (only adding is implemented for now). The class
also takes care of choosing which mechanism to use (pull queue vs. SQL)
based on the current time and the database migration schedule map.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1941)

### DIFF
--- a/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
@@ -32,7 +32,7 @@ import com.google.common.collect.Sets;
 import com.google.common.flogger.FluentLogger;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.config.RegistryEnvironment;
-import google.registry.dns.DnsQueue;
+import google.registry.dns.DnsUtils;
 import google.registry.model.CreateAutoTimestamp;
 import google.registry.model.EppResourceUtils;
 import google.registry.model.domain.Domain;
@@ -98,7 +98,7 @@ public class DeleteProberDataAction implements Runnable {
   /** Number of domains to retrieve and delete per SQL transaction. */
   private static final int BATCH_SIZE = 1000;
 
-  @Inject DnsQueue dnsQueue;
+  @Inject DnsUtils dnsUtils;
 
   @Inject
   @Parameter(PARAM_DRY_RUN)
@@ -264,6 +264,6 @@ public class DeleteProberDataAction implements Runnable {
     // messages, or auto-renews because those will all be hard-deleted the next time the job runs
     // anyway.
     tm().putAll(ImmutableList.of(deletedDomain, historyEntry));
-    dnsQueue.addDomainRefreshTask(deletedDomain.getDomainName());
+    dnsUtils.requestDomainDnsRefresh(deletedDomain.getDomainName());
   }
 }
--- a/core/src/main/java/google/registry/dns/DnsConstants.java
+++ b/core/src/main/java/google/registry/dns/DnsConstants.java
@@ -34,5 +34,8 @@ public class DnsConstants {
   public static final String DNS_TARGET_CREATE_TIME_PARAM = "Create-Time";
 
   /** The possible values of the {@code DNS_TARGET_TYPE_PARAM} parameter. */
-  public enum TargetType { DOMAIN, HOST, ZONE }
+  public enum TargetType {
+    DOMAIN,
+    HOST
+  }
 }
--- a/core/src/main/java/google/registry/dns/DnsQueue.java
+++ b/core/src/main/java/google/registry/dns/DnsQueue.java
@@ -77,9 +77,13 @@ public class DnsQueue {
   private static final RateLimiter rateLimiter = RateLimiter.create(9);
 
   @Inject
-  public DnsQueue(@Named(DNS_PULL_QUEUE_NAME) Queue queue, Clock clock) {
+  DnsQueue(@Named(DNS_PULL_QUEUE_NAME) Queue queue, Clock clock) {
     this.queue = queue;
     this.clock = clock;
+  }
+
+  Clock getClock() {
+    return clock;
   }
 
   @VisibleForTesting
@@ -108,7 +112,7 @@ public class DnsQueue {
   }
 
   /** Adds a task to the queue to refresh the DNS information for the specified subordinate host. */
-  public TaskHandle addHostRefreshTask(String hostName) {
+  TaskHandle addHostRefreshTask(String hostName) {
     Optional<InternetDomainName> tld = Registries.findTldForName(InternetDomainName.from(hostName));
     checkArgument(
         tld.isPresent(), String.format("%s is not a subordinate host to a known tld", hostName));
@@ -116,22 +120,17 @@ public class DnsQueue {
   }
 
   /** Enqueues a task to refresh DNS for the specified domain now. */
-  public TaskHandle addDomainRefreshTask(String domainName) {
+  TaskHandle addDomainRefreshTask(String domainName) {
     return addDomainRefreshTask(domainName, Duration.ZERO);
   }
 
   /** Enqueues a task to refresh DNS for the specified domain at some point in the future. */
-  public TaskHandle addDomainRefreshTask(String domainName, Duration countdown) {
+  TaskHandle addDomainRefreshTask(String domainName, Duration countdown) {
     return addToQueue(
         TargetType.DOMAIN,
         domainName,
         assertTldExists(getTldFromDomainName(domainName)),
         countdown);
-  }
-
-  /** Adds a task to the queue to refresh the DNS information for the specified zone. */
-  public TaskHandle addZoneRefreshTask(String zoneName) {
-    return addToQueue(TargetType.ZONE, zoneName, zoneName, Duration.ZERO);
   }
 
   /**
--- a/core/src/main/java/google/registry/dns/DnsUtils.java
+++ b/core/src/main/java/google/registry/dns/DnsUtils.java
@@ -1,0 +1,74 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.dns;
+
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+
+import com.google.common.net.InternetDomainName;
+import google.registry.dns.DnsConstants.TargetType;
+import google.registry.model.common.DatabaseMigrationStateSchedule;
+import google.registry.model.common.DatabaseMigrationStateSchedule.MigrationState;
+import google.registry.model.common.DnsRefreshRequest;
+import google.registry.model.tld.Registries;
+import javax.inject.Inject;
+import org.joda.time.Duration;
+
+/** Utility class to handle DNS refresh requests. */
+// TODO: Make this a static util function once we are done with the DNS pull queue migration.
+public class DnsUtils {
+
+  private final DnsQueue dnsQueue;
+
+  @Inject
+  DnsUtils(DnsQueue dnsQueue) {
+    this.dnsQueue = dnsQueue;
+  }
+
+  private void requestDnsRefresh(String name, TargetType type, Duration delay) {
+    // Throws an IllegalArgumentException if the name is not under a managed TLD -- we only update
+    // DNS for names that are under our management.
+    String tld = Registries.findTldForNameOrThrow(InternetDomainName.from(name)).toString();
+    if (usePullQueue()) {
+      if (TargetType.HOST.equals(type)) {
+        dnsQueue.addHostRefreshTask(name);
+      } else {
+        dnsQueue.addDomainRefreshTask(name, delay);
+      }
+    } else {
+      tm().transact(
+              () ->
+                  tm().insert(
+                          new DnsRefreshRequest(
+                              type, name, tld, tm().getTransactionTime().plus(delay))));
+    }
+  }
+
+  public void requestDomainDnsRefresh(String domainName, Duration delay) {
+    requestDnsRefresh(domainName, TargetType.DOMAIN, delay);
+  }
+
+  public void requestDomainDnsRefresh(String domainName) {
+    requestDomainDnsRefresh(domainName, Duration.ZERO);
+  }
+
+  public void requestHostDnsRefresh(String hostName) {
+    requestDnsRefresh(hostName, TargetType.HOST, Duration.ZERO);
+  }
+
+  private boolean usePullQueue() {
+    return !DatabaseMigrationStateSchedule.getValueAtTime(dnsQueue.getClock().nowUtc())
+        .equals(MigrationState.DNS_SQL);
+  }
+}
--- a/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
+++ b/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java
@@ -85,7 +85,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
-  private final DnsQueue dnsQueue;
+  private final DnsUtils dnsUtils;
   private final DnsWriterProxy dnsWriterProxy;
   private final DnsMetrics dnsMetrics;
   private final Duration timeout;
@@ -139,7 +139,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
       @Config("gSuiteOutgoingEmailAddress") InternetAddress gSuiteOutgoingEmailAddress,
       @Header(APP_ENGINE_RETRY_HEADER) Optional<Integer> appEngineRetryCount,
       @Header(CLOUD_TASKS_RETRY_HEADER) Optional<Integer> cloudTasksRetryCount,
-      DnsQueue dnsQueue,
+      DnsUtils dnsUtils,
       DnsWriterProxy dnsWriterProxy,
       DnsMetrics dnsMetrics,
       LockHandler lockHandler,
@@ -147,7 +147,7 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
       CloudTasksUtils cloudTasksUtils,
       SendEmailService sendEmailService,
       Response response) {
-    this.dnsQueue = dnsQueue;
+    this.dnsUtils = dnsUtils;
     this.dnsWriterProxy = dnsWriterProxy;
     this.dnsMetrics = dnsMetrics;
     this.timeout = timeout;
@@ -356,10 +356,10 @@ public final class PublishDnsUpdatesAction implements Runnable, Callable<Void> {
   private void requeueBatch() {
     logger.atInfo().log("Requeueing batch for retry.");
     for (String domain : nullToEmpty(domains)) {
-      dnsQueue.addDomainRefreshTask(domain);
+      dnsUtils.requestDomainDnsRefresh(domain);
     }
     for (String host : nullToEmpty(hosts)) {
-      dnsQueue.addHostRefreshTask(host);
+      dnsUtils.requestHostDnsRefresh(host);
     }
   }
 
--- a/core/src/main/java/google/registry/dns/RefreshDnsAction.java
+++ b/core/src/main/java/google/registry/dns/RefreshDnsAction.java
@@ -39,7 +39,7 @@ import javax.inject.Inject;
 public final class RefreshDnsAction implements Runnable {
 
   private final Clock clock;
-  private final DnsQueue dnsQueue;
+  private final DnsUtils dnsUtils;
   private final String domainOrHostName;
   private final TargetType type;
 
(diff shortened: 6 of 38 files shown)


jianglai (commented):
> Reviewed 8 of 8 files at r1, 5 of 5 files at r2, 1 of 14 files at r3, 42 of 42 files at r4, 15 of 15 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

jianglai (commented):
> Reviewed 2 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

jianglai (commented):
Gentle ping.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941#-:-NQ5oSi23iol7kP1Ge9D:bq2fntq)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 1 of 8 files at r1, 5 of 5 files at r2, 1 of 14 files at r3, 28 of 42 files at r4, 12 of 15 files at r5, 1 of 3 files at r6, 2 of 2 files at r7, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: all files reviewed, 7 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/dns/DnsQueue.java` line 115 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQa8WkEMn_T5b5K2IP:-NQQa8WkEMn_T5b5K2IQ:bjm0nsd) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/dns/DnsQueue.java#L115)):*
> ```Java
> 
>   /** Adds a task to the queue to refresh the DNS information for the specified subordinate host. */
>   protected TaskHandle addHostRefreshTask(String hostName) {
> ```

I might be missing it, but does anything extend this / is there any reason to make these methods protected? 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 717 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQaY6vCSbpq9owHIjT:-NQQaY6vCSbpq9owHIjU:b-k22g0l) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L717)):*
> ```Java
>       dnsUtils.requestDomainDnsRefresh(newDomain.getDomainName());
>     }
>     if (!(DatabaseMigrationStateSchedule.getValueAtTime(tm().getTransactionTime())
> ```

hmm it feels like there must be a better way of doing this, especially if we keep adding more states in the future 

Maybe something using the ordering of the states to see if we are currently at-or-after a particular state (in this case, NORDN_SQL)? 

To be honest, at some point we may want to change it to be a feature-flag system rather than a single map with stages. 
___
*[`core/src/main/java/google/registry/model/common/DnsRefreshRequest.java` line 59 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQdZ1j2t9PoV4WLNKD:-NQQdZ1j2t9PoV4WLNKE:bab4m5i) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/model/common/DnsRefreshRequest.java#L59)):*
> ```Java
> 
>   @Column(nullable = false)
>   private DateTime processTime;
> ```

the naming of this is vague, since it could be referring to something either in the past or the future 

something like `lastProcessedTime` would be clearer as to exactly what it's referring to (this would apply for the things like messages that refer to this variable too of course)
___
*[`core/src/test/java/google/registry/dns/DnsUtilsTest.java` line 149 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQbzr29Ms7QX2_tgX5:-NQQbzr29Ms7QX2_tgX6:bbdv4a3) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/DnsUtilsTest.java#L149)):*
> ```Java
>   }
> 
>   private static void useDnsSql() {
> ```

what's up with this method? we shouldn't need all these calls 
___
*[`core/src/test/java/google/registry/dns/RefreshDnsOnHostRenameActionTest.java` line 63 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQcGKk6OEsvWBmIfF6:-NQQcGKk6OEsvWBmIfF7:bb69l0f) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/RefreshDnsOnHostRenameActionTest.java#L63)):*
> ```Java
>     createTld("tld");
>   }
> ```

extra newline
___
*[`core/src/test/java/google/registry/model/common/DnsRefreshRequestTest.java` line 36 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQcZ5P9pxBzzLlUPQl:-NQQcZ5P9pxBzzLlUPQm:b-aexbm6) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/model/common/DnsRefreshRequestTest.java#L36)):*
> ```Java
>   }
> 
>   DnsRefreshRequest request;
> ```

nit: "private", and also I don't see any reason why we couldn't just create the request here inline 
___
*[`core/src/test/java/google/registry/schema/integration/SqlIntegrationTestSuite.java` line 91 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQcs2f2AF--bTrmtL5:-NQQcs2f2AF--bTrmtL6:b-8fbynr) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/schema/integration/SqlIntegrationTestSuite.java#L91)):*
> ```Java
>   DomainSqlTest.class,
>   DomainHistoryTest.class,
>   DnsRefreshRequestTest.class,
> ```

nit: alphabetization

jianglai (commented):
> Reviewed 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: all files reviewed, 6 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/dns/DnsQueue.java` line 115 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQa8WkEMn_T5b5K2IP:-NQRCg8XA3Hz3b9PD-Wv:ba2jxvb) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/dns/DnsQueue.java#L115)):*

I might be missing it, but does anything extend this / is there any reason to make these methods protected? 
</blockquote></details>

I made it protected so that only the newly added `DnsUtils` can use it. The goal is to not call this method directly and rely on `DnsUtils` because the latter is able to decide whether to use the pull queue-based or the SQL-based flow based on the current time and the database migration schedule.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 717 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQaY6vCSbpq9owHIjT:-NQRDNzF5SaJb9DB9tZ7:bfpqe4r) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L717)):*

hmm it feels like there must be a better way of doing this, especially if we keep adding more states in the future 

Maybe something using the ordering of the states to see if we are currently at-or-after a particular state (in this case, NORDN_SQL)? 

To be honest, at some point we may want to change it to be a feature-flag system rather than a single map with stages. 
</blockquote></details>

I completely agree with you that this is ugly as ****. As you said the best way to handle this to make `DatabaseMigrationStateSchedule` support `isBefore(MigrationState)`/`isAfter(MigrationState)`.

However the migration state schedule class is never meant for this and we kind of shoehorned it in there to give us the ability to set a flag at a given time (again, like you suggested). The switching logic regarding Nordn will be deleted as soon as we confirm that SQL mode is working properly on production.

It might be worth considering if we want to set up a proper flag system (with time triggers) so that we can turn on features orthogonally. For now I think the best way forward is to piggy back on the migration schedule, which we intentionally didn't delete because of one-off migration related switches like this. I don't foresee us needing it any more after the Nordn and DNS SQL migration.
___
*[`core/src/main/java/google/registry/model/common/DnsRefreshRequest.java` line 59 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQdZ1j2t9PoV4WLNKD:-NQRTZNo54SfXEXEjk6l:bjnfoyb) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/model/common/DnsRefreshRequest.java#L59)):*

the naming of this is vague, since it could be referring to something either in the past or the future 

something like `lastProcessedTime` would be clearer as to exactly what it's referring to (this would apply for the things like messages that refer to this variable too of course)
</blockquote></details>

Yeah, `lastProcessedTime` is probably a more apt name. I did explain what this field is supposed to store in the getter method's javadoc. If we change the name now, we'd need to do another schema deployment (unless we want to update the filed name but not update the column name accordingly, which seems like a bad idea).

I kinda feel that it could work as is, unless you feel strongly that we should change the name.
___
*[`core/src/test/java/google/registry/dns/DnsUtilsTest.java` line 149 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQbzr29Ms7QX2_tgX5:-NQRIpP820tfWGjfOR1N:b-d31l39) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/DnsUtilsTest.java#L149)):*

what's up with this method? we shouldn't need all these calls 
</blockquote></details>

We do need it because otherwise the `DnsUtils` will always use the pull queue mode in tests. We have to configure the migration schedule so to control which mode to use.
___
*[`core/src/test/java/google/registry/dns/RefreshDnsOnHostRenameActionTest.java` line 63 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQcGKk6OEsvWBmIfF6:-NQRJC7pA7UKB85sZw1w:b-896fix) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/RefreshDnsOnHostRenameActionTest.java#L63)):*

extra newline
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/model/common/DnsRefreshRequestTest.java` line 36 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQcZ5P9pxBzzLlUPQl:-NQRVP7gB-WZzf7hBo7e:b-896fix) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/model/common/DnsRefreshRequestTest.java#L36)):*

nit: "private", and also I don't see any reason why we couldn't just create the request here inline 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/schema/integration/SqlIntegrationTestSuite.java` line 91 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQcs2f2AF--bTrmtL5:-NQRJMTw6m1wxFnO-2sd:b-kl8g4c) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/schema/integration/SqlIntegrationTestSuite.java#L91)):*

nit: alphabetization 
</blockquote></details>

good catch!

gbrodman (changes requested):
> Reviewed 3 of 4 files at r10.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/dns/DnsQueue.java` line 115 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQa8WkEMn_T5b5K2IP:-NQRXrqbCmTRDwbUKdIN:bctlzm0) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/dns/DnsQueue.java#L115)):*

I made it protected so that only the newly added `DnsUtils` can use it. The goal is to not call this method directly and rely on `DnsUtils` because the latter is able to decide whether to use the pull queue-based or the SQL-based flow based on the current time and the database migration schedule.
</blockquote></details>

Oh I see, can we make it package-private instead then? I was confused because I was like "wait, this class doesn't have any subclasses" so I thought it might have been a typo or something. 
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 717 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQaY6vCSbpq9owHIjT:-NQVVZ0t8BvSgMxG8OSU:b-e35evn) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L717)):*

I completely agree with you that this is ugly as ****. As you said the best way to handle this to make `DatabaseMigrationStateSchedule` support `isBefore(MigrationState)`/`isAfter(MigrationState)`.

However the migration state schedule class is never meant for this and we kind of shoehorned it in there to give us the ability to set a flag at a given time (again, like you suggested). The switching logic regarding Nordn will be deleted as soon as we confirm that SQL mode is working properly on production.

It might be worth considering if we want to set up a proper flag system (with time triggers) so that we can turn on features orthogonally. For now I think the best way forward is to piggy back on the migration schedule, which we intentionally didn't delete because of one-off migration related switches like this. I don't foresee us needing it any more after the Nordn and DNS SQL migration.
</blockquote></details>

yeah I think the pull queue migration(s) are hopefully the last time we need it :( 
___
*[`core/src/main/java/google/registry/model/common/DnsRefreshRequest.java` line 59 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQdZ1j2t9PoV4WLNKD:-NQVYnzr5ihCIM7SQFSb:byas5z6) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/model/common/DnsRefreshRequest.java#L59)):*

Yeah, `lastProcessedTime` is probably a more apt name. I did explain what this field is supposed to store in the getter method's javadoc. If we change the name now, we'd need to do another schema deployment (unless we want to update the filed name but not update the column name accordingly, which seems like a bad idea).

I kinda feel that it could work as is, unless you feel strongly that we should change the name.
</blockquote></details>

yeah, unfortunately I think we should probably change it since (hopefully) this will be around forever, right? 

If it sounds good to you, we could easily do that schema change today and then a second release this week so that we can merge this PR in before the end of the week?

sorry about this, I recognize it's inconvenient 
___
*[`core/src/test/java/google/registry/dns/DnsUtilsTest.java` line 149 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQbzr29Ms7QX2_tgX5:-NQVW1nu5xCKNGz0lr2o:b-7qeeps) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/DnsUtilsTest.java#L149)):*

We do need it because otherwise the `DnsUtils` will always use the pull queue mode in tests. We have to configure the migration schedule so to control which mode to use.
</blockquote></details>

My point is that I wouldn't expect that it would require multiple calls, since you're just overwriting the schedule each time

jianglai (commented):
> Reviewed 10 of 10 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/dns/DnsQueue.java` line 115 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQa8WkEMn_T5b5K2IP:-NQVbASX0cKESJ1HFMQj:b-qweozj) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/dns/DnsQueue.java#L115)):*

Oh I see, can we make it package-private instead then? I was confused because I was like "wait, this class doesn't have any subclasses" so I thought it might have been a typo or something. 
</blockquote></details>

You are right, they should have been package-private.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 717 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQaY6vCSbpq9owHIjT:-NQVbNMQ56l1tlESnVAU:b-gj224x) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L717)):*

yeah I think the pull queue migration(s) are hopefully the last time we need it :( 
</blockquote></details>

Made b/273512400 to track the feature request to implement a flag system.
___
*[`core/src/main/java/google/registry/model/common/DnsRefreshRequest.java` line 59 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQdZ1j2t9PoV4WLNKD:-NQViC9t97E5cIMHd1cU:bby8ebf) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/model/common/DnsRefreshRequest.java#L59)):*

yeah, unfortunately I think we should probably change it since (hopefully) this will be around forever, right? 

If it sounds good to you, we could easily do that schema change today and then a second release this week so that we can merge this PR in before the end of the week?

sorry about this, I recognize it's inconvenient 
</blockquote></details>

That's fine. I'll be out for the next two weeks and it's unlikely that the whole migration will be done before EOW anyway. The good thing is that this column is not used yet so there's need for a data migration.

Created #1962 to rename the column. Note that this PR currently contains changes in #1962, which would be diffbased away once #1962 is merged.
___
*[`core/src/test/java/google/registry/dns/DnsUtilsTest.java` line 149 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQbzr29Ms7QX2_tgX5:-NQViaj3BNF5tneT7UCd:bvupnl6) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/DnsUtilsTest.java#L149)):*

My point is that I wouldn't expect that it would require multiple calls, since you're just overwriting the schedule each time 
</blockquote></details>

You'd think so :P but no, `DatabaseMigrationStateSchedule` also checks that you only transition to the next state from your current state as of now, in the new schedule. Since the time is set to 2020, we cannot add the entire schedule at once because that would make us jump directly from the initial state (`DATASTORE_ONLY`) to the latest state in the new schedule, which is illegal.

I was lazy to figure out a better way, but setting the clock to `START_OF_TIME` temporarily seems to be more elegant solution.

gbrodman (approved):
> Reviewed 10 of 10 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/common/DnsRefreshRequest.java` line 59 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQdZ1j2t9PoV4WLNKD:-NQWVn9N9VXoq9Fwxlr2:b-idfm8j) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/main/java/google/registry/model/common/DnsRefreshRequest.java#L59)):*

That's fine. I'll be out for the next two weeks and it's unlikely that the whole migration will be done before EOW anyway. The good thing is that this column is not used yet so there's need for a data migration.

Created #1962 to rename the column. Note that this PR currently contains changes in #1962, which would be diffbased away once #1962 is merged.
</blockquote></details>

thanks! 
___
*[`core/src/test/java/google/registry/dns/DnsUtilsTest.java` line 149 at r9](https://reviewable.io/reviews/google/nomulus/1941#-NQQbzr29Ms7QX2_tgX5:-NQWVXeHAdx_BIq3T7oe:bo55zaw) ([raw file](https://github.com/google/nomulus/blob/36134b7bd0d2f4e01c09fce4bf6212fa57432aaa/core/src/test/java/google/registry/dns/DnsUtilsTest.java#L149)):*

You'd think so :P but no, `DatabaseMigrationStateSchedule` also checks that you only transition to the next state from your current state as of now, in the new schedule. Since the time is set to 2020, we cannot add the entire schedule at once because that would make us jump directly from the initial state (`DATASTORE_ONLY`) to the latest state in the new schedule, which is illegal.

I was lazy to figure out a better way, but setting the clock to `START_OF_TIME` temporarily seems to be more elegant solution.
</blockquote></details>

that works! yeah this is definitely better than setting it multiple times, even though it's slightly weird to manipulate time

gbrodman (approved):
> Reviewed 8 of 8 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1941)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1942, opened 2023-02-23
Add allowed license for json

jianglai:
For some reason `./gradlew clean build` on master is failing for me on
multiple machines due to a new org.json:json version triggering license
violations, even though the lock files are not changing.

Note that the old versions are still present because if I remove
"The JSON license", which the old versions use, the check also fails...

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1942)

### DIFF
--- a/config/dependency-license/allowed_licenses.json
+++ b/config/dependency-license/allowed_licenses.json
@@ -271,6 +271,10 @@
       "moduleName": "org.tukaani:xz"
     },
     {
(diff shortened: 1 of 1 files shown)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1942)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1948, opened 2023-02-28
Add a process time column to DnsRefreshRequest

jianglai:
The value of the column would be set to START_OF_TIME for new entries.
Every time a row is read, the value is updated to the current time. This
allows concurrent reads to not repeatedly read the same entry that has the
earliest request time, because they would only look for rows that have a value
of process time that is before current time - some padding time.

This basically fulfills the same function that LEASE_PADDING gives us
when using a pull queue, whereas a task would be leased for a certain
time, during which time they would not be leased by anyone else.

See: https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/dns/ReadDnsQueueAction.java;l=99?q=readdnsqueue&ss=nomulus%2Fnomulus

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1948)


weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1948)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #1952, opened 2023-03-03
Separate DEFAULT_TOKEN type into CREATE and RENEW types

sarahcaseybot:
This separates the DEFAULT_TOKEN type into DEFAULT_CREATE_TOKEN and DEFAULT_RENEW_TOKEN. This will allow us to store default tokens for both types of commands on the same field of the TLD but distinguish what specific command that token should be applied to. 

This marks the DEFAULT_TOKEN type as deprecated, however I do not believe there currently are any tokens in production or sandbox with that type, so it may be safe to remove it entirely.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1952)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -118,8 +118,17 @@ public class AllocationToken extends UpdateAutoTimestampEntity implements Builda
 
   /** Type of the token that indicates how and where it should be used. */
   public enum TokenType {
+    /**
+     * Token soved on a TLD to use on a create request if no other token is passed from the client
+     */
+    DEFAULT_CREATE_PROMO,
     /** Token saved on a TLD to use if no other token is passed from the client */
+    @Deprecated
     DEFAULT_PROMO,
+    /**
+     * Token soved on a TLD to use on a renew request if no other token is passed from the client
+     */
+    DEFAULT_RENEW_PROMO,
     /** Token used for package pricing */
     PACKAGE,
     /** Invalid after use */
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -123,6 +123,7 @@ public class AllocationToken extends UpdateAutoTimestampEntity implements Builda
      */
     DEFAULT_CREATE_PROMO,
     /** Token saved on a TLD to use if no other token is passed from the client */
+    // TODO(sarahbot@): Remove this type once we are sure it is safe
     @Deprecated
     DEFAULT_PROMO,
     /**
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -119,14 +119,14 @@ public class AllocationToken extends UpdateAutoTimestampEntity implements Builda
   /** Type of the token that indicates how and where it should be used. */
   public enum TokenType {
     /**
-     * Token soved on a TLD to use on a create request if no other token is passed from the client
+     * Token saved on a TLD to use on a create request if no other token is passed from the client
      */
     DEFAULT_CREATE_PROMO,
     /** Token saved on a TLD to use if no other token is passed from the client */
     @Deprecated
     DEFAULT_PROMO,
     /**
-     * Token soved on a TLD to use on a renew request if no other token is passed from the client
+     * Token saved on a TLD to use on a renew request if no other token is passed from the client
      */
     DEFAULT_RENEW_PROMO,
     /** Token used for package pricing */
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -22,6 +22,7 @@ import static google.registry.model.domain.token.AllocationToken.TokenStatus.CAN
 import static google.registry.model.domain.token.AllocationToken.TokenStatus.ENDED;
 import static google.registry.model.domain.token.AllocationToken.TokenStatus.NOT_STARTED;
 import static google.registry.model.domain.token.AllocationToken.TokenStatus.VALID;
+import static google.registry.model.domain.token.AllocationToken.TokenType.DEFAULT_PROMO;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.CollectionUtils.forceEmptyToNull;
 import static google.registry.util.CollectionUtils.nullToEmptyImmutableCopy;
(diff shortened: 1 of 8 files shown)


gbrodman (changes requested):
Thoughts on adding a check to make sure that no new instances of `DEFAULT_TOKEN` are created? That way, if there are none in prod / sandbox right now, we can guarantee that it's safe to remove it entirely. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1952#-:-NPxktLmB7ChBlVdUeGz:bhkaxon)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java` line 379 at r1](https://reviewable.io/reviews/google/nomulus/1952#-NPy1VgF8LNRKqRCznIs:-NPy1VgGAN-1ohPdyiYG:b-q5hs1z) ([raw file](https://github.com/google/nomulus/blob/199a9e96f98b8c908bae2da93d96430e32be7ca5/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L379)):*
> ```Java
>         .hasMessageThat()
>         .isEqualTo(
>             "Invalid value for -t parameter. Allowed values:[DEFAULT_CREATE_PROMO, DEFAULT_PROMO,"
> ```

we probably want to update this message too (don't want to allow DEFAULT_PROMO)

sarahcaseybot (commented):
Good idea, added a check

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1952#-:-NQ1n7JH8vfWQs0S4IN2:b1ikqu8)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java` line 379 at r1](https://reviewable.io/reviews/google/nomulus/1952#-NPy1VgF8LNRKqRCznIs:-NQ1oxhYA_7dE8g7d-S9:bwzg3em) ([raw file](https://github.com/google/nomulus/blob/199a9e96f98b8c908bae2da93d96430e32be7ca5/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L379)):*

we probably want to update this message too (don't want to allow DEFAULT_PROMO) 
</blockquote></details>

I'm not sure how to change this message since this is autogenerated based on the values in the enum

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1952)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java` line 283 at r2](https://reviewable.io/reviews/google/nomulus/1952#-NQ6kfqR02kHVnbHpqIi:-NQ6kfqR02kHVnbHpqIj:br7lk3t) ([raw file](https://github.com/google/nomulus/blob/03eab5bf05a48276389bdafd0cc82f6b92c160e4/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L283)):*
> ```Java
>     // Don't allow DEFAULT_TOKEN token type on new tokens so the deprecated type is safe to remove
>     // in the future
>     checkArgument(
> ```

Can we just put this in the AllocationToken.Builder class so that we can be 100% sure it's not created anywhere? 
___
*[`core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java` line 379 at r1](https://reviewable.io/reviews/google/nomulus/1952#-NPy1VgF8LNRKqRCznIs:-NQ6kkyy10GzKo80FIJH:bjo6cx4) ([raw file](https://github.com/google/nomulus/blob/199a9e96f98b8c908bae2da93d96430e32be7ca5/core/src/test/java/google/registry/tools/GenerateAllocationTokensCommandTest.java#L379)):*

I'm not sure how to change this message since this is autogenerated based on the values in the enum
</blockquote></details>

oh dang

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1952)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 124 at r2](https://reviewable.io/reviews/google/nomulus/1952#-NQ6l4ES-QMVxW13dzo_:-NQ6l4ES-QMVxW13dzoa:b3xj911) ([raw file](https://github.com/google/nomulus/blob/03eab5bf05a48276389bdafd0cc82f6b92c160e4/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L124)):*
> ```Java
>      * Token saved on a TLD to use on a create request if no other token is passed from the client
>      */
>     DEFAULT_CREATE_PROMO,
> ```

Is this the right approach?  I believe our current planned promo will use all the exact same discounts on creates and renews, so won't this make more work for us?

I think adding a separate field that, if specified, lists just the allowable actions (which could be expanded to include TRANSFERS in the future as well) would make more sense, but I'm also not clear if we even need this quite yet.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1952)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/token/AllocationToken.java` line 124 at r2](https://reviewable.io/reviews/google/nomulus/1952#-NQ6l4ES-QMVxW13dzo_:-NQ7285J2ovg6Rl7oeyi:b-4ru31m) ([raw file](https://github.com/google/nomulus/blob/03eab5bf05a48276389bdafd0cc82f6b92c160e4/core/src/main/java/google/registry/model/domain/token/AllocationToken.java#L124)):*

I did not know the plan was to use the same discounts for creates and renews, but I assumed if we did that we would create separate default tokens  for these actions in order to fine tune the promotions by action. However I suppose it could also be done by adding a separate field to the AllocationToken object. However, that will require a database change which the token type change does not require.
</blockquote></details>

It's worth talking to Bruno, but, it seems somewhat counter-intuitive to me that we'd have multiple tokens to manage for a single promotion.  That means that every time a partner is added (or drops out), we'd have to update multiple tokens.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1952)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java` line 283 at r2](https://reviewable.io/reviews/google/nomulus/1952#-NQ6kfqR02kHVnbHpqIi:-NQ6wNW35CPsMcssp2sM:b-896fix) ([raw file](https://github.com/google/nomulus/blob/03eab5bf05a48276389bdafd0cc82f6b92c160e4/core/src/main/java/google/registry/tools/GenerateAllocationTokensCommand.java#L283)):*

Can we just put this in the AllocationToken.Builder class so that we can be 100% sure it's not created anywhere? 
</blockquote></details>

Done.

CydeWeys:
Yes, so after discussion with Bruno the final decision is to leave `DEFAULT` here alone and to add a new `Set<Enum>` field (or similar) that governs the list of allowable EPP actions for a given token, which would default to "all actions" if unspecified.

This seems to be the best way forward as it prevents combinatorial explosion, e.g. it seems plausible we could also need the ability to limit explicitly specified tokens to CREATES, RENEWS, TRANSFERS, etc., only (or any combination thereof), and in order to do so it wouldn't be desirable to need to keep splitting off the types like this.

CydeWeys:
Oh, and by abandoning the PR and starting anew, you're losing the history of discussion on this existing PR.  It would probably be better to continue using this PR, but to change the code in it to implement the new idea.

PR #1954, opened 2023-03-06
Double the number of frontend instances on production from 12 to 24

CydeWeys:
It seems like we're hitting App Engine capacity issues resulting in actual pages now (for whatever reason, but likely one customer), and we obviously don't want that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1954)

### DIFF
--- a/core/src/main/java/google/registry/env/production/default/WEB-INF/appengine-web.xml
+++ b/core/src/main/java/google/registry/env/production/default/WEB-INF/appengine-web.xml
@@ -7,7 +7,7 @@
   <sessions-enabled>true</sessions-enabled>
   <instance-class>B4_1G</instance-class>
   <manual-scaling>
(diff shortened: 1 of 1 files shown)


jianglai (approved):

ptkach:
Now, that we know what happened, do we still want to double the number of instances? What we probably want is a rate limiting / throttling instead.

CydeWeys (commented):
We can always reduce the number of instances back down again once rate limiting and throttling is up and running, but until then, we don't want unnecessary pages.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1954#-:-NPsjDL21gxLizYajQug:b-bf2tpz)* status: 0 of 1 files reviewed, all discussions resolved

PR #1955, opened 2023-03-09
Use builder image as a base for schema-deployer and schema-verifier

ptkach:
See b/270836911 for more details

Tested it locally by running both images with no issues

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1955)

### DIFF
--- a/release/builder/Dockerfile
+++ b/release/builder/Dockerfile
@@ -29,6 +29,13 @@ RUN go build -o /cloudSchedulerDeployer
 
 FROM marketplace.gcr.io/google/debian10
 ENV DEBIAN_FRONTEND=noninteractive LANG=en_US.UTF-8
+# Add script for cloud scheduler deployer
 COPY --from=cloudSchedulerBuilder /cloudSchedulerDeployer /usr/local/bin/cloudSchedulerDeployer
+# Add Cloud sql connector
+ADD https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 \
+    /usr/local/bin/cloud_sql_proxy
+RUN chmod +x /usr/local/bin/cloud_sql_proxy
+
 ADD ./build.sh .
 RUN ["bash", "./build.sh"]
+
--- a/release/builder/build.sh
+++ b/release/builder/build.sh
@@ -40,9 +40,16 @@ apt-get install lsb-release -y
 export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)"
 echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" \
   | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
-curl https://packages.cloud.google.com/apt/doc/apt-key.gpg \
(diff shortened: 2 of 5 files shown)


ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1955)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @jianglai and @weiminyu)

___
*[`-- commits` line 2 at r1](https://reviewable.io/reviews/google/nomulus/1955#-NQQGVAd9HjATou3tZPa):*

Please give a more descriptive PR description. Something like "Use builder as base image for ..."
</blockquote></details>

Done.
___
*[`release/cloudbuild-release.yaml` line 110 at r1](https://reviewable.io/reviews/google/nomulus/1955#-NQQGlRkDhZiycsDBm7p:-NQR1ltv4feD9vuk-V7Y:b-hu95jv) ([raw file](https://github.com/google/nomulus/blob/4cdbf8c0fb4d314aaa2ace56776c0c79416ae4c7/release/cloudbuild-release.yaml#L110)):*

Use braces for consistency. Same below.

Also, using `--build-arg` seems to be a better approach than using `sed` to replace the variables (in the previous step). We should considering to move our existing use case to `--build-arg` as well. (In a separate PR of course).
</blockquote></details>

braces updated, for sed I will create a task for myself
___
*[`release/builder/build.sh` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1955#-NQQHSpp6Q0l65jN7sE5:-NQR38gT3t8-5C6apHrN:bz17p4w) ([raw file](https://github.com/google/nomulus/blob/4cdbf8c0fb4d314aaa2ace56776c0c79416ae4c7/release/builder/build.sh#L45)):*

Did you try this? The builder base image is Debian 10. Is adding the custom APT repo still necessary?
</blockquote></details>

It'd been Debian 10  with this apt-key before my change, so I was under assumption that whoever made the change kept it for a reason. It works as is though, but I will make a separate change to check if it's still needed

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1955)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach and @weiminyu)

___
*[`release/builder/build.sh` line 45 at r1](https://reviewable.io/reviews/google/nomulus/1955#-NQQHSpp6Q0l65jN7sE5:-NQR4j-LEF4x67SelQe3:b-5r6ngp) ([raw file](https://github.com/google/nomulus/blob/4cdbf8c0fb4d314aaa2ace56776c0c79416ae4c7/release/builder/build.sh#L45)):*

It'd been Debian 10  with this apt-key before my change, so I was under assumption that whoever made the change kept it for a reason. It works as is though, but I will make a separate change to check if it's still needed
</blockquote></details>

FWIW it was ubuntu1804, which started breaking as 1804 is EOL. So I moved the base image to debian for the two two schema images. I didn't check if the added repo is still necessary when I changed the base image. (I would not be surprised if it still is needed, as we all know how "stable" debian is.)

PR #1956, opened 2023-03-09
Move CloudTasksUtils to core/ project

gbrodman:
This does nothing for now, but in the future this will allow us to refer to the RegistryConfig and/or Service objects from the core project. This will be necessary when changing CloudTasksUtils to not use the AppEngine built-in connection (it will need to use a standard HTTP request instead).

This allows us, in this PR, to use the Service object in CloudTasksUtils which is nicer than plain strings. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1956)


gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1956)* status: 0 of 27 files reviewed, all discussions resolved (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/config/CloudTasksUtilsModule.java` line 50 at r1](https://reviewable.io/reviews/google/nomulus/1956#-NQ7A4c7PMkY7bB1Susg-r1-50:-NQBu8VA1mM4xguC85_i:b-896fix) ([raw file](https://github.com/google/nomulus/blob/94cb30e415ef8460d6349a1387bb5682be612ab8/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java#L50)):*

## Useless parameter

The parameter 'iapClientId' is never used.

[Show more details](https://github.com/google/nomulus/security/code-scanning/770)
</blockquote></details>

Done.

jianglai (approved):
I vaguely remembered that there was a reason `CloudTasksUtils` was in the `util` subpackage, because I faced the same issue with not having access to `RegistryConfig`... But I forgot what the reason was. Maybe because the predecessor `TaskQueueUtils` was in that subproject...? 

I guess as long as we don't think there's a chance that some other subproject would depend on `CloudTasksUtils` (therefore creating a cyclic dependency on `core`), it's fine for it to live inside `core`.

> Reviewed 25 of 27 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1956#-:-NQQJ4gvEx2p_BpwpQrm:bt2jdnb)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/config/CloudTasksUtilsModule.java` line 38 at r2](https://reviewable.io/reviews/google/nomulus/1956#-NQQK3sK7sb61UWlG3I-:-NQQK3sK7sb61UWlG3I0:bc7q4sm) ([raw file](https://github.com/google/nomulus/blob/05cf5ac22f42f044a51e526b4367ca9e3461c39f/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java#L38)):*
> ```Java
>  * A {@link Module} that provides {@link CloudTasksUtils}.
>  *
>  * <p>The class itself cannot be annotated as {@code Inject} because its requested dependencies use
> ```

I think you can remove the provider for `CloudTasksUtils` now.



_Code quote:_
```Java
 * <p>The class itself cannot be annotated as {@code Inject} because its requested dependencies use
 * the {@link Config} qualifier which is not available in the {@code util} package.
```

gbrodman (commented):
Maybe it was the predecessor? 

If there were any circular dependency the build would fail, right? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1956#-:-NQRSemw3KKgQ4Yio4_A:bupc9hn)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/config/CloudTasksUtilsModule.java` line 38 at r2](https://reviewable.io/reviews/google/nomulus/1956#-NQQK3sK7sb61UWlG3I-:-NQRWSJsEAqR1nrCazQL:b9iv5di) ([raw file](https://github.com/google/nomulus/blob/05cf5ac22f42f044a51e526b4367ca9e3461c39f/core/src/main/java/google/registry/config/CloudTasksUtilsModule.java#L38)):*

I think you can remove the provider for `CloudTasksUtils` now.
</blockquote></details>

Yep, definitely. We'll keep the rest of the module around though, better for testing.

PR #1958, opened 2023-03-10
Add allowedEppActions to AllocationToken Java classes

sarahcaseybot:
Schema change is in go/r3pr/1957 Original plan was to implement this using varying Token Types, but decided instead to add this field. See discussion in go/r3pr/1952.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1958)

PR #1961, opened 2023-03-13
Add pricing logic for allocation tokens in domain renew

sarahcaseybot:


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1961)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 413 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQbC4jY2I2zRsksNe7m:-NQbC4jY2I2zRsksNe7n:b7r6l5u) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L413)):*
> ```Java
>                 .setTargetId(recurring.getTargetId())
>                 .build();
>       } catch (EppException e) {
> ```

Let's move the comment explaining the exception handling here. 

Also, doesn't this mean that we'd only drop the one date from the expansion? It's a bit unclear to me why this would happen or what the exact series of actions are in this circumstance (now and in the future) 
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 262 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBJXc6IWtJaIcbmaD:-NQgBJXdEE9jZ9y9Ui6e:bwt08xd) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L262)):*
> ```Java
>   }
> 
>   private Money getDomainCostWithDiscount(
> ```

Unless I'm mistaken, we can continue just passing in the `DomainPrices`object into this method instead of having multiple parameters pulled from it (the upside is that we don't have a boolean param either) 
___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 175 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBdua2nEVQfJiW1Xg:-NQgBdua2nEVQfJiW1Xh:barso79) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L175)):*
> ```Java
>     // Loads the target resource if it exists
>     Domain existingDomain = loadAndVerifyExistence(Domain.class, targetId, now);
>     // TODO(sarahbot@): Add check for valid EPP actions on the token once PR #1958 is merged
> ```

1958 is merged now right? 
___
*[`core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java` line 594 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgCKkx-5Wsvi2wCyti:-NQgCKkx-5Wsvi2wCytj:bcmssw6) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java#L594)):*
> ```Java
>                 persistDomainAndSetRecurringBillingEvent(
>                     "standard.example", SPECIFIED, Optional.of(Money.of(USD, 1))),
>                 Optional.of(allocationToken)))
> ```

It seems like there's a mismatch here between the renewal price behavior we pass into the `persistDomain...` method (SPECIFIED) and the one that's actually on the allocation token (this might happen in other places too, I'm not sure) 

Is this a particularly bad thing? We probably want to not do that, at the very least, I don't know if it has any repercussions 
___
*[`core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java` line 616 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBlGw5lbhveBlva-3:-NQgBlGw5lbhveBlva-4:b-qikxtq) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/test/java/google/registry/flows/domain/DomainRenewFlowTest.java#L616)):*
> ```Java
>             ImmutableMap.of("DOMAIN", "example.tld", "EXDATE", "2002-04-03T22:00:00.0Z")));
>     assertThat(DatabaseHelper.loadByEntity(allocationToken).getRedemptionHistoryId()).isPresent();
>     BillingEvent.OneTime billingEvent =
> ```

fyi i don't really care much either way but if you want you can use the DatabaseHelper methods (e.g. `loadAllOf`) to wrap the transactions so we don't have to create all these transactions in tests

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1961)* status: 1 of 8 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java` line 413 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQbC4jY2I2zRsksNe7m:-NQkyCAxDA1WTFlgUA8S:bvhzir0) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/beam/billing/ExpandRecurringBillingEventsPipeline.java#L413)):*

Let's move the comment explaining the exception handling here. 

Also, doesn't this mean that we'd only drop the one date from the expansion? It's a bit unclear to me why this would happen or what the exact series of actions are in this circumstance (now and in the future) 
</blockquote></details>

Really the only reason this needed to be added was due to the allocation token check, but since this is always passing in an empty allocation token, there shouldn't be an EPPException thrown by the getRenewPrice method. Changed to a more exact exception and indicated in comments that the catch should not be reached.
___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 262 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBJXc6IWtJaIcbmaD:-NQl1CtLDsOx_pCASWUN:bl3ygq1) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L262)):*

Unless I'm mistaken, we can continue just passing in the `DomainPrices`object into this method instead of having multiple parameters pulled from it (the upside is that we don't have a boolean param either) 
</blockquote></details>

I needed to separate the isPremium variable out because in the case of names with the NONPREMIUM renewalPriceBehavior they should get the price as if they are not premium even though DomainPrices has them labeled as premium.
___
*[`core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java` line 175 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBdua2nEVQfJiW1Xg:-NQl1guU7313hK1WfSuR:b-9i5oul) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/flows/domain/DomainRenewFlow.java#L175)):*

1958 is merged now right? 
</blockquote></details>

yes, I just merged it yesterday. However, to keep this PR from getting to large, I will add these checks in a separate PR.
___
*[`core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java` line 594 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgCKkx-5Wsvi2wCyti:-NQl28WvAk-5hTMK2Loz:b601kc5) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java#L594)):*

It seems like there's a mismatch here between the renewal price behavior we pass into the `persistDomain...` method (SPECIFIED) and the one that's actually on the allocation token (this might happen in other places too, I'm not sure) 

Is this a particularly bad thing? We probably want to not do that, at the very least, I don't know if it has any repercussions 
</blockquote></details>

Yes, we went back and forth on what the behavior in this case should be a lot in the team meeting last week and over chat between Ben, Bruno, and I. We decided for now that discounts should not stack and since the SPECIFIED price is already a discounted price, the allocation token's price should not be applied to the SPECIFIED price.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1961)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 262 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBJXc6IWtJaIcbmaD:-NQlH4Rd5Qj500rSHnq3:b-4iazmv) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L262)):*

I needed to separate the isPremium variable out because in the case of names with the NONPREMIUM renewalPriceBehavior they should get the price as if they are not premium even though DomainPrices has them labeled as premium.
</blockquote></details>

ahhhh tricky tricky 

this is nitty, but we should use the primitive boolean here so that it's not possible to pass in null 
___
*[`core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java` line 594 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgCKkx-5Wsvi2wCyti:-NQlG7le6vAljdhh8biF:b-5wd0vf) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java#L594)):*

Yes, we went back and forth on what the behavior in this case should be a lot in the team meeting last week and over chat between Ben, Bruno, and I. We decided for now that discounts should not stack and since the SPECIFIED price is already a discounted price, the allocation token's price should not be applied to the SPECIFIED price.
</blockquote></details>

Yeah I remember that, which makes sense, what I'm referring to is more the structure of the tests themselves. Like, in most cases in tests I assume we'll want the token to have the same price behavior as the billing event just as a matter of "this is the case that should usually happen", then perhaps a separate test that uses a different status on the token to explicitly call out "even if the token has a different price behavior, we should use the one on the billing event". Does that make sense / seem reasonable?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1961)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 262 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgBJXc6IWtJaIcbmaD:-NR-w0SP2sRnDnhzMboq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L262)):*

ahhhh tricky tricky 

this is nitty, but we should use the primitive boolean here so that it's not possible to pass in null 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java` line 594 at r1](https://reviewable.io/reviews/google/nomulus/1961#-NQgCKkx-5Wsvi2wCyti:-NR0-jQ21WxQAY13Jc5e:bxgrs1r) ([raw file](https://github.com/google/nomulus/blob/386f8e23547787519ae942c2e44c78fb3a674b31/core/src/test/java/google/registry/flows/domain/DomainPricingLogicTest.java#L594)):*

Yeah I remember that, which makes sense, what I'm referring to is more the structure of the tests themselves. Like, in most cases in tests I assume we'll want the token to have the same price behavior as the billing event just as a matter of "this is the case that should usually happen", then perhaps a separate test that uses a different status on the token to explicitly call out "even if the token has a different price behavior, we should use the one on the billing event". Does that make sense / seem reasonable? 
</blockquote></details>

Added a test for this

gbrodman (approved):
> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1961)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #1970, opened 2023-03-16
Add tool changes to modify allowedEppActions on allocation tokens

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1970)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/fee/FeeQueryCommandExtensionItem.java
+++ b/core/src/main/java/google/registry/model/domain/fee/FeeQueryCommandExtensionItem.java
@@ -43,8 +43,9 @@ public abstract class FeeQueryCommandExtensionItem extends ImmutableObject {
 
     public static CommandName parseKnownCommand(String string) {
       try {
-        checkArgument(!string.equals("UNKNOWN"));
-        return valueOf(string);
+        CommandName command = valueOf(string);
+        checkArgument(!command.equals(UNKNOWN));
+        return command;
       } catch (IllegalArgumentException e) {
         throw new IllegalArgumentException(
             "Invalid EPP action name. Valid actions are CREATE, RENEW, TRANSFER, RESTORE, and"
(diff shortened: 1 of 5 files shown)


gbrodman (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1970)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/fee/FeeQueryCommandExtensionItem.java` line 42 at r2](https://reviewable.io/reviews/google/nomulus/1970#-NQlKH-g8tbuFkqNPidt:-NQlKH-g8tbuFkqNPidu:b-ha153e) ([raw file](https://github.com/google/nomulus/blob/eb911211e583f882b118f3b3fa6dfe9ec5a05f6d/core/src/main/java/google/registry/model/domain/fee/FeeQueryCommandExtensionItem.java#L42)):*
> ```Java
>     UPDATE;
> 
>     public static CommandName fromString(String string) {
> ```

Is the point of this method (versus just using valueOf) that we don't want to include UNKNOWN in the error message?  I guess I'm not sure where UNKNOWN command names would normally be used, but if we're trying to prevent the use of them then we should probably also 
1. not allow users to pass in the string "UNKNOWN" 
2. change the name to be something like `parseKnownCommand" or something? idk

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1970)* status: all files reviewed, 2 unresolved discussions (waiting on @github-code-scanning[bot] and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/domain/fee/FeeQueryCommandExtensionItem.java` line 42 at r2](https://reviewable.io/reviews/google/nomulus/1970#-NQlKH-g8tbuFkqNPidt:-NR9khlf0OxV0wOlaOgU:bcwr2f6) ([raw file](https://github.com/google/nomulus/blob/eb911211e583f882b118f3b3fa6dfe9ec5a05f6d/core/src/main/java/google/registry/model/domain/fee/FeeQueryCommandExtensionItem.java#L42)):*

I originally just made this to create a neater error message, but good point that this should also prevent UNKNOWN commands.
</blockquote></details>

oh, is the error message that much nicer? Either way, I might recommend 
1. allowing any exception thrown by valueOf to propagate down, we don't have to rethrow it here 
2. attempting to do the parsing first then comparing it to the UNKNOWN enum value rather than string comparison

PR #1971, opened 2023-03-17
Add SQL-based DNS refresh processing mechanism

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1971)


jianglai (commented):
> Reviewed 10 of 12 files at r1, 7 of 8 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
Ping.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971#-:-NSBfB9t6UpPXh9Hvf9i:b-82oqa1)* status: 17 of 18 files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 8 of 12 files at r1, 6 of 8 files at r2, 1 of 2 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: all files reviewed, 5 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 82 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCUE3PDwoNGL8C5W_E:-NSCUE3PDwoNGL8C5W_F:b-5t7ai2) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L82)):*
> ```Java
>    * <ul>
>    *   <li>They are for the given TLD.
>    *   <li>Their request time not in the future.
> ```

nit: "Their request time <is> not in the future"
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 86 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCYgpc4fBpxQ_HL-um:-NSCYgpc4fBpxQ_HL-un:b8xmh8d) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L86)):*
> ```Java
>    * </ul>
>    */
>   public ImmutableList<DnsRefreshRequest> processRequests(
> ```

bit nitty, but w/r/t the method name this isn't actually processing the requests, just reading them 
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCXLJX4nX-NYkVej7A:-NSCXLJX4nX-NYkVej7B:b-3ihfwa) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L102)):*
> ```Java
>                       .setMaxResults(batchSize)
>                       .getResultStream()
>                       .map(e -> e.updateProcessTime(transactionTime))
> ```

It seems a bit odd to update the processed-time here when we're loading the requests, rather than after we've actually processed them -- what if they fail? 

Technically wouldn't the processed-time also be slightly later, when we actually enqueue the requests?
___
*[`core/src/main/java/google/registry/dns/ReadDnsRefreshRequestsAction.java` line 117 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSC_vSb7FtNfI_LbmL8:-NSC_vSb7FtNfI_LbmL9:b8swsqe) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/ReadDnsRefreshRequestsAction.java#L117)):*
> ```Java
>           dnsUtils.processRequests(tld, requestedMaximumDuration, processBatchSize);
>       logger.atInfo().log("Read %d DNS update requests for TLD %s.", requests.size(), tld);
>       if (!requests.isEmpty()) {
> ```

doesn't seem like this "if" block is necessary, because the method will just do nothing if it is empty
___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 143 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCTElv846z08iJvrZz:-NSCTElv846z08iJvr_-:b-wqven9) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L143)):*
> ```XML
>   <task>
>     <url>
>       <![CDATA[/_dr/cron/fanout?queue=dns-refresh&forEachRealTld&ForEachTestTld&endpoint=/_dr/task/readDnsRefreshRequests&dnsJitterSeconds=45]]></url>
> ```

Do we currently refresh DNS for test TLDs? I wouldn't expect that that's necessary generally 
___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 148 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCT1-nAiqF8TMct22D:-NSCT1-nAiqF8TMct22E:b2y3g41) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L148)):*
> ```XML
>       Enqueue a ReadDnsRefreshRequestAction for each TLD.
>     </description>
>     <schedule>*/1 * * * *</schedule>
> ```

out of curiosity is this just every minute?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 82 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCUE3PDwoNGL8C5W_E:-NSGwAhJ2Rm-NCnLDbu1:boi32lc) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L82)):*

nit: "Their request time <is> not in the future"
</blockquote></details>

errr, I think you might mistyped. That is exactly what the line currently says.
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 86 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCYgpc4fBpxQ_HL-um:-NSGwNYG6-FxnqzOktJ0:b-r6jeir) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L86)):*

bit nitty, but w/r/t the method name this isn't actually processing the requests, just reading them 
</blockquote></details>

The issue with naming the method `readRequests` is that it implies the method is immutable. But in fact the requests are read, updated, and then written back to the db with the updated process time. So in a sense these requests are indeed processed. We can name it `readAndUpdateRequests`, but `processRequests` seems to be doing a good enough job. Plus there is also the column `last_process_time` the corresponds to the method.
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCXLJX4nX-NYkVej7A:-NSGxdxb7l7pE2PtOQno:btflum3) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L102)):*

It seems a bit odd to update the processed-time here when we're loading the requests, rather than after we've actually processed them -- what if they fail? 

Technically wouldn't the processed-time also be slightly later, when we actually enqueue the requests?
</blockquote></details>

This timestamp is only used to ensure that a recently processed (read) request is not immediately re-processed in a subsequent cycle, should some error happened in the downstream that resulted in the requests not being removed from the database. It is necessary because we order by request time in ascending order when reading the pending requests, so if we had stale records they would be repeatedly retried in succession. The idea is that once a record is read, it will have a cool down period before it is retried again.

After the tasks are enqueued, the records should be deleted.
___
*[`core/src/main/java/google/registry/dns/ReadDnsRefreshRequestsAction.java` line 117 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSC_vSb7FtNfI_LbmL8:-NSGzKnN69zaNxAfy7yV:b-2tpu7l) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/ReadDnsRefreshRequestsAction.java#L117)):*

doesn't seem like this "if" block is necessary, because the method will just do nothing if it is empty
</blockquote></details>

It saves a call to `Registry.get(tld).getNumDnsPublishLocks()` at the expense of a conditional. I guess it's a wash. Removed the if block.
___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 143 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCTElv846z08iJvrZz:-NSGuuvc2lNpZbLwJqFb:b81qd77) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L143)):*

Do we currently refresh DNS for test TLDs? I wouldn't expect that that's necessary generally 
</blockquote></details>

Currently we use `ReadDnsQueueAction` that leases all pending DNS update tasks and batches them by TLD. That action does not treat test TLDs any differently. The new action is designed to work on a per-TLD basis, so the TLD needs to be provided as a query parameter. I think we do need to include test TLDs because prober TLDs  are test TLDs and the probers do look at if DNS updates are successful.
___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 148 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCT1-nAiqF8TMct22D:-NSGvnTr7OEhviU7SOoF:b-5jpocz) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L148)):*

out of curiosity is this just every minute? 
</blockquote></details>

Correct.

See: https://crontab.guru/#*/1_*_*_*_*



_Code quote:_
```XML
*/1 * * * *<
```

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 82 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCUE3PDwoNGL8C5W_E:-NSHDtOgE97c0TW-7zyB:b-jufnzp) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L82)):*

errr, I think you might mistyped. That is exactly what the line currently says.
</blockquote></details>

haha yup there's meant to be an "is" before "not" 
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 86 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCYgpc4fBpxQ_HL-um:-NSIGQfI4t_VF_SK0r75:b-5oie5z) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L86)):*

The issue with naming the method `readRequests` is that it implies the method is immutable. But in fact the requests are read, updated, and then written back to the db with the updated process time. So in a sense these requests are indeed processed. We can name it `readAndUpdateRequests`, but `processRequests` seems to be doing a good enough job. Plus there is also the column `last_process_time` the corresponds to the method.
</blockquote></details>

i mean, it means that in the caller of this we call DnsUtils.processRequests then immediately call another `processRequests` method, meaning that at the very least one or both of them should be renamed. 

The point of "this method doesn't actually process the requests" still stands, since all it does is update one timestamp. 
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCXLJX4nX-NYkVej7A:-NSIGu9vBmlSi4pJHrYN:bnfudc6) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L102)):*

This timestamp is only used to ensure that a recently processed (read) request is not immediately re-processed in a subsequent cycle, should some error happened in the downstream that resulted in the requests not being removed from the database. It is necessary because we order by request time in ascending order when reading the pending requests, so if we had stale records they would be repeatedly retried in succession. The idea is that once a record is read, it will have a cool down period before it is retried again.

After the tasks are enqueued, the records should be deleted.
</blockquote></details>

hmm yeah I see, can we add a small comment here explaining that this is to essentially act as a cooldown? it's not immediately clear from reading it that that's the case 
___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 143 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCTElv846z08iJvrZz:-NSIGzcXCX4oDG8iaF1X:b-fd3w1c) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L143)):*

Currently we use `ReadDnsQueueAction` that leases all pending DNS update tasks and batches them by TLD. That action does not treat test TLDs any differently. The new action is designed to work on a per-TLD basis, so the TLD needs to be provided as a query parameter. I think we do need to include test TLDs because prober TLDs  are test TLDs and the probers do look at if DNS updates are successful.
</blockquote></details>

sounds good to me!

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 82 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCUE3PDwoNGL8C5W_E:-NSL_Jd0EQDoistYwgbU:b-7ve0yx) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L82)):*

haha yup there's meant to be an "is" before "not" 
</blockquote></details>

Ah, got you. Fixed.
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 86 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCYgpc4fBpxQ_HL-um:-NSLbJog9J8ZdUlri0yp:bz7p16l) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L86)):*

i mean, it means that in the caller of this we call DnsUtils.processRequests then immediately call another `processRequests` method, meaning that at the very least one or both of them should be renamed. 

The point of "this method doesn't actually process the requests" still stands, since all it does is update one timestamp. 
</blockquote></details>

Updated the method name.
___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSCXLJX4nX-NYkVej7A:-NSLaeaQ2jVpiwqkYgG7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/DnsUtils.java#L102)):*

hmm yeah I see, can we add a small comment here explaining that this is to essentially act as a cooldown? it's not immediately clear from reading it that that's the case 
</blockquote></details>

Done.

jianglai (commented):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman)

gbrodman (approved):
> Reviewed 1 of 1 files at r4, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
> Reviewed 1 of 1 files at r5, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1971)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/dns/ReadDnsRefreshRequestsAction.java` line 117 at r5](https://reviewable.io/reviews/google/nomulus/1971#-NSC_vSb7FtNfI_LbmL8:-NSSP74s4HvuVsS8dad6:bf4eo97) ([raw file](https://github.com/google/nomulus/blob/50e74d27966825fd28a4ee694b0f4763073caa52/core/src/main/java/google/registry/dns/ReadDnsRefreshRequestsAction.java#L117)):*

It saves a call to `Registry.get(tld).getNumDnsPublishLocks()` at the expense of a conditional. I guess it's a wash. Removed the if block.
</blockquote></details>

Actually it is easier for testing purpose to have the check there. Otherwise instead of simply checking if `processRequests` is called, we have to capture the argument and check if it is empty. Reverted.

PR #1972, opened 2023-03-22
Check allowedEppActions when validating tokens

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1972)

### DIFF
--- a/core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml
+++ b/core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml
@@ -19,7 +19,7 @@
           </fee:object>
           <fee:command name="create">
             <fee:period unit="y">1</fee:period>
-            <fee:fee description="create">6.50</fee:fee>
+            <fee:fee description="create">13.00</fee:fee>
           </fee:command>
         </fee:cd>
         <fee:cd>
@@ -28,7 +28,7 @@
           </fee:object>
           <fee:command name="renew">
             <fee:period unit="y">1</fee:period>
-            <fee:fee description="renew">11.00</fee:fee>
+            <fee:class>Token Not Supported</fee:class>
           </fee:command>
         </fee:cd>
         <fee:cd>
@@ -46,7 +46,7 @@
           </fee:object>
           <fee:command name="restore">
             <fee:period unit="y">1</fee:period>
-            <fee:fee description="restore">17.00</fee:fee>
+            <fee:class>Token Not Supported</fee:class>
           </fee:command>
         </fee:cd>
         <fee:cd>
@@ -55,7 +55,7 @@
           </fee:object>
           <fee:command name="update">
             <fee:period unit="y">1</fee:period>
-            <fee:fee description="update">0.00</fee:fee>
+            <fee:class>Token Not Supported</fee:class>
           </fee:command>
         </fee:cd>
       </fee:chkData>
--- a/core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml
+++ b/core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml
@@ -28,7 +28,7 @@
           </fee:object>
           <fee:command name="renew">
             <fee:period unit="y">1</fee:period>
-            <fee:class>Token Not Supported</fee:class>
+            <fee:class>token-not-supported</fee:class>
           </fee:command>
         </fee:cd>
         <fee:cd>
@@ -46,7 +46,7 @@
           </fee:object>
           <fee:command name="restore">
             <fee:period unit="y">1</fee:period>
-            <fee:class>Token Not Supported</fee:class>
+            <fee:class>token-not-supported</fee:class>
           </fee:command>
         </fee:cd>
         <fee:cd>
@@ -55,7 +55,7 @@
           </fee:object>
           <fee:command name="update">
             <fee:period unit="y">1</fee:period>
-            <fee:class>Token Not Supported</fee:class>
+            <fee:class>token-not-supported</fee:class>
           </fee:command>
         </fee:cd>
       </fee:chkData>
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -42,9 +42,7 @@ import com.google.common.flogger.FluentLogger;
 import com.google.common.net.InternetDomainName;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.flows.EppException;
-import google.registry.flows.EppException.AssociationProhibitsOperationException;
 import google.registry.flows.EppException.ParameterValuePolicyErrorException;
-import google.registry.flows.EppException.StatusProhibitsOperationException;
 import google.registry.flows.ExtensionManager;
 import google.registry.flows.Flow;
 import google.registry.flows.FlowModule.RegistrarId;
@@ -55,6 +53,11 @@ import google.registry.flows.custom.DomainCheckFlowCustomLogic.BeforeResponsePar
 import google.registry.flows.custom.DomainCheckFlowCustomLogic.BeforeResponseReturnData;
 import google.registry.flows.domain.token.AllocationTokenDomainCheckResults;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils;
+import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotInPromotionException;
+import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForCommandException;
+import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForDomainException;
+import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForRegistrarException;
+import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForTldException;
 import google.registry.model.EppResource;
 import google.registry.model.ForeignKeyUtils;
 import google.registry.model.billing.BillingEvent;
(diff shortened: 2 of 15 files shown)


core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml
```diff
@@ -0,0 +1,68 @@
...
+            <fee:fee description="renew">11.00</fee:fee>
+          </fee:command>
+        </fee:cd>
+        <fee:cd>
+          <fee:object>
+            <domain:name>example1.tld</domain:name>
+          </fee:object>
+          <fee:command name="restore">
+            <fee:period unit="y">1</fee:period>
+            <fee:class>Token Not Supported</fee:class>
```
CydeWeys, line 49: This is inconsistent with the existing fee:classes, which are all of the form lower-case-hyphenated.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: 0 of 17 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml` line 49 at r1](https://reviewable.io/reviews/google/nomulus/1972#-NRFWviLL0EiRtniFiIj-r1-49:-NRJdwP5DnXyzol4u3AK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/adde52a259db9cbbd431d16882a34957e428faf2/core/src/test/resources/google/registry/flows/domain/domain_check_fee_multiple_commands_allocationtoken_response_v12.xml#L49)):*

This is inconsistent with the existing fee:classes, which are all of the form lower-case-hyphenated.
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 12 of 17 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 308 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKPhkg4aoyqhCTFrYe:-NRKPhkg4aoyqhCTFrYf:bui1xyx) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L308)):*
> ```Java
>               recurrences.getOrDefault(domainName, null));
>           responseItems.add(builder.setDomainNameIfSupported(domainName).build());
>         } catch (AssociationProhibitsOperationException | StatusProhibitsOperationException e) {
> ```

what's the type of issue that could cause this? We probably want to explain that in a comment at the start of the catch block so that it's clear how this might come about to any readers. 
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 306 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKQWXBAe3w6FlO3RGD:-NRKQWXBAe3w6FlO3RGE:b-rl9lbt) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L306)):*
> ```Java
>       extends AssociationProhibitsOperationException {
>     AllocationTokenNotValidForCommandException() {
>       super("Allocation token not valid for the EPP command");
> ```

This can only happen on default tokens, right? Could it possibly be confusing to the user if they try to do something without specifying an allocation token and they get this message? 
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 881 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKR8VnEeLT2h19v1Ze:-NRKR8VnEeLT2h19v1Zf:b74m9f2) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L881)):*
> ```Java
> 
>   @Test
>   void testFeeExtension_multipleCommands_tokenNotValidForAll_v06() throws Exception {
> ```

apologies for the nittiness, but can we use "some" instead of "all" in the test name? "all" can be a bit ambiguous (could mean either "some aren't valid" or "all aren't valid") 

same below

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: 15 of 17 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 308 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKPhkg4aoyqhCTFrYe:-NRe1IdL23IVogV2q8r2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L308)):*

what's the type of issue that could cause this? We probably want to explain that in a comment at the start of the catch block so that it's clear how this might come about to any readers. 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 881 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKR8VnEeLT2h19v1Ze:-NRe1ViS9Z7P-0G5a5Cz:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L881)):*

apologies for the nittiness, but can we use "some" instead of "all" in the test name? "all" can be a bit ambiguous (could mean either "some aren't valid" or "all aren't valid") 

same below
</blockquote></details>

Done.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 308 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKPhkg4aoyqhCTFrYe:-NRt4MWf7w0i-KqVjBEG:budhfa8) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L308)):*

Done.
</blockquote></details>

In that case wouldn't it be the case that we only need to surround the call to AllocationTokenFlowUtils.validateToken with the try/catch block? 

Also, we'd want to catch the specific exceptions that we expect could be thrown, e.g. AllocationTokenNotValidForTldException

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 308 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKPhkg4aoyqhCTFrYe:-NS7IAtj0hMcEVB0OZf7:bdtdjdu) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L308)):*

No, I think we need to keep it this way because we do not want to call handleFeeRequest if the token is not valid. Instead, we just want to return a fee response item with the fee class of "token-not-supported" which is what's done in the catch block. 

AllocationTokenNotValidForTldException is not the only exception we want to catch. We want to catch all the AssociationProhibitsOperationExceptions and StatusProhibitsOperationExceptions since those would fail on actual EPP requests. This is actually an improvement from previous behavior. Before this change, you could pass in an invalid allocation token to the feeCheck command and it would still return a fee in the fee extension even though the actual create would fail. 
</blockquote></details>

1. i think it might be better to put a `continue` in the catch block, that way it's clear that `handleFeeRequest` isn't throwing the exception and it'll reduce indentation 
2. are those the actual exceptions that get thrown, or is a subclass of those like AllocationTokenNotValidForTldException? My point is that we should be as specific as possible 
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 306 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKQWXBAe3w6FlO3RGD:-NS7KznZ6bkukaRl-4uN:b-kz85lq) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L306)):*

No, this should not  happen with default tokens. This should happen if a token was created for only one (or some) command(s) and a user tries to use it for a different flow. Currently we only have tokens implemented for CREATE and RENEW, but could add more in the future. Perhaps if we want to temporarily discount RESTORE like we did during covid, we could create an allocation token making RESTORE free, but that token should not work on CREATE and RENEW.
</blockquote></details>

yeah right I think I misread this, my bad

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java` line 308 at r2](https://reviewable.io/reviews/google/nomulus/1972#-NRKPhkg4aoyqhCTFrYe:-NS7pJXp9KMmR9n_ur0m:b-wtesft) ([raw file](https://github.com/google/nomulus/blob/9b4c2e2ec489ba7b8395e20a70585517a5d93038/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java#L308)):*

1. i think it might be better to put a `continue` in the catch block, that way it's clear that `handleFeeRequest` isn't throwing the exception and it'll reduce indentation 
2. are those the actual exceptions that get thrown, or is a subclass of those like AllocationTokenNotValidForTldException? My point is that we should be as specific as possible 
</blockquote></details>

1. I don't think that the continue statement makes any sense as it does not do anything as the last line in the catch block and it will not reduce the indentation.
2. Changed to all exceptions that could be thrown

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1972)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

PR #1984, opened 2023-04-06
Allow a credit of 0 when deleting a domain during a grace period

gbrodman:
There can be situations (anchor tenants, test tokens, other ways of getting a domain to cost $0) where we may want to delete a domain during the add grace period but the credit applied is 0. We should not fail on those cases.

See b/277115241 for an example.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1984)


CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1984#-:-NSN7mWR90xgsHv0YNak:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

PR #1985, opened 2023-04-07
Add tool commands to modify TTLs on a TLD

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1985)

### DIFF
--- a/core/src/main/java/google/registry/model/tld/Registry.java
+++ b/core/src/main/java/google/registry/model/tld/Registry.java
@@ -785,13 +785,13 @@ public class Registry extends ImmutableObject implements Buildable, UnsafeSerial
       return this;
     }
 
-    public Builder setDnsNsAtl(Duration dnsNsAtl) {
-      getInstance().dnsNsTtl = dnsNsAtl;
+    public Builder setDnsNsTtl(Duration dnsNsTtl) {
+      getInstance().dnsNsTtl = dnsNsTtl;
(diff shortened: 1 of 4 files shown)


jianglai (approved):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1985)* status: all files reviewed, 3 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java` line 243 at r2](https://reviewable.io/reviews/google/nomulus/1985#-NSfUpmb5CF0X6DnYiV1:-NSfUpmb5CF0X6DnYiV2:bx0ewxz) ([raw file](https://github.com/google/nomulus/blob/94dfeeb6ad7feb6210bede54083e665fe229405e/core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java#L243)):*
> ```Java
>   @Parameter(
>       names = {"--dns_a_plus_aaaa_ttl"},
>       description = "The time to live for DNS A and AAAA records")
> ```

I think it might be worth it to include an example in the description. `PT240S` for example is not something that people would know on top of their mind.

Same for the other parameters below.



_Code quote:_
```Java
The time to live for DNS A and AAAA records
```

PR #1987, opened 2023-04-10
Add a command to update Recurrence objects' behavior

gbrodman:
We want to basically be able to change the renewal behavior, either setting the behavior type (e.g. NONPREMIUM) or the specified renewal price.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1987)

### DIFF
--- a/core/src/main/java/google/registry/tools/RegistryTool.java
+++ b/core/src/main/java/google/registry/tools/RegistryTool.java
@@ -108,6 +108,7 @@ public final class RegistryTool {
           .put("update_keyring_secret", UpdateKeyringSecretCommand.class)
           .put("update_package_promotion", UpdatePackagePromotionCommand.class)
           .put("update_premium_list", UpdatePremiumListCommand.class)
+          .put("update_recurrence", UpdateRecurrenceCommand.class)
           .put("update_registrar", UpdateRegistrarCommand.class)
           .put("update_reserved_list", UpdateReservedListCommand.class)
           .put("update_server_locks", UpdateServerLocksCommand.class)
--- a/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java
+++ b/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java
@@ -1,0 +1,195 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.model.IdService.allocateId;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.util.DateTimeUtils.END_OF_TIME;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import google.registry.model.EppResourceUtils;
+import google.registry.model.billing.BillingEvent.Recurring;
+import google.registry.model.billing.BillingEvent.RenewalPriceBehavior;
+import google.registry.model.domain.Domain;
+import google.registry.model.domain.DomainHistory;
+import google.registry.model.reporting.HistoryEntry;
+import google.registry.model.reporting.HistoryEntry.HistoryEntryId;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.joda.money.Money;
+import org.joda.time.DateTime;
+
+/**
+ * Command to update {@link Recurring} billing events with a new behavior and/or price.
+ *
+ * <p>More specifically, this closes the existing recurrence object and creates a new, similar,
+ * object as well as a corresponding synthetic {@link DomainHistory} object. This is done to
+ * preserve the recurrence's history.
+ */
+@Parameters(separators = " =", commandDescription = "Update a billing recurrence")
+public class UpdateRecurrenceCommand extends ConfirmingCommand {
+
+  private static final String HISTORY_REASON =
+      "Administrative update of billing recurrence behavior";
+
+  @Parameter(
+      description = "Domain name(s) for which we wish to update the recurrence(s)",
+      required = true)
+  private List<String> mainParameters;
+
+  @Nullable
+  @Parameter(
+      names = "--renewal_price_behavior",
+      description = "New RenewalPriceBehavior value to use with this recurrence")
+  RenewalPriceBehavior renewalPriceBehavior;
+
+  @Nullable
+  @Parameter(
+      names = "--specified_renewal_price",
+      description = "Exact renewal price to use if the behavior is SPECIFIED")
+  Money specifiedRenewalPrice;
+
+  @Override
+  protected String prompt() throws Exception {
+    checkArgument(
+        renewalPriceBehavior != null || specifiedRenewalPrice != null,
+        "Must specify a behavior and/or a price");
+
+    if (renewalPriceBehavior != null) {
+      if (renewalPriceBehavior.equals(RenewalPriceBehavior.SPECIFIED)) {
+        checkArgument(
+            specifiedRenewalPrice != null,
+            "Renewal price must be set when using SPECIFIED behavior");
+      } else {
+        checkArgument(
+            specifiedRenewalPrice == null,
+            "Renewal price can have a value if and only if the renewal price behavior is"
+                + " SPECIFIED");
+      }
+    }
+    ImmutableMap<Domain, Recurring> domainsAndRecurrings =
+        tm().transact(this::loadDomainsAndRecurrings);
+    if (renewalPriceBehavior == null) {
+      // Allow users to specify only a price only if all renewals are already SPECIFIED
+      domainsAndRecurrings.forEach(
+          (d, r) ->
+              checkArgument(
+                  r.getRenewalPriceBehavior().equals(RenewalPriceBehavior.SPECIFIED),
+                  "When specifying only a price, all domains must have SPECIFIED behavior. Domain"
+                      + " %s does not",
+                  d.getDomainName()));
+    }
+    String specifiedPriceString =
+        specifiedRenewalPrice == null ? "" : String.format(" and price %s", specifiedRenewalPrice);
+    return String.format(
+        "Update the following with behavior %s%s?\n%s",
+        renewalPriceBehavior,
+        specifiedPriceString,
+        Joiner.on('\n').withKeyValueSeparator(':').join(domainsAndRecurrings));
+  }
+
+  @Override
+  protected String execute() throws Exception {
+    ImmutableList<Recurring> newRecurrings = tm().transact(this::internalExecute);
+    return "Updated new recurring(s): " + newRecurrings;
+  }
+
+  private ImmutableList<Recurring> internalExecute() {
+    ImmutableMap<Domain, Recurring> domainsAndRecurrings = loadDomainsAndRecurrings();
+    DateTime now = tm().getTransactionTime();
+    ImmutableList.Builder<Recurring> resultBuilder = new ImmutableList.Builder<>();
+    domainsAndRecurrings.forEach(
+        (domain, existingRecurring) -> {
+          // Make a new history ID to break the (recurring, history, domain) circular dep chain
+          long newHistoryId = allocateId();
+          HistoryEntryId newDomainHistoryId = new HistoryEntryId(domain.getRepoId(), newHistoryId);
+          Recurring endingNow = existingRecurring.asBuilder().setRecurrenceEndTime(now).build();
+          Recurring.Builder newRecurringBuilder =
+              existingRecurring
+                  .asBuilder()
+                  // set the ID to be 0 (null) to create a new object
+                  .setId(0)
+                  .setDomainHistoryId(newDomainHistoryId);
+          if (renewalPriceBehavior != null) {
+            newRecurringBuilder.setRenewalPriceBehavior(renewalPriceBehavior);
+            newRecurringBuilder.setRenewalPrice(null);
+          }
+          if (specifiedRenewalPrice != null) {
+            newRecurringBuilder.setRenewalPrice(specifiedRenewalPrice);
+          }
+          Recurring newRecurring = newRecurringBuilder.build();
+          Domain newDomain =
+              domain.asBuilder().setAutorenewBillingEvent(newRecurring.createVKey()).build();
+          DomainHistory newDomainHistory =
+              new DomainHistory.Builder()
+                  .setRevisionId(newDomainHistoryId.getRevisionId())
+                  .setDomain(newDomain)
+                  .setReason(HISTORY_REASON)
+                  .setRegistrarId(domain.getCurrentSponsorRegistrarId())
+                  .setBySuperuser(true)
+                  .setRequestedByRegistrar(false)
+                  .setType(HistoryEntry.Type.SYNTHETIC)
+                  .setModificationTime(now)
+                  .build();
+          tm().putAll(endingNow, newRecurring, newDomain, newDomainHistory);
+          resultBuilder.add(newRecurring);
+        });
+    return resultBuilder.build();
+  }
+
+  private ImmutableMap<Domain, Recurring> loadDomainsAndRecurrings() {
+    ImmutableMap.Builder<Domain, Recurring> result = new ImmutableMap.Builder<>();
+    DateTime now = tm().getTransactionTime();
+    for (String domainName : mainParameters) {
+      Domain domain =
+          EppResourceUtils.loadByForeignKey(Domain.class, domainName, now)
+              .orElseThrow(
+                  () ->
+                      new IllegalArgumentException(
+                          String.format(
+                              "Domain %s does not exist or has been deleted", domainName)));
+      checkArgument(
+          domain.getDeletionTime().equals(END_OF_TIME),
+          "Domain %s has already had a deletion time set",
+          domainName);
+      checkArgument(
+          domain.getTransferData().isEmpty(),
+          "Domain %s has a pending transfer: %s",
+          domainName,
+          domain.getTransferData());
+      Optional<DateTime> domainAutorenewEndTime = domain.getAutorenewEndTime();
+      domainAutorenewEndTime.ifPresent(
+          endTime ->
+              checkArgument(
+                  endTime.isAfter(now),
+                  "Domain %s autorenew ended prior to now at %s",
+                  domainName,
+                  endTime));
+      Recurring recurring = tm().loadByKey(domain.getAutorenewBillingEvent());
+      checkArgument(
+          recurring.getRecurrenceEndTime().equals(END_OF_TIME),
+          "Domain %s's recurrence's end date is not END_OF_TIME",
+          domainName);
+      result.put(domain, recurring);
+    }
+    return result.build();
+  }
+}
--- a/core/src/test/java/google/registry/tools/UpdateRecurrenceCommandTest.java
+++ b/core/src/test/java/google/registry/tools/UpdateRecurrenceCommandTest.java
@@ -1,0 +1,222 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.tools;
+
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.truth.Truth8.assertThat;
+import static google.registry.testing.DatabaseHelper.createTld;
+import static google.registry.testing.DatabaseHelper.loadAllOf;
+import static google.registry.testing.DatabaseHelper.loadByEntity;
+import static google.registry.testing.DatabaseHelper.loadByKey;
+import static google.registry.testing.DatabaseHelper.persistActiveContact;
+import static google.registry.testing.DatabaseHelper.persistDomainWithDependentResources;
+import static google.registry.testing.DatabaseHelper.persistDomainWithPendingTransfer;
+import static google.registry.testing.DatabaseHelper.persistResource;
+import static google.registry.util.DateTimeUtils.END_OF_TIME;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import com.google.common.collect.Iterables;
+import google.registry.model.billing.BillingEvent.Reason;
+import google.registry.model.billing.BillingEvent.Recurring;
+import google.registry.model.billing.BillingEvent.RenewalPriceBehavior;
+import google.registry.model.domain.Domain;
+import google.registry.model.domain.DomainHistory;
+import google.registry.model.reporting.HistoryEntry;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.joda.money.CurrencyUnit;
+import org.joda.money.Money;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/** Tests for {@link UpdateRecurrenceCommand}. */
+public class UpdateRecurrenceCommandTest extends CommandTestCase<UpdateRecurrenceCommand> {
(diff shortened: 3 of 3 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 46 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgWRPK4ZxWWb33wNhu:-NSgWRPK4ZxWWb33wNhv:b-mch16o) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L46)):*
> ```Java
> 
>   @Parameter(
>       description = "ID(s) of the BillingRecurrence event to modify",
> ```

This would be much more useful if it took a domain name.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 54 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgWaAJ1oJJaX25tYRP:-NSgWaAJ1oJJaX25tYRQ:b-iahzfo) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L54)):*
> ```Java
>       names = "--renewal_price_behavior",
>       description = "New RenewalPriceBehavior value to use with this recurrence")
>   RenewalPriceBehavior renewalPriceBehavior;
> ```

These should either be tagged with `@Nullable`, as they are not required parameters, and the null case means that the existing state stays the same.

Or they should be marked as `required` in `@Parameter`.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 59 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgWqxaDea6IX1Ssjvk:-NSgWqxaDea6IX1Ssjvl:bkr69je) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L59)):*
> ```Java
>       names = "--specified_renewal_price",
>       description = "Exact renewal price to use if the behavior is SPECIFIED")
>   Money specifiedRenewalPrice;
> ```

This one should definitely be `@Nullable` as it's only applicable to `SPECIFIED`.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgX0-589449mLZwC9N:-NSgX0-589449mLZwC9O:b-flxpkp) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L64)):*
> ```Java
>   protected String prompt() throws Exception {
>     checkArgument(
>         renewalPriceBehavior.equals(RenewalPriceBehavior.SPECIFIED) ^ specifiedRenewalPrice == null,
> ```

For readability I would include the second clause in parentheses, as I don't know without having to look it up whether `==` takes precedence over `^`. (And if it doesn't, then this is a bug.)
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 81 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgW2bjCgOwQI8KNd1x:-NSgW2bjCgOwQI8KNd1y:b-6c8o0d) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L81)):*
> ```Java
> 
>   private ImmutableList<Recurring> internalExecute() {
>     ImmutableList<Recurring> existingRecurrings = loadRecurrings();
> ```

I think you need to be doing some additional checks here.  Your logic does not work correctly for domains with pending transfers (and potentially other pending states), and the easiest thing to do (rather than trying to handle that correctly) is simply to fail out on any pending states.

There also needs to be a test that the existing recurring is not already closed (this counts as pending state, e.g. maybe because a deletion has been requested).
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 91 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgXCnk-4rEfvaj3scj:-NSgXCnk-4rEfvaj3sck:b-wnvi42) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L91)):*
> ```Java
>               .setReason(
>                   String.format(
>                       "Updating recurring with ID %s to have behavior %s",
> ```

This isn't how `reason`s work in our codebase.  We have a small number of distinct reasons (42 non-null ones currently, and only 17 that are used by 10 or more resources).  But this would grow the # of reasons in an unbounded manner as the ID would always be unique.  It's not worth including the ID in the reason here because the ID is already in a different column in this row anyway.

It should be something simpler that is globally applicable to all edits performed by this command, something like "Administrator update of billing recurrence behavior"
___
*[`core/src/test/java/google/registry/tools/UpdateRecurrenceCommandTest.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgUMSuCQam9UHsbFlb:-NSgUMSuCQam9UHsbFlc:bexcqru) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/test/java/google/registry/tools/UpdateRecurrenceCommandTest.java#L64)):*
> ```Java
>             .findFirst()
>             .get();
>     assertThat(newRecurring.getRecurrenceEndTime()).isEqualTo(END_OF_TIME);
> ```

Add assertions about the `reason` et al in these tests.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 54 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgWaAJ1oJJaX25tYRP:-NSg_gybAjKud9ZZfJam:bcpi1f6) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L54)):*

These should either be tagged with `@Nullable`, as they are not required parameters, and the null case means that the existing state stays the same.

Or they should be marked as `required` in `@Parameter`.
</blockquote></details>

Ah, I see why it might be null -- if you're only changing the renewal price, but leaving the price behavior as `SPECIFIED`.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 54 at r1](https://reviewable.io/reviews/google/nomulus/1987#-NSgWaAJ1oJJaX25tYRP:-NSm3ZGz0D1pk-MS3c3s:b-jrtcc1) ([raw file](https://github.com/google/nomulus/blob/27107ed47e6073ccef72826b18749d7e76df29e9/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L54)):*

yep 
</blockquote></details>

yeah added more checks, basically either can be null but not both (and the behavior can only be null if you're setting an already-SPECIFIED price to be a different price)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 42 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqSxNd1pp24o0RnRdE:-NSqSxNd1pp24o0RnRdF:by6bl2b) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L42)):*
> ```Java
> 
> /**
>  * Command to update a {@link Recurring} billing event with a new behavior or price.
> ```

Should be `Command to update {@link Recurring} billing events ...`
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 116 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTe8G4DTpLojuzdl6:-NSqTe8G4DTpLojuzdl7:bslkux7) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L116)):*
> ```Java
>     ImmutableList.Builder<Recurring> resultBuilder = new ImmutableList.Builder<>();
>     for (Recurring existing : existingRecurrings) {
>       Domain domain = tm().loadByKey(VKey.create(Domain.class, existing.getDomainRepoId()));
> ```

Why are you doing this here when you're already loading the Domain earlier in `loadDomainsAndRecurrings()`?  You need to do some refactoring.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 163 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTwcX1FeqXcPEXmyb:-NSqTwcX1FeqXcPEXmyc:b-jmeyvs) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L163)):*
> ```Java
>       ImmutableSet<StatusValue> domainStatuses = domain.getStatusValues();
>       checkArgument(
>           domainStatuses.stream().noneMatch(s -> s.name().startsWith("PENDING")),
> ```

This isn't the best way to check for pending state.  Statuses exist solely for the purpose of EPP/WHOIS and aren't considered the canonical source of truth about resources in our code base.  You should be looking at `domain.getTransferData()` instead.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 167 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTLlf-dBo1JajJ4a9:-NSqTLlf-dBo1JajJ4aA:bkoqsih) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L167)):*
> ```Java
>           domainName,
>           domainStatuses);
>       Recurring recurring = tm().loadByKey(domain.getAutorenewBillingEvent());
> ```

I'm not seeing anywhere in the code above where you're re-saving the `Domain` with a modified autorenewBillingEvent pointing to the new one you're writing out?
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 168 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqToXp2skZ37zIQ-DZ:-NSqToXp2skZ37zIQ-D_:b6tfis3) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L168)):*
> ```Java
>           domainStatuses);
>       Recurring recurring = tm().loadByKey(domain.getAutorenewBillingEvent());
>       checkArgument(
> ```

You also need a `checkArgument` that the domain's deletionTime is `END_OF_TIME`.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 169 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqUWbW-mKGEq2X6asF:-NSqUWbW-mKGEq2X6asG:btffai) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L169)):*
> ```Java
>       Recurring recurring = tm().loadByKey(domain.getAutorenewBillingEvent());
>       checkArgument(
>           recurring.getRecurrenceEndTime().equals(DateTimeUtils.END_OF_TIME),
> ```

How about a check on the domain's `autorenewEndTime` as well?  That might be relevant.  Or would that info also be captured in the recurrence?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 112 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqVfj997MAQU0HG22E:-NSqVfj997MAQU0HG22F:b-xbjdxi) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L112)):*
> ```Java
> 
>   private ImmutableList<Recurring> internalExecute() {
>     ImmutableCollection<Recurring> existingRecurrings = loadDomainsAndRecurrings().values();
> ```

Yes, see, you're immediately throwing away the `Domain` objects themselves, which you shouldn't be doing, as you need them below.  It doesn't really ever make sense to have a function, used only in one place, that returns a map, but then you don't actually use both sides of the map.  It either shouldn't be returning a map (not correct for this use case), or you should actually be using the whole map.

gbrodman (commented):
i don't think we've sufficiently discussed how ever-so-slightly inconvenient it is for naming that the Java object is `Recurring` and the SQL table's name is `BillingRecurrence` 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987#-:-NSqkozLE6gteGRE02je:bmualuq)* status: 0 of 3 files reviewed, 7 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 42 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqSxNd1pp24o0RnRdE:-NSqh2l3-Y3cH7EEtMt2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L42)):*

Should be `Command to update {@link Recurring} billing events ...`
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 112 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqVfj997MAQU0HG22E:-NSqh4aqDdb_UDMNYYoK:b-geftsa) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L112)):*

Yes, see, you're immediately throwing away the `Domain` objects themselves, which you shouldn't be doing, as you need them below.  It doesn't really ever make sense to have a function, used only in one place, that returns a map, but then you don't actually use both sides of the map.  It either shouldn't be returning a map (not correct for this use case), or you should actually be using the whole map.
</blockquote></details>

yeah you're right, this was a screwup 
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 116 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTe8G4DTpLojuzdl6:-NSqh3i47P85s0j00I9r:b-8ol16a) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L116)):*

Why are you doing this here when you're already loading the Domain earlier in `loadDomainsAndRecurrings()`?  You need to do some refactoring.
</blockquote></details>

ach yeah 
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 163 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTwcX1FeqXcPEXmyb:-NSqj9fi8IgB2zlAabkS:b-8izxm9) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L163)):*

This isn't the best way to check for pending state.  Statuses exist solely for the purpose of EPP/WHOIS and aren't considered the canonical source of truth about resources in our code base.  You should be looking at `domain.getTransferData()` instead.
</blockquote></details>

yeah fair enough, and that plus a deletion time check would cover PENDING_TRANSFER and PENDING_DELETE statuses 
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 167 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTLlf-dBo1JajJ4a9:-NSqhB5S3Arouj0JaaWQ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L167)):*

I'm not seeing anywhere in the code above where you're re-saving the `Domain` with a modified autorenewBillingEvent pointing to the new one you're writing out?
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 168 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqToXp2skZ37zIQ-DZ:-NSqiZZr1JbzQ53LJNM5:bgaey6v) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L168)):*

You also need a `checkArgument` that the domain's deletionTime is `END_OF_TIME`.
</blockquote></details>

Done.

yeah we already checked for deleted-in-the-past domains, need to check for this 
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 169 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqUWbW-mKGEq2X6asF:-NSqijkZ7teqNObGSrxX:bppcaul) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L169)):*

How about a check on the domain's `autorenewEndTime` as well?  That might be relevant.  Or would that info also be captured in the recurrence?
</blockquote></details>

wouldn't that already be captured in 
- the recurrence 
-  the deletionTime on the domain 
- ?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 163 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqTwcX1FeqXcPEXmyb:-NSve-A17VIhus1lhI77:b-veduom) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L163)):*

yeah fair enough, and that plus a deletion time check would cover PENDING_TRANSFER and PENDING_DELETE statuses 
</blockquote></details>

Yeah. Just remember, checking the deletionTime to know that a domain isn't being deleted is a super common pattern and is the preferred/canonical way to do this.
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 169 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqUWbW-mKGEq2X6asF:-NSvele7AO8b4ZXj7c8G:b-r5codb) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L169)):*

wouldn't that already be captured in 
- the recurrence 
-  the deletionTime on the domain 
- ? 
</blockquote></details>

In checking the code, it doesn't seem that it is?  Double-check this but this might be a different independent situation.  Note that the way this field is used is that a nightly batch job checks it and then issues the delete, so until that happens, the recurrence will not be modified (and the deletion time not set).  But still, in this situation, the existing recurrence is essentially already destined to be stale, and changes to it won't actually take effect, as the domain will be deleted before then.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 169 at r2](https://reviewable.io/reviews/google/nomulus/1987#-NSqUWbW-mKGEq2X6asF:-NSvotHOBcJo1sC0K5lX:biuzcev) ([raw file](https://github.com/google/nomulus/blob/aaac56593b3e523271dde6b925cf0c6dee1a6da1/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L169)):*

In checking the code, it doesn't seem that it is?  Double-check this but this might be a different independent situation.  Note that the way this field is used is that a nightly batch job checks it and then issues the delete, so until that happens, the recurrence will not be modified (and the deletion time not set).  But still, in this situation, the existing recurrence is essentially already destined to be stale, and changes to it won't actually take effect, as the domain will be deleted before then.
</blockquote></details>

yeah this seems like an odd edge case, especially since Domain's autorenewEndTime is only ever modified by a superuser update (or cleared by a restore request) 

but sure we can check it, doesn't hurt anyone to do so

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 75 at r4](https://reviewable.io/reviews/google/nomulus/1987#-NSwNQruCvCYHRZfN6dV:-NSwNQruCvCYHRZfN6dW:bn6k2tt) ([raw file](https://github.com/google/nomulus/blob/f82b9ba8c5a2c89342f1e52f47039378ad76b5c0/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L75)):*
> ```Java
>         tm().transact(this::loadDomainsAndRecurrings);
>     checkArgument(
>         renewalPriceBehavior != null || specifiedRenewalPrice != null,
> ```

These checks should be before the `domainsAndRecurrings` are loaded, as they don't require hitting the DB and can fail instantly (whereas opening a DB connection and getting info from it might take a non-trivial amount of time).
___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 161 at r4](https://reviewable.io/reviews/google/nomulus/1987#-NSwODrvAhw3e5G2gO2E:-NSwODrvAhw3e5G2gO2F:b-pe76v5) ([raw file](https://github.com/google/nomulus/blob/f82b9ba8c5a2c89342f1e52f47039378ad76b5c0/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L161)):*
> ```Java
>     for (String domainName : mainParameters) {
>       Domain domain =
>           EppResourceUtils.loadByForeignKey(Domain.class, domainName, now)
> ```

Just note that doing these individually is not as performant as possible, but I don't think it matters that much for a manual command that will typically be called with a single domain.

CydeWeys (approved):
Rename the Java object.  That was always the plan, it just hasn't been gotten around to yet.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987#-:-NT-FjsDBdDRek2DlWtF:b-63po0n)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 161 at r4](https://reviewable.io/reviews/google/nomulus/1987#-NSwODrvAhw3e5G2gO2E:-NT-ovtv7xU0yzVk6OdO:bk1ss0h) ([raw file](https://github.com/google/nomulus/blob/f82b9ba8c5a2c89342f1e52f47039378ad76b5c0/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L161)):*

Just note that doing these individually is not as performant as possible, but I don't think it matters that much for a manual command that will typically be called with a single domain.
</blockquote></details>

yep, this was explicitly written to be simpler code-wise with the idea that we don't really care about bulk performance in such a one-off manual command like this.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1987)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java` line 75 at r4](https://reviewable.io/reviews/google/nomulus/1987#-NSwNQruCvCYHRZfN6dV:-NT-q4zk5UttT35K7onf:b-7j86re) ([raw file](https://github.com/google/nomulus/blob/f82b9ba8c5a2c89342f1e52f47039378ad76b5c0/core/src/main/java/google/registry/tools/UpdateRecurrenceCommand.java#L75)):*

These checks should be before the `domainsAndRecurrings` are loaded, as they don't require hitting the DB and can fail instantly (whereas opening a DB connection and getting info from it might take a non-trivial amount of time).
</blockquote></details>

sure

PR #1999, opened 2023-04-19
Add cloud scheduler and tasks deployer

ptkach:
This updates what used to be a cloudSchedulerDeployer script with changes necessary to support both Cloud Scheduler Tasks and Cloud Tasks Queues. The entire script has been refactored to provide more structure and better support for different types of deployments. 
I've considered different scripts for different deployments, however both deployments share a lot of similarities, so it's reusing. 

Verified locally by running the script and checking the outcome for both Cloud Scheduler tasks (for regression) and Cloud Tasks Queues (new functionality). 

Update 04.25: Verified cloud build deployment to alpha, confirmed no regression to cloud scheduler and successfully deployed queues 

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/1999)


ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1999)* status: 2 of 15 files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`build.gradle` line 562 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo0Tc_DWTI9nXzksBE:-NTszgaH5EU6HunmUako:ba3o3s1) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/build.gradle#L562)):*

TBT I feel this name is a bit misleading, especially when in the .xml file you have "Cloud Scheduler tasks begin" and "Cloud Scheduler tasks end". It's not clear what the "task" here is for (whether Cloud Scheduler tasks, or Cloud Tasks).

I think a better name is `deployCloudSchedulerAndCloudTasksQueue`, or Just `deployCloudSchedulerAndQueue`.

Same goes to the golang file name.
</blockquote></details>

Renamed
___
*[`docs/configuration.md` line 42 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo8xk9917p1s_gWFFh:-NTt6Zn43zspQd7WyZkg:b-cy8lsw) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/docs/configuration.md#L42)):*

This again might be misleading. "Cloud Tasks" is a proper noun and should not be broken up like that. It should be more like "Cloud Scheduler and Cloud Tasks queues". But as with previous comment, I think we should just have a separate file for queue configurations for all environments.
</blockquote></details>

Renamed, replied to separate files above
___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 153 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo0MVP4MUyuT5WUx3V:-NTt1eiH-dzwvM-pT509:b-cbnpbt) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L153)):*

Maybe add a link to the reference page so people know what these parameters are for.

Also, as I mentioned in the other file, this file should be called something like "`cloud-scheduler-and-queue.xml`. Or we can have two separate XMLs (which I think is preferable). There is no reason to lump them together, even if we use the same script to handle them.
</blockquote></details>

Added the link and renamed. 
Re separate the files, I've considered that, but there're practically no downsides and multiple upsides in keeping them in a single file, given that we always want to deploy both.
To name a few benefits - it's more resourceful as script only reads and parses single xml tree; we reuse existing rails to deploy and parse the file; we get to keep the deployer script simpler, by reducing the number of arguments; 
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml` line 2 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo8Y0DB4BgXJmW2xxK:-NTt4Ihr21WzJtzfCptO:b-js7qy0) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/core/src/main/java/google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml#L2)):*

Come to think of it, we used to have just one `queue.xml` file for all environments. Why duplicate the same content in each environment-specific file now? For cron tasks I can see that there are tasks that we only want to run in production, like billing generation, etc, but for queues they really out to be the same for all environments.
</blockquote></details>

This is again something I've been contemplating quite a bit before making the change. The main reason for it being per env is actually the 7 days no re-create policy for deleted queues. This way queue updates (delete and recreate) for one env won't affect the other ones, like when I've deleted sheet queue in alpha 3 days ago and no required to deploy updated config. 
Separately, as I've also decided to reuse already existing cloud-scheduler xml config, we're not gaining any more files, we're actually going to loose one eventually.

jianglai (changes requested):
> Reviewed 3 of 13 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1999)* status: 5 of 15 files reviewed, 4 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 153 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo0MVP4MUyuT5WUx3V:-NU25nLM1a3q2Pg9ABXq:bz6m8b6) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L153)):*

Added the link and renamed. 
Re separate the files, I've considered that, but there're practically no downsides and multiple upsides in keeping them in a single file, given that we always want to deploy both.
To name a few benefits - it's more resourceful as script only reads and parses single xml tree; we reuse existing rails to deploy and parse the file; we get to keep the deployer script simpler, by reducing the number of arguments; 
</blockquote></details>

Since this is something that happens rather infrequently (once a week during deployment), performance is not really a concern here. Also the two XML elements are completely separate, so I doubt saving them in one file vs. two files would make much of a difference when it comes to how the tool script parses it. That said, if we are only talking about splitting this per-env file into two, for each environment, I don't feel particularly strongly one way or another.

What I do think we should consider more carefully is the whether to share the queue config across environments. See comments below.
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml` line 2 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo8Y0DB4BgXJmW2xxK:-NU25pIYDXA8jj6xNQbj:bgiwpui) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/core/src/main/java/google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml#L2)):*

This is again something I've been contemplating quite a bit before making the change. The main reason for it being per env is actually the 7 days no re-create policy for deleted queues. This way queue updates (delete and recreate) for one env won't affect the other ones, like when I've deleted sheet queue in alpha 3 days ago and no required to deploy updated config. 
Separately, as I've also decided to reuse already existing cloud-scheduler xml config, we're not gaining any more files, we're actually going to loose one eventually. 
</blockquote></details>

Yes, the issue with the "7 days no re-receate" policy is a point against sharing the queue config file. However, I think arguments can be made for sharing the same file too. Namely I consider queues to be part of the Nomulus infrastructure that is similar to how paths are mapped to difference actions. There really should not be any difference between each environment when it comes to which queues they have and what are the parameters they are set to. This is in contrast to cron jobs where you would want to turn on/off different jobs based on the environment you are in. i. e. cron jobs define how you use a particular environment, whereas queues are part of the setup of each environment, which should be identical.

So keeping them as separate files is just asking for trouble when you need to update a queue. You have to remember to do the same for all environments or lest we'll end up with diverging environments down the line.

I've been thinking about how to overcome the issue with the somewhat annoying 7-days rule. We don't have it now when using `gcloud app deploy queue.yaml`, right? I think with App Engine queues, the deployment never actually delete a queue, even if you remove it from the config file. It will make sure whatever is described in the config file is reflected in the system after deployment, but not necessarily that the system doesn't have more queues that are not "managed" by the config file. It will just leave them there. This is probably the best way to handle queues with our new script too. Deleting queues out right makes things irreversible (within 7 days at least) and I can see it being rather undesirable. For example what if you need to do an emergency rollback after a release that deleted a queue? You are then in a pickle because you can't add that queue back. Maybe we can ask the script to pause a queue if it is not included in the config file, to keep things tidier. A queue itself doesn't cost anything anyway (it's the operations in the queue that are [billed](https://cloud.google.com/tasks/pricing)).

Sorry for the back and forth, but the more I think of this, the more I believe deleting queues is a bad idea. If we don't have to worry about that, then sharing the same config for all environments is also neater and fits better with the overall architecture with the system.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1999)* status: 3 of 16 files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml` line 153 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo0MVP4MUyuT5WUx3V:-NU3Pisc5SFFDREIlT8k:b2nhb2c) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/core/src/main/java/google/registry/env/alpha/default/WEB-INF/cloud-scheduler-tasks.xml#L153)):*

Since this is something that happens rather infrequently (once a week during deployment), performance is not really a concern here. Also the two XML elements are completely separate, so I doubt saving them in one file vs. two files would make much of a difference when it comes to how the tool script parses it. That said, if we are only talking about splitting this per-env file into two, for each environment, I don't feel particularly strongly one way or another.

What I do think we should consider more carefully is the whether to share the queue config across environments. See comments below.
</blockquote></details>

Ack
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml` line 2 at r1](https://reviewable.io/reviews/google/nomulus/1999#-NTo8Y0DB4BgXJmW2xxK:-NU3PlqT2j6B7pzXR0Ae:bmj4aoq) ([raw file](https://github.com/google/nomulus/blob/e26b8ffab011645cf7e121176271e89940878dcb/core/src/main/java/google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml#L2)):*

Yes, the issue with the "7 days no re-receate" policy is a point against sharing the queue config file. However, I think arguments can be made for sharing the same file too. Namely I consider queues to be part of the Nomulus infrastructure that is similar to how paths are mapped to difference actions. There really should not be any difference between each environment when it comes to which queues they have and what are the parameters they are set to. This is in contrast to cron jobs where you would want to turn on/off different jobs based on the environment you are in. i. e. cron jobs define how you use a particular environment, whereas queues are part of the setup of each environment, which should be identical.

So keeping them as separate files is just asking for trouble when you need to update a queue. You have to remember to do the same for all environments or lest we'll end up with diverging environments down the line.

I've been thinking about how to overcome the issue with the somewhat annoying 7-days rule. We don't have it now when using `gcloud app deploy queue.yaml`, right? I think with App Engine queues, the deployment never actually delete a queue, even if you remove it from the config file. It will make sure whatever is described in the config file is reflected in the system after deployment, but not necessarily that the system doesn't have more queues that are not "managed" by the config file. It will just leave them there. This is probably the best way to handle queues with our new script too. Deleting queues out right makes things irreversible (within 7 days at least) and I can see it being rather undesirable. For example what if you need to do an emergency rollback after a release that deleted a queue? You are then in a pickle because you can't add that queue back. Maybe we can ask the script to pause a queue if it is not included in the config file, to keep things tidier. A queue itself doesn't cost anything anyway (it's the operations in the queue that are [billed](https://cloud.google.com/tasks/pricing)).

Sorry for the back and forth, but the more I think of this, the more I believe deleting queues is a bad idea. If we don't have to worry about that, then sharing the same config for all environments is also neater and fits better with the overall architecture with the system.
</blockquote></details>

I agree, delete accompanied with the 7 days no-receate policy was bugging me as well, but I've not thought about a "pause" alternative, which I think is a good idea. Updated

jianglai (changes requested):
> Reviewed 2 of 15 files at r2, 12 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1999)* status: all files reviewed, 7 unresolved discussions (waiting on @ptkach)

___
*[`core/src/test/java/google/registry/testing/TaskQueueExtension.java` line 40 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURfYE0ARcPNdhF2us7:-NURfYE1DQz6GCL-WR0C:bpoqrxx) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/core/src/test/java/google/registry/testing/TaskQueueExtension.java#L40)):*
> ```Java
>    */
>   private static final String QUEUE_XML =
>       readResourceUtf8("google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml");
> ```

Actually this file will go away after #2000 is merged.



_Code quote:_
```Java
google/registry/env/common/default/WEB-INF/to_be_deleted_queue.xml"
```
___
*[`docs/architecture.md` line 114 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURfuFl6rpkmB9m20Lf:-NURfuFm7zqviOUo1Jwx:brsgm4u) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/docs/architecture.md#L114)):*
> ```Markdown
> 
> The task queues used by Nomulus are configured in the 
> `cloud-tasks-queue.xml` file. Note
> ```

You can probably re-flow the paragraph.



_Code quote:_
```Markdown
`cloud-tasks-queue.xml` file. Note
```
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml` line 15 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NUReWKd-uO96vG0eMm-:-NUReWKe247dT_0AJlyK:ba5bo3z) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml#L15)):*
> ```XML
>       <max-retry-duration></max-retry-duration>
>       <min-backoff></min-backoff>
>       <routing-override></routing-override>
> ```

We probably don't need this. I don't believe we have used this, and it is specific to App Engine anyway.
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml` line 96 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURfHnD1Ajr4ll3YXeP:-NURfHnD1Ajr4ll3YXeQ:b40chgw) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml#L96)):*
> ```XML
>     <name>sheet</name>
>     <max-dispatches-per-second>1</max-dispatches-per-second>
>     
> ```

max-concurrent-dispatches



_Code quote:_
```XML
max-concurrent-requests
```
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 162 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NUSYp3wEkLlETEXSkbg:-NUSYp3wEkLlETEXSkbh:b47jq3e) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L162)):*
> ```Go
> 	// Cloud Schedule doesn't allow description of more than 499 chars and \n
> 	var description string
> 	if len(task.Description) > 499 {
> ```

Maybe give a warning here?



_Code quote:_
```Go
	if len(task.Description) > 499 {
		description = task.Description[:499]
```
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 173 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NUSZGSoEHlCU3QDVIou:-NUSZGSoEHlCU3QDVIov:brfol3v) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L173)):*
> ```Go
> 		"scheduler", "jobs", operationType,
> 		"http", task.Name,
> 		"--location", "us-central1",
> ```

Since you have `us-central1` in several places, you can just make it a constant so if we ever need to change it we don't have to change it everywhere. Ideally it would be a command line parameter (with a default value as `us-central1`).



_Code quote:_
```Go
us-central1
```
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURtPIoE_TUyCnt_Xfo:-NURtPIoE_TUyCnt_Xfp:b-6i4fg5) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L320)):*
> ```Go
> 	var xmlEntries XmlEntries
> 	if err := xml.Unmarshal(byteValue, &xmlEntries); err != nil {
> 		panic("Failed to parse xml file entries: " + err.Error())
> ```

I have seen in Cloud Build that when an error occurred, no error messages are actually output to stderr (at least not captured by Cloud Build), for example here:

https://pantheon.corp.google.com/cloud-build/builds/3aa5069e-a7bc-4c29-a386-1fee21dffe7b;step=1?project=domain-registry-alpha

Have you tried if the error message is actually printed out in in the shell? I wonder if it's Cloud Build that somehow missed it, or if it's not being outputed at all. 



_Code quote:_
```Go
panic("Failed to parse xml file entries: " + err.Error())
```
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 325 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURyjW0AxwbnP-mOSCn:-NURyjW0AxwbnP-mOSCo:b-tvnw0n) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L325)):*
> ```Go
> 	if len(xmlEntries.Tasks) > 0 {
> 		log.Default().Println("Syncing cloud scheduler tasks from xml...")
> 		taskSyncManager := TasksSyncManager{
> ```

tasksSyncManager



_Code quote:_
```Go
taskSyncManager
```

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/1999)* status: 13 of 17 files reviewed, 7 unresolved discussions (waiting on @jianglai)

___
*[`docs/architecture.md` line 114 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURfuFl6rpkmB9m20Lf:-NUSy6Q43je3hGTTwUNd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/docs/architecture.md#L114)):*

You can probably re-flow the paragraph.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml` line 15 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NUReWKd-uO96vG0eMm-:-NUSxK6c2gy7Vs8EDs8u:ba3dfob) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml#L15)):*

We probably don't need this. I don't believe we have used this, and it is specific to App Engine anyway.
</blockquote></details>

Removed
___
*[`core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml` line 96 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURfHnD1Ajr4ll3YXeP:-NUSxLrkDHrGsn2-0eYv:bh7gqfq) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml#L96)):*

max-concurrent-dispatches
</blockquote></details>

Right, removed as it was a left over from find+replace on old field with new fields
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 162 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NUSYp3wEkLlETEXSkbg:-NUSz_Ne7GfgM1HBI3oN:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L162)):*

Maybe give a warning here?
</blockquote></details>

Done.
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 173 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NUSZGSoEHlCU3QDVIou:-NUT-G1_CiVZKLYkt6ze:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L173)):*

Since you have `us-central1` in several places, you can just make it a constant so if we ever need to change it we don't have to change it everywhere. Ideally it would be a command line parameter (with a default value as `us-central1`).
</blockquote></details>

Done.
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 320 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURtPIoE_TUyCnt_Xfo:-NUT2KTg8GmexHaLlpAl:bjipdsk) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L320)):*

I have seen in Cloud Build that when an error occurred, no error messages are actually output to stderr (at least not captured by Cloud Build), for example here:

https://pantheon.corp.google.com/cloud-build/builds/3aa5069e-a7bc-4c29-a386-1fee21dffe7b;step=1?project=domain-registry-alpha

Have you tried if the error message is actually printed out in in the shell? I wonder if it's Cloud Build that somehow missed it, or if it's not being outputed at all. 
</blockquote></details>

This place prints out fine, the issue you pointed is actually in a different place, which I've also addressed in this PR



_Code quote:_
```Go
xml.Unmarshal(
```
___
*[`release/builder/deployCloudSchedulerAndQueue.go` line 325 at r3](https://reviewable.io/reviews/google/nomulus/1999#-NURyjW0AxwbnP-mOSCn:-NUT03EvDHfRxAYPQ4ys:bhn2a73) ([raw file](https://github.com/google/nomulus/blob/4f4971e00cc9fcc970ee34dfe904131c7cb0e762/release/builder/deployCloudSchedulerAndQueue.go#L325)):*

tasksSyncManager
</blockquote></details>

Not sure why not plural? We're syncing all tasks, not one of them, so plural appears to be more suitable

PR #2007, opened 2023-04-24
Don't apply non-premium default tokens to premium names

sarahcaseybot:
This is a fix for b/279438732

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2007)

### DIFF
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -81,6 +81,7 @@ import google.registry.model.tld.Tld;
 import google.registry.model.tld.Tld.TldState;
 import google.registry.model.tld.label.ReservationType;
 import google.registry.persistence.VKey;
+import google.registry.pricing.PricingEngineProxy;
 import google.registry.util.Clock;
 import java.util.HashSet;
 import java.util.Optional;
@@ -93,6 +94,8 @@ import org.joda.time.DateTime;
  *
  * <p>This flow also supports the EPP fee extension and can return pricing information.
  *
+ * @error {@link
+ *     google.registry.flows.domain.DomainPricingLogic.AllocationTokenInvalidForPremiumNameException}
  * @error {@link google.registry.flows.exceptions.TooManyResourceChecksException}
  * @error {@link google.registry.flows.FlowUtils.NotLoggedInException}
  * @error {@link google.registry.flows.FlowUtils.UnknownCurrencyEppException}
@@ -291,6 +294,7 @@ public final class DomainCheckFlow implements Flow {
                 allocationToken.get(),
                 feeCheckItem.getCommandName(),
                 registrarId,
+                PricingEngineProxy.isDomainPremium(domainName, now),
                 now);
           }
           handleFeeRequest(
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -54,6 +54,7 @@ import google.registry.flows.domain.DomainPricingLogic.AllocationTokenInvalidFor
 import google.registry.flows.domain.token.AllocationTokenDomainCheckResults;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotInPromotionException;
+import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForCommandException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForDomainException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForRegistrarException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForTldException;
@@ -310,6 +311,7 @@ public final class DomainCheckFlow implements Flow {
               recurrences.getOrDefault(domainName, null));
           responseItems.add(builder.setDomainNameIfSupported(domainName).build());
         } catch (AllocationTokenInvalidForPremiumNameException
+            | AllocationTokenNotValidForCommandException
             | AllocationTokenNotValidForDomainException
             | AllocationTokenNotValidForRegistrarException
             | AllocationTokenNotValidForTldException
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -50,10 +50,10 @@ import google.registry.flows.annotations.ReportingSpec;
 import google.registry.flows.custom.DomainCheckFlowCustomLogic;
 import google.registry.flows.custom.DomainCheckFlowCustomLogic.BeforeResponseParameters;
 import google.registry.flows.custom.DomainCheckFlowCustomLogic.BeforeResponseReturnData;
+import google.registry.flows.domain.DomainPricingLogic.AllocationTokenInvalidForPremiumNameException;
 import google.registry.flows.domain.token.AllocationTokenDomainCheckResults;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotInPromotionException;
-import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForCommandException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForDomainException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForRegistrarException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils.AllocationTokenNotValidForTldException;
@@ -309,7 +309,7 @@ public final class DomainCheckFlow implements Flow {
               availableDomains.contains(domainName),
               recurrences.getOrDefault(domainName, null));
           responseItems.add(builder.setDomainNameIfSupported(domainName).build());
-        } catch (AllocationTokenNotValidForCommandException
+        } catch (AllocationTokenInvalidForPremiumNameException
             | AllocationTokenNotValidForDomainException
             | AllocationTokenNotValidForRegistrarException
             | AllocationTokenNotValidForTldException
--- a/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainCheckFlow.java
@@ -95,8 +95,6 @@ import org.joda.time.DateTime;
  *
  * <p>This flow also supports the EPP fee extension and can return pricing information.
  *
- * @error {@link
- *     google.registry.flows.domain.DomainPricingLogic.AllocationTokenInvalidForPremiumNameException}
  * @error {@link google.registry.flows.exceptions.TooManyResourceChecksException}
  * @error {@link google.registry.flows.FlowUtils.NotLoggedInException}
  * @error {@link google.registry.flows.FlowUtils.UnknownCurrencyEppException}
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -1219,13 +1219,14 @@ public class DomainFlowUtils {
       try {
         AllocationTokenFlowUtils.validateToken(
             InternetDomainName.from(domainName), token.get(), commandName, registrarId, now);
-        if (PricingEngineProxy.isDomainPremium(domainName, now)
-            && !token.get().shouldDiscountPremiums()) {
-          continue;
-        }
       } catch (AssociationProhibitsOperationException | StatusProhibitsOperationException e) {
         // Allocation token was not valid for this registration, continue to check the next token in
         // the list
+        continue;
+      }
+      // Don't apply token to premium names if token not valid for premiums
+      if (PricingEngineProxy.isDomainPremium(domainName, now)
+          && !token.get().shouldDiscountPremiums()) {
         continue;
       }
       // Only use the first valid token in the list
--- a/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
+++ b/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java
@@ -74,6 +74,7 @@ import google.registry.flows.EppException.ParameterValueSyntaxErrorException;
 import google.registry.flows.EppException.RequiredParameterMissingException;
 import google.registry.flows.EppException.StatusProhibitsOperationException;
 import google.registry.flows.EppException.UnimplementedOptionException;
+import google.registry.flows.domain.DomainPricingLogic.AllocationTokenInvalidForPremiumNameException;
 import google.registry.flows.domain.token.AllocationTokenFlowUtils;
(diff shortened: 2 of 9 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2043 at r1](https://reviewable.io/reviews/google/nomulus/2007#-NTo2laP01a1ew43EgMZ:-NTo2laQ0xQagOVwZc1d:b2iaw7s) ([raw file](https://github.com/google/nomulus/blob/ee4fb83ded01172187979be2833b49444b890727/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2043)):*
> ```Java
>     doSuccessfulTest();
>   }
> ```

should we check renewals as well? Presumably that could possibly run into issues as well (without this fix)

gbrodman (approved):
> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1224 at r2](https://reviewable.io/reviews/google/nomulus/2007#-NTo9mL075zBDeHk7Gta:-NTo9mL075zBDeHk7Gtb:bqmm7ee) ([raw file](https://github.com/google/nomulus/blob/87b715fcd5c56a31eec0266175acb34c6dcae075/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1224)):*
> ```Java
>         AllocationTokenFlowUtils.validateToken(
>             InternetDomainName.from(domainName), token.get(), commandName, registrarId, now);
>         if (PricingEngineProxy.isDomainPremium(domainName, now)
> ```

nit, but let's put this outside of the try/catch block with a little note on why we're performing the check maybe

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1222 at r3](https://reviewable.io/reviews/google/nomulus/2007#-NToUz5OC_h_5kIfh6Ns:-NToUz5OC_h_5kIfh6Nt:b-dc9faj) ([raw file](https://github.com/google/nomulus/blob/8ddec5bff2642ed143226c04d1e39fd29c8a96b2/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1222)):*
> ```Java
>     for (Optional<AllocationToken> token : tokenList) {
>       try {
>         AllocationTokenFlowUtils.validateToken(
> ```

How come this doesn't do premium handling?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1222 at r3](https://reviewable.io/reviews/google/nomulus/2007#-NToUz5OC_h_5kIfh6Ns:-NToZoYL4vHE137zntkU:b-7z0gzn) ([raw file](https://github.com/google/nomulus/blob/8ddec5bff2642ed143226c04d1e39fd29c8a96b2/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1222)):*

There are other places this method is called (such as https://cs.opensource.google/nomulus/nomulus/+/refs/heads/master:core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java;l=82;drc=e1afe007585aac2b88afb0ba431c454652df2a63;bpv=1;bpt=1) where we would want to allow the token for a premium name. We could add a check if the token type is DEFAULT_PROMO and the token does not discount premiums, and the domain is premium, then the token is not valid. But it seemed simpler to do it here since validateToken returns an EPPException and the AllocationTokenInvalidForPremiumNameException is a CommandUseErrorException which would require more refactoring.
</blockquote></details>

basically, the distinction has been between "is this token valid now in general for this tld / registrar / time" and "is this token valid for this particular name?" 

I don't think there are any situations where that distinction is strictly necessary so it probably makes sense to add the domain name to the validation (we'll just need to add in `isPremium` from wherever) but that's why it exists

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1222 at r3](https://reviewable.io/reviews/google/nomulus/2007#-NToUz5OC_h_5kIfh6Ns:-NTtQoSF0MkjomSObFTW:b-s5jh6w) ([raw file](https://github.com/google/nomulus/blob/8ddec5bff2642ed143226c04d1e39fd29c8a96b2/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1222)):*

@sarahcaseybot what's the exact business logic there in the check flow requiring a non-premium token to work on a premium domain name?  If a token doesn't support premiums, and a `domain:check` is run on a premium name, the expected result is an error message, right?  Surely we have a test case covering this.

The larger issue is that we have a validate method that takes a domain name and a token, but that is returning true even if that token is not valid for that domain name.  This seems like a flat out bug (and indeed has caused the bug we are seeing now).  I would rather correct that bug so it doesn't bite us in the future, rather than fix it just at one call-site.
</blockquote></details>

We definitely have test cases for creations etc using non-premium tokens on premium names, but we might not have one for checks I suppose (the ability to use a token on a check was added fairly recently in the grand scheme of things I believe). 

But yeah I think it would make sense to do the premium check in the AllocationTokenFlowUtils method

gbrodman (changes requested):
> Reviewed 5 of 5 files at r4, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: 7 of 8 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r4](https://reviewable.io/reviews/google/nomulus/2007#-NTuDj1y1fWusWAj9jE2:-NTuDj1y1fWusWAj9jE3:b-ej8hx1) ([raw file](https://github.com/google/nomulus/blob/914c3456169592c4ee7dc404ac1a0a31f9368555/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*
> ```Java
>   }
> 
>   /**
> ```

we should update the comment here too
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 134 at r4](https://reviewable.io/reviews/google/nomulus/2007#-NTuDuvk4YeU_Zq3x8n-:-NTuDuvk4YeU_Zq3x8n0:b-cgm0oz) ([raw file](https://github.com/google/nomulus/blob/914c3456169592c4ee7dc404ac1a0a31f9368555/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L134)):*
> ```Java
>       return;
>     }
>     if (!token.shouldDiscountPremiums()
> ```

anchor tenant tokens don't have a discount fraction set, so we can use the same check here that we do in DomainPricingLogic. Better yet, maybe make that logic a little method in AllocationTokenFlowUtils or somewhere and call it here and in DPL. 
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 410 at r4](https://reviewable.io/reviews/google/nomulus/2007#-NTuEa9Y-pnkP6inUEk3:-NTuEa9Y-pnkP6inUEk4:b-n69ic6) ([raw file](https://github.com/google/nomulus/blob/914c3456169592c4ee7dc404ac1a0a31f9368555/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L410)):*
> ```Java
>                 .build()));
>   }
> ```

we probably want a test for an anchor tenant token on a premium domain (unless we have one already?)

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 114 at r4](https://reviewable.io/reviews/google/nomulus/2007#-NTuDj1y1fWusWAj9jE2:-NTywK6Y00Y99N4zQtof:b-896fix) ([raw file](https://github.com/google/nomulus/blob/914c3456169592c4ee7dc404ac1a0a31f9368555/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L114)):*

we should update the comment here too
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 134 at r4](https://reviewable.io/reviews/google/nomulus/2007#-NTuDuvk4YeU_Zq3x8n-:-NTyw3fS4FGfAFu0iprT:b8z584h) ([raw file](https://github.com/google/nomulus/blob/914c3456169592c4ee7dc404ac1a0a31f9368555/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L134)):*

anchor tenant tokens don't have a discount fraction set, so we can use the same check here that we do in DomainPricingLogic. Better yet, maybe make that logic a little method in AllocationTokenFlowUtils or somewhere and call it here and in DPL. 
</blockquote></details>

Done. And this changes the logic to also not require that the registration behavior specifically be set to Anchor Tenant, but to just require that there be no discount
___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 410 at r4](https://reviewable.io/reviews/google/nomulus/2007#-NTuEa9Y-pnkP6inUEk3:-NTyyZVe9fAtb6A2TMz0:b-896fix) ([raw file](https://github.com/google/nomulus/blob/914c3456169592c4ee7dc404ac1a0a31f9368555/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L410)):*

we probably want a test for an anchor tenant token on a premium domain (unless we have one already?)
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed (commit messages unreviewed), 3 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 159 at r6](https://reviewable.io/reviews/google/nomulus/2007#-NU3LSRe9-ntWA3IzIRU:-NU3LSRe9-ntWA3IzIRV:blasfgu) ([raw file](https://github.com/google/nomulus/blob/d62a14888462279db02fd24ca8706bb6bc6ebbab/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L159)):*
> ```Java
> 
>   /** Validates that the given token is valid for a premium name. */
>   public static void validateForPremiums(Optional<AllocationToken> token, boolean isPremium)
> ```

given that this is now its own method, can we be a little more detailed with the naming? e.g. `domainIsPremium` for the param, `validateTokenOnPossiblePremiumName` for the method name or something like that? Just so that it's clear what we're doing when we see it called from another class

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java` line 159 at r6](https://reviewable.io/reviews/google/nomulus/2007#-NU3LSRe9-ntWA3IzIRU:-NU3XiYN3Mtieoa4nHDa:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d62a14888462279db02fd24ca8706bb6bc6ebbab/core/src/main/java/google/registry/flows/domain/token/AllocationTokenFlowUtils.java#L159)):*

given that this is now its own method, can we be a little more detailed with the naming? e.g. `domainIsPremium` for the param, `validateTokenOnPossiblePremiumName` for the method name or something like that? Just so that it's clear what we're doing when we see it called from another class
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @github-code-scanning[bot])

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @github-code-scanning[bot])

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: all files reviewed, 2 unresolved discussions (waiting on @github-code-scanning[bot] and @sarahcaseybot)

___
*[`core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java` line 425 at r9](https://reviewable.io/reviews/google/nomulus/2007#-NU7zBjL5ZEf5OTbrle2:-NU7zBjL5ZEf5OTbrle3:b-vc6929) ([raw file](https://github.com/google/nomulus/blob/ee52c043ba356fd6634f48abff5bf8473686bdd0/core/src/test/java/google/registry/flows/domain/DomainCheckFlowTest.java#L425)):*
> ```Java
>     setEppInput(
>         "domain_check_allocationtoken_promotion.xml", ImmutableMap.of("DOMAIN", "rich.example"));
>     assertThrows(AllocationTokenInvalidForPremiumNameException.class, this::runFlow);
> ```

Is this the correct behavior?  What about a `domain:check` on multiple domain names where only some are premium but the rest are valid for the specified token; should that domain check succeed for the valid domain names, and then just return a status message for the invalid ones?  How is this handled for other cases of multiple-domain checks where the passed token isn't applicable to all of them?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-code-scanning[bot], and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java` line 281 at r10](https://reviewable.io/reviews/google/nomulus/2007#-NU8UTv_0nkOfizD2z6T:-NU8UTv_0nkOfizD2z6U:b-xl294l) ([raw file](https://github.com/google/nomulus/blob/7c286d97d0c4952a4f66262a075c583dbb4c7484/core/src/main/java/google/registry/flows/domain/DomainPricingLogic.java#L281)):*
> ```Java
>       extends CommandUseErrorException {
>     public AllocationTokenInvalidForPremiumNameException() {
>       super("Token invalid for premium name");
> ```

I think a slightly better message would be "Token not valid for premium names" or similar (if it fits).

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2007#-:-NU8Ub7O5WhYfXo3B-Qx:bnfp4nl)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-code-scanning[bot], and @sarahcaseybot)

PR #2011, opened 2023-04-26
Don't allow a list of the empty string in List<String> fields

gbrodman:
If the user does, e.g. `--allowed_nameservers=` (or contact ids) that shouldn't mean a list consisting solely of the empty string.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2011)

### DIFF
--- a/core/src/main/java/google/registry/tools/CreateOrUpdateDomainCommand.java
+++ b/core/src/main/java/google/registry/tools/CreateOrUpdateDomainCommand.java
@@ -21,6 +21,7 @@ import com.beust.jcommander.Parameter;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
 import google.registry.tools.params.NameserversParameter;
+import google.registry.tools.params.StringListParameter;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
@@ -53,15 +54,15 @@ abstract class CreateOrUpdateDomainCommand extends MutatingEppToolCommand {
   String registrant;
 
   @Parameter(
-    names = {"-a", "--admins"},
-    description = "Comma-separated list of admin contacts."
-  )
+      names = {"-a", "--admins"},
+      description = "Comma-separated list of admin contacts.",
+      listConverter = StringListParameter.class)
   List<String> admins = new ArrayList<>();
 
   @Parameter(
-    names = {"-t", "--techs"},
-    description = "Comma-separated list of technical contacts."
-  )
+      names = {"-t", "--techs"},
+      description = "Comma-separated list of technical contacts.",
+      listConverter = StringListParameter.class)
   List<String> techs = new ArrayList<>();
 
   @Parameter(
--- a/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java
+++ b/core/src/main/java/google/registry/tools/CreateOrUpdateRegistrarCommand.java
@@ -18,6 +18,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Predicates.isNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.util.DomainNameUtils.canonicalizeHostname;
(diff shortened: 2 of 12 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2011)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java` line 210 at r1](https://reviewable.io/reviews/google/nomulus/2011#-NTyDp8lBZ6r-5ku3TV2:-NTyDp8lBZ6r-5ku3TV3:bbdnzev) ([raw file](https://github.com/google/nomulus/blob/837ffa3cc973a00ef6163058abef99c98efa8ea5/core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java#L210)):*
> ```Java
>   @Parameter(
>       names = "--allowed_nameservers",
>       description = "A comma-separated list of allowed nameservers for the TLD")
> ```

I think a better way to fix this would be to add a `validateWith` directive here, and create a new validator for a list of strings that disallows empty strings being in the list (is that ever a valid use case? it seems like it's always a bug, analogous to the default Java List implementation allowing null as a value).

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2011)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java` line 210 at r1](https://reviewable.io/reviews/google/nomulus/2011#-NTyDp8lBZ6r-5ku3TV2:-NTybAqLCsp9PkVlpJd_:b-65tzqi) ([raw file](https://github.com/google/nomulus/blob/837ffa3cc973a00ef6163058abef99c98efa8ea5/core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java#L210)):*

I think a better way to fix this would be to add a `validateWith` directive here, and create a new validator for a list of strings that disallows empty strings being in the list (is that ever a valid use case? it seems like it's always a bug, analogous to the default Java List implementation allowing null as a value).
</blockquote></details>

the proper use is the listConverter (not just "converter"), not validateWith (since we're converting, not validating). 

this ends up being a significantly larger PR but it's probably for the best

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2011)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/params/StringListParameter.java` line 23 at r3](https://reviewable.io/reviews/google/nomulus/2011#-NU7xXNX1tAbuX3cN8Je:-NU7xXNYD6JEoN0lv-1u:b-w7djpz) ([raw file](https://github.com/google/nomulus/blob/d4aee0a7b92adce855489a501057d0b2d37e0b24/core/src/main/java/google/registry/tools/params/StringListParameter.java#L23)):*
> ```Java
> 
> /**
>  * Converter for lists of String params that doesn't allow a list of the empty string.
> ```

This description isn't exactly correct.  It's not that it isn't allowed (which implies an error would be thrown), it's that it converts it.  A better description would be something about how all empty string elements are removed.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2011#-:-NU7xpAm6I4VY3XV2W36:bnfp4nl)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2011)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/tools/params/StringListParameter.java` line 23 at r3](https://reviewable.io/reviews/google/nomulus/2011#-NU7xXNX1tAbuX3cN8Je:-NU85gLz0FqyhxG4BHav:b-896fix) ([raw file](https://github.com/google/nomulus/blob/d4aee0a7b92adce855489a501057d0b2d37e0b24/core/src/main/java/google/registry/tools/params/StringListParameter.java#L23)):*

This description isn't exactly correct.  It's not that it isn't allowed (which implies an error would be thrown), it's that it converts it.  A better description would be something about how all empty string elements are removed.
</blockquote></details>

Done.

PR #2013, opened 2023-04-27
Allow usage of standard HTTP requests in CloudTasksUtils

gbrodman:
This adds a possible configuration point "defaultServiceAccount" (which in GAE will be the standard GAE service account). If this is configured, CloudTasksUtils can create tasks with standard HTTP requests with an OIDC token corresponding to that service account, as opposed to using the AppEngine-specific request methods.

This also works with IAP, in that if IAP is on and we specify the IAP client ID in the config, CloudTasksUtils will use the IAP client ID as the token audience and the request will successfully be passed through the IAP layer. Unfortunately, unlike other areas (e.g. RequestFactoryModule), we cannot use both IAP and standard authentication; we must choose one. This means that we will need to turn on IAP at roughly the same time that we deploy with the internal configuration change that adds the IAP client ID.

Tetsted in QA.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2013)

### DIFF
--- a/core/src/main/java/google/registry/batch/CloudTasksUtils.java
+++ b/core/src/main/java/google/registry/batch/CloudTasksUtils.java
@@ -16,6 +16,7 @@ package google.registry.batch;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static google.registry.tools.ServiceConnection.getServer;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 import com.google.api.gax.rpc.ApiException;
@@ -23,6 +24,8 @@ import com.google.cloud.tasks.v2.AppEngineHttpRequest;
 import com.google.cloud.tasks.v2.AppEngineRouting;
 import com.google.cloud.tasks.v2.CloudTasksClient;
 import com.google.cloud.tasks.v2.HttpMethod;
+import com.google.cloud.tasks.v2.HttpRequest;
+import com.google.cloud.tasks.v2.OidcToken;
 import com.google.cloud.tasks.v2.QueueName;
 import com.google.cloud.tasks.v2.Task;
 import com.google.common.base.Joiner;
@@ -46,7 +49,10 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.Random;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import org.joda.time.Duration;
 
@@ -61,6 +67,9 @@ public class CloudTasksUtils implements Serializable {
   private final Clock clock;
   private final String projectId;
   private final String locationId;
+  // defaultServiceAccount and iapClientId are nullable because Optional isn't serializable
+  @Nullable private final String defaultServiceAccount;
+  @Nullable private final String iapClientId;
   private final SerializableCloudTasksClient client;
 
   @Inject
@@ -69,11 +78,15 @@ public class CloudTasksUtils implements Serializable {
       Clock clock,
       @Config("projectId") String projectId,
       @Config("locationId") String locationId,
+      @Config("defaultServiceAccount") Optional<String> defaultServiceAccount,
+      @Config("iapClientId") Optional<String> iapClientId,
       SerializableCloudTasksClient client) {
     this.retrier = retrier;
     this.clock = clock;
     this.projectId = projectId;
     this.locationId = locationId;
+    this.defaultServiceAccount = defaultServiceAccount.orElse(null);
+    this.iapClientId = iapClientId.orElse(null);
     this.client = client;
   }
 
@@ -96,6 +109,74 @@ public class CloudTasksUtils implements Serializable {
 
   public ImmutableList<Task> enqueue(String queue, Task... tasks) {
     return enqueue(queue, Arrays.asList(tasks));
+  }
+
+  /**
+   * Converts a (possible) set of params into an HTTP request via the appropriate method.
+   *
+   * <p>For GET requests we add them on to the URL, and for POST requests we add them in the body of
+   * the request.
+   *
+   * <p>The parameters {@code putHeadersFunction} and {@code setBodyFunction} are used so that this
+   * method can be called with either an AppEngine HTTP request or a standard non-AppEngine HTTP
+   * request. The two objects do not have the same methods, but both have ways of setting headers /
+   * body.
+   *
+   * @return the resulting path (unchanged for POST requests, with params added for GET requests)
+   */
+  private String processRequestParameters(
+      String path,
+      HttpMethod method,
+      Multimap<String, String> params,
+      BiConsumer<String, String> putHeadersFunction,
+      Consumer<ByteString> setBodyFunction) {
+    if (CollectionUtils.isNullOrEmpty(params)) {
+      return path;
+    }
+    Escaper escaper = UrlEscapers.urlPathSegmentEscaper();
+    String encodedParams =
+        Joiner.on("&")
+            .join(
+                params.entries().stream()
+                    .map(
+                        entry ->
+                            String.format(
+                                "%s=%s",
+                                escaper.escape(entry.getKey()), escaper.escape(entry.getValue())))
+                    .collect(toImmutableList()));
+    if (method.equals(HttpMethod.GET)) {
+      return String.format("%s?%s", path, encodedParams);
+    }
+    putHeadersFunction.accept(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString());
+    setBodyFunction.accept(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
+    return path;
+  }
+
+  /**
+   * Creates a {@link Task} that does not use AppEngine for submission.
+   *
+   * <p>This uses the standard Cloud Tasks auth format to create and send an OIDC ID token set to
+   * the default service account. That account must have permission to submit tasks to Cloud Tasks.
+   */
+  private Task createNonAppEngineTask(
+      String path, HttpMethod method, Service service, Multimap<String, String> params) {
+    HttpRequest.Builder requestBuilder = HttpRequest.newBuilder().setHttpMethod(method);
+    path =
+        processRequestParameters(
+            path, method, params, requestBuilder::putHeaders, requestBuilder::setBody);
+    OidcToken.Builder oidcTokenBuilder =
+        OidcToken.newBuilder().setServiceAccountEmail(defaultServiceAccount);
+    // If the service is using IAP, add that as the audience for the token so the request can be
+    // appropriately authed. Otherwise, use the project name.
+    if (iapClientId != null) {
+      oidcTokenBuilder.setAudience(iapClientId);
+    } else {
+      oidcTokenBuilder.setAudience(projectId);
+    }
+    requestBuilder.setOidcToken(oidcTokenBuilder.build());
+    String totalPath = String.format("%s%s", getServer(service), path);
+    requestBuilder.setUrl(totalPath);
+    return Task.newBuilder().setHttpRequest(requestBuilder.build()).build();
   }
 
   /**
@@ -123,34 +204,21 @@ public class CloudTasksUtils implements Serializable {
         method.equals(HttpMethod.GET) || method.equals(HttpMethod.POST),
         "HTTP method %s is used. Only GET and POST are allowed.",
         method);
-    AppEngineHttpRequest.Builder requestBuilder =
-        AppEngineHttpRequest.newBuilder()
-            .setHttpMethod(method)
-            .setAppEngineRouting(
-                AppEngineRouting.newBuilder().setService(service.toString()).build());
-
-    if (!CollectionUtils.isNullOrEmpty(params)) {
-      Escaper escaper = UrlEscapers.urlPathSegmentEscaper();
-      String encodedParams =
-          Joiner.on("&")
-              .join(
-                  params.entries().stream()
-                      .map(
-                          entry ->
-                              String.format(
-                                  "%s=%s",
-                                  escaper.escape(entry.getKey()), escaper.escape(entry.getValue())))
-                      .collect(toImmutableList()));
-      if (method == HttpMethod.GET) {
-        path = String.format("%s?%s", path, encodedParams);
-      } else {
-        requestBuilder
-            .putHeaders(HttpHeaders.CONTENT_TYPE, MediaType.FORM_DATA.toString())
-            .setBody(ByteString.copyFrom(encodedParams, StandardCharsets.UTF_8));
-      }
+    // If the default service account is configured, send a standard non-AppEngine HTTP request
+    if (defaultServiceAccount != null) {
+      return createNonAppEngineTask(path, method, service, params);
+    } else {
+      AppEngineHttpRequest.Builder requestBuilder =
+          AppEngineHttpRequest.newBuilder()
+              .setHttpMethod(method)
+              .setAppEngineRouting(
+                  AppEngineRouting.newBuilder().setService(service.toString()).build());
+      path =
+          processRequestParameters(
+              path, method, params, requestBuilder::putHeaders, requestBuilder::setBody);
+      requestBuilder.setRelativeUri(path);
+      return Task.newBuilder().setAppEngineHttpRequest(requestBuilder.build()).build();
     }
-    requestBuilder.setRelativeUri(path);
-    return Task.newBuilder().setAppEngineHttpRequest(requestBuilder.build()).build();
   }
 
   /**
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -109,12 +109,6 @@ public final class RegistryConfig {
     }
 
     @Provides
-    @Config("serviceAccountEmails")
-    public static ImmutableList<String> provideServiceAccountEmails(RegistryConfigSettings config) {
-      return ImmutableList.copyOf(config.gcpProject.serviceAccountEmails);
-    }
-
-    @Provides
     @Config("projectIdNumber")
     public static long provideProjectIdNumber(RegistryConfigSettings config) {
       return config.gcpProject.projectIdNumber;
@@ -124,6 +118,18 @@ public final class RegistryConfig {
     @Config("locationId")
     public static String provideLocationId(RegistryConfigSettings config) {
       return config.gcpProject.locationId;
+    }
+
+    @Provides
+    @Config("serviceAccountEmails")
+    public static ImmutableList<String> provideServiceAccountEmails(RegistryConfigSettings config) {
+      return ImmutableList.copyOf(config.gcpProject.serviceAccountEmails);
+    }
+
+    @Provides
+    @Config("defaultServiceAccount")
+    public static Optional<String> provideDefaultServiceAccount(RegistryConfigSettings config) {
+      return Optional.ofNullable(config.gcpProject.defaultServiceAccount);
     }
 
     /**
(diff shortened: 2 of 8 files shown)


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2013)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 168 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSgf3oBSGX4-_T4WPV:-NUSgf3oBSGX4-_T4WPW:byxn5rq) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L168)):*
> ```Java
>             path, method, params, requestBuilder::putHeaders, requestBuilder::setBody);
>     OidcToken.Builder oidcTokenBuilder =
>         OidcToken.newBuilder().setServiceAccountEmail(defaultServiceAccount);
> ```

I'm somewhat surprised that this would just work. `defaultServiceAccount` is just a string, right? So what if we give it an account that we are trying to impersonate? Would it mint a OIDC token for `potus@wh.gov` just because we set `defaultServiceAccount` to it? My guess the answer is not. So I assume there's some magic going under the hood to figure out that the runtime is on App Engine and the default service account matches the string that's given, so it actually has sufficient permission to act as that account.

This is part of the consolidation that I would like to do, as you can see we have at least three ways to mint OIDC token now: [for nomulus tool](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/RequestFactoryModule.java;l=105?q=requestfac&ss=nomulus%2Fnomulus), [for the proxy](https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/handler/HttpsRelayServiceHandler.java;l=125?q=httpsre&ss=nomulus%2Fnomulus) and now here.

I'll have a more detailed write-up in the bug, so we don't necessary have to make changes here, but it would be more re-assuring to confirm that we cannot mint arbitrary OIDC token here.



_Code quote:_
```Java
    OidcToken.Builder oidcTokenBuilder =
        OidcToken.newBuilder().setServiceAccountEmail(defaultServiceAccount);
```
___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 171 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSi5GM0NVGx2ammiCn:-NUSi5GM0NVGx2ammiCo:b-n7nogw) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L171)):*
> ```Java
>     // If the service is using IAP, add that as the audience for the token so the request can be
>     // appropriately authed. Otherwise, use the project name.
>     if (iapClientId != null) {
> ```

When you tested it in QA, was it behind IAP? The way the auth mechanism is written, if the you use [IAP](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/IapHeaderAuthenticationMechanism.java;l=57?q=iap&ss=nomulus%2Fnomulus), the `AuthResult` is at `USER` level and requires a corresponding `User` entity. Did you create an entity for the service account in SQL?

On the other hand, if you don't use IAP and relies on the [mechanism](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/IdTokenAuthenticationBase.java) Pavlo wrote, you need the service account to be included in the `serviceAccountEmails` section in the config file and will get an `APP` level `AuthResult`. So that means you need to have the service account email on two fields in the config file (`serviceAccountEmails` and `defaultServiceAccount`). Is that how you tested it?

Again this is something that I'm working on consolidating, so there's no need to make extensive changes just yet. But still I would like to confirm that my theory is working correctly, so it is helpful to know what kind of tests you did and how you got the requests to be authenticated. 



_Code quote:_
```Java
    if (iapClientId != null) {
      oidcTokenBuilder.setAudience(iapClientId);
    } else {
      oidcTokenBuilder.setAudience(projectId);
    }
```
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 32 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUScOy234ahZOiao8ly:-NUScOy234ahZOiao8lz:b-64g4r9) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/config/files/default-config.yaml#L32)):*
> ```YAML
>   # The default service account with which the service is running. For example,
>   # on GAE this would be {project-id}@appspot.gserviceaccount.com
>   defaultServiceAccount: null
> ```

This could be just a boolean I think. Something like `sendHttpCloudTasks`. You can get the service account by injecting an [`ApplicationDefaultCredential`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/config/CredentialModule.java;l=163?q=defaultcredential&ss=nomulus%2Fnomulus), so there is no need to specify one here.

This would be temporary as well, as I am hopeful that we can achieve a seamless transition.



_Code quote:_
```YAML
defaultServiceAccount: null
```
___
*[`core/src/main/java/google/registry/tools/ServiceConnection.java` line 144 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSjlRRBvw2r8ds1DAm:-NUSjlRRBvw2r8ds1DAn:b-jfyu8s) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/tools/ServiceConnection.java#L144)):*
> ```Java
>   }
> 
>   public static URL getServer(Service service) {
> ```

Seems like we should move this to the util package, given that it is now being used for more than just the tool.



_Code quote:_
```Java
public static URL getServer(Service service) {
```

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2013)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 168 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSgf3oBSGX4-_T4WPV:-NUSmUv4DTciZaMaTe1-:b-4xjq78) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L168)):*

I'm somewhat surprised that this would just work. `defaultServiceAccount` is just a string, right? So what if we give it an account that we are trying to impersonate? Would it mint a OIDC token for `potus@wh.gov` just because we set `defaultServiceAccount` to it? My guess the answer is not. So I assume there's some magic going under the hood to figure out that the runtime is on App Engine and the default service account matches the string that's given, so it actually has sufficient permission to act as that account.

This is part of the consolidation that I would like to do, as you can see we have at least three ways to mint OIDC token now: [for nomulus tool](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/RequestFactoryModule.java;l=105?q=requestfac&ss=nomulus%2Fnomulus), [for the proxy](https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/handler/HttpsRelayServiceHandler.java;l=125?q=httpsre&ss=nomulus%2Fnomulus) and now here.

I'll have a more detailed write-up in the bug, so we don't necessary have to make changes here, but it would be more re-assuring to confirm that we cannot mint arbitrary OIDC token here.
</blockquote></details>

Yes, it is required that the account that we're using is the same account that we're currently running as (the default service account). If you try to do anything else it won't work. 
___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 171 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSi5GM0NVGx2ammiCn:-NUSmoT59YGdQwucV-ii:b-3fzewh) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L171)):*
I tested it with both IAP and non-IAP. 

>  So that means you need to have the service account email on two fields in the config file (serviceAccountEmails and defaultServiceAccount). Is that how you tested it?
yes

The only necessary change to make it work with IAP (once I got it working with non-IAP) was to add the IAP client ID as the audience 
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 32 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUScOy234ahZOiao8ly:-NUSn1872m1dLrbeJzp5:bd5uwyp) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/config/files/default-config.yaml#L32)):*
> you can get the service account by injecting an ApplicationDefaultCredential

How sure are you of this? The credentials don't have an API point for determining the email address of the user. In addition, how that information is actually stored depends on the subclass of `GoogleCredential` that we're actually using (which will vary, and isn't guaranteed).  
___
*[`core/src/main/java/google/registry/tools/ServiceConnection.java` line 144 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSjlRRBvw2r8ds1DAm:-NUSnkgU0PBenVnj6gpo:bqyv5sx) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/tools/ServiceConnection.java#L144)):*

Seems like we should move this to the util package, given that it is now being used for more than just the tool.
</blockquote></details>

the `util` module can't use config points

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2013)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 168 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSgf3oBSGX4-_T4WPV:-NUSqxie9Lhn__soNBLA:b-57s4kn) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L168)):*

Yes, it is required that the account that we're using is the same account that we're currently running as (the default service account). If you try to do anything else it won't work. 
</blockquote></details>

let me check to see if we can use something similar to the proxy authentication mechanism here (it would be replacing the OidcToken.Builder)

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2013)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 168 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSgf3oBSGX4-_T4WPV:-NUTK6ri1YjiyY6bD_Sv:b9s24pg) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L168)):*

let me check to see if we can use something similar to the proxy authentication mechanism here (it would be replacing the OidcToken.Builder) 
</blockquote></details>

very weird, if I try to do this (without IAP) and just set the access token on the request like in HttpsRelayServiceHandler, the OAuth authentication mechanism says that the OAuth client ID is "anonymous" when parsing the access token. I don't really know why that's happening here and not in HttpsRelayServiceHandler in the proxy.

jianglai (approved):
Given that this works now, I'm inclined to let it merge as it so we have a record of it, before we embark on taking a holistic view of our usage of OIDC tokens and making it consistent accross the code base.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2013#-:-NUvrAAcDI6OM4VnJZwO:brkdy53)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 168 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSgf3oBSGX4-_T4WPV:-NUvnHqw34ChitOQEzWT:bq0qujn) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L168)):*

very weird, if I try to do this (without IAP) and just set the access token on the request like in HttpsRelayServiceHandler, the OAuth authentication mechanism says that the OAuth client ID is "anonymous" when parsing the access token. I don't really know why that's happening here and not in HttpsRelayServiceHandler in the proxy. 
</blockquote></details>

Did you mean that you called `GoogleCredentials.getAccessToken().getTokenValue()` on the application default credential? I think that is to be expected as the ADC you get from App Engine is different from that which you get from GAE/GKE. The OAuth token minted for GAE ADC might not have client ID in it, even though supposedly the service account itself does have client IDs associated to it:

https://pantheon.corp.google.com/iam-admin/serviceaccounts?project=domain-registry

But we should not bother with OAuth here I think, the goal should be to have a single method to create OIDC tokens, not OAuth tokens.

No need to fixate on this part now. As long as it is working, we can proceed and work on the consolidation later.
___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 171 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSi5GM0NVGx2ammiCn:-NUvpTVy1s_AOk2Zw4_4:byl6kmr) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L171)):*

I tested it with both IAP and non-IAP. 

>  So that means you need to have the service account email on two fields in the config file (serviceAccountEmails and defaultServiceAccount). Is that how you tested it?
yes

The only necessary change to make it work with IAP (once I got it working with non-IAP) was to add the IAP client ID as the audience 
</blockquote></details>

But how would it work with IAP when you don't have the `User` entity that's mapped to the email address? Doesn't it require it to work?

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/IapHeaderAuthenticationMechanism.java;l=53;drc=55243e7cf62ead0c792f5f68c9d155e899261ed7;bpv=0;bpt=0?q=iapheader&ss=nomulus%2Fnomulus
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 32 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUScOy234ahZOiao8ly:-NUvqNwYDxTgfIEQvHGb:bz2wybx) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/config/files/default-config.yaml#L32)):*

> you can get the service account by injecting an ApplicationDefaultCredential

How sure are you of this? The credentials don't have an API point for determining the email address of the user. In addition, how that information is actually stored depends on the subclass of `GoogleCredential` that we're actually using (which will vary, and isn't guaranteed).  
</blockquote></details>

I was thinking to use the `GoogleCredentials` as an `IdTokenProvider` like here:

https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/handler/HttpsRelayServiceHandler.java;l=121;drc=55243e7cf62ead0c792f5f68c9d155e899261ed7;bpv=0;bpt=0?q=http&ss=nomulus%2Fnomulus

But you are right that we can't really tell until we tried. The fact that this is about the only point in the code base where we want the GAE ADC to make an OIDC token may muddle things a bit, given how weird the App Engine default credential is.
___
*[`core/src/main/java/google/registry/tools/ServiceConnection.java` line 144 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSjlRRBvw2r8ds1DAm:-NUvr--26il8bKJTlbXO:b-xhyb3) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/tools/ServiceConnection.java#L144)):*

the `util` module can't use config points 
</blockquote></details>

Yeah, that's annoying. It is almost as if we should have another util-like package inside core that hosts common classes and methods that are used by multiple packages.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2013)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 168 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSgf3oBSGX4-_T4WPV:-NV0jFX436k26cOtSMEO:b1v0q5a) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L168)):*

Did you mean that you called `GoogleCredentials.getAccessToken().getTokenValue()` on the application default credential? I think that is to be expected as the ADC you get from App Engine is different from that which you get from GAE/GKE. The OAuth token minted for GAE ADC might not have client ID in it, even though supposedly the service account itself does have client IDs associated to it:

https://pantheon.corp.google.com/iam-admin/serviceaccounts?project=domain-registry

But we should not bother with OAuth here I think, the goal should be to have a single method to create OIDC tokens, not OAuth tokens.

No need to fixate on this part now. As long as it is working, we can proceed and work on the consolidation later.
</blockquote></details>

Well, if we want to have a common way of creating OIDC tokens in all of the clients, we will need to figure out why generating a non-IAP OIDC token doesn't contain the required audience when we use the credentials that we have here. Otherwise, this client won't be able to have the "works in both IAP and non-IAP worlds" characteristic that we both want
___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 171 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUSi5GM0NVGx2ammiCn:-NV0i8yg9mSUWSyqjO9H:btlru4) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L171)):*

But how would it work with IAP when you don't have the `User` entity that's mapped to the email address? Doesn't it require it to work?

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/IapHeaderAuthenticationMechanism.java;l=53;drc=55243e7cf62ead0c792f5f68c9d155e899261ed7;bpv=0;bpt=0?q=iapheader&ss=nomulus%2Fnomulus
</blockquote></details>

Yes, it requires the User entity because currently the IAP authenticator only checks for User objects (for a proof of concept / first step this is fine, of course)
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 32 at r1](https://reviewable.io/reviews/google/nomulus/2013#-NUScOy234ahZOiao8ly:-NV0kAh79mEaifD3oVxJ:b-jprxwt) ([raw file](https://github.com/google/nomulus/blob/4643f9b0415feb962ccd20814ba52abe59fb0400/core/src/main/java/google/registry/config/files/default-config.yaml#L32)):*

I was thinking to use the `GoogleCredentials` as an `IdTokenProvider` like here:

https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/handler/HttpsRelayServiceHandler.java;l=121;drc=55243e7cf62ead0c792f5f68c9d155e899261ed7;bpv=0;bpt=0?q=http&ss=nomulus%2Fnomulus

But you are right that we can't really tell until we tried. The fact that this is about the only point in the code base where we want the GAE ADC to make an OIDC token may muddle things a bit, given how weird the App Engine default credential is.
</blockquote></details>

Ah yes. Good news: using the default credentials as an IdTokenProvider then requesting an ID token with the proper audience (`domain-registry-qa`, in this case) does seem to work. 

Less good news: oddly, it doesn't seem like it works in the IAP case even when doing basically the same thing that we do in RequestFactoryModule -- setting both Authorization and Proxy-Authorization. Oddly enough, I even printed out the token we were attempting to use on Proxy-Authorization and was able to make an IAP-enabled curl request to the registrar page just fine using that token.

PR #2014, opened 2023-04-28
Prepare switch of credential annotation

weiminyu:
Prepare the switch from DefaultCredential to ApplicationCredential.
    
In nomulus tools, start using the new annotation. This is tested by
successfully using the nomulus curl command, which actually needs a
valid credential to work.
    
For remaining use cases of the old annotation in Nomulus server, add
some harmless code that relies on the new credential to work. Such code can be
triggered using the curl command. Once they are tested in sandbox and production,
we will switch the all uses of the old annotation to the new one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2014)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2014)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2025, opened 2023-05-09
Refactor OIDC-based auth mechanism

jianglai:
IAP and regular OIDC auth mechanisms are unified under a base class that
produces either APP or USER level AuthResult based on the principal email
found in the OIDC token.

Also moved some enum classes to better organize code structure.

TESTED=deployed to alpha and verified Cloud Scheduler (which uses regular OIDC) still works.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2025)


gbrodman (changes requested):
> Reviewed 23 of 32 files at r1, 1 of 1 files at r2, 1 of 2 files at r6, 1 of 1 files at r7, 1 of 1 files at r9, 1 of 1 files at r10, 2 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: 30 of 32 files reviewed, 6 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/Auth.java` line 26 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1AxAb6NgRJaFH6F8j:-NV1AxAb6NgRJaFH6F8k:b-2yibs8) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/Auth.java#L26)):*
> ```Java
> 
>   /**
>    * Allows anyone to access, doesn't attempt to authenticate user.
> ```

this is valid both with or without the "to" because "access" is both a verb and a noun. English is great! 
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 71 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1JFaA9oqdUbqnR1Y-:-NV1JFaA9oqdUbqnR1Y0:b3k7zkr) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L71)):*
> ```Java
>   TokenVerifier provideIapTokenVerifier(
>       @Config("projectId") String projectId, @Config("projectIdNumber") long projectIdNumber) {
>     String audience = String.format("/projects/%d/apps/%s", projectIdNumber, projectId);
> ```

we will need to change this at some point, since I believe the audience will change once we move from GAE to GKE. That shouldn't be an issue right? 
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 78 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1EwhN6pkBpUm1g_YR:-NV1EwhN6pkBpUm1g_YS:bgurmep) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L78)):*
> ```Java
>   @Bearer
>   @Singleton
>   TokenVerifier provideBearerkenVerifier(@Config("projectId") String projectId) {
> ```

something weird with the method name here
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 90 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1ELC--P_m7lQAPujF:-NV1ELC--P_m7lQAPujG:b-r35bjc) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L90)):*
> ```Java
> 
>   @Provides
>   @Bearer
> ```

"Bearer" doesn't seem like a good name because it doesn't really seem to describe what it's talking about, which is specifically a non-IAP standard OAuth ID token
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 94 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1E2qW72zbSu8xQh8u:-NV1E2qX0vRDjsTL9ytk:bru1yt7) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L94)):*
> ```Java
>   TokenExtractor providesBearerTokenExtractor() {
>     return request -> {
>       String rawToken = request.getHeader(PROXY_AUTHORIZATION);
> ```

we should never be reading the proxy-authorization header, since the only things setting it should be for IAP
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 118 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1HxCl8JbZiE_SK1hv:-NV1HxCl8JbZiE_SK1hw:bded02o) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L118)):*
> ```Java
>   @FunctionalInterface
>   interface TokenExtractor {
>     String extract(HttpServletRequest request);
> ```

probably should be an Optional<String> to avoid nulls 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 153 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1Ivgf8-4PDrCtufwT:-NV1Ivgf8-4PDrCtufwU:biw8nx2) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L153)):*
> ```Java
>    * for certain GAE endpoints).
>    *
>    * <p>Note that this mechanism will try to first extract the token under the "proxy-authorization"
> ```

This is backward. We should be setting the Authorization header to be the non-IAP token and the Proxy-Authorization header to be the IAP token.

jianglai (commented):
> Reviewed 27 of 32 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 2 files at r6, 1 of 1 files at r7, 1 of 1 files at r9, 1 of 1 files at r10, 4 of 4 files at r11, 4 of 6 files at r12, 1 of 1 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 6 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/request/auth/Auth.java` line 26 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1AxAb6NgRJaFH6F8j:-NV555YHBcJDRTUCdCe2:b-18nym2) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/Auth.java#L26)):*

this is valid both with or without the "to" because "access" is both a verb and a noun. English is great! 
</blockquote></details>

lol. I think IntelliJ was trying to be too helpful :D
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 71 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1JFaA9oqdUbqnR1Y-:-NV57agNEqWBNoZDT7d-:bw5tisu) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L71)):*

we will need to change this at some point, since I believe the audience will change once we move from GAE to GKE. That shouldn't be an issue right? 
</blockquote></details>

Looks like it:

https://cloud.google.com/iap/docs/signed-headers-howto#verifying_the_jwt_payload

Added a TODO.
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 78 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1EwhN6pkBpUm1g_YR:-NV5G6Zd1dspKoK7Jyhb:bt2vr1b) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L78)):*

something weird with the method name here
</blockquote></details>

Fixed.
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 90 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1ELC--P_m7lQAPujF:-NV5G7Ls8Hm76Z12zetO:b-3r09pt) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L90)):*

"Bearer" doesn't seem like a good name because it doesn't really seem to describe what it's talking about, which is specifically a non-IAP standard OAuth ID token
</blockquote></details>

Updated the naming scheme throughout the PR. Hopefully it is clearer now.
___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 94 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1E2qW72zbSu8xQh8u:-NV5IWSm5qU7-Xgik57E:blk7hib) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L94)):*

we should never be reading the proxy-authorization header, since the only things setting it should be for IAP
</blockquote></details>

Discussed this a bit offline. The thinking behind checking both headers is to allow for an easier transition from OAuth to regular OIDC to IAP OIDC.

1. OAuth under Authorization + Regular OIDC under Proxy-Authroization
2. Regular OIDC under Proxy-Authroization + IAP OIDC under Authorization
3. IAP under Authorization (or Proxy-Authorization).

Because OAuth only checks for Authorization, if we want both OAuth and *any* OIDC to work at the same time, OIDC must be under Proxy-Authorization. I think having the server allow both auth AND having the client send both headers will give us the most flexibility during the migration as the likelihood that a re-deployment is needed to fix any auth-related issue is  minimized.

This noticeably does not include Cloud Tasks and Cloud Scheduler, which only supports the Authorization header, as far as we can tell.

After the migration is over, we can decide if we want to keep both IAP and regular OIDC. I think it's worthwhile to do so as it doesn't really take much effort to support both anyway. In that case I agree that it is more semantically correct to have IAP tokens occupy the Proxy-Authorization header and the regular occupy the Authorization header. But still, for maximum flexibility, isn't it better for both auth to support *either* header? We'd need to rework the regular OIDC auth a bit for it to try *validate* either headers, instead of just trying to *extract* either, in that case, and we'd be hijacking the meaning of Proxy-Authorization for regular OIDC auth. It seems a worthy trade off for the flexibility it provides.

What are you thoughts on the matter?
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 118 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1HxCl8JbZiE_SK1hv:-NV5I4Cu54bJncqNt435:b9s8lzi) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L118)):*

probably should be an Optional<String> to avoid nulls 
</blockquote></details>

You meant a null `request`? That should not be possible because when this function is called, there must be a non-null `request` that is routed to the action that attempts to authenticate it.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 153 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1Ivgf8-4PDrCtufwT:-NV5ITxyCYKw_OJ5f61o:bsjte78) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L153)):*

This is backward. We should be setting the Authorization header to be the non-IAP token and the Proxy-Authorization header to be the IAP token. 
</blockquote></details>

See above comment.

jianglai (commented):
> Reviewed 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 6 unresolved discussions (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 2 of 6 files at r12, 1 of 1 files at r13, 1 of 2 files at r14, 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 94 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1E2qW72zbSu8xQh8u:-NVBNN9l8JnJ6RMqPEyW:b-iqu7km) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L94)):*

Discussed this a bit offline. The thinking behind checking both headers is to allow for an easier transition from OAuth to regular OIDC to IAP OIDC.

1. OAuth under Authorization + Regular OIDC under Proxy-Authroization
2. Regular OIDC under Proxy-Authroization + IAP OIDC under Authorization
3. IAP under Authorization (or Proxy-Authorization).

Because OAuth only checks for Authorization, if we want both OAuth and *any* OIDC to work at the same time, OIDC must be under Proxy-Authorization. I think having the server allow both auth AND having the client send both headers will give us the most flexibility during the migration as the likelihood that a re-deployment is needed to fix any auth-related issue is  minimized.

This noticeably does not include Cloud Tasks and Cloud Scheduler, which only supports the Authorization header, as far as we can tell.

After the migration is over, we can decide if we want to keep both IAP and regular OIDC. I think it's worthwhile to do so as it doesn't really take much effort to support both anyway. In that case I agree that it is more semantically correct to have IAP tokens occupy the Proxy-Authorization header and the regular occupy the Authorization header. But still, for maximum flexibility, isn't it better for both auth to support *either* header? We'd need to rework the regular OIDC auth a bit for it to try *validate* either headers, instead of just trying to *extract* either, in that case, and we'd be hijacking the meaning of Proxy-Authorization for regular OIDC auth. It seems a worthy trade off for the flexibility it provides.

What are you thoughts on the matter?
</blockquote></details>

It seems like it might possibly be overkill since the only place the GAE based OAuth is used is with the Nomulus tool, but I suppose it might prevent us from getting in the situation where we forget to add our own users to the DB before we require our-own-user based OAuth for the nomulus tool 

Question: have you, by any chance, tested whether or not the regular (non-IAP) token verifier can verify the tokens provided by the existing credentials bundle's HTTP request initializer (the step that is done in RequestFactoryModule line ~73, where it sets the standard Authorization header? If so, I don't think we'd need to bother checking Proxy-Authorization here because we could check+verify+use the Authorization header value in all non-IAP cases. 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 118 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1HxCl8JbZiE_SK1hv:-NVB5EBZ79r6b-dCj0yp:b-cel0yj) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L118)):*

You meant a null `request`? That should not be possible because when this function is called, there must be a non-null `request` that is routed to the action that attempts to authenticate it.
</blockquote></details>

no, it should return an Optional because the result can be null
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 91 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5Ory7N8PViue68hK:-NVB5Ory7N8PViue68hL:btn42ds) ([raw file](https://github.com/google/nomulus/blob/77c740364d10b1ecd57bc2055790918ef03f9539/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L91)):*
> ```Java
>     String email = (String) token.getPayload().get("email");
>     if (email == null) {
>       logger.atInfo().log("No email address from the OIDC token");
> ```

nitty, but this should probably be a warning since it's very unexpected. Maybe print out the payload's key set as well to see what fields we do have? 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 94 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVBKdor7r2Fc6pYkDK_:-NVBKdos3WRZZTkmtvqP:b-kkaaqw) ([raw file](https://github.com/google/nomulus/blob/77c740364d10b1ecd57bc2055790918ef03f9539/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L94)):*
> ```Java
>       return AuthResult.NOT_AUTHENTICATED;
>     }
>     Optional<User> maybeUser = UserDao.loadUser(email);
> ```

One benefit of this is that it will allow us to use our DB-stored users when not in IAP mode -- this is necessary for the Nomulus tool. I'm not sure if that was something you directly thought of, but yay. 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 102 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5tqWDlp7erFFX7Qc:-NVB5tqWDlp7erFFX7Qd:b-9952g4) ([raw file](https://github.com/google/nomulus/blob/77c740364d10b1ecd57bc2055790918ef03f9539/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L102)):*
> ```Java
>       return AuthResult.create(APP);
>     }
>     logger.atInfo().log("No service account found for email address %s", email);
> ```

maybe this should be a warning as well? We probably won't get many situations in which a person authenticates but isn't one of our users

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 94 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1E2qW72zbSu8xQh8u:-NVVurGm1AvqpoFwvSI0:b-zcjavk) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L94)):*

It seems like it might possibly be overkill since the only place the GAE based OAuth is used is with the Nomulus tool, but I suppose it might prevent us from getting in the situation where we forget to add our own users to the DB before we require our-own-user based OAuth for the nomulus tool 

Question: have you, by any chance, tested whether or not the regular (non-IAP) token verifier can verify the tokens provided by the existing credentials bundle's HTTP request initializer (the step that is done in RequestFactoryModule line ~73, where it sets the standard Authorization header? If so, I don't think we'd need to bother checking Proxy-Authorization here because we could check+verify+use the Authorization header value in all non-IAP cases. 
</blockquote></details>

OAuth is used in at least two cases: nomulus tool and proxy.

I just did some testing. I didn't try nomulus tool because it requires USER level auth with entries in the DB, but on the proxy side, I changed the line [here](https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/handler/HttpsRelayServiceHandler.java;l=126?q=httpsre&ss=nomulus%2Fnomulus) to used the project Id as the audience and under the Authorization header. I then deployed this PR on the service side, adding the proxy service account to the config allowlist, and verified that the requests are successfully authenticated by regular OIDC auth. I suppose it should work similarly with nomulus tool if we set up the DB properly. This is good news because it means OIDC token generation can mostly likely be simplifed to a static function that takes a `GoogleCredentials` and a `String` audience and generate the appropriate IAP or non-IAP token based on the audience.

As far calling the request initializer directly in `RequestFactoryModule`, that injects an OAuth access token to the request and would not be usable by OIDC-based auth methods.

So if we want to keep OAuth working with regular OIDC, the OIDC token has to be under a different header, it doesn't have to by proxy-authroization because we control which headers to check. In this case I think it's probably convenient to use proxy-authorization as it is an established header name. I've added a TODO to remove it after the migration is complete. It seems like web-based authentication (for the console) would always require IAP as otherwise we'd have to implement not only the three-legged OAuth flow, but also session cookies to keep track of logged in sessions. That's too much work for optionally support the non-IAP based OIDC flow.

So my proposal is to keep regular OIDC around, but only checking the Authorization header once everything has migrated. We should have no need for it but it doesn't hurt to keep it around. It also benefits the open source effort as it is not tied to IAP.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 118 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1HxCl8JbZiE_SK1hv:-NVZhogG3Zn9dtpvyzMn:b5lz9xk) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L118)):*

no, it should return an Optional because the result can be null
</blockquote></details>

Added `@Nullable` annotation. Since this class is only ever used here, and the nullness is fully taken into account, I don't see how `Optional` would make a difference. To me, you make the returned value optional if you want to signal to users that you don't necessarily control that there's a chance the returned value is empty, which is not the case here, where the user is immediately proceeding it. 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 91 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5Ory7N8PViue68hK:-NVW5Rg068F-CYIwT88U:b-khbr18) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L91)):*

nitty, but this should probably be a warning since it's very unexpected. Maybe print out the payload's key set as well to see what fields we do have? 
</blockquote></details>

Since the end point (without IAP) can be reached by anyone, I think it is fair to log it at info because we expect invalid tokens to happen from time to time. Added logging.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 94 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVBKdor7r2Fc6pYkDK_:-NVW5CvzEY7wJxKH9EhH:b-el18nj) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L94)):*

One benefit of this is that it will allow us to use our DB-stored users when not in IAP mode -- this is necessary for the Nomulus tool. I'm not sure if that was something you directly thought of, but yay. 
</blockquote></details>

Yes, I think IAP/non-IAP and USER/APP level auth should be orthogonal. There is no reason why one type of auth level is tied to a particular auth mechanism.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 102 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5tqWDlp7erFFX7Qc:-NVZeKDM1LV3s89sBTpo:bjirfyl) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L102)):*

maybe this should be a warning as well? We probably won't get many situations in which a person authenticates but isn't one of our users 
</blockquote></details>

Similarly, I feel that if we are looking for a particular login issue, we obviously would know which lines to check. Otherwise, it probably doesn't make sense to be spammed for failed logins that are out side of our control anyway.

jianglai (commented):
> Reviewed 2 of 2 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

jianglai (commented):
> Reviewed 1 of 1 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 4 unresolved discussions (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 1 of 6 files at r12, 2 of 2 files at r16, 1 of 1 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/AuthModule.java` line 94 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1E2qW72zbSu8xQh8u:-NV_kJhS1-Dws0B65sex:b-v4st90) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/AuthModule.java#L94)):*

OAuth is used in at least two cases: nomulus tool and proxy.

I just did some testing. I didn't try nomulus tool because it requires USER level auth with entries in the DB, but on the proxy side, I changed the line [here](https://cs.opensource.google/nomulus/nomulus/+/master:proxy/src/main/java/google/registry/proxy/handler/HttpsRelayServiceHandler.java;l=126?q=httpsre&ss=nomulus%2Fnomulus) to used the project Id as the audience and under the Authorization header. I then deployed this PR on the service side, adding the proxy service account to the config allowlist, and verified that the requests are successfully authenticated by regular OIDC auth. I suppose it should work similarly with nomulus tool if we set up the DB properly. This is good news because it means OIDC token generation can mostly likely be simplifed to a static function that takes a `GoogleCredentials` and a `String` audience and generate the appropriate IAP or non-IAP token based on the audience.

As far calling the request initializer directly in `RequestFactoryModule`, that injects an OAuth access token to the request and would not be usable by OIDC-based auth methods.

So if we want to keep OAuth working with regular OIDC, the OIDC token has to be under a different header, it doesn't have to by proxy-authroization because we control which headers to check. In this case I think it's probably convenient to use proxy-authorization as it is an established header name. I've added a TODO to remove it after the migration is complete. It seems like web-based authentication (for the console) would always require IAP as otherwise we'd have to implement not only the three-legged OAuth flow, but also session cookies to keep track of logged in sessions. That's too much work for optionally support the non-IAP based OIDC flow.

So my proposal is to keep regular OIDC around, but only checking the Authorization header once everything has migrated. We should have no need for it but it doesn't hurt to keep it around. It also benefits the open source effort as it is not tied to IAP.
</blockquote></details>

Ah right I forgot about the proxy, my mistake. 

Yep, that sounds good, we should add the capability of using a non-IAP OIDC token in the Proxy-Authorization header in the Nomulus tool. 

You're 100% correct about the front end always using IAP because it means there are a toooon of other problems we don't have to deal with. Using Proxy-Authorization here seems unfortunate but then probably the best solution, yeah. 

And as discussed in person, once we're done moving everything over to GKE and IAP we can just use the Authorization header for everything. I'm not so much concerned about tying ourselves to IAP because we're going to be relying on that for the registrar console anyway. 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 118 at r11](https://reviewable.io/reviews/google/nomulus/2025#-NV1HxCl8JbZiE_SK1hv:-NVa-8EC3ly5qVd1i65p:bleskji) ([raw file](https://github.com/google/nomulus/blob/a54be27146f850026271495412b36b4e20d18b73/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L118)):*

Added `@Nullable` annotation. Since this class is only ever used here, and the nullness is fully taken into account, I don't see how `Optional` would make a difference. To me, you make the returned value optional if you want to signal to users that you don't necessarily control that there's a chance the returned value is empty, which is not the case here, where the user is immediately proceeding it. 
</blockquote></details>

eh yeah I guess this isn't a public method so it's fine 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 91 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5Ory7N8PViue68hK:-NV_xHmg5mgRP8xhE7MG:b-uu0egi) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L91)):*

Since the end point (without IAP) can be reached by anyone, I think it is fair to log it at info because we expect invalid tokens to happen from time to time. Added logging.
</blockquote></details>

But if we get to this point it means that the user passed in a valid JWT, just one that doesn't have an email. If they passed in a nonexistent or invalid token we would have already returned out a few lines ago. I don't think we'd ever expect invalid tokens? 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 102 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5tqWDlp7erFFX7Qc:-NV_xxlb5ESGHOMyM52g:bo3ie1t) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L102)):*

Similarly, I feel that if we are looking for a particular login issue, we obviously would know which lines to check. Otherwise, it probably doesn't make sense to be spammed for failed logins that are out side of our control anyway.
</blockquote></details>

similar to above, we'll already have exited out for standard failed logins at this point so I think it's probably helpful to log at warning

jianglai (commented):
> Reviewed 2 of 2 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 91 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5Ory7N8PViue68hK:-NVa41_UC2YF3AU_BBkx:b-obizmm) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L91)):*

But if we get to this point it means that the user passed in a valid JWT, just one that doesn't have an email. If they passed in a nonexistent or invalid token we would have already returned out a few lines ago. I don't think we'd ever expect invalid tokens? 
</blockquote></details>

I guess you have a point, the JWTs that we are expecting (signed by Google, with or without IAP) should all have an email field. Changed to warning.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 102 at r15](https://reviewable.io/reviews/google/nomulus/2025#-NVB5tqWDlp7erFFX7Qc:-NVa54gq7dWApcbQjnv6:b-un2gp) ([raw file](https://github.com/google/nomulus/blob/0b2fba11ada6aa98045128ce1dfddd710a6b286a/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L102)):*

similar to above, we'll already have exited out for standard failed logins at this point so I think it's probably helpful to log at warning
</blockquote></details>

With IAP I don't think this would every happen, as the email must be whitelisted in IAP already before we receive the JWT. But if regular OIDC auth is used here, anyone could ask Google to mint a token for their email address, that would be a valid token, but we'd reject the request because we don't recognize that email.

It doesn't feel right to log this part as warning but the end user log message at info, so I added another warning after this.

gbrodman (approved):
> Reviewed 2 of 2 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2025)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2030, opened 2023-05-10
Wrap tm().loadByKey() in a transaction when caching is not enabled.

jianglai:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2030)

### DIFF
--- a/core/src/main/java/google/registry/model/EppResource.java
+++ b/core/src/main/java/google/registry/model/EppResource.java
@@ -403,7 +403,7 @@ public abstract class EppResource extends UpdateAutoTimestampEntity implements B
   public static ImmutableMap<VKey<? extends EppResource>, EppResource> loadCached(
       Iterable<VKey<? extends EppResource>> keys) {
     if (!RegistryConfig.isEppResourceCachingEnabled()) {
-      return tm().loadByKeys(keys);
+      return tm().transact(() -> tm().loadByKeys(keys));
     }
     return ImmutableMap.copyOf(cacheEppResources.getAll(keys));
   }
@@ -416,7 +416,7 @@ public abstract class EppResource extends UpdateAutoTimestampEntity implements B
    */
   public static <T extends EppResource> T loadCached(VKey<T> key) {
     if (!RegistryConfig.isEppResourceCachingEnabled()) {
-      return tm().loadByKey(key);
+      return tm().transact(() -> tm().loadByKey(key));
     }
     // Safe to cast because loading a Key<T> returns an entity of type T.
     @SuppressWarnings("unchecked")


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2030)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/EppResource.java` line 406 at r1](https://reviewable.io/reviews/google/nomulus/2030#-NV6JBgO8ddlhCNPV-ak:-NV6JBgP0w_3UpUuQxeK:bdvw8ud) ([raw file](https://github.com/google/nomulus/blob/38da3059469ef69e8221e5af53da3126b803c28d/core/src/main/java/google/registry/model/EppResource.java#L406)):*
> ```Java
>       Iterable<VKey<? extends EppResource>> keys) {
>     if (!RegistryConfig.isEppResourceCachingEnabled()) {
>       return tm().loadByKeys(keys);
> ```

What about here?
___
*[`core/src/main/java/google/registry/model/EppResource.java` line 419 at r1](https://reviewable.io/reviews/google/nomulus/2030#-NV6JCZX0-pIXp-74DWa:-NV6JCZX0-pIXp-74DWb:b6kw0dl) ([raw file](https://github.com/google/nomulus/blob/38da3059469ef69e8221e5af53da3126b803c28d/core/src/main/java/google/registry/model/EppResource.java#L419)):*
> ```Java
>   public static <T extends EppResource> T loadCached(VKey<T> key) {
>     if (!RegistryConfig.isEppResourceCachingEnabled()) {
>       return tm().transact(() -> tm().loadByKey(key));
> ```

Why not at callsites?  In most contexts (outside WHOIS apparently), when this is called, it will still be inside a transaction already anyway, right?  E.g. in other EPP flows?  Is there an additional cost to always attempting to start a new tx?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2030)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/model/EppResource.java` line 406 at r1](https://reviewable.io/reviews/google/nomulus/2030#-NV6JBgO8ddlhCNPV-ak:-NV6KMW1EUTQnxgDWMK8:b-ypcbo4) ([raw file](https://github.com/google/nomulus/blob/38da3059469ef69e8221e5af53da3126b803c28d/core/src/main/java/google/registry/model/EppResource.java#L406)):*

What about here?
</blockquote></details>

Both call sites are in flows which are already in a transaction. But wrapping them here doesn't seem to hurt.
___
*[`core/src/main/java/google/registry/model/EppResource.java` line 419 at r1](https://reviewable.io/reviews/google/nomulus/2030#-NV6JCZX0-pIXp-74DWa:-NV6JwySBnUAJ-1CN1Xn:b-3r7jmy) ([raw file](https://github.com/google/nomulus/blob/38da3059469ef69e8221e5af53da3126b803c28d/core/src/main/java/google/registry/model/EppResource.java#L419)):*

Why not at callsites?  In most contexts (outside WHOIS apparently), when this is called, it will still be inside a transaction already anyway, right?  E.g. in other EPP flows?  Is there an additional cost to always attempting to start a new tx?
</blockquote></details>

Starting a new transaction inside a transaction is a no-op:

https://cs.opensource.google/nomulus/nomulus/+/refs/heads/master:core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java;l=167;drc=1aef31efff48fe656b6429d65257a5611e980b10?q=jpatran&ss=nomulus%2Fnomulus

Given that, it probably doesn't hurt to always preemptively wrap database access in a transaction, so call sites don't have to worry about it.

Technically, we could even go a step further and change the implementation of `tm().loadByKey()` and other methods to always run in a transaction. But I guess we do like to have to explicitly create transactions most of times so the programmers are aware of and more in control of the scope of the transactions?

jianglai:
Friendly ping.

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2030)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/model/EppResource.java` line 419 at r1](https://reviewable.io/reviews/google/nomulus/2030#-NV6JCZX0-pIXp-74DWa:-NVpEas92ivgMpHEh79a:bmthk1r) ([raw file](https://github.com/google/nomulus/blob/38da3059469ef69e8221e5af53da3126b803c28d/core/src/main/java/google/registry/model/EppResource.java#L419)):*

Starting a new transaction inside a transaction is a no-op:

https://cs.opensource.google/nomulus/nomulus/+/refs/heads/master:core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java;l=167;drc=1aef31efff48fe656b6429d65257a5611e980b10?q=jpatran&ss=nomulus%2Fnomulus

Given that, it probably doesn't hurt to always preemptively wrap database access in a transaction, so call sites don't have to worry about it.

Technically, we could even go a step further and change the implementation of `tm().loadByKey()` and other methods to always run in a transaction. But I guess we do like to have to explicitly create transactions most of times so the programmers are aware of and more in control of the scope of the transactions?
</blockquote></details>

Yeah we definitely don't want to always create transactions automatically like that, as it's often not the right thing to do because it'll result in running lots of little transactions instead of one atomic big one.

PR #2034, opened 2023-05-11
Remove internal auth mechanism

jianglai:
This auth is only used by Task Queue tasks and Cron jobs, which have migrated
to Cloud Tasks and Cloud Scheduler, respectively. Both are sending
standard HTTP requests authenticated by OIDC.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2034)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2034)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
blocking just in case because we don't want to submit this until we have pushed the relevant internal changes to production 

This basically makes the default service account configuration point required, right? Assuming so, we probably want to change the RegistryConfig (and CloudTasksUtils) to make it a straight-up non-nullable String instead of an Optional

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2034#-:-NVBZDJt6K8sqw-1y_rG:b-rtc0k0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2037, opened 2023-05-18
Add batching to the RefreshDnsForAllDomainsAction

sarahcaseybot:
This PR batches the domains read and processed in the RefreshDnsForAllDomainsAction. The batching is necessary to prevent issues with predicate locking as described in go/dr-psql-index-pitfall.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2037)

### DIFF
--- a/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
+++ b/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
@@ -21,10 +21,10 @@ import static google.registry.model.tld.Tlds.assertTldsExist;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.RequestParameters.PARAM_TLDS;
 import static google.registry.util.DateTimeUtils.END_OF_TIME;
+import static com.google.common.collect.Iterables.getLast;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
 import com.google.common.flogger.FluentLogger;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
@@ -115,11 +115,11 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
                       } catch (Throwable t) {
                         logger.atSevere().withCause(t).log(
                             "Error while enqueuing DNS refresh for domain '%s'.", domainName);
-                        response.setStatus(HttpStatus.SC_INTERNAL_SERVER_ERROR);
+                        response.setStatus(HttpStatus.SC_OK);
                       }
                     });
               });
-      lastInPreviousBatch = Iterables.getLast(domainBatch);
+      lastInPreviousBatch = getLast(domainBatch);
     } while (lastBatchSize == batchSize);
   }
 }
--- a/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
+++ b/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
@@ -30,6 +30,7 @@ import google.registry.request.Action;
 import google.registry.request.Parameter;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
+import java.util.Optional;
 import java.util.Random;
 import javax.inject.Inject;
 import org.apache.arrow.util.VisibleForTesting;
@@ -58,6 +59,8 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
+  private static final int DEFAULT_BATCH_SIZE = 250;
+
   private final Response response;
   private final ImmutableSet<String> tlds;
 
@@ -69,11 +72,11 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
   RefreshDnsForAllDomainsAction(
       Response response,
       @Parameter(PARAM_TLDS) ImmutableSet<String> tlds,
-      @Parameter("batchSize") int batchSize,
+      @Parameter("batchSize") Optional<Integer> batchSize,
       Random random) {
     this.response = response;
     this.tlds = tlds;
-    this.batchSize = batchSize;
+    this.batchSize = batchSize.orElse(DEFAULT_BATCH_SIZE);
     this.random = random;
   }
 
@@ -84,11 +87,8 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
     int smearMinutes = tm().transact(this::calculateSmearMinutes);
     ImmutableList<String> previousBatch = ImmutableList.of("");
     do {
-      // This temp variable is required since Java requires that an effectively final variable is
-      // used in the lambda expression below
-      ImmutableList<String> finalPreviousBatch = previousBatch;
-      previousBatch =
-          tm().transact(() -> refreshBatch(ImmutableList.copyOf(finalPreviousBatch), smearMinutes));
+      String lastInPreviousBatch = getLast(previousBatch);
+      previousBatch = tm().transact(() -> refreshBatch(lastInPreviousBatch, smearMinutes));
     } while (previousBatch.size() == batchSize);
   }
 
@@ -107,7 +107,7 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
     return Math.max(activeDomains.intValue() / 1000, 1);
   }
 
-  private ImmutableList<String> getBatch(ImmutableList<String> previousBatch) {
+  private ImmutableList<String> getBatch(String lastInPreviousBatch) {
     return tm().query(
             "SELECT domainName FROM Domain WHERE tld IN (:tlds) AND"
                 + " deletionTime = :endOfTime  AND domainName >"
@@ -115,15 +115,15 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
             String.class)
         .setParameter("tlds", tlds)
         .setParameter("endOfTime", END_OF_TIME)
-        .setParameter("lastInPreviousBatch", getLast(previousBatch))
+        .setParameter("lastInPreviousBatch", lastInPreviousBatch)
         .setMaxResults(batchSize)
         .getResultStream()
         .collect(toImmutableList());
   }
 
   @VisibleForTesting
-  ImmutableList<String> refreshBatch(ImmutableList<String> previousBatch, int smearMinutes) {
-    ImmutableList<String> domainBatch = getBatch(previousBatch);
+  ImmutableList<String> refreshBatch(String lastInPreviousBatch, int smearMinutes) {
+    ImmutableList<String> domainBatch = getBatch(lastInPreviousBatch);
     try {
       // Smear the task execution time over the next N minutes.
       requestDomainDnsRefresh(domainBatch, Duration.standardMinutes(random.nextInt(smearMinutes)));
--- a/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
+++ b/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
@@ -20,9 +20,11 @@ import static google.registry.dns.DnsUtils.requestDomainDnsRefresh;
 import static google.registry.model.tld.Tlds.assertTldsExist;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.RequestParameters.PARAM_TLDS;
+import static google.registry.util.DateTimeUtils.END_OF_TIME;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.flogger.FluentLogger;
 import google.registry.request.Action;
 import google.registry.request.Parameter;
@@ -58,7 +60,7 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
-  static final int BATCH_SIZE = 10;
+  static final int BATCH_SIZE = 250;
 
   @Inject Response response;
 
@@ -81,38 +83,40 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
     assertTldsExist(tlds);
     checkArgument(smearMinutes > 0, "Must specify a positive number of smear minutes");
     String lastInPreviousBatch = "";
-    while (true) {
+    int lastBatchSize;
+    do {
       String previousLastName = lastInPreviousBatch;
       ImmutableList<String> domainBatch =
-          ImmutableList.copyOf(
-              tm().transact(
-                      () ->
-                          tm().query(
-                                  "SELECT domainName FROM Domain WHERE tld IN (:tlds) AND"
-                                      + " deletionTime > :now  AND domainName >"
-                                      + " :lastInPreviousBatch ORDER BY domainName ASC",
-                                  String.class)
-                              .setParameter("tlds", tlds)
-                              .setParameter("now", clock.nowUtc())
-                              .setParameter("lastInPreviousBatch", previousLastName)
-                              .setMaxResults(BATCH_SIZE)
-                              .getResultStream()
-                              .collect(toImmutableList())));
-
-      if (domainBatch.isEmpty()) break;
-      domainBatch.forEach(
-          domainName -> {
-            try {
-              // Smear the task execution time over the next N minutes.
-              requestDomainDnsRefresh(
-                  domainName, Duration.standardMinutes(random.nextInt(smearMinutes)));
-            } catch (Throwable t) {
-              logger.atSevere().withCause(t).log(
-                  "Error while enqueuing DNS refresh for domain '%s'.", domainName);
-              response.setStatus(HttpStatus.SC_INTERNAL_SERVER_ERROR);
-            }
-          });
-      lastInPreviousBatch = domainBatch.get(domainBatch.size() - 1);
-    }
+          tm().transact(
+                  () ->
+                      tm().query(
+                              "SELECT domainName FROM Domain WHERE tld IN (:tlds) AND"
+                                  + " deletionTime = :endOfTime  AND domainName >"
+                                  + " :lastInPreviousBatch ORDER BY domainName ASC",
+                              String.class)
+                          .setParameter("tlds", tlds)
+                          .setParameter("endOfTime", END_OF_TIME)
+                          .setParameter("lastInPreviousBatch", previousLastName)
+                          .setMaxResults(BATCH_SIZE)
+                          .getResultStream()
+                          .collect(toImmutableList()));
+      lastBatchSize = domainBatch.size();
+      tm().transact(
+              () -> {
+                domainBatch.forEach(
+                    domainName -> {
+                      try {
+                        // Smear the task execution time over the next N minutes.
+                        requestDomainDnsRefresh(
+                            domainName, Duration.standardMinutes(random.nextInt(smearMinutes)));
+                      } catch (Throwable t) {
+                        logger.atSevere().withCause(t).log(
+                            "Error while enqueuing DNS refresh for domain '%s'.", domainName);
+                        response.setStatus(HttpStatus.SC_INTERNAL_SERVER_ERROR);
+                      }
+                    });
+              });
+      lastInPreviousBatch = Iterables.getLast(domainBatch);
+    } while (lastBatchSize == BATCH_SIZE);
   }
 }
--- a/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
+++ b/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
@@ -16,12 +16,12 @@ package google.registry.tools.server;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.Iterables.getLast;
 import static google.registry.dns.DnsUtils.requestDomainDnsRefresh;
 import static google.registry.model.tld.Tlds.assertTldsExist;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.RequestParameters.PARAM_TLDS;
 import static google.registry.util.DateTimeUtils.END_OF_TIME;
-import static com.google.common.collect.Iterables.getLast;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -30,7 +30,6 @@ import google.registry.request.Action;
 import google.registry.request.Parameter;
 import google.registry.request.Response;
 import google.registry.request.auth.Auth;
-import google.registry.util.Clock;
 import java.util.Random;
 import javax.inject.Inject;
 import org.apache.http.HttpStatus;
@@ -51,6 +50,9 @@ import org.joda.time.Duration;
  * doesn't get overloaded. A rough rule of thumb for Cloud DNS is 1 minute per every 1,000 domains.
  * This smears the updates out over the next N minutes. For small TLDs consisting of fewer than
  * 1,000 domains, passing in 1 is fine (which will execute all the updates immediately).
+ *
+ * <p>You may pass in a {@code batchSize} for the batched read of domains from the database. This is
+ * recommended to be somewhere between 200 and 500. The default value is 250.
  */
 @Action(
     service = Action.Service.TOOLS,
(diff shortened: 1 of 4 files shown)


weiminyu (commented):
Thought a bit more about this:  if the analysis in go/dr-psql-index-pitfall is correct, then the 'tld' index on "Domain" will cause the same problem.

Also, .app has 660K entries live domains and 1.5 million total. Refreshing all of them in a tight loop will add a big load to the SQL database, since each domain to be refreshed means one DnsRefreshRequest inserted. Also, when done in one transaction, we ill have to flush the JPA entitymanager to reduce the size of the persistent context.

Now I think the quickest fix is to break up the single transaction in batches.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037#-:-NVos_FbC9NUWIl3jrEz:b-xd51qs)* status: 0 of 3 files reviewed, all discussions resolved

core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
```diff
@@ -76,27 +80,39 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
...
-                            requestDomainDnsRefresh(
-                                domainName, Duration.standardMinutes(random.nextInt(smearMinutes)));
-                          } catch (Throwable t) {
-                            logger.atSevere().withCause(t).log(
-                                "Error while enqueuing DNS refresh for domain '%s'.", domainName);
-                            response.setStatus(HttpStatus.SC_INTERNAL_SERVER_ERROR);
-                          }
-                        }));
+    String lastInPreviousBatch = "";
+    while (true) {
```
CydeWeys, line 84: Infinite while loops are dangerous and aren't all that readable, as from a moment's glance it's not clear what conditions would cause the loop to terminate.

This will be much more readable using a do-while, which would look like, in pseudo-code:

```
do {
  get new batch;
  process batch;
} while { size of latest batch == BATCH_SIZE };
```

core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
```diff
@@ -76,27 +80,39 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
...
+      String previousLastName = lastInPreviousBatch;
+      ImmutableList<String> domainBatch =
+          ImmutableList.copyOf(
+              tm().transact(
+                      () ->
+                          tm().query(
+                                  "SELECT domainName FROM Domain WHERE tld IN (:tlds) AND"
+                                      + " deletionTime > :now  AND domainName >"
```
CydeWeys, line 92: Do we need to refresh DNS on domains that might be in pending delete?  Wouldn't their DNS already be yanked?  It's possible this condition could simply be deletionTime = END OF TIME.

core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java
```diff
@@ -56,6 +58,8 @@ public class RefreshDnsForAllDomainsAction implements Runnable {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
+  static final int BATCH_SIZE = 10;
```
CydeWeys, line 61: This seems like an awfully small batch size.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 4 files reviewed, 11 unresolved discussions (waiting on @github-code-scanning[bot] and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 93 at r2](https://reviewable.io/reviews/google/nomulus/2037#-NW8ozrI-hlZlw8zdCs2:-NW8ozrI-hlZlw8zdCs3:bd801fm) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L93)):*
> ```Java
>                                   "SELECT domainName FROM Domain WHERE tld IN (:tlds) AND"
>                                       + " deletionTime > :now  AND domainName >"
>                                       + " :lastInPreviousBatch ORDER BY domainName ASC",
> ```

It seems more logical to do the batching based on primary key (the repo ID) rather than one of the other columns.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 4 files reviewed, 10 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 61 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201104562:-NW9F20ZCf5KCb_uSD_Y:bhfuzl3) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L61)):*

This seems like an awfully small batch size.
</blockquote></details>

Talked to @weiminyu who recommended a batch size between 200 and 500. 
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 84 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201095605:-NW8p748-rqStGILGh5Z:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L84)):*

Infinite while loops are dangerous and aren't all that readable, as from a moment's glance it's not clear what conditions would cause the loop to terminate.

This will be much more readable using a do-while, which would look like, in pseudo-code:

```
do {
  get new batch;
  process batch;
} while { size of latest batch == BATCH_SIZE };
```
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 85 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201073712:-NW8qUgA5MCPjbbXJH7e:bcvfqyx) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L85)):*

Why is this extra variable necessary?
</blockquote></details>

Added the extra variable because a variable in a lambda expression should be effectivel final
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 87 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201092480:-NW8q19c2-UfMKjN8v7Q:b-g05ihi) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L87)):*

Why are you calling `ImmutableList.copyOf()` when the collector you're using is already `toImmutableList()`?
</blockquote></details>

Removed extra copy
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 92 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201103807:-NW8vjkYCbzp9f2xojxr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L92)):*

Do we need to refresh DNS on domains that might be in pending delete?  Wouldn't their DNS already be yanked?  It's possible this condition could simply be deletionTime = END OF TIME.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 93 at r2](https://reviewable.io/reviews/google/nomulus/2037#-NW8ozrI-hlZlw8zdCs2:-NW8wTh531vBTQQVavKO:b-kvvke0) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L93)):*

It seems more logical to do the batching based on primary key (the repo ID) rather than one of the other columns.
</blockquote></details>

I think that would be less efficient as that would require storing the domainBatch as a map of domainName->repoId when just the domainName is all we need to be loading in.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 107 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201075398:-NW9DEGuBr7rOLitVVS8:bxh2bgt) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L107)):*

Isn't this horribly inefficient?  This is individually updating one row at a time.  If you run a refresh on a big TLD this will be running hundreds of thousands of individual transactions.
</blockquote></details>

Added a transaction around the processing part of each batch so all the refreshes for a batch are combined into one transaction. 

Note, this creates a nested transaction which we discussed recently that we should try to avoid. It may be useful in the future to refactor all the call sites of requestDomainDnsRefresh to be in a transaction so that the transaction creation can be removed from the method.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 112 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201090851:-NW9EVfr3-b7iGRQYNhZ:b-hbdcif) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L112)):*

Note that, since you're swallowing the Exception here, I'm not sure that it makes sense to return an HTTP error status.
</blockquote></details>

Are you recommending we not set any response status in the catch block?
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 115 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201084060:-NW9Eo7u8qWsYp9rbQyX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L115)):*

`Iterables.getLast(...)` is a nicer way of doing this (and when you run it on a List it's O(1)).
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 115 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201093540:-NW9Euew9xgfdBBo02AB:b-y75y2p) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L115)):*

This logic seems to be flawed to me, as it will always run one more batch than necessary and do a final query against the DB that will return zero results.  You can stop looping once you've gotten a batch that is smaller than `BATCH_SIZE`.
</blockquote></details>

fixed with do while loop

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 4 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 87 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201092480:-NWJJdK23l0qp7VxkIo-:b-xuyf99) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L87)):*

Removed extra copy
</blockquote></details>

This looks really awkward. Can you move the transaction statement to a method?
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 107 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201075398:-NWJFQ5x1imclqUGUBTg:b-gy6pzv) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L107)):*

Added a transaction around the processing part of each batch so all the refreshes for a batch are combined into one transaction. 

Note, this creates a nested transaction which we discussed recently that we should try to avoid. It may be useful in the future to refactor all the call sites of requestDomainDnsRefresh to be in a transaction so that the transaction creation can be removed from the method.
</blockquote></details>

I think we can try cleaning it up now. If we add a `createDnsRefreshRequest(name, type, delay)` method in DnsUtil, we can easily turn this into a bulk insert.

As a site node, I don't think the timestamp in `DnsRefreshRequest` needs to be the transaction time. However, if we want to keep it this way, we can pass it as a parameter in the new method. 
 
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 112 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201090851:-NWJHEoyAIiTTPdlvIN4:bfxu0j8) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L112)):*

Are you recommending we not set any response status in the catch block?
</blockquote></details>

I think Ben meant to reply 'OK', since we are not going to do anything.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 122 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJIGB09l__PySro27D:-NWJIGB09l__PySro27E:b-j1upz7) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L122)):*
> ```Java
>                     });
>               });
>       lastInPreviousBatch = Iterables.getLast(domainBatch);
> ```

We can statically import this method, since its meaning is obvious and the class name does not give much additional information.

 



_Code quote:_
```Java
getLast
```

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 4 files reviewed, 11 unresolved discussions (waiting on @github-code-scanning[bot], @sarahcaseybot, and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 71 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJKaEV55Wpk03Yfjaz:-NWJKaEV55Wpk03Yfjb-:b-84ynhh) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L71)):*
> ```Java
>   @Inject
>   @Parameter("batchSize")
>   int batchSize;
> ```

Make these `Optional<Integer>` parameters and then have reasonable defaults?  This action isn't going to be super easy to use if someone has to specify all of these every time, plus you don't even have any guidance here on what would be reasonable for people to pass in.  Use reasonable defaults.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @github-code-scanning[bot], @sarahcaseybot, and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 93 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJbzYADH50qH5cC3GJ:-NWJbzYADH50qH5cC3GK:b-i58sul) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L93)):*
> ```Java
>       String previousLastName = lastInPreviousBatch;
>       ImmutableList<String> domainBatch =
>           tm().transact(
> ```

Why are there two separate transactions per iteration of the loop instead of just one?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @github-code-scanning[bot], @sarahcaseybot, and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 114 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJg25P4pO-Ni2fO3x4:-NWJg25P4pO-Ni2fO3x5:by05rp) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L114)):*
> ```Java
>                         // Smear the task execution time over the next N minutes.
>                         requestDomainDnsRefresh(
>                             domainName, Duration.standardMinutes(random.nextInt(smearMinutes)));
> ```

Just something to ponder on -- for a large TLD like .app, with, say, a smear of 10 minutes, if enqueueing the DNS refresh tasks itself takes longer than on the order of 10 minutes, then the smear itself isn't actually accomplishing that much.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 107 at r2](https://reviewable.io/reviews/google/nomulus/2037#gh-1201075398:-NWxERnCAaViM6jDnpiR:bk5q6y4) ([raw file](https://github.com/google/nomulus/blob/6adaad8aab5e1d9ef5cee8395b4bf022ed8a01dc/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L107)):*

I think we can try cleaning it up now. If we add a `createDnsRefreshRequest(name, type, delay)` method in DnsUtil, we can easily turn this into a bulk insert.

As a site node, I don't think the timestamp in `DnsRefreshRequest` needs to be the transaction time. However, if we want to keep it this way, we can pass it as a parameter in the new method. 
 
</blockquote></details>

Addressed in go/r3pr/2044
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 71 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJKaEV55Wpk03Yfjaz:-NWsccFQB-iQ3XvIj6Rf:b-wpctnh) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L71)):*

Make these `Optional<Integer>` parameters and then have reasonable defaults?  This action isn't going to be super easy to use if someone has to specify all of these every time, plus you don't even have any guidance here on what would be reasonable for people to pass in.  Use reasonable defaults.
</blockquote></details>

Made batch size optional, but did not make smear minutes optional as there is no good default value that works well for any TLD.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 93 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJbzYADH50qH5cC3GJ:-NWsli3RC73xhiGWr959:b-urt4x8) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L93)):*

Why are there two separate transactions per iteration of the loop instead of just one?
</blockquote></details>

Was able to combine into one transaction but was unable to avoid needing the temp variable in order to use the variables in the lambda needed for the larger transaction.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 114 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJg25P4pO-Ni2fO3x4:-NWxEjE-7-qdFXbvLtUd:b-vsg5lp) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L114)):*

Just something to ponder on -- for a large TLD like .app, with, say, a smear of 10 minutes, if enqueueing the DNS refresh tasks itself takes longer than on the order of 10 minutes, then the smear itself isn't actually accomplishing that much.
</blockquote></details>

The guidance in the javadoc for this action recommends "1 minute per every 1,000 domains" which I think should avoid a case like this. 
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 122 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJIGB09l__PySro27D:-NWso9Cf9gAOpzbWLkH3:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L122)):*

We can statically import this method, since its meaning is obvious and the class name does not give much additional information.

 


</blockquote></details>

Done.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 78 at r5](https://reviewable.io/reviews/google/nomulus/2037#-NWyA4RE3SiX_t11PAvo:-NWyA4RE3SiX_t11PAvp:b5d19n7) ([raw file](https://github.com/google/nomulus/blob/fc63e1fa89e0a9da3145fb3c70fd23e4c22dea96/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L78)):*
> ```Java
>       @Parameter(PARAM_TLDS) ImmutableSet<String> tlds,
>       @Parameter("batchSize") int batchSize,
>       @Parameter("smearMinutes") int smearMinutes,
> ```

Please see my comment to ToolsServerModule for issues with this parameter.

An alternative is to calcualte the smear based on the number of names to refresh (you can run a select count(*) query first) and a configured fixed rate.



_Code quote:_
```Java
@Parameter("smearMinutes") int smearMinutes
```
___
*[`core/src/main/java/google/registry/tools/server/ToolsServerModule.java` line 82 at r5](https://reviewable.io/reviews/google/nomulus/2037#-NWy8mQW-_Ui5b3aHlgf:-NWy8mQW-_Ui5b3aHlgg:b-b6bzvl) ([raw file](https://github.com/google/nomulus/blob/fc63e1fa89e0a9da3145fb3c70fd23e4c22dea96/core/src/main/java/google/registry/tools/server/ToolsServerModule.java#L82)):*
> ```Java
>   @Parameter("smearMinutes")
>   static int provideSmearMinutes(HttpServletRequest req) {
>     return extractIntParameter(req, "smearMinutes");
> ```

You'll need to either making this an optional parameter with a default value, or change the all request senders to add this parameter.
Currently PublishDnsUpdateAction is invoked by itself (the `enqueue` method) and ReadDnsRefreshRequestsAction.

Alternatively, you can drop this parameter and calculate the smear on the fly.



_Code quote:_
```Java
extractIntParameter(req, "smearMinutes");
```

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot], @sarahcaseybot, and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 71 at r4](https://reviewable.io/reviews/google/nomulus/2037#-NWJKaEV55Wpk03Yfjaz:-NXHTKvj8W0htaw4L4xg:b-3mzw5k) ([raw file](https://github.com/google/nomulus/blob/c2c9249a9e1d7c84b89a79708e47a88c673252f2/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L71)):*

Made batch size optional, but did not make smear minutes optional as there is no good default value that works well for any TLD.
</blockquote></details>

Smear minutes already existed and wasn't required though, so how was that handled?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 85 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYPCEhX6v6CyQLLjQgf:-NYPCEhX6v6CyQLLjQgg:b-bwt1o0) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L85)):*
> ```Java
>     checkArgument(batchSize > 0, "Must specify a positive number for batch size");
>     int smearMinutes = tm().transact(this::calculateSmearMinutes);
>     ImmutableList<String> previousBatch = ImmutableList.of("");
> ```

I'm not understanding the logic here of initializing the previous batch to a list containing a single empty string?

It seems like this should just be a `String`?
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 110 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYPCVs801zzi84yBeDU:-NYPCVs801zzi84yBeDV:bi2xkto) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L110)):*
> ```Java
>   }
> 
>   private ImmutableList<String> getBatch(ImmutableList<String> previousBatch) {
> ```

Why are you passing in `previousBatch` here when all you're using it for is calculating `getLast()` on it?  Do that when processing the previous batch and just store the last element in it.  No need to keep the entire last batch around.
___
*[`core/src/main/java/google/registry/tools/server/ToolsServerModule.java` line 81 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYPBoMo2hXHPnkEZTWe:-NYPBoMo2hXHPnkEZTWf:bvoibew) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/ToolsServerModule.java#L81)):*
> ```Java
>   @Provides
>   @Parameter("batchSize")
>   static int provideBatchSize(HttpServletRequest req) {
> ```

This isn't the right way to handle batch size in a way that is reusable for different actions.  (And the generic naming implies that it is.)  The default of 250 may not be suitable for other actions and thus should be applied within the refresh DNS action itself.  Also it should not be a bare number, but rather, a named constant value clearly explaining what it is.  So this provide should return an `Optional<Integer>`

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 110 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYP8L6H8oNNpBeofPvo:-NYP8L6H8oNNpBeofPvp:b-ru74m7) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L110)):*
> ```Java
>   }
> 
>   private ImmutableList<String> getBatch(ImmutableList<String> previousBatch) {
> ```

This parameter shouldn't be a list anymore. 



_Code quote:_
```Java
ImmutableList<String> previousBatch
```

weiminyu (approved):
LGTM now, but please address comment about parameter of get_batch().

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037#-:-NYPGsZG74h2VUIoGovV:b-uec346)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @sarahcaseybot)

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2037)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @github-code-scanning[bot], and @weiminyu)

___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 85 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYPCEhX6v6CyQLLjQgf:-NYPKN-96WNg897QXhiF:b-imhmkm) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L85)):*

I'm not understanding the logic here of initializing the previous batch to a list containing a single empty string?

It seems like this should just be a `String`?
</blockquote></details>

Changed refreshBatch() to only take in a String but this variable still needs to be a list instead of just a string since it is needed to calculate String lastInPreviousBatch in the Do-While loop. The lastInPreviousBatch variable cannot be initialized before the do-while loop and updated in each iteration after previousBatch is updated since Java requires an effectively final variable be used in the lambda expression.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 110 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYP8L6H8oNNpBeofPvo:-NYPIwBMA5zxZLTB88MC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L110)):*

This parameter shouldn't be a list anymore. 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java` line 110 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYPCVs801zzi84yBeDU:-NYPKFoa-H571InyyxW9:b-1d2l7) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/RefreshDnsForAllDomainsAction.java#L110)):*

Why are you passing in `previousBatch` here when all you're using it for is calculating `getLast()` on it?  Do that when processing the previous batch and just store the last element in it.  No need to keep the entire last batch around.
</blockquote></details>

Changed to only pass in String lastInPreviousBatch
___
*[`core/src/main/java/google/registry/tools/server/ToolsServerModule.java` line 81 at r6](https://reviewable.io/reviews/google/nomulus/2037#-NYPBoMo2hXHPnkEZTWe:-NYPMY8s8qlrrw856byc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ab05eadce923dca619ce6b4cb752fc22e69f42f0/core/src/main/java/google/registry/tools/server/ToolsServerModule.java#L81)):*

This isn't the right way to handle batch size in a way that is reusable for different actions.  (And the generic naming implies that it is.)  The default of 250 may not be suitable for other actions and thus should be applied within the refresh DNS action itself.  Also it should not be a bare number, but rather, a named constant value clearly explaining what it is.  So this provide should return an `Optional<Integer>`
</blockquote></details>

Done.

PR #2043, opened 2023-05-25
Find the most recent prefix for RdeReportAction

jianglai:
When RdeReportAction is invoked without a prefix parameter (as in the
case when it is kicked off by cron jobs for potential catch ups), we
need to used the same heuristics that's employed in RdeUploadAction to
find the most recent prefix for the given watermark, otherwise the job
will not find any deposits to upload.

Also renamed RdeUtil to RdeUtils, to be consistent with our naming
conventions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2043)

### DIFF
--- a/core/src/main/java/google/registry/beam/rde/RdeIO.java
+++ b/core/src/main/java/google/registry/beam/rde/RdeIO.java
@@ -43,7 +43,7 @@ import google.registry.rde.RdeMarshaller;
 import google.registry.rde.RdeModule;
 import google.registry.rde.RdeResourceType;
 import google.registry.rde.RdeUploadAction;
-import google.registry.rde.RdeUtil;
+import google.registry.rde.RdeUtils;
 import google.registry.request.Action.Service;
 import google.registry.request.RequestParameters;
 import google.registry.tldconfig.idn.IdnTableEnum;
@@ -166,7 +166,7 @@ public class RdeIO {
       final int revision =
           Optional.ofNullable(key.revision())
               .orElseGet(() -> RdeRevision.getNextRevision(tld, watermark, mode));
-      String id = RdeUtil.timestampToId(watermark);
+      String id = RdeUtils.timestampToId(watermark);
       String prefix =
           options.getJobName()
               + '/'
--- a/core/src/main/java/google/registry/rde/ContactToXjcConverter.java
+++ b/core/src/main/java/google/registry/rde/ContactToXjcConverter.java
@@ -112,8 +112,8 @@ final class ContactToXjcConverter {
   private static XjcRdeContactTransferDataType convertTransferData(TransferData model) {
     XjcRdeContactTransferDataType bean = new XjcRdeContactTransferDataType();
     bean.setTrStatus(XjcEppcomTrStatusType.fromValue(model.getTransferStatus().getXmlName()));
-    bean.setReRr(RdeUtil.makeXjcRdeRrType(model.getGainingRegistrarId()));
-    bean.setAcRr(RdeUtil.makeXjcRdeRrType(model.getLosingRegistrarId()));
+    bean.setReRr(RdeUtils.makeXjcRdeRrType(model.getGainingRegistrarId()));
+    bean.setAcRr(RdeUtils.makeXjcRdeRrType(model.getLosingRegistrarId()));
     bean.setReDate(model.getTransferRequestTime());
     bean.setAcDate(model.getPendingTransferExpirationTime());
     return bean;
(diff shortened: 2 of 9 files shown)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2043)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2043)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/rde/RdeUploadAction.java` line 180 at r1](https://reviewable.io/reviews/google/nomulus/2043#-NWJ8_SdEsR39TNAyBEd:-NWJ8_SdEsR39TNAyBEe:b7srbdd) ([raw file](https://github.com/google/nomulus/blob/00a83067f0de2e7226fc6e79af7b70af8ea444c5/core/src/main/java/google/registry/rde/RdeUploadAction.java#L180)):*
> ```Java
>     final String nameWithoutPrefix =
>         RdeNamingUtils.makeRydeFilename(tld, watermark, FULL, 1, revision);
>     final String name = prefix.get() + nameWithoutPrefix;
> ```

seems a little odd to use get() here -- I know we are guaranteed that the Optional will hold a value, maybe we even skip the check above and do something like 

`final String name = prefix.orElseGet(() -> findMostRecentPrefixForWatermark(...)) + nameWithoutPrefix`? Or is that too long for one statement? 

Same comment for RdeReportAction I guess, in that we don't need to handle Optionals there either

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2043)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/rde/RdeUploadAction.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/2043#-NWJCzrwEDiDIgMbLh8i:-NWJCzrwEDiDIgMbLh8j:bcypnnb) ([raw file](https://github.com/google/nomulus/blob/00a83067f0de2e7226fc6e79af7b70af8ea444c5/core/src/main/java/google/registry/rde/RdeUploadAction.java#L131)):*
> ```Java
>     HashMultimap<String, String> params = HashMultimap.create();
>     params.put(RequestParameters.PARAM_TLD, tld);
>     prefix.ifPresent(s -> params.put(RdeModule.PARAM_PREFIX, s));
> ```

Here is where the (potentially modified) `prefix` variable is used.
___
*[`core/src/main/java/google/registry/rde/RdeUploadAction.java` line 180 at r1](https://reviewable.io/reviews/google/nomulus/2043#-NWJ8_SdEsR39TNAyBEd:-NWJC0rb9XZk0y60c2Cj:bphc3bp) ([raw file](https://github.com/google/nomulus/blob/00a83067f0de2e7226fc6e79af7b70af8ea444c5/core/src/main/java/google/registry/rde/RdeUploadAction.java#L180)):*

seems a little odd to use get() here -- I know we are guaranteed that the Optional will hold a value, maybe we even skip the check above and do something like 

`final String name = prefix.orElseGet(() -> findMostRecentPrefixForWatermark(...)) + nameWithoutPrefix`? Or is that too long for one statement? 

Same comment for RdeReportAction I guess, in that we don't need to handle Optionals there either 
</blockquote></details>

Using `Optional.orElseGet()` does look better here, but in this action specifically we also use the `prefix` variable in the `run()` function, after `runWithLock()` is called (see comment above). So we do want to change the variable itself so the prefix can be passed down to the next action as an URL parameter. It is not strictly necessary because we just made the next action (`RdeReportAction`) work without a prefix. However, that would just make the next action do the same work twice to find the prefix. So there's a bit of savings to be had by changing the `prefix` variable after the value has been found.

We can certainly use `Optional.orElseGet()` in `RdeReportAction`, as the variable is not used later, but I felt, for consistency, its better to call the helper function the same way in both places.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2043)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/rde/RdeUploadAction.java` line 131 at r1](https://reviewable.io/reviews/google/nomulus/2043#-NWJCzrwEDiDIgMbLh8i:-NWJFAie-izoeiZ3L6-X:b-a74kv) ([raw file](https://github.com/google/nomulus/blob/00a83067f0de2e7226fc6e79af7b70af8ea444c5/core/src/main/java/google/registry/rde/RdeUploadAction.java#L131)):*

Here is where the (potentially modified) `prefix` variable is used.
</blockquote></details>

ah right and we can't get the prefix before `runWithLock` because we use the watermark to get it, and the watermark is provided by the superclass. I was going to suggest acquiring the prefix in an injected constructor so that we only have to do it once, but we can't do that. Tricky!

PR #2044, opened 2023-05-26
Remove nested transaction from requestDnsRefresh

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2044)

### DIFF
--- a/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
@@ -262,6 +262,6 @@ public class DeleteProberDataAction implements Runnable {
     // messages, or auto-renews because those will all be hard-deleted the next time the job runs
     // anyway.
     tm().putAll(ImmutableList.of(deletedDomain, historyEntry));
-    requestDomainDnsRefresh(deletedDomain.getDomainName());
+    requestDomainDnsRefresh(deletedDomain.getDomainName(), tm().getTransactionTime());
   }
 }
--- a/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteProberDataAction.java
@@ -262,6 +262,6 @@ public class DeleteProberDataAction implements Runnable {
     // messages, or auto-renews because those will all be hard-deleted the next time the job runs
     // anyway.
     tm().putAll(ImmutableList.of(deletedDomain, historyEntry));
-    requestDomainDnsRefresh(deletedDomain.getDomainName(), tm().getTransactionTime());
+    requestDomainDnsRefresh(deletedDomain.getDomainName());
   }
 }
--- a/core/src/main/java/google/registry/dns/DnsUtils.java
+++ b/core/src/main/java/google/registry/dns/DnsUtils.java
@@ -35,17 +35,16 @@ public final class DnsUtils {
 
   private DnsUtils() {}
 
-  private static void requestDnsRefresh(
-      String name, TargetType type, Duration delay, DateTime transactionTime) {
+  private static void requestDnsRefresh(String name, TargetType type, Duration delay) {
     tm().assertInTransaction();
     // Throws an IllegalArgumentException if the name is not under a managed TLD -- we only update
     // DNS for names that are under our management.
     String tld = Tlds.findTldForNameOrThrow(InternetDomainName.from(name)).toString();
-    tm().insert(new DnsRefreshRequest(type, name, tld, transactionTime.plus(delay)));
+    tm().insert(new DnsRefreshRequest(type, name, tld, tm().getTransactionTime().plus(delay)));
   }
 
   private static void requestDnsRefresh(
-      ImmutableList<String> names, TargetType type, Duration delay, DateTime transactionTime) {
+      ImmutableList<String> names, TargetType type, Duration delay) {
     tm().assertInTransaction();
     tm().insertAll(
             names.stream()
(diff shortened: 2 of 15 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2044)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 42 at r1](https://reviewable.io/reviews/google/nomulus/2044#-NWOGwky-k-plgYdEYii:-NWOGwky-k-plgYdEYij:b-fktol7) ([raw file](https://github.com/google/nomulus/blob/6345d1c77bd5e8a17ebbe7ecab1a9694ac0dd2aa/core/src/main/java/google/registry/dns/DnsUtils.java#L42)):*
> ```Java
>     // Throws an IllegalArgumentException if the name is not under a managed TLD -- we only update
>     // DNS for names that are under our management.
>     String tld = Tlds.findTldForNameOrThrow(InternetDomainName.from(name)).toString();
> ```

Now that I think of this, rather than reusing this method in the bulk update case, it may be better to define a dedicated method:

public static void requestDnsRefresh(ImmutableCollection<String> names, type, delay) {
  String tld = ...;
  tm().transact(
      () ->
          tm().insertAll(names.stream().,map(name -> new DnsRefreshRquest(...)).collect(...));

}

We can add shorthand methods like requestDomainDnsRefresh(names, ...) etc, and start using it in the RequestRefreshForAllDomainsAction without having to change other use cases at this moment.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2044)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 42 at r1](https://reviewable.io/reviews/google/nomulus/2044#-NWOGwky-k-plgYdEYii:-NWOZp2y6rUkUXAupqrr:bqxk5u) ([raw file](https://github.com/google/nomulus/blob/6345d1c77bd5e8a17ebbe7ecab1a9694ac0dd2aa/core/src/main/java/google/registry/dns/DnsUtils.java#L42)):*

Are you suggesting not making the changes to the existing calls of requestDnsRefresh? It seemed to me like its not a good idea to leave as it is since its hiding a transaction creation when some calls to requestDnsRefresh are already in a transaction and some are not. By moving the transaction creation outside of requestDnsRefresh we can force all calls to requestDnsRefresh to already be in a transaction.

I agree that we should add a requestDnsRefresh method that takes in a collection of names, but wouldn't it still be a better idea to keep the transaction creation outside of the method?
</blockquote></details>

This is fine.  I was thinking that you don't have to do this in the context of the RefreshForAll fix. 

To improve readablity we should have a assertInTraction before the insert call.

Please add a bulk version of this method.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2044)* status: 0 of 15 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/test/java/google/registry/dns/DnsUtilsTest.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/2044#-NWObT0O9nRrr2Yw0sQf:-NWObT0O9nRrr2Yw0sQg:b-dzkh65) ([raw file](https://github.com/google/nomulus/blob/6345d1c77bd5e8a17ebbe7ecab1a9694ac0dd2aa/core/src/test/java/google/registry/dns/DnsUtilsTest.java#L65)):*
> ```Java
>             tm().transact(
>                     () ->
>                         DnsUtils.requestHostDnsRefresh(
> ```

static import.

Same below



_Code quote:_
```Java
requestHostDnsRefresh
```

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2044)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 39 at r2](https://reviewable.io/reviews/google/nomulus/2044#-NWmVrjl-1ivETGGMJib:-NWmsESJ-PrZfq_0IAfX:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9c0fc7bddcc772fb75801c472ffeba7014e80f87/core/src/main/java/google/registry/dns/DnsUtils.java#L39)):*

With tm().assertInTransaction() below, do we still need this parameter? 
I think we can just call tm().getTransactionTime() instead.

Same for the bulk request method below.
</blockquote></details>

Done.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2044)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 58 at r2](https://reviewable.io/reviews/google/nomulus/2044#-NWmWRxr50z9RdjpSpcz:-NWmzKUSDamUZHMSNHLm:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9c0fc7bddcc772fb75801c472ffeba7014e80f87/core/src/main/java/google/registry/dns/DnsUtils.java#L58)):*

It's value doesn't change in this method, since delay and transactionTime (or tm().getTransactionTime()) do not change in this method.
So we don't have to calculate afresh for each DnsRefreshRequest constructor call.

We can add one line before the insertAll: DateTime requestTime = transactionTime.plus(delay); and use requestTime in the statement.
</blockquote></details>

Done.

weiminyu (changes requested):
> Reviewed 1 of 15 files at r3.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2044)* status: 1 of 15 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/dns/DnsUtils.java` line 46 at r4](https://reviewable.io/reviews/google/nomulus/2044#-NWy3OKnBUrlIs5yXyWf:-NWy3OKnBUrlIs5yXyWg:b5sl93e) ([raw file](https://github.com/google/nomulus/blob/4d99aecdac4fd782bf28415c5ad887fb46adca96/core/src/main/java/google/registry/dns/DnsUtils.java#L46)):*
> ```Java
>   }
> 
>   private static void requestDnsRefresh(
> ```

Do we need this method? I don't see anywhere in current code base where this new method would make a difference.



_Code quote:_
```Java
requestDnsRefresh
```
___
*[`core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java` line 363 at r4](https://reviewable.io/reviews/google/nomulus/2044#-NWy1znl8Wlr2R_lCl5A:-NWy1znl8Wlr2R_lCl5B:b-d6b10g) ([raw file](https://github.com/google/nomulus/blob/4d99aecdac4fd782bf28415c5ad887fb46adca96/core/src/main/java/google/registry/dns/PublishDnsUpdatesAction.java#L363)):*
> ```Java
>     for (String host : nullToEmpty(hosts)) {
>       tm().transact(() -> requestHostDnsRefresh(host));
>     }
> ```

These two loops are starting a real transaction for each name now. We should move up the transact call. E.g.
`
    tm().transact(
        () -> nullToEmpty(domains).forEach(DnsUtils::requestDomainDnsRefresh)
    );
`



_Code quote:_
```Java
    for (String domain : nullToEmpty(domains)) {
      tm().transact(() -> requestDomainDnsRefresh(domain));
    }
    for (String host : nullToEmpty(hosts)) {
      tm().transact(() -> requestHostDnsRefresh(host));
    }
```

PR #2045, opened 2023-06-02
Remove slash from console contacts endpoint

gbrodman:
Endpoints shouldn't themselves end in slashes

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2045)


ptkach (commented):
I don't think always removing a trailing slash is written in stone. As with UNIX directories analogy and according to Google webmasters guidelines - https://developers.google.com/search/blog/2010/04/to-slash-or-not-to-slash it should be used to better identify requested resource, which in this case feels more like a folder, rather than a file. 

I understand that it's not inline with the rest of resources, so if you strongly believe it trailing slash should be removed I'll legit

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2045#-:-NWxFwyhBb17PSVdNFHS:b-inwl96)* status: 0 of 3 files reviewed, all discussions resolved

ptkach (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2045#-:-NWxISuN0GRpHUoh2Rsb:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

PR #2049, opened 2023-06-07
Refactor OIDC-based auth mechanism

jianglai:
This PR changes the two flavors of OIDC authentication mechanisms to
verify the same audience. This allows the same token to pass both
mechanisms. Previously the regular OIDC flavor uses the project id as
its required audience, which does not work for local user credentials
(such as ones used by the nomulus tool), which requires a valid OAuth
client ID as audience when minting the token (project id is NOT a valid
OAuth client ID).

I considered allowing multiple audiences, but the result is not as clean
as just using the same everywhere, because the fall-through logic would
have generated a lot of noises for failed attempts.

This PR also changes the client side to solely use OIDC token whenever
possible, including the proxy, cloud scheduler and cloud tasks. The nomulus
tool still uses OAuth access token by default because it requires USER level
authentication, which in turn requires us to fill the User table with objects
corresponding to the email address of everyone needing access to the tool.

TESTED=verified each client is able to make authenticated calls on QA with or
without IAP.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2049)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -564,14 +564,18 @@ task deployCloudSchedulerAndQueue {
     def env = environment
     if (!prodOrSandboxEnv) {
       exec {
+        workingDir "${rootDir}/release/builder/"
         commandLine 'go', 'run',
-                "${rootDir}/release/builder/deployCloudSchedulerAndQueue.go",
+                "./deployCloudSchedulerAndQueue.go",
+                "${rootDir}/core/src/main/java/google/registry/config/files/nomulus-config-${env}.yaml",
                 "${rootDir}/core/src/main/java/google/registry/env/${env}/default/WEB-INF/cloud-scheduler-tasks.xml",
                 "domain-registry-${env}"
       }
       exec {
+        workingDir "${rootDir}/release/builder/"
         commandLine 'go', 'run',
-                "${rootDir}/release/builder/deployCloudSchedulerAndQueue.go",
+                "./deployCloudSchedulerAndQueue.go",
+                "${rootDir}/core/src/main/java/google/registry/config/files/nomulus-config-${env}.yaml",
                 "${rootDir}/core/src/main/java/google/registry/env/common/default/WEB-INF/cloud-tasks-queue.xml",
                 "domain-registry-${env}"
       }
--- a/core/src/main/java/google/registry/batch/CloudTasksUtils.java
+++ b/core/src/main/java/google/registry/batch/CloudTasksUtils.java
@@ -20,8 +20,6 @@ import static google.registry.tools.ServiceConnection.getServer;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 import com.google.api.gax.rpc.ApiException;
-import com.google.cloud.tasks.v2.AppEngineHttpRequest;
-import com.google.cloud.tasks.v2.AppEngineRouting;
 import com.google.cloud.tasks.v2.CloudTasksClient;
 import com.google.cloud.tasks.v2.HttpMethod;
 import com.google.cloud.tasks.v2.HttpRequest;
@@ -39,10 +37,12 @@ import com.google.common.net.MediaType;
 import com.google.common.net.UrlEscapers;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.util.Timestamps;
+import google.registry.config.CredentialModule.ApplicationDefaultCredential;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.request.Action.Service;
 import google.registry.util.Clock;
 import google.registry.util.CollectionUtils;
+import google.registry.util.GoogleCredentialsBundle;
 import google.registry.util.Retrier;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
@@ -52,7 +52,6 @@ import java.util.Random;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import javax.annotation.Nullable;
 import javax.inject.Inject;
 import org.joda.time.Duration;
 
@@ -67,9 +66,8 @@ public class CloudTasksUtils implements Serializable {
   private final Clock clock;
   private final String projectId;
   private final String locationId;
-  // defaultServiceAccount and iapClientId are nullable because Optional isn't serializable
-  @Nullable private final String defaultServiceAccount;
-  @Nullable private final String iapClientId;
+  private final String oauthClientId;
+  private final GoogleCredentialsBundle credential;
   private final SerializableCloudTasksClient client;
 
   @Inject
@@ -78,15 +76,15 @@ public class CloudTasksUtils implements Serializable {
       Clock clock,
       @Config("projectId") String projectId,
       @Config("locationId") String locationId,
-      @Config("defaultServiceAccount") Optional<String> defaultServiceAccount,
-      @Config("iapClientId") Optional<String> iapClientId,
+      @Config("oauthClientId") String oauthClientId,
+      @ApplicationDefaultCredential GoogleCredentialsBundle credential,
       SerializableCloudTasksClient client) {
     this.retrier = retrier;
     this.clock = clock;
     this.projectId = projectId;
     this.locationId = locationId;
-    this.defaultServiceAccount = defaultServiceAccount.orElse(null);
-    this.iapClientId = iapClientId.orElse(null);
+    this.oauthClientId = oauthClientId;
+    this.credential = credential;
     this.client = client;
   }
 
@@ -124,7 +122,7 @@ public class CloudTasksUtils implements Serializable {
    *
    * @return the resulting path (unchanged for POST requests, with params added for GET requests)
    */
-  private String processRequestParameters(
+  private static String processRequestParameters(
       String path,
       HttpMethod method,
       Multimap<String, String> params,
@@ -153,34 +151,11 @@ public class CloudTasksUtils implements Serializable {
   }
 
   /**
-   * Creates a {@link Task} that does not use AppEngine for submission.
-   *
-   * <p>This uses the standard Cloud Tasks auth format to create and send an OIDC ID token set to
-   * the default service account. That account must have permission to submit tasks to Cloud Tasks.
-   */
-  private Task createNonAppEngineTask(
-      String path, HttpMethod method, Service service, Multimap<String, String> params) {
-    HttpRequest.Builder requestBuilder = HttpRequest.newBuilder().setHttpMethod(method);
-    path =
-        processRequestParameters(
-            path, method, params, requestBuilder::putHeaders, requestBuilder::setBody);
-    OidcToken.Builder oidcTokenBuilder =
-        OidcToken.newBuilder().setServiceAccountEmail(defaultServiceAccount);
-    // If the service is using IAP, add that as the audience for the token so the request can be
-    // appropriately authed. Otherwise, use the project name.
-    if (iapClientId != null) {
-      oidcTokenBuilder.setAudience(iapClientId);
-    } else {
-      oidcTokenBuilder.setAudience(projectId);
-    }
-    requestBuilder.setOidcToken(oidcTokenBuilder.build());
-    String totalPath = String.format("%s%s", getServer(service), path);
-    requestBuilder.setUrl(totalPath);
-    return Task.newBuilder().setHttpRequest(requestBuilder.build()).build();
-  }
-
-  /**
    * Create a {@link Task} to be enqueued.
+   *
+   * <p>This uses the standard Cloud Tasks auth format to create and send an OIDC ID token with the
+   * default service account as the principal. That account must have permission to submit tasks to
+   * Cloud Tasks.
    *
    * @param path the relative URI (staring with a slash and ending without one).
    * @param method the HTTP method to be used for the request, only GET and POST are supported.
@@ -188,7 +163,7 @@ public class CloudTasksUtils implements Serializable {
    *     Queue API if no service is specified, the service which enqueues the task will be used to
    *     process the task. Cloud Tasks API does not support this feature so the service will always
    *     needs to be explicitly specified.
-   * @param params a multi-map of URL query parameters. Duplicate keys are saved as is, and it is up
+   * @param params a multimap of URL query parameters. Duplicate keys are saved as is, and it is up
    *     to the server to process the duplicate keys.
    * @return the enqueued task.
    * @see <a
@@ -204,21 +179,18 @@ public class CloudTasksUtils implements Serializable {
         method.equals(HttpMethod.GET) || method.equals(HttpMethod.POST),
         "HTTP method %s is used. Only GET and POST are allowed.",
         method);
-    // If the default service account is configured, send a standard non-AppEngine HTTP request
-    if (defaultServiceAccount != null) {
-      return createNonAppEngineTask(path, method, service, params);
-    } else {
-      AppEngineHttpRequest.Builder requestBuilder =
-          AppEngineHttpRequest.newBuilder()
-              .setHttpMethod(method)
-              .setAppEngineRouting(
-                  AppEngineRouting.newBuilder().setService(service.toString()).build());
-      path =
-          processRequestParameters(
-              path, method, params, requestBuilder::putHeaders, requestBuilder::setBody);
-      requestBuilder.setRelativeUri(path);
-      return Task.newBuilder().setAppEngineHttpRequest(requestBuilder.build()).build();
-    }
+    HttpRequest.Builder requestBuilder = HttpRequest.newBuilder().setHttpMethod(method);
+    path =
+        processRequestParameters(
+            path, method, params, requestBuilder::putHeaders, requestBuilder::setBody);
+    OidcToken.Builder oidcTokenBuilder =
+        OidcToken.newBuilder()
+            .setServiceAccountEmail(credential.serviceAccount())
+            .setAudience(oauthClientId);
+    requestBuilder.setOidcToken(oidcTokenBuilder.build());
+    String totalPath = String.format("%s%s", getServer(service), path);
+    requestBuilder.setUrl(totalPath);
+    return Task.newBuilder().setHttpRequest(requestBuilder.build()).build();
   }
 
   /**
@@ -230,7 +202,7 @@ public class CloudTasksUtils implements Serializable {
    *     Queue API if no service is specified, the service which enqueues the task will be used to
    *     process the task. Cloud Tasks API does not support this feature so the service will always
    *     needs to be explicitly specified.
-   * @param params a multi-map of URL query parameters. Duplicate keys are saved as is, and it is up
+   * @param params a multimap of URL query parameters. Duplicate keys are saved as is, and it is up
    *     to the server to process the duplicate keys.
    * @param jitterSeconds the number of seconds that a task is randomly delayed up to.
    * @return the enqueued task.
@@ -264,7 +236,7 @@ public class CloudTasksUtils implements Serializable {
    *     Queue API if no service is specified, the service which enqueues the task will be used to
    *     process the task. Cloud Tasks API does not support this feature so the service will always
    *     needs to be explicitly specified.
-   * @param params a multi-map of URL query parameters. Duplicate keys are saved as is, and it is up
+   * @param params a multimap of URL query parameters. Duplicate keys are saved as is, and it is up
    *     to the server to process the duplicate keys.
    * @param delay the amount of time that a task needs to delayed for.
    * @return the enqueued task.
@@ -330,6 +302,9 @@ public class CloudTasksUtils implements Serializable {
   }
 
   public abstract static class SerializableCloudTasksClient implements Serializable {
+
+    private static final long serialVersionUID = 7872861868968535498L;
+
     public abstract Task enqueue(String projectId, String locationId, String queueName, Task task);
   }
 
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -102,6 +102,8 @@ public final class RegistryConfig {
   @Module
   public static final class ConfigModule {
 
+    private ConfigModule() {}
+
     @Provides
     @Config("projectId")
     public static String provideProjectId(RegistryConfigSettings config) {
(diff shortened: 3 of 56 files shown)


jianglai (commented):
> Reviewed 21 of 35 files at r1, 5 of 5 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 3 of 4 files at r5, 3 of 3 files at r6, 2 of 2 files at r7, 1 of 1 files at r8.
> Dismissed @github-code-scanning[bot] from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: 33 of 36 files reviewed, 1 unresolved discussion (waiting on @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 86 at r1](https://reviewable.io/reviews/google/nomulus/2049#gh-1221650325:-NXLPfAB1-sOwoO6Qk2h:bwxkk1v) ([raw file](https://github.com/google/nomulus/blob/578663a7405497b7bd0d01a2f7350602c30438cc/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L86)):*

## Insertion of sensitive information into log files

This [potentially sensitive information](1) is written to a log file.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1502)
</blockquote></details>

We need to see the token in case of authentication failure.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 86 at r4](https://reviewable.io/reviews/google/nomulus/2049#gh-1221799400:-NXLajj-BuwMCP4wJhcI:b-896fix) ([raw file](https://github.com/google/nomulus/blob/3f30d17c5541c45c53ebe0387d200917a43cccba/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L86)):*

## Insertion of sensitive information into log files

This [potentially sensitive information](1) is written to a log file.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1503)
</blockquote></details>

Done.
___
*[`util/src/main/java/google/registry/util/OidcTokenUtils.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/2049#gh-1221650299:-NXLdIUq3kF1Mk4vjTmS:b-848932) ([raw file](https://github.com/google/nomulus/blob/578663a7405497b7bd0d01a2f7350602c30438cc/util/src/main/java/google/registry/util/OidcTokenUtils.java#L63)):*

## Dereferenced variable may be null

Variable [idToken](1) may be null at this access because of [this](2) assignment.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1494)
</blockquote></details>

Moot.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: 33 of 36 files reviewed, all discussions resolved (waiting on @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 46 at r1](https://reviewable.io/reviews/google/nomulus/2049#gh-1221650305:-NXLdW_a24kOWbN_wpBq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/578663a7405497b7bd0d01a2f7350602c30438cc/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L46)):*

## Useless parameter

The parameter 'projectId' is never used.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1501)
</blockquote></details>

Done.

jianglai (commented):
> Reviewed 2 of 35 files at r1, 1 of 1 files at r9, 21 of 21 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
> Reviewed 7 of 7 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: all files reviewed, 12 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 67 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRMNlv6vwRGfeG1loN:-NXkD7Ny7iA9bk747-aC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L67)):*

can we just use the name `oauthClientId` everywhere so that it's clear what type of client ID it is and what it's used for? 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 181 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRLwOv2K2Qlz32ctGy:-NXkDBLf8THXmW9A0LoQ:b-34lz9t) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L181)):*

this means that switching to GKE will require an actual code change (rather than just a config change) right? this is just the App Engine default service account
</blockquote></details>

Correct. On GKE we should be able to use `OidcTokenUtils` to get the OIDC token directly from the default credential, instead of having to rely on the `OidcToken.Builder` class which requires the service account explicitly. On GAE the ADC is an [`AppIdentityCredential`](https://github.com/googleapis/google-api-java-client/blob/main/google-api-client-appengine/src/main/java/com/google/api/client/googleapis/extensions/appengine/auth/oauth2/AppIdentityCredential.java) that does not implement `IdTokenProvider`.

I think after we get the token from `OidcTokenUtils`, we can use [`OidcToken.parseFrom`](https://cloud.google.com/java/docs/reference/google-cloud-tasks/2.13.0/com.google.cloud.tasks.v2.OidcToken#com_google_cloud_tasks_v2_OidcToken_parseFrom_byte___) to inject the token into the request. It will need to be tested. If it doesn't work, we can always explicitly provide the service account and still use `OidcToken.Builder`, though the former option is more desirable.
___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 183 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRMqjJD5GNIarsrXxK:-NXkG0y5DGfgfWowGS2L:badrdmb) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L183)):*

the nittiest comment, but we can just include this line in the previous statement
</blockquote></details>

Done. Builder pattern FTW.
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 294 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRce7U3YZnY8zcv4nr:-NXkG6NkDq9EseebSTPZ:b-coqw4w) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/config/files/default-config.yaml#L294)):*

This is going to force changes in all of the config files right? 
</blockquote></details>

Right, I'll have a corresponding change to the internal config files that go with this PR.
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 295 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRNARf6Kn0Wx0QnM_X:-NXkITbqDC49L4FjxkR7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/config/files/default-config.yaml#L295)):*

can we have a quick note explaining the replacement for the deprecated fields? 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 326 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRbFTe-flcHP3-3AJ_:-NXkIU7D9PALKoeqmUsy:b-1ythyy) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/config/files/default-config.yaml#L326)):*

what will be the ID in the case where we don't have IAP? Are you thinking that we just "create" it in every environment and use that client ID even when it's not enabled? 
</blockquote></details>

We just use the IAP client ID. IAP does not have to be enabled for the project. In principle we can use any client ID, but using the IAP client ID has the added benefit that there would be no change at the client side when we turn on/off IAP, the same token will be accepted with or without IAP.
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 86 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRc84XBqnMnmry6HbE:-NXkIpXq9KOke3aVVGC0:b-5njwzn) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L86)):*

hmmmm, is it sketchy to log the raw token itself? 
</blockquote></details>

I felt that if the token doesn't work, we might as well log it. Changed it to only log on non-production environment.
___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 45 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRcpYT9YQTTxcXgl11:-NXkJRwk7twO82Hj_Ri5:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L45)):*

similar to another comment can we still call it `oauthClientId` for clarity? 
</blockquote></details>

Done.
___
*[`proxy/src/main/java/google/registry/proxy/ProxyModule.java` line 263 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRg7-82PBisN6GqA5y:-NXkJ_xn9N8zi41mbBvA:b-rvpcgh) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/proxy/src/main/java/google/registry/proxy/ProxyModule.java#L263)):*

Does the ID token creation process refresh the credentials bundle automatically? 
</blockquote></details>

I believe refresh tokens and access tokens are only relevant for OAuth. With OIDC tokens they are always [fetched from the metadata server when requested](https://github.com/googleapis/google-auth-library-java/blob/5fa7039482298cc1c0ea09889db47081c83d990b/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L248).
___
*[`proxy/src/main/java/google/registry/proxy/config/default-config.yaml` line 11 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRgmYFE_1K6Vf5kVnn:-NXkK74Y94gAi147N8Om:b-77n76r) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/proxy/src/main/java/google/registry/proxy/config/default-config.yaml#L11)):*

This says "usually", is there any situation where it wouldn't be? 
</blockquote></details>

It needs to be the same as the value of `auth.oauthClientId` in the nomulus config file. Because that value should usually be the IAP client ID (unless we want to use different client IDs for IAP on/off). Updated the comment to make it clearer.
___
*[`util/src/main/java/google/registry/util/OidcTokenUtils.java` line 35 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRdfDnEVSFg47U5oJS:-NXkiDsXE7Oy2bh0_1ju:b-oopbyb) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/util/src/main/java/google/registry/util/OidcTokenUtils.java#L35)):*

out of curiosity, is there any particular reason to make this class abstract? 
</blockquote></details>

I guess not. Final with a private constructor should be good enough.
___
*[`util/src/main/java/google/registry/util/OidcTokenUtils.java` line 51 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRdjxzBXjgVnlkPfRH:-NXkiOMFDWNLS437wVxH:b7bkq53) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/util/src/main/java/google/registry/util/OidcTokenUtils.java#L51)):*

nitty but technically we don't need the `else` block 
</blockquote></details>

yes. But I find it improves readability when you have a corresponding `else` block, especially when there's not a lot of indentation involved.

gbrodman (changes requested):
> Reviewed 6 of 35 files at r1, 7 of 7 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 181 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRLwOv2K2Qlz32ctGy:-NXks--1-hqamdJzN9vd:b6zmino) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L181)):*

Correct. On GKE we should be able to use `OidcTokenUtils` to get the OIDC token directly from the default credential, instead of having to rely on the `OidcToken.Builder` class which requires the service account explicitly. On GAE the ADC is an [`AppIdentityCredential`](https://github.com/googleapis/google-api-java-client/blob/main/google-api-client-appengine/src/main/java/com/google/api/client/googleapis/extensions/appengine/auth/oauth2/AppIdentityCredential.java) that does not implement `IdTokenProvider`.

I think after we get the token from `OidcTokenUtils`, we can use [`OidcToken.parseFrom`](https://cloud.google.com/java/docs/reference/google-cloud-tasks/2.13.0/com.google.cloud.tasks.v2.OidcToken#com_google_cloud_tasks_v2_OidcToken_parseFrom_byte___) to inject the token into the request. It will need to be tested. If it doesn't work, we can always explicitly provide the service account and still use `OidcToken.Builder`, though the former option is more desirable.
</blockquote></details>

Right yeah that makes sense. 

I suppose the route I was going down was "it seems a bit dangerous to require a code / behavioral change (as opposed to only a config change) at the same time that we switch to GKE". Am I missing something? 
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 294 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRce7U3YZnY8zcv4nr:-NXksgXpALoqU_Zb9e11:bo2mp2t) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/config/files/default-config.yaml#L294)):*

Right, I'll have a corresponding change to the internal config files that go with this PR.
</blockquote></details>

got it, i've seen it now 
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 304 at r12](https://reviewable.io/reviews/google/nomulus/2049#-NXkspZfCEHEoNo38z8x:-NXkspZfCEHEoNo38z8y:b-elm8p0) ([raw file](https://github.com/google/nomulus/blob/0436ecd6f6d1cad23a30529e6029fb977768dfd2/core/src/main/java/google/registry/config/files/default-config.yaml#L304)):*
> ```YAML
> 
>   # Deprecated: Use OIDC-based auth instead. This field is for OAuth-based auth
>   # that should only be used by the old registrar console.
> ```

I don't think these comments are correct, since OAuth was used primarily by the Nomulus tool and the proxy 

also this is a bit nitty but can we refer them specifically to the `oauthClientId` and `allowedServiceAccountEmails` fields in the comment? 
___
*[`core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java` line 90 at r12](https://reviewable.io/reviews/google/nomulus/2049#gh-1227057640:-NXktX864UCxkjhwDfld:bjeuwhg) ([raw file](https://github.com/google/nomulus/blob/0436ecd6f6d1cad23a30529e6029fb977768dfd2/core/src/main/java/google/registry/request/auth/OidcTokenAuthenticationMechanism.java#L90)):*

## Insertion of sensitive information into log files

This [potentially sensitive information](1) is written to a log file.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1509)
</blockquote></details>

sigh, i love that this was only flagged AFTER you removed the prod data 
___
*[`proxy/src/main/java/google/registry/proxy/config/default-config.yaml` line 15 at r12](https://reviewable.io/reviews/google/nomulus/2049#-NXlE-tj0dviYSCoShmw:-NXlE-tj0dviYSCoShmx:bvrtzsr) ([raw file](https://github.com/google/nomulus/blob/0436ecd6f6d1cad23a30529e6029fb977768dfd2/proxy/src/main/java/google/registry/proxy/config/default-config.yaml#L15)):*
> ```YAML
> # the IAP client ID, to allow the request to access IAP protected endpoints.
> # Regular OIDC authentication mechanism also checks for this audience.
> clientId: iap-client-id
> ```

just a thought looking at the other PR, thoughts on changing this to `oauthClientId` to match the variable names (since the values have to be the same)?

jianglai (commented):
> Reviewed 7 of 8 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: 56 of 57 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 181 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRLwOv2K2Qlz32ctGy:-NYTKNshDVMIxq0mDG6C:b-mh8j5) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L181)):*

Right yeah that makes sense. 

I suppose the route I was going down was "it seems a bit dangerous to require a code / behavioral change (as opposed to only a config change) at the same time that we switch to GKE". Am I missing something? 
</blockquote></details>

We could add some logic to check which runtime we are in. But I think I have something better. Both `AppEngineCredentials` (not `AppIdentiyCredential`, which is a `GoogleCredential`, not `GoogleCredentials`...) and `ComputeEngineCrednetials` implement `ServiceAccountSigner`, which exposes the service account email address, so we can just grab that from the credential, instead having to construct our own. This should allow a seamlessly without code change.
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 304 at r12](https://reviewable.io/reviews/google/nomulus/2049#-NXkspZfCEHEoNo38z8x:-NYTKtQC4yofSsm0wExB:b-f0y2yx) ([raw file](https://github.com/google/nomulus/blob/6f6e7bea5162ad1093d636000fbb7782dda51b35/core/src/main/java/google/registry/config/files/default-config.yaml#L304)):*

I don't think these comments are correct, since OAuth was used primarily by the Nomulus tool and the proxy 

also this is a bit nitty but can we refer them specifically to the `oauthClientId` and `allowedServiceAccountEmails` fields in the comment? 
</blockquote></details>

Right now OAuth is used by the tool and the proxy, but as of this PR the proxy would start using OIDC exclusively (after a deployment), and we should plan to add the user entities to allow the tool to use OIDC as well. Going forward, OAuth should only be used by the legacy console.

Can you clarify what is "them" in "refer them specifically to the `oauthClientId` and `allowedServiceAccountEmails` fields in the comment"?
___
*[`proxy/src/main/java/google/registry/proxy/config/default-config.yaml` line 15 at r12](https://reviewable.io/reviews/google/nomulus/2049#-NXlE-tj0dviYSCoShmw:-NYTRcBS-6Di2qUj2M5B:b-896fix) ([raw file](https://github.com/google/nomulus/blob/6f6e7bea5162ad1093d636000fbb7782dda51b35/proxy/src/main/java/google/registry/proxy/config/default-config.yaml#L15)):*

just a thought looking at the other PR, thoughts on changing this to `oauthClientId` to match the variable names (since the values have to be the same)? 
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 8 of 8 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/batch/CloudTasksUtils.java` line 181 at r11](https://reviewable.io/reviews/google/nomulus/2049#-NXRLwOv2K2Qlz32ctGy:-NYZV7pHA_E6vxykMNW2:b-qn5lbb) ([raw file](https://github.com/google/nomulus/blob/4d9b8182b00c8fe54af4cf9eaa7d7e0b7a125118/core/src/main/java/google/registry/batch/CloudTasksUtils.java#L181)):*

We could add some logic to check which runtime we are in. But I think I have something better. Both `AppEngineCredentials` (not `AppIdentiyCredential`, which is a `GoogleCredential`, not `GoogleCredentials`...) and `ComputeEngineCrednetials` implement `ServiceAccountSigner`, which exposes the service account email address, so we can just grab that from the credential, instead having to construct our own. This should allow a seamlessly without code change.
</blockquote></details>

Perfect!
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 304 at r12](https://reviewable.io/reviews/google/nomulus/2049#-NXkspZfCEHEoNo38z8x:-NYZVNKd2qXTkX57Huoa:blhj2gg) ([raw file](https://github.com/google/nomulus/blob/6f6e7bea5162ad1093d636000fbb7782dda51b35/core/src/main/java/google/registry/config/files/default-config.yaml#L304)):*

Right now OAuth is used by the tool and the proxy, but as of this PR the proxy would start using OIDC exclusively (after a deployment), and we should plan to add the user entities to allow the tool to use OIDC as well. Going forward, OAuth should only be used by the legacy console.

Can you clarify what is "them" in "refer them specifically to the `oauthClientId` and `allowedServiceAccountEmails` fields in the comment"?
</blockquote></details>

My point is that "should only be used by the old registrar console" is actually not true, since the old console uses the AppEngine built-in `LegacyAuthenticationMechanism`. 

```
The client_id value
  # used in registryTool.clientId field for associated tooling should be included
  # in this list.
```

this is a bit awkward and I'm not sure what it's trying to say, can we just say "the registryTool.clientId field should be included in this list"?

My second comment was related to directing the user to the `allowedServiceAccountEmails` and `oauthClientId` fields, since those are the replacement for the deprecated fields.

jianglai (commented):
> Reviewed 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2049)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @ptkach)

PR #2052, opened 2023-06-08
Remove contacts with empty type from console GET /contacts response

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2052)

### DIFF
--- a/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java
@@ -92,9 +92,12 @@ public class ContactAction implements JsonGetAction {
     ImmutableList<RegistrarPoc> am =
         tm().transact(
                 () ->
-                    tm().createQueryComposer(RegistrarPoc.class)
+                    tm()
+                        .createQueryComposer(RegistrarPoc.class)
                         .where("registrarId", Comparator.EQ, registrarId)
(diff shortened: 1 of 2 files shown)


ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2052)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java` line 77 at r1](https://reviewable.io/reviews/google/nomulus/2052#-NXRvr0b5LZnYmdDWCEJ:-NXWIPwVCmWc7Mw32XTx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a902e3ad55535ddc76e453f4b9b37ae3620636b9/core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java#L77)):*

if this ever gets run in parallel this could result in some issues if we reset it in a non-static way but only have one static variable. How about having this var be non-static? 

also if we're setting it elsewhere the name probably shouldn't be in all caps 
</blockquote></details>

Done.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2052)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ptkach)

PR #2054, opened 2023-06-12
Add registrar selection functionality

ptkach:
This adds the registrar selection screen which appears as a first step after landing on UI. The implementation was demoed to MB and confirmed as it should be good. 

This also fixes tests for existing components and adds covers added features with tests. 

Mock data removed, back-end integration for /contacts and /registrars endpoints added

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2054)

### DIFF
--- a/console-webapp/dev-proxy.config.json
+++ b/console-webapp/dev-proxy.config.json
@@ -1,7 +1,7 @@
 {
-  "/registrar":
+  "/console-api":
   {
-    "target": "http://localhost:8080/registrar",
-    "secure": false
+    "target": "http://localhost:8080",
+    "secure": true
   }
 }
--- a/console-webapp/src/app/app-routing.module.ts
+++ b/console-webapp/src/app/app-routing.module.ts
@@ -22,14 +22,18 @@ import SettingsRegistrarsComponent from './settings/registrars/registrars.compon
 import SettingsWhoisComponent from './settings/whois/whois.component';
 import SettingsUsersComponent from './settings/users/users.component';
 import SettingsSecurityComponent from './settings/security/security.component';
+import { RegistrarGuard } from './registrar/registrar.guard';
+import { RegistrarComponent } from './registrar/registrar.component';
 
 const routes: Routes = [
-  { path: '', redirectTo: '/settings/contact', pathMatch: 'full' },
-  { path: 'home', component: HomeComponent },
-  { path: 'tlds', component: TldsComponent },
+  { path: '', redirectTo: '/home', pathMatch: 'full' },
+  { path: 'registrars', component: RegistrarComponent },
+  { path: 'home', component: HomeComponent, canActivate: [RegistrarGuard] },
+  { path: 'tlds', component: TldsComponent, canActivate: [RegistrarGuard] },
   {
     path: 'settings',
     component: SettingsComponent,
+    canActivate: [RegistrarGuard],
     children: [
       {
         path: '',
--- a/console-webapp/src/app/app.component.html
+++ b/console-webapp/src/app/app.component.html
@@ -12,6 +12,9 @@
         <a mat-list-item [routerLink]="'/settings'" routerLinkActive="active">
           Settings
         </a>
+        <a mat-list-item [routerLink]="'/registrars'" routerLinkActive="active">
+          Select Registrar
+        </a>
       </mat-nav-list>
     </mat-sidenav>
     <mat-sidenav-content class="console-app__content">
--- a/console-webapp/src/app/app.module.ts
+++ b/console-webapp/src/app/app.module.ts
@@ -31,6 +31,8 @@ import SettingsContactComponent, {
   ContactDetailsDialogComponent,
 } from './settings/contact/contact.component';
 import { HttpClientModule } from '@angular/common/http';
+import { RegistrarComponent } from './registrar/registrar.component';
+import { RegistrarGuard } from './registrar/registrar.guard';
 
 @NgModule({
   declarations: [
@@ -41,6 +43,7 @@ import { HttpClientModule } from '@angular/common/http';
     SettingsComponent,
     SettingsContactComponent,
     ContactDetailsDialogComponent,
+    RegistrarComponent,
   ],
   imports: [
     HttpClientModule,
@@ -50,7 +53,7 @@ import { HttpClientModule } from '@angular/common/http';
     AppRoutingModule,
     BrowserAnimationsModule,
   ],
-  providers: [BackendService],
+  providers: [BackendService, RegistrarGuard],
   bootstrap: [AppComponent],
 })
 export class AppModule {}
--- a/console-webapp/src/app/registrar/registrar.component.html
+++ b/console-webapp/src/app/registrar/registrar.component.html
@@ -1,0 +1,17 @@
+<div class="console-app__registrar">
+  <h4 class="console-app__title">
+    {{ registrarService.activeRegistrarId === "" ? "Select" : "Switch" }}
+    registrar:
+  </h4>
+  <mat-form-field>
+    <mat-label>Registrar</mat-label>
+    <mat-select [(ngModel)]="registrarService.activeRegistrarId">
+      <mat-option
+        *ngFor="let registrar of registrarService.registrars"
+        [value]="registrar"
+      >
+        {{ registrar }}
+      </mat-option>
+    </mat-select>
+  </mat-form-field>
+</div>
--- a/console-webapp/src/app/registrar/registrar.component.less
+++ b/console-webapp/src/app/registrar/registrar.component.less
@@ -1,0 +1,11 @@
+.console-app {
+  &__registrar {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    margin-top: 1rem;
+  }
+  &__title {
+    margin-right: 1rem;
+  }
+}
--- a/console-webapp/src/app/registrar/registrar.component.spec.ts
+++ b/console-webapp/src/app/registrar/registrar.component.spec.ts
@@ -14,18 +14,26 @@
 
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 
-import SecurityComponent from './security.component';
+import { RegistrarComponent } from './registrar.component';
+import { BackendService } from '../shared/services/backend.service';
+import { ActivatedRoute } from '@angular/router';
+import { HttpClientTestingModule } from '@angular/common/http/testing';
 
-describe('SecurityComponent', () => {
-  let component: SecurityComponent;
-  let fixture: ComponentFixture<SecurityComponent>;
+describe('RegistrarComponent', () => {
+  let component: RegistrarComponent;
+  let fixture: ComponentFixture<RegistrarComponent>;
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
-      declarations: [SecurityComponent],
+      declarations: [RegistrarComponent],
+      imports: [HttpClientTestingModule],
+      providers: [
+        BackendService,
+        { provide: ActivatedRoute, useValue: {} as ActivatedRoute },
+      ],
     }).compileComponents();
 
-    fixture = TestBed.createComponent(SecurityComponent);
+    fixture = TestBed.createComponent(RegistrarComponent);
     component = fixture.componentInstance;
     fixture.detectChanges();
   });
--- a/console-webapp/src/app/registrar/registrar.component.ts
+++ b/console-webapp/src/app/registrar/registrar.component.ts
@@ -1,0 +1,45 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Component } from '@angular/core';
+import { RegistrarService } from './registrar.service';
+import { ActivatedRoute, Router } from '@angular/router';
+
+@Component({
+  selector: 'app-registrar',
+  templateUrl: './registrar.component.html',
+  styleUrls: ['./registrar.component.less'],
+})
+export class RegistrarComponent {
+  private lastActiveRegistrarId: string;
+
+  constructor(
+    private route: ActivatedRoute,
+    protected registrarService: RegistrarService,
+    private router: Router
+  ) {
+    this.lastActiveRegistrarId = registrarService.activeRegistrarId;
+  }
+
+  ngDoCheck() {
+    if (
+      this.registrarService.activeRegistrarId &&
+      this.registrarService.activeRegistrarId !== this.lastActiveRegistrarId &&
+      this.route.snapshot.paramMap.get('nextUrl')
+    ) {
+      this.lastActiveRegistrarId = this.registrarService.activeRegistrarId;
+      this.router.navigate([this.route.snapshot.paramMap.get('nextUrl')]);
+    }
+  }
+}
--- a/console-webapp/src/app/registrar/registrar.guard.spec.ts
+++ b/console-webapp/src/app/registrar/registrar.guard.spec.ts
@@ -1,0 +1,64 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TestBed } from '@angular/core/testing';
+
+import { RegistrarGuard } from './registrar.guard';
+import { Router, RouterStateSnapshot } from '@angular/router';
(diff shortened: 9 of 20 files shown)


gbrodman (changes requested):
> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: all files reviewed, 14 unresolved discussions (waiting on @ptkach)

___
*[`console-webapp/src/app/app-routing.module.ts` line 29 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlZkWu90aFZijFuTz9:-NXlZkWu90aFZijFuTzA:bs2fav1) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/app-routing.module.ts#L29)):*
> ```XML
> 
> const routes: Routes = [
>   { path: '', redirectTo: '/settings/contact', pathMatch: 'full' },
> ```

should we redirect this to home? 
___
*[`console-webapp/src/app/registrar/registrar.component.less` line 1 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlamrRD0DzIGlpL2Wi:-NXlamrRD0DzIGlpL2Wj:bqc89v6) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.less#L1)):*
> ```Less
> .console-app {
> ```

out of curiosity, how do you go about setting all these / figuring them out? Do you just set it up with some base values and play around with adding / removing characteristics? 
___
*[`console-webapp/src/app/registrar/registrar.component.ts` line 25 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlb469APGUxGtSv_k2:-NXlb469APGUxGtSv_k3:bak2540) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.ts#L25)):*
> ```XML
> })
> export class RegistrarComponent {
>   private lastActiveRegistrarId: string;
> ```

I'm not sure I fully understand, what's the point of having both a lastActiveRegistrarId here and an activeRegistrarId in the registrar service? I would imagine that it might be ideal to only store that piece of data in one place. 
___
*[`console-webapp/src/app/registrar/registrar.component.ts` line 35 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlbjvtA1hSIDV2vodH:-NXlbjvtA1hSIDV2vodI:b-h799z1) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.ts#L35)):*
> ```XML
>   }
> 
>   ngDoCheck() {
> ```

Is the point of this that it's called whenever the user selects a registrar (thus modifying the state)? I'm a noob :D 
___
*[`console-webapp/src/app/registrar/registrar.component.ts` line 46 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlccHT-PvtoWvhVEr-:-NXlccHT-PvtoWvhVEr0:b-13yqvt) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.ts#L46)):*
> ```XML
>   }
> 
>   compareFunc(value: string, registrarId: string) {
> ```

what is this for? 
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 23 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlW29m53WVC6qw-Mna:-NXlW29m53WVC6qw-Mnb:bsuep33) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.guard.ts#L23)):*
> ```XML
>   providedIn: 'root',
> })
> export class RegistrarGuard {
> ```

should we implement CanActivate? It doesn't seem like it's technically required but it might prevent any errors from happening if we modify the `canActivate` method signature 
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 32 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlVlZU5AqsSdtiE39O:-NXlVlZU5AqsSdtiE39P:b-zga28b) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.guard.ts#L32)):*
> ```XML
>     state: RouterStateSnapshot
>   ):
>     | Observable<boolean | UrlTree>
> ```

this is tiny but i don't think we need the | at the very beginning here 

also, I'm looking at https://angular.io/api/router/CanActivate and I see it uses the same return type (with all the |s) but since we know we're only returning a `Promise<boolean>` is it possible to only have a boolean in the method signature? No worries if we have to use the whole thing 
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 39 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlaGLNDWMPHYhGaLsv:-NXlaGLNDWMPHYhGaLsw:b-e5ohaq) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.guard.ts#L39)):*
> ```XML
>       return true;
>     }
>     return this.router.navigate([`/registrars`, { nextUrl: state.url }]);
> ```

i guess a TODO would be to possibly skip this if the user only has access to one registrar, right? Not super necessary for MVP I suppose 
___
*[`console-webapp/src/app/registrar/registrar.service.ts` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlXRIq33D9A6VoD_EU:-NXlXRIq33D9A6VoD_EV:b-wpuw3c) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.service.ts#L22)):*
> ```XML
> })
> export class RegistrarService {
>   activeRegistrarId: string = '';
> ```

Are there guidelines on using a getter method vs just accessing the variable directly? 

Also, I don't think we ever set this variable?

gbrodman (changes requested):
> Reviewed 6 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: all files reviewed, 8 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

___
*[`console-webapp/src/app/registrar/registrar.component.less` line 1 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlamrRD0DzIGlpL2Wi:-NXq5z5I5T4oOK2263oH:b7o2myb) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.less#L1)):*

No sure I understand the question - are you asking about naming of classes or css rules? Naming is based on BEM methodology https://getbem.com/naming/ BLOCK__ELEMENT_MODIFIER, css rules and values based on what the design is 
</blockquote></details>

Yeah I was thinking about the CSS rules themselves, sort of  a "how do you get started?" sort of question
___
*[`console-webapp/src/app/registrar/registrar.component.ts` line 25 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlb469APGUxGtSv_k2:-NXqS5r58Zj3YU2ZZfdL:bv2zjir) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.ts#L25)):*

Good question! I've decided to make redirect to the page that user aimed for in the first place and for that we need to understand when user actually makes a selection, so that we can initiate a redirect. There're multiple way to achieve that:
1) As you suggest - listen and deal with the `activeRegistrarId` stored in the service. However, you see, service is not a good place to make a redirect, because a) it can and in fact will be reused in a different component, where redirect is not desirable b) service is just not the right place for this kind of logic
2) Slight modification of option 1 - instead of dealing with redirect in the service, we can store the value in the service but notify the component when the value is changed, so that the component then can redirect or do anything it pleases. That'd be a great option and I was aiming for it at first, however as it is Angular 15 requires rather a cumbersome way to simply notify of the value change in the service, which would make the entire usage of the service a lot more unpleasant. This will change in Angular 16 though wit signals - https://angular.io/guide/signals
3) Keep the `activeRegistrarId` in the service, but then with use of `ngDoCheck` compare the value with the last one stored in the component. This doesn't break single source of truth principle, because `lastActiveRegistrarId` gets synced with the value in `activeRegistrarId` and it doesn't produce any side effects in component, redirect is by definition is not in the component :) 

We will update option 3 as soon as signals are out of dev preview, for now though it's the simplest and the most straightforward way to handle this task. 
</blockquote></details>

ok so basically this is essentially polling the select component to see if the user has changed the element and if so, do the redirect.

That makes sense I think, even though as you mention signals would be a nicer way of doing it. 
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 23 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlW29m53WVC6qw-Mna:-NXq7yQz-qwLrA5FsymS:b-p5e4w8) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.guard.ts#L23)):*

Nope, the reason I've removed the check is that it requires canActivate to receive a parameter of type `ActivatedRouteSnapshot`, which we don't use and don't need. This would complicate testing as well, so I've decided to get rid of the param and `implement` type checking
</blockquote></details>

OK so I guess it can even be called as a `CanActivate` from the router even when it doesn't implement the interface or accept the exact parameters in the call? That's kind of neat. 
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 32 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlVlZU5AqsSdtiE39O:-NXq6zkw9L9LBD3yxmP0:bnazqs1) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.guard.ts#L32)):*

Removed unused. BTW we use Promise and boolean 
</blockquote></details>

right yeah I meant `Promise<boolean>` in both places in the comment above. Cool! 
___
*[`console-webapp/src/app/registrar/registrar.service.ts` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlXRIq33D9A6VoD_EU:-NXqRQ7A0mFm-kMjI5Jl:b-9teaif) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.service.ts#L22)):*

As long as there's no "extra" login on top of the value setting direct access is fine. set would be used once say you want to add logging, etc 

We do - in registrar.component -> `[(ngModel)]`it's a 2 way binding
</blockquote></details>

That kind of makes sense, but I see that we set `[ngModelOptions]="{ standalone: true }"` there and I thought that means that it disconnected it from the two-way binding? I'm sure I'm misunderstanding something, just trying to figure it all out.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: 18 of 19 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`console-webapp/src/app/registrar/registrar.component.less` line 1 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlamrRD0DzIGlpL2Wi:-NXu5_iq2ncy--oA1Hj1:b-cby6v7) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.component.less#L1)):*

Yeah I was thinking about the CSS rules themselves, sort of  a "how do you get started?" sort of question
</blockquote></details>

I usually add positioning styles first (flexbox or grid) for the whole page or screen, then move to the same for the parts of the screen adding flexbox or grid rules to specific blocks and then work on colors and fonts. 
___
*[`console-webapp/src/app/registrar/registrar.component.ts` line 42 at r2](https://reviewable.io/reviews/google/nomulus/2054#gh-1228567824:-NXu6NjkDZBr-HWZJV67:b-896fix) ([raw file](https://github.com/google/nomulus/blob/80eb92632ab03fe2ba9a5720b3cba3bbbe089437/console-webapp/src/app/registrar/registrar.component.ts#L42)):*

## Client-side cross-site scripting (experimental)

(Experimental) This may be a cross-site scripting vulnerability due to [a user-provided value](1). Identified using machine learning.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1516)
</blockquote></details>

Done.
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 23 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlW29m53WVC6qw-Mna:-NXu3lgt-xBgFutEMZNO:b2nhswv) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.guard.ts#L23)):*

OK so I guess it can even be called as a `CanActivate` from the router even when it doesn't implement the interface or accept the exact parameters in the call? That's kind of neat. 
</blockquote></details>

Yep
___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 16 at r2](https://reviewable.io/reviews/google/nomulus/2054#gh-1228567820:-NXu6OZR5_mdQ5NhGsmP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/80eb92632ab03fe2ba9a5720b3cba3bbbe089437/console-webapp/src/app/registrar/registrar.guard.ts#L16)):*

## Unused variable, import, function or class

Unused import UrlTree.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1515)
</blockquote></details>

Done.
___
*[`console-webapp/src/app/registrar/registrar.service.ts` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlXRIq33D9A6VoD_EU:-NXu4hkR3TqapMfl2bBs:bwg6hyp) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.service.ts#L22)):*

That kind of makes sense, but I see that we set `[ngModelOptions]="{ standalone: true }"` there and I thought that means that it disconnected it from the two-way binding? I'm sure I'm misunderstanding something, just trying to figure it all out. 
</blockquote></details>

`[(ngModel)]` is for binding
`[ngModelOptions]="{ standalone: true }"` is only to separate the field from formgroup. In other words to allow work with fields separately and not as a part of some form

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: 17 of 21 files reviewed, 8 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

___
*[`console-webapp/src/app/registrar/registrar.service.ts` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlXRIq33D9A6VoD_EU:-NXvF7vV5c3l2J8pRm3l:bu9l0kq) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.service.ts#L22)):*

`[(ngModel)]` is for binding
`[ngModelOptions]="{ standalone: true }"` is only to separate the field from formgroup. In other words to allow work with fields separately and not as a part of some form
</blockquote></details>

ahhh I think I see. 

It seems like the registrar component HTML file doesn't have any forms though so will we still need the `standalone: true` bit?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: 17 of 21 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`console-webapp/src/app/registrar/registrar.service.ts` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlXRIq33D9A6VoD_EU:-NXvGs1J-kkqy5XXCBOl:b-rke2u3) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.service.ts#L22)):*

ahhh I think I see. 

It seems like the registrar component HTML file doesn't have any forms though so will we still need the `standalone: true` bit? 
</blockquote></details>

Unfortunately it doesn't matter for angular, so yeah, it's still required

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: 16 of 21 files reviewed, 8 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`console-webapp/src/app/registrar/registrar.service.ts` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2054#-NXlXRIq33D9A6VoD_EU:-NXvLCqY8Gp8zsYDKZoH:bh55ze2) ([raw file](https://github.com/google/nomulus/blob/a227b744ec53f01e3dba60987a3db2752f64a57a/console-webapp/src/app/registrar/registrar.service.ts#L22)):*

Unfortunately it doesn't matter for angular, so yeah, it's still required
</blockquote></details>

After double checking it seems like as long as control is not wrapped in a form `standalone: true` is not required. Thanks for the tip!

gbrodman (approved):
> Reviewed 4 of 4 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2054)* status: all files reviewed, 7 unresolved discussions (waiting on @github-code-scanning[bot])

___
*[`console-webapp/src/app/settings/contact/contact-details.component.html` line 3 at r6](https://reviewable.io/reviews/google/nomulus/2054#-NXvbke65aCLuaS8_Rm7:-NXvbke65aCLuaS8_Rm8:b9sz23w) ([raw file](https://github.com/google/nomulus/blob/a0ea65248e2a95eb9346bcae06d5d8f4ab97a3f9/console-webapp/src/app/settings/contact/contact-details.component.html#L3)):*
> ```HTML
> <h3 mat-dialog-title>Contact details</h3>
> <div mat-dialog-content>
>   <form (ngSubmit)="saveAndClose($event)">
> ```

oh neat

PR #2057, opened 2023-06-17
Add console-api/settings/security endpoint

ptkach:
Tested on alpha, get and post work fine

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2057)

### DIFF
--- a/core/src/main/java/google/registry/model/registrar/Registrar.java
+++ b/core/src/main/java/google/registry/model/registrar/Registrar.java
@@ -50,6 +50,7 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Range;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
+import com.google.gson.annotations.Expose;
 import com.google.re2j.Pattern;
 import google.registry.model.Buildable;
 import google.registry.model.CreateAutoTimestamp;
@@ -253,7 +254,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   // Authentication.
 
   /** X.509 PEM client certificate(s) used to authenticate registrar to EPP service. */
-  String clientCertificate;
+  @Expose String clientCertificate;
 
   /** Base64 encoded SHA256 hash of {@link #clientCertificate}. */
   String clientCertificateHash;
@@ -263,13 +264,13 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    *
    * <p>This allows registrars to migrate certificates without downtime.
    */
-  String failoverClientCertificate;
+  @Expose String failoverClientCertificate;
 
   /** Base64 encoded SHA256 hash of {@link #failoverClientCertificate}. */
   String failoverClientCertificateHash;
 
   /** An allow list of netmasks (in CIDR notation) which the client is allowed to connect from. */
-  List<CidrAddressBlock> ipAddressAllowList;
+  @Expose List<CidrAddressBlock> ipAddressAllowList;
 
   /** A hashed password for EPP access. The hash is a base64 encoded SHA256 string. */
   String passwordHash;
--- a/core/src/main/java/google/registry/module/frontend/FrontendRequestComponent.java
+++ b/core/src/main/java/google/registry/module/frontend/FrontendRequestComponent.java
@@ -28,6 +28,7 @@ import google.registry.request.RequestScope;
 import google.registry.ui.server.console.ConsoleDomainGetAction;
 import google.registry.ui.server.console.RegistrarsAction;
 import google.registry.ui.server.console.settings.ContactAction;
+import google.registry.ui.server.console.settings.SecurityAction;
 import google.registry.ui.server.registrar.ConsoleOteSetupAction;
 import google.registry.ui.server.registrar.ConsoleRegistrarCreatorAction;
 import google.registry.ui.server.registrar.ConsoleUiAction;
@@ -69,6 +70,8 @@ interface FrontendRequestComponent {
   ContactAction contactAction();
 
   RegistrarsAction registrarsAction();
+
+  SecurityAction securityAction();
 
   @Subcomponent.Builder
   abstract class Builder implements RequestComponentBuilder<FrontendRequestComponent> {
--- a/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java
@@ -1,0 +1,171 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.ui.server.console.settings;
+
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+import static google.registry.request.Action.Method.GET;
+import static google.registry.request.Action.Method.POST;
+
+import avro.shaded.com.google.common.collect.ImmutableList;
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.gson.Gson;
+import google.registry.flows.certs.CertificateChecker;
+import google.registry.flows.certs.CertificateChecker.InsecureCertificateException;
+import google.registry.model.console.ConsolePermission;
+import google.registry.model.console.User;
+import google.registry.model.registrar.Registrar;
+import google.registry.request.Action;
+import google.registry.request.Parameter;
+import google.registry.request.Response;
+import google.registry.request.auth.Auth;
+import google.registry.request.auth.AuthResult;
+import google.registry.request.auth.AuthenticatedRegistrarAccessor;
+import google.registry.request.auth.AuthenticatedRegistrarAccessor.RegistrarAccessDeniedException;
+import google.registry.ui.server.registrar.JsonGetAction;
+import java.util.Optional;
+import javax.inject.Inject;
+import javax.servlet.http.HttpServletRequest;
+
+@Action(
+    service = Action.Service.DEFAULT,
+    path = SecurityAction.PATH,
+    method = {GET, POST},
+    auth = Auth.AUTH_PUBLIC_LOGGED_IN)
+public class SecurityAction implements JsonGetAction {
+
+  static final String PATH = "/console-api/settings/security";
+  private final HttpServletRequest req;
+  private final AuthResult authResult;
+  private final Response response;
+  private final Gson gson;
+  private final String registrarId;
+  private AuthenticatedRegistrarAccessor registrarAccessor;
+  private Optional<Registrar> registrar;
+  private CertificateChecker certificateChecker;
+
+  @Inject
+  public SecurityAction(
+      HttpServletRequest req,
+      AuthResult authResult,
+      Response response,
+      Gson gson,
+      CertificateChecker certificateChecker,
+      AuthenticatedRegistrarAccessor registrarAccessor,
+      @Parameter("registrarId") String registrarId,
+      @Parameter("registrar") Optional<Registrar> registrar) {
+    this.req = req;
+    this.authResult = authResult;
+    this.response = response;
+    this.gson = gson;
+    this.registrarId = registrarId;
+    this.registrarAccessor = registrarAccessor;
+    this.registrar = registrar;
+    this.certificateChecker = certificateChecker;
+  }
+
+  @Override
+  public void run() {
+    if (req.getMethod().equals(GET.toString())) {
+      getHandler();
+    } else {
+      postHandler();
+    }
+  }
+
+  private void getHandler() {
+    try {
+      Registrar registrar = registrarAccessor.getRegistrar(registrarId);
+      response.setStatus(HttpStatusCodes.STATUS_CODE_OK);
+      response.setPayload(gson.toJson(registrar));
+    } catch (RegistrarAccessDeniedException e) {
+      response.setStatus(HttpStatusCodes.STATUS_CODE_FORBIDDEN);
+      response.setPayload(e.getMessage());
+    }
+  }
+
+  private void postHandler() {
+    User user = authResult.userAuthInfo().get().consoleUser().get();
+    if (!user.getUserRoles().hasPermission(registrarId, ConsolePermission.EDIT_REGISTRAR_DETAILS)) {
+      response.setStatus(HttpStatusCodes.STATUS_CODE_FORBIDDEN);
+      return;
+    }
+
+    if (!registrar.isPresent()) {
+      response.setStatus(HttpStatusCodes.STATUS_CODE_BAD_REQUEST);
+      response.setPayload(gson.toJson("'registrar' parameter is not present"));
+      return;
+    }
+
+    Registrar savedRegistrar;
+    try {
+      savedRegistrar = registrarAccessor.getRegistrar(registrarId);
+    } catch (RegistrarAccessDeniedException e) {
+      response.setStatus(HttpStatusCodes.STATUS_CODE_FORBIDDEN);
+      response.setPayload(e.getMessage());
+      return;
+    }
+
+    tm().transact(() -> setResponse(savedRegistrar));
+  }
+
+  private void setResponse(Registrar savedRegistrar) {
+    Registrar registrarParameter = registrar.get();
+    Registrar.Builder updatedRegistrar =
+        savedRegistrar
+            .asBuilder()
+            .setIpAddressAllowList(registrarParameter.getIpAddressAllowList());
+
+    boolean hasInvalidCerts =
+        ImmutableList.of(
+                registrarParameter.getClientCertificate(),
+                registrarParameter.getFailoverClientCertificate())
+            .stream()
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .anyMatch(
+                cert -> {
+                  try {
+                    certificateChecker.validateCertificate(cert);
+                    return false;
+                  } catch (InsecureCertificateException e) {
+                    return true;
+                  }
+                });
+
+    if (hasInvalidCerts) {
+      response.setStatus(HttpStatusCodes.STATUS_CODE_BAD_REQUEST);
+      response.setPayload("Insecure Certificate in parameter");
+      return;
+    }
+
+    registrarParameter
+        .getClientCertificate()
+        .ifPresent(
+            newClientCert -> {
+              updatedRegistrar.setClientCertificate(newClientCert, tm().getTransactionTime());
+            });
+
+    registrarParameter
+        .getFailoverClientCertificate()
+        .ifPresent(
+            failoverCert -> {
+              updatedRegistrar.setFailoverClientCertificate(
+                  failoverCert, tm().getTransactionTime());
+            });
+
+    tm().put(updatedRegistrar.build());
+    response.setStatus(HttpStatusCodes.STATUS_CODE_OK);
+  }
+}
--- a/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java
+++ b/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java
@@ -24,6 +24,7 @@ import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import dagger.Module;
 import dagger.Provides;
+import google.registry.model.registrar.Registrar;
 import google.registry.model.registrar.RegistrarPoc;
 import google.registry.request.OptionalJsonPayload;
 import google.registry.request.Parameter;
@@ -186,5 +187,16 @@ public final class RegistrarConsoleModule {
   @Parameter("registrarId")
   static String provideRegistrarId(HttpServletRequest req) {
     return extractRequiredParameter(req, "registrarId");
+  }
+
+  @Provides
+  @Parameter("registrar")
+  public static Optional<Registrar> provideRegistrar(
+      Gson gson, @OptionalJsonPayload Optional<JsonObject> payload) {
+    if (payload.isPresent() && payload.get().has("registrar")) {
+      return Optional.of(gson.fromJson(payload.get().get("registrar"), Registrar.class));
+    }
+
+    return Optional.empty();
   }
 }
--- a/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java
+++ b/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java
@@ -1,0 +1,183 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.ui.server.console.settings;
+
+import static com.google.common.truth.Truth.assertThat;
+import static google.registry.testing.CertificateSamples.SAMPLE_CERT;
+import static google.registry.testing.CertificateSamples.SAMPLE_CERT2;
+import static google.registry.testing.DatabaseHelper.loadRegistrar;
+import static google.registry.testing.DatabaseHelper.persistResource;
+import static google.registry.testing.SqlHelper.saveRegistrar;
+import static google.registry.util.DateTimeUtils.START_OF_TIME;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSetMultimap;
+import com.google.common.collect.ImmutableSortedMap;
+import com.google.common.net.InetAddresses;
+import com.google.gson.Gson;
+import google.registry.flows.certs.CertificateChecker;
+import google.registry.model.console.GlobalRole;
+import google.registry.model.console.User;
+import google.registry.model.console.UserRoles;
+import google.registry.model.registrar.Registrar;
+import google.registry.persistence.transaction.JpaTestExtensions;
+import google.registry.request.Action;
+import google.registry.request.RequestModule;
+import google.registry.request.auth.AuthResult;
+import google.registry.request.auth.AuthSettings.AuthLevel;
+import google.registry.request.auth.AuthenticatedRegistrarAccessor;
+import google.registry.request.auth.UserAuthInfo;
+import google.registry.testing.FakeClock;
+import google.registry.testing.FakeResponse;
+import google.registry.ui.server.registrar.RegistrarConsoleModule;
+import google.registry.util.CidrAddressBlock;
+import google.registry.util.UtilsModule;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.util.Optional;
+import javax.servlet.http.HttpServletRequest;
+import org.joda.time.DateTime;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+/** Tests for {@link google.registry.ui.server.console.settings.SecurityAction}. */
+class SecurityActionTest {
+
+  private static String jsonRegistrar1 =
+      String.format(
+          "{\"registrarId\": \"registrarId\", \"clientCertificate\": \"%s\","
+              + " \"ipAddressAllowList\": [\"192.168.1.1/32\"]}",
+          SAMPLE_CERT2);
+  private static final Gson GSON = UtilsModule.provideGson();
+  private final HttpServletRequest request = mock(HttpServletRequest.class);
+  private final FakeClock clock = new FakeClock();
+  private Registrar testRegistrar;
+  private FakeResponse response = new FakeResponse();
+
+  private AuthenticatedRegistrarAccessor registrarAccessor =
+      AuthenticatedRegistrarAccessor.createForTesting(
+          ImmutableSetMultimap.of("registrarId", AuthenticatedRegistrarAccessor.Role.ADMIN));
+
+  private CertificateChecker certificateChecker =
+      new CertificateChecker(
+          ImmutableSortedMap.of(START_OF_TIME, 20825, DateTime.parse("2020-09-01T00:00:00Z"), 398),
+          30,
+          15,
+          2048,
+          ImmutableSet.of("secp256r1", "secp384r1"),
+          clock);
+
+  @RegisterExtension
+  final JpaTestExtensions.JpaIntegrationTestExtension jpa =
+      new JpaTestExtensions.Builder().withClock(clock).buildIntegrationTestExtension();
+
+  @BeforeEach
+  void beforeEach() {
+    testRegistrar = saveRegistrar("registrarId");
+  }
+
+  @Test
+  void testSuccess_getRegistrarInfo() throws IOException {
+    persistResource(
+        testRegistrar
+            .asBuilder()
+            .setClientCertificate(SAMPLE_CERT, clock.nowUtc())
+            .setIpAddressAllowList(
+                ImmutableSet.of(
+                    CidrAddressBlock.create(InetAddresses.forString("192.168.1.1"), 32),
+                    CidrAddressBlock.create(InetAddresses.forString("2001:db8::1"), 128)))
+            .build());
+    SecurityAction action =
+        createAction(
+            Action.Method.GET,
+            AuthResult.create(
+                AuthLevel.USER,
+                UserAuthInfo.create(
+                    createUser(new UserRoles.Builder().setGlobalRole(GlobalRole.FTE).build()))),
+            testRegistrar.getRegistrarId());
+    action.run();
+    assertThat(response.getStatus()).isEqualTo(HttpStatusCodes.STATUS_CODE_OK);
+    String payload = response.getPayload().replace("\\n", "").replace("\\u003d", "=");
+    assertThat(payload).contains(SAMPLE_CERT.replace("\n", ""));
+    assertThat(payload).contains("192.168.1.1/32");
+    assertThat(payload).contains("2001:db8:0:0:0:0:0:1/128");
+  }
+
+  @Test
+  void testSuccess_postRegistrarInfo() throws IOException {
+    clock.setTo(DateTime.parse("2020-11-01T00:00:00Z"));
+    SecurityAction action =
+        createAction(
+            Action.Method.POST,
+            AuthResult.create(
+                AuthLevel.USER,
+                UserAuthInfo.create(
+                    createUser(new UserRoles.Builder().setGlobalRole(GlobalRole.FTE).build()))),
+            testRegistrar.getRegistrarId());
+    action.run();
+    assertThat(response.getStatus()).isEqualTo(HttpStatusCodes.STATUS_CODE_OK);
+    Registrar r = loadRegistrar(testRegistrar.getRegistrarId());
+    assertThat(r.getClientCertificateHash().get())
+        .isEqualTo("GNd6ZP8/n91t9UTnpxR8aH7aAW4+CpvufYx9ViGbcMY");
+    assertThat(r.getIpAddressAllowList().get(0).getIp()).isEqualTo("192.168.1.1");
+    assertThat(r.getIpAddressAllowList().get(0).getNetmask()).isEqualTo(32);
+  }
+
+  private User createUser(UserRoles userRoles) {
+    return new User.Builder()
+        .setEmailAddress("email@email.com")
+        .setGaiaId("TestUserId")
+        .setUserRoles(userRoles)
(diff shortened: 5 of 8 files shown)


ptkach:
@gbrodman Kind reminder to review this PR. Thanks!

gbrodman (changes requested):
> Reviewed 7 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: all files reviewed, 14 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 87 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbwCwm-gsAual-bl6E:-NZbwCwm-gsAual-bl6F:bblme79) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L87)):*
> ```Java
>   private void getHandler() {
>     try {
>       Registrar r = registrarAccessor.getRegistrar(registrarId);
> ```

probably should use a longer variable name go/java-practices/naming#single-letter-vars just in case (or just inline it, with no variable)
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxAzl9M2YnQ9fWONp:-NZbxAzl9M2YnQ9fWONq:b5ka6ua) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L113)):*
> ```Java
>       tm().transact(
>               () -> {
>                 tm().getEntityManager().detach(savedRegistrar);
> ```

What's the purpose of detaching it here? I'm sure there is one, I just want to know the details 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 119 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxlbK7Tc3dmpVib8Y:-NZbxlbK7Tc3dmpVib8Z:b-puwtld) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L119)):*
> ```Java
>                         .asBuilder()
>                         .setIpAddressAllowList(registrar.get().getIpAddressAllowList());
>                 if (registrar.get().getClientCertificate().isPresent()
> ```

there's a few duplicated calls to `Optional::get` at various points in this method (e.g. the registrar) and we could probably reduce complexity of some individual lines if we separate those into separate variables? 

tbh this might only apply to the registrar and transaction time 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbySzbEeXamL9UMrYZ:-NZbySzbEeXamL9UMrY_:bu3bqc5) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L139)):*
> ```Java
> 
>       response.setStatus(HttpStatusCodes.STATUS_CODE_OK);
>     } catch (RegistrarAccessDeniedException e) {
> ```

I think we can probably move this out so that only the `registrarAccessor` method is covered, right? AFAIK this should never happen because we've checked the permission for EDIT_REGISTRAR_DETAILS above 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 147 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbzmUN8SbvybaEBHHo:-NZbzmUN8SbvybaEBHHp:baf2iiz) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L147)):*
> ```Java
>   private boolean validateCertificate(
>       Optional<String> existingCertificate, String certificateString) {
>     if (!existingCertificate.isPresent() || !existingCertificate.get().equals(certificateString)) {
> ```

i know this is just copied from RegistrarSettingsAction but i think we'd probably just be better served by always calling `validateCertificate` if the "new" registrar has that particular cert present, then setting it (handling any InsecureCertificateExceptions) 

maybe like 

```
newRegistrar.getClientCertificate().ifPresent(newClientCert -> {
  try {
    certificateChecker.validateCertificate(newClientCert);
    updatedRegistrar.setClientCertificate(newClientCert, transactionTime);
  catch (ICE e) {
    // handling
  }
}
```

maybe having both calls inside a single try block (instead of having two try blocks inside the `ifPresent` calls?

seems like that might end up being more concise / clear at the cost of only a bit of extra validation 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 151 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbyw-D0Z-Q_Qsp5tDL:-NZbyw-D0Z-Q_Qsp5tDM:b-r8wctp) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L151)):*
> ```Java
>         certificateChecker.validateCertificate(certificateString);
>       } catch (InsecureCertificateException e) {
>         throw new IllegalArgumentException(e.getMessage());
> ```

we'll want exception handling here right? 
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 91 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc4cxQ9y9uUoBMnRUd:-NZc4cxQ9y9uUoBMnRUe:b5tj7jz) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L91)):*
> ```Java
>   @RegisterExtension
>   final JpaTestExtensions.JpaIntegrationTestExtension jpa =
>       new JpaTestExtensions.Builder().buildIntegrationTestExtension();
> ```

if we're creating / using a clock we should have it be the JPA / SQL clock as well 
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 95 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5FIe7eZFtx0s5lm1:-NZc5FIe7eZFtx0s5lm2:bymfs5i) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L95)):*
> ```Java
>   @BeforeEach
>   void beforeEach() {
>     response = new FakeResponse();
> ```

just create this inline
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 96 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5Srn8bFXlMsZ_3Q3:-NZc5Srn8bFXlMsZ_3Q4:b-1j42gm) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L96)):*
> ```Java
>   void beforeEach() {
>     response = new FakeResponse();
>     testRegistrar = saveRegistrar("registrarId");
> ```

Is there a reason to use this instead of using the built-in JpaIntegrationExtension canned data? 
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 120 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5i5t0gt4chNHr9Xv:-NZc5i5t0gt4chNHr9Xw:bwt1ro7) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L120)):*
> ```Java
>     action.run();
>     assertThat(response.getStatus()).isEqualTo(HttpStatusCodes.STATUS_CODE_OK);
>     String payload = response.getPayload().replace("\\n", "").replace("\\u003d", "=");
> ```

what's up with these, out of curiosity? 

I assume it's necessary, especially since in an ideal world we wouldn't have to replace the newlines in SAMPLE_CERT in the next line 
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 489 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc8esD-L2llJWOu2UD:-NZc8esD-L2llJWOu2UE:becdqoi) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L489)):*
> ```Java
> 
>     @Override
>     public CidrAddressBlock read(JsonReader reader) throws IOException {
> ```

Do we want to close the reader? 
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 493 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc87EH6PJYGcAC8_x7:-NZc87EH6PJYGcAC8_x8:b-fek382) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L493)):*
> ```Java
>       CidrAddressBlock cidrAddressBlock = null;
>       String fieldname = null;
>       while (reader.hasNext()) {
> ```

i'm a bit confused by this, I think this may be a bit overcomplicating the type adapter? 

Shouldn't it just be getting the `nextString` on the reader, which returns the next (only) value? It seems like this is writing this as a single object `{"cidrAddressBlock": "someString"}`
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 499 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc7WZl8X4PGcPh5xp7:-NZc7WZl8X4PGcPh5xp8:b-7k9d9g) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L499)):*
> ```Java
>         }
>         if (FIELD_NAME.equals(fieldname)) {
>           reader.peek();
> ```

unnecessary line?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: 5 of 8 files reviewed, 14 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 87 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbwCwm-gsAual-bl6E:-NZgOWiJD0BROYA5fPbz:bczwrjp) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L87)):*

probably should use a longer variable name go/java-practices/naming#single-letter-vars just in case (or just inline it, with no variable)
</blockquote></details>

Updated, although I think single letter variable considered to be ok when it's short-lived, like in this case - 2 lines only
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxAzl9M2YnQ9fWONp:-NZgPADPC6fvYs2U57rx:brp3snv) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L113)):*

What's the purpose of detaching it here? I'm sure there is one, I just want to know the details 
</blockquote></details>

I believe it's to avoid update collisions. I've followed an existing approach - https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java;l=245 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 119 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxlbK7Tc3dmpVib8Y:-NZgQYWM6qE3I3lxC5Sj:b-896fix) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L119)):*

there's a few duplicated calls to `Optional::get` at various points in this method (e.g. the registrar) and we could probably reduce complexity of some individual lines if we separate those into separate variables? 

tbh this might only apply to the registrar and transaction time 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbySzbEeXamL9UMrYZ:-NZgRC51-Ue2KUSnMp8r:b-762jw9) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L139)):*

I think we can probably move this out so that only the `registrarAccessor` method is covered, right? AFAIK this should never happen because we've checked the permission for EDIT_REGISTRAR_DETAILS above 
</blockquote></details>

I'm not sure removing the catch would be correct. There's multiple checks which throw in https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java;l=222 and permission (which we cover above) is just one of them 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 147 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbzmUN8SbvybaEBHHo:-NZgvXno2d3ACVo3ZqVu:bnpykjp) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L147)):*

i know this is just copied from RegistrarSettingsAction but i think we'd probably just be better served by always calling `validateCertificate` if the "new" registrar has that particular cert present, then setting it (handling any InsecureCertificateExceptions) 

maybe like 

```
newRegistrar.getClientCertificate().ifPresent(newClientCert -> {
  try {
    certificateChecker.validateCertificate(newClientCert);
    updatedRegistrar.setClientCertificate(newClientCert, transactionTime);
  catch (ICE e) {
    // handling
  }
}
```

maybe having both calls inside a single try block (instead of having two try blocks inside the `ifPresent` calls?

seems like that might end up being more concise / clear at the cost of only a bit of extra validation 
</blockquote></details>

Good suggestion, updated
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 151 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbyw-D0Z-Q_Qsp5tDL:-NZgvgHnBmYDnEL8j4Pp:b1kwlvm) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L151)):*

we'll want exception handling here right? 
</blockquote></details>

Correct, updated
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/2057#gh-1233080968:-NZh3sBY3pmwqSjJydab:b-896fix) ([raw file](https://github.com/google/nomulus/blob/30adc6ddc9a19fd5812ada09722cb3406b370a0b/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L63)):*

Unused class: SecurityActionTest is not referenced within this codebase. If not used as an external API it should be removed.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1517)
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 91 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc4cxQ9y9uUoBMnRUd:-NZgw2zB1JKyap5s1phq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L91)):*

if we're creating / using a clock we should have it be the JPA / SQL clock as well 
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 95 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5FIe7eZFtx0s5lm1:-NZgwGVLEBXnNEcxcrDA:blya761) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L95)):*

just create this inline
</blockquote></details>

Why? it's common to all suits and it seems easier this way
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 96 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5Srn8bFXlMsZ_3Q3:-NZgwpDt6T0nefrcUJwT:bnwgplu) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L96)):*

Is there a reason to use this instead of using the built-in JpaIntegrationExtension canned data? 
</blockquote></details>

I prefer to rely on data scoped in the test, I believe this makes it easier to follow what's going on and more importantly makes less prone to be broken by changes outside of the test. 
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 120 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5i5t0gt4chNHr9Xv:-NZgxRSPBFTXBe7-YpFT:bz4v0od) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L120)):*

what's up with these, out of curiosity? 

I assume it's necessary, especially since in an ideal world we wouldn't have to replace the newlines in SAMPLE_CERT in the next line 
</blockquote></details>

Ugh, it took me a some time to figure out. GSON is doing escapement and inlining of the data in the request, so I had to match the changes to make a proper comparison. Couldn't find a better way
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 489 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc8esD-L2llJWOu2UD:-NZgywn8-HnsGlQ47f2Z:b-drjyp0) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L489)):*

Do we want to close the reader? 
</blockquote></details>

As far as I understand closing the reader is not part of data adapters responsibility, at least I've not seen it done in an adapter:
**Data adapter you added** - https://cs.opensource.google/nomulus/nomulus/+/master:util/src/main/java/google/registry/util/DateTimeTypeAdapter.java;l=35 
**Google Maps project** - https://github.com/googlemaps/google-maps-services-java/blob/main/src/main/java/com/google/maps/internal/FareAdapter.java#L40
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 493 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc87EH6PJYGcAC8_x7:-NZh2PAgAzx9c1vC2_fH:b-bu91xm) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L493)):*

i'm a bit confused by this, I think this may be a bit overcomplicating the type adapter? 

Shouldn't it just be getting the `nextString` on the reader, which returns the next (only) value? It seems like this is writing this as a single object `{"cidrAddressBlock": "someString"}`
</blockquote></details>

Updated
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 499 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc7WZl8X4PGcPh5xp7:-NZh2Pzy1MbTZhw-wtmp:ba3dfob) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L499)):*

unnecessary line?
</blockquote></details>

Removed

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: all files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxAzl9M2YnQ9fWONp:-NZhAlDs3A3bX_tm7G8y:b-9rw7vw) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L113)):*

I believe it's to avoid update collisions. I've followed an existing approach - https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java;l=245 
</blockquote></details>

ah ok, I think we might not need that because we create a new object here. Have you tried not detaching it here? I think in RegistrarSettingsAction we have to do that because when we update linked objects (e.g. contacts) the original registrar object updates, even if we use `asBuilder` 

Basically, there we only did it because we needed to make 100% sure that we had a reference to the old object (which shouldn't be necessary here) 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbySzbEeXamL9UMrYZ:-NZhCMbK2TdsgHStBRi_:bl96cy7) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L139)):*

I'm not sure removing the catch would be correct. There's multiple checks which throw in https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java;l=222 and permission (which we cover above) is just one of them 
</blockquote></details>

It looks like all those checks are already covered by the permission check we do above though (which makes sense, since if you have EDIT_ on a registrar, you have access to that registrar) 
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 95 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5FIe7eZFtx0s5lm1:-NZhCriO5_egBqhcr1eB:bkd11ps) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L95)):*

Why? it's common to all suits and it seems easier this way
</blockquote></details>

it's the same amount of work (it's created once per test), it's a line of code removed, and it means that we create + declare at the same time which is good
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 96 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5Srn8bFXlMsZ_3Q3:-NZhIPiNC6xUc-X-u9Y2:bb7g23t) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L96)):*

I prefer to rely on data scoped in the test, I believe this makes it easier to follow what's going on and more importantly makes less prone to be broken by changes outside of the test. 
</blockquote></details>

fair
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 120 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5i5t0gt4chNHr9Xv:-NZhITYE-ADdTdu9KRbv:berhkv1) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L120)):*

Ugh, it took me a some time to figure out. GSON is doing escapement and inlining of the data in the request, so I had to match the changes to make a proper comparison. Couldn't find a better way
</blockquote></details>

interesting -- do you happen to know if this will cause any issues when trying to read/write the data on the front end? 
___
*[`util/src/main/java/google/registry/util/CidrAddressBlock.java` line 489 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc8esD-L2llJWOu2UD:-NZhFT8-5zdp22PQTxB5:baee7s2) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/util/src/main/java/google/registry/util/CidrAddressBlock.java#L489)):*

As far as I understand closing the reader is not part of data adapters responsibility, at least I've not seen it done in an adapter:
**Data adapter you added** - https://cs.opensource.google/nomulus/nomulus/+/master:util/src/main/java/google/registry/util/DateTimeTypeAdapter.java;l=35 
**Google Maps project** - https://github.com/googlemaps/google-maps-services-java/blob/main/src/main/java/com/google/maps/internal/FareAdapter.java#L40

</blockquote></details>

haha fair enough!

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxAzl9M2YnQ9fWONp:-NZlbYmd0NVlKn00t-_q:b-ynxlda) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L113)):*

ah ok, I think we might not need that because we create a new object here. Have you tried not detaching it here? I think in RegistrarSettingsAction we have to do that because when we update linked objects (e.g. contacts) the original registrar object updates, even if we use `asBuilder` 

Basically, there we only did it because we needed to make 100% sure that we had a reference to the old object (which shouldn't be necessary here) 
</blockquote></details>

I see, thanks for the clarification. Removed.
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbySzbEeXamL9UMrYZ:-NZlcAsy9G0vyZl04FUn:bvdpm6q) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L139)):*

It looks like all those checks are already covered by the permission check we do above though (which makes sense, since if you have EDIT_ on a registrar, you have access to that registrar) 
</blockquote></details>

This is the case I don't see covered in perm check - https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java;l=231

However per your suggestion I have moved catch block
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 95 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5FIe7eZFtx0s5lm1:-NZlcWnM2jBm0WTIxWGP:b-pqesmj) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L95)):*

it's the same amount of work (it's created once per test), it's a line of code removed, and it means that we create + declare at the same time which is good
</blockquote></details>

Moved the line to `createAction` method.
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 120 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5i5t0gt4chNHr9Xv:-NZlhr5b0ZHrCO70L7o5:b74sbio) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L120)):*

interesting -- do you happen to know if this will cause any issues when trying to read/write the data on the front end? 
</blockquote></details>

No, it's just an innocence here, because I have to compare it to the string I declared.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: 5 of 8 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 143 at r4](https://reviewable.io/reviews/google/nomulus/2057#gh-1256344060:-NZm0MSF1s7MHNWCihMj:b87f0hv) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L143)):*

## Boxed variable is never null

The variable 'hasInvalidCerts' is only assigned values of primitive type and is never 'null', but it is declared with the boxed type 'Boolean'.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1537)
</blockquote></details>

Seems like this should be `boolean`

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: all files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 113 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbxAzl9M2YnQ9fWONp:-NZlw4kZD0iob_kTW3T5:b-pmf5hk) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L113)):*

I see, thanks for the clarification. Removed.
</blockquote></details>

yep -- I meant to mention this in a previous comment but specifically that's because we needed / wanted to send out email updates 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 139 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZbySzbEeXamL9UMrYZ:-NZlvPtB49dpg6UIkG7w:bby6ggc) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L139)):*

This is the case I don't see covered in perm check - https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/auth/AuthenticatedRegistrarAccessor.java;l=231

However per your suggestion I have moved catch block
</blockquote></details>

yeah that is....a super weird exception that should never ever happen hahaha (we probably could even just remove it) 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 99 at r4](https://reviewable.io/reviews/google/nomulus/2057#-NZm0_Bg0BeN1d44fFac:-NZm0_Bg0BeN1d44fFad:b-ox7igc) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L99)):*
> ```Java
> 
>   private void postHandler() {
>     Registrar savedRegistrar;
> ```

variables should be declared as close as possible to when they're used go/java-style#s4.8.2.2-variables-limited-scope 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 121 at r4](https://reviewable.io/reviews/google/nomulus/2057#-NZm0nBC1TkP4w6k9Vzw:-NZm0nBC1TkP4w6k9Vzx:b-mvxswn) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L121)):*
> ```Java
> 
>     Registrar registrarParameter = registrar.get();
>     tm().transact(
> ```

optional: it's possible that it would be neater whitespace-wise to have the within-transaction code in a separate method.

i don't know if that's actually the case, worth thinking about if you haven't 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 143 at r4](https://reviewable.io/reviews/google/nomulus/2057#gh-1256344060:-NZlvHE99fkddq0q5575:b2nigms) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L143)):*

Seems like this should be `boolean`
</blockquote></details>

yee
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 95 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5FIe7eZFtx0s5lm1:-NZlzotEBOtZr0rcmDua:b-volgvc) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L95)):*

Moved the line to `createAction` method.
</blockquote></details>

that's still not at the declaration, which is the best way of doing it 

each class variable is created anew before each and every test

https://www.youtube.com/watch?v=cs2Wu9Co-2s
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 73 at r4](https://reviewable.io/reviews/google/nomulus/2057#-NZm-VC0E688xKN4HTkv:-NZm-VC0E688xKN4HTkw:bmauuyu) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L73)):*
> ```Java
>   private final HttpServletRequest request = mock(HttpServletRequest.class);
>   private static final Gson GSON = UtilsModule.provideGson();
>   private FakeResponse response;
> ```

somewhat nitty, but variable declarations should be grouped by variable type (meaning private/public, static/non-static)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: all files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 99 at r4](https://reviewable.io/reviews/google/nomulus/2057#-NZm0_Bg0BeN1d44fFac:-N_-xEhT1pj8TIfh0xZm:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L99)):*

variables should be declared as close as possible to when they're used go/java-style#s4.8.2.2-variables-limited-scope 
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 121 at r4](https://reviewable.io/reviews/google/nomulus/2057#-NZm0nBC1TkP4w6k9Vzw:-N_9Qr_XEn3-wGR_kylH:b-bu91xm) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L121)):*

optional: it's possible that it would be neater whitespace-wise to have the within-transaction code in a separate method.

i don't know if that's actually the case, worth thinking about if you haven't 
</blockquote></details>

Updated
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 143 at r4](https://reviewable.io/reviews/google/nomulus/2057#gh-1256344060:-N_0OjIcB8nr1rNAiBp8:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L143)):*

yee
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 95 at r2](https://reviewable.io/reviews/google/nomulus/2057#-NZc5FIe7eZFtx0s5lm1:-N_9_RwlCB4BoAOQKIPD:b-896fix) ([raw file](https://github.com/google/nomulus/blob/31cc5037fd518983e1a1affd3c935cc725f1ce07/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L95)):*

that's still not at the declaration, which is the best way of doing it 

each class variable is created anew before each and every test

https://www.youtube.com/watch?v=cs2Wu9Co-2s
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java` line 73 at r4](https://reviewable.io/reviews/google/nomulus/2057#-NZm-VC0E688xKN4HTkv:-N_0P4LR0XoUVokuerVO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4ebe52bf470bb39cb9ce9d4286fabea4e5f587b9/core/src/test/java/google/registry/ui/server/console/settings/SecurityActionTest.java#L73)):*

somewhat nitty, but variable declarations should be grouped by variable type (meaning private/public, static/non-static)  
</blockquote></details>

Done.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2057)* status: all files reviewed, 2 unresolved discussions (waiting on @github-code-scanning[bot])

PR #2061, opened 2023-06-21
Add a TldSettings class for modifying Tld objects with YAML files

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2061)

### DIFF
--- a/core/src/main/java/google/registry/model/CreateAutoTimestamp.java
+++ b/core/src/main/java/google/registry/model/CreateAutoTimestamp.java
@@ -46,6 +46,14 @@ public class CreateAutoTimestamp extends ImmutableObject implements UnsafeSerial
     return creationTime;
   }
 
+  @Override
+  public String toString() {
+    if (creationTime == null) {
+      return null;
+    }
+    return creationTime.toString();
+  }
+
   public static CreateAutoTimestamp create(@Nullable DateTime creationTime) {
     CreateAutoTimestamp instance = new CreateAutoTimestamp();
     instance.creationTime = creationTime;
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -296,6 +296,7 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
    * <p>When this field is null, the "dnsDefaultDsTtl" value from the config file will be used.
    */
   Duration dnsDsTtl;
+
   /**
    * The unicode-aware representation of the TLD associated with this {@link Tld}.
    *
@@ -688,6 +689,14 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
     return Optional.ofNullable(dnsDsTtl);
   }
 
+  public String getTldUnicode() {
+    return tldUnicode;
+  }
+
+  public boolean getInvoicingEnabled() {
+    return invoicingEnabled;
+  }
+
   public ImmutableSet<String> getAllowedRegistrantContactIds() {
     return nullToEmptyImmutableCopy(allowedRegistrantContactIds);
   }
@@ -806,6 +815,14 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
           addGracePeriodLength.isLongerThan(Duration.ZERO),
           "addGracePeriodLength must be non-zero");
       getInstance().addGracePeriodLength = addGracePeriodLength;
+      return this;
+    }
+
+    public Builder setAnchorTenantAddGracePeriodLength(Duration anchorTenantAddGracePeriodLength) {
+      checkArgument(
+          anchorTenantAddGracePeriodLength.isLongerThan(Duration.ZERO),
+          "anchorTenantAddGracePeriodLength must be non-zero");
+      getInstance().anchorTenantAddGracePeriodLength = anchorTenantAddGracePeriodLength;
       return this;
     }
 
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -888,7 +888,8 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
 
     public Builder setNullableReservedListsByName(@Nullable Set<String> reservedListNames) {
       if (reservedListNames == null) {
-        return setNullReservedLists();
+        getInstance().reservedListNames = null;
+        return this;
       }
       return setReservedListsByName(reservedListNames);
     }
@@ -919,11 +920,6 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
         nameBuilder.add(reservedList.getName());
       }
       getInstance().reservedListNames = nameBuilder.build();
-      return this;
-    }
-
-    public Builder setNullReservedLists() {
-      getInstance().reservedListNames = null;
       return this;
     }
 
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -886,6 +886,13 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
       return this;
     }
 
+    public Builder setNullableReservedListsByName(@Nullable Set<String> reservedListNames) {
+      if (reservedListNames == null) {
+        return setNullReservedLists();
+      }
+      return setReservedListsByName(reservedListNames);
+    }
+
     public Builder setReservedListsByName(Set<String> reservedListNames) {
       checkArgument(reservedListNames != null, "reservedListNames must not be null");
       ImmutableSet.Builder<ReservedList> builder = new ImmutableSet.Builder<>();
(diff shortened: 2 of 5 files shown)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2061)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/CreateAutoTimestamp.java` line 29 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_JrQsFDNJl4TLULvAr:-N_JrQsFDNJl4TLULvAs:b-42o3vs) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/CreateAutoTimestamp.java#L29)):*
> ```Java
> /** A timestamp that auto-updates when first saved to the database. */
> @Embeddable
> public class CreateAutoTimestamp extends ImmutableObject implements UnsafeSerializable {
> ```

Better make this class `final` now that `toString` method is overwritten.



_Code quote:_
```Java
public class
```
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 894 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_JrwPq408HNNQ2SGal:-N_JrwPq408HNNQ2SGam:bfntjmh) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/tld/Tld.java#L894)):*
> ```Java
>         return this;
>       }
>       return setReservedListsByName(reservedListNames);
> ```

External mutable containers should not be saved by reference.

Either make this `ImmutableSet.copyOf(reservedListNames)`, or make the parameter type Immutable.



_Code quote:_
```Java
reservedListName
```

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2061)* status: 0 of 5 files reviewed, 7 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[`core/src/main/java/google/registry/model/CreateAutoTimestamp.java` line 52 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_L89SRDopdLrys3EJ0:-N_L89SRDopdLrys3EJ1:bycr4) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/CreateAutoTimestamp.java#L52)):*
> ```Java
>   public String toString() {
>     if (creationTime == null) {
>       return null;
> ```

I don't think it makes sense for a `toString()` method to return `null`?  What's the use case for this?  Returning `null` or `(null)` or similar makes more sense.  See more info on this type of issue here:  https://stackoverflow.com/questions/29513497/why-tostring-on-an-object-instance-which-is-null-is-not-throwing-npe
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 889 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_L8t5tAheWCAE7YRqU:-N_L8t5tAheWCAE7YRqV:bqqpxms) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/tld/Tld.java#L889)):*
> ```Java
>     }
> 
>     public Builder setNullableReservedListsByName(@Nullable Set<String> reservedListNames) {
> ```

Why does this need to be able to take `null`?  That seems like an error; if you want to zero this out, you should have to pass an empty set.
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 1013 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_L8yGd-GSErziVaakp:-N_L8yGd-GSErziVaakq:bpwjitq) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/tld/Tld.java#L1013)):*
> ```Java
> 
>     public Builder setDefaultPromoTokens(
>         @Nullable ImmutableList<VKey<AllocationToken>> promoTokens) {
> ```

Same issue with null; I don't see why that should be valid.  Prefer `nullToEmptyImmutableCopy()` instead.
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 1039 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_L95F7B5gO0oxywEh5:-N_L95F7B5gO0oxywEh6:b-du3cv) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/tld/Tld.java#L1039)):*
> ```Java
>     }
> 
>     public Builder setIdnTables(@Nullable ImmutableSet<IdnTableEnum> idnTables) {
> ```

Ditto on null.
___
*[`core/src/main/java/google/registry/model/tld/TldSettings.java` line 48 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_L9FBU5o4l_F09X2FH:-N_L9FBU5o4l_F09X2FI:bh7xsnn) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/main/java/google/registry/model/tld/TldSettings.java#L48)):*
> ```Java
> 
> /**
>  * A POJO that can be used to convert {@link google.registry.model.tld.Tld} objects to YAML files
> ```

Why does this class need to exist at all?  Can `Tld` not be reused for this purpose?  What are the issues there?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2061)* status: 0 of 5 files reviewed, 8 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[`core/src/test/resources/google/registry/model/tld/tld.yaml` line 23 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_L9pAbDE73Ha4qNQgv:-N_L9pAcDJngmJ22ibjh:b-dldc3f) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/test/resources/google/registry/model/tld/tld.yaml#L23)):*
> ```YAML
> driveFolderId: null
> eapFeeSchedule:
>   '2000-06-02T00:00:00.000Z': USD 0.00
> ```

Use a sorted map so that these are ordered chronologically in the YAML.

You might want to use sorted maps for everything, otherwise there might be unnecessary churn in the YAML representations during a round trip from Java to YAML and back.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2061)* status: 0 of 5 files reviewed, 11 unresolved discussions (waiting on @sarahcaseybot and @weiminyu)

___
*[`core/src/test/resources/google/registry/model/tld/tld.yaml` line 3 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_LAkXuCtlKEf0EFi9B:-N_LAkXuCtlKEf0EFi9C:b2qqjo1) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/test/resources/google/registry/model/tld/tld.yaml#L3)):*
> ```YAML
> addGracePeriodLength: PT432000S
> allowedFullyQualifiedHostNames:
> - foo
> ```

Shouldn't these list items be indented?  They are in our nomulus YAML config.  Seems like something is going wrong with the YAML serialization.
___
*[`core/src/test/resources/google/registry/model/tld/tld.yaml` line 10 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_LAUJr3ng5nu2BgHOP:-N_LAUJr3ng5nu2BgHOQ:b-h1fqkt) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/test/resources/google/registry/model/tld/tld.yaml#L10)):*
> ```YAML
> claimsPeriodEnd: 294247-01-10T04:00:54.775Z
> createBillingCost: USD 13.00
> creationTime: '1970-01-01T00:00:00.000Z'
> ```

Why is this in quote marks but claimsPeriodEnd above isn't.  Are these quote marks necessary at all?  I don't think they are; our YAML nomulus config file doesn't use them anywhere.
___
*[`core/src/test/resources/google/registry/model/tld/tld.yaml` line 15 at r1](https://reviewable.io/reviews/google/nomulus/2061#-N_LAJQ79F2ZdyeJv6Id:-N_LAJQ79F2ZdyeJv6Ie:b-a4snuz) ([raw file](https://github.com/google/nomulus/blob/4c5552d11378a2f8df592eb5cf155e60aa3d9d10/core/src/test/resources/google/registry/model/tld/tld.yaml#L15)):*
> ```YAML
> - bbbbb
> dnsAPlusAaaaTtl: PT3600S
> dnsDsTtl: ''
> ```

I don't think empty string makes sense for these values.  `null` would be better.  The durations are not strings (as you can see below where one that does have a value is not string-escaped).  Also strings in YAML don't need quotes anyway.

PR #2066, opened 2023-07-05
Remove internal auth mechanism

jianglai:
It was used by cron job and task queues, which now use OIDC-based auth.

  Also renamed and consolidated auth enums to make them easier to
  understand. Ultimately we should get rid of the AuthMethod part as OIDC
  will be the only auth method used.

  Based on the updated routing map:

  Backend and tools: the only change is that INTERNAL is removed from allowed
  auth methods. Should be an no-op.

  Pubapi: INTERNAL is removed from allowed auth. For endpoints that only
  allowed INTERNAL before, API and LEGACY become the allowed methods.
  However this should not affect anything because regardless of which auth
  method is ultimately used, the required auth level is always NONE for
  pubapi endpoints. Therefore any auth result is discarded anyway.

  Frontend: INTERNAL is removed. RegistryLockVerifyAction has lowered
  its required auth level to NONE because it extends HtmlAction, which can
  redirect the user to login if necessary. All other endpoints extending
  HtmlAction require NONE, so it's better to keep things consistent.

  PS: I'm not sure where all the frontend code change came from. I didn't
  make any changes, so presumably from the formatting tool.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2066)


jianglai (commented):
> Reviewed 78 of 78 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, 1 of 1 files at r5, 1 of 1 files at r6, 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> Reviewed 70 of 78 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, 1 of 1 files at r6, 7 of 7 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/RequestAuthenticator.java` line 77 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgqs-o-tOzA3A0fz9N:-NZgqs-o-tOzA3A0fz9O:b3h92oe) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/request/auth/RequestAuthenticator.java#L77)):*
> ```Java
>         break;
>     }
>     switch (auth.userPolicy()) {
> ```

we can probably just make this a single if statement right, since we're only doing something if it's ADMIN?
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleUiAction.java` line 127 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgt0SO1HHDVFWwMgKy:-NZgt0SO1HHDVFWwMgKz:b2ni78k) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/ui/server/registrar/ConsoleUiAction.java#L127)):*
> ```Java
>       //
>       // Note that not doing so (and just passing the "clientId" as given) isn't a security issue
>       // since we double-check the access to the registrar on any read / update request. We have to
> ```

lol
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java` line 37 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgub9CEPAduBzBeSCy:-NZgub9CEPAduBzBeSCz:bbai9ne) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java#L37)):*
> ```Java
>     service = Action.Service.DEFAULT,
>     path = RegistryLockVerifyAction.PATH,
>     auth = Auth.AUTH_PUBLIC)
> ```

i think we still want AUTH_PUBLIC_LOGGED_IN here since we want it to be the same person

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/request/auth/RequestAuthenticator.java` line 77 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgqs-o-tOzA3A0fz9N:-N_-KLZ29ZSHqOSWDm_u:b-17021p) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/request/auth/RequestAuthenticator.java#L77)):*

we can probably just make this a single if statement right, since we're only doing something if it's ADMIN?
</blockquote></details>

Changed this and the one above to if statements.

Though to be honest, I kinda felt the switch statements were clearer, albeit more verbose.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java` line 37 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgub9CEPAduBzBeSCy:-N_-O-9F6N_3eMab3mGo:b-r35op9) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java#L37)):*

i think we still want AUTH_PUBLIC_LOGGED_IN here since we want it to be the same person
</blockquote></details>

The same person as whom? Does it matter if a different person with the same permission completes this action?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/RequestAuthenticator.java` line 77 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgqs-o-tOzA3A0fz9N:-N_-p8i4EuCu28CM6Kbz:bsdgn5j) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/request/auth/RequestAuthenticator.java#L77)):*

Changed this and the one above to if statements.

Though to be honest, I kinda felt the switch statements were clearer, albeit more verbose.
</blockquote></details>

yeah i kinda only meant this one because once you end up having more than one clause (which the one above does, this one doesn't) I agree that the `switch` is very possibly better. I'm good either way I think. 
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java` line 37 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgub9CEPAduBzBeSCy:-N_-pIGj1XzOqcY909CI:b2kewdz) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java#L37)):*

The same person as whom? Does it matter if a different person with the same permission completes this action?
</blockquote></details>

In this PR, AUTH_PUBLIC means that the user doesn't need to be logged in to access the endpoint (e.g. RDAP) 

When verifying the registry lock we want / need the user to be logged in to the front end

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/request/auth/RequestAuthenticator.java` line 77 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgqs-o-tOzA3A0fz9N:-N_0NqzoBXVdjqe5MVOh:bz3o62c) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/request/auth/RequestAuthenticator.java#L77)):*

yeah i kinda only meant this one because once you end up having more than one clause (which the one above does, this one doesn't) I agree that the `switch` is very possibly better. I'm good either way I think. 
</blockquote></details>

The nested IFs are not necessary, so I removed them now. I think it looks OK now.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java` line 37 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgub9CEPAduBzBeSCy:-N_0OMz39c909LOHYid9:bvhhvuv) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java#L37)):*

In this PR, AUTH_PUBLIC means that the user doesn't need to be logged in to access the endpoint (e.g. RDAP) 

When verifying the registry lock we want / need the user to be logged in to the front end 
</blockquote></details>

We have multiple level of authentications (for better or for worse). `AUTH_PUBLIC` just means that the request will not be rejected by `RequestAuthenticator` if it is not logged in. However, like other actions extending `HtmlAction`, if the request is unauthenticated, the action will redirect the user to login, before proceeding to the rest of its logic.

The reason that I changed this one from `AUTH_PUBLIC_LOGGED_IN` to `AUTH_PUBLIC` is for consistency: all other `HtmlAction` use that auth. The difference between the two is that the former will require the initial request to be logged in, otherwise it will be rejected right away. The latter, on the other hand, will attempt to login if it is not already.

So unless there is a reason to reject an un-logged-in request right away, I think it is better to maintain the consistency and make all `HtmlAction` `AUTH_PUBLIC`.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2066)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java` line 37 at r7](https://reviewable.io/reviews/google/nomulus/2066#-NZgub9CEPAduBzBeSCy:-N_0SCe6BcUNdJHGVNnQ:bbyuqb6) ([raw file](https://github.com/google/nomulus/blob/09eba1fb08619221c26fb932ea980bcf4cf4d646/core/src/main/java/google/registry/ui/server/registrar/RegistryLockVerifyAction.java#L37)):*

We have multiple level of authentications (for better or for worse). `AUTH_PUBLIC` just means that the request will not be rejected by `RequestAuthenticator` if it is not logged in. However, like other actions extending `HtmlAction`, if the request is unauthenticated, the action will redirect the user to login, before proceeding to the rest of its logic.

The reason that I changed this one from `AUTH_PUBLIC_LOGGED_IN` to `AUTH_PUBLIC` is for consistency: all other `HtmlAction` use that auth. The difference between the two is that the former will require the initial request to be logged in, otherwise it will be rejected right away. The latter, on the other hand, will attempt to login if it is not already.

So unless there is a reason to reject an un-logged-in request right away, I think it is better to maintain the consistency and make all `HtmlAction` `AUTH_PUBLIC`.
</blockquote></details>

Gotcha yeah, I see what you mean with regards to HtmlAction redirecting to the login page. Cool!

PR #2079, opened 2023-07-17
Add Console Settings -> Security front-end

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2079)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ tmp/
 local.properties
 .settings/
 .loadpath
+.DS_Store
 
 # Eclipse Core
 .project
--- a/console-webapp/dev-proxy.config.json
+++ b/console-webapp/dev-proxy.config.json
@@ -2,6 +2,8 @@
   "/console-api":
   {
     "target": "http://localhost:8080",
-    "secure": true
+    "secure": false,
+    "logLevel": "debug",
+    "changeOrigin": true
   }
 }
--- a/console-webapp/package.json
+++ b/console-webapp/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
+    "start": "ng serve --proxy-config dev-proxy.config.json",
     "build": "ng build --base-href=/console/",
     "build:local": "ng build --base-href=/default/console/",
     "watch": "ng build --watch --configuration development",
--- a/console-webapp/src/app/app.component.html
+++ b/console-webapp/src/app/app.component.html
@@ -1,6 +1,6 @@
 <div class="console-app">
   <app-header (toggleNavOpen)="sidenav.toggle()"></app-header>
-  <mat-sidenav-container class="console-app__content-wrapper">
+  <mat-sidenav-container class="console-app__container">
     <mat-sidenav #sidenav class="console-app__sidebar">
       <mat-nav-list>
         <a mat-list-item [routerLink]="'/home'" routerLinkActive="active">
@@ -17,8 +17,10 @@
         </a>
       </mat-nav-list>
     </mat-sidenav>
-    <mat-sidenav-content class="console-app__content">
-      <router-outlet></router-outlet>
+    <mat-sidenav-content class="console-app__content-wrapper">
+      <div class="console-app__content">
+        <router-outlet></router-outlet>
+      </div>
     </mat-sidenav-content>
   </mat-sidenav-container>
 </div>
--- a/console-webapp/src/app/app.component.less
+++ b/console-webapp/src/app/app.component.less
@@ -26,7 +26,7 @@
   display: flex;
   flex-direction: column;
   height: 100%;
-  &__content-wrapper {
+  &__container {
     flex: 1;
     margin-top: -12px;
     padding-bottom: 36px;
@@ -40,7 +40,11 @@
       background: #eae1e1;
     }
   }
-  &__content {
+  &__content-wrapper {
     margin: 12px 24px;
+  }
+  &__content {
+    max-width: 1340px;
+    margin: 0 auto;
   }
 }
--- a/console-webapp/src/app/app.component.spec.ts
+++ b/console-webapp/src/app/app.component.spec.ts
@@ -15,11 +15,13 @@
 import { TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 import { AppComponent } from './app.component';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { MaterialModule } from './material.module';
 
 describe('AppComponent', () => {
   beforeEach(async () => {
     await TestBed.configureTestingModule({
-      imports: [RouterTestingModule],
+      imports: [RouterTestingModule, MaterialModule, BrowserAnimationsModule],
       declarations: [AppComponent],
     }).compileComponents();
   });
--- a/console-webapp/src/app/app.module.ts
+++ b/console-webapp/src/app/app.module.ts
@@ -33,6 +33,7 @@ import SettingsContactComponent, {
 import { HttpClientModule } from '@angular/common/http';
 import { RegistrarComponent } from './registrar/registrar.component';
 import { RegistrarGuard } from './registrar/registrar.guard';
+import SecurityComponent from './settings/security/security.component';
 
 @NgModule({
   declarations: [
@@ -44,6 +45,7 @@ import { RegistrarGuard } from './registrar/registrar.guard';
     SettingsContactComponent,
     ContactDetailsDialogComponent,
     RegistrarComponent,
+    SecurityComponent,
   ],
   imports: [
     HttpClientModule,
--- a/console-webapp/src/app/header/header.component.spec.ts
+++ b/console-webapp/src/app/header/header.component.spec.ts
@@ -15,6 +15,8 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { HeaderComponent } from './header.component';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { MaterialModule } from '../material.module';
 
 describe('HeaderComponent', () => {
   let component: HeaderComponent;
@@ -22,6 +24,7 @@ describe('HeaderComponent', () => {
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
+      imports: [MaterialModule, BrowserAnimationsModule],
       declarations: [HeaderComponent],
     }).compileComponents();
 
--- a/console-webapp/src/app/registrar/registrar.component.spec.ts
+++ b/console-webapp/src/app/registrar/registrar.component.spec.ts
@@ -18,6 +18,8 @@ import { RegistrarComponent } from './registrar.component';
 import { BackendService } from '../shared/services/backend.service';
 import { ActivatedRoute } from '@angular/router';
 import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { MaterialModule } from '../material.module';
 
 describe('RegistrarComponent', () => {
   let component: RegistrarComponent;
@@ -26,7 +28,11 @@ describe('RegistrarComponent', () => {
   beforeEach(async () => {
     await TestBed.configureTestingModule({
       declarations: [RegistrarComponent],
-      imports: [HttpClientTestingModule],
+      imports: [
+        HttpClientTestingModule,
+        MaterialModule,
+        BrowserAnimationsModule,
+      ],
       providers: [
         BackendService,
         { provide: ActivatedRoute, useValue: {} as ActivatedRoute },
--- a/console-webapp/src/app/settings/contact/contact-details.component.html
+++ b/console-webapp/src/app/settings/contact/contact-details.component.html
@@ -97,7 +97,7 @@
       >
     </section>
     <mat-dialog-actions>
-      <button mat-button (click)="onClose()">Cancel</button>
+      <button mat-button (click)="onClose($event)">Cancel</button>
       <button type="submit" mat-button>Save</button>
     </mat-dialog-actions>
   </form>
--- a/console-webapp/src/app/settings/contact/contact.component.ts
+++ b/console-webapp/src/app/settings/contact/contact.component.ts
@@ -80,11 +80,11 @@ class ContactDetailsEventsResponder {
   styleUrls: ['./contact.component.less'],
 })
 export class ContactDetailsDialogComponent {
-  onClose!: Function;
   contact: Contact;
   contactTypes = contactTypes;
   operation: Operations;
   contactIndex: number;
+  onCloseCallback: Function;
 
   constructor(
     public contactService: ContactService,
@@ -96,7 +96,7 @@ export class ContactDetailsDialogComponent {
       operation: Operations;
     }
   ) {
-    this.onClose = data.onClose;
+    this.onCloseCallback = data.onClose;
     this.contactIndex = contactService.contacts.findIndex(
       (c) => c === data.contact
     );
@@ -104,9 +104,14 @@ export class ContactDetailsDialogComponent {
     this.operation = data.operation;
   }
 
-  saveAndClose(e: any) {
+  onClose(e: MouseEvent) {
     e.preventDefault();
-    if (!e.target.checkValidity()) {
+    this.onCloseCallback.call(this);
+  }
+
+  saveAndClose(e: SubmitEvent) {
+    e.preventDefault();
+    if (!(e.target as HTMLFormElement).checkValidity()) {
       return;
     }
     let operationObservable;
@@ -122,7 +127,7 @@ export class ContactDetailsDialogComponent {
     }
 
     operationObservable.subscribe({
-      complete: this.onClose.bind(this),
+      complete: this.onCloseCallback.bind(this),
       error: (err: HttpErrorResponse) => {
         this._snackBar.open(err.statusText, undefined, {
           duration: 1500,
@@ -169,7 +174,7 @@ export default class ContactComponent {
     }
   }
 
-  openCreateNew(e: Event) {
+  openCreateNew(e: MouseEvent) {
     const newContact: Contact = {
       name: '',
       phoneNumber: '',
@@ -180,7 +185,7 @@ export default class ContactComponent {
   }
 
   openDetails(
-    e: Event,
+    e: MouseEvent,
     contact: Contact,
     operation: Operations = Operations.UPDATE
   ) {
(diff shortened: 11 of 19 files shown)


gbrodman (changes requested):
> Reviewed 10 of 13 files at r1, 1 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2079)* status: all files reviewed, 9 unresolved discussions (waiting on @ptkach)

___
*[`console-webapp/dev-proxy.config.json` line 4 at r2](https://reviewable.io/reviews/google/nomulus/2079#-N_eFPuj6rUV9lk3SrwY:-N_eFPuj6rUV9lk3SrwZ:b-gmbsda) ([raw file](https://github.com/google/nomulus/blob/ec41c6d1fd475ccae7ecb1caf8350a6dd0a5c51d/console-webapp/dev-proxy.config.json#L4)):*
> ```JSON
>   "/console-api":
>   {
>     "target": "https://domain-registry-alpha.appspot.com",
> ```

hm i'm not sure we want to push this if we're usually developing everything against the local dev server? 

also, technically this won't work for other open-source users 
___
*[`console-webapp/src/app/settings/contact/contact.component.ts` line 107 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_ipx1-5t7OzA0UiGjn:-N_ipx100Cue8Y-7kdHK:btacek) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/contact/contact.component.ts#L107)):*
> ```XML
>   }
> 
>   onClose(e: any) {
> ```

this (and the other event methods) should probably have an actual type of the event -- I'm not sure what that is, but it seems like there are some kinda-generic Event types in the Typescript spec that correspond to DOM events. Have you tried to figure out exactly what type this should be? 
___
*[`console-webapp/src/app/settings/security/security.component.html` line 7 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_j8Ge06C86M28WLAYJ:-N_j8Ge06C86M28WLAYK:b8c4502) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L7)):*
> ```HTML
>   <div class="settings-security__section">
>     <div class="settings-security__section-description">
>       <h2>IP Allow Whitelist</h2>
> ```

technically we'll want to use the term `IP Allowlist`, even if it's a bit awkward 
___
*[`console-webapp/src/app/settings/security/security.component.html` line 16 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jAsbw8W22pPxlsg7M:-N_jAsbw8W22pPxlsg7N:bci911e) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L16)):*
> ```HTML
>       <div
>         *ngIf="
>           dataSource.ipAddressAllowList &&
> ```

Do we need to check for both null/undefined and an empty list? What happens if the list is empty? and we don't have this check? 
___
*[`console-webapp/src/app/settings/security/security.component.html` line 18 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jAQq6Bw50qQAwspJZ:-N_jAQq6Bw50qQAwspJ_:b-91z1iz) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L18)):*
> ```HTML
>           dataSource.ipAddressAllowList &&
>           dataSource.ipAddressAllowList.length > 0
>         "
> ```

out of curiosity, does the formatter add these newlines? 
___
*[`console-webapp/src/app/settings/security/security.component.html` line 33 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jBIw20gjD3tJ3aXx-:-N_jBIw37Tw6RSj4w5aS:b-7t9g4q) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L33)):*
> ```HTML
>               mat-icon-button
>               aria-label="Remove"
>               (click)="removeIpEntry(index)"
> ```

out of curiosity did you give any thought to removal of the IP address by index (what you're doing here) vs by value? I assume that there's some way of saying "remove this element from this list" 
___
*[`console-webapp/src/app/settings/security/security.service.ts` line 50 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jE_tw11Tn8kSSI70Y:-N_jEKGwCZcVvekIqu4d:b-e25y6y) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.ts#L50)):*
> ```XML
>         tap((securitySettings: SecuritySettingsBackendModel) => {
>           if(securitySettings) {
>             this.securitySettings = Object.assign({}, securitySettings, {
> ```

What does the `Object.assign` bit do? It seems to me like this is converting the response from the API into the form we want, is that right / why is that necessary? 
___
*[`console-webapp/src/app/settings/security/security.service.ts` line 72 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jF-S14SU_BgsZT-7s:-N_jF-S14SU_BgsZT-7t:blke4a0) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.ts#L72)):*
> ```XML
>     )
>     .pipe(
>       tap((_) => {
> ```

why are we using `tap` if we're not using the value? I thought that `tap` was essentially for doing something with a value as we pass it through
___
*[`console-webapp/src/app/settings/security/security.service.spec.ts` line 14 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jCT7o2PtM-EQzl4eW:-N_jCT7o2PtM-EQzl4eX:bqlkgzn) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.spec.ts#L14)):*
> ```XML
> // See the License for the specific language governing permissions and
> ```

Did you mention wanting to write tests for this, or are you thinking of saving that for the future?
___
*[`console-webapp/src/app/shared/services/backend.service.ts` line 67 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jCpfJ4sRc1UylMW4D:-N_jCpfJ4sRc1UylMW4E:bkswj5f) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/shared/services/backend.service.ts#L67)):*
> ```XML
>       .pipe(catchError((err) => this.errorCatcher<string[]>(err)));
>   }
> ```

at some point should we consider retrieving the entire registrar object once when the user selects it, then just passing around the required bits of data? I think I remember you mentioning it, but to be honest if it doesn't create problems to retrieve bits at a time it's probably safer to do it this way.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2079)* status: 8 of 19 files reviewed, 9 unresolved discussions (waiting on @gbrodman)

___
*[`console-webapp/dev-proxy.config.json` line 4 at r2](https://reviewable.io/reviews/google/nomulus/2079#-N_eFPuj6rUV9lk3SrwY:-N_ncs9RCrEY8IyEm1tb:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ec41c6d1fd475ccae7ecb1caf8350a6dd0a5c51d/console-webapp/dev-proxy.config.json#L4)):*

hm i'm not sure we want to push this if we're usually developing everything against the local dev server? 

also, technically this won't work for other open-source users 
</blockquote></details>

Done.
___
*[`console-webapp/src/app/settings/contact/contact.component.ts` line 107 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_ipx1-5t7OzA0UiGjn:-N_nctDJ7p2GWaQ2dNyJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/contact/contact.component.ts#L107)):*

this (and the other event methods) should probably have an actual type of the event -- I'm not sure what that is, but it seems like there are some kinda-generic Event types in the Typescript spec that correspond to DOM events. Have you tried to figure out exactly what type this should be? 
</blockquote></details>

Done.
___
*[`console-webapp/src/app/settings/security/security.component.html` line 7 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_j8Ge06C86M28WLAYJ:-N_ncuCb0aZqcbo-M8py:b-896fix) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L7)):*

technically we'll want to use the term `IP Allowlist`, even if it's a bit awkward 
</blockquote></details>

Done.
___
*[`console-webapp/src/app/settings/security/security.component.html` line 16 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jAsbw8W22pPxlsg7M:-N_nY_NyEqcC9LObOYnG:b1w0qdm) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L16)):*

Do we need to check for both null/undefined and an empty list? What happens if the list is empty? and we don't have this check? 
</blockquote></details>

I wasn't going to check for empty list, but typescript just won't let it compile without this check, so I had to add it. 
___
*[`console-webapp/src/app/settings/security/security.component.html` line 18 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jAQq6Bw50qQAwspJZ:-N_p4Td50WiTVeluSuK6:b2nhswy) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L18)):*

out of curiosity, does the formatter add these newlines? 
</blockquote></details>

Yes
___
*[`console-webapp/src/app/settings/security/security.component.html` line 33 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jBIw20gjD3tJ3aXx-:-N_nZ6Mn7WUcJJun2yW5:b7h6nhi) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.component.html#L33)):*

out of curiosity did you give any thought to removal of the IP address by index (what you're doing here) vs by value? I assume that there's some way of saying "remove this element from this list" 
</blockquote></details>

There's no such method in javascript to remove element from the list, `splice` does it by index and it's mutable. Generally `filter` is considered to be a way around it, which you can use with either index or reference. To me index seems more intuitive. 
___
*[`console-webapp/src/app/settings/security/security.service.ts` line 50 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jE_tw11Tn8kSSI70Y:-N_n_67B25Ql3QxbwFWB:b-vpj3vi) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.ts#L50)):*

What does the `Object.assign` bit do? It seems to me like this is converting the response from the API into the form we want, is that right / why is that necessary? 
</blockquote></details>

Correct .`Object.assign` converts list of strings (API) to list of objects (Frontend) and back. It's necessary because from Java side the `ipAddressAllowList` is naturally and historically a list of strings, however Angular binding doesn't work well with list of strings, causing rerenders and focus being constantly lost on the field. Making a simple conversion seemed like the easiest solution. 
___
*[`console-webapp/src/app/settings/security/security.service.ts` line 72 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jF-S14SU_BgsZT-7s:-N_naCac3nM1cED07t2k:b-k5gh6u) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.ts#L72)):*

why are we using `tap` if we're not using the value? I thought that `tap` was essentially for doing something with a value as we pass it through
</blockquote></details>

We're "doing something with a value" - we're saving it. `tap` is used for any side effect in observable, which is exactly our case here. 
___
*[`console-webapp/src/app/settings/security/security.service.spec.ts` line 14 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jCT7o2PtM-EQzl4eW:-N_p441L5sSMZ0rLKzQH:b-1yzgki) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.spec.ts#L14)):*

Did you mention wanting to write tests for this, or are you thinking of saving that for the future?
</blockquote></details>

I'm working on tests in parallel, but I have a feeling that at some point, probably as soon as old console is reimplemented, we will revisit the tests topic
___
*[`console-webapp/src/app/shared/services/backend.service.ts` line 67 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jCpfJ4sRc1UylMW4D:-N_nbz3J1yIVMgRVC_-H:b-wo4otd) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/shared/services/backend.service.ts#L67)):*

at some point should we consider retrieving the entire registrar object once when the user selects it, then just passing around the required bits of data? I think I remember you mentioning it, but to be honest if it doesn't create problems to retrieve bits at a time it's probably safer to do it this way.
</blockquote></details>

That's absolutely correct, my initial idea was to retrieve registrar info entirely and work with it on frontend, however the more I thought about it during both frontend and backend implementations I realized we'd be better off dealing with parts of the object instead.

ptkach:
@gbrodman  Please let me know if you have any suggestions here, as I believe it's in a good shape now and I'd like to get it merged. Thanks!

1proprogrammerchant:
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/2079)_ status: 7 of 19 files reviewed, 13 unresolved discussions (waiting on @gbrodman and @github-code-scanning[bot])
> 
> _[`console-webapp/src/app/settings/security/security.component.spec.ts` line 30 at r5](https://reviewable.io/reviews/google/nomulus/2079#gh-1270884364:-Na84Xt17RrM6svivt29:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f6b63f38b046328fe8658c5efc8b14470dfe2c1/console-webapp/src/app/settings/security/security.component.spec.ts#L30)):_
> 
> Previously, github-code-scanning[bot] wrote…
> Done.
> 
> _[`console-webapp/src/app/settings/security/security.service.spec.ts` line 19 at r5](https://reviewable.io/reviews/google/nomulus/2079#gh-1270884366:-Na84fntEaEsOagG42av:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f6b63f38b046328fe8658c5efc8b14470dfe2c1/console-webapp/src/app/settings/security/security.service.spec.ts#L19)):_
> 
> Previously, github-code-scanning[bot] wrote…
> Done.
> 
> _[`console-webapp/src/app/settings/security/security.service.spec.ts` line 20 at r5](https://reviewable.io/reviews/google/nomulus/2079#gh-1270884372:-Na84gGdDy8QIO8mFfCd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f6b63f38b046328fe8658c5efc8b14470dfe2c1/console-webapp/src/app/settings/security/security.service.spec.ts#L20)):_
> 
> Previously, github-code-scanning[bot] wrote…
> Done.
> 
> _[`console-webapp/src/app/settings/security/security.service.spec.ts` line 21 at r5](https://reviewable.io/reviews/google/nomulus/2079#gh-1270884375:-Na84gjk1HbNGTAz6Q3V:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0f6b63f38b046328fe8658c5efc8b14470dfe2c1/console-webapp/src/app/settings/security/security.service.spec.ts#L21)):_
> 
> Previously, github-code-scanning[bot] wrote…
> Done. 
this is a nice change

gbrodman (changes requested):
> Reviewed 8 of 11 files at r4, 4 of 4 files at r6, all commit messages.
> Dismissed @gbrodman from a discussion.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2079)* status: all files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

___
*[`console-webapp/src/app/settings/security/security.component.spec.ts` line 151 at r6](https://reviewable.io/reviews/google/nomulus/2079#-NaDkcu8AO59uoXbOIEl:-NaDkcu8AO59uoXbOIEm:b-7m9pri) ([raw file](https://github.com/google/nomulus/blob/4dc55341a707eaf892b05ef73980acd61b6debe8/console-webapp/src/app/settings/security/security.component.spec.ts#L151)):*
> ```XML
>       .querySelector('.settings-security__actions-save')
>       .click();
>     expect(saveSpy).toHaveBeenCalledOnceWith({
> ```

neat!
___
*[`console-webapp/src/app/settings/security/security.service.ts` line 72 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jF-S14SU_BgsZT-7s:-NaDf4Ao24J7hvNFAAdC:bo8jrhz) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.ts#L72)):*

We're "doing something with a value" - we're saving it. `tap` is used for any side effect in observable, which is exactly our case here. 
</blockquote></details>

Maybe I'm missing something but why would we do this instead of just calling subscribe? 
___
*[`console-webapp/src/app/shared/services/backend.service.ts` line 67 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jCpfJ4sRc1UylMW4D:-NaDgLr48N-RFUUJkhty:b-4u9l00) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/shared/services/backend.service.ts#L67)):*

That's absolutely correct, my initial idea was to retrieve registrar info entirely and work with it on frontend, however the more I thought about it during both frontend and backend implementations I realized we'd be better off dealing with parts of the object instead. 
</blockquote></details>

yeah that's fair, just noting what we talked about where if we get to the point where there's more than a couple different little bits we're dealing with we might want to end up just using the whole registrar

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2079)* status: all files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot])

___
*[`console-webapp/src/app/settings/security/security.service.ts` line 72 at r3](https://reviewable.io/reviews/google/nomulus/2079#-N_jF-S14SU_BgsZT-7s:-NaI3LC95paXS8vzMX_c:buj1z7n) ([raw file](https://github.com/google/nomulus/blob/739207431855e0f5d0127e1794c79c541098f99d/console-webapp/src/app/settings/security/security.service.ts#L72)):*

Because we have 2 separate things we need to do based on the completion of the the observer:
1) Stop loading indicator - that's a view responsibility to deal with view-related logic
2) Save data - that's a service responsibility to deal with the data

Given that there's only a single subscriber allowed on a observer I don't think we have a better way to deal with this logic. 

</blockquote></details>

got it, yeah i didn't realize that we were returning an observable that someone else subscribed to -- in that case, this makes complete sense

PR #2080, opened 2023-07-18
Allow EPP password to be changed during login flow

CydeWeys:
This is part of the spec in RFC 5730 that we hadn't implemented until now. Note that this requires changing LoginFlow to be transactional, but I don't think that should cause any issues.

See https://b.corp.google.com/issues/291759824

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2080)

### DIFF
--- a/core/src/main/java/google/registry/flows/FlowRunner.java
+++ b/core/src/main/java/google/registry/flows/FlowRunner.java
@@ -72,21 +72,20 @@ public class FlowRunner {
     }
     eppMetricBuilder.setCommandNameFromFlow(flowClass.getSimpleName());
     if (!isTransactional) {
-      EppOutput eppOutput = EppOutput.create(flowProvider.get().run());
-      if (flowClass.equals(LoginFlow.class)) {
-        // In LoginFlow, registrarId isn't known until after the flow executes, so save it then.
-        eppMetricBuilder.setRegistrarId(sessionMetadata.getRegistrarId());
-      }
-      return eppOutput;
+      return EppOutput.create(flowProvider.get().run());
     }
     try {
-      return tm()
-          .transact(
+      return tm().transact(
               () -> {
                 try {
                   EppOutput output = EppOutput.create(flowProvider.get().run());
                   if (isDryRun) {
                     throw new DryRunException(output);
+                  }
+                  if (flowClass.equals(LoginFlow.class)) {
+                    // In LoginFlow, registrarId isn't known until after the flow executes, so save
+                    // it then.
+                    eppMetricBuilder.setRegistrarId(sessionMetadata.getRegistrarId());
                   }
                   return output;
                 } catch (EppException e) {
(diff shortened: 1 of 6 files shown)


CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2080)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/flows/session/LoginFlow.java` line 142 at r1](https://reviewable.io/reviews/google/nomulus/2080#-Na8VmO23HHUDQvjdnDC:-Na8uXcR6fAdZ7RjUFiS:b-heifa0) ([raw file](https://github.com/google/nomulus/blob/08978ce96558732c1c434b91c00624a5b52ef7ea/core/src/main/java/google/registry/flows/session/LoginFlow.java#L142)):*

I know that setPassword() already checks that this is a valid password and throws an IllegalArgumentException, but shouldn't we add a catch that throws an EPP exception to explain to the registrar why this would fail if the password is bad.
</blockquote></details>

The password requirements are actually enforced in the XSD, which requires a token-type string consisting of 6-16 characters:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/xml/xsd/epp.xsd;l=242-247;drc=a7a983bfed42634985176141b3a073350eb97267

So the incoming EPP will fail to marshal, rather than the error being caught in the setter.  The restrictions in the setter just affect the `nomulus` command line tool, which already have tests.  The newly added test I've added should help to make this clear.

And note that the password requirements are super lenient as defined in the EPP RFC.  They basically amount to "anything that is a valid XML token (i.e. will parse correctly) with the length of 6-16 characters".

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2080)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/test/java/google/registry/flows/session/LoginFlowTestCase.java` line 87 at r1](https://reviewable.io/reviews/google/nomulus/2080#-Na8b-I8BzReq8xsx9JA:-Na9-XFv60tyalT6U1Tu:bs5y37s) ([raw file](https://github.com/google/nomulus/blob/08978ce96558732c1c434b91c00624a5b52ef7ea/core/src/test/java/google/registry/flows/session/LoginFlowTestCase.java#L87)):*

Why not just do this in this PR?
</blockquote></details>

I had meant to, and this TODO was left for myself in the context of finishing this PR, but then over the weekend I clearly just forgot about it :/   Anyway, done now.

CydeWeys (commented):
PTAL

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2080#-:-Na9-er_3ER_Cr9Q-s9_:bb71pro)* status: 3 of 6 files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

PR #2081, opened 2023-07-21
Fix ICANN reporting and add rdap-queries field

gbrodman:
This includes two changes, the second necessary for testing the first.
1. We add the rdap-queries field as mandated by the amendment to the registry agreement, https://itp.cdn.icann.org/en/files/registry-agreement/proposed-global-amendment-base-gtld-registry-agreement-12-04-2023-en.pdf. This is fairly similar to the whois-queries field where we just query the logs, but instead of searching for "whois" we search for "rdap".
2. BigQuery doesn't use MAX to refer to the bigger of two fields; MAX accepts an array as an argument. In order to do what we want (and to have the BigQuery statements succeed), we need to use GREATEST. Tested both versions in alpha and production BigQuery instances.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2081)


jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2081)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #2086, opened 2023-07-25
 Fix time inversion error when writing metrics 

jianglai:
The instance ID used to be uniquely determined by App Engine SDK. Since         
we no longer calls the SDK, we need a way to distinguish instances so           
that their metrics would not stump on each other and result in a time           
inversion error (as we have seen frequently in the logs since the               
removal of the App Engine SDK).

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2086)

### DIFF
--- a/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java
+++ b/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java
@@ -25,17 +25,31 @@ import dagger.Module;
 import dagger.Provides;
 import google.registry.config.CredentialModule.ApplicationDefaultCredential;
 import google.registry.config.RegistryConfig.Config;
+import google.registry.util.Clock;
 import google.registry.util.GoogleCredentialsBundle;
+import javax.inject.Named;
+import javax.inject.Singleton;
 import org.joda.time.Duration;
 
 /** Dagger module for Google Stackdriver service connection objects. */
 @Module
 public final class StackdriverModule {
 
+  private StackdriverModule() {}
+
   // We need a fake GCE zone to appease Stackdriver's resource model.
   // TODO(b/265973059): Switch to resource type "gke_container".
   private static final String SPOOFED_GCE_ZONE = "us-central1-f";
-  private static final String SPOOFED_GCE_INSTANCE = "fake-instance";
+
+  // We cannot use a static fake intance ID which is shared by all instances, because metrics might
+  // be flushed to stackdriver with delays, which lead to time inversion erros when another instance
+  // has already written a data point at a later time.
+  @Singleton
+  @Provides
+  @Named("spoofedGceInstanceId")
+  static String providesSpoofedGceInstanceId(Clock clock) {
+    return clock.nowUtc().toString();
+  }
 
   @Provides
   static Monitoring provideMonitoring(
@@ -54,7 +68,8 @@ public final class StackdriverModule {
       Monitoring monitoringClient,
       @Config("projectId") String projectId,
       @Config("stackdriverMaxQps") int maxQps,
-      @Config("stackdriverMaxPointsPerRequest") int maxPointsPerRequest) {
+      @Config("stackdriverMaxPointsPerRequest") int maxPointsPerRequest,
+      @Named("spoofedGceInstanceId") String instanceId) {
     // The MonitoredResource for GAE apps is not writable (and missing fields anyway) so we just
     // use the gce_instance resource type instead.
     return new StackdriverWriter(
@@ -65,7 +80,7 @@ public final class StackdriverModule {
             .setLabels(
                 ImmutableMap.of(
                     // The "zone" field MUST be a valid GCE zone, so we fake one.
-                    "zone", SPOOFED_GCE_ZONE, "instance_id", SPOOFED_GCE_INSTANCE)),
+                    "zone", SPOOFED_GCE_ZONE, "instance_id", instanceId)),
         maxQps,
         maxPointsPerRequest);
   }


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2086)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java` line 47 at r1](https://reviewable.io/reviews/google/nomulus/2086#-NaD05syBh2na0aPWtHo:-NaD05syBh2na0aPWtHp:b2ni3xx) ([raw file](https://github.com/google/nomulus/blob/bba8b3a7573717b18d5ae63c06f37efb81ef9448/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java#L47)):*
> ```Java
>   // We cannot use a static fake intance ID which is shared by all instances, because metrics might
>   // be flushed to stackdriver with delays, which lead to time inversion erros when another instance
>   // had already written a data point at a later time.
> ```

has
___
*[`core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java` line 50 at r1](https://reviewable.io/reviews/google/nomulus/2086#-NaD0MiQ011brNnEC2-j:-NaD0MiR9-5VVqkBssP-:b-fmb1wp) ([raw file](https://github.com/google/nomulus/blob/bba8b3a7573717b18d5ae63c06f37efb81ef9448/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java#L50)):*
> ```Java
>   @Singleton
>   @Provides
>   @Named("spoofedGceInstance")
> ```

It's not just "Instance", it's "InstanceId"
___
*[`core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java` line 52 at r1](https://reviewable.io/reviews/google/nomulus/2086#-NaD0BSW5po9psQirXO9:-NaD0BSW5po9psQirXOA:beww3is) ([raw file](https://github.com/google/nomulus/blob/bba8b3a7573717b18d5ae63c06f37efb81ef9448/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java#L52)):*
> ```Java
>   @Named("spoofedGceInstance")
>   static String providesSpoofedGceInstanceId(Random random, Clock clock) {
>     return String.format("%s-%d", clock.nowUtc(), random.nextInt());
> ```

So it's ISO 8601 time and then a random int?  Why like this?  Does the time need to be in it at all?  And given that the random int is the fake instance number, shouldn't that be first?

CydeWeys (commented):
The PR title needs work.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2086#-:-NaD0VNH8ZELxYBk7Fl_:bu7dgy0)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @jianglai)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2086)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java` line 52 at r1](https://reviewable.io/reviews/google/nomulus/2086#-NaD0BSW5po9psQirXO9:-NaD9opGEN_jVHGxKjYX:b-vxj26x) ([raw file](https://github.com/google/nomulus/blob/bba8b3a7573717b18d5ae63c06f37efb81ef9448/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java#L52)):*

When we plot the time series, we aggregate the data points across all instance IDs. So the format of the fake ID does not matter much. But I think if for whatever reason we do want to look at the metrics just for a given instance, it's much easier to filter the instance by the time period during which it might be live. Hence why I decided to start the fake ID with the timestamp.
</blockquote></details>

But how will you filter the metrics to just those for a given instance?  Isn't that what the instance ID is for?  And here the instance ID contains a big datetime at the beginning of it, making everything effectively unique?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2086)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java` line 52 at r1](https://reviewable.io/reviews/google/nomulus/2086#-NaD0BSW5po9psQirXO9:-NaDAAwt6TG0pGbNi4c4:bvgws7e) ([raw file](https://github.com/google/nomulus/blob/bba8b3a7573717b18d5ae63c06f37efb81ef9448/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java#L52)):*

But how will you filter the metrics to just those for a given instance?  Isn't that what the instance ID is for?  And here the instance ID contains a big datetime at the beginning of it, making everything effectively unique?
</blockquote></details>

Yes, every instance will be unique. They were unique before we changed them to all be "fake-instance":

```
        new MonitoredResource()
            .setType("gce_instance")
            .setLabels(
                ImmutableMap.of(
                    // The "zone" field MUST be a valid GCE zone, so we fake one.
                    "zone",
                    SPOOFED_GCE_ZONE,
                    // Overload the GCE "instance_id" field with the GAE module name, version and
                    // instance_id.
                    "instance_id",
                    modulesService.getCurrentModule()
                        + ":"
                        + modulesService.getCurrentVersion()
                        + ":"
                        + modulesService.getCurrentInstanceId())),
```

The point is that by having some more or less descriptive instance ID, its easier to find the ones that are likely relevant, instead of just seeing a bunch of random numbers. Because this value is set when the first metric of an instance is written (or when the instance starts up, depending on if the injection is lazy), you at least have an idea when approximately was the instance spun up, like here:

CydeWeys (dismissed):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2086)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java` line 52 at r1](https://reviewable.io/reviews/google/nomulus/2086#-NaD0BSW5po9psQirXO9:-NaE70FS88rjRAWLmvJH:bc3irq6) ([raw file](https://github.com/google/nomulus/blob/bba8b3a7573717b18d5ae63c06f37efb81ef9448/core/src/main/java/google/registry/monitoring/whitebox/StackdriverModule.java#L52)):*

Yes, every instance will be unique. They were unique before we changed them to all be "fake-instance":

```
        new MonitoredResource()
            .setType("gce_instance")
            .setLabels(
                ImmutableMap.of(
                    // The "zone" field MUST be a valid GCE zone, so we fake one.
                    "zone",
                    SPOOFED_GCE_ZONE,
                    // Overload the GCE "instance_id" field with the GAE module name, version and
                    // instance_id.
                    "instance_id",
                    modulesService.getCurrentModule()
                        + ":"
                        + modulesService.getCurrentVersion()
                        + ":"
                        + modulesService.getCurrentInstanceId())),
```

The point is that by having some more or less descriptive instance ID, its easier to find the ones that are likely relevant, instead of just seeing a bunch of random numbers. Because this value is set when the first metric of an instance is written (or when the instance starts up, depending on if the injection is lazy), you at least have an idea when approximately was the instance spun up, like here:


 
</blockquote></details>

Ahaahhhh, got it, thank you.  That screenshot made it clear.

It now occurs to me that we don't need the random integer at all.  Just stick with the `DateTime`.  It's exceedingly unlikely that two instances will generate this at the exact same millisecond.  If you're worried about collisions still, then go down to microseconds (which is still valid ISO 8601).

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2086)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

PR #2095, opened 2023-08-04
Update /console-api/registrars endpoint with POST and GET

ptkach:
This changes the response with the following:

1. Adds POST to create a registrar
2. Updates GET to provide the list of entire registrar objects instead of list of ids

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2095)

### DIFF
--- a/core/src/main/java/google/registry/model/adapters/CurrencyJsonAdapter.java
+++ b/core/src/main/java/google/registry/model/adapters/CurrencyJsonAdapter.java
@@ -1,0 +1,41 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.model.adapters;
+
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import google.registry.model.adapters.CurrencyUnitAdapter.UnknownCurrencyException;
+import java.io.IOException;
+import org.joda.money.CurrencyUnit;
+
+public class CurrencyJsonAdapter extends TypeAdapter<CurrencyUnit> {
+
+  @Override
+  public void write(JsonWriter out, CurrencyUnit value) throws IOException {
+    String currency = CurrencyUnitAdapter.convertFromCurrency(value);
+    out.value(currency);
+  }
+
+  @Override
+  public CurrencyUnit read(JsonReader in) throws IOException {
+    String currency = in.nextString();
+    try {
+      return CurrencyUnitAdapter.convertFromString(currency);
+    } catch (UnknownCurrencyException e) {
+      throw new IOException("Unknown currency");
+    }
+  }
+}
--- a/core/src/main/java/google/registry/model/adapters/CurrencyUnitAdapter.java
+++ b/core/src/main/java/google/registry/model/adapters/CurrencyUnitAdapter.java
@@ -1,4 +1,4 @@
-// Copyright 2017 The Nomulus Authors. All Rights Reserved.
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,9 +22,7 @@ import org.joda.money.CurrencyUnit;
 /** Adapter to use Joda {@link CurrencyUnit} when marshalling strings. */
 public class CurrencyUnitAdapter extends XmlAdapter<String, CurrencyUnit> {
 
-  /** Parses a string into a {@link CurrencyUnit} object. */
-  @Override
-  public CurrencyUnit unmarshal(String currency) throws UnknownCurrencyException {
+  public static CurrencyUnit convertFromString(String currency) throws UnknownCurrencyException {
     try {
       return CurrencyUnit.of(nullToEmpty(currency).trim());
     } catch (IllegalArgumentException e) {
@@ -32,10 +30,20 @@ public class CurrencyUnitAdapter extends XmlAdapter<String, CurrencyUnit> {
     }
   }
 
+  public static String convertFromCurrency(CurrencyUnit currency) {
+    return currency == null ? null : currency.toString();
+  }
+
+  /** Parses a string into a {@link CurrencyUnit} object. */
+  @Override
+  public CurrencyUnit unmarshal(String currency) throws UnknownCurrencyException {
+    return convertFromString(currency);
+  }
+
   /** Converts {@link CurrencyUnit} to a string. */
   @Override
   public String marshal(CurrencyUnit currency) {
-    return currency == null ? null : currency.toString();
+    return convertFromCurrency(currency);
   }
 
   /** Exception to throw when failing to parse a currency. */
--- a/core/src/main/java/google/registry/model/eppcommon/Address.java
+++ b/core/src/main/java/google/registry/model/eppcommon/Address.java
@@ -20,6 +20,7 @@ import static google.registry.util.CollectionUtils.nullToEmptyImmutableCopy;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.gson.annotations.Expose;
 import google.registry.model.Buildable;
 import google.registry.model.ImmutableObject;
 import google.registry.model.JsonMapBuilder;
@@ -87,6 +88,7 @@ public class Address extends ImmutableObject implements Jsonifiable, UnsafeSeria
    */
   @XmlJavaTypeAdapter(NormalizedStringAdapter.class)
   @Transient
+  @Expose
   protected List<String> street;
 
   @XmlTransient @IgnoredInDiffableMap protected String streetLine1;
@@ -96,18 +98,22 @@ public class Address extends ImmutableObject implements Jsonifiable, UnsafeSeria
   @XmlTransient @IgnoredInDiffableMap protected String streetLine3;
 
   @XmlJavaTypeAdapter(NormalizedStringAdapter.class)
+  @Expose
   protected String city;
 
   @XmlElement(name = "sp")
   @XmlJavaTypeAdapter(NormalizedStringAdapter.class)
+  @Expose
   protected String state;
 
   @XmlElement(name = "pc")
   @XmlJavaTypeAdapter(CollapsedStringAdapter.class)
+  @Expose
   protected String zip;
 
   @XmlElement(name = "cc")
   @XmlJavaTypeAdapter(CollapsedStringAdapter.class)
+  @Expose
   protected String countryCode;
 
   public ImmutableList<String> getStreet() {
--- a/core/src/main/java/google/registry/model/registrar/Registrar.java
+++ b/core/src/main/java/google/registry/model/registrar/Registrar.java
@@ -211,6 +211,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    */
   @Id
   @Column(nullable = false)
+  @Expose
   String registrarId;
 
   /**
@@ -224,6 +225,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    * @see <a href="http://www.icann.org/registrar-reports/accredited-list.html">ICANN-Accredited
    *     Registrars</a>
    */
+  @Expose
   @Column(nullable = false)
   String registrarName;
 
@@ -237,7 +239,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   State state;
 
   /** The set of TLDs which this registrar is allowed to access. */
-  Set<String> allowedTlds;
+  @Expose Set<String> allowedTlds;
 
   /** Host name of WHOIS server. */
   String whoisServer;
@@ -287,6 +289,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    * unrestricted UTF-8.
    */
   @Embedded
+  @Expose
   @AttributeOverrides({
     @AttributeOverride(
         name = "streetLine1",
@@ -329,7 +332,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   String faxNumber;
 
   /** Email address. */
-  String emailAddress;
+  @Expose String emailAddress;
 
   // External IDs.
 
@@ -345,7 +348,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    * @see <a href="http://www.iana.org/assignments/registrar-ids/registrar-ids.txt">Registrar
    *     IDs</a>
    */
-  @Nullable Long ianaIdentifier;
+  @Expose @Nullable Long ianaIdentifier;
 
   /** Purchase Order number used for invoices in external billing system, if applicable. */
   @Nullable String poNumber;
@@ -358,7 +361,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    * accessed by {@link #getBillingAccountMap}, a sorted map is returned to guarantee deterministic
    * behavior when serializing the map, for display purpose for instance.
    */
-  @Nullable Map<CurrencyUnit, String> billingAccountMap;
+  @Expose @Nullable Map<CurrencyUnit, String> billingAccountMap;
 
   /** URL of registrar's website. */
   String url;
@@ -369,10 +372,10 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    * <p>This value is specified in the initial registrar contact. It can't be edited in the web GUI,
    * and it must be specified when the registrar account is created.
    */
-  String icannReferralEmail;
+  @Expose String icannReferralEmail;
 
   /** Id of the folder in drive used to publish information for this registrar. */
-  String driveFolderId;
+  @Expose String driveFolderId;
 
   // Metadata.
 
@@ -400,7 +403,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   boolean contactsRequireSyncing = true;
 
   /** Whether or not registry lock is allowed for this registrar. */
-  boolean registryLockAllowed = false;
+  @Expose boolean registryLockAllowed = false;
 
   public String getRegistrarId() {
     return registrarId;
--- a/core/src/main/java/google/registry/request/RequestModule.java
+++ b/core/src/main/java/google/registry/request/RequestModule.java
@@ -31,21 +31,28 @@ import com.google.common.io.ByteStreams;
 import com.google.common.io.CharStreams;
 import com.google.common.net.MediaType;
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
 import com.google.protobuf.ByteString;
 import dagger.Module;
 import dagger.Provides;
+import google.registry.model.adapters.CurrencyJsonAdapter;
 import google.registry.request.HttpException.BadRequestException;
 import google.registry.request.HttpException.UnsupportedMediaTypeException;
 import google.registry.request.auth.AuthResult;
 import google.registry.request.lock.LockHandler;
 import google.registry.request.lock.LockHandlerImpl;
+import google.registry.util.CidrAddressBlock;
+import google.registry.util.CidrAddressBlock.CidrAddressBlockAdapter;
+import google.registry.util.DateTimeTypeAdapter;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Optional;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
+import org.joda.money.CurrencyUnit;
+import org.joda.time.DateTime;
 import org.json.simple.JSONValue;
 import org.json.simple.parser.ParseException;
 
@@ -67,6 +74,18 @@ public final class RequestModule {
     this.req = req;
     this.rsp = rsp;
     this.authResult = authResult;
+  }
+
+  @RequestScope
+  @VisibleForTesting
+  @Provides
+  public static Gson provideGson() {
+    return new GsonBuilder()
+        .registerTypeAdapter(DateTime.class, new DateTimeTypeAdapter())
+        .registerTypeAdapter(CidrAddressBlock.class, new CidrAddressBlockAdapter())
+        .registerTypeAdapter(CurrencyUnit.class, new CurrencyJsonAdapter())
+        .excludeFieldsWithoutExposeAnnotation()
+        .create();
   }
 
   @Provides
--- a/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java
@@ -14,7 +14,11 @@
 
 package google.registry.ui.server.console;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.request.Action.Method.GET;
+import static google.registry.request.Action.Method.POST;
 
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.common.collect.ImmutableList;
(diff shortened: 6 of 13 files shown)


gbrodman (changes requested):
> Reviewed 11 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: 11 of 12 files reviewed, 10 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGfWrC9zn9g2va9DXC:-NbGfWrC9zn9g2va9DXD:ba6e58w) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L64)):*
> ```Java
>   }
> 
>   public static class RegistrarAddressAdapter extends TypeAdapter<RegistrarAddress> {
> ```

out of curiosity what happens if we don't have this? I would hope that GSON is able to just deserialize the fields into the object but that might not work for some reason 
___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 72 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGhvzxCJn5E84zEObr:-NbGhvzxCJn5E84zEObs:b-qmywm4) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L72)):*
> ```Java
>         switch (reader.nextName()) {
>           case "street":
>             builder.setStreet(ImmutableList.of(reader.nextString()));
> ```

see other comment, we'll want a list of strings
___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 95 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGhjTV0w-lfngYnvQm:-NbGhjTV0w-lfngYnvQn:bxyrbb5) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L95)):*
> ```Java
>     public void write(JsonWriter writer, RegistrarAddress registrarAddress) throws IOException {
>       writer.beginObject();
>       writer.name("street").value(StringUtils.join(registrarAddress.getStreet(), ""));
> ```

We probably want to write a list of strings here explicitly instead of joining them together because that would allow us to deserialize them (like how we do `putListOfStrings` in `Address:toJsonMap`
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 42 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGiEL9DpTvuAE6KrJc:-NbGiEL9DpTvuAE6KrJd:bmo6em3) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L42)):*
> ```Java
> import javax.inject.Named;
> import javax.servlet.http.HttpServletRequest;
> import org.apache.commons.lang3.StringUtils;
> ```

given that they're equal we should use Guava's Strings.isNullOrEmpty to keep everything the same / because it's Google
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 90 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGiY-A1EZA22N6351J:-NbGiY-A1EZA22N6351K:b-rjdc3m) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L90)):*
> ```Java
>   }
> 
>   private void getHandler(User user) {
> ```

any thoughts on how we'll want to handle this when it's a registrar user, who will presumably have access to at least one but not all registrar objects? 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 119 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGk_txC0P9VHuenW_X:-NbGk_txC0P9VHuenW_Y:bplg7pn) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L119)):*
> ```Java
>     String errorMsg = "Missing value for %s";
>     try {
>       checkState(
> ```

these should all be checkArgument because they're info provided by the user

also for checkArgument (probably for checkState too but idk) you don't actually need to use String.format, you can just pass in the format-string and the arguments directly 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 121 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGjwb52K0DNj5H8-5D:-NbGjwb52K0DNj5H8-5E:b-untkdt) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L121)):*
> ```Java
>       checkState(
>           !StringUtils.isEmpty(registrarParam.getRegistrarId()),
>           String.format(errorMsg, "clientId"));
> ```

Why not just use the actual field names in these? 

(at the very least we should definitely use `registrarId`, we would like to get rid of the term `clientId` anywhere possible 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 146 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGlld31xbh8kT3hx4I:-NbGlld31xbh8kT3hx4J:b-uprvkq) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L146)):*
> ```Java
>           String.format(errorMsg, "address"));
> 
>       String password = passwordGenerator.createString(PASSWORD_LENGTH);
> ```

Remind me, did we talk about having the user provide this rather than it being generated? It looks like we require that the phone passcode and the password are provided by the "user" in the nomulus `create_registrar` tool. 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 162 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGm33MC7jR6MUHxgXY:-NbGm33MC7jR6MUHxgXZ:b-j4n0yi) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L162)):*
> ```Java
>               .setPhonePasscode(phonePasscode)
>               .setState(State.PENDING)
>               .setLocalizedAddress(registrarParam.getLocalizedAddress().asBuilder().build())
> ```

don't think `asBuilder` is necessary? 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 182 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGmI4N7EL3eFwjQT3M:-NbGmI4N7EL3eFwjQT3N:b-8los9d) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L182)):*
> ```Java
>               });
> 
>     } catch (Throwable e) {
> ```

Throwable is kinda generic and could mask other, less expected, types of things going wrong, can we be more specific?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: 9 of 12 files reviewed, 10 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGfWrC9zn9g2va9DXC:-NbKatNT4KJGGFjaT_kJ:b-rtvuus) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L64)):*

out of curiosity what happens if we don't have this? I would hope that GSON is able to just deserialize the fields into the object but that might not work for some reason 
</blockquote></details>

I was expecting that GSON would be capable to deserialize a simple object, but it just fills all the fields with `null`, so yeah the adapter is required 
___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 72 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGhvzxCJn5E84zEObr:-NbV3xee1waXseQFwKAd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L72)):*

see other comment, we'll want a list of strings
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 95 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGhjTV0w-lfngYnvQm:-NbV4-QyCQJu5eJ1EXE3:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L95)):*

We probably want to write a list of strings here explicitly instead of joining them together because that would allow us to deserialize them (like how we do `putListOfStrings` in `Address:toJsonMap`
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 42 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGiEL9DpTvuAE6KrJc:-NbQSiS0-pI3k9C9ToK0:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L42)):*

given that they're equal we should use Guava's Strings.isNullOrEmpty to keep everything the same / because it's Google
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 90 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGiY-A1EZA22N6351J:-NbQSyuB7S-vuivuvdOb:b-3b6zxp) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L90)):*

any thoughts on how we'll want to handle this when it's a registrar user, who will presumably have access to at least one but not all registrar objects? 
</blockquote></details>

We will provide only the registrars for which user would have access 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 119 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGk_txC0P9VHuenW_X:-NbQh9-e3xUfrlfhZ3UN:bm6vf53) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L119)):*

these should all be checkArgument because they're info provided by the user

also for checkArgument (probably for checkState too but idk) you don't actually need to use String.format, you can just pass in the format-string and the arguments directly 
</blockquote></details>

Updated, followed the same approach as in [https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java;l=141](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java;l=141 ) 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 121 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGjwb52K0DNj5H8-5D:-NbQhoGXAhiA8IwxSQeC:bbohfl7) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L121)):*

Why not just use the actual field names in these? 

(at the very least we should definitely use `registrarId`, we would like to get rid of the term `clientId` anywhere possible 
</blockquote></details>

That's how it's done in `ConsoleRegistrarCreatorAction.java`  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java;l=170
There's no such field on mockup - `registrarId`, it's called `clientId` on UI, which is why the response should match what's on UI, otherwise it'd be confusing for the user
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 146 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGlld31xbh8kT3hx4I:-NbQiNg40eZdOXOfvcZ6:bn3ljdg) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L146)):*

Remind me, did we talk about having the user provide this rather than it being generated? It looks like we require that the phone passcode and the password are provided by the "user" in the nomulus `create_registrar` tool. 
</blockquote></details>

I first had a chat with mbrem about it, because it's not on mockup and then I relayed this information to you as well. 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 162 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGm33MC7jR6MUHxgXY:-NbQiyrj1t78MB7HEh1T:b-5796px) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L162)):*

don't think `asBuilder` is necessary? 
</blockquote></details>

Why not? Otherwise we will have a reference to localizedAddress in registrarParam, instead of an immutable copy
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 182 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGmI4N7EL3eFwjQT3M:-NbQjJ-U79Wk6I_t5mxt:bjbs5h7) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L182)):*

Throwable is kinda generic and could mask other, less expected, types of things going wrong, can we be more specific? 
</blockquote></details>

That's how it currently written in `ConsoleRegistrarCreatorAction.java` . I don't think it's an issue, because we relay the **message** from throwable, throwable itself or other error type conveys lesser of a message, comparing to error message text. 
Please let me know how would you like to see it otherwise? Thanks

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: 11 of 12 files reviewed, 6 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGfWrC9zn9g2va9DXC:-NbWD-3n-bjwDy-P-EGH:b-6m8wbp) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L64)):*

I was expecting that GSON would be capable to deserialize a simple object, but it just fills all the fields with `null`, so yeah the adapter is required 
</blockquote></details>

Hmm I think something else is going wrong then because this definitely should work as far as I'm aware

I wrote up a very simple test:

```
String jsonString = "{\"registrarName\": \"TheRegistrar\", \"localizedAddress\": {\"street\\": [\"123 Fake St\", \"Apt 12\"], \"city\": \"New York\", \"state\": \"NY\", \"zip\": \"10011\", \"countryCode\": \"US\"}}";  
Registrar registrar = GSON.fromJson(jsonString, Registrar.class);
```
and it gave me a Registrar object with registrarName and localizedAddress filled out (and everything else null) as expected. I wonder what's going wrong
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 90 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGiY-A1EZA22N6351J:-NbWOutP4PFT4yOskEHo:b-3s8ts5) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L90)):*

We will provide only the registrars for which user would have access 
</blockquote></details>

yeah fair enough, i guess that'll just be another block that goes here if the caller doesn't have VIEW_REGISTRARS permissions
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 119 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGk_txC0P9VHuenW_X:-NbWEmtc8TvLyqTkGjQf:b-iu9uc5) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L119)):*

Updated, followed the same approach as in [https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java;l=141](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java;l=141 ) 
</blockquote></details>

oh lame, that one is probably not correct either but oh well 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 121 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGjwb52K0DNj5H8-5D:-NbWFEhIDKpoC6A4o_Zo:b-8ins06) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L121)):*

That's how it's done in `ConsoleRegistrarCreatorAction.java`  https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java;l=170
There's no such field on mockup - `registrarId`, it's called `clientId` on UI, which is why the response should match what's on UI, otherwise it'd be confusing for the user

</blockquote></details>

ah yeah, I must have missed that on the mockup -- my bad. In that case the mockup is incorrect; we want to avoid using `clientId` and we should use `registrarId` instead 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 162 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGm33MC7jR6MUHxgXY:-NbWFS0z9FmTHDn7wb-2:br8s8z0) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L162)):*

Why not? Otherwise we will have a reference to localizedAddress in registrarParam, instead of an immutable copy
</blockquote></details>

What's bad about having that reference? It cannot be modified and we're saving it to the DB anyway 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 182 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGmI4N7EL3eFwjQT3M:-NbWGjMBAPrk9fVtsqIz:b-x16per) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L182)):*

That's how it currently written in `ConsoleRegistrarCreatorAction.java` . I don't think it's an issue, because we relay the **message** from throwable, throwable itself or other error type conveys lesser of a message, comparing to error message text. 
Please let me know how would you like to see it otherwise? Thanks
</blockquote></details>

yeah maybe we have two separate catch clauses: one for an IllegalArgumentException that returns a 400 status code and one generic one that returns 500 for a server error (for unexpected issues)? What do you think about that? 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 167 at r2](https://reviewable.io/reviews/google/nomulus/2095#-NbWGy7qBbcakNDBN3-y:-NbWGy7qBbcakNDBN3-z:b-tnumnu) ([raw file](https://github.com/google/nomulus/blob/bdf7f2d95edae9271047619fe5a651a96dad9009/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L167)):*
> ```Java
>       tm().transact(
>               () -> {
>                 checkState(
> ```

we probably want `checkArgument` here too sorry

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: 10 of 12 files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGfWrC9zn9g2va9DXC:-Nba0wcu8EMcqJpRzk52:buizkk9) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L64)):*

That's because you explicitly specify the class to use for mapping - `Registrar.class`. We use a generic 
`JsonObject.class` to provide the body once - https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/request/RequestModule.java;l=247. 

So as far as I can tell it's either the way we do it now or we have multiple providers. I prefer the current approach. 
</blockquote></details>

ok yeah so we have three options 
1. create TypeAdapters for every single class that can be nested inside an outer class (what this is doing)
2. Parse the body as a string and have one provider method per object type, where each one of those just returns `gson.fromJson(jsonString, TheType.class)` 
3. Forget the providing entirely, provide only the `HttpServletRequest` in each of the `Action`s and have a static helper method that parses the body as a JSON string, i.e.

```
public static <T> T parseRequestBody(HttpServletRequest req, Class<T> clazz) {
  return GSON.fromJson(req.getReader(), clazz);
}
```
maybe converting any JSON parsing exceptions to IllegalArgumentExceptions, not sure 



I suppose the feasibility of 2 would depend solely on how many types of objects we might be passing in but as the console expands there could very possibly be many. What do you think about option 3? It would mean that we'd only need to provide the request (which we do anyway) and then we have only one utils method that needs to be called. 
___
*[`core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java` line 121 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGjwb52K0DNj5H8-5D:-Nba4ogv9T-pIab1toBe:bdtyuum) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/ui/server/console/RegistrarsAction.java#L121)):*

Done.
</blockquote></details>

thaaaaaaanks

gbrodman (changes requested):
> Reviewed 2 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/model/registrar/RegistrarAddress.java` line 64 at r1](https://reviewable.io/reviews/google/nomulus/2095#-NbGfWrC9zn9g2va9DXC:-Nbz0NBHBKqqb_v2KAM-:b-shpz19) ([raw file](https://github.com/google/nomulus/blob/e32a81a0e567f21588cd9735a4139bdac6c9c7c1/core/src/main/java/google/registry/model/registrar/RegistrarAddress.java#L64)):*

Updated, turns out all it needed was `Expose` annotation on Address fields. It works perfectly now without an adapter.
</blockquote></details>

yaaaaay  (sorry i didn't notice the lack of annotations earlier!) 
___
*[`core/src/test/java/google/registry/ui/server/console/RegistrarsActionTest.java` line 194 at r4](https://reviewable.io/reviews/google/nomulus/2095#-Nbz3FycCyBr9HlbH5J5:-Nbz3FycCyBr9HlbH5J6:b1ra247) ([raw file](https://github.com/google/nomulus/blob/3565f21f8d8a7057f2b62e908b72072241ebdbde/core/src/test/java/google/registry/ui/server/console/RegistrarsActionTest.java#L194)):*
> ```Java
>                           String.format(
>                               "Missing value for %s", userFriendlyKeysToRegistrarKeys.get(key))));
>               registrarParamMap = copy;
> ```

small comment, this line shouldn't be necessary

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ptkach)

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2095)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ptkach)

PR #2100, opened 2023-08-09
Add backend for editing whois-visible fields

gbrodman:
This includes a bit of refactoring of the GSON creation. There can exist
some objects (e.g. Address) where the JSON representation is not equal to the
representation that we store in the database. For these objects, when
deserializing, we should update the objects so that they reflect the
proper DB structure (indeed, this is already what we do for the XML
parsing of Address).

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2100)


gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2100)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsAction.java` line 87 at r1](https://reviewable.io/reviews/google/nomulus/2100#-NbVHmmA7n_DANs2wa9k:-NbWNmRY3DCjVfxouOmk:b-rgqttk) ([raw file](https://github.com/google/nomulus/blob/8a8b97b6b7cbd0d8fc4e0f5e4f6d502286fcc95b/core/src/main/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsAction.java#L87)):*

I'm not sure what will happen if `loadAndModifyRegistrar` catches an error and sets the status to `HttpStatusCodes.STATUS_CODE_FORBIDDEN` and then this line tries to set it again? 
</blockquote></details>

good point -- any method that can possibly set a response needs to either be the last line in the calling method or it needs to have some sort of checked return value
___
*[`core/src/test/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsActionTest.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/2100#-NbVIXvcA39yinsczw4m:-NbWJDhG6yMjfKUuSAlA:b-oleao7) ([raw file](https://github.com/google/nomulus/blob/8a8b97b6b7cbd0d8fc4e0f5e4f6d502286fcc95b/core/src/test/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsActionTest.java#L65)):*

Although that works, I'd prefer to be closer to the real-life scenario and work with the JSON instead of an object, same as I did in my tests. 
</blockquote></details>

I think if we do that for every action we're going to end up testing the same area of code (parsing JSON strings into a partially-filled Registrar object) a bunch of times unnecessarily.  Maybe we should have one area that tests that functionality and the rest of the code can use the simpler format we do here? 

It is the case that in many other POST actions (ones that pre-date the console) we don't test the ability to parse the request properly, and I'm not sure how I feel about that. It's not ideal, but I suppose in most of those cases we're not actually parsing the request body, just the request parameters.

ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2100)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsActionTest.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/2100#-NbVIXvcA39yinsczw4m:-Nb_Dd-Z81qzbnEcc_f_:b4g33cy) ([raw file](https://github.com/google/nomulus/blob/8a8b97b6b7cbd0d8fc4e0f5e4f6d502286fcc95b/core/src/test/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsActionTest.java#L65)):*

I think if we do that for every action we're going to end up testing the same area of code (parsing JSON strings into a partially-filled Registrar object) a bunch of times unnecessarily.  Maybe we should have one area that tests that functionality and the rest of the code can use the simpler format we do here? 

It is the case that in many other POST actions (ones that pre-date the console) we don't test the ability to parse the request properly, and I'm not sure how I feel about that. It's not ideal, but I suppose in most of those cases we're not actually parsing the request body, just the request parameters.
</blockquote></details>

I don't think it'd be parsing a bunch of times unnecessary, because each post action ultimately is responsible for only piece of data related to it. Apart from that overhead for json is not that big and I think it's worth working with it, like in my case (https://github.com/google/nomulus/pull/2095) - I only was able to figured out that localizedAddress has null fields because I was working with json and not the object directly.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2100)* status: 1 of 12 files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot] and @ptkach)

___
*[`core/src/test/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsActionTest.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/2100#-NbVIXvcA39yinsczw4m:-NcO6rzSB9xCsXYy88HM:b-f0rukq) ([raw file](https://github.com/google/nomulus/blob/8a8b97b6b7cbd0d8fc4e0f5e4f6d502286fcc95b/core/src/test/java/google/registry/ui/server/console/settings/WhoisRegistrarFieldsActionTest.java#L65)):*

I don't think it'd be parsing a bunch of times unnecessary, because each post action ultimately is responsible for only piece of data related to it. Apart from that overhead for json is not that big and I think it's worth working with it, like in my case (https://github.com/google/nomulus/pull/2095) - I only was able to figured out that localizedAddress has null fields because I was working with json and not the object directly. 
</blockquote></details>

We should be all set here

ptkach (approved):
> Reviewed 10 of 11 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2100)* status: all files reviewed, 1 unresolved discussion (waiting on @github-code-scanning[bot])

PR #2104, opened 2023-08-14
Add the ability to specify transaction isolation level on a per-transaction basis

jianglai:
A config file field is added to control if per-transaction isolation
level is actually used. If set to true, nested transactions will throw
a runtime exception as the enclosing transaction may run at a different
isolation level.

In this PR we only add the ability to specify the isolation level,
without enabling it in any environment (including unit tests), or
actually specifying one for any query. This should allow us to set up
the system without impacting anything currently working.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2104)

### DIFF
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -1547,6 +1547,11 @@ public final class RegistryConfig {
     return CONFIG_SETTINGS.get().hibernate.connectionIsolation;
   }
 
+  /** Returns true if per-transaction isolation level is enabled. */
+  public static boolean getHibernatePerTransactionIsolationEnabled() {
+    return CONFIG_SETTINGS.get().hibernate.perTransactionIsolation;
+  }
+
   /** Returns true if hibernate.show_sql is enabled. */
   public static String getHibernateLogSqlQueries() {
     return CONFIG_SETTINGS.get().hibernate.logSqlQueries;
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -115,6 +115,7 @@ public class RegistryConfigSettings {
 
   /** Configuration for Hibernate. */
   public static class Hibernate {
+    public boolean perTransactionIsolation;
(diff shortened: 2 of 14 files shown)


jianglai (commented):
> Reviewed 14 of 14 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2104)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2104)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2104)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/persistence/PersistenceModule.java` line 336 at r3](https://reviewable.io/reviews/google/nomulus/2104#-Nc8B-F1CzePk3VmhpHK:-NcNKgTT-GEr70epQ8ZJ:bmo2ech) ([raw file](https://github.com/google/nomulus/blob/f5044cb92caa15fdbfea350223a46fb9e7f07db0/core/src/main/java/google/registry/persistence/PersistenceModule.java#L336)):*

Maybe just drop this prefix so we don't need the mode property?
</blockquote></details>

Unfortunately it is not that simple. We need the `TRANSACTION_` prefix because the fields in `java.sql.Connection` have it, and these fields are used to validate the enum names:

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/persistence/PersistenceModule.java;l=347?q=persistencemodule&ss=nomulus%2Fnomulus

So, if we drop the prefix, we'll need to manually add them for validation. Also, the value returned from PostgreSQL is whitespace-deliminated, therefore converting that to understore-deliminated is always necessary, anyway.

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2104)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2113, opened 2023-08-17
Refactor the way that the console BE parses POST bodies

gbrodman:
This includes two changes:
1. Creating a base string-type adapter for use parsing to/from JSON classes that are represented as simple strings
2. Changing the object-provider methods so that the POST bodies should contain precisely the expected object(s) and nothing else. This way, it's easier for the frontend and backend to agree that, for instance, one POST endpoint might accept exactly a Registrar object, or a list of Contact objects.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2113)

### DIFF
--- a/console-webapp/src/app/shared/services/backend.service.ts
+++ b/console-webapp/src/app/shared/services/backend.service.ts
@@ -12,11 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Injectable } from '@angular/core';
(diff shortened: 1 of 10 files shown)


ptkach (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2113)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #2125, opened 2023-08-25
Add tm().reTransact() methods and refactor away some inner transactions

CydeWeys:
In the future, reTransact() will be the only way to initiate a transaction that doesn't fail when called inside an outer wrapping transaction (when wrapped, it's a no-op). It should be used sparingly, with a preference towards refactoring the code to move transactions outwards (which this PR also contains).

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2125)


jianglai (approved):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2125)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #2129, opened 2023-08-25
Add a new MutatingFlow interface and make most flows use Transactiona…

CydeWeys:
…lFlow

The old semantics for TransactionalFlow meant "anything that needs to mutate the database", but then FlowRunner was not creating transactions for non-transactional flows even though nearly every flow needs a transaction (as nearly every flow needs to hit the database for some purpose).  So now TransactionalFlow simply means "any flow that needs the database", and MutatingFlow means "a flow that mutates the database". In the future we will have FlowRunner use a read-only transaction for TransactionalFlow and then a normal writes-allowed transaction for MutatingFlow. That is a TODO.

This also fixes up some transact() calls inside caches to be reTransact(), as we rightly can't move the transaction outside them as from some callsites we legitimately do not know whether a transaction will be needed at all (depending on whether said data is already in memory). And it removes the replicaTm() calls which weren't actually doing anything as they were always nested inside of normal tm()s, thus causing confusion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2129)

### DIFF
--- a/core/src/main/java/google/registry/flows/FlowRunner.java
+++ b/core/src/main/java/google/registry/flows/FlowRunner.java
@@ -78,6 +78,8 @@ public class FlowRunner {
       return EppOutput.create(flowProvider.get().run());
     }
     try {
+      // TODO(mcilwain/weiminyu): Use transactReadOnly() here for TransactionalFlow and transact()
+      //                          for MutatingFlow.
       return tm().transact(
               () -> {
                 try {
--- a/core/src/main/java/google/registry/flows/MutatingFlow.java
+++ b/core/src/main/java/google/registry/flows/MutatingFlow.java
@@ -1,0 +1,23 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.flows;
+
+/**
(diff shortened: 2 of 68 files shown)


CydeWeys (commented):
PTAL, had to make additional changes to get tests to pass.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2129#-:-NciKLnADS87sQ8U2ZzI:b-5jsy1y)* status: 36 of 68 files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
Are there any remaining instances of something implementing only `Flow`? If not (and I think there probably shouldn't be?), should we make Flow package-private to indicate that one should not implement it directly? 

> Reviewed 9 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2129#-:-NciRbD89x9xrS6xMSY6:bl98lt)* status: 36 of 68 files reviewed, all discussions resolved (waiting on @weiminyu)

CydeWeys (commented):
`HelloFlow` and `LogoutFlow` are the two remaining flows implementing `Flow` directly, as they don't require any DB interaction at all.  Then there's some other stuff for tests only.  It can't be made package private however because it is referred to from many other packages already, e.g. `google.registry.flows.contact`, `google.registry.flows.domain`, and more.

Note that the only reason to write more classes implementing flow would be if the base EPP spec itself were amended to add more types of commands, and that hasn't happened since well before I even started working on this stuff.  So I don't think it's much of a concern.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2129#-:-NciYx3h5yrp_b-5o06D:b3m57dd)* status: 36 of 68 files reviewed, all discussions resolved (waiting on @gbrodman and @weiminyu)

CydeWeys (commented):
PTAL.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2129#-:-NcxX2DTDU58c6nnRKDQ:b-83330m)* status: 36 of 68 files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (approved):
> Reviewed 27 of 36 files at r1, 32 of 32 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2129)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #2133, opened 2023-09-01
Add a transactional parameter to @Action

CydeWeys:
This starts the process of removing inner transactions on some Registrar methods with some resultant refactoring, but then I gave up after awhile and turned them into reTransact() calls just to be able to get in the Action changes quickly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2133)

### DIFF
--- a/core/src/main/java/google/registry/tools/ConfirmingCommand.java
+++ b/core/src/main/java/google/registry/tools/ConfirmingCommand.java
@@ -14,6 +14,7 @@
 
 package google.registry.tools;
 
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.tools.CommandUtilities.promptForYes;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
@@ -44,20 +45,26 @@ public abstract class ConfirmingCommand implements Command {
 
   @Override
   public final void run() throws Exception {
-    if (checkExecutionState()) {
-      init();
-      printLineIfNotEmpty(prompt(), printStream);
-      if (dontRunCommand()) {
-        // This typically happens when all of the work is accomplished inside of prompt(), so do
-        // nothing further.
-        return;
-      } else if (force || promptForYes("Perform this command?")) {
-        printStream.println("Running ... ");
-        printStream.println(execute());
-        printLineIfNotEmpty(postExecute(), printStream);
-      } else {
-        printStream.println("Command aborted.");
-      }
+    if (!checkExecutionState()) {
+      return;
+    }
+
+    tm().transact(
+            () -> {
+              init();
+              printLineIfNotEmpty(prompt(), printStream);
+            });
+
+    if (dontRunCommand()) {
+      // This typically happens when all the work is accomplished inside of prompt(), so do
+      // nothing further.
+      return;
+    } else if (force || promptForYes("Perform this command?")) {
+      printStream.println("Running ... ");
+      tm().transact(() -> printStream.println(execute()));
+      tm().transact(() -> printLineIfNotEmpty(postExecute(), printStream));
+    } else {
+      printStream.println("Command aborted.");
     }
     printStream.close();
     errorPrintStream.close();
--- a/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java
+++ b/core/src/main/java/google/registry/batch/DeleteLoadTestDataAction.java
@@ -89,12 +89,9 @@ public class DeleteLoadTestDataAction implements Runnable {
         !RegistryEnvironment.get().equals(PRODUCTION),
         "This action is not safe to run on PRODUCTION.");
 
-    tm().transact(
-            () -> {
-              LOAD_TEST_REGISTRARS.forEach(this::deletePollMessages);
-              tm().loadAllOfStream(Contact.class).forEach(this::deleteContact);
-              tm().loadAllOfStream(Host.class).forEach(this::deleteHost);
-            });
+    LOAD_TEST_REGISTRARS.forEach(this::deletePollMessages);
+    tm().loadAllOfStream(Contact.class).forEach(this::deleteContact);
+    tm().loadAllOfStream(Host.class).forEach(this::deleteHost);
   }
 
   private void deletePollMessages(String registrarId) {
--- a/core/src/main/java/google/registry/model/registrar/Registrar.java
+++ b/core/src/main/java/google/registry/model/registrar/Registrar.java
@@ -585,7 +585,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   }
 
   private ImmutableSet<RegistrarPoc> getContactPocs() {
-    return tm().transact(
+    return tm().reTransact(
             () ->
                 tm().query("FROM RegistrarPoc WHERE registrarId = :registrarId", RegistrarPoc.class)
                     .setParameter("registrarId", registrarId)
@@ -944,7 +944,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
 
   /** Loads all registrar entities directly from the database. */
   public static Iterable<Registrar> loadAll() {
-    return tm().transact(() -> tm().loadAllOf(Registrar.class));
+    return tm().loadAllOf(Registrar.class);
   }
 
   /** Loads all registrar entities using an in-memory cache. */
@@ -962,7 +962,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   /** Loads and returns a registrar entity by its id directly from the database. */
   public static Optional<Registrar> loadByRegistrarId(String registrarId) {
     checkArgument(!Strings.isNullOrEmpty(registrarId), "registrarId must be specified");
-    return tm().transact(() -> tm().loadByKeyIfPresent(createVKey(registrarId)));
+    return tm().reTransact(() -> tm().loadByKeyIfPresent(createVKey(registrarId)));
   }
 
   /**
--- a/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java
+++ b/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java
@@ -69,7 +69,7 @@ public final class UpdateRegistrarRdapBaseUrlsAction implements Runnable {
   public void run() {
     try {
       ImmutableMap<String, String> ianaIdsToUrls = getIanaIdsToUrls();
-      tm().transact(() -> processAllRegistrars(ianaIdsToUrls));
+      processAllRegistrars(ianaIdsToUrls);
     } catch (Exception e) {
       throw new InternalServerErrorException("Error when retrieving RDAP base URL CSV file", e);
     }
--- a/core/src/main/java/google/registry/request/Action.java
+++ b/core/src/main/java/google/registry/request/Action.java
@@ -59,6 +59,9 @@ public @interface Action {
   /** HTTP methods that request processor should allow. */
   Method[] method() default Method.GET;
 
+  /** Whether to create a wrapping DB transaction around the execution of the action. */
+  boolean transactional() default true;
+
   /**
    * Indicates request processor should print "OK" to the HTTP client on success.
    *
--- a/core/src/main/java/google/registry/request/RequestHandler.java
+++ b/core/src/main/java/google/registry/request/RequestHandler.java
@@ -16,6 +16,7 @@ package google.registry.request;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.net.MediaType.PLAIN_TEXT_UTF_8;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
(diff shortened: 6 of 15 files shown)


core/src/main/java/google/registry/tools/ConfirmingCommand.java
```diff
@@ -29,20 +31,46 @@ public abstract class ConfirmingCommand implements Command {
...
-        System.out.println(execute());
-        printLineIfNotEmpty(postExecute());
-      } else {
-        System.out.println("Command aborted.");
-      }
+    if (!checkExecutionState()) {
+      return;
+    }
+
+    tm().transact(
```
CydeWeys, line 38: I really wish there were a way to make these 3 repeated calls nicer, but I couldn't find any.  You can't extract to a helper function that takes a lambda.

jianglai (changes requested):
It seems like most of the change are to various commands. Maybe you should consider changing the title of the PR to reflect that.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2133#-:-Nei5kWU9mtiZd5NCgRG:b-l5x51l)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/request/Action.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/2133#-NehwmSB6C-I0m_lkZ6k:-NehwmSB6C-I0m_lkZ6l:bihdrmn) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/request/Action.java#L63)):*
> ```Java
> 
>   /** Whether to create a wrapping DB transaction around the execution of the action. */
>   boolean transactional() default true;
> ```

Are we sure we want to set the default to true? Did we survey the actions and conclude that the majority of them need to be wrapped in a transaction in their entirety?
___
*[`core/src/main/java/google/registry/tools/ConfirmingCommand.java` line 38 at r1](https://reviewable.io/reviews/google/nomulus/2133#gh-1321668584:-Nei1bMLCuXAh8fN6-j1:b-6hihol) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/tools/ConfirmingCommand.java#L38)):*

I really wish there were a way to make these 3 repeated calls nicer, but I couldn't find any.  You can't extract to a helper function that takes a lambda.
</blockquote></details>

The root problem is that `tm().transact()` expects a `Runnable`, whose `run()` method doesn't declare a checked exception, so your lambda cannot either.

You can get around it by making your lambda a `Callable<Void>`, then feeding it to a wrapper method that converts it back to a `Runnable` but rethrows the checked exception as unchecked. The key here is that the `Callable.call()` method declares a checked exception, so you are allowed to supply a lambda that throws a checked exception.

So it would be like this:

```
private static Runnable wrapper(Callable<Void> callable) {  
  return new Runnable() {  
    @Override  
    public void run() {  
      try {  
        callable.call();  
      } catch (Exception e) {  
        rethrow(e);  
      }  
    }  
  };  
}

tm().transact(wrapper(  
    () -> {  
      init();  
      printLineIfNotEmpty(prompt());  
      return null;  
    }));
```

Whether it is preferable to just do the rethrow manually is up for debate.

Alternatively, I see no reason why we cannot just have `tm().transact()` take a `Callable<T>` instead of a `Supplier<T>` (which doesn't allow checked exceptions), so you don't need a wrapper method and can just do:

```
tm().transact(  
        () -> {  
          init();  
          printLineIfNotEmpty(prompt());  
          return null;  
        });
```

We can maybe go a step further and define our own `ThrowableRunnable` that allows throwing, and make `tm().transact()` take it instead of the vanilla `Runnable`, in which case you can just provide the lambda as is.

It seems to me that there really isn't a need to force the caller to handle checked exceptions when supplying a lambda to `tm().transact()`, if all they do is to just re-throw them as unchecked. `tm().transact()` wraps the work in its own try block so any exception thrown by the workload would be handled by the transaction manager anyway.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java` line 221 at r1](https://reviewable.io/reviews/google/nomulus/2133#-NehynuuApHEmg7E-lnM:-NehynuuApHEmg7E-lnN:b-ehf02l) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/ui/server/registrar/RegistrarSettingsAction.java#L221)):*
> ```Java
> 
>   private Map<String, Object> read(String registrarId) {
>     return tm().transact(
> ```

What's wrong with just wrapping `loadRegistrarUnchecked(registrarId)` inside the transaction? That way you don't need to change the return value here.

Also, the entire `RegistrarResult` class seems redundant at this point, since you are just calling the `toJsonResponse()` method on it to get the map anyway.
___
*[`core/src/test/java/google/registry/whois/WhoisCommandFactoryTest.java` line 100 at r1](https://reviewable.io/reviews/google/nomulus/2133#-Nei5d_v8dlQ8J-nev8q:-Nei5d_v8dlQ8J-nev8r:bydbv1g) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/test/java/google/registry/whois/WhoisCommandFactoryTest.java#L100)):*
> ```Java
>             () -> {
>               WhoisResponse response =
>                   assertDoesNotThrow(
> ```

Why do we need this assertion? If the lambda throws, wouldn't the whole test method fail anyway?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2133)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/Action.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/2133#-NehwmSB6C-I0m_lkZ6k:-Nj9TeYdAwLuu1afNwAa:bm3o3r8) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/request/Action.java#L63)):*

Are we sure we want to set the default to true? Did we survey the actions and conclude that the majority of them need to be wrapped in a transaction in their entirety?
</blockquote></details>

Yes, I did survey the actions and determine that almost all of them are transactional.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2133)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/test/java/google/registry/whois/WhoisCommandFactoryTest.java` line 100 at r1](https://reviewable.io/reviews/google/nomulus/2133#-Nei5d_v8dlQ8J-nev8q:-Nj9Tory-dALGhyjLVxc:bg2qbuu) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/test/java/google/registry/whois/WhoisCommandFactoryTest.java#L100)):*

Why do we need this assertion? If the lambda throws, wouldn't the whole test method fail anyway?
</blockquote></details>

If the lambda throws then it can't be passed to the lambda in `transact()`.  The checked exception needs to be handled.  `assertDoesNotThrow()` handles the checked exception and does not rethrow a checked exception.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2133)* status: 10 of 15 files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/ConfirmingCommand.java` line 38 at r1](https://reviewable.io/reviews/google/nomulus/2133#gh-1321668584:-Nj9_zKhBHznpq-MPFdi:bs0i1xe) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/tools/ConfirmingCommand.java#L38)):*

The root problem is that `tm().transact()` expects a `Runnable`, whose `run()` method doesn't declare a checked exception, so your lambda cannot either.

You can get around it by making your lambda a `Callable<Void>`, then feeding it to a wrapper method that converts it back to a `Runnable` but rethrows the checked exception as unchecked. The key here is that the `Callable.call()` method declares a checked exception, so you are allowed to supply a lambda that throws a checked exception.

So it would be like this:

```
private static Runnable wrapper(Callable<Void> callable) {  
  return new Runnable() {  
    @Override  
    public void run() {  
      try {  
        callable.call();  
      } catch (Exception e) {  
        rethrow(e);  
      }  
    }  
  };  
}

tm().transact(wrapper(  
    () -> {  
      init();  
      printLineIfNotEmpty(prompt());  
      return null;  
    }));
```

Whether it is preferable to just do the rethrow manually is up for debate.

Alternatively, I see no reason why we cannot just have `tm().transact()` take a `Callable<T>` instead of a `Supplier<T>` (which doesn't allow checked exceptions), so you don't need a wrapper method and can just do:

```
tm().transact(  
        () -> {  
          init();  
          printLineIfNotEmpty(prompt());  
          return null;  
        });
```

We can maybe go a step further and define our own `ThrowableRunnable` that allows throwing, and make `tm().transact()` take it instead of the vanilla `Runnable`, in which case you can just provide the lambda as is.

It seems to me that there really isn't a need to force the caller to handle checked exceptions when supplying a lambda to `tm().transact()`, if all they do is to just re-throw them as unchecked. `tm().transact()` wraps the work in its own try block so any exception thrown by the workload would be handled by the transaction manager anyway.
</blockquote></details>

Thanks for your `ThrowingRunnable` PR, this is nicer now.

jianglai (changes requested):
> Reviewed 4 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2133)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/request/Action.java` line 63 at r1](https://reviewable.io/reviews/google/nomulus/2133#-NehwmSB6C-I0m_lkZ6k:-NjP97vb3x1U3vi5LTFx:b7cmgd) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/request/Action.java#L63)):*

Yes, I did survey the actions and determine that almost all of them are transactional.
</blockquote></details>

See my other comment below, `EppTlsAction` needs to be marked as non-transactional, and there needs to be another field to specify the transaction isolation level override, if any. 
___
*[`core/src/main/java/google/registry/request/RequestHandler.java` line 161 at r3](https://reviewable.io/reviews/google/nomulus/2133#-NjP5YCV-u16tH_oE-DU:-NjP5YCV-u16tH_oE-DV:bdmyqg2) ([raw file](https://github.com/google/nomulus/blob/045655d35f2ddba7cba4a4e11bd351be5c7c4e8f/core/src/main/java/google/registry/request/RequestHandler.java#L161)):*
> ```Java
>       Runnable action = route.get().instantiator().apply(component);
>       if (route.get().action().transactional()) {
>         tm().transact(action::run);
> ```

Now that we have a good understanding on how to handle nested transactions, I have some concern on how this would work together with the transaction isolation level override in the long run.

The override can only be applied at the top level transaction, which would be here. We therefore need to allow setting the override (possibly via another field in the`@Action` annotation).

However, all EPP flows are executed by `EppTlsAction`, and we have envisioned using the [`@IsolationLevel` annotation](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/persistence/IsolationLevel.java) on a flow to signal to the [`FlowRunner`](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/flows/FlowRunner.java;l=100?q=flowrunner&ss=nomulus%2Fnomulus).

In order to allow different EPP flows to use different isolation levels, we need to annotate the `EppTlsAction` as non-transactional, and define the transaction behavior at the flow level. This is doable, but needs to be clearly documented to avoid confusion.
___
*[`core/src/main/java/google/registry/tools/ConfirmingCommand.java` line 38 at r1](https://reviewable.io/reviews/google/nomulus/2133#gh-1321668584:-NjP9Jng2wzZqf8-PyZx:b-90m8us) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/main/java/google/registry/tools/ConfirmingCommand.java#L38)):*

Thanks for your `ThrowingRunnable` PR, this is nicer now.
</blockquote></details>

It's it sooooo much nicer!
___
*[`core/src/test/java/google/registry/whois/WhoisCommandFactoryTest.java` line 100 at r1](https://reviewable.io/reviews/google/nomulus/2133#-Nei5d_v8dlQ8J-nev8q:-NjP8YfG7_NkLZTqyPti:bytwnjt) ([raw file](https://github.com/google/nomulus/blob/7ab6511ec312d13767ab9ba7356df43e03886bb0/core/src/test/java/google/registry/whois/WhoisCommandFactoryTest.java#L100)):*

If the lambda throws then it can't be passed to the lambda in `transact()`.  The checked exception needs to be handled.  `assertDoesNotThrow()` handles the checked exception and does not rethrow a checked exception.
</blockquote></details>

Does it still apply now that we can throw checked exceptions in the lambda supplied to `tm().transact()`? It will just be wrapped in a `RuntimeException` if it throws, correct?

PR #2134, opened 2023-09-01
Disable Hibernate error logging

weiminyu:
Hibernate logs certain information at the ERROR level, which for the purpose of troubleshooting is misleading, since most affected operations succeed after retry. ERROR-level logging should only be added by Nomulus code.

This PR does two things:
1. Disable all logging in two Hibernate classes: we cannot disable logging at a finer granularity, and we cannot preserve lower-level logging while disabling ERROR.
2. Adds a DatabaseException which captures all error details that may escape the typical loggers' attention: SQLException instances can be chained in a different way from Throwable's `getCause()` method.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2134)

### DIFF
--- a/core/src/main/java/google/registry/env/common/backend/WEB-INF/logging.properties
+++ b/core/src/main/java/google/registry/env/common/backend/WEB-INF/logging.properties
@@ -11,3 +11,7 @@
 
 # Set the default logging level for all loggers to INFO.
 .level = INFO
+
+# Turn off logging in Hibernate classes for misleading ERROR-level logs
+org.hibernate.engine.jdbc.batch.internal.BatchingBatch.level=OFF
+org.hibernate.engine.jdbc.spi.SqlExceptionHelper.level=OFF
--- a/core/src/main/java/google/registry/env/common/default/WEB-INF/logging.properties
+++ b/core/src/main/java/google/registry/env/common/default/WEB-INF/logging.properties
@@ -11,3 +11,7 @@
 
 # Set the default logging level for all loggers to INFO.
 .level = INFO
+
+# Turn off logging in Hibernate classes for misleading ERROR-level logs
+org.hibernate.engine.jdbc.batch.internal.BatchingBatch.level=OFF
+org.hibernate.engine.jdbc.spi.SqlExceptionHelper.level=OFF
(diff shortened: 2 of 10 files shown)


sarahcaseybot (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2134)* status: all files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java` line 218 at r1](https://reviewable.io/reviews/google/nomulus/2134#-NdaqLGEB6zJwxVFVgmt:-NdaqLGEB6zJwxVFVgmu:bnizhja) ([raw file](https://github.com/google/nomulus/blob/a3e2e3056454e2b8980a2dd0d3f1667ca1d7ee2f/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L218)):*
> ```Java
>       }
>       tryWrapAndThrow(e);
>       throw e;
> ```

Is this line necessary here? Why not add a `throw e` at the end of `tryWrapAndThrow`?

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2134)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java` line 218 at r1](https://reviewable.io/reviews/google/nomulus/2134#-NdaqLGEB6zJwxVFVgmt:-NdarJN1B4dh4GXymtSV:buyyrgh) ([raw file](https://github.com/google/nomulus/blob/a3e2e3056454e2b8980a2dd0d3f1667ca1d7ee2f/core/src/main/java/google/registry/persistence/transaction/JpaTransactionManagerImpl.java#L218)):*

Is this line necessary here? Why not add a `throw e` at the end of `tryWrapAndThrow`?
</blockquote></details>

'e' could be either an unchecked exception or an `Error` (which is also unchecked).
However, when passing `e` `tryWrapAndThrow`, we cannot find a type that is unchecked and a super class of both, so have to settle on `Throwable`.  If we try to throw e from inside the method, we'll have to do unnecessary type check.

sarahcaseybot (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2134#-:-NdfUwkuEZt0YFWG2N2F:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #2138, opened 2023-09-06
Use registrar ID, not name, in selector

gbrodman:
We set the active registrar ID based on this choice so it needs to be the ID, not the name

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2138)

### DIFF
--- a/console-webapp/src/app/registrar/registrar-selector.component.html
+++ b/console-webapp/src/app/registrar/registrar-selector.component.html
@@ -8,9 +8,9 @@
       >
         <mat-option
           *ngFor="let registrar of registrarService.registrars"
-          [value]="registrar.registrarName"
+          [value]="registrar.registrarId"
         >
-          {{ registrar.registrarName }}
+          {{ registrar.registrarId }}
         </mat-option>
       </mat-select>
     </mat-form-field>


gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2138)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/registrar/registrar-selector.component.html` line 13 at r1](https://reviewable.io/reviews/google/nomulus/2138#-Ndjsiv1EwYzc6Ykvb9M:-Ne9Nn0SDiknKKKB0taA:b-l7d81e) ([raw file](https://github.com/google/nomulus/blob/d5c528b46a44a4de67fc59165e3f26a35b345c1d/console-webapp/src/app/registrar/registrar-selector.component.html#L13)):*

Although I agree that we should use id as a value, I think we should stick with name as a label in selector, as it's more user friendly
</blockquote></details>

Some of the names can be pretty long / involved and are thus (i think?) not suitable for a dropdown -- have you looked at the list of names in prod? Like, some are even around 50 chars

ptkach (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2138)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/registrar/registrar-selector.component.html` line 13 at r1](https://reviewable.io/reviews/google/nomulus/2138#-Ndjsiv1EwYzc6Ykvb9M:-Ne9_Mzm3XzYWWcctaaE:b-4c50x7) ([raw file](https://github.com/google/nomulus/blob/d5c528b46a44a4de67fc59165e3f26a35b345c1d/console-webapp/src/app/registrar/registrar-selector.component.html#L13)):*

Some of the names can be pretty long / involved and are thus (i think?) not suitable for a dropdown -- have you looked at the list of names in prod? Like, some are even around 50 chars
</blockquote></details>

Ugh, yeah I see, names we have are not suitable for this dropdown. Usually id should be hidden from the representation, but I guess not this time.

PR #2149, opened 2023-09-14
Don't include a nextUrl when accessing the console homepage

gbrodman:
In this case we should just display the standard page, no need to redirect anywhere since there's nothing to redirect to.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2149)

### DIFF
--- a/console-webapp/src/app/registrar/emptyRegistrar.component.ts
+++ b/console-webapp/src/app/registrar/emptyRegistrar.component.ts
@@ -14,8 +14,9 @@
 
 import { Component } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
-import { RegistrarService } from './registrar.service';
 import { Subscription } from 'rxjs';
+
+import { RegistrarService } from './registrar.service';
 
 @Component({
   selector: 'app-empty-registrar',
--- a/console-webapp/src/app/registrar/registrar.guard.ts
+++ b/console-webapp/src/app/registrar/registrar.guard.ts
@@ -31,7 +31,8 @@ export class RegistrarGuard {
       return true;
     }
     // Get the full URL including any nested children (skip the initial '#/')
-    const nextUrl = location.hash.split('#/')[1];
+    // NB: an empty nextUrl takes the user to the home page
+    const nextUrl = location.hash.split('#/')[1] || '';
     return this.router.navigate([`/empty-registrar`, { nextUrl }]);
   }
 }


ptkach (changes requested):
Have you tested it on alpha? I don't think that's what we want, as far as I can predict it's going to navigate to "home page" instead of an "empty registrar page", which is what we want

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2149#-:-NeKVQR-4BJnfZkainoP:b-s9u011)* status: 0 of 1 files reviewed, all discussions resolved

gbrodman:
just tested this locally and it seems like we just want to tell it go to to nothing (the home page)

ptkach (commented):
we do want it to go to home page after selecting the registrar, you're correct. But that's not what your change does, it allows it to go to home page before making any registrar selection and it goes straight to home page, while it has to go to empty registrar page instead. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2149#-:-NeKdc5y01meI8ecNG0k:b-bruy1a)* status: 1 of 2 files reviewed, all discussions resolved

gbrodman (commented):
I don't think you saw the update to the code that I made when writing my previous comment. because that's what it does now. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2149#-:-NeL_kuU8mf9Vr3Hq8NH:b-bt6xud)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @ptkach)

ptkach (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2149)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 35 at r2](https://reviewable.io/reviews/google/nomulus/2149#-NeOBGO7AtKn9TyYeAqQ:-NeOBGO87LZDVLQipOHg:bw7a309) ([raw file](https://github.com/google/nomulus/blob/4ab472d6aa152c49562ad60e6b03ac75ea3ad0f2/console-webapp/src/app/registrar/registrar.guard.ts#L35)):*
> ```XML
>     // Get the full URL including any nested children (skip the initial '#/')
>     const splits = location.hash.split('#/');
>     const nextUrl = splits.length > 1 ? splits[1] : '';
> ```

Oh, I see. I think it requires a comment that empty nextUrl takes user to a router default page, which in our case is a home page.
Also it can be shortened I believe - `const nextUrl = location.hash.split('#/')[1] || ""`

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2149)* status: all files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/registrar/registrar.guard.ts` line 35 at r2](https://reviewable.io/reviews/google/nomulus/2149#-NeOBGO7AtKn9TyYeAqQ:-NeOOK9N91FdYfTFoXvZ:b-inwiso) ([raw file](https://github.com/google/nomulus/blob/4ab472d6aa152c49562ad60e6b03ac75ea3ad0f2/console-webapp/src/app/registrar/registrar.guard.ts#L35)):*

Oh, I see. I think it requires a comment that empty nextUrl takes user to a router default page, which in our case is a home page.
Also it can be shortened I believe - `const nextUrl = location.hash.split('#/')[1] || ""`
</blockquote></details>

huh I did not know that you could use `||` like that, interesting

ptkach (approved):
Yeah and no IndexOutOfBounds :)

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2149#-:-NeOQohO0Y_pkKdtW7_-:b-p86pwn)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @gbrodman)

PR #2150, opened 2023-09-14
Defend against deserialization-based attacks

weiminyu:
Defend against deserialization-based attacks
    
Added the `SafeObjectInputStream` class that defends against attacks using
malformed serialized data, including remote code execution and
denial-of-service attacks.

Started using the new class to handle EPP resource VKeys and
PendingDeposits, which are passed across credential-boundaries: between
TaskQueue and AppEngine server, and between AppEngine server and the RDE
pipeline on GCE. Note that the wireformat of VKeys do not change,
therefore existing tasks sitting in the TaskQueue are not affected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2150)


jianglai (changes requested):
> defends attacks

defends against attacks

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2150#-:-Neclp5D3vHB9j7FhvGJ:b-tubn7c)* status: 7 of 11 files reviewed, 10 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 118 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedpxN6BGyaj7zDl90m:-NedpxN6BGyaj7zDl90n:b-ec5z7t) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L118)):*
> ```Java
>   private static class SerializedForm implements Serializable {
> 
>     private static final long serialVersionUID = 1L;
> ```

Can't we use a randomly generated number here?
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 146 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedvEpH4I1Y6UXWenfQ:-Nedsybv5ZCkiSe9kxI3:buohyya) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L146)):*
> ```Java
>    *
>    * <p>We cannot use a {@code SerializableCoder} directly for two reasons: it does not guarantee
>    * determinism, which is required by GroupBy; and the Joda-time classes are not allowed by {@link
> ```

If Joda Time is the only offender here, why do we need to the coders for other classes like `RdeMode` and `CursorType`?
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 183 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Nedx-Rm1ipRPruH60c9:-Nedx-Rm1ipRPruH60cA:b-sr063q) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L183)):*
> ```Java
>           BooleanCoder.of().decode(inStream),
>           StringUtf8Coder.of().decode(inStream),
>           DateTime.parse(StringUtf8Coder.of().decode(inStream)),
> ```

I don't fully understand how serialization works, so this might be a bit naive: here you are returning a class with a joda time as one of the field. Assuming that the program works, this obviously didn't go through `SafeInputObjectStream.resolveClass()`, otherwise it would have resulted in an exception.

So it looks to me that as long as you override the `readObject` method for the class in question (in this case relying on Beam coders), you don't have to rely on Java to resolve the class to be deserialized and therefore bypass the check. So why is it that we need to be careful not to serialize a Joda time in the first place? The checks only apply during deserialization,, and we I don't see a reason why just calling `SerializableCoder.of(DateTime.class)` would not work?

Futher more, by that logic, why do we have to provide a `SerializableForm` which just stores `PendingDeposit` as a delegate? Can't we override `readObject` and `writeObject` on `PendingDeposit` and use the `encode` and `decode` method from the coder? I guess it's a bit more work for decoding as you'd get another `PendingDeposit` and have to set the value of each fields of `this`?
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 88 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Necqz3N86vQEExkAaTC:-Necqz3N86vQEExkAaTD:buc0xb8) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L88)):*
> ```Java
> 
>   private boolean isNomulusClass(String clazz) {
>     return clazz.startsWith("google.registry.");
> ```

What's preventing a compromised third party dependency to include a class that starts with "google.registry"? My understanding is that as long as that class does not conflict with existing legitimate classes, it will not result in any error during initialization. We can't prevent others from using the namespace. 
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 92 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NecsE--40ndlSeWI9XX:-NecsE--40ndlSeWI9XY:b-bmafom) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L92)):*
> ```Java
> 
>   private boolean isContainer(Class<?> clazz) {
>     return Collection.class.isAssignableFrom(clazz) || Map.class.isAssignableFrom(clazz);
> ```

Similarly, what if the compromised jar contains a re-implementation of `ArrayList` without implementing the `List` interface? It would be able to sidestep this check and result in DoS attacks, in theory, right?
___
*[`util/src/main/java/google/registry/util/SafeSerializationUtils.java` line 55 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Ned5L-CBwu_47YwBboO:-Ned5L-DCX8UINs4-7pO:buznixl) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeSerializationUtils.java#L55)):*
> ```Java
>    * <p>If any element of the collection cannot be safely-deserialized, deserialization will fail.
>    */
>   public static byte[] serializeCollection(Collection<?> collection) {
> ```

Just to confirm, this would only allow serialization/deserialization of top-level collections, right? You still cannot have collection of collections.
___
*[`util/src/main/java/google/registry/util/SafeSerializationUtils.java` line 87 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedPu-GBPIu-7ph_pPx:-NedPu-GBPIu-7ph_pPy:bz7ewb2) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeSerializationUtils.java#L87)):*
> ```Java
>    * #serializeCollection}.
>    */
>   public static <T> ImmutableList<T> safeDeserializeCollection(Class<T> elementType, byte[] bytes) {
> ```

Is it necessary to provide the element type? How does normal deserialization work? I don't think you need to provide the element type if you rely on built-in deserialization. Can we not save the type info in `serializeCollection"?
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 66 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedOfpK84-_ow1cjYe8:-NedOfpL7Xv95er7Ev0H:bb76mu4) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L66)):*
> ```Java
>     try (SafeObjectInputStream sois =
>         new SafeObjectInputStream(new ByteArrayInputStream(serialize(orig)))) {
>       // For array, The ClassNotFoundException thrown by us may be converted to something else.
> ```

Why?
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 81 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedP4df9ZRwA_4ykdMX:-NedP4df9ZRwA_4ykdMY:b-jnyyt4) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L81)):*
> ```Java
> 
>   @Test
>   void nomulusEntitySuccess() throws Exception {
> ```

Need a test for non-Nomulus, non-collection objects.
___
*[`util/src/test/java/google/registry/util/SafeSerializationUtilsTest.java` line 35 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedQZrN6nj9-2waCaKI:-NedQZrN6nj9-2waCaKJ:byv8o6t) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeSerializationUtilsTest.java#L35)):*
> ```Java
> 
>   @Test
>   void deserialize_array_failure() {
> ```

I think some of the tests here are redundant, as the behavior of deserializing arrays, lists, maps, etc are already tested in `SafeObjectInputStream`.

If we do want to duplicate the tests here, shouldn't be duplicate all the cases, including deserializing nomulus classes, java classes, and other classes?

weiminyu (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2150#-:-NeiespJ7c7BBWvIAnEq:bb6un1j)* status: 7 of 11 files reviewed, 10 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 118 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedpxN6BGyaj7zDl90m:-NeecJhP7T_Y0KVRTX44:b-vgiwo7) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L118)):*

Can't we use a randomly generated number here?
</blockquote></details>

Are there good reasons for using random numbers? I don't see any when adding new classes.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 146 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedvEpH4I1Y6UXWenfQ:-NeiqG6-0BJNerrHVOoX:bvdl9cm) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L146)):*

If Joda Time is the only offender here, why do we need to the coders for other classes like `RdeMode` and `CursorType`?
</blockquote></details>

Updated the comments. SerializableCoder calls `java.io.ObjectInputStream`, which is not safe against attacks.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 183 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Nedx-Rm1ipRPruH60c9:-NeeehoQBagvj7kEWavH:b2wwt2e) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L183)):*

I don't fully understand how serialization works, so this might be a bit naive: here you are returning a class with a joda time as one of the field. Assuming that the program works, this obviously didn't go through `SafeInputObjectStream.resolveClass()`, otherwise it would have resulted in an exception.

So it looks to me that as long as you override the `readObject` method for the class in question (in this case relying on Beam coders), you don't have to rely on Java to resolve the class to be deserialized and therefore bypass the check. So why is it that we need to be careful not to serialize a Joda time in the first place? The checks only apply during deserialization,, and we I don't see a reason why just calling `SerializableCoder.of(DateTime.class)` would not work?

Futher more, by that logic, why do we have to provide a `SerializableForm` which just stores `PendingDeposit` as a delegate? Can't we override `readObject` and `writeObject` on `PendingDeposit` and use the `encode` and `decode` method from the coder? I guess it's a bit more work for decoding as you'd get another `PendingDeposit` and have to set the value of each fields of `this`?
</blockquote></details>

Updated the comments. PendingDeposit is serialized in two ways, in the RdePipeline using Beam serialization with Beam coders; and as a field in the RdeOptions class, using normal java serialization. This PR addresses the second case. The change to PendingDepositCoder is to support the second case, so that we don't have to write another set of serialization/deserialization code.

We cannot overide readObject/writeObject in `PendingDepoisit`. What we are really dealing with is `AutoValue_PendingDeposit`, whose fields are `private final`. A proxy like `SerializedForm` is the only way to support custom serialization of immutable class with builders.

There is a valid alternative to drop serialization altogether, and explicitly use the PendingDepositCoder to convert PendingDeposit instances to/from string. However, this is the case where a collection of instances are sent, so we will have to remember to perform the collection size check, which is easy to forget. I think doing it in one place in SafeSerializationUtils is better.









 
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 88 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Necqz3N86vQEExkAaTC:-NeejKdz617_3dlshujL:byl3rdm) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L88)):*

What's preventing a compromised third party dependency to include a class that starts with "google.registry"? My understanding is that as long as that class does not conflict with existing legitimate classes, it will not result in any error during initialization. We can't prevent others from using the namespace.
</blockquote></details>

I would guess that most such attempts are general purpose instead of targeted so a `google.registry` class in a third-party jar is not likely.
If we are paranoid enough, since we already scan all third-party classes for java version, we can easily check them for their packages.
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 92 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NecsE--40ndlSeWI9XX:-Neekskp6BN6rx7JbfvH:b8v8h9q) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L92)):*

Similarly, what if the compromised jar contains a re-implementation of `ArrayList` without implementing the `List` interface? It would be able to sidestep this check and result in DoS attacks, in theory, right?
</blockquote></details>

They'd have to slip it in as a `google.registry` class in some third-party jars, which is preventable.
___
*[`util/src/main/java/google/registry/util/SafeSerializationUtils.java` line 55 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Ned5L-CBwu_47YwBboO:-Neel7Ph6tqHK9LU30d4:bb77ufg) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeSerializationUtils.java#L55)):*

Just to confirm, this would only allow serialization/deserialization of top-level collections, right? You still cannot have collection of collections.
</blockquote></details>

Yes.
___
*[`util/src/main/java/google/registry/util/SafeSerializationUtils.java` line 87 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedPu-GBPIu-7ph_pPx:-NeelDqX3kaJeDmb-O43:bkeawe9) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeSerializationUtils.java#L87)):*

Is it necessary to provide the element type? How does normal deserialization work? I don't think you need to provide the element type if you rely on built-in deserialization. Can we not save the type info in `serializeCollection"?
</blockquote></details>

It just gives a nice type signature without casting of collection with type signature.  Casts like `(ImmutableList<PendingDeposit>)` is not safe if we are not certain about the element type.
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 66 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedOfpK84-_ow1cjYe8:-NeelLqw3d0zvBduY0Pf:b-544znx) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L66)):*

Why?
</blockquote></details>

Updated the comment. Just the internal implementation of the Java's ObjectInputStream.
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 81 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedP4df9ZRwA_4ykdMX:-Neelsmd277euI7dXeo1:b-fgyj3p) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L81)):*

Need a test for non-Nomulus, non-collection objects.
</blockquote></details>

Renamed `nonJavaUnitaryFailure` two methods above to `nonJavaNonNomulusUnitaryFailure`
___
*[`util/src/test/java/google/registry/util/SafeSerializationUtilsTest.java` line 35 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedQZrN6nj9-2waCaKI:-NeemFhmEuIcMeDgcRtd:bepdggi) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeSerializationUtilsTest.java#L35)):*

I think some of the tests here are redundant, as the behavior of deserializing arrays, lists, maps, etc are already tested in `SafeObjectInputStream`.

If we do want to duplicate the tests here, shouldn't be duplicate all the cases, including deserializing nomulus classes, java classes, and other classes?
</blockquote></details>

These methods test the interface of SerializationUtils: user can pass in a list of `safe` objects but not an array of them. But someday we may support sending an array of `safe` objects.

weiminyu (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2150#-:-NeiespJ7c7BBWvIAnEq:bb6un1j)* status: 7 of 11 files reviewed, 10 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 118 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedpxN6BGyaj7zDl90m:-NeecJhP7T_Y0KVRTX44:b-vgiwo7) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L118)):*

Can't we use a randomly generated number here?
</blockquote></details>

Are there good reasons for using random numbers? I don't see any when adding new classes.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 146 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedvEpH4I1Y6UXWenfQ:-NeiqG6-0BJNerrHVOoX:bvdl9cm) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L146)):*

If Joda Time is the only offender here, why do we need to the coders for other classes like `RdeMode` and `CursorType`?
</blockquote></details>

Updated the comments. SerializableCoder calls `java.io.ObjectInputStream`, which is not safe against attacks.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 183 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Nedx-Rm1ipRPruH60c9:-NeeehoQBagvj7kEWavH:b2wwt2e) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L183)):*

I don't fully understand how serialization works, so this might be a bit naive: here you are returning a class with a joda time as one of the field. Assuming that the program works, this obviously didn't go through `SafeInputObjectStream.resolveClass()`, otherwise it would have resulted in an exception.

So it looks to me that as long as you override the `readObject` method for the class in question (in this case relying on Beam coders), you don't have to rely on Java to resolve the class to be deserialized and therefore bypass the check. So why is it that we need to be careful not to serialize a Joda time in the first place? The checks only apply during deserialization,, and we I don't see a reason why just calling `SerializableCoder.of(DateTime.class)` would not work?

Futher more, by that logic, why do we have to provide a `SerializableForm` which just stores `PendingDeposit` as a delegate? Can't we override `readObject` and `writeObject` on `PendingDeposit` and use the `encode` and `decode` method from the coder? I guess it's a bit more work for decoding as you'd get another `PendingDeposit` and have to set the value of each fields of `this`?
</blockquote></details>

Updated the comments. PendingDeposit is serialized in two ways, in the RdePipeline using Beam serialization with Beam coders; and as a field in the RdeOptions class, using normal java serialization. This PR addresses the second case. The change to PendingDepositCoder is to support the second case, so that we don't have to write another set of serialization/deserialization code.

We cannot overide readObject/writeObject in `PendingDepoisit`. What we are really dealing with is `AutoValue_PendingDeposit`, whose fields are `private final`. A proxy like `SerializedForm` is the only way to support custom serialization of immutable class with builders.

There is a valid alternative to drop serialization altogether, and explicitly use the PendingDepositCoder to convert PendingDeposit instances to/from string. However, this is the case where a collection of instances are sent, so we will have to remember to perform the collection size check, which is easy to forget. I think doing it in one place in SafeSerializationUtils is better.









 
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 88 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Necqz3N86vQEExkAaTC:-NeejKdz617_3dlshujL:byl3rdm) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L88)):*

What's preventing a compromised third party dependency to include a class that starts with "google.registry"? My understanding is that as long as that class does not conflict with existing legitimate classes, it will not result in any error during initialization. We can't prevent others from using the namespace.
</blockquote></details>

I would guess that most such attempts are general purpose instead of targeted so a `google.registry` class in a third-party jar is not likely.
If we are paranoid enough, since we already scan all third-party classes for java version, we can easily check them for their packages.
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 92 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NecsE--40ndlSeWI9XX:-Neekskp6BN6rx7JbfvH:b8v8h9q) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L92)):*

Similarly, what if the compromised jar contains a re-implementation of `ArrayList` without implementing the `List` interface? It would be able to sidestep this check and result in DoS attacks, in theory, right?
</blockquote></details>

They'd have to slip it in as a `google.registry` class in some third-party jars, which is preventable.
___
*[`util/src/main/java/google/registry/util/SafeSerializationUtils.java` line 55 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Ned5L-CBwu_47YwBboO:-Neel7Ph6tqHK9LU30d4:bb77ufg) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeSerializationUtils.java#L55)):*

Just to confirm, this would only allow serialization/deserialization of top-level collections, right? You still cannot have collection of collections.
</blockquote></details>

Yes.
___
*[`util/src/main/java/google/registry/util/SafeSerializationUtils.java` line 87 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedPu-GBPIu-7ph_pPx:-NeelDqX3kaJeDmb-O43:bkeawe9) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeSerializationUtils.java#L87)):*

Is it necessary to provide the element type? How does normal deserialization work? I don't think you need to provide the element type if you rely on built-in deserialization. Can we not save the type info in `serializeCollection"?
</blockquote></details>

It just gives a nice type signature without casting of collection with type signature.  Casts like `(ImmutableList<PendingDeposit>)` is not safe if we are not certain about the element type.
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 66 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedOfpK84-_ow1cjYe8:-NeelLqw3d0zvBduY0Pf:b-544znx) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L66)):*

Why?
</blockquote></details>

Updated the comment. Just the internal implementation of the Java's ObjectInputStream.
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 81 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedP4df9ZRwA_4ykdMX:-Neelsmd277euI7dXeo1:b-fgyj3p) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L81)):*

Need a test for non-Nomulus, non-collection objects.
</blockquote></details>

Renamed `nonJavaUnitaryFailure` two methods above to `nonJavaNonNomulusUnitaryFailure`
___
*[`util/src/test/java/google/registry/util/SafeSerializationUtilsTest.java` line 35 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedQZrN6nj9-2waCaKI:-NeemFhmEuIcMeDgcRtd:bepdggi) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeSerializationUtilsTest.java#L35)):*

I think some of the tests here are redundant, as the behavior of deserializing arrays, lists, maps, etc are already tested in `SafeObjectInputStream`.

If we do want to duplicate the tests here, shouldn't be duplicate all the cases, including deserializing nomulus classes, java classes, and other classes?
</blockquote></details>

These methods test the interface of SerializationUtils: user can pass in a list of `safe` objects but not an array of them. But someday we may support sending an array of `safe` objects.

jianglai (approved):
> Reviewed 2 of 4 files at r2, 1 of 1 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2150)* status: all files reviewed, 7 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 118 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedpxN6BGyaj7zDl90m:-NenOvIs7b7fXj7TBihp:bw6jzw2) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L118)):*

Are there good reasons for using random numbers? I don't see any when adding new classes.
</blockquote></details>

The same class name with different serial version ID would be treated as different classes. I'm not sure what would happen if an automatically generated one (when none is specified) collide with 1L. I doubt it's anything bad, but you IDE should easily offer generating a random one for you.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 183 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Nedx-Rm1ipRPruH60c9:-NenlSk6CYt2wKMruJ40:bga3xv9) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L183)):*

Updated the comments. PendingDeposit is serialized in two ways, in the RdePipeline using Beam serialization with Beam coders; and as a field in the RdeOptions class, using normal java serialization. This PR addresses the second case. The change to PendingDepositCoder is to support the second case, so that we don't have to write another set of serialization/deserialization code.

We cannot overide readObject/writeObject in `PendingDepoisit`. What we are really dealing with is `AutoValue_PendingDeposit`, whose fields are `private final`. A proxy like `SerializedForm` is the only way to support custom serialization of immutable class with builders.

There is a valid alternative to drop serialization altogether, and explicitly use the PendingDepositCoder to convert PendingDeposit instances to/from string. However, this is the case where a collection of instances are sent, so we will have to remember to perform the collection size check, which is easy to forget. I think doing it in one place in SafeSerializationUtils is better.
</blockquote></details>

Ah, I think the part that I missed is that it's the `AutoValue` class that is actually serialized/deserialized.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 110 at r4](https://reviewable.io/reviews/google/nomulus/2150#-Nenm7X_EjBj_RNyObmt:-Nenm7X_EjBj_RNyObmu:b-bgdamy) ([raw file](https://github.com/google/nomulus/blob/023eba874a676b58b661500d1a5f4b1f2e3bb039/core/src/main/java/google/registry/rde/PendingDeposit.java#L110)):*
> ```Java
> 
>   /**
>    * Specifies that {@link SerializedForm} be used for @code SafeObjectInputStream}-compatible
> ```

Missing an open paren.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 111 at r4](https://reviewable.io/reviews/google/nomulus/2150#-NenmGAnDWDpuFApuhug:-NenmGAnDWDpuFApuhuh:b-4eckom) ([raw file](https://github.com/google/nomulus/blob/023eba874a676b58b661500d1a5f4b1f2e3bb039/core/src/main/java/google/registry/rde/PendingDeposit.java#L111)):*
> ```Java
>   /**
>    * Specifies that {@link SerializedForm} be used for @code SafeObjectInputStream}-compatible
>    * custom-serialization of {@link AutoValue_PendingDeposit the AutoValue implementation class}.
> ```

Missing a close paren.
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 88 at r1](https://reviewable.io/reviews/google/nomulus/2150#-Necqz3N86vQEExkAaTC:-NenoQZq00tPfTyL_Gxu:b8njk2p) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L88)):*

I would guess that most such attempts are general purpose instead of targeted so a `google.registry` class in a third-party jar is not likely.
If we are paranoid enough, since we already scan all third-party classes for java version, we can easily check them for their packages.
</blockquote></details>

I guess that's a good enough solution, though a bit reactive. 
___
*[`util/src/main/java/google/registry/util/SafeObjectInputStream.java` line 73 at r2](https://reviewable.io/reviews/google/nomulus/2150#-Neno9N_6V15y3q_or4B:-Neno9N_6V15y3q_or4C:b7yp2qp) ([raw file](https://github.com/google/nomulus/blob/dbd10f93951d412f9042b486bd3ae859a9e76273/util/src/main/java/google/registry/util/SafeObjectInputStream.java#L73)):*
> ```Java
>       return checkNotArrayOrContainer(super.resolveClass(desc));
>     }
>     // TODO(b/297587959): after upgrading to Java 11, consider replace if-block below with
> ```

I don't understand why you split the if block into two. You are calling the same checks in both blocks.
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 66 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedOfpK84-_ow1cjYe8:-NenoyDy5EdoDN2XuIUk:brl86qt) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L66)):*

Updated the comment. Just the internal implementation of the Java's ObjectInputStream.
</blockquote></details>

Does the exception thrown contain `ClassNotFoundException` as its cause? You can assert on that.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2150)* status: 10 of 11 files reviewed, 7 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 118 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedpxN6BGyaj7zDl90m:-Neo9-595rrrgkQ7z8ko:b-skul6h) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/core/src/main/java/google/registry/rde/PendingDeposit.java#L118)):*

The same class name with different serial version ID would be treated as different classes. I'm not sure what would happen if an automatically generated one (when none is specified) collide with 1L. I doubt it's anything bad, but you IDE should easily offer generating a random one for you.
</blockquote></details>

With a new class there won't be earlier versions needing automatically generated ones.

I guess this is one of those subjective matters that have no definite answer, and Google best practice guide
is silent on this. I'll follow existing code then.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 110 at r4](https://reviewable.io/reviews/google/nomulus/2150#-Nenm7X_EjBj_RNyObmt:-Neo6s709HwvwM12olib:b-896fix) ([raw file](https://github.com/google/nomulus/blob/023eba874a676b58b661500d1a5f4b1f2e3bb039/core/src/main/java/google/registry/rde/PendingDeposit.java#L110)):*

Missing an open paren.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/rde/PendingDeposit.java` line 111 at r4](https://reviewable.io/reviews/google/nomulus/2150#-NenmGAnDWDpuFApuhug:-Neo6wPsEDI7Wcdqj05J:bt54sl2) ([raw file](https://github.com/google/nomulus/blob/023eba874a676b58b661500d1a5f4b1f2e3bb039/core/src/main/java/google/registry/rde/PendingDeposit.java#L111)):*

Missing a close paren.
</blockquote></details>

This intentional. The phrase after class name is the intended display text.
___
*[`util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java` line 66 at r1](https://reviewable.io/reviews/google/nomulus/2150#-NedOfpK84-_ow1cjYe8:-Neo7PkbCfFuZHb5Nz32:btspqji) ([raw file](https://github.com/google/nomulus/blob/06703ccc56dd9e141d8126fa758fd7606dd37fd6/util/src/test/java/google/registry/util/SafeObjectInputStreamTest.java#L66)):*

Does the exception thrown contain `ClassNotFoundException` as its cause? You can assert on that.
</blockquote></details>

No it stores it, then finds the input stream corrupt before checking its existence. Seems like a bug to me.

PR #2151, opened 2023-09-18
Add resources widget front-end, enable settings widget navigation

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2151)

### DIFF
--- a/console-webapp/src/app/app-routing.module.ts
+++ b/console-webapp/src/app/app-routing.module.ts
@@ -24,6 +24,10 @@ import SettingsSecurityComponent from './settings/security/security.component';
 import { RegistrarGuard } from './registrar/registrar.guard';
 import { RegistrarComponent } from './registrar/registrarsTable.component';
 import { EmptyRegistrar } from './registrar/emptyRegistrar.component';
+import ContactComponent from './settings/contact/contact.component';
+import WhoisComponent from './settings/whois/whois.component';
+import SecurityComponent from './settings/security/security.component';
+import UsersComponent from './settings/users/users.component';
 
 const routes: Routes = [
   { path: '', redirectTo: '/home', pathMatch: 'full' },
@@ -32,7 +36,7 @@ const routes: Routes = [
   { path: 'home', component: HomeComponent, canActivate: [RegistrarGuard] },
   { path: 'tlds', component: TldsComponent, canActivate: [RegistrarGuard] },
   {
-    path: 'settings',
+    path: SettingsComponent.PATH,
     component: SettingsComponent,
     children: [
       {
@@ -41,32 +45,27 @@ const routes: Routes = [
         pathMatch: 'full',
       },
       {
-        path: 'contact',
+        path: ContactComponent.PATH,
         component: SettingsContactComponent,
         canActivate: [RegistrarGuard],
       },
       {
-        path: 'whois',
+        path: WhoisComponent.PATH,
         component: SettingsWhoisComponent,
         canActivate: [RegistrarGuard],
       },
       {
-        path: 'security',
+        path: SecurityComponent.PATH,
         component: SettingsSecurityComponent,
         canActivate: [RegistrarGuard],
       },
       {
-        path: 'epp-password',
-        component: SettingsSecurityComponent,
-        canActivate: [RegistrarGuard],
-      },
-      {
-        path: 'users',
+        path: UsersComponent.PATH,
         component: SettingsUsersComponent,
         canActivate: [RegistrarGuard],
       },
       {
-        path: 'registrars',
+        path: RegistrarComponent.PATH,
         component: RegistrarComponent,
       },
     ],
--- a/console-webapp/src/app/app.component.ts
+++ b/console-webapp/src/app/app.component.ts
@@ -14,6 +14,7 @@
 
 import { Component } from '@angular/core';
 import { RegistrarService } from './registrar/registrar.service';
+import { UserDataService } from './shared/services/userData.service';
 import { GlobalLoaderService } from './shared/services/globalLoader.service';
 
 @Component({
@@ -25,6 +26,7 @@ export class AppComponent {
   renderRouter: boolean = true;
   constructor(
     protected registrarService: RegistrarService,
+    protected userDataService: UserDataService,
     protected globalLoader: GlobalLoaderService
   ) {
     registrarService.activeRegistrarIdChange.subscribe(() => {
--- a/console-webapp/src/app/app.module.ts
+++ b/console-webapp/src/app/app.module.ts
@@ -46,6 +46,7 @@ import { EppWidgetComponent } from './home/widgets/epp-widget.component';
 import { BillingWidgetComponent } from './home/widgets/billing-widget.component';
 import { DomainsWidgetComponent } from './home/widgets/domains-widget.component';
 import { SettingsWidgetComponent } from './home/widgets/settings-widget.component';
+import { UserDataService } from './shared/services/userData.service';
 
 @NgModule({
   declarations: [
@@ -81,6 +82,7 @@ import { SettingsWidgetComponent } from './home/widgets/settings-widget.componen
     BackendService,
     GlobalLoaderService,
     RegistrarGuard,
+    UserDataService,
     {
       provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
       useValue: {
--- a/console-webapp/src/app/home/widgets/contact-widget.component.html
+++ b/console-webapp/src/app/home/widgets/contact-widget.component.html
@@ -13,13 +13,13 @@
           Give us a Call
         </button>
         <p class="secondary-text">
-          Call Google Registry support at +1 (404) 978 8419
+          Call Google Registry support at <b>+1 (404) 978 8419</b>
         </p>
         <button mat-button color="primary" class="console-app__widget-link">
           Send us an Email
         </button>
         <p class="secondary-text">
-          Email Google Registry at support@google.com
+          Email Google Registry at <b>support@google.com</b>
         </p>
       </div>
     </div>
--- a/console-webapp/src/app/home/widgets/resources-widget.component.html
+++ b/console-webapp/src/app/home/widgets/resources-widget.component.html
@@ -1,13 +1,16 @@
 <mat-card>
   <mat-card-content>
     <div class="console-app__widget">
-      <div class="console-app__widget_left">
+      <a
+        class="console-app__widget_left"
+        href="{{ userDataService.userData?.technicalDocsUrl }}"
+      >
         <mat-icon class="console-app__widget-icon">menu_book</mat-icon>
         <h1 class="console-app__widget-title">Resources</h1>
         <h4 class="secondary-text text-center">
           Use Google Drive to view onboarding FAQs, and technical documentation.
         </h4>
-      </div>
+      </a>
     </div>
   </mat-card-content>
 </mat-card>
--- a/console-webapp/src/app/home/widgets/resources-widget.component.ts
+++ b/console-webapp/src/app/home/widgets/resources-widget.component.ts
@@ -13,11 +13,12 @@
 // limitations under the License.
 
 import { Component } from '@angular/core';
+import { UserDataService } from 'src/app/shared/services/userData.service';
 
 @Component({
   selector: '[app-resources-widget]',
   templateUrl: './resources-widget.component.html',
 })
 export class ResourcesWidgetComponent {
-  constructor() {}
+  constructor(public userDataService: UserDataService) {}
 }
--- a/console-webapp/src/app/home/widgets/settings-widget.component.html
+++ b/console-webapp/src/app/home/widgets/settings-widget.component.html
@@ -10,11 +10,21 @@
         </h4>
       </div>
       <div class="console-app__widget_right">
-        <button mat-button color="primary" class="console-app__widget-link">
+        <button
+          mat-button
+          color="primary"
+          class="console-app__widget-link"
+          (click)="openContactsPage()"
+        >
           Contact Information
         </button>
         <p class="secondary-text">Manage Primary, Technical, etc contacts.</p>
-        <button mat-button color="primary" class="console-app__widget-link">
+        <button
+          mat-button
+          color="primary"
+          class="console-app__widget-link"
+          (click)="openSecurityPage()"
+        >
           Security
         </button>
         <p class="secondary-text">
@@ -28,7 +38,12 @@
           User Management
         </button>
         <p class="secondary-text">Create and manage console user accounts</p>
-        <button mat-button color="primary" class="console-app__widget-link">
+        <button
+          mat-button
+          color="primary"
+          class="console-app__widget-link"
+          (click)="openRegistrarsPage()"
+        >
           Registrar Management
         </button>
         <p class="secondary-text">Create and manage registrar accounts</p>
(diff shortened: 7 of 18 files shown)


gbrodman (changes requested):
> Reviewed 4 of 7 files at r1, 9 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 13 of 18 files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/home/widgets/contact-widget.component.html` line 16 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeieRcQ1ZVfodBTN0Hm:-NeieRcQ1ZVfodBTN0Hn:bv1ircz) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/contact-widget.component.html#L16)):*
> ```HTML
>         </button>
>         <p class="secondary-text">
>           Call Google Registry support at <b>+1 (404) 978 8419</b>
> ```

won't be in this PR, just flagging that we'll want to make this configurable as well 
___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-NeiemEg-XrTpdNUAh7h:b-fyvmo6) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*
> ```HTML
>       <a
>         class="console-app__widget_left"
>         href="{{ userDataService.userData?.technicalDocsUrl }}"
> ```

i think we possibly might want to introduce a separate endpoint for retrieving all the informational fields e.g. technical docs URL, support phone number and email address

It seems like that'd fit better than including it in the user information.

Thoughts?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 13 of 18 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-Neig5E01E-Ez_aAhBxR:btxugyr) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*

i think we possibly might want to introduce a separate endpoint for retrieving all the informational fields e.g. technical docs URL, support phone number and email address

It seems like that'd fit better than including it in the user information.

Thoughts?
</blockquote></details>

Although I agree that having a separate endpoint might be better in ideal circumstances, than trying to fit technical docs url, phone num, email etc into user data, but due the **specifics of our situation** I don't think we should do that. Few things we need to consider:
1) This data immediately available and doesn't require no  DB queries, etc
2) We can make an argument that this data is to be shown pretty much all the time, as it's positioned on the home (default) page
3) We require at least the following data from the server in order to provide meaningful functionality to the user:
    3.1) list of registrars (in order to render a selector); 
    3.2) user data (in order to know what widgets/pages available to the user);
    3.3) misc data, like tech url, etc (in order to render widgets like resources, contacts )

Now, without all that data we can not make webapp fully available to the user and given that we already have at least 2 blocking requests (3.1 and 3.2) I strongly prefer not to make another blocking request given that the data for it is immediately available to us (see 1)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 13 of 18 files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-Neipmov9tPtwyG2NTjJ:b-8jdk23) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*

Although I agree that having a separate endpoint might be better in ideal circumstances, than trying to fit technical docs url, phone num, email etc into user data, but due the **specifics of our situation** I don't think we should do that. Few things we need to consider:
1) This data immediately available and doesn't require no  DB queries, etc
2) We can make an argument that this data is to be shown pretty much all the time, as it's positioned on the home (default) page
3) We require at least the following data from the server in order to provide meaningful functionality to the user:
    3.1) list of registrars (in order to render a selector); 
    3.2) user data (in order to know what widgets/pages available to the user);
    3.3) misc data, like tech url, etc (in order to render widgets like resources, contacts )

Now, without all that data we can not make webapp fully available to the user and given that we already have at least 2 blocking requests (3.1 and 3.2) I strongly prefer not to make another blocking request given that the data for it is immediately available to us (see 1)
</blockquote></details>

I think the immediate availability of the simple documentation data indicates even more so that it should be a separate network request. Because it doesn't require a DB query it should return pretty much immediately, meaning that it wouldn't block the loading of the front end (while we wait for registrars and users info). 

I don't see any other reasoning not to make the documentation data a separate request -- the fact that it's frequently visible doesn't seem to make a difference to me. 

Plus, we'll need to continue adding more information to it (and possibly more information to the users API), such as the area that I commented in elsewhere in this PR.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 13 of 18 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-NeisBjHD0HUrQmYGa2I:b-anp3yl) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*
Still not convinced. See my comments below:
Before letting user to open home page we need to gather 3 things - registrars, user data, misc. Order of those things is irrelevant because the requests have to go out in parallel not sequential, regardless whether it's 2 requests or 3, otherwise the wait time will grow event more. 
Now you say:
> it wouldn't block the loading of the front end (while we wait for registrars and users info).
Well it would, because after all it's an extra http request, which aside from obvious network overhead doesn't have to be processed in order it was sent. That and the fact that handling that on front-end will become more complicated.

As I mentioned I see where you're coming from, but I don't think the downside in this case worth data separation even if we continue to extend the data in response.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 13 of 18 files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-Neiugu-BfbEUdRJ7H-7:b2rh66o) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*

Still not convinced. See my comments below:
Before letting user to open home page we need to gather 3 things - registrars, user data, misc. Order of those things is irrelevant because the requests have to go out in parallel not sequential, regardless whether it's 2 requests or 3, otherwise the wait time will grow event more. 
Now you say:
> it wouldn't block the loading of the front end (while we wait for registrars and users info).
Well it would, because after all it's an extra http request, which aside from obvious network overhead doesn't have to be processed in order it was sent. That and the fact that handling that on front-end will become more complicated.

As I mentioned I see where you're coming from, but I don't think the downside in this case worth data separation even if we continue to extend the data in response.
</blockquote></details>

My point is that I don't believe the addition of the third very-simple request for the documentation information will actually slow things down, since we're going to be bound by the other two requests that are going out in parallel (registrars and users) no matter what. 

At least, I'd hope they go out in parallel!

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 13 of 18 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-Nej04HX3LBAzfI-A1i6:b6aouxn) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*

My point is that I don't believe the addition of the third very-simple request for the documentation information will actually slow things down, since we're going to be bound by the other two requests that are going out in parallel (registrars and users) no matter what. 

At least, I'd hope they go out in parallel!
</blockquote></details>

It will slow things down, I don't see how it wouldn't as it's extra http request. Even sending 3 requests (/registrars, /userdata and say /miscdata) in parallel (what you're suggesting here) we have no guarantees that /miscdata will be processed first, in fact it might as well be received and processed last by the server depending on the network trafficking, that's what I've been trying to tell.

We're talking here about things that directly affect TTI of our web app - https://developer.chrome.com/en/docs/lighthouse/performance/interactive/ and I'd very much prefer to stay in the green zone.

My point is that given that we make user wait for the entire resource loading (html, css, js, etc) and necessary data (registrars, users) already, which we can't do anything anyway, we should consider not making another blocking (read TTI affecting) request unless it's **absolutely necessary**. And in this case I'm not convinced that it's in fact absolutely necessary.

gbrodman (changes requested):
> Reviewed 3 of 7 files at r1, 2 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`console-webapp/src/app/app.component.ts` line 43 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeoOhgb3HPLNn4n_xnk:-NeoOhgb3HPLNn4n_xnl:bxs4bvr) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/app.component.ts#L43)):*
> ```XML
>       this.globalLoader.stopGlobalLoader(this);
>     });
>     this.globalLoader.startGlobalLoader(this);
> ```

1. It probably doesn't really matter too much, but shouldn't we start the global loader first before sending off the call to the user service? 
2. It seems odd to me that we have the one GlobalLoader blocking call here in the app instantiation and one call in the registrar service in the constructor. Maybe do both of them here? 
___
*[`console-webapp/src/app/home/widgets/resources-widget.component.html` line 6 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeiemEg-XrTpdNUAh7g:-NeoKthUAebvvfvZuc9y:b-44ykr4) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/resources-widget.component.html#L6)):*

It will slow things down, I don't see how it wouldn't as it's extra http request. Even sending 3 requests (/registrars, /userdata and say /miscdata) in parallel (what you're suggesting here) we have no guarantees that /miscdata will be processed first, in fact it might as well be received and processed last by the server depending on the network trafficking, that's what I've been trying to tell.

We're talking here about things that directly affect TTI of our web app - https://developer.chrome.com/en/docs/lighthouse/performance/interactive/ and I'd very much prefer to stay in the green zone.

My point is that given that we make user wait for the entire resource loading (html, css, js, etc) and necessary data (registrars, users) already, which we can't do anything anyway, we should consider not making another blocking (read TTI affecting) request unless it's **absolutely necessary**. And in this case I'm not convinced that it's in fact absolutely necessary.
</blockquote></details>

discussed offline and we should hopefully be OK here 
___
*[`console-webapp/src/app/home/widgets/settings-widget.component.ts` line 33 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeoNWZM-UNs7bQ46JRi:-NeoNWZM-UNs7bQ46JRj:bb55moo) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/settings-widget.component.ts#L33)):*
> ```XML
>   }
> 
>   openSecPage() {
> ```

i know this is nitty, but can we use the full term `Security`? It removes ambiguity and 
___
*[`console-webapp/src/app/shared/services/userData.service.ts` line 28 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeoPAOP8zmTmiheNABn:-NeoPAOP8zmTmiheNABo:bwrhgft) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/shared/services/userData.service.ts#L28)):*
> ```XML
>   providedIn: 'root',
> })
> export class UserDataService {
> ```

Shouldn't this implement GlobalLoader?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: 14 of 18 files reviewed, 3 unresolved discussions (waiting on @gbrodman)

___
*[`console-webapp/src/app/app.component.ts` line 43 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeoOhgb3HPLNn4n_xnk:-Nes47R86Q9kvz0KWgur:bebxjrf) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/app.component.ts#L43)):*

1. It probably doesn't really matter too much, but shouldn't we start the global loader first before sending off the call to the user service? 
2. It seems odd to me that we have the one GlobalLoader blocking call here in the app instantiation and one call in the registrar service in the constructor. Maybe do both of them here?
</blockquote></details>

1) It matters functionally - loader is usually started after the call to cover the potential issue with the call itself that might arise and leave the loader hanging. In other words if `getUserData()` throws runtime error for whatever reason we won't have to deal with the ghost webapp loading state
2) Good point, however this is not a good place for loaders, in fact let me remove `getUserData` loader from here and put it in the service,  where it belongs.
___
*[`console-webapp/src/app/home/widgets/settings-widget.component.ts` line 33 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeoNWZM-UNs7bQ46JRi:-NesAZ_f7_Q6mCY2zHnj:b-896fix) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/home/widgets/settings-widget.component.ts#L33)):*

i know this is nitty, but can we use the full term `Security`? It removes ambiguity and
</blockquote></details>

Done.
___
*[`console-webapp/src/app/shared/services/userData.service.ts` line 28 at r2](https://reviewable.io/reviews/google/nomulus/2151#-NeoPAOP8zmTmiheNABn:-NesA_4Z8RoB5tHYRE_-:b-7gvx66) ([raw file](https://github.com/google/nomulus/blob/16e80f6492252608034b19fe6d2f1a8c18464448/console-webapp/src/app/shared/services/userData.service.ts#L28)):*

Shouldn't this implement GlobalLoader?
</blockquote></details>

It does now as I moved globalLoader here call from `app.component.ts`

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2151)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ptkach)

PR #2158, opened 2023-09-27
Use equals() method to compare equality

jianglai:
It will call equalsImmutableObject(), which seems the right thing to do.
We only care if the two Tld objects have the same fields, not if they
are the same object. ErrorProne complained about comparison by identity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2158)


sarahcaseybot (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2158#-:-NfRpBZ26mJi7f4KHc3h:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2160, opened 2023-09-28
Deprecate the OAuth header in Nomulus tool

jianglai:
Unless an --oauth flag is used, the nomulus tool will only send the OIDC
header. The server still accepts both headers and the user should use
`create_user` command to create an admin User (with the --oauth flag on), which
will then allow one to use the nomulus tool without the --oauth flag.

The --oauth flag and the server's ability to support OAuth-based
authentication will be removed soon. Users are urged to create the User
object in time to avoid service interruption.

TESTED=verified on alpha.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2160)

### DIFF
--- a/core/src/main/java/google/registry/tools/RegistryCli.java
+++ b/core/src/main/java/google/registry/tools/RegistryCli.java
@@ -43,11 +43,19 @@ final class RegistryCli implements CommandRunner {
   // The environment parameter is parsed twice: once here, and once with {@link
   // RegistryToolEnvironment#parseFromArgs} in the {@link RegistryTool#main} function.
   //
-  // The flag names must be in sync between the two, and also - this is ugly and we should feel bad.
+  // The flag names must be in sync between the two, and also - this is ugly, and we should feel
+  // bad.
   @Parameter(
       names = {"-e", "--environment"},
       description = "Sets the default environment to run the command.")
   private RegistryToolEnvironment environment = RegistryToolEnvironment.PRODUCTION;
+
+  @Parameter(
+      names = "--oauth",
+      description =
+          "Turn on OAuth-based authentication, the usage of which is to be deprecated. Use"
+              + " `create_user` to create an Admin user that allows for OIDC-based authentication.")
+  private boolean oAuth = false;
 
   @Parameter(
       names = {"-c", "--commands"},
@@ -55,22 +63,21 @@ final class RegistryCli implements CommandRunner {
   private boolean showAllCommands;
 
   @Parameter(
-      names = {"--credential"},
+      names = "--credential",
       description =
           "Name of a JSON file containing credential information used by the tool. "
               + "If not set, credentials saved by running `nomulus login' will be used.")
   private String credentialJson = null;
 
   @Parameter(
-      names = {"--sql_access_info"},
+      names = "--sql_access_info",
       description =
           "Name of a file containing space-separated SQL access info used when deploying "
               + "Beam pipelines")
   private String sqlAccessInfoFile = null;
 
   // Do not make this final - compile-time constant inlining may interfere with JCommander.
-  @ParametersDelegate
-  private LoggingParameters loggingParams = new LoggingParameters();
+  @ParametersDelegate private LoggingParameters loggingParams = new LoggingParameters();
 
   RegistryToolComponent component;
 
@@ -105,8 +112,8 @@ final class RegistryCli implements CommandRunner {
     jcommander.setProgramName(programName);
 
     // Create all command instances. It would be preferrable to do this in the constructor, but
-    // JCommander mutates the command instances and doesn't reset them so we have to do it for every
-    // run.
+    // JCommander mutates the command instances and doesn't reset them, so we have to do it for
+    // every run.
     try {
       for (Map.Entry<String, ? extends Class<? extends Command>> entry : commands.entrySet()) {
         Command command = entry.getValue().getDeclaredConstructor().newInstance();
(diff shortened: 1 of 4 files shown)


jianglai (commented):
> Reviewed 4 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2160)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (changes requested):
is there a mechanism in place in the future for people to create User objects if they were starting with a fresh installation post-oauth flag? 

> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2160#-:-NfRM3NWCykY-lqWzBYz:b-pfwkq)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 62 at r2](https://reviewable.io/reviews/google/nomulus/2160#-NfRMLDJ7lXuiduiGhGg:-NfRMLDJ7lXuiduiGhGh:b-ihsmki) ([raw file](https://github.com/google/nomulus/blob/a5ab3a0e18c438dae5a34206e16dbd45db789ea3/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L62)):*
> ```Java
>                   credentialsBundle.getHttpRequestInitializer().initialize(request);
>                 }
>                 // Set OIDC token as the alternative bearer token.
> ```

probably want to update this comment since it's the primary, not the alternate?

jianglai (commented):
That's a very good question that I have not thought of. I guess with OAuth we just need to add the account to the group that has App Engine admin role, but here the User entity has no connections to IAM. The only way to bootstrap would be to create the User entity directly in SQL, I think. As long as you have one user with access to the tool, they can create users for other people.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2160#-:-NfRQFel14Vm165oNCCE:b-n8md1)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 62 at r2](https://reviewable.io/reviews/google/nomulus/2160#-NfRMLDJ7lXuiduiGhGg:-NfRS42C8wD3M0rOuqO-:b-2tvz79) ([raw file](https://github.com/google/nomulus/blob/a5ab3a0e18c438dae5a34206e16dbd45db789ea3/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L62)):*

probably want to update this comment since it's the primary, not the alternate?
</blockquote></details>

I was planning to change it to primary after we get rid of OAuth. Right now it still uses the `PROXY_AUTHORIZATION` header name, even if the OAuth header is absent. I can add more conditionals to make it use `AUTHORIZATION` when it is the only one present, but it doesn't seem to be worth the effort given that all the conditionals will go away soon, and the current setup proves to be working.

gbrodman (changes requested):
Maybe we should add a configuration option with a set of bootstrap email addresses? That would work; I don't know if it's the best option 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2160#-:-NfRxbgS9kea_em2Opb9:be7ofxk)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 62 at r2](https://reviewable.io/reviews/google/nomulus/2160#-NfRMLDJ7lXuiduiGhGg:-NfRxaCO-l7FqwCq003n:b-kshstr) ([raw file](https://github.com/google/nomulus/blob/a5ab3a0e18c438dae5a34206e16dbd45db789ea3/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L62)):*

I was planning to change it to primary after we get rid of OAuth. Right now it still uses the `PROXY_AUTHORIZATION` header name, even if the OAuth header is absent. I can add more conditionals to make it use `AUTHORIZATION` when it is the only one present, but it doesn't seem to be worth the effort given that all the conditionals will go away soon, and the current setup proves to be working.
</blockquote></details>

hmm fair enough

jianglai (commented):
I doubt it is the best option, for a couple of reasons:

1. Putting it in the config file would tie it to the deployment cycle. I think we'd want more flexibility than that.
2. Currently the config file supports whitelisting service accounts used for "APP" level authentication.  We'd need to make more changes to support configuring user level authentication with admin permissions, just for the purpose of bootstrapping.

Honestly, having to create a user in the database doesn't seem too big of a hassle, if you only have to do it once when setting things up from scratch. It'll just be part of the "installation guide", if we finally had time to update it :P

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2160#-:-NfS5ZbP0CgsBubodPzs:b-3b1mtv)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tools/RequestFactoryModule.java` line 62 at r2](https://reviewable.io/reviews/google/nomulus/2160#-NfRMLDJ7lXuiduiGhGg:-NfS5UeaACI72VkE3HaY:b-896fix) ([raw file](https://github.com/google/nomulus/blob/a5ab3a0e18c438dae5a34206e16dbd45db789ea3/core/src/main/java/google/registry/tools/RequestFactoryModule.java#L62)):*

hmm fair enough
</blockquote></details>

Done.

gbrodman (approved):
Hahaha. Yeah, we'll definitely need to update that.....

but fair enough 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2160#-:-NflwCzcDiWvXVnYJRwp:b-xgoje7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2161, opened 2023-09-28
Add custom YAML serializer for Duration

sarahcaseybot:
This addresses b/301119144. This changes the YAML representation of a TLD to show Duration fields as a String reperesntation using the Java Duration object's toString() format. This eliminates the previous ambiguity over the time unit that is being used for each duration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2161)

### DIFF
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -51,7 +51,6 @@ import google.registry.model.CreateAutoTimestamp;
 import google.registry.model.EntityYamlUtils.CreateAutoTimestampDeserializer;
 import google.registry.model.EntityYamlUtils.CurrencyDeserializer;
 import google.registry.model.EntityYamlUtils.CurrencySerializer;
-import google.registry.model.EntityYamlUtils.DurationSerializer;
 import google.registry.model.EntityYamlUtils.OptionalDurationSerializer;
 import google.registry.model.EntityYamlUtils.OptionalStringSerializer;
 import google.registry.model.EntityYamlUtils.SortedEnumSetSerializer;
@@ -417,42 +416,34 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
    * amount of time following creation.
    */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration addGracePeriodLength = DEFAULT_ADD_GRACE_PERIOD;
 
   /** The length of the anchor tenant add grace period for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration anchorTenantAddGracePeriodLength = DEFAULT_ANCHOR_TENANT_ADD_GRACE_PERIOD;
 
   /** The length of the autorenew grace period for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration autoRenewGracePeriodLength = DEFAULT_AUTO_RENEW_GRACE_PERIOD;
 
   /** The length of the redemption grace period for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration redemptionGracePeriodLength = DEFAULT_REDEMPTION_GRACE_PERIOD;
 
   /** The length of the renew grace period for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration renewGracePeriodLength = DEFAULT_RENEW_GRACE_PERIOD;
 
   /** The length of the transfer grace period for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration transferGracePeriodLength = DEFAULT_TRANSFER_GRACE_PERIOD;
 
   /** The length of time before a transfer is automatically approved for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration automaticTransferLength = DEFAULT_AUTOMATIC_TRANSFER_LENGTH;
 
   /** The length of time a domain spends in the non-redeemable pending delete phase for this TLD. */
   @Column(nullable = false)
-  @JsonSerialize(using = DurationSerializer.class)
   Duration pendingDeleteLength = DEFAULT_PENDING_DELETE_LENGTH;
 
   /** The currency unit for all costs associated with this TLD. */
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -51,6 +51,7 @@ import google.registry.model.CreateAutoTimestamp;
 import google.registry.model.EntityYamlUtils.CreateAutoTimestampDeserializer;
 import google.registry.model.EntityYamlUtils.CurrencyDeserializer;
 import google.registry.model.EntityYamlUtils.CurrencySerializer;
+import google.registry.model.EntityYamlUtils.DurationSerializer;
 import google.registry.model.EntityYamlUtils.OptionalDurationSerializer;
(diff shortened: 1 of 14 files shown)


core/src/test/java/google/registry/tools/GetTldCommandTest.java
```diff
@@ -45,7 +45,7 @@ void testSuccess() throws Exception {
     PremiumList premiumList = persistPremiumList("test", USD, "silver,USD 50", "gold,USD 80");
     persistResource(
         tld.asBuilder()
-            .setDnsAPlusAaaaTtl(Duration.millis(900))
```
CydeWeys, line 48: 😬

Well this definitely fixes a pre-existing bug then, huh, where 900 in the YAML looked like it should've been 15 minutes (which would make sense) but it actually represented 900ms.

core/src/test/java/google/registry/tools/GetTldCommandTest.java
```diff
@@ -45,7 +45,7 @@ void testSuccess() throws Exception {
...
+            .setDnsAPlusAaaaTtl(Duration.standardSeconds(900))
```
CydeWeys, line 48: It's tiny, but I think `Duration.standardMinutes(15)` would be clearer here.

core/src/main/java/google/registry/model/tld/Tld.java
```diff
@@ -416,34 +417,42 @@ public ImmutableSet<String> getReservedListNames() {
    * amount of time following creation.
    */
   @Column(nullable = false)
+  @JsonSerialize(using = DurationSerializer.class)
```
CydeWeys, line 420: And there's no way to apply this by default to all `Duration`s?  Having to include it individually for each field seems like it might be prone to forgetting to do it.  There's no `Duration` we wouldn't want to serialize in this way, right?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2161)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 420 at r1](https://reviewable.io/reviews/google/nomulus/2161#gh-1340697578:-NfSp4sU1LnMOSGMUvHg:bubqdkw) ([raw file](https://github.com/google/nomulus/blob/4980c9cbbdf2d5fe0c152d41d6bfb0731b1d95c0/core/src/main/java/google/registry/model/tld/Tld.java#L420)):*

And there's no way to apply this by default to all `Duration`s?  Having to include it individually for each field seems like it might be prone to forgetting to do it.  There's no `Duration` we wouldn't want to serialize in this way, right?
</blockquote></details>

Agreed that we probably would want to apply this to all Durations. I tried applying this to the entire object mapper in `EntityYamlUtils.createObjectMapper` similarly to how it was done for Money objects, but this did not work. I'm assuming this is because it defaults to the built in Duration serializer. 
___
*[`core/src/test/java/google/registry/tools/GetTldCommandTest.java` line 0 at r1](https://reviewable.io/reviews/google/nomulus/2161#gh-1340696920:-NfSrCyEDJLNcIGWAxXB:b-gb1z7d) ([raw file](https://github.com/google/nomulus/blob/4980c9cbbdf2d5fe0c152d41d6bfb0731b1d95c0/core/src/test/java/google/registry/tools/GetTldCommandTest.java#L0)):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

😬

Well this definitely fixes a pre-existing bug then, huh, where 900 in the YAML looked like it should've been 15 minutes (which would make sense) but it actually represented 900ms.
</blockquote></details>

eh, not exactly a bug since I don't think this broke anything, but yes it was an odd value to use in a test and unclear with just a number what the time unit was
___
*[`core/src/test/java/google/registry/tools/GetTldCommandTest.java` line 48 at r1](https://reviewable.io/reviews/google/nomulus/2161#gh-1340697173:-NfSs3qO-IJBQtYzREGC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4980c9cbbdf2d5fe0c152d41d6bfb0731b1d95c0/core/src/test/java/google/registry/tools/GetTldCommandTest.java#L48)):*

It's tiny, but I think `Duration.standardMinutes(15)` would be clearer here.
</blockquote></details>

Done.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2161)* status: 0 of 14 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/EntityYamlUtils.java` line 67 at r2](https://reviewable.io/reviews/google/nomulus/2161#-NfX80ZiAuboIvj--yY2:-NfX80ZiAuboIvj--yY3:b-mpi3ay) ([raw file](https://github.com/google/nomulus/blob/9aeeb29616aed32a893f896ea2b7fd04f3f4d427/core/src/main/java/google/registry/model/EntityYamlUtils.java#L67)):*
> ```Java
>             .build()
>             .registerModule(module);
>     mapper.findAndRegisterModules();
> ```

This should be like below. Otherwise your customization will be overridden by defaults.. 

        .....    
        .build();
    mapper.findAndRegisterModules()
            .registerModule(module);



_Code quote:_
```Java
            .build()
            .registerModule(module);
    mapper.findAndRegisterModules();
```
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 420 at r1](https://reviewable.io/reviews/google/nomulus/2161#gh-1340697578:-NfX8XUF0lqhGRxyIRyJ:b-cvqcrq) ([raw file](https://github.com/google/nomulus/blob/4980c9cbbdf2d5fe0c152d41d6bfb0731b1d95c0/core/src/main/java/google/registry/model/tld/Tld.java#L420)):*

Agreed that we probably would want to apply this to all Durations. I tried applying this to the entire object mapper in `EntityYamlUtils.createObjectMapper` similarly to how it was done for Money objects, but this did not work. I'm assuming this is because it defaults to the built in Duration serializer.
</blockquote></details>

See my comments above in EntityYamlUtility.java.
You should be able to override the default if you move `registryModule()` behind `findAndRegistryModules()`.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2161)* status: 0 of 14 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/model/EntityYamlUtils.java` line 67 at r2](https://reviewable.io/reviews/google/nomulus/2161#-NfX80ZiAuboIvj--yY2:-NfXOzxZ4_PRMSU3Our7:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9aeeb29616aed32a893f896ea2b7fd04f3f4d427/core/src/main/java/google/registry/model/EntityYamlUtils.java#L67)):*

This should be like below. Otherwise your customization will be overridden by defaults.. 

        .....    
        .build();
    mapper.findAndRegisterModules()
            .registerModule(module);
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 420 at r1](https://reviewable.io/reviews/google/nomulus/2161#gh-1340697578:-NfXP02T7D1egaR66XOZ:b-ygmfmu) ([raw file](https://github.com/google/nomulus/blob/4980c9cbbdf2d5fe0c152d41d6bfb0731b1d95c0/core/src/main/java/google/registry/model/tld/Tld.java#L420)):*

See my comments above in EntityYamlUtility.java.
You should be able to override the default if you move `registryModule()` behind `findAndRegistryModules()`.
</blockquote></details>

Thanks for pointing that out

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2161)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

PR #2171, opened 2023-10-04
Remove OAuthAuthenticationMechanism

jianglai:
Also made some refactoring to various Auth related classes to clean up things a bit and make the logic less convoluted:

1. In `Auth`, remove `AUTH_API_PUBLIC` as it is only used by the WHOIS and EPP endpoints accessed by the proxy. Previously, the proxy relies on OAuth and its service account is not given admin role (in OAuth parlance), so we made them accessible by a public user, deferring authorization to the actions themselves. In practice, OAuth checks for allowlisted client IDs and only the proxy client ID was allowlisted, which effectively limited access to only the proxy anyway.

2. In `AuthResult`, expose the service account email if it is at `APP` level. `RequestAuthenticator` will print out the auth result and therefore log the email, making it easy to identify which account was used. This field is mutually exclusive to the user auth info field. As a result, the factory methods are refactored to explicitly create either `APP` or `USER` level auth result.

3. Completely re-wrote `RequestAuthenticatorTest`. Previously, the test mingled testing functionalities of the target class with testing how various authentication mechanisms work. Now they are cleanly decoupled, and each method in `RequestAuthenticator` is tested individually.

4. Removed nomulus-config-production-sample.yaml as it is vastly out of date.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2171)


jianglai (commented):
> Reviewed 45 of 45 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2171)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (changes requested):
I think we refer to allowedOauthClientIds in some of the docs too. Let's fix those references / docs here too; it shouldn't be too significant hopefully 


Note for 1 that I believe in your last sentence you mean "allowlisted service account IDs", not "client IDs" -- the proxy-service-account identifiers being allowlisted are what allows the proxy to work with OAuth. 

> Reviewed 42 of 45 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2171#-:-Ng-oxC8-U2OxlOYdvfT:b7dr8vd)* status: all files reviewed, 6 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/request/auth/Auth.java` line 56 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng04-xs2Gs0d4M2qbJF:-Ng04-xs2Gs0d4M2qbJG:b-4mmcl6) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/main/java/google/registry/request/auth/Auth.java#L56)):*
> ```Java
> 
>   /**
>    * Allows only the app or admins to access.
> ```

"the app" is unclear -- what do we mean here? I'm assuming we meant the service accounts that are listed? We can just say that instead (which we do below) 
___
*[`core/src/main/java/google/registry/request/auth/AuthSettings.java` line 75 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng05xan7WyoKy46PwLy:-Ng05xan7WyoKy46PwLz:b-ncd95q) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/main/java/google/registry/request/auth/AuthSettings.java#L75)):*
> ```Java
>      * Authentication required, but user not required.
>      *
>      * <p>In Auth: Authentication is required, but App-internal authentication (which isn't
> ```

nit, but i don't think we need to capitalize A 
___
*[`core/src/main/java/google/registry/request/auth/RequestAuthenticator.java` line 81 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng07iaD-cClwpem0rTJ:-Ng07iaD-cClwpem0rTK:b-gqnik5) ([raw file](https://github.com/google/nomulus/blob/2da8ea0185e3449d3c069376c1a7c5d1ca64e325/core/src/main/java/google/registry/request/auth/RequestAuthenticator.java#L81)):*
> ```Java
>    *
>    * @param auth the auth settings of the action, which determine what authentication is allowed
>    * @param req the {@link HttpServletRequest}; some authentication mechanisms use HTTP headers
> ```

why remove this comment? It's still definitely true since we use the Authorization header 
___
*[`core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java` line 91 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng0F8kH9X4oCNNDcM18:-Ng0F8kH9X4oCNNDcM19:b-d170mc) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java#L91)):*
> ```Java
>                 ImmutableList.of(apiAuthenticationMechanism1, apiAuthenticationMechanism2),
>                 legacyAuthenticationMechanism)
>             .authorize(
> ```

why are we setting a class-level variable here? If we can, we should just return the value and test it directly 
___
*[`core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java` line 106 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng0E3M84lI032UT1yvb:-Ng0E3M84lI032UT1yvc:b9kfbw1) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java#L106)):*
> ```Java
>       assertThat(authResult).isEmpty();
>     } else {
>       assertThat(authResult).isPresent();
> ```

small thing, but fun: we can use `Truth8.assertThat(someOptional).hasValue(...)`


or we can just do a single line assertThat(authResult).isEqualTo(Optional.ofNullable(foundResult));
___
*[`core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java` line 113 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng0EFiY1a74z2cdq5Sq:-Ng0EFiY1a74z2cdq5Sr:bdw9a4v) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java#L113)):*
> ```Java
>   @BeforeEach
>   void beforeEach() {
>     when(apiAuthenticationMechanism1.authenticate(eq(req))).thenReturn(NOT_AUTHENTICATED);
> ```

I don't think we need the `eq()` calls here? From what I recall those are only necessary when you're mixing `any()` and concrete types in the same call 

(same comment for all other locations of `eq` i think)

jianglai (commented):
I changed the reference to `allowedOauthClientIds` in the proxy config doc, as it is still relatively up to date. The other mentions of the field are in docs that are terribly out of date. There's no point only updating the field name when an entire section needs to be re-written. It might be worthwhile revisiting these docs once the changes to auth are complete.

For (1), I did mean allowed client IDs. What I was trying to say that even though we had `AUTH_API_PUBLIC` for the WHOIS and EPP endpoints, they were not really "public" as you still need the client ID to be whitelisted, so changing them to `AUTH_API` doesn't really change anything, except that OIDC doesn't care about client IDs and the service accounts are indeed whitelisted and checked, compared to before when any service account could in theory connect, but they wouldn't have access to a whiltelisted client ID to send their requests with.

> Reviewed 7 of 7 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2171#-:-NgTvj0J-S7qUSmARpTz:bln4txh)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/request/auth/Auth.java` line 56 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng04-xs2Gs0d4M2qbJF:-NgTz-Bz3hTkGLy9E0qV:b-896fix) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/main/java/google/registry/request/auth/Auth.java#L56)):*

"the app" is unclear -- what do we mean here? I'm assuming we meant the service accounts that are listed? We can just say that instead (which we do below)
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/request/auth/RequestAuthenticator.java` line 81 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng07iaD-cClwpem0rTJ:-NgTzPC_81wqYA3kTNX6:bv9gw21) ([raw file](https://github.com/google/nomulus/blob/2da8ea0185e3449d3c069376c1a7c5d1ca64e325/core/src/main/java/google/registry/request/auth/RequestAuthenticator.java#L81)):*

why remove this comment? It's still definitely true since we use the Authorization header
</blockquote></details>

I removed it because it implied that some mechanisms DON'T use headers (such as internal auth, which we are getting rid of next). But you are right, it's better to keep it there.
___
*[`core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java` line 91 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng0F8kH9X4oCNNDcM18:-NgU-7MjC438CvF0j_Mi:bmndxp2) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java#L91)):*

why are we setting a class-level variable here? If we can, we should just return the value and test it directly
</blockquote></details>

It makes it easier to call `assertAuthResult` later, otherwise you'd have to pass the expected results and the actual result. But using Truth8 makes it moot.
___
*[`core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java` line 106 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng0E3M84lI032UT1yvb:-NgU0RDQBJbG9Djx0oMQ:bsfmw37) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java#L106)):*

small thing, but fun: we can use `Truth8.assertThat(someOptional).hasValue(...)`


or we can just do a single line assertThat(authResult).isEqualTo(Optional.ofNullable(foundResult));
</blockquote></details>

It looks like the Truth8 variant doesn't deal with empty and you have to call `isEmpty` directly.
___
*[`core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java` line 113 at r3](https://reviewable.io/reviews/google/nomulus/2171#-Ng0EFiY1a74z2cdq5Sq:-NgU1qpOACzBPn6R6PLy:b-5atzwx) ([raw file](https://github.com/google/nomulus/blob/45556aef2698dad5f02370a3da8db9b800f26a30/core/src/test/java/google/registry/request/auth/RequestAuthenticatorTest.java#L113)):*

I don't think we need the `eq()` calls here? From what I recall those are only necessary when you're mixing `any()` and concrete types in the same call 

(same comment for all other locations of `eq` i think)
</blockquote></details>

I think you are right, if you pass the argument directly, it is as if they are wrapped in `eq()`.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2171)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman)

gbrodman (approved):
re: 1 I think you might be mixing up "service account email" and "client ID" in some places? The OIDC auth mechanism definitely cares about what client ID we're using, since it has to be the same when creating the token (e.g. in the proxy) and when we verify the token. 

sounds good re: the super-out-of-date documentation in the non-proxy docs, we should probably bring this up at some point in a planning meeting 

> Reviewed 6 of 7 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2171#-:-NgVawpC9wcZtATAP2Na:belxugl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (commented):
The issue with client ID is indeed confusing. It comes from the fact that they are used differently between OAuth and OIDC.

In OAuth, the client ID is determined by the "client" that contacts the identity server to request the access token. Depending on the client type and the authentication flow type, it could be tied to the service account in case of the proxy, or the nomulus tool's client secret. The OAuth authentication flow then verifies that the request bears an access token that was generated for that client ID, by checking the (deleted in this PR) `auth.allowedOauthClientIds` field.

In OIDC, on the other hand, the client ID is something that you supply to the identity server, to be put in the audience field of the returned ID token. The verifier then checks if the audience is what is expected, by checking the `auth.oauthClientId` field.

The key difference is that for OAuth token, you say "this request is from client X representing principal Y", where both X and Y are none-impersonate-able -- if you don't have the nomulus tool's client secret, you cannot claim to be the nomulus tool. The server in turn cares about both parts to be allowed. That's what I meant when I say "in practice, OAuth checks for allowlisted client IDs and only the proxy client ID was allowlisted, which effectively limited access to only the proxy anyway." Even though the configured auth for EPP and WHOIS allows any principal Y to connect, the request can only come from whitelisted client IDs, effectively limiting who can.

For OIDC tokens, on the other hand, you say "this request is from principal Y, intended for audience Z". Here Z can be anything (as long as it is legitimate, as we have found out that Cloud Scheduler only allows valid Client IDs to be set as the audience). You do NOT have to have some secret to prove that you have control over Z. The verifier cares about the token being intended for itself, not for someone else, but there's nothing preventing you from asking the OIDC identity server to mint you an ID for any legitimate audience. The fact that we use the OAuth client ID created for IAP as the audience is because 1) as stated before, some service validates the format of the audience and requires it to be an client ID, and 2) it is what IAP looks for when it inspects an incoming request. By setting the audience to this particular client ID, the same request can pass both IAP and non-IAP protected endpoints, as long as the regular OIDC authentication mechanism also expects that client ID as the audience.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2171#-:-NgVu_7iCxqqYFZmp6se:b-30q50a)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2177, opened 2023-10-12
Refine error handling in RequestHandler and the console slightly

gbrodman:
If we don't explicitly handle random unexpected exceptions, the error that the front end receives includes a big ole stacktrace, which is unhelpful for regular users and possibly bad to expose. Instead, we should provide a vague "something went wrong" message.

Separately, we can create a default SnackBar options and use that (we want it longer than 1.5 seconds because that's pretty short).

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2177)


ptkach (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2177)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #2179, opened 2023-10-12
Use reTransact when loading the cache for database objects

sarahcaseybot:
Cache loads will likely always be inner transactions, if they have a transaction at all. Cache loads do not always call a transaction since they are only necessary if the cache is not fresh at the time it is called. Since the cache itself needs to decide whether or not a DB transaction is necessary, it should use the reTransact method to safely indicate that the isolation level of the outer transaction is what should be used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2179)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
+++ b/core/src/main/java/google/registry/model/domain/token/AllocationToken.java
@@ -305,14 +305,14 @@ public class AllocationToken extends UpdateAutoTimestampEntity implements Builda
                   new CacheLoader<VKey<AllocationToken>, Optional<AllocationToken>>() {
                     @Override
                     public Optional<AllocationToken> load(VKey<AllocationToken> key) {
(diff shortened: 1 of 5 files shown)


CydeWeys (commented):
Commit message should also explain that the caches themselves need to handle whether a transaction is necessary or not, as if the cache is fresh no transaction should be started but if the cache needs to be loaded then it has to do a DB transaction.  Hence why these inner "might be" transaction calls are necessary.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2179#-:-NgeRPoE7D7u5fpWhya3:b-wmx2vl)* status: 0 of 5 files reviewed, all discussions resolved

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #2181, opened 2023-10-16
Remove URLFetch

jianglai:
We previously needed to use URLFetch in some instances where TLS 1.3 is
required (mostly when connecting to ICANN servers),and the JDK-bundled SSL
engine that came with App Engine runtime did not support TLS 1.3.

It appears now that the Java 8 runtime on App Engine supports TLS 1.3
out of the box, which allows us to get rid of URLFetch, which depends on
App Engine APIs.

Also removed some redundant retry and logging logic, now that we know
the HTTP client behaves correctly.

TESTED=modified the CannedScriptExecutionAction and deployed to alpha, used the
new HTTP client to connect to the three URL endpoints that were
problematic before and confirmed that TLS connections can be established. HTTP
sessions were rejected in some cases when authentication failed, but
that was expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2181)

### DIFF
--- a/core/src/main/java/google/registry/batch/BatchModule.java
+++ b/core/src/main/java/google/registry/batch/BatchModule.java
@@ -44,6 +44,12 @@ public class BatchModule {
   public static final String PARAM_FAST = "fast";
 
   @Provides
+  @Parameter("url")
+  static String provideUrl(HttpServletRequest req) {
+    return extractRequiredParameter(req, "url");
+  }
+
+  @Provides
   @Parameter("jobName")
   static Optional<String> provideJobName(HttpServletRequest req) {
     return extractOptionalParameter(req, "jobName");
--- a/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java
+++ b/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java
@@ -14,26 +14,20 @@
 
 package google.registry.batch;
 
-import static com.google.common.collect.ImmutableList.toImmutableList;
+import static google.registry.request.Action.Method.GET;
 import static google.registry.request.Action.Method.POST;
-import static google.registry.util.RegistrarUtils.normalizeRegistrarId;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Streams;
 import com.google.common.flogger.FluentLogger;
-import google.registry.config.RegistryConfig.Config;
-import google.registry.groups.GmailClient;
-import google.registry.groups.GroupsConnection;
-import google.registry.model.registrar.Registrar;
-import google.registry.model.registrar.RegistrarPoc;
 import google.registry.request.Action;
+import google.registry.request.Parameter;
+import google.registry.request.Response;
+import google.registry.request.UrlConnectionService;
+import google.registry.request.UrlConnectionUtils;
 import google.registry.request.auth.Auth;
-import google.registry.util.EmailMessage;
-import java.io.IOException;
-import java.util.Set;
+import java.net.URL;
 import javax.inject.Inject;
-import javax.mail.internet.AddressException;
-import javax.mail.internet.InternetAddress;
+import javax.net.ssl.HttpsURLConnection;
 
 /**
  * Action that executes a canned script specified by the caller.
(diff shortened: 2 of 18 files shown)


jianglai (commented):
> Reviewed 11 of 15 files at r1, 5 of 5 files at r2, 2 of 2 files at r3, 2 of 2 files at r5, 2 of 2 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2181)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java` line 75 at r5](https://reviewable.io/reviews/google/nomulus/2181#gh-1362755103:-Nh2Hw4b0ocCqE2Cjx9f:blz4gsp) ([raw file](https://github.com/google/nomulus/blob/de08d5d9b51f79941afc45a78da290f3319c6906/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java#L75)):*

## Log Injection

This log entry depends on a [user-provided value](1).

[Show more details](https://github.com/google/nomulus/security/code-scanning/1574)
</blockquote></details>

Only used for testing.
___
*[`core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java` line 109 at r1](https://reviewable.io/reviews/google/nomulus/2181#gh-1360952505:-Ngz5OxRBIem3IobgECX:bss4a2b) ([raw file](https://github.com/google/nomulus/blob/3f8bbf249a67cbc2570e303ba9d97dd78b60d6fb/core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java#L109)):*

## Log Injection

This log entry depends on a [user-provided value](1).

[Show more details](https://github.com/google/nomulus/security/code-scanning/1573)
</blockquote></details>

Removed.

gbrodman (changes requested):
> Reviewed 8 of 15 files at r1, 2 of 5 files at r2, 2 of 2 files at r3, 1 of 2 files at r5, 1 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2181)* status: all files reviewed, 7 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java` line 47 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7PQ_2-Wb74f1RcMr0:-Nh7PQ_2-Wb74f1RcMr1:be2t45e) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java#L47)):*
> ```Java
>     service = Action.Service.BACKEND,
>     path = "/_dr/task/executeCannedScript",
>     method = {POST, GET},
> ```

We shouldn't need both POST and GET, right? 
___
*[`core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java` line 61 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7OvW45FiYX6xAeXLv:-Nh7OvW45FiYX6xAeXLw:blnvm54) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java#L61)):*
> ```Java
> 
>   @Inject
>   CannedScriptExecutionAction() {}
> ```

i don't think we need the constructor if the class variables are injected
___
*[`core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java` line 120 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7Rta8Evu5gRe_TV2m:-Nh7Rta8Evu5gRe_TV2n:b-l13u4) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsAction.java#L120)):*
> ```Java
>       csvString = new String(UrlConnectionUtils.getResponseBytes(connection), UTF_8);
>     } finally {
>       connection.disconnect();
> ```

I don't think we do the try-finally-disconnect pattern in the other locations that we use HttpURLConnection. Would it be difficult to add them in this PR? 
___
*[`core/src/main/java/google/registry/rde/RdeReporter.java` line 100 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7SUvLCD_pul0gQz-h:-Nh7SUvLCD_pul0gQz-i:b1gxwhz) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/rde/RdeReporter.java#L100)):*
> ```Java
>     // We know that an HTTP 200 response can only contain a result code of
>     // 1000 (i. e. success), there is no need to parse it.
>     if (responseCode != STATUS_CODE_OK) {
> ```

The only way we get this is BAD_REQUEST right? Can we make that more explicit? Also, do you know why that might occur? 
___
*[`core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java` line 110 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7TGmP-ukq6Ljo28Pi:-Nh7TGmP-ukq6Ljo28Pj:bz35uuc) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/reporting/icann/IcannHttpReporter.java#L110)):*
> ```Java
>     // 1000 (i. e. success), there is no need to parse it.
>     // See: https://tools.ietf.org/html/draft-lozano-icann-registry-interfaces-13#page-16
>     if (responseCode != STATUS_CODE_OK) {
> ```

same note, maybe we should make it explicit that this is BAD_REQUEST?

gbrodman (approved):
> Reviewed 1 of 5 files at r2, 3 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2181)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java` line 47 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7PQ_2-Wb74f1RcMr0:-Nh7qPMJEzlzFomTCeSw:b-f3smhe) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java#L47)):*

It's easier to pass the URL parameters with GET when they are part of the URL. There might be cases for POST if you want to provide something come complex like a JSON object in the body.

This class is used for testing things that can only be done on GAE. I don't think we can be a bit more flexible here.
</blockquote></details>

My point was that we could probably only have GET (and not have POST at all), because we don't use the request body. But it doesn't matter either way since this is kinda a scrap class anyway 
___
*[`core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java` line 61 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7OvW45FiYX6xAeXLv:-Nh7hY0VDzhdSTahkgUe:b2b6hkn) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/batch/CannedScriptExecutionAction.java#L61)):*

You still need a @Inject-annotated constructor for Dagger to create the instance for you.
</blockquote></details>

huh, I guess I never noticed
___
*[`core/src/main/java/google/registry/rde/RdeReporter.java` line 100 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7SUvLCD_pul0gQz-h:-Nh7qi9xAcIX7TYBINJN:b4no186) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/main/java/google/registry/rde/RdeReporter.java#L100)):*

My understanding is that if you get 400, it means the server understands what you sent but rejects it, so the reply would contain a reason. The reasons can be like the file you send over is malformed, for example. So you'd want to parse the response and see what the server says.

On the other hand, if you get something like a 401, you didn't present the right credential to the server and there would be nothing in the response body.

I actually think that checking for non-200 status is more robust here, as if in the future there arises a case where we need to parse the body for another non-200 response, we just need to change the if block above to not return early for that code either, and don't have to modify the code here.
</blockquote></details>

yeah, i think i see what you mean, cool 
___
*[`core/src/test/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsActionTest.java` line 91 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7pCi_2EH-YQaKaCG-:-Nh7pCia7JAARCKywsoI:bm622g6) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/test/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsActionTest.java#L91)):*
> ```Java
>         .containsExactly(
>             new URL("https://www.iana.org/assignments/registrar-ids/registrar-ids-1.csv"));
>     verify(connection).setRequestProperty(eq("Accept-Encoding"), eq("gzip"));
> ```

this doesn't really matter much but you don't need to use `eq` (that's only necessary when mixing `any()` and known values)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2181)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

___
*[`core/src/test/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsActionTest.java` line 91 at r7](https://reviewable.io/reviews/google/nomulus/2181#-Nh7pCi_2EH-YQaKaCG-:-Nh86XaKE9AyWmZbLH_C:b-ioe1i9) ([raw file](https://github.com/google/nomulus/blob/c89b0ec19a38b1df43fa7367b76c98e7f0096f0f/core/src/test/java/google/registry/rdap/UpdateRegistrarRdapBaseUrlsActionTest.java#L91)):*

this doesn't really matter much but you don't need to use `eq` (that's only necessary when mixing `any()` and known values)
</blockquote></details>

Good catch. I guess I'll have another PR to remove any use of `eq` there might be in the code base.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2181)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

PR #2182, opened 2023-10-17
Lower the isolation level for RefreshDnsForAllDomainsAction

sarahcaseybot:
This lowers the isolation level to TRANSACTION_REPEATABLE_READ which will hopefully allow the action to run the entire action without timing out on our larger TLDs.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2182)

### DIFF
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -193,7 +193,7 @@ hibernate:
   # to true, nested transactions will throw an exception. If set to false, a
   # transaction with the isolation override specified will still execute at the
   # default level (specified below).
-  perTransactionIsolation: true
+  perTransactionIsolation: false
 
   # Make 'SERIALIZABLE' the default isolation level to ensure correctness.
   #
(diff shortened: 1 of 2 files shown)


jianglai (approved):
Did you verify that the specified transaction isolation level is actually used?

Also, I would caution against changing the default config, as it would apply the setting to all projects, including production. Even though you are only setting the isolation level in this action, there might be potential issues with allowing it, if somewhere else in the code base we set an incompatible isolation level, which would throw an error instead of a warning.

The chances of this happening is pretty low (because I don't think we explicitly set the level anywhere else, currently) , but it's still non-zero. If there's no rush, I would feel more comfortable enabling it on non-prod environment first and then prod with a week-long delay. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2182#-:-NgyhX7k6eDtdSXoMe-D:be6ina6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

sarahcaseybot (commented):
I did verify via logs that the correct isolation level was used when I ran this in alpha. Since we don't explicitly set the isolation level anywhere other than tests, I don't think it is necessary to hold off on changing the default value, but to be safe I removed the change to default. Config change per-environment is added in: https://team-review.git.corp.google.com/c/domain-registry-eng/nomulus-internal/+/2059127 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2182#-:-NgyyIcQ-WxXi0uUffQ5:bl4ptoe)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @jianglai)

PR #2191, opened 2023-10-24
Switch to a stronger algorithm for password hashing

jianglai:
We have been using SHA256 to hash passwords (for both EPP and registry lock),
which is now considered too weak.

This PR switches to using Scrypt, a memory-hard slow hash function, with
recommended parameters per go/crypto-password-hash.

To ease the transition, when a password is being verified, both Scrypt
and SHA256 are tried. If SHA256 verification is successful, we re-hash
the verified password with Scrypt and replace the stored SHA256 hash
with the new one. This way, as long as a user uses the password once
before the transition period ends (when Scrypt becomes the only valid
algorithm), there would be no need for manual intervention from them.

We will send out notifications to users to remind them of the transition
and urge them to use the password (which should not be a problem with
EPP, but less so with the registry lock). After the transition,
out-of-band reset for EPP password, or remove-and-add on the console for
registry lock password, would be required for the hashes that have not
been re-saved.

Note that the re-save logic is not present for console user's registry
lock password, as there is no production data for console users yet.
Only legacy GAE user's password requires re-save.

TESTED=deployed to crash and verified that passwords are correctly rehashed to Scrypt for both EPP logins and registry lock requests from the legacy console.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2191)

### DIFF
--- a/core/src/main/java/google/registry/flows/session/LoginFlow.java
+++ b/core/src/main/java/google/registry/flows/session/LoginFlow.java
@@ -28,11 +28,17 @@ import google.registry.flows.EppException.CommandUseErrorException;
 import google.registry.flows.EppException.ParameterValuePolicyErrorException;
 import google.registry.flows.EppException.UnimplementedExtensionException;
 import google.registry.flows.EppException.UnimplementedObjectServiceException;
+import google.registry.flows.EppException.UnimplementedProtocolVersionException;
 import google.registry.flows.ExtensionManager;
 import google.registry.flows.FlowModule.RegistrarId;
+import google.registry.flows.FlowUtils.GenericXmlSyntaxErrorException;
 import google.registry.flows.MutatingFlow;
 import google.registry.flows.SessionMetadata;
+import google.registry.flows.TlsCredentials.BadRegistrarCertificateException;
+import google.registry.flows.TlsCredentials.BadRegistrarIpAddressException;
+import google.registry.flows.TlsCredentials.MissingRegistrarCertificateException;
 import google.registry.flows.TransportCredentials;
+import google.registry.flows.TransportCredentials.BadRegistrarPasswordException;
 import google.registry.model.eppcommon.ProtocolDefinition;
 import google.registry.model.eppcommon.ProtocolDefinition.ServiceExtension;
 import google.registry.model.eppinput.EppInput;
@@ -41,6 +47,7 @@ import google.registry.model.eppinput.EppInput.Options;
 import google.registry.model.eppinput.EppInput.Services;
 import google.registry.model.eppoutput.EppResponse;
 import google.registry.model.registrar.Registrar;
+import google.registry.util.PasswordUtils.HashAlgorithm;
 import java.util.Optional;
 import java.util.Set;
 import javax.inject.Inject;
@@ -48,14 +55,14 @@ import javax.inject.Inject;
 /**
  * An EPP flow for login.
  *
- * @error {@link google.registry.flows.EppException.UnimplementedExtensionException}
- * @error {@link google.registry.flows.EppException.UnimplementedObjectServiceException}
- * @error {@link google.registry.flows.EppException.UnimplementedProtocolVersionException}
- * @error {@link google.registry.flows.FlowUtils.GenericXmlSyntaxErrorException}
- * @error {@link google.registry.flows.TlsCredentials.BadRegistrarCertificateException}
- * @error {@link google.registry.flows.TlsCredentials.BadRegistrarIpAddressException}
- * @error {@link google.registry.flows.TlsCredentials.MissingRegistrarCertificateException}
- * @error {@link google.registry.flows.TransportCredentials.BadRegistrarPasswordException}
+ * @error {@link UnimplementedExtensionException}
+ * @error {@link UnimplementedObjectServiceException}
+ * @error {@link UnimplementedProtocolVersionException}
+ * @error {@link GenericXmlSyntaxErrorException}
+ * @error {@link BadRegistrarCertificateException}
+ * @error {@link BadRegistrarIpAddressException}
+ * @error {@link MissingRegistrarCertificateException}
+ * @error {@link BadRegistrarPasswordException}
  * @error {@link LoginFlow.AlreadyLoggedInException}
  * @error {@link BadRegistrarIdException}
  * @error {@link LoginFlow.TooManyFailedLoginsException}
@@ -134,13 +141,24 @@ public class LoginFlow implements MutatingFlow {
     if (!registrar.get().isLive()) {
       throw new RegistrarAccountNotActiveException();
     }
-    if (login.getNewPassword().isPresent()) {
+
+    if (login.getNewPassword().isPresent()
+        || registrar.get().getCurrentHashAlgorithm(login.getPassword()).orElse(null)
+            != HashAlgorithm.SCRYPT) {
+      String newPassword =
+          login
+              .getNewPassword()
+              .orElseGet(
+                  () -> {
+                    logger.atInfo().log("Rehashing existing registrar password with Scrypt");
+                    return login.getPassword();
+                  });
       // Load fresh from database (bypassing the cache) to ensure we don't save stale data.
       Optional<Registrar> freshRegistrar = Registrar.loadByRegistrarId(login.getClientId());
       if (!freshRegistrar.isPresent()) {
         throw new BadRegistrarIdException(login.getClientId());
       }
-      tm().put(freshRegistrar.get().asBuilder().setPassword(login.getNewPassword().get()).build());
+      tm().put(freshRegistrar.get().asBuilder().setPassword(newPassword).build());
     }
 
     // We are in!
@@ -152,35 +170,35 @@ public class LoginFlow implements MutatingFlow {
 
   /** Registrar with this ID could not be found. */
   static class BadRegistrarIdException extends AuthenticationErrorException {
-    public BadRegistrarIdException(String registrarId) {
+    BadRegistrarIdException(String registrarId) {
       super("Registrar with this ID could not be found: " + registrarId);
     }
   }
 
   /** Registrar login failed too many times. */
   static class TooManyFailedLoginsException extends AuthenticationErrorClosingConnectionException {
-    public TooManyFailedLoginsException() {
+    TooManyFailedLoginsException() {
       super("Registrar login failed too many times");
     }
   }
 
   /** Registrar account is not active. */
   static class RegistrarAccountNotActiveException extends AuthorizationErrorException {
-    public RegistrarAccountNotActiveException() {
+    RegistrarAccountNotActiveException() {
       super("Registrar account is not active");
     }
   }
 
   /** Registrar is already logged in. */
   static class AlreadyLoggedInException extends CommandUseErrorException {
-    public AlreadyLoggedInException() {
+    AlreadyLoggedInException() {
       super("Registrar is already logged in");
     }
   }
 
   /** Specified language is not supported. */
   static class UnsupportedLanguageException extends ParameterValuePolicyErrorException {
-    public UnsupportedLanguageException() {
+    UnsupportedLanguageException() {
       super("Specified language is not supported");
     }
   }
--- a/core/src/main/java/google/registry/model/console/User.java
+++ b/core/src/main/java/google/registry/model/console/User.java
@@ -25,6 +25,7 @@ import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
 import google.registry.model.Buildable;
 import google.registry.model.UpdateAutoTimestampEntity;
 import google.registry.persistence.VKey;
+import google.registry.util.PasswordUtils;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
@@ -84,8 +85,9 @@ public class User extends UpdateAutoTimestampEntity implements Buildable {
         || isNullOrEmpty(registryLockPasswordHash)) {
       return false;
     }
-    return hashPassword(registryLockPassword, registryLockPasswordSalt)
-        .equals(registryLockPasswordHash);
+    return PasswordUtils.verifyPassword(
+            registryLockPassword, registryLockPasswordHash, registryLockPasswordSalt)
+        .isPresent();
   }
 
   /**
@@ -154,9 +156,9 @@ public class User extends UpdateAutoTimestampEntity implements Buildable {
           !getInstance().hasRegistryLockPassword(), "User already has a password, remove it first");
       checkArgument(
           !isNullOrEmpty(registryLockPassword), "Registry lock password was null or empty");
-      getInstance().registryLockPasswordSalt = base64().encode(SALT_SUPPLIER.get());
-      getInstance().registryLockPasswordHash =
-          hashPassword(registryLockPassword, getInstance().registryLockPasswordSalt);
+      byte[] salt = SALT_SUPPLIER.get();
+      getInstance().registryLockPasswordSalt = base64().encode(salt);
+      getInstance().registryLockPasswordHash = hashPassword(registryLockPassword, salt);
       return this;
     }
   }
--- a/core/src/main/java/google/registry/model/registrar/Registrar.java
+++ b/core/src/main/java/google/registry/model/registrar/Registrar.java
@@ -60,6 +60,8 @@ import google.registry.model.tld.Tld;
 import google.registry.model.tld.Tld.TldType;
 import google.registry.persistence.VKey;
 import google.registry.util.CidrAddressBlock;
+import google.registry.util.PasswordUtils;
+import google.registry.util.PasswordUtils.HashAlgorithm;
 import java.security.cert.CertificateParsingException;
 import java.util.Comparator;
 import java.util.List;
@@ -97,7 +99,7 @@ import org.joda.time.DateTime;
     column = @Column(nullable = false, name = "lastUpdateTime"))
 public class Registrar extends UpdateAutoTimestampEntity implements Buildable, Jsonifiable {
 
-  /** Represents the type of a registrar entity. */
+  /** Represents the type of registrar entity. */
   public enum Type {
     /** A real-world, third-party registrar. Should have non-null IANA and billing account IDs. */
     REAL(Objects::nonNull),
@@ -376,7 +378,7 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
    */
   @Expose String icannReferralEmail;
 
-  /** Id of the folder in drive used to publish information for this registrar. */
+  /** ID of the folder in drive used to publish information for this registrar. */
   @Expose String driveFolderId;
 
   // Metadata.
@@ -639,7 +641,11 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
   }
 
   public boolean verifyPassword(String password) {
-    return hashPassword(password, salt).equals(passwordHash);
+    return getCurrentHashAlgorithm(password).isPresent();
+  }
+
+  public Optional<HashAlgorithm> getCurrentHashAlgorithm(String password) {
+    return PasswordUtils.verifyPassword(password, passwordHash, salt);
   }
 
   public String getPhonePasscode() {
@@ -861,8 +867,9 @@ public class Registrar extends UpdateAutoTimestampEntity implements Buildable, J
       checkArgument(
           Range.closed(6, 16).contains(nullToEmpty(password).length()),
           "Password must be 6-16 characters long.");
-      getInstance().salt = base64().encode(SALT_SUPPLIER.get());
-      getInstance().passwordHash = hashPassword(password, getInstance().salt);
+      byte[] salt = SALT_SUPPLIER.get();
+      getInstance().salt = base64().encode(salt);
+      getInstance().passwordHash = hashPassword(password, salt);
       return this;
     }
 
--- a/core/src/main/java/google/registry/model/registrar/RegistrarPoc.java
+++ b/core/src/main/java/google/registry/model/registrar/RegistrarPoc.java
@@ -37,6 +37,8 @@ import google.registry.model.Jsonifiable;
 import google.registry.model.UnsafeSerializable;
 import google.registry.model.registrar.RegistrarPoc.RegistrarPocId;
 import google.registry.persistence.VKey;
+import google.registry.util.PasswordUtils;
+import google.registry.util.PasswordUtils.HashAlgorithm;
 import java.io.Serializable;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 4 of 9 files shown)


jianglai (commented):
> Reviewed 7 of 7 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 130 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx-os6Z7jObZY7tyP:-NhWx-os6Z7jObZY7tyQ:bya7a9v) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L130)):*
> ```Java
>               .orElseThrow(() -> new ForbiddenException("User is not logged in"));
> 
>       String userEmail = verifyPasswordAndGetEmail(userAuthInfo, postInput);
> ```

Just move this line inside the `transact()` block.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 227 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx2c66SdCHNi3Ep8K:-NhWx2c66SdCHNi3Ep8L:b-4i9tb5) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L227)):*
> ```Java
>     if (registrarPoc.getCurrentHashAlgorithm(postInput.password) != HashAlgorithm.SCRYPT) {
>       logger.atInfo().log("Rehashing existing registry lock password with Scrypt.");
>       tm().transact(
> ```

You won't need this inner transaction if you make my recommended change above.

In general we shouldn't be adding transactions anymore in existing code, merely taking the existing transactions and making them broader.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 44 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWxDzl0oR3hD4zaKjf:-NhWxDzl0oR3hD4zaKjg:b4bxkec) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/util/src/main/java/google/registry/util/PasswordUtils.java#L44)):*
> ```Java
>     } catch (NoSuchAlgorithmException e) {
>       // All implementations of MessageDigest are required to support SHA-256.
>       throw new RuntimeException(
> ```

I'm worried that throwing an exception like this in the static initializer might have broad consequences, potentially causing instances to fail to spin up at all, and thus be unable to process anything, even stuff unrelated to password utils?

It would be better to memoize the creation of this -- initialize it at first use (and throw an exception then on failure).

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 130 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx-os6Z7jObZY7tyP:-NhXISd99peEbNO1_3gz:b-eillbs) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L130)):*

Just move this line inside the `transact()` block.
</blockquote></details>

I tried it first, but this method throws a checked exception, which needs to be handled. It doesn't seem worth the hassle because:

1) Re-save a temporary measure.
2) This action is not on a critical path and is not likely to be used a lot.
3) We need to solve the issue of the `transact()` method(s) doesn't take a lambda that throws check exceptions more broadly anyway, if we want to achieve the goal of eliminating most of nested transactions, realistically, without having to add a massive amount of rethrows. As I suggested in #2133, I think the solution is to change the signatures of the `transact()` method(s) for them to allow lambdas with checked exceptions, and handle that checked exception within `transact()` itself. That PR seems to have stalled, so I am working on a PR that refactors `reTransact` and makes the necessary changes to the function signatures to allow for checked exceptions.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 227 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx2c66SdCHNi3Ep8K:-NhXJuG2AdSu9aPrvOnL:b4c3l3q) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L227)):*

You won't need this inner transaction if you make my recommended change above.

In general we shouldn't be adding transactions anymore in existing code, merely taking the existing transactions and making them broader.
</blockquote></details>

See above.

Also, this PR is still WIP. I need to add more unit tests and test re-save feature on alpha.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 44 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWxDzl0oR3hD4zaKjf:-NhXGd2w3I2OQRhkn6Ng:b-slxl9y) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/util/src/main/java/google/registry/util/PasswordUtils.java#L44)):*

I'm worried that throwing an exception like this in the static initializer might have broad consequences, potentially causing instances to fail to spin up at all, and thus be unable to process anything, even stuff unrelated to password utils?

It would be better to memoize the creation of this -- initialize it at first use (and throw an exception then on failure).
</blockquote></details>

In theory you are right. In practice we know that it's been working for years.

Changed to supplier with memoization.

jianglai (commented):
Tested on crash. Ready for review now.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191#-:-NhbVKJg3AoCy-zPlHcN:bilt6ve)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @ptkach)

jianglai:
Gentle ping.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman, @jianglai, and @ptkach)

___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 52 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAyyI65QxLEgRYsSGZ:-NiAyyI65QxLEgRYsSG_:bvzesf) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L52)):*
> ```Java
>   private PasswordUtils() {}
> 
>   public enum HashAlgorithm {
> ```

Needs a Javadoc.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 84 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAywcK49ZM8BfrifBP:-NiAywcK49ZM8BfrifBQ:bvzesf) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L84)):*
> ```Java
> 
>   public static String hashPassword(String password, byte[] salt, HashAlgorithm algorithm) {
>     return base64().encode(algorithm.hash(password.getBytes(US_ASCII), salt));
> ```

Needs a Javadoc.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAyvCgDM6IAFhi-x23:-NiAyvCgDM6IAFhi-x24:bvzesf) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L88)):*
> ```Java
> 
>   @Nullable
>   public static HashAlgorithm verifyPassword(String password, String hash, String salt) {
> ```

Needs a Javadoc.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, @jianglai, and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 130 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx-os6Z7jObZY7tyP:-NiAz3p53E9GXna2RnjX:b-sdr66f) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L130)):*

I tried it first, but this method throws a checked exception, which needs to be handled. It doesn't seem worth the hassle because:

1) Re-save a temporary measure.
2) This action is not on a critical path and is not likely to be used a lot.
3) We need to solve the issue of the `transact()` method(s) doesn't take a lambda that throws check exceptions more broadly anyway, if we want to achieve the goal of eliminating most of nested transactions, realistically, without having to add a massive amount of rethrows. As I suggested in #2133, I think the solution is to change the signatures of the `transact()` method(s) for them to allow lambdas with checked exceptions, and handle that checked exception within `transact()` itself. That PR seems to have stalled, so I am working on a PR that refactors `reTransact` and makes the necessary changes to the function signatures to allow for checked exceptions.
</blockquote></details>

So this should wait until after #2195 then?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 130 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx-os6Z7jObZY7tyP:-NiFOQwSAuHwzU1VbzdP:bgie7mg) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L130)):*

So this should wait until after #2195 then?
</blockquote></details>

Ideally, yes. But I think registry lock action is not used frequently enough for us to worry about the performance hit with one more transaction. More importantly, it is better for us to get this change deployed so we can start converting SHA256 to Scrypt hashes and communicating with registrars regarding a conversion deadline. Even after #2195 is merged there will likely be non-trivial amount of work to fully unravel the nested transactions. I'd prefer to work on that separately and not have this change wait on that. 
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 52 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAyyI65QxLEgRYsSGZ:-NiFQFzo-tXQZS3-osXJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L52)):*

Needs a Javadoc.
</blockquote></details>

Done.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 84 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAywcK49ZM8BfrifBP:-NiFQGHaDUEKuDnO3ZAJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L84)):*

Needs a Javadoc.
</blockquote></details>

Done.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAyvCgDM6IAFhi-x23:-NiFRHXaEOjHC8xvBVQP:b-896fix) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L88)):*

Needs a Javadoc.
</blockquote></details>

Done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman, @jianglai, and @ptkach)

___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFb74gD9wNpTYi3OpC:-NiFb74gD9wNpTYi3OpD:b70zea0) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/util/src/main/java/google/registry/util/PasswordUtils.java#L102)):*
> ```Java
> 
>   /**
>    * Verifies a password by regerating the hash with the provided salt and comparing it to the
> ```

`regerating`?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman, @jianglai, and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 232 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFbq9n3FRGDmBGN6tg:-NiFbq9n3FRGDmBGN6th:b-rqndbb) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L232)):*
> ```Java
>                         tm().loadByEntity(registrarPoc)
>                             .asBuilder()
>                             .setAllowedToSetRegistryLockPassword(true)
> ```

What is this coming from?
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFb74gD9wNpTYi3OpC:-NiFcDj628Li2A9m3q8v:bqk4wpr) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/util/src/main/java/google/registry/util/PasswordUtils.java#L102)):*

`regerating`?
</blockquote></details>

Also the description needs to make it clear that it returns the algorithm if the hash is valid, otherwise returns null (or empty as suggested below).
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 111 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFc3jQ48NkFWYD-WUQ:-NiFc3jQ48NkFWYD-WUR:bomz80z) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/util/src/main/java/google/registry/util/PasswordUtils.java#L111)):*
> ```Java
>    *     works.
>    */
>   public static HashAlgorithm verifyPassword(String password, String hash, String salt) {
> ```

`Optional<HashAlgorithm>` as the return type will make this easier to understand.

jianglai (commented):
> Reviewed 9 of 9 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 130 at r2](https://reviewable.io/reviews/google/nomulus/2191#-NhWx-os6Z7jObZY7tyP:-NiG15DP-RZJJ4wxQ6Ly:b-896fix) ([raw file](https://github.com/google/nomulus/blob/31f77bb80f27e95bc6e9f10d569a21ee978a3814/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L130)):*

Leave a TODO then at least?
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 232 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFbq9n3FRGDmBGN6tg:-NiG19bf25MM8AEhtihg:bczd0ft) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L232)):*

What is this coming from?
</blockquote></details>

LoL. I'm glad you asked. For some reason we made it so that one can only set the password once. This boolean will need to be true (setting it to be true also [nulls out any existing password](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/registrar/RegistrarPoc.java;l=433?q=registrarpoc&ss=nomulus%2Fnomulus)) for one to set the password. Once the password is set, the boolean is [reverted back to false](https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/model/registrar/RegistrarPoc.java;l=433?q=registrarpoc&ss=nomulus%2Fnomulus), preventing any further change to the password by the registrar contact, unless they contact the admins to reset the password.

I honestly don't know why it was configured this way (basically preventing self-service password change or reset). Gus doesn't seem to know the full reason behind it either.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFb74gD9wNpTYi3OpC:-NiG2oBY0t638l98KhHK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/util/src/main/java/google/registry/util/PasswordUtils.java#L102)):*

Also the description needs to make it clear that it returns the algorithm if the hash is valid, otherwise returns null (or empty as suggested below).
</blockquote></details>

Done.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 111 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFc3jQ48NkFWYD-WUQ:-NiG2n3qCF5t-NoIwZMK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/util/src/main/java/google/registry/util/PasswordUtils.java#L111)):*

`Optional<HashAlgorithm>` as the return type will make this easier to understand.
</blockquote></details>

Done.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ptkach)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 5 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ptkach)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman, @jianglai, and @ptkach)

___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 165 at r8](https://reviewable.io/reviews/google/nomulus/2191#-NiKt3Hb2IVe-_tI7X0a:-NiKt3Hb2IVe-_tI7X0b:b-jw13iw) ([raw file](https://github.com/google/nomulus/blob/bd171abfaf3b224a879f09e1a936c3e638fdcec4/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L165)):*
> ```Java
>     assertThat(registrarPoc.getCurrentHashAlgorithm("hi").get()).isEqualTo(SHA256);
>     Map<String, ?> response = action.handleJsonRequest(lockRequest());
>     registrarPoc =
> ```

Don't reuse the existing variable here. This should be `RegistrarPoc updatedRegistrarPoc`

jianglai (commented):
> Reviewed 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @ptkach)

___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 165 at r8](https://reviewable.io/reviews/google/nomulus/2191#-NiKt3Hb2IVe-_tI7X0a:-NiLbqoH6UPS9Kl_S1ax:b85lwkw) ([raw file](https://github.com/google/nomulus/blob/bd171abfaf3b224a879f09e1a936c3e638fdcec4/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L165)):*

Don't reuse the existing variable here. This should be `RegistrarPoc updatedRegistrarPoc`
</blockquote></details>

Done. Personally I don't see why this is necessary. It's in test and we are reloading the registrar POC after the action. 
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 84 at r4](https://reviewable.io/reviews/google/nomulus/2191#-NiAywcK49ZM8BfrifBP:-NiLc7lB7FVcYIR-jzGP:b-h83gap) ([raw file](https://github.com/google/nomulus/blob/ff11f0af9a0fea1a2068fc6cc433f396ffdbeda2/util/src/main/java/google/registry/util/PasswordUtils.java#L84)):*

Is it?
</blockquote></details>

Sorry, forgot to push the latest commit.
___
*[`util/src/main/java/google/registry/util/PasswordUtils.java` line 102 at r5](https://reviewable.io/reviews/google/nomulus/2191#-NiFb74gD9wNpTYi3OpC:-NiLcEq28Y8Frq3hzroc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/f57a3b13cd0e4bb0dd8c24a77f7a01dbadb2f9f0/util/src/main/java/google/registry/util/PasswordUtils.java#L102)):*

It still says `regerating`
</blockquote></details>

Done.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman and @ptkach)

___
*[`core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java` line 165 at r8](https://reviewable.io/reviews/google/nomulus/2191#-NiKt3Hb2IVe-_tI7X0a:-NiLfRJL05ipIhdl4ljq:b-xa1928) ([raw file](https://github.com/google/nomulus/blob/bd171abfaf3b224a879f09e1a936c3e638fdcec4/core/src/test/java/google/registry/ui/server/registrar/RegistryLockPostActionTest.java#L165)):*

Done. Personally I don't see why this is necessary. It's in test and we are reloading the registrar POC after the action.
</blockquote></details>

Even tests are read by people though, and reusing variables makes the code harder to understand.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2191#-:-NiLfcfO0KacRl4QlzmM:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @gbrodman and @ptkach)

PR #2192, opened 2023-10-24
Remove node.js download instruction

ptkach:
node.js is already installed as a part of gradle-runner.sh, so I don't think this download is necessary

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2192)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -60,9 +60,8 @@ dependencyLocking {
 }
 
 node {
-  download = true
-  version = "16.14.0"
-  npmVersion = "6.14.11"
+  download = false
+  version = "16.19.0"
 }
 
 wrapper {


jianglai (changes requested):
Isn't it better to download the newer version here? This way the newest version will always be present, regardless where you run it. We can then delete the download directive in gradle-runner.sh.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2192#-:-NhXaWGv3Of2EKMoPlgi:b30akjy)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

ptkach (commented):
no, we don't want automatically switch to a newer version of node or npm, as that can cause inconsistencies with builds and runtime tasks, like linter, tests and dependencies inconsistencies between kokoro and local/dev machines 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2192#-:-NhXb-9A1uc8z4P-flxf:bvdz6bu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai (changes requested):
Doesn't the config here lock to Node version 16.14.0 (which is what will be used to build, I assume)? If anything, it's better to lock the version via Gradle so we always use the same version. In your example, there's no guarantee that whoever is running the build locally happens to have the same version of Node as Kokoro, which is locked to 16.19.0.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2192#-:-NhY-T1I8brwpc4zbuvg:bnsi1mu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

ptkach (commented):
I agree with you, however I don't see how bumping the version to 16.19.0 will resolve the issue we're observing with timeouts. It periodically hangs while downloading the 16.14.0 and periodically it doesn't, so chances are the same will happen to 16.19.0 if updated. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2192#-:-NhYFJ6e5xfQlrKwAoSc:b-2w5o43)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai (changes requested):
I'm a bit confused. Didn't you say that 16.19.0 was downloaded successfully? As long as we download from NodeJS's website, there's always a chance that this happens.

My point was more that if we are going to only download Node once, here is the better place to do it, rather than on kokoro. Whether we should do something to make sure the download is an orthogonal issue (that we should consider solving now that we've encountered problems with downloading several times). 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2192#-:-Nhb0dH_6j-Ql3Lg4r7K:b4duf3v)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai (approved):
Ah, I missed the part where the node version was downloaded from GCS. That surely is more reliable than from the Internet.

Since that will be the version that is eventually used on kokoro, should we change the version here to match that? I don't know what use does the version number here have if we are not asking gradle to download it, but still, it's better to keep things consistent, if just for documentation, right?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2192#-:-NhbtXGF3OP6mIo_7kBc:blm57c4)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #2193, opened 2023-10-24
Add a console action to retrieve a paged list of domains

gbrodman:
In the future we'll want to add searching capability but for now we can go with straightforward pagination. 

Note: we will need to add a GSON VKey converter in the future to parse domains, but for now we're not using the domains anywhere

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2193)

### DIFF
--- a/core/src/main/java/google/registry/model/domain/DomainBase.java
+++ b/core/src/main/java/google/registry/model/domain/DomainBase.java
@@ -128,14 +128,14 @@ public class DomainBase extends EppResource
   String tld;
 
   /** References to hosts that are the nameservers for the domain. */
-  @Expose @Transient Set<VKey<Host>> nsHosts;
+  @Transient Set<VKey<Host>> nsHosts;
 
   /** Contacts. */
-  @Expose VKey<Contact> adminContact;
+  VKey<Contact> adminContact;
 
-  @Expose VKey<Contact> billingContact;
-  @Expose VKey<Contact> techContact;
-  @Expose VKey<Contact> registrantContact;
+  VKey<Contact> billingContact;
+  VKey<Contact> techContact;
+  VKey<Contact> registrantContact;
 
   /** Authorization info (aka transfer secret) of the domain. */
   @Embedded
--- a/core/src/main/java/google/registry/module/frontend/FrontendRequestComponent.java
+++ b/core/src/main/java/google/registry/module/frontend/FrontendRequestComponent.java
@@ -26,6 +26,7 @@ import google.registry.request.RequestComponentBuilder;
 import google.registry.request.RequestModule;
 import google.registry.request.RequestScope;
 import google.registry.ui.server.console.ConsoleDomainGetAction;
+import google.registry.ui.server.console.ConsoleDomainListAction;
 import google.registry.ui.server.console.ConsoleUserDataAction;
 import google.registry.ui.server.console.RegistrarsAction;
 import google.registry.ui.server.console.settings.ContactAction;
@@ -54,6 +55,8 @@ import google.registry.ui.server.registrar.RegistryLockVerifyAction;
     })
 interface FrontendRequestComponent {
   ConsoleDomainGetAction consoleDomainGetAction();
+
+  ConsoleDomainListAction consoleDomainListAction();
 
   ConsoleOteSetupAction consoleOteSetupAction();
   ConsoleRegistrarCreatorAction consoleRegistrarCreatorAction();
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java
@@ -1,0 +1,148 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.ui.server.console;
+
+import static google.registry.model.console.ConsolePermission.DOWNLOAD_DOMAINS;
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.Gson;
+import com.google.gson.annotations.Expose;
+import google.registry.model.CreateAutoTimestamp;
+import google.registry.model.console.User;
+import google.registry.model.domain.Domain;
+import google.registry.request.Action;
+import google.registry.request.Parameter;
+import google.registry.request.Response;
+import google.registry.request.auth.Auth;
+import google.registry.request.auth.AuthResult;
+import google.registry.ui.server.registrar.JsonGetAction;
+import java.util.List;
+import java.util.Optional;
+import javax.inject.Inject;
+import org.joda.time.DateTime;
+
+/** Returns a (paginated) list of domains for a particular registrar. */
+@Action(
+    service = Action.Service.DEFAULT,
+    path = ConsoleDomainListAction.PATH,
+    method = Action.Method.GET,
+    auth = Auth.AUTH_PUBLIC_LOGGED_IN)
+public class ConsoleDomainListAction implements JsonGetAction {
+
+  public static final String PATH = "/console-api/domain-list";
+
+  private static final int DEFAULT_RESULTS_PER_PAGE = 50;
+  private static final String DOMAIN_QUERY_TEMPLATE =
+      "FROM Domain WHERE currentSponsorRegistrarId = :registrarId AND deletionTime >"
+          + " :deletedAfterTime AND creationTime <= :createdBeforeTime";
+
+  private final AuthResult authResult;
+  private final Response response;
+  private final Gson gson;
+  private final String registrarId;
+  private final Optional<DateTime> checkpointTime;
+  private final int pageNumber;
+  private final int resultsPerPage;
+  private final Optional<Long> totalResults;
+
+  @Inject
+  public ConsoleDomainListAction(
+      AuthResult authResult,
+      Response response,
+      Gson gson,
+      @Parameter("registrarId") String registrarId,
+      @Parameter("checkpointTime") Optional<DateTime> checkpointTime,
+      @Parameter("pageNumber") Optional<Integer> pageNumber,
+      @Parameter("resultsPerPage") Optional<Integer> resultsPerPage,
+      @Parameter("totalResults") Optional<Long> totalResults) {
+    this.authResult = authResult;
+    this.response = response;
+    this.gson = gson;
+    this.registrarId = registrarId;
+    this.checkpointTime = checkpointTime;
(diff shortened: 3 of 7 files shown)


ptkach (commented):
That looks great, is it deployed somewhere? 

> Reviewed 2 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193#-:-Nhc9QRN5kNvRkMuiXtM:b-jxl4ue)* status: 2 of 7 files reviewed, all discussions resolved

gbrodman (commented):
nah, I didn't deploy it anywhere because it doesn't have a front end component 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193#-:-NhcMeJt6WFbRu6N4C2O:b-jm56rn)* status: 2 of 7 files reviewed, all discussions resolved (waiting on @ptkach)

ptkach (changes requested):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/2193#-NhhQavO2wWUB_FuMUPP:-NhhQavO2wWUB_FuMUPQ:b-837oi1) ([raw file](https://github.com/google/nomulus/blob/1492964fab8b0c0e2e900dc8121ac5c5530e2005/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java#L98)):*
> ```Java
>     // We have to use a constant created-before time in order to have stable pagination, since new
>     // domains can be constantly created
>     CreateAutoTimestamp createdBeforeTime;
> ```

To keep track, as discussed, we probably want to reset this if it comes older than an hour as someone might get stuck with older results if the same url reused

gbrodman (commented):
Discussed offline: I shifted this to be a list of individual parameters because we cannot use the body of GET requests. 

Deploying to crash now 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193#-:-NhhQnIr9-11MkRObgNZ:bxe4dit)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

gbrodman (commented):
some sample curl commands on crash take about 2.5-3.5 seconds which is not great but not terrible 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193#-:-Nhlnl0o9hP8lqgIDBlJ:boflke6)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/2193#-NhhQavO2wWUB_FuMUPP:-NhmM-QKErUP-OpUJspt:bg5cmfo) ([raw file](https://github.com/google/nomulus/blob/1492964fab8b0c0e2e900dc8121ac5c5530e2005/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java#L98)):*

To keep track, as discussed, we probably want to reset this if it comes older than an hour as someone might get stuck with older results if the same url reused
</blockquote></details>

i was thinking about this for a while and I think it makes more sense to handle that on the front end 

1. it's quicker to display to the user what's going on that way 
2. it means that the back end is giving back to the front end exactly what it requested every time 
3. in the case where we'd want to reset it, we don't have a great way of communicating to the front end what we're doing. Whereas if the front end does it, we can have a check there that says basically "if the time is old, reload the thing without a time" 
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 214 at r2](https://reviewable.io/reviews/google/nomulus/2193#gh-1373731513:-Nhhrf_Y5AEkNM0rBZfx:b-nbuf10) ([raw file](https://github.com/google/nomulus/blob/d2abfde74edc4a589005ff4c818878de16d9e8ac/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L214)):*

## Missing catch of NumberFormatException

Potential uncaught 'java.lang.NumberFormatException'.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1578)
</blockquote></details>

this exception shouldn't be relevant

ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/2193#-NhhQavO2wWUB_FuMUPP:-Ni0DBM6B4A2OnY5JW9H:bln0swv) ([raw file](https://github.com/google/nomulus/blob/1492964fab8b0c0e2e900dc8121ac5c5530e2005/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java#L98)):*
I think we need to think it through a little more. 
I agree with you that handing it on front-end side makes more sense and will provide a better transparency to the user. However from user's perspective there's still inconsistency in this implementation:

> it means that the back end is giving back to the front end exactly what it requested every time

Not exactly, as after the first visit user gets the "limited" list of `createdBeforeNow` domains. User doesn't get to request this value, but gets to be responsible for resetting it, which in turn will provide a new `createdBeforeNow` for the user

Another point is that this implementation takes care of new domains created after user's request, but it doesn't take into account the domains deleted during user session. That enables a possibility with providing the user with N domains and P pages on the first load and then during navigation that can switch to N - D domains and potentially less pages.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/2193#-NhhQavO2wWUB_FuMUPP:-Ni0NV2LC-AHnkdCY8MO:b-nkshv3) ([raw file](https://github.com/google/nomulus/blob/1492964fab8b0c0e2e900dc8121ac5c5530e2005/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java#L98)):*

I think we need to think it through a little more. 
I agree with you that handing it on front-end side makes more sense and will provide a better transparency to the user. However from user's perspective there's still inconsistency in this implementation:

> it means that the back end is giving back to the front end exactly what it requested every time

Not exactly, as after the first visit user gets the "limited" list of `createdBeforeNow` domains. User doesn't get to request this value, but gets to be responsible for resetting it, which in turn will provide a new `createdBeforeNow` for the user

Another point is that this implementation takes care of new domains created after user's request, but it doesn't take into account the domains deleted during user session. That enables a possibility with providing the user with N domains and P pages on the first load and then during navigation that can switch to N - D domains and potentially less pages.
</blockquote></details>

yeah there can be inconsistency, but after thinking about it this is pretty straightforward I think because we can also use deletion time (created before or at creationTime, deletion after deletionTime, both of which default to the time at the first request) 

the front end can check whether or not the creationTime that it's about to request is reasonable and remove it if not -- my goal is to make the backend API do exactly what the front end requests, and it's the front end's responsibility to request non-silly things. Things get very complex quickly if the backend returns things that don't match what the frontend requested.

ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/2193#-NhhQavO2wWUB_FuMUPP:-Ni0blH6E35xmPkoRBJO:b-l4djva) ([raw file](https://github.com/google/nomulus/blob/1492964fab8b0c0e2e900dc8121ac5c5530e2005/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java#L98)):*

yeah there can be inconsistency, but after thinking about it this is pretty straightforward I think because we can also use deletion time (created before or at creationTime, deletion after deletionTime, both of which default to the time at the first request) 

the front end can check whether or not the creationTime that it's about to request is reasonable and remove it if not -- my goal is to make the backend API do exactly what the front end requests, and it's the front end's responsibility to request non-silly things. Things get very complex quickly if the backend returns things that don't match what the frontend requested.
</blockquote></details>

As discussed offline, please rename the variable to be something like 'snapshotTime' and update the query to account for the deleted after that said time. The rest should be taken care by front-end

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java` line 98 at r1](https://reviewable.io/reviews/google/nomulus/2193#-NhhQavO2wWUB_FuMUPP:-Ni1-z31AG3xdm4Geki_:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1492964fab8b0c0e2e900dc8121ac5c5530e2005/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java#L98)):*

As discussed offline, please rename the variable to be something like 'snapshotTime' and update the query to account for the deleted after that said time. The rest should be taken care by front-end
</blockquote></details>

Done.

ptkach (approved):
> Reviewed 2 of 7 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 194 at r3](https://reviewable.io/reviews/google/nomulus/2193#-Ni15zPP7t5JJDZBUL57:-Ni15zPP7t5JJDZBUL58:b-k5ia1e) ([raw file](https://github.com/google/nomulus/blob/bb95a939d2059435cebcc305e8b3100573519be7/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L194)):*
> ```Java
> 
>   @Provides
>   @Parameter("checkpointTime")
> ```

Not major, but since this is global to a request namespace in order to avoid collision in the future, we might want to have have it named after a request, like `domains-checkointTime` unless it's in fact expected to be used across different screens. Thoughts?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2193)* status: all files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 194 at r3](https://reviewable.io/reviews/google/nomulus/2193#-Ni15zPP7t5JJDZBUL57:-Ni1AUGKB1qQIVhLyj4M:bh6gq84) ([raw file](https://github.com/google/nomulus/blob/bb95a939d2059435cebcc305e8b3100573519be7/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L194)):*

Not major, but since this is global to a request namespace in order to avoid collision in the future, we might want to have have it named after a request, like `domains-checkointTime` unless it's in fact expected to be used across different screens. Thoughts?
</blockquote></details>

I don't think that particularly matters. This way we can reuse the injection if we, for some reason, want to have a checkpoint time somewhere else. Even if the usage is entirely different, the "inject an optional DateTime object" can be the same.

PR #2196, opened 2023-10-26
Change the default config for perTransactionIsolation to true

sarahcaseybot:
This was already set to true in all environments except prod last week. Now that the release has gone out and we have not seen any issues, we should feel safe turning this on in production as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2196)


jianglai:
Remember to remove the overrides in the private repo as well.

jianglai (approved):

PR #2204, opened 2023-11-01
Add BSA schema

weiminyu:
Add database tables for BSA integration.

Entity classes are in go/r3pr/2205

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2204)


weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2204)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @jianglai and @sarahcaseybot)

___
*[`db/src/main/resources/sql/flyway/V148__add_bsa_download_and_label_tables.sql` line 26 at r2](https://reviewable.io/reviews/google/nomulus/2204#-NiBVRK56eo3-xgvCSow:-NiBhsMV1UIWR4wsYmKs:b-cneqd6) ([raw file](https://github.com/google/nomulus/blob/e1a81fa05a3889a6a75b2fa9540322a849260f61/db/src/main/resources/sql/flyway/V148__add_bsa_download_and_label_tables.sql#L26)):*

Sorry if this was already asked and discussed in a design doc somewhere, but why is there a creation time stored here? Why not use the job_id that the label first showed up in? Wouldn't  we potentially want to link the BsaLabel table to the BsaDownload table?
</blockquote></details>

Good question. The creation time will be set to the job's creationTime, which will be used as folder name on GCS.
The association with the download files is the only important one.

You question reminds me to change the entity class of BsaLabel to not use CreationAutoTimestamp.

sarahcaseybot (approved):
:lgtm:

> Reviewed 4 of 5 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2204#-:-NiGSl-VD3cCmdK38d1Z:bnfp4nl)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`db/src/main/resources/sql/flyway/V149__add_bsa_domain_in_use_table.sql` line 18 at r3](https://reviewable.io/reviews/google/nomulus/2204#-NiGRXQBAZNyhZj1O1f0:-NiGRXQBAZNyhZj1O1f1:b-bkrbgs) ([raw file](https://github.com/google/nomulus/blob/ce4a33061421a12def508228337fe9991046bf29/db/src/main/resources/sql/flyway/V149__add_bsa_domain_in_use_table.sql#L18)):*
> ```TSQL
>     label text not null,
>     tld text not null,
>     creation_time timestamptz not null,
> ```

This is a confusing field name since I initially thought it was referring to the creation time of the domain in use's registration if it was already registered, but looking at the code in go/r3pr/2205 it looks like it is referring to the time the name was first requested to be blocked. Maybe a better name would be bsa_creation_time or block_creation_time?

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2204)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on @jianglai and @sarahcaseybot)

___
*[`db/src/main/resources/sql/flyway/V149__add_bsa_domain_in_use_table.sql` line 18 at r3](https://reviewable.io/reviews/google/nomulus/2204#-NiGRXQBAZNyhZj1O1f0:-NiGWejiCAQFGaUmX4Dq:b-aftjuq) ([raw file](https://github.com/google/nomulus/blob/ce4a33061421a12def508228337fe9991046bf29/db/src/main/resources/sql/flyway/V149__add_bsa_domain_in_use_table.sql#L18)):*

This is a confusing field name since I initially thought it was referring to the creation time of the domain in use's registration if it was already registered, but looking at the code in go/r3pr/2205 it looks like it is referring to the time the name was first requested to be blocked. Maybe a better name would be bsa_creation_time or block_creation_time?
</blockquote></details>

It is the time when we detect that it is in use. It could be during download processing, but could also be during one of the regular reconciliation cron jobs. It is for troubleshooting and has no real use in code. I'm worrying that giving more details will only invite more questions.

PR #2206, opened 2023-11-01
Update Debian 10 -> 11, replace node.js installation method

ptkach:
Updated builder docker base image from Debian10 to Debian11.
Updated node.js installation method. The version it used to install wasn't inline with the version we use everywhere else, now that's fixed.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2206)

### DIFF
--- a/release/builder/build.sh
+++ b/release/builder/build.sh
@@ -24,7 +24,7 @@ apt-get install gnupg2 -y
 # Install Java
 apt-get install openjdk-11-jdk-headless -y
 # Install Python
-apt-get install python -y
+apt-get install python3 -y
 # As of March 2021 python3 is at v3.6. Get pip then install dataclasses
 # (introduced in 3.7) for nom_build
 apt-get install python3-pip -y
@@ -32,8 +32,10 @@ python3 -m pip install dataclasses
 # Install curl.
 apt-get install curl -y
 # Install Node
-curl -sL https://deb.nodesource.com/setup_current.x | bash -
-apt-get install -y nodejs
+apt-get install npm -y
+npm cache clean -f
+npm install -g n
+n 16.19.0
 # Install gcloud
 # Cribbed from https://cloud.google.com/sdk/docs/quickstart-debian-ubuntu
 apt-get install lsb-release -y
--- a/release/builder/Dockerfile
+++ b/release/builder/Dockerfile
@@ -28,7 +28,7 @@ COPY go.sum ./
 COPY go.mod ./
 RUN go build -o /deployCloudSchedulerAndQueue
 
-FROM marketplace.gcr.io/google/debian10
+FROM marketplace.gcr.io/google/debian11
 ENV DEBIAN_FRONTEND=noninteractive LANG=en_US.UTF-8
 # Add script for cloud scheduler and cloud tasks deployment
 COPY --from=deployCloudSchedulerAndQueueBuilder /deployCloudSchedulerAndQueue /usr/local/bin/deployCloudSchedulerAndQueue


release/builder/build.sh
```diff
@@ -24,16 +24,22 @@ apt-get install gnupg2 -y
 # Install Java
 apt-get install openjdk-11-jdk-headless -y
 # Install Python
-apt-get install python -y
+echo ">>>>"
+apt-get install python
+echo ">>>> 2"
+apt-get install python3.8
```
mindhog, line 30: I don't think there is a python3.8 package in the repo, at least there's not in any of the ones I'm using -- `apt-get install python3 -y` is what you want.  In fact, there's no "python" package any more, the old version has become "python2".  This might be the reason for the problem you're seeing.

You need the -y to suppress the "are you sure you want to install..." prompt.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2206)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

___
*[`release/builder/build.sh` line 30 at r1](https://reviewable.io/reviews/google/nomulus/2206#gh-1379927974:-NiFHlo654cB32hD_zAY:b-6eyl29) ([raw file](https://github.com/google/nomulus/blob/0ddb4b7e1881ba70f1d119089a0e700f82ebe668/release/builder/build.sh#L30)):*

I don't think there is a python3.8 package in the repo, at least there's not in any of the ones I'm using -- `apt-get install python3 -y` is what you want.  In fact, there's no "python" package any more, the old version has become "python2".  This might be the reason for the problem you're seeing.

You need the -y to suppress the "are you sure you want to install..." prompt.  
</blockquote></details>

I concur with Mike, this change needs to be tested first. For example the echo commands obviously should not exist in the final PR.

Use gcloud to submit a build with uncommitted code to confirm that things are working, and clean up the echos.

PS: we should explain why Python needs to be update to 3.8 when the PR's purpose is to make changes to how Node is installed. Further more, it probably is a good time to update the base image to debian 11 as well. I suspect we will run into more and more issues if we stayed on the old version.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2206)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)

___
*[`release/builder/build.sh` line 30 at r1](https://reviewable.io/reviews/google/nomulus/2206#gh-1379927974:-NiFV5FV-TFPQl2FVJLw:bfyv1oi) ([raw file](https://github.com/google/nomulus/blob/0ddb4b7e1881ba70f1d119089a0e700f82ebe668/release/builder/build.sh#L30)):*

I concur with Mike, this change needs to be tested first. For example the echo commands obviously should not exist in the final PR.

Use gcloud to submit a build with uncommitted code to confirm that things are working, and clean up the echos.

PS: we should explain why Python needs to be update to 3.8 when the PR's purpose is to make changes to how Node is installed. Further more, it probably is a good time to update the base image to debian 11 as well. I suspect we will run into more and more issues if we stayed on the old version.
</blockquote></details>

Thanks Mike. What you see here is in fact me testing the change, thus `echo` and `python3.8` installation attempts. I created a PR once I saw that nomulus-release succeeded with the new Node.js installation, which's what was broken in the first place. Still no clue why python even came into picture, but I will update. However I don't feel like including Debian update here as well, to many moving parts at the same time.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2206)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jianglai and @mindhog)

___
*[`release/builder/build.sh` line 30 at r1](https://reviewable.io/reviews/google/nomulus/2206#gh-1379927974:-NiG27mZ5yadj0AIeE2k:bf3wnd2) ([raw file](https://github.com/google/nomulus/blob/0ddb4b7e1881ba70f1d119089a0e700f82ebe668/release/builder/build.sh#L30)):*

Thanks Mike. What you see here is in fact me testing the change, thus `echo` and `python3.8` installation attempts. I created a PR once I saw that nomulus-release succeeded with the new Node.js installation, which's what was broken in the first place. Still no clue why python even came into picture, but I will update. However I don't feel like including Debian update here as well, to many moving parts at the same time.
</blockquote></details>

Latest update succeeded both pipelines - `prepare-release` and `release-nomulus`. Just a simple `apt-get install python3` didn't work under Debian10, as it pulled Python3.7. Installing Python3.8 on Debian10 turned out to be a bit complicated, so I decided to go for Debian update to 11, which now has Python 3.9. Node.js seems to work fine as well.

jianglai (approved):
Update the PR description to note the fact that the base image was also updated.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2206#-:-NiG5eG1ETiCkFCL81zL:b-vkxtn6)* status: all files reviewed, 1 unresolved discussion (waiting on @mindhog)

PR #2211, opened 2023-11-06
Add handler for Console API requests and XSRF token creation and verification

ptkach:
While working on XSRF protection I realized that we're lacking a better way to handle API request to new console. For instance I never really liked that we used `JsonGetAction` to handle **POST** and GET requests, as well as the fact that we needed to verify `consoleUser` every time and check if the request is GET or POST, etc. On top of that there's no good place where to actually do the XSRF verification. All that prompted me to create a base class that we can use for Console Api requests - `ConsoleApiAction`. For now I've migrated `ConsoleUserDataAction` to work with the new `ConsoleApiAction` and if you're ok with this approach I will continue updating existing endpoints.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2211)

### DIFF
--- a/core/src/main/java/google/registry/request/Response.java
+++ b/core/src/main/java/google/registry/request/Response.java
@@ -15,6 +15,7 @@
 package google.registry.request;
 
 import com.google.common.net.MediaType;
+import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletResponse;
 import org.joda.time.DateTime;
 
@@ -51,4 +52,11 @@ public interface Response {
    * @see HttpServletResponse#setDateHeader(String, long)
    */
   void setDateHeader(String header, DateTime timestamp);
+
+  /**
+   * Adds a cookie to the response
+   *
+   * @see HttpServletResponse#addCookie(Cookie)
+   */
+  void addCookie(Cookie cookie);
 }
--- a/core/src/main/java/google/registry/request/ResponseImpl.java
+++ b/core/src/main/java/google/registry/request/ResponseImpl.java
@@ -17,6 +17,7 @@ package google.registry.request;
 import com.google.common.net.MediaType;
 import java.io.IOException;
 import javax.inject.Inject;
+import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletResponse;
 import org.joda.time.DateTime;
 
@@ -57,5 +58,10 @@ public final class ResponseImpl implements Response {
   @Override
   public void setDateHeader(String header, DateTime timestamp) {
     rsp.setDateHeader(header, timestamp.getMillis());
+  }
+
+  @Override
+  public void addCookie(Cookie cookie) {
+    rsp.addCookie(cookie);
   }
 }
--- a/core/src/main/java/google/registry/security/XsrfTokenManager.java
+++ b/core/src/main/java/google/registry/security/XsrfTokenManager.java
@@ -34,7 +34,7 @@ import org.joda.time.Duration;
 /** Helper class for generating and validate XSRF tokens. */
 public final class XsrfTokenManager {
 
-  /** HTTP header used for transmitting XSRF tokens. */
+  /** HTTP header or cookie name used for transmitting XSRF tokens. */
   public static final String X_CSRF_TOKEN = "X-CSRF-Token";
 
   /** POST parameter used for transmitting XSRF tokens. */
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java
@@ -1,0 +1,72 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.ui.server.console;
+
+import static google.registry.request.Action.Method.GET;
+
+import com.google.api.client.http.HttpStatusCodes;
+import google.registry.model.console.User;
+import google.registry.security.XsrfTokenManager;
+import google.registry.ui.server.registrar.ConsoleApiParams;
+import java.util.Arrays;
+import java.util.Optional;
+import javax.servlet.http.Cookie;
+
+/** Base class for handling Console API requests */
+public abstract class ConsoleApiAction implements Runnable {
+  protected ConsoleApiParams consoleApiParams;
+
+  public ConsoleApiAction(ConsoleApiParams consoleApiParams) {
+    this.consoleApiParams = consoleApiParams;
+  }
+
+  @Override
+  public final void run() {
+    // Shouldn't be even possible because of Auth annotations on the various implementing classes
+    if (!consoleApiParams.authResult().userAuthInfo().get().consoleUser().isPresent()) {
+      consoleApiParams.response().setStatus(HttpStatusCodes.STATUS_CODE_UNAUTHORIZED);
+      return;
+    }
+    User user = consoleApiParams.authResult().userAuthInfo().get().consoleUser().get();
+    if (consoleApiParams.request().getMethod().equals(GET.toString())) {
+      getHandler(user);
+    } else {
+      if (verifyXSRF()) {
+        postHandler(user);
+      }
+    }
+  }
+
+  protected void postHandler(User user) {
+    throw new UnsupportedOperationException("Console API POST handler not implemented");
+  }
+
+  protected void getHandler(User user) {
+    throw new UnsupportedOperationException("Console API GET handler not implemented");
+  }
+
+  private boolean verifyXSRF() {
+    Optional<Cookie> maybeCookie =
+        Arrays.stream(consoleApiParams.request().getCookies())
+            .filter(c -> XsrfTokenManager.X_CSRF_TOKEN.equals(c.getName()))
+            .findFirst();
+    if (!maybeCookie.isPresent()
+        || !consoleApiParams.xsrfTokenManager().validateToken(maybeCookie.get().getValue())) {
+      consoleApiParams.response().setStatus(HttpStatusCodes.STATUS_CODE_UNAUTHORIZED);
+      return false;
+    }
+    return true;
+  }
+}
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java
@@ -21,12 +21,10 @@ import com.google.common.collect.ImmutableMap;
 import google.registry.config.RegistryConfig.Config;
 import google.registry.model.console.User;
 import google.registry.request.Action;
-import google.registry.request.Response;
 import google.registry.request.auth.Auth;
-import google.registry.request.auth.AuthResult;
-import google.registry.request.auth.UserAuthInfo;
-import google.registry.ui.server.registrar.JsonGetAction;
+import google.registry.ui.server.registrar.ConsoleApiParams;
 import javax.inject.Inject;
+import javax.servlet.http.Cookie;
 import org.json.JSONObject;
 
 @Action(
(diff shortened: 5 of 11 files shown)


gbrodman (changes requested):
> Reviewed 3 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2211)* status: 3 of 9 files reviewed, 11 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/request/ResponseImpl.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/2211#gh-1383856575:-Nia_6vd59ceDQgFNtxt:b-5znq7q) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/request/ResponseImpl.java#L65)):*

## Failure to use secure cookies

Cookie is added to response without the 'secure' flag being set.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1597)
</blockquote></details>

this seems like the wrong place for this note, because one sets the "secure" status on the cookie itself, not here.....
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 1 at r1](https://reviewable.io/reviews/google/nomulus/2211#-NiakR1JElPCSjj3PO6b:-NiakR1JElPCSjj3PO6c:b-wqhi78) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L1)):*
> ```Java
> package google.registry.ui.server.console;
> ```

license
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 13 at r1](https://reviewable.io/reviews/google/nomulus/2211#-NiakQ7VDC2PD3rRrNpQ:-NiakQ7W8Br8lHDNzztc:bpnc71r) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L13)):*
> ```Java
> import javax.servlet.http.Cookie;
> 
> public class ConsoleApiAction implements Runnable {
> ```

needs a javadoc
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 22 at r1](https://reviewable.io/reviews/google/nomulus/2211#-NiaaoPp0eOe2KJKO9Ag:-NiaaoPp0eOe2KJKO9Ah:bke3s6m) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L22)):*
> ```Java
>   public final void run() {
>     if (!consoleApiParams.getAuthResult().userAuthInfo().get().consoleUser().isPresent()) {
>       consoleApiParams.getRsp().setStatus(HttpStatusCodes.STATUS_CODE_UNAUTHORIZED);
> ```

maybe note that this shouldn't ever be possible because of the Auth annotations on the various implementing classes 
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 34 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Niad59h3yDg3JdgiGa8:-Niad59i9IDM_s1dfYL3:b69pwey) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L34)):*
> ```Java
>     }
>   }
>   ;
> ```

extra ; 
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 37 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Niab0YREhRw-gJtSG03:-Niab0YREhRw-gJtSG04:bielhxq) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L37)):*
> ```Java
> 
>   protected void postHandler(User user) {
>     throw new RuntimeException("Console API POST handler not implemented");
> ```

UnsupportedOperationException (same below) 
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java` line 39 at r1](https://reviewable.io/reviews/google/nomulus/2211#-NiakkEME2pKHcfJaA0g:-NiakkEME2pKHcfJaA0h:bhaz4bz) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java#L39)):*
> ```Java
>   public static final String PATH = "/console-api/userdata";
> 
>   private final ConsoleApiParams consoleApiParams;
> ```

just make the superclass's variable protected
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java` line 61 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Nialz7CCcNxku0WM1Ch:-Nialz7D-04-gDDpj5u4:by5zdge) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java#L61)):*
> ```Java
> 
>   @Override
>   protected void getHandler(User user) {
> ```

Shouldn't we do this on all GET requests? Like, why not just do this in the base class for every GET request before we actually run the code? 
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 43 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Nia_aqHCpTTOIuXtWTj:-Nia_aqIDqQBUABl9d1-:bsegenm) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L43)):*
> ```Java
>   static final String PARAM_CLIENT_ID = "clientId";
> 
>   public static class ConsoleApiParams {
> ```

let's just move this to a new file and use an AutoValue, it doesn't need to be in here   
___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 46 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Nial_dY3nc3uFRs6Aw-:-Nial_dY3nc3uFRs6Aw0:bbbtboa) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L46)):*
> ```Java
>     private final XsrfTokenManager xsrfTokenManager;
>     private final HttpServletRequest req;
>     private final Response rsp;
> ```

No point in saving a few characters, just use `response` everywhere

Same for `request`

gbrodman (changes requested):
> Reviewed 2 of 9 files at r1, 5 of 6 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2211)* status: all files reviewed, 5 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 28 at r3](https://reviewable.io/reviews/google/nomulus/2211#-Nil-2Yp-1Jn0eD8zWd6:-Nil-2Yp-1Jn0eD8zWd7:bma2451) ([raw file](https://github.com/google/nomulus/blob/c42b59a6acadcabfda0ceceaff75a3906a7ee50e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L28)):*
> ```Java
> 
> /** Base class for handling Console API requests */
> public class ConsoleApiAction implements Runnable {
> ```

I don't think we'd ever want to instantiate this directly, right? If not we can make it an abstract class
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java` line 61 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Nialz7CCcNxku0WM1Ch:-NikyrkHEhHioH67ETR7:b-p08ijy) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java#L61)):*

Not sure why would we do that on all GET requests. See Angular XSRF doc - https://angular.io/guide/http-security-xsrf-protection.
>> To take advantage of this, your server needs to set a token in a JavaScript readable session cookie called `XSRF-TOKEN` on either the page load or the first GET request.
</blockquote></details>

Yeah I could have phrased it better, I think, my bad. 

We always want to just set the cookie (only) on the first GET request right? I'm a bit concerned that we're only doing this in the UserDataAction. What if we did "add an XSRF cookie if and only if it doesn't already exist" on every single GET request, guaranteeing that we always set it on the first GET request no matter where it is? w
___
*[`core/src/test/java/google/registry/ui/server/ActionMembershipTest.java` line 38 at r3](https://reviewable.io/reviews/google/nomulus/2211#-Nikzouf7dxiyXE61k-M:-Nikzouf7dxiyXE61k-N:b9drs36) ([raw file](https://github.com/google/nomulus/blob/c42b59a6acadcabfda0ceceaff75a3906a7ee50e/core/src/test/java/google/registry/ui/server/ActionMembershipTest.java#L38)):*
> ```Java
>     // 2. Extending JsonAction to show that we are serving JSON POST requests
>     // 3. Extending JsonGetAction to serve JSON GET requests
>     // 3. Extending ConsoleApiAction to serve JSON requests
> ```

sorry major nit, should be 4 (and maybe specifying for the new console but eh fine either way)

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2211)* status: 9 of 11 files reviewed, 5 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/request/ResponseImpl.java` line 65 at r1](https://reviewable.io/reviews/google/nomulus/2211#gh-1383856575:-NipsITFDzuaUju-mqWk:b-sko33n) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/request/ResponseImpl.java#L65)):*

this seems like the wrong place for this note, because one sets the "secure" status on the cookie itself, not here.....
</blockquote></details>

Done, secured at the cookie creation
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 28 at r3](https://reviewable.io/reviews/google/nomulus/2211#-Nil-2Yp-1Jn0eD8zWd6:-NipsO9j-VWkP48FQioa:b-wo63cu) ([raw file](https://github.com/google/nomulus/blob/c42b59a6acadcabfda0ceceaff75a3906a7ee50e/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L28)):*

I don't think we'd ever want to instantiate this directly, right? If not we can make it an abstract class
</blockquote></details>

Good suggestion
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java` line 61 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Nialz7CCcNxku0WM1Ch:-NipsQ9d3fKeEqTZD2Q1:b-f74oqb) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java#L61)):*

Yeah I could have phrased it better, I think, my bad. 

We always want to just set the cookie (only) on the first GET request right? I'm a bit concerned that we're only doing this in the UserDataAction. What if we did "add an XSRF cookie if and only if it doesn't already exist" on every single GET request, guaranteeing that we always set it on the first GET request no matter where it is? w
</blockquote></details>

Nah, then we'd need to start carrying about cookie expiration time to make sure that the cookie that we see is not old or expired and that is a completely new level of complexity we don't want. 
___
*[`core/src/test/java/google/registry/ui/server/ActionMembershipTest.java` line 38 at r3](https://reviewable.io/reviews/google/nomulus/2211#-Nikzouf7dxiyXE61k-M:-NipsoeM2IWYlZ5uxxCr:b-896fix) ([raw file](https://github.com/google/nomulus/blob/c42b59a6acadcabfda0ceceaff75a3906a7ee50e/core/src/test/java/google/registry/ui/server/ActionMembershipTest.java#L38)):*

sorry major nit, should be 4 (and maybe specifying for the new console but eh fine either way)
</blockquote></details>

Done.

gbrodman (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2211)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java` line 61 at r1](https://reviewable.io/reviews/google/nomulus/2211#-Nialz7CCcNxku0WM1Ch:-NipyQO06eBOtocsC0Ri:bfq903) ([raw file](https://github.com/google/nomulus/blob/458446b0c6f40711deb9b09cc5ad227be145028e/core/src/main/java/google/registry/ui/server/console/ConsoleUserDataAction.java#L61)):*

Nah, then we'd need to start carrying about cookie expiration time to make sure that the cookie that we see is not old or expired and that is a completely new level of complexity we don't want.
</blockquote></details>

sounds good!

PR #2214, opened 2023-11-08
Add Bsa Download Action

weiminyu:
Add a skeletion download action.

Also added a downloader (BlockListFetcher) and a GcsClient.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2214)

### DIFF
--- a/core/src/main/java/google/registry/bsa/BlockListFetcher.java
+++ b/core/src/main/java/google/registry/bsa/BlockListFetcher.java
@@ -1,0 +1,128 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.bsa;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
+
+import com.google.api.client.http.HttpMethods;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.io.ByteStreams;
+import google.registry.bsa.api.BsaCredential;
+import google.registry.bsa.api.BsaException;
+import google.registry.config.RegistryConfig.Config;
+import google.registry.request.UrlConnectionService;
+import google.registry.util.Retrier;
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.security.GeneralSecurityException;
+import java.util.function.BiConsumer;
+import javax.inject.Inject;
+import javax.net.ssl.HttpsURLConnection;
+
+/** Fetches data from the BSA API. */
+public class BlockListFetcher {
+
+  private final UrlConnectionService urlConnectionService;
+  private final BsaCredential credential;
+
+  private final ImmutableMap<String, String> blockListUrls;
+  private final Retrier retrier;
+
+  @Inject
+  BlockListFetcher(
+      UrlConnectionService urlConnectionService,
+      BsaCredential credential,
+      @Config("bsaDataUrls") ImmutableMap<String, String> blockListUrls,
+      Retrier retrier) {
+    this.urlConnectionService = urlConnectionService;
+    this.credential = credential;
+    this.blockListUrls = blockListUrls;
+    this.retrier = retrier;
+  }
+
+  LazyBlockList fetch(BlockList blockList) {
+    // TODO: use more informative exceptions to describe retriable errors
+    return retrier.callWithRetry(
+        () -> tryFetch(blockList),
+        e -> e instanceof BsaException && ((BsaException) e).isRetriable());
+  }
+
+  LazyBlockList tryFetch(BlockList blockList) {
+    try {
+      HttpsURLConnection connection =
+          (HttpsURLConnection)
+              urlConnectionService.createConnection(
+                  new java.net.URL(blockListUrls.get(blockList.name())));
+      connection.setRequestMethod(HttpMethods.GET);
+      connection.setRequestProperty("Authorization", "Bearer " + credential.getAuthToken());
+      int code = connection.getResponseCode();
+      if (code != SC_OK) {
+        String errorDetails = "";
+        try (InputStream errorStream = connection.getErrorStream()) {
+          errorDetails = new String(ByteStreams.toByteArray(errorStream), UTF_8);
+        } catch (Exception e) {
+          // ignore
+        }
+        throw new BsaException(
+            String.format(
+                "Status code: [%s], error: [%s], details: [%s]",
+                code, connection.getResponseMessage(), errorDetails),
+            /* retriable= */ true);
+      }
+      return new LazyBlockList(blockList, connection);
+    } catch (IOException e) {
+      throw new BsaException(e, /* retriable= */ true);
+    } catch (GeneralSecurityException e) {
+      throw new BsaException(e, /* retriable= */ false);
+    }
+  }
+
+  static class LazyBlockList implements Closeable {
+
+    private final BlockList blockList;
+
+    private final HttpsURLConnection connection;
+
+    LazyBlockList(BlockList blockList, HttpsURLConnection connection) {
+      this.blockList = blockList;
+      this.connection = connection;
+    }
+
+    BlockList getName() {
+      return blockList;
+    }
+
+    String peekChecksum() {
+      return "TODO"; // Depends on BSA impl: header or first line of file
+    }
+
+    void consumeAll(BiConsumer<byte[], Integer> consumer) throws IOException {
+      try (InputStream inputStream = connection.getInputStream()) {
+        byte[] buffer = new byte[1024];
+        int bytesRead;
+        while ((bytesRead = inputStream.read(buffer)) != -1) {
+          consumer.accept(buffer, bytesRead);
+        }
+      }
+    }
+
+    @Override
+    public void close() {
+      connection.disconnect();
+    }
+  }
+}
--- a/core/src/main/java/google/registry/bsa/BsaDownloadAction.java
+++ b/core/src/main/java/google/registry/bsa/BsaDownloadAction.java
@@ -32,7 +32,7 @@ import java.util.Optional;
 import javax.inject.Inject;
 
 @Action(
-    service = Action.Service.BACKEND,
+    service = Action.Service.BSA,
     path = BsaDownloadAction.PATH,
     method = POST,
     auth = Auth.AUTH_API_ADMIN)
--- a/core/src/main/java/google/registry/bsa/BsaDownloadAction.java
+++ b/core/src/main/java/google/registry/bsa/BsaDownloadAction.java
@@ -1,0 +1,129 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.bsa;
+
+import static google.registry.bsa.BlockList.BLOCK;
+import static google.registry.bsa.BlockList.BLOCK_PLUS;
+import static google.registry.request.Action.Method.POST;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.flogger.FluentLogger;
+import google.registry.bsa.BlockListFetcher.LazyBlockList;
+import google.registry.bsa.persistence.DownloadSchedule;
+import google.registry.bsa.persistence.DownloadScheduler;
+import google.registry.request.Action;
+import google.registry.request.Response;
+import google.registry.request.auth.Auth;
+import java.util.Optional;
+import javax.inject.Inject;
+
+@Action(
+    service = Action.Service.BACKEND,
+    path = BsaDownloadAction.PATH,
+    method = POST,
+    auth = Auth.AUTH_API_ADMIN)
+public class BsaDownloadAction implements Runnable {
+
+  private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+  static final String PATH = "/_dr/task/bsaDownload";
+
+  private final DownloadScheduler downloadScheduler;
+  private final BlockListFetcher blockListFetcher;
+  private final GcsClient gcsClient;
+  private final BsaLock bsaLock;
+  private final Response response;
+
+  @Inject
+  BsaDownloadAction(
+      DownloadScheduler downloadScheduler,
+      BlockListFetcher blockListFetcher,
+      GcsClient gcsClient,
+      BsaLock bsaLock,
+      Response response) {
+    this.downloadScheduler = downloadScheduler;
+    this.blockListFetcher = blockListFetcher;
+    this.gcsClient = gcsClient;
+    this.bsaLock = bsaLock;
+    this.response = response;
+  }
+
+  @Override
+  public void run() {
+    try {
+      if (!bsaLock.executeWithLock(this::runWithinLock)) {
+        logger.atInfo().log("Job is being executed by another worker.");
+      }
+    } catch (Throwable throwable) {
+      logger.atWarning().withCause(throwable).log("Failed to update block lists.");
+    }
+    // Always return OK. Let the next cron job retry.
+    response.setStatus(SC_OK);
+  }
+
+  Void runWithinLock() {
+    Optional<DownloadSchedule> scheduleOptional = downloadScheduler.schedule();
+    if (!scheduleOptional.isPresent()) {
+      logger.atInfo().log("Nothing to do.");
+      return null;
+    }
+    DownloadSchedule schedule = scheduleOptional.get();
+    switch (schedule.stage()) {
+      case DOWNLOAD:
+        try (LazyBlockList block = blockListFetcher.fetch(BLOCK);
+            LazyBlockList blockPlus = blockListFetcher.fetch(BLOCK_PLUS)) {
+          ImmutableMap<BlockList, String> fetchedChecksums =
+              ImmutableMap.of(BLOCK, block.peekChecksum(), BLOCK_PLUS, blockPlus.peekChecksum());
+          ImmutableMap<BlockList, String> prevChecksums =
+              schedule
+                  .latestCompleted()
+                  .map(DownloadSchedule.CompletedJob::checksums)
+                  .orElseGet(ImmutableMap::of);
+          // TODO(11/30/2023): uncomment below block when BSA checksums are available.
+          // if (!schedule.forceDownload() && Objects.equals(fetchedChecksums, prevChecksums)) {
+          //   schedule.updateJobStage(DownloadStage.NOP, fetchedChecksums);
+          //   return null;
+          // }
+          ImmutableMap<BlockList, String> actualChecksum =
+              gcsClient.saveAndChecksumBlockList(
+                  schedule.jobName(), ImmutableList.of(block, blockPlus));
+          // TODO(11/30/2023): uncomment below block when BSA checksums are available.
+          // if (!Objects.equals(fetchedChecksums, actualChecksum)) {
+          //   logger.atSevere().log(
+          //       "Mismatching checksums: BSA's is [%s], ours is [%s]",
+          //       fetchedChecksums, actualChecksum);
+          //   schedule.updateJobStage(DownloadStage.CHECKSUMS_NOT_MATCH);
+          //   return null;
+          // }
+          schedule.updateJobStage(DownloadStage.MAKE_DIFF, actualChecksum);
+        }
+        // Fall through
+      case MAKE_DIFF:
+        // TODO(11/30/2023): fill out the rest stages.
+      case APPLY_DIFF:
+      case START_UPLOADING:
+      case UPLOAD_DOMAINS_IN_USE:
+      case FINISH_UPLOADING:
+      case DONE:
+      case NOP:
+      case CHECKSUMS_NOT_MATCH:
+        logger.atWarning().log("Unexpectedly reached the %s stage.", schedule.stage());
+        break;
+    }
+    return null;
+  }
+}
--- a/core/src/main/java/google/registry/bsa/BsaLock.java
+++ b/core/src/main/java/google/registry/bsa/BsaLock.java
@@ -1,0 +1,40 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.bsa;
+
+import google.registry.config.RegistryConfig.Config;
+import google.registry.request.lock.LockHandler;
+import java.util.concurrent.Callable;
+import javax.inject.Inject;
+import org.joda.time.Duration;
+
+/** Helper for guarding all BSA related work with a common lock. */
+public class BsaLock {
+
+  private static final String LOCK_NAME = "all-bsa-jobs";
+
+  private final LockHandler lockHandler;
+  private final Duration leaseExpiry;
+
+  @Inject
+  BsaLock(LockHandler lockHandler, @Config("bsaLockLeaseExpiry") Duration leaseExpiry) {
+    this.lockHandler = lockHandler;
+    this.leaseExpiry = leaseExpiry;
+  }
+
+  boolean executeWithLock(Callable<Void> callable) {
+    return lockHandler.executeWithLocks(callable, null, leaseExpiry, LOCK_NAME);
+  }
+}
--- a/core/src/main/java/google/registry/bsa/GcsClient.java
+++ b/core/src/main/java/google/registry/bsa/GcsClient.java
@@ -1,0 +1,167 @@
+// Copyright 2023 The Nomulus Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package google.registry.bsa;
+
+import static com.google.common.io.BaseEncoding.base16;
+
+import com.google.cloud.storage.BlobId;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import google.registry.bsa.BlockListFetcher.LazyBlockList;
+import google.registry.bsa.api.Label;
+import google.registry.bsa.api.NonBlockedDomain;
+import google.registry.bsa.api.Order;
+import google.registry.config.RegistryConfig.Config;
+import google.registry.gcs.GcsUtils;
+import java.io.BufferedOutputStream;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.stream.Stream;
+import javax.inject.Inject;
+
+/** Stores and accesses BSA-related data, including original downloads and processed data. */
+public class GcsClient {
+
+  static final String LABELS_DIFF_FILE = "labels_diff.csv";
+  public static final String DOMAINS_IN_USE_FILE = "domains_in_use.csv";
+  public static final String ORDERS_DIFF_FILE = "orders_diff.csv";
+  private final GcsUtils gcsUtils;
+  private final String bucketName;
+
+  private final String checksumAlgorithm;
+
+  @Inject
+  GcsClient(
+      GcsUtils gcsUtils,
+      @Config("bsaGcsBucket") String bucketName,
+      @Config("bsaChecksumAlgorithm") String checksumAlgorithm) {
+    this.gcsUtils = gcsUtils;
+    this.bucketName = bucketName;
+    this.checksumAlgorithm = checksumAlgorithm;
+  }
+
+  static String getBlockListFileName(BlockList blockList) {
+    return blockList.name() + ".csv";
+  }
+
+  ImmutableMap<BlockList, String> saveAndChecksumBlockList(
+      String jobName, ImmutableList<LazyBlockList> blockLists) {
+    // Downloading sequentially, since one is expected to be much smaller than the other.
+    return blockLists.stream()
+        .collect(
+            ImmutableMap.toImmutableMap(
+                LazyBlockList::getName, blockList -> saveAndChecksumBlockList(jobName, blockList)));
+  }
+
+  private String saveAndChecksumBlockList(String jobName, LazyBlockList blockList) {
+    BlobId blobId = getBlobId(jobName, getBlockListFileName(blockList.getName()));
+    try (BufferedOutputStream gcsWriter =
+        new BufferedOutputStream(gcsUtils.openOutputStream(blobId))) {
+      MessageDigest messageDigest = MessageDigest.getInstance(checksumAlgorithm);
+      blockList.consumeAll(
+          (byteArray, length) -> {
+            try {
+              gcsWriter.write(byteArray, 0, length);
+            } catch (IOException e) {
+              throw new RuntimeException(e);
+            }
+            messageDigest.update(byteArray, 0, length);
+          });
+      return base16().lowerCase().encode(messageDigest.digest());
+    } catch (IOException | NoSuchAlgorithmException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static void writeWithNewline(BufferedWriter writer, String line) {
+    try {
+      writer.write(line);
+      if (!line.endsWith("\n")) {
+        writer.write('\n');
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  Stream<String> readBlockList(String jobName, BlockList blockList) {
+    return readStream(getBlobId(jobName, getBlockListFileName(blockList)));
+  }
+
+  Stream<Order> readOrderDiffs(String jobName) {
+    BlobId blobId = getBlobId(jobName, ORDERS_DIFF_FILE);
+    return readStream(blobId).map(Order::deserialize);
+  }
+
+  void writeOrderDiffs(String jobName, Stream<Order> orders) {
+    BlobId blobId = getBlobId(jobName, ORDERS_DIFF_FILE);
+    try (BufferedWriter gcsWriter = getWriter(blobId)) {
+      orders.map(Order::serialize).forEach(line -> writeWithNewline(gcsWriter, line));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  Stream<Label> readLabelDiffs(String jobName) {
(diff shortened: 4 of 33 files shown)


ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 7 unresolved discussions (waiting on @github-advanced-security[bot] and @weiminyu)

___
*[`core/src/main/java/google/registry/module/backend/BackendRequestComponent.java` line 105 at r3](https://reviewable.io/reviews/google/nomulus/2214#-Nipgv_9DfKDt3dyHjCt:-Nipgv_9DfKDt3dyHjCu:bistxml) ([raw file](https://github.com/google/nomulus/blob/cca9663dd0034e5389c5b09371053a95d2f947c6/core/src/main/java/google/registry/module/backend/BackendRequestComponent.java#L105)):*
> ```Java
> interface BackendRequestComponent {
> 
>   BsaDownloadAction bsaDownloadAction();
> ```

We probably want to make it a separate service altogether. This way static IP change is more safe, because as it turned out we have multiple upload actions that require whitelisting with ICANN. Given the infrequent nature of those actions and possibility of messing them up we would be better of limiting the scope of change to a separate service specifically to BSA

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 7 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/module/backend/BackendRequestComponent.java` line 105 at r3](https://reviewable.io/reviews/google/nomulus/2214#-Nipgv_9DfKDt3dyHjCt:-Niprmxu-kr9aUimqwSX:bqrk7il) ([raw file](https://github.com/google/nomulus/blob/cca9663dd0034e5389c5b09371053a95d2f947c6/core/src/main/java/google/registry/module/backend/BackendRequestComponent.java#L105)):*

We probably want to make it a separate service altogether. This way static IP change is more safe, because as it turned out we have multiple upload actions that require whitelisting with ICANN. Given the infrequent nature of those actions and possibility of messing them up we would be better of limiting the scope of change to a separate service specifically to BSA
</blockquote></details>

Makes sense. Could you set it up?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 7 unresolved discussions (waiting on @github-advanced-security[bot] and @weiminyu)

___
*[`core/src/main/java/google/registry/module/backend/BackendRequestComponent.java` line 105 at r3](https://reviewable.io/reviews/google/nomulus/2214#-Nipgv_9DfKDt3dyHjCt:-NiptgRK9FbZ0fZj1xZK:bip2z71) ([raw file](https://github.com/google/nomulus/blob/cca9663dd0034e5389c5b09371053a95d2f947c6/core/src/main/java/google/registry/module/backend/BackendRequestComponent.java#L105)):*

Makes sense. Could you set it up?
</blockquote></details>

Yeah, I can do it

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 10 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92OHADMCsS8v3MqCW:-Nj92OHADMCsS8v3MqCX:beqxmzn) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L88)):*
> ```
>     );
> 
>     create table "BsaDomainInUse" (
> ```

Add `domain_repo_id` as a pointer so that these can easily be checked in the future to see which might no longer be in use if the domain gets deleted?
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 101 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92GrV3Bl_MEpsbNd0:-Nj92GrV3Bl_MEpsbNd1:bu3iov4) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L101)):*
> ```
>         creation_time timestamptz not null,
>         stage text not null,
>         update_timestamp timestamptz,
> ```

Why is there a creation time and an update time?  Aren't these immutable?  I would expect there to just be a download time (could also be called creation time).
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 107 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92Y6g6ksQ2O9N_A5K:-Nj92Y6g6ksQ2O9N_A5L:b-6t3je4) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L107)):*
> ```
>     create table "BsaLabel" (
>        label text not null,
>         creation_time timestamptz not null,
> ```

Do we want a pointer to the first `BsaDownload` the label occurred in?  Or would you do that by matching up the creation times?

I think it might be better to have an revision/download ID here, and then join it to the `BsaDownload` table if you need to know the creation time.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 11 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java` line 29 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9DeAY6qjRqUize4Hd:-Nj9DeAY6qjRqUize4He:bcrt2v5) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java#L29)):*
> ```Java
> import javax.persistence.IdClass;
> 
> /** A domain matching a BSA label but is in use (registered or reserved), so cannot be blocked. */
> ```

I wouldn't necessarily call a reserved domain an "domain in use".  `BsaUnblockableDomain` with a reason or either registered or reserved makes more sense to me semantically.

Also what about domains that are unblockable by reason of the IDN table not supporting them? We don't want to track those here, and will instead depend on regenerating the entire set (with notifications) on the (very rare) chance of the IDN table on the TLD being changed?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 13 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/BlockList.java` line 18 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9EFMXB0WKR68crvL5:-Nj9EFMXB0WKR68crvL6:b8p0ynb) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/BlockList.java#L18)):*
> ```Java
> 
> /** Identifiers of the BSA lists with blocking labels. */
> public enum BlockList {
> ```

You can put these small enums as static inner members of the classes in which they're used rather than having them be their own files, like `NonBlockedDomain.Reason` below.

Also I think the name here needs to be workshopped.  Maybe `BlockListType`, `BlockListName` or similar?
___
*[`core/src/main/java/google/registry/bsa/api/NonBlockedDomain.java` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9E6cRDITpU-L-TWtZ:-Nj9E6cRDITpU-L-TWt_:b-lgec88) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/api/NonBlockedDomain.java#L27)):*
> ```Java
>  */
> @AutoValue
> public abstract class NonBlockedDomain {
> ```

`UnblockableDomain` makes more sense to me here as a name.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 13 unresolved discussions (waiting on @CydeWeys, @github-advanced-security[bot], and @ptkach)

___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92OHADMCsS8v3MqCW:-NjAAcaG3ob0Mm7LjXod:b-kfq9z3) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L88)):*

Add `domain_repo_id` as a pointer so that these can easily be checked in the future to see which might no longer be in use if the domain gets deleted?
</blockquote></details>

Domain_name is the foreign key so checking should be efficient too. Also, this may be a reserved name that is not registered yet.
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 101 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92GrV3Bl_MEpsbNd0:-NjA6nRq2Aws78KYxp9G:b1wd295) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L101)):*

Why is there a creation time and an update time?  Aren't these immutable?  I would expect there to just be a download time (could also be called creation time).
</blockquote></details>

A download progresses through multiple stages. If it somehow fails in the middle and retries, the update_time will be different from the
creation time.
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 107 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92Y6g6ksQ2O9N_A5K:-NjA7POeAXCr50rzIKFN:b-8dwvg) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L107)):*

Do we want a pointer to the first `BsaDownload` the label occurred in?  Or would you do that by matching up the creation times?

I think it might be better to have an revision/download ID here, and then join it to the `BsaDownload` table if you need to know the creation time.
</blockquote></details>

The creation_time should have been called first_download_time. It's the same timestamp of the first BsaDownload it appears in.
It's a bit too late to change the name if we don't want to wait another week.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 35 files reviewed, 13 unresolved discussions (waiting on @CydeWeys, @github-advanced-security[bot], and @ptkach)

___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 107 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92Y6g6ksQ2O9N_A5K:-NjACqMxEsrGC3gPQhLa:b-17cwg7) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L107)):*

The creation_time should have been called first_download_time. It's the same timestamp of the first BsaDownload it appears in.
It's a bit too late to change the name if we don't want to wait another week.
</blockquote></details>

By the way, this file is carried over from go/r3pr/2205, which is approved and will be submitted next.
Please make comments there if you want to discuss more about the schema.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 12 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92OHADMCsS8v3MqCW:-Njn8Tqo-I0HJuF8l1bu:btdk1ss) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L88)):*

Domain_name is the foreign key so checking should be efficient too. Also, this may be a reserved name that is not registered yet.
</blockquote></details>

But you don't have `domain_name` in here either -- it would need to be constructed through string concatenation.  And the primary key is better than the foreign key, as `domain_name` on the `Domain` table is not unique (whereas repo id is).  If a domain has existed several times, then it will have several entries in the `Domain` table, and now you have order them by create time and take the most recent one.  It's just a lot harder than `JOIN` to the `Domain` table on `domain_repo_id`.

And for the case of a reservation, the `domain_repo_id` field can be nullable.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @github-advanced-security[bot], and @ptkach)

___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92OHADMCsS8v3MqCW:-Njs4g9a2vE_K3HbLzZb:b-2etmqe) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L88)):*

But you don't have `domain_name` in here either -- it would need to be constructed through string concatenation.  And the primary key is better than the foreign key, as `domain_name` on the `Domain` table is not unique (whereas repo id is).  If a domain has existed several times, then it will have several entries in the `Domain` table, and now you have order them by create time and take the most recent one.  It's just a lot harder than `JOIN` to the `Domain` table on `domain_repo_id`.

And for the case of a reservation, the `domain_repo_id` field can be nullable.
</blockquote></details>

`ForeignKeyUtils.load` already  does the filtering on timestamps, and it is used when we create BsaDomainInUse entities from labels.
We can reuse the same code when doing the refresh.  `ForeignKeyUtils.load` also already supports bulk query, which is also needed by the refresh check.
___
*[`core/src/main/java/google/registry/bsa/BlockList.java` line 18 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9EFMXB0WKR68crvL5:-Njs6sR13aFSslE6sFmO:b-2xd66h) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/BlockList.java#L18)):*

You can put these small enums as static inner members of the classes in which they're used rather than having them be their own files, like `NonBlockedDomain.Reason` below.

Also I think the name here needs to be workshopped.  Maybe `BlockListType`, `BlockListName` or similar?
</blockquote></details>

These enums are used in multiple classes, and there is not an obvious choice of enclosing class for each of them.
Unless you want a `Definitions` interface or class to hold all of them.
___
*[`core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java` line 29 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9DeAY6qjRqUize4Hd:-Njs7SIW2F7jiUPuO9Dp:bzdzgqc) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java#L29)):*

I wouldn't necessarily call a reserved domain an "domain in use".  `BsaUnblockableDomain` with a reason or either registered or reserved makes more sense to me semantically.

Also what about domains that are unblockable by reason of the IDN table not supporting them? We don't want to track those here, and will instead depend on regenerating the entire set (with notifications) on the (very rare) chance of the IDN table on the TLD being changed?
</blockquote></details>

I just want a name that is distinguishable from the one used for communicating with BSA. We can change the name of this class, but changing the name of the underlying
schema table would be time-consuming.

And yes, invalid (due to idn) domains are not tracked here. See the Reason enum below: it doesn't have `INVALID`, which is present in the API object.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 12 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92OHADMCsS8v3MqCW:-NjsdcB17HPkz1WeXCht:b-qwoacr) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L88)):*

`ForeignKeyUtils.load` already  does the filtering on timestamps, and it is used when we create BsaDomainInUse entities from labels.
We can reuse the same code when doing the refresh.  `ForeignKeyUtils.load` also already supports bulk query, which is also needed by the refresh check.
</blockquote></details>

I was thinking it might be useful for manual queries or dashboards to do analytics on the list (to join to the exact `Domain` row in question easily).  You don't have access to ``ForeignKeyUtils.load()` from a SQL query sadly; it gets much more complicated (need to concatenate domain + label, join to the Domain table based on that, sort by create time, pick the most recent, etc.).

But maybe we wouldn't use this table for that, and just use the `BsaLabel` table?  The only reason we really need this table is to keep track of what we already sent to the BSA API, right?

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @github-advanced-security[bot], and @ptkach)

___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 90 at r1](https://reviewable.io/reviews/google/nomulus/2214#gh-1387227428:-NkMT3ePB6JBobMkIndI:bi01p9k) ([raw file](https://github.com/google/nomulus/blob/16478ab9776ec5acd6a58829a6476dfcef81cf86/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L90)):*

## Unread local variable

Variable 'ImmutableMap<BlockList,String> fetchedChecksums' is never read.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1603)
</blockquote></details>

Will address will BSA feature is available.
___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 95 at r1](https://reviewable.io/reviews/google/nomulus/2214#gh-1387227429:-NkMTH2vClEsCApS37Ro:bi01p9k) ([raw file](https://github.com/google/nomulus/blob/16478ab9776ec5acd6a58829a6476dfcef81cf86/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L95)):*

## Unread local variable

Variable 'ImmutableMap<BlockList,String> prevChecksums' is never read.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1604)
</blockquote></details>

Will address will BSA feature is available.
___
*[`core/src/main/java/google/registry/bsa/api/NonBlockedDomain.java` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9E6cRDITpU-L-TWtZ:-NkMTJPr6rJ-5sefbRHL:b-iq524s) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/api/NonBlockedDomain.java#L27)):*

`UnblockableDomain` makes more sense to me here as a name.
</blockquote></details>

Will address in a future PR.
___
*[`core/src/main/java/google/registry/bsa/api/Order.java` line 42 at r1](https://reviewable.io/reviews/google/nomulus/2214#gh-1387227420:-NkMThWgC-1TD-i9Gaj2:bx8kvxv) ([raw file](https://github.com/google/nomulus/blob/16478ab9776ec5acd6a58829a6476dfcef81cf86/core/src/main/java/google/registry/bsa/api/Order.java#L42)):*

## Missing catch of NumberFormatException

Potential uncaught 'java.lang.NumberFormatException'.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1599)
</blockquote></details>

Addressed in a follow-up PR
___
*[`core/src/test/java/google/registry/bsa/api/LabelTest.java` line 25 at r1](https://reviewable.io/reviews/google/nomulus/2214#gh-1387227421:-NkMTlDm5Dfz6MSwEStq:bq1tew8) ([raw file](https://github.com/google/nomulus/blob/16478ab9776ec5acd6a58829a6476dfcef81cf86/core/src/test/java/google/registry/bsa/api/LabelTest.java#L25)):*

Unused class: LabelTest is not referenced within this codebase. If not used as an external API it should be removed.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1600)
</blockquote></details>

Github check gone wrong
. This is test.
___
*[`core/src/test/java/google/registry/bsa/api/NonBlockedDomainTest.java` line 24 at r1](https://reviewable.io/reviews/google/nomulus/2214#gh-1387227424:-NkMTpgXCn2KXSFH-TPS:bg7fv0a) ([raw file](https://github.com/google/nomulus/blob/16478ab9776ec5acd6a58829a6476dfcef81cf86/core/src/test/java/google/registry/bsa/api/NonBlockedDomainTest.java#L24)):*

Unused class: NonBlockedDomainTest is not referenced within this codebase. If not used as an external API it should be removed.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1601)
</blockquote></details>

Github check gone wrong. This is test.
___
*[`core/src/test/java/google/registry/bsa/api/OrderTest.java` line 24 at r1](https://reviewable.io/reviews/google/nomulus/2214#gh-1387227425:-NkMTzg_89Nnurht0WOI:bg7fv0a) ([raw file](https://github.com/google/nomulus/blob/16478ab9776ec5acd6a58829a6476dfcef81cf86/core/src/test/java/google/registry/bsa/api/OrderTest.java#L24)):*

Unused class: OrderTest is not referenced within this codebase. If not used as an external API it should be removed.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1602)
</blockquote></details>

Github check gone wrong. This is test.
___
*[`db/src/main/resources/sql/schema/db-schema.sql.generated` line 88 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj92OHADMCsS8v3MqCW:-NkMUJ2u42MDw5007doW:b-zayvw6) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/db/src/main/resources/sql/schema/db-schema.sql.generated#L88)):*

I was thinking it might be useful for manual queries or dashboards to do analytics on the list (to join to the exact `Domain` row in question easily).  You don't have access to ``ForeignKeyUtils.load()` from a SQL query sadly; it gets much more complicated (need to concatenate domain + label, join to the Domain table based on that, sort by create time, pick the most recent, etc.).

But maybe we wouldn't use this table for that, and just use the `BsaLabel` table?  The only reason we really need this table is to keep track of what we already sent to the BSA API, right?
</blockquote></details>

Yes, this table is just to track what we sent to BSA.

If we want to do a present-time analysis we can just join the "Domain" table with the "BsaLabel" table, with a `deletion_time > now()` filter.
___
*[`core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java` line 29 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9DeAY6qjRqUize4Hd:-NkMU7rO2vKKcRLtoxVA:b-g3d1t) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java#L29)):*

I just want a name that is distinguishable from the one used for communicating with BSA. We can change the name of this class, but changing the name of the underlying
schema table would be time-consuming.

And yes, invalid (due to idn) domains are not tracked here. See the Reason enum below: it doesn't have `INVALID`, which is present in the API object.
</blockquote></details>

Started the renaming process. New table checked in.
___
*[`core/src/main/java/google/registry/module/backend/BackendRequestComponent.java` line 105 at r3](https://reviewable.io/reviews/google/nomulus/2214#-Nipgv_9DfKDt3dyHjCt:-NkMUERm0zwwRAIe4F5W:b-886gsx) ([raw file](https://github.com/google/nomulus/blob/cca9663dd0034e5389c5b09371053a95d2f947c6/core/src/main/java/google/registry/module/backend/BackendRequestComponent.java#L105)):*

Yeah, I can do it
</blockquote></details>

Fixed

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 12 unresolved discussions (waiting on @CydeWeys, @github-advanced-security[bot], and @ptkach)

___
*[`core/src/main/java/google/registry/bsa/BlockList.java` line 18 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9EFMXB0WKR68crvL5:-NkOOEtF4XbpI9jPJCD9:b-cupdzs) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/BlockList.java#L18)):*

These enums are used in multiple classes, and there is not an obvious choice of enclosing class for each of them.
Unless you want a `Definitions` interface or class to hold all of them.
</blockquote></details>

We can have a refactor PR for naming etc when all is done.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 21 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/BlockListFetcher.java` line 78 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRLRgaEVDc3vxjHFCN:-NkRLRgaEVDc3vxjHFCO:b-okofxs) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BlockListFetcher.java#L78)):*
> ```Java
>           errorDetails = new String(ByteStreams.toByteArray(errorStream), UTF_8);
>         } catch (Exception e) {
>           // ignore
> ```

Set `errorDetails` in here to `"Couldn't get error details from connection"` (or whatever else is correctly descriptive), just so that the exception message below has the full info.

And there's really nothing of value on `e` here ever?  Not its message or anything?
___
*[`core/src/main/java/google/registry/bsa/BlockListFetcher.java` line 94 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRLjozE5tDXC6nUrgl:-NkRLjozE5tDXC6nUrgm:b-megzs6) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BlockListFetcher.java#L94)):*
> ```Java
>   }
> 
>   static class LazyBlockList implements Closeable {
> ```

I guess `Lazy<BlockList>` wasn't good enough?
___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 69 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRLt-R8ENjiqh48mFK:-NkRLt-R8ENjiqh48mFL:b-k8objp) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L69)):*
> ```Java
>     try {
>       if (!bsaLock.executeWithLock(this::runWithinLock)) {
>         logger.atInfo().log("Job is being executed by another worker.");
> ```

Error message should say the action being taken.  E.g.: `"Job is being executed by another worker; terminating."`
___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 74 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRM3562CC-854yIc_2:-NkRM3562CC-854yIc_3:b72k4gm) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L74)):*
> ```Java
>       logger.atWarning().withCause(throwable).log("Failed to update block lists.");
>     }
>     // Always return OK. Let the next cron job retry.
> ```

How often is the cron entry going to run this?  Might we want to retry a few times in the case of an error rather than waiting for the next at least, what, hour?
___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 81 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRMQoC4V_km36_N35o:-NkRMQoC4V_km36_N35p:btnissq) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L81)):*
> ```Java
>     Optional<DownloadSchedule> scheduleOptional = downloadScheduler.schedule();
>     if (!scheduleOptional.isPresent()) {
>       logger.atInfo().log("Nothing to do.");
> ```

Log message needs more clarity.  Why might there be nothing to do and what is happening as a result?  Is this a configuration thing, or what causes this case?
___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 85 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRMdOI7IQGZy1X_8LW:-NkRMdOI7IQGZy1X_8LX:bi4h37w) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L85)):*
> ```Java
>     }
>     DownloadSchedule schedule = scheduleOptional.get();
>     switch (schedule.stage()) {
> ```

So this, what, keeps running over and over again, ratcheting through the stages and updating the progress each time?  And this supports resuming from the previous stage it left off at I assume?
___
*[`core/src/main/java/google/registry/bsa/BsaLock.java` line 23 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRMoCr-GL1yfvz34Xi:-NkRMoCr-GL1yfvz34Xj:b-467bom) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BsaLock.java#L23)):*
> ```Java
> import org.joda.time.Duration;
> 
> /** Helper for guarding all BSA related work with a common lock. */
> ```

So this is a global singleton lock?  If so, mention that here.
___
*[`core/src/main/java/google/registry/bsa/api/BsaCredential.java` line 43 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRNJr4B-6Gsub4sgbD:-NkRNJr4B-6Gsub4sgbE:b8hd10f) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/api/BsaCredential.java#L43)):*
> ```Java
> import org.joda.time.Instant;
> 
> /** Self-refreshing credential for accessing the BSA API. */
> ```

Add more documentation here about what the credential consists of and how/why it self-refreshes (e.g. because it expires, and also possibly other reasons?).
___
*[`core/src/main/java/google/registry/bsa/api/BsaException.java` line 17 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRNd7u9J8oA1jmd-vw:-NkRNd7u9J8oA1jmd-vx:b-jo2119) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/api/BsaException.java#L17)):*
> ```Java
> package google.registry.bsa.api;
> 
> public class BsaException extends RuntimeException {
> ```

Why are they all `RuntimeException`s?

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2214)* status: 0 of 44 files reviewed, 26 unresolved discussions (waiting on @github-advanced-security[bot], @ptkach, and @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/BsaDownloadAction.java` line 87 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRjRz9D9bxHTlTOg28:-NkRjRz9D9bxHTlTOg29:b-lfajkw) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/BsaDownloadAction.java#L87)):*
> ```Java
>     switch (schedule.stage()) {
>       case DOWNLOAD:
>         try (LazyBlockList block = blockListFetcher.fetch(BLOCK);
> ```

You'll probably want to break these out into helper methods but I assume that comes in a subsequent PR.
___
*[`core/src/main/java/google/registry/bsa/api/BsaCredential.java` line 88 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRjiTkA4jIUc1p2TxE:-NkRjiTkA4jIUc1p2TxF:b-cxivx1) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/api/BsaCredential.java#L88)):*
> ```Java
>       ensureAuthTokenValid();
>       throw new BsaException(e, /* retriable= */ true);
> ```

Maybe there's a case here for a subclass of `BsaException`, something like `RetriableBsaException` or `TransientBsaException`?  Rather than just having it as a parameter to the constructor?
___
*[`core/src/main/java/google/registry/bsa/api/BsaCredential.java` line 123 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRjxkjANBycS59bup4:-NkRjxkjANBycS59bup5:b-734z8c) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/api/BsaCredential.java#L123)):*
> ```Java
>           errorDetails = new String(ByteStreams.toByteArray(errorStream), UTF_8);
>         } catch (Exception e) {
>           // ignore
> ```

This always makes me a bit wary ... what is the case that's being ignored here?
___
*[`core/src/main/java/google/registry/bsa/api/Label.java` line 28 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRk2j84dYWjPp2TssP:-NkRk2j84dYWjPp2TssQ:bbycqby) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/api/Label.java#L28)):*
> ```Java
>  */
> @AutoValue
> public abstract class Label {
> ```

`Label` is quite generic.  Why not `BsaLabel`?  Is that because it overloads the DB table name?
___
*[`core/src/main/java/google/registry/bsa/api/NonBlockedDomain.java` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9E6cRDITpU-L-TWtZ:-NkRkF2DAW-Mf233e7J-:b32w0x) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/api/NonBlockedDomain.java#L27)):*

Will address in a future PR.
</blockquote></details>

OK
___
*[`core/src/main/java/google/registry/bsa/api/Order.java` line 27 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRkHig7wWA-DDztfyZ:-NkRkHig7wWA-DDztfy_:b2p0l75) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/api/Order.java#L27)):*
> ```Java
>  */
> @AutoValue
> public abstract class Order {
> ```

Perhaps `BsaOrder`?
___
*[`core/src/main/java/google/registry/bsa/persistence/DownloadSchedule.java` line 30 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRkOISD9tMVaU8MIJ1:-NkRkOISD9tMVaU8MIJ2:b-1iuajn) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/bsa/persistence/DownloadSchedule.java#L30)):*
> ```Java
> import java.util.Optional;
> 
> /** Information needed when handling a download from BSA. */
> ```

Can you expand this documentation?  This class is not 100% clear.
___
*[`core/src/main/java/google/registry/config/files/default-config.yaml` line 620 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRk_ae9EX3B1NeJhB7:-NkRk_ae9EX3B1NeJhB8:blyug4l) ([raw file](https://github.com/google/nomulus/blob/065fc6635d5e33c369fac868f5dfdc23704b7301/core/src/main/java/google/registry/config/files/default-config.yaml#L620)):*
> ```YAML
>   # Max time period during which downloads can be skipped because checksums have
>   # not changed from the previous one.
>   bsaMaxNopIntervalHours: 24
> ```

`Nop`?  Should be `Noop` perhaps?
___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 577 at r6](https://reviewable.io/reviews/google/nomulus/2214#-NkRkgar3qV0aKBdfw-q:-NkRkgar3qV0aKBdfw-r:b-ifxl5e) ([raw file](https://github.com/google/nomulus/blob/853e571d014011afde20d105f24d75a8907b2dee/core/src/main/java/google/registry/model/tld/Tld.java#L577)):*
> ```Java
> 
>   /** Returns the time when this TLD was enrolled in the Brand Safety Alliance (BSA) program. */
>   @JsonIgnore // Annotation can be removed once we add the field and annotate it.
> ```

So is this going to be removed or not?  I think it should be.
___
*[`core/src/main/java/google/registry/bsa/BlockList.java` line 18 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9EFMXB0WKR68crvL5:-NkRktkl1MjTHHVmcDoe:b-ojw9mm) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/BlockList.java#L18)):*

We can have a refactor PR for naming etc when all is done.
</blockquote></details>

Yeah it's fine, this is low importance obviously and trivial to refactor later on.  It's just that there's quite a lot of different files floating around and I think some things could be consolidated eventually.
___
*[`core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java` line 29 at r4](https://reviewable.io/reviews/google/nomulus/2214#-Nj9DeAY6qjRqUize4Hd:-NkRl2v4A5_PI0-vKJa7:b-9755fn) ([raw file](https://github.com/google/nomulus/blob/c918fa183aaa814e547b533981f1414a19ec81bf/core/src/main/java/google/registry/bsa/persistence/BsaDomainInUse.java#L29)):*

Started the renaming process. New table checked in.
</blockquote></details>

Thank you!

PR #2224, opened 2023-11-16
Add a dedicated IP header to accommodate Java 17 on GAE

jianglai:
For reasons unclear at this point, Java 17's servlet implementation on
GAE injects IP addresses (including unroutable private IPs) into the
standard X-Forwarded-For header, which we currently use to embed
registrar IP addresses to check against the allow list. This results in
the server not properly parsing the header and rejecting legitimate
connections.

This PR sets a custom header that should not be interfered with by any
JVM implementation to store the IP address, while maintaining the old
header as a fallback. The proxy will set both headers to allow the
server to gracefully migrate from Java 8 and Java 17 (and potentially
rollback).

Also removed some headers and logic that are not used.

TESTED=deployed both the proxy and nomulus to crash. Verified that EPP login correctly parsed the client IP in both Java 8 and Java 17 runtimes

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2224)

### DIFF
--- a/core/src/main/java/google/registry/flows/EppRequestHandler.java
+++ b/core/src/main/java/google/registry/flows/EppRequestHandler.java
@@ -75,14 +75,6 @@ public class EppRequestHandler {
           && eppOutput.getResponse().getResult().getCode() == SUCCESS_AND_CLOSE) {
         response.setHeader(ProxyHttpHeaders.EPP_SESSION, "close");
       }
-      // If a login request returns a success, a logged-in header is added to the response to inform
-      // the proxy that it is no longer necessary to send the full client certificate to the backend
-      // for this connection.
-      if (eppOutput.isResponse()
-          && eppOutput.getResponse().isLoginResponse()
-          && eppOutput.isSuccess()) {
-        response.setHeader(ProxyHttpHeaders.LOGGED_IN, "true");
-      }
     } catch (Exception e) {
       logger.atWarning().withCause(e).log("handleEppCommand general exception.");
       response.setStatus(SC_BAD_REQUEST);
--- a/core/src/main/java/google/registry/flows/TlsCredentials.java
+++ b/core/src/main/java/google/registry/flows/TlsCredentials.java
@@ -26,6 +26,7 @@ import com.google.common.net.InetAddresses;
 import dagger.Module;
 import dagger.Provides;
 import google.registry.config.RegistryConfig.Config;
+import google.registry.config.RegistryEnvironment;
 import google.registry.flows.EppException.AuthenticationErrorException;
 import google.registry.flows.certs.CertificateChecker;
 import google.registry.flows.certs.CertificateChecker.InsecureCertificateException;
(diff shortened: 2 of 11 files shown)


jianglai (commented):
> Reviewed 10 of 11 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2224)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2224)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @jianglai)

___
*[`util/src/main/java/google/registry/util/ProxyHttpHeaders.java` line 37 at r3](https://reviewable.io/reviews/google/nomulus/2224#-NjOcdtK3iWbhb5-XBL_:-NjOcdtL7QutMBzYVMeP:bqn4578) ([raw file](https://github.com/google/nomulus/blob/f5dd053e3b0e7319e3a1e7680ab1e4f7b533f826/util/src/main/java/google/registry/util/ProxyHttpHeaders.java#L37)):*
> ```Java
> 
>   /** HTTP header name used to pass the client IP address from the proxy to Nomulus. */
>   public static final String IP_ADDRESS = "Client-Address";
> ```

Should probably be `X-` something based on established convention (if that's actually correct), and maybe something not quite so generic?  Put `Nomulus` in it perhaps?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2224)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`util/src/main/java/google/registry/util/ProxyHttpHeaders.java` line 37 at r3](https://reviewable.io/reviews/google/nomulus/2224#-NjOcdtK3iWbhb5-XBL_:-NjP2dy35vdPcKX6D1iN:b-xzo72q) ([raw file](https://github.com/google/nomulus/blob/f5dd053e3b0e7319e3a1e7680ab1e4f7b533f826/util/src/main/java/google/registry/util/ProxyHttpHeaders.java#L37)):*

Should probably be `X-` something based on established convention (if that's actually correct), and maybe something not quite so generic?  Put `Nomulus` in it perhaps?
</blockquote></details>

According to MDN the `X-` prefix convention is [deprecated as of 2012](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).

Ideally we would want to change all of our custom headers to be `Nomulus-` prefixed, perhaps.

gbrodman (approved):
> Reviewed 10 of 11 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2224)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2224)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys)

PR #2225, opened 2023-11-16
Add an optional search term for ConsoleDomainListAction

gbrodman:
It's a case-insensitive query and it can appear anywhere (including TLDs)

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2225)

### DIFF
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleDomainListAction.java
@@ -19,6 +19,7 @@ import static google.registry.persistence.transaction.TransactionManagerFactory.
 
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Ascii;
 import com.google.gson.Gson;
 import com.google.gson.annotations.Expose;
 import google.registry.model.CreateAutoTimestamp;
@@ -33,6 +34,7 @@ import google.registry.ui.server.registrar.JsonGetAction;
 import java.util.List;
 import java.util.Optional;
 import javax.inject.Inject;
+import javax.persistence.TypedQuery;
 import org.joda.time.DateTime;
 
 /** Returns a (paginated) list of domains for a particular registrar. */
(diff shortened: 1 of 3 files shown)


ptkach (changes requested):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2225)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 233 at r1](https://reviewable.io/reviews/google/nomulus/2225#-NkLflsLCEn89EHdqeDs:-NkLflsLCEn89EHdqeDt:b-vjj8jc) ([raw file](https://github.com/google/nomulus/blob/3546d138c478b0698c4f8dda0c30461fef1744e0/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L233)):*
> ```Java
>   @Parameter("searchTerm")
>   public static Optional<String> provideSearchTerm(HttpServletRequest req) {
>     return extractOptionalParameter(req, "searchTerm");
> ```

Although I agree that's very neat approach to extract a single param, I'm worried about SQL injection or even just unintended use, which that definitely allows. I think we should break it down into separate parameters

ptkach (approved):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2225)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java` line 233 at r1](https://reviewable.io/reviews/google/nomulus/2225#-NkLflsLCEn89EHdqeDs:-NkS00z117m_GH32Ci1O:b-8ku2mb) ([raw file](https://github.com/google/nomulus/blob/3546d138c478b0698c4f8dda0c30461fef1744e0/core/src/main/java/google/registry/ui/server/registrar/RegistrarConsoleModule.java#L233)):*

Break down what into separate parameters? This is just a single string field; I'm not sure what you mean. 

And SQL injection is handled at the SQL query level as we always do
</blockquote></details>

Discussed offline, all clear

PR #2233, opened 2023-11-29
Make read-only transactions more performant

jianglai:
Since the replica SQL instance is read-only, any transaction performed
  on it should be explicitly read-only, which would allow PostgreSQL to
  optimize away (some) use of predicate locks.

  Also changed the EPP cache to read from the replica. The foreign key
  cache already behaves this way.

See: https://www.postgresql.org/docs/current/transaction-iso.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2233)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2233)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (changes requested):
`the large number predicate  
locks needed for higher transaction isolation level.`

Only SERIALIZABLE level uses predicate locks.



Btw, with whois, the foreignkey cache uses replica, but the EppResource cache seems to use the primary.
We should see some improvement here.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2233#-:-NkRELx59i9Djoqj_OTM:biufzks)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/persistence/PersistenceModule.java` line 269 at r1](https://reviewable.io/reviews/google/nomulus/2233#-NkREiPiCgV2xIJdSklh:-NkREiPiCgV2xIJdSkli:b-n9jqil) ([raw file](https://github.com/google/nomulus/blob/f5c4c758e00a9e0d8eed087bd4fa973d1b66a975/core/src/main/java/google/registry/persistence/PersistenceModule.java#L269)):*
> ```Java
>         name -> overrides.put(HIKARI_DS_CLOUD_SQL_INSTANCE, name));
>     overrides.put(
>         Environment.ISOLATION, TransactionIsolationLevel.TRANSACTION_READ_UNCOMMITTED.name());
> ```

I'm not comfortable with this change. 

Although a replica only has a single stream of mutation, it happens concurrent with the read queries.

With READ_COMMITTED, we don't get a consistent snapshot across an entire data set.

E.g., assume two vars A, and B

On Primary:

T1: A=1, B=1
T2: A=2, B=2

On replica, a Txn starting at T2-delta may see A=1, B=2 if we use READ_COMMITTED.



_Code quote:_
```Java
TRANSACTION_READ_UNCOMMITTED
```

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2233)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/persistence/PersistenceModule.java` line 269 at r1](https://reviewable.io/reviews/google/nomulus/2233#-NkREiPiCgV2xIJdSklh:-NkRTdyb7zv2qk2Qw1Tj:bckhmen) ([raw file](https://github.com/google/nomulus/blob/f5c4c758e00a9e0d8eed087bd4fa973d1b66a975/core/src/main/java/google/registry/persistence/PersistenceModule.java#L269)):*

I think you are referring to READ_UNCOMMITTED and are concerned about dirty read. In Postgres READ_COMMITTED and READ_UNCOMMITTED are actually the same and dirty read is never possible.

But given the fact that REPEATABLE_READ doesn't use locks, I think it is fine to keep using it.
</blockquote></details>

Not dirty read, but read-write race condition.

Timeline on Replica:

                        
T1:  replicator: replicate transaction: A=1, B=1
T2: user query starts: read A (get 1)
T3: replicator: replicate another transaction: A=2, B-2
T4: user query continues: read B (get 2)

PR #2236, opened 2023-11-30
Forbid domain creation with label blocked by BSA

weiminyu:
Add a BSA label check in the DomainCreation flow.

Also define a new exception to be thrown on blocked labels. Its error message will be copied to the Epp output.


core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
+
+import org.joda.time.DateTime;
+
+/** Testing utils for users of {@link BsaLabel}. */
+public final class BsaLabelUtils {
+
+  private BsaLabelUtils() {}
+
+  public static void persistBsaLabel(String domainLabel, DateTime creationTime) {
```
CydeWeys, line 26: Can you just add this method to `DatabaseHelper` or would that create bad dependencies?  I definitely don't think we need a whole class just for a few test helper methods.

core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java
```diff
@@ -251,6 +252,7 @@ public EppResponse run() throws EppException {
     verifyResourceDoesNotExist(Domain.class, targetId, now, registrarId);
     // Validate that this is actually a legal domain name on a TLD that the registrar has access to.
     InternetDomainName domainName = validateDomainName(command.getDomainName());
+    verifyNotBlockedByBsa(domainName);
```
CydeWeys, line 255: Add an `@error` comment above on the class def.  (Which will require modifying another file that a build test checks, I think `flows.md`.)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/persistence/BsaLabel.java` line 84 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWJd6vBboHzEWruKSy:-NkWJd6wC8OZXBbuOebf:b-u0buij) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/bsa/persistence/BsaLabel.java#L84)):*
> ```Java
>   /** Checks if the {@code domainLabel} (the leading `part` of a domain name) is blocked by BSA. */
>   public static boolean isLabelBlocked(String domainLabel) {
>     return tm().loadByKeyIfPresent(vKey(domainLabel)).isPresent();
> ```

Because this is in the critical path of domain creation, it needs to use an in-memory cache so that we don't have to keep going to the DB when the same name is being checked over and over (as it would be during drop-catching).  Refer to other examples in the codebase of caching in the create path, e.g. whether a given string is premium is cached.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1756 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMQ0P66yfVsY21hlf:-NkWMQ0P66yfVsY21hlg:b-nn6gpf) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1756)):*
> ```Java
>   static class DomainLabelBlockedByBsaException extends ParameterValuePolicyErrorException {
>     public DomainLabelBlockedByBsaException() {
>       super("Domain label is blocked by the Brand Safety Alliance.");
> ```

Shouldn't end with a period, and also, this needs to be the exact error message that the BSA wants?  Or is this it?

And there should absolutely be a test with an EPP response XML showing what this looks like.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 263 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMdFn3I2s_mCeLlP2:-NkWMdFn3I2s_mCeLlP3:b-66dnag) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L263)):*
> ```Java
>   }
> 
>   public static void verifyNotBlockedByBsa(InternetDomainName domainName)
> ```

Make this just take a `domainLabel`, since we're already parsing out `domainLabel` in the flow above.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 255 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMjAv0In6vHuYyKs5:-NkWMjAv0In6vHuYyKs6:bu3tqez) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L255)):*
> ```Java
>     // Validate that this is actually a legal domain name on a TLD that the registrar has access to.
>     InternetDomainName domainName = validateDomainName(command.getDomainName());
>     verifyNotBlockedByBsa(domainName);
> ```

This check (which involves hitting the DB) needs to be moved farther down.  E.g. if a create fails because the TLD doesn't exist, then that's of higher precedence than merely the label being blocked.  And if they try to reference a contact or nameserver that doesn't exist, that's also of higher precedence than a label being blocked.  Or if it's the wrong launch phase, same issue.  And it certainly needs to be below the reserved check, right?  As reservations take precedence over BSA?  (Or am I wrong about that?)  If so, it should be at the very end, right before `flowCustomLogic.afterValidation()`.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 263 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWNq1K5gaAiyDPYIpb:-NkWNq1K5gaAiyDPYIpc:b-idm8u7) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L263)):*
> ```Java
>   }
> 
>   public static void verifyNotBlockedByBsa(InternetDomainName domainName)
> ```

This should also take the `Tld` object and verify that the BSA blocking is actually enabled on this TLD and has begun already (check the `DateTime`).

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/bsa/persistence/BsaLabel.java` line 84 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWJd6vBboHzEWruKSy:-NkWKbMXDKn5nUyuItxe:b-d36rnd) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/bsa/persistence/BsaLabel.java#L84)):*

Because this is in the critical path of domain creation, it needs to use an in-memory cache so that we don't have to keep going to the DB when the same name is being checked over and over (as it would be during drop-catching).  Refer to other examples in the codebase of caching in the create path, e.g. whether a given string is premium is cached.
</blockquote></details>

I intentionally avoided caching because caching would allow blocked labels to be created after we've installed the labels.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 255 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMjAv0In6vHuYyKs5:-NkX5BV-37dRhV0ymhbc:b-896fix) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L255)):*

This check (which involves hitting the DB) needs to be moved farther down.  E.g. if a create fails because the TLD doesn't exist, then that's of higher precedence than merely the label being blocked.  And if they try to reference a contact or nameserver that doesn't exist, that's also of higher precedence than a label being blocked.  Or if it's the wrong launch phase, same issue.  And it certainly needs to be below the reserved check, right?  As reservations take precedence over BSA?  (Or am I wrong about that?)  If so, it should be at the very end, right before `flowCustomLogic.afterValidation()`.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java` line 255 at r1](https://reviewable.io/reviews/google/nomulus/2236#gh-1411054085:-NkX4gB05JRtd0TVDs1Y:b-896fix) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainCreateFlow.java#L255)):*

Add an `@error` comment above on the class def.  (Which will require modifying another file that a build test checks, I think `flows.md`.)
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 263 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMdFn3I2s_mCeLlP2:-NkX5eCd6sKF6lEWnVo0:b-896fix) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L263)):*

Make this just take a `domainLabel`, since we're already parsing out `domainLabel` in the flow above.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 263 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWNq1K5gaAiyDPYIpb:-NkX6dzxEMC5X1Z30DVD:b-896fix) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L263)):*

This should also take the `Tld` object and verify that the BSA blocking is actually enabled on this TLD and has begun already (check the `DateTime`).
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1756 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMQ0P66yfVsY21hlf:-NkWQfQH7GIRYhWGoFMI:bx5uzb3) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1756)):*

Shouldn't end with a period, and also, this needs to be the exact error message that the BSA wants?  Or is this it?

And there should absolutely be a test with an EPP response XML showing what this looks like.
</blockquote></details>

BSA doesn't specify EPP response. We are on our own.

Removed the period.
___
*[`core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/2236#gh-1411052908:-NkWJYF9DSmk4-oMVdky:bey0e9o) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java#L26)):*

Can you just add this method to `DatabaseHelper` or would that create bad dependencies?  I definitely don't think we need a whole class just for a few test helper methods.
</blockquote></details>

I don't like over-exposure of classes just for tests' sake.

Most of the BSA entity model intentionally have non-public constructors and getters.

I can change if you think this is overboard.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/persistence/BsaLabel.java` line 84 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWJd6vBboHzEWruKSy:-NkXK8tFEbgIw74wGy7T:b-f806pv) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/bsa/persistence/BsaLabel.java#L84)):*

I intentionally avoided caching because caching would allow blocked labels to be created after we've installed the labels.
</blockquote></details>

So there's 3 existing shorter cache durations in our codebase that we could reuse here, depending on which we think is most suitable:

* 60 minutes: the default domain label list cache interval (used for premium and reserved list entries)
* 10 minutes: the default singleton cache interval, used for Registrar and Tld entities amongst others
* 1 minute: the default EPP resource cache duration, used to cache associated contacts/hosts as part of the domain create flow to prevent too many associated lookups

I think either 1 or 10 minutes is appropriate for the caching interval here.  But not having a cache is just gonna blow up the database every time we get lots of attempted creates (e.g. like we are seeing now during EAP).  We have several hours in which to block a label after a BSA order comes through, right?  So a cache on the order of minutes should be fine.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1756 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMQ0P66yfVsY21hlf:-NkXLTWH4guVH3E63JuG:bpfzjx0) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1756)):*

BSA doesn't specify EPP response. We are on our own.

Removed the period.
</blockquote></details>

Weird.  They SHOULD specify the EPP response!  We should ask Quoc.  Maybe it's an oversight in their docs?  They specify the response in WHOIS/RDAP.

Still needs the XML btw.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 271 at r2](https://reviewable.io/reviews/google/nomulus/2236#-NkXLtBY8E4pHF4SkBRu:-NkXLtBY8E4pHF4SkBRv:b-4qbem4) ([raw file](https://github.com/google/nomulus/blob/6a07738e4e1fcf8c5a7f0d7181085937004318ff/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L271)):*
> ```Java
>   public static void verifyNotBlockedByBsa(String domainLabel, Tld tld, DateTime now)
>       throws DomainLabelBlockedByBsaException {
>     if (tld.getBsaEnrollStartTime() == null || tld.getBsaEnrollStartTime().isAfter(now)) {
> ```

I know it's not strictly being added in this PR, but `getBsaEnrollStartTime()` should be returning an `Optional<DateTime>` as is our standard pattern for turning `null` in the DB realm into something more friendly to use in the Java realm.

And then this check could be something like:

```
if (tld.getBsaEnrollStartTime().orElse(END_OF_TIME).isBeforeOrAt(now) && BsaLabel.isLabelBlocked(domainLabel)) {
  throw new DomainLabelBlockedByBsaException();
}
```

No need for the short circuit return at that point (which is unnecessary in such a small function), and intuitively I find it makes more sense to check if the BSA is enabled, and if so, do it, than to check if it's disabled, and if so, don't do it.
___
*[`core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/2236#gh-1411052908:-NkXLgpgCFEV7Sj8bOla:bpcqn1d) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java#L26)):*

I don't like over-exposure of classes just for tests' sake.

Most of the BSA entity model intentionally have non-public constructors and getters.

I can change if you think this is overboard.
</blockquote></details>

Ah so it's here in this package because you're using package-private.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/bsa/persistence/BsaLabel.java` line 84 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWJd6vBboHzEWruKSy:-Nl4pK3e2A8RSFZnFI_C:b-nprqgw) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/bsa/persistence/BsaLabel.java#L84)):*

So there's 3 existing shorter cache durations in our codebase that we could reuse here, depending on which we think is most suitable:

* 60 minutes: the default domain label list cache interval (used for premium and reserved list entries)
* 10 minutes: the default singleton cache interval, used for Registrar and Tld entities amongst others
* 1 minute: the default EPP resource cache duration, used to cache associated contacts/hosts as part of the domain create flow to prevent too many associated lookups

I think either 1 or 10 minutes is appropriate for the caching interval here.  But not having a cache is just gonna blow up the database every time we get lots of attempted creates (e.g. like we are seeing now during EAP).  We have several hours in which to block a label after a BSA order comes through, right?  So a cache on the order of minutes should be fine.
</blockquote></details>

Using the one-minute cache. Also moved method to another class.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1756 at r1](https://reviewable.io/reviews/google/nomulus/2236#-NkWMQ0P66yfVsY21hlf:-NkXN84X9z1PE5tkEqRl:b-m99agn) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1756)):*

Weird.  They SHOULD specify the EPP response!  We should ask Quoc.  Maybe it's an oversight in their docs?  They specify the response in WHOIS/RDAP.

Still needs the XML btw.
</blockquote></details>

I asked Quoc about DomainCheckFlow a while ago. He said "not required for EPP flows".

Added a check for response text.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 271 at r2](https://reviewable.io/reviews/google/nomulus/2236#-NkXLtBY8E4pHF4SkBRu:-Nl0FbPQ0WgJ7g_AOgtn:bb6un1j) ([raw file](https://github.com/google/nomulus/blob/6a07738e4e1fcf8c5a7f0d7181085937004318ff/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L271)):*

I know it's not strictly being added in this PR, but `getBsaEnrollStartTime()` should be returning an `Optional<DateTime>` as is our standard pattern for turning `null` in the DB realm into something more friendly to use in the Java realm.

And then this check could be something like:

```
if (tld.getBsaEnrollStartTime().orElse(END_OF_TIME).isBeforeOrAt(now) && BsaLabel.isLabelBlocked(domainLabel)) {
  throw new DomainLabelBlockedByBsaException();
}
```

No need for the short circuit return at that point (which is unnecessary in such a small function), and intuitively I find it makes more sense to check if the BSA is enabled, and if so, do it, than to check if it's disabled, and if so, don't do it.
</blockquote></details>

Done
___
*[`core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java` line 26 at r1](https://reviewable.io/reviews/google/nomulus/2236#gh-1411052908:-Nl0FMg22pQdYRWomtT5:bb77ufg) ([raw file](https://github.com/google/nomulus/blob/92b18ba3f7a95221ac5b7ad2f864e6cd413b2082/core/src/test/java/google/registry/bsa/persistence/BsaLabelUtils.java#L26)):*

Ah so it's here in this package because you're using package-private.
</blockquote></details>

Yes.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/bsa/persistence/BsaLabelUtils.java` line 46 at r6](https://reviewable.io/reviews/google/nomulus/2236#-Nl9AamR0UcyXrc7I5PL:-Nl9AamR0UcyXrc7I5PM:b1wlich) ([raw file](https://github.com/google/nomulus/blob/be7d991d5bf7de6b172bf76d9b1fe95bd5f3d791/core/src/main/java/google/registry/bsa/persistence/BsaLabelUtils.java#L46)):*
> ```Java
>         public Map<VKey<BsaLabel>, Optional<BsaLabel>> loadAll(
>             Iterable<? extends VKey<BsaLabel>> keys) {
>           throw new UnsupportedOperationException(
> ```

You are going to need to support this for the `domain:check` operation (as up to 50 domains can be check simultaneously), though that can be in a follow-up PR that implements the changes to the `DomainCheckFlow`.
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 1764 at r6](https://reviewable.io/reviews/google/nomulus/2236#-Nl9BIsB6BeSB7sLR7af:-Nl9BIsB6BeSB7sLR7ag:b-p4vc7x) ([raw file](https://github.com/google/nomulus/blob/be7d991d5bf7de6b172bf76d9b1fe95bd5f3d791/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L1764)):*
> ```Java
>   static class DomainLabelBlockedByBsaException extends ParameterValuePolicyErrorException {
>     public DomainLabelBlockedByBsaException() {
>       super("Domain label is blocked by the Brand Safety Alliance");
> ```

We'll probably want to workshop the exact text here once we hear back from the BSA on their recommendation.  But that can wait for later.
___
*[`core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java` line 2616 at r6](https://reviewable.io/reviews/google/nomulus/2236#-Nl9BcTf40fSNc2AYBDF:-Nl9BcTf40fSNc2AYBDG:bjxgfyk) ([raw file](https://github.com/google/nomulus/blob/be7d991d5bf7de6b172bf76d9b1fe95bd5f3d791/core/src/test/java/google/registry/flows/domain/DomainCreateFlowTest.java#L2616)):*
> ```Java
>                     .build()),
>             ValidationMode.STRICT);
>     assertThat(new String(responseXmlBytes, StandardCharsets.UTF_8))
> ```

Just use an XML testdata file for this. It's better to see exactly what that error looks like in the EPP response.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2236#-:-NlPiWvuD0Kx_5qkYQqC:bnfp4nl)* status: 0 of 16 files reviewed, all discussions resolved

PR #2238, opened 2023-11-30
Return the correct renewal fee for anchor tenants in domain checks

gbrodman:
The code as previously written assumed that creation fees would be the same as renewal fees -- this is not the case for anchor tenants, where the renewal fee is always the standard cost for the TLD (instead of any premium cost). This was already handled properly in the actual billing implementation, but we didn't tell the user the right renewal cost in domain checks.

This also removes some warning logs related to nested transactions

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2238)


gbrodman:
note: testSuccess_allocationToken_premiumAnchorTenant_noFee (with the new test renew check added) incorrectly returns $100 as the renewal cost without the logic changes in this PR

PR #2243, opened 2023-12-04
Increase delete allocation tokens command DB batch size

CydeWeys:
The existing figure of 20 is likely a remnant from Datastore (what with its max 25 entity groups per transaction limit), but in PostgreSQL this limit doesn't exist and we want to do bigger batches so that the command executes faster.


gbrodman:
i don't think this was the issue, as it never even finished deleting one batch

PR #2253, opened 2023-12-12
Do not double-enqueue NordnVerifyAction

jianglai:
Currently, a verify action is enqueued every time the upload method
succeeds. Because the upload job is wrapped in a transaction, the
same task will be enqueued again if the transaction retries.

We cannot move the upload method outside the transaction because the
read-upload-write logic needs to be atomic, and the upload part itself
is idempotent (therefore retri-able). We can, however, move the
enqueuing part outside the transaction as we only need to enqueue the
verify task once the transaction succeeds. This should fix the issue
where multiple verify jobs try to hit the same marksdb endpoints,
resulting in 429 (Too Many Requests) errors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2253)

### DIFF
--- a/core/src/main/java/google/registry/tmch/NordnUploadAction.java
+++ b/core/src/main/java/google/registry/tmch/NordnUploadAction.java
@@ -127,7 +127,7 @@ public final class NordnUploadAction implements Runnable {
         phase.equals(PARAM_LORDN_PHASE_SUNRISE) || phase.equals(PARAM_LORDN_PHASE_CLAIMS),
         "Invalid phase specified to NordnUploadAction: %s.",
         phase);
-    Optional<URL> verifyUrl =
+    Optional<URL> uploadUrl =
         tm().transact(
                 () -> {
                   // Note here that we load all domains pending Nordn in one batch, which should not
@@ -181,7 +181,7 @@ public final class NordnUploadAction implements Runnable {
                     throw new RuntimeException(e);
                   }
                 });
-    verifyUrl.ifPresent(
+    uploadUrl.ifPresent(
         url -> cloudTasksUtils.enqueue(NordnVerifyAction.QUEUE, makeVerifyTask(url)));
   }
 
--- a/core/src/main/java/google/registry/tmch/NordnUploadAction.java
+++ b/core/src/main/java/google/registry/tmch/NordnUploadAction.java
@@ -53,6 +53,7 @@ import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.SecureRandom;
 import java.util.List;
+import java.util.Optional;
 import java.util.Random;
 import javax.inject.Inject;
 import org.joda.time.Duration;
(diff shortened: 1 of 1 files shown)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2253)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2253)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tmch/NordnUploadAction.java` line 178 at r2](https://reviewable.io/reviews/google/nomulus/2253#-NlUI-v7EPFkUA-9DVYL:-NlUI-v89wXOI196iGOT:b-m6q1ge) ([raw file](https://github.com/google/nomulus/blob/30fc2d17817683024c94c1ab98e2f468f63a09c2/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L178)):*
> ```Java
>                     URL url =
>                         uploadCsvToLordn(String.format("/LORDN/%s/%s", tld, phase), header + csv);
>                     tm().updateAll(newDomains.build());
> ```

these two lines don't need to be within the try block but i don't actually care that much since it'd require declaring the url outside of it 
___
*[`core/src/main/java/google/registry/tmch/NordnUploadAction.java` line 184 at r2](https://reviewable.io/reviews/google/nomulus/2253#-NlUJ4jp8oI5UEqVmWAk:-NlUJ4jp8oI5UEqVmWAl:blbmrv5) ([raw file](https://github.com/google/nomulus/blob/30fc2d17817683024c94c1ab98e2f468f63a09c2/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L184)):*
> ```Java
>                   }
>                 });
>     verifyUrl.ifPresent(
> ```

I think this maybe should be something like "uploadUrl", or "urlToVerify"? Just to be as descriptive as possible about what the url is actually representing

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2253)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/tmch/NordnUploadAction.java` line 178 at r2](https://reviewable.io/reviews/google/nomulus/2253#-NlUI-v7EPFkUA-9DVYL:-NlUKXrr7q79P-PLazj8:b-rk2j4n) ([raw file](https://github.com/google/nomulus/blob/30fc2d17817683024c94c1ab98e2f468f63a09c2/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L178)):*

these two lines don't need to be within the try block but i don't actually care that much since it'd require declaring the url outside of it
</blockquote></details>

Yeah, that's the reason that I moved it inside the try block. Otherwise I also have to deal with potential nulls even it would never actually occur because the statement would have thrown before the null value is accessed.
___
*[`core/src/main/java/google/registry/tmch/NordnUploadAction.java` line 184 at r2](https://reviewable.io/reviews/google/nomulus/2253#-NlUJ4jp8oI5UEqVmWAk:-NlUKk9NEHG60WjacJ4O:b-ojm0h4) ([raw file](https://github.com/google/nomulus/blob/30fc2d17817683024c94c1ab98e2f468f63a09c2/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L184)):*

I think this maybe should be something like "uploadUrl", or "urlToVerify"? Just to be as descriptive as possible about what the url is actually representing
</blockquote></details>

Ironically, I had wanted to call it `verifyUrl`, but because of closure that variable name is already taken...

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2253)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/tmch/NordnUploadAction.java` line 184 at r2](https://reviewable.io/reviews/google/nomulus/2253#-NlUJ4jp8oI5UEqVmWAk:-NlULMZVD_zW2EhNRt2q:bjrxk2s) ([raw file](https://github.com/google/nomulus/blob/30fc2d17817683024c94c1ab98e2f468f63a09c2/core/src/main/java/google/registry/tmch/NordnUploadAction.java#L184)):*

Ironically, I had wanted to call it `verifyUrl`, but because of closure that variable name is already taken...
</blockquote></details>

oh I meant the outer variable that the transact block returns

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2253)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2255, opened 2023-12-14
 Build Nomulus with Java 17 

jianglai:
This PR makes it possible to build the Nomulus code base using Java 17.
Building with Java 11 continue to be possible and the resulting bytecodes are
still at Java 8 level. Also upgraded Gradle to 8.5.

There are several necessary changes to make this happen:

1. Some Gradle plugins need to be upgraded to support Java 17, notably
errorprone. As a result, a lot more "errors" were caught and corrected.

2. All test code are now built and run at Java 8 level. Previously it was left
undefined (which defaults to the version of the compiler) and had led to
situations where we inadvertently called Java 8+ features in production that
are not caught by tests. The change also made the java8compatibility subproject
obsolete, which is therefore removed.

3. Removed the docs subproject. Its main use is to generate flows.md, but it
relies heavily on Java internal APIs that have changed significant with each
version. Upgrading to Java 11 required extensive refactoring of the code there,
and Java 17 again removed many APIs that were used. I don't think it is worth
the maintenance effort just to have a tool to generate flows.md which no one
actually reads.

4. Capped a few GCP dependencies because the latest version depends on
 grpc-java >= 1.59.0, which includes a runtime incompatibility
 (https://github.com/grpc/grpc-java/releases/tag/v1.59.0).

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2255)


weiminyu (commented):
Are you sure java8compatibility can be removed? It checks third-party dependencies too. 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-NliOV_zAln5a67wUMUx:braquvk)* status: 0 of 139 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

weiminyu (changes requested):
You may want to check with ben about removing docs.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-NliPBgS3iMVVjINA9Sb:bvo77ns)* status: 0 of 139 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @jianglai)

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255)* status: 0 of 139 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @jianglai)

___
*[`build.gradle` line 368 at r6](https://reviewable.io/reviews/google/nomulus/2255#-NliPbdJ7ytZnWq9mDpG:-NliPbdJ7ytZnWq9mDpH:b4bwves) ([raw file](https://github.com/google/nomulus/blob/217af8ad71d5bb96326d7194abeee2785600a84d/build.gradle#L368)):*
> ```Gradle
>     compileTestJava {
>       // TODO: Remove this once we migrate off AppEngine.
>       options.release = 8
> ```

I don't see the needs for this. It doesn't affect production code.



_Code quote:_
```Gradle
 options.release = 8
```

jianglai (commented):
I would have kept java8compatibility tests in place, were it not for the fact that it also has come incompatibility issues with Java 17. I have a hard time understanding why it was needed in the first place, though. I would have expected that since all prod jars are built at Java 8 level, it would not have been possible for them to have a compile time dependency on any jars that requires Java 9+. Since this class is checking the jar files, it would not have caught any runtime dependency that requires a higher version of Java, would it?

Regarding docs, it is my firm belief (having spent considerable time upgrading it from Java 8 to Java 11) that it is not worth keeping around. @CydeWeys feel free to chime in.

> Reviewed 125 of 128 files at r1, 13 of 13 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, 7 of 7 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-NlxYixE14KyrziWAZIh:b-cp7o7v)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman, @github-advanced-security[bot], and @weiminyu)

___
*[`build.gradle` line 368 at r6](https://reviewable.io/reviews/google/nomulus/2255#-NliPbdJ7ytZnWq9mDpG:-Nlx_xb6Dl6Qp-nO8f3H:bpjek8h) ([raw file](https://github.com/google/nomulus/blob/217af8ad71d5bb96326d7194abeee2785600a84d/build.gradle#L368)):*

I don't see the needs for this. It doesn't affect production code.
</blockquote></details>

It doesn't. But when building with Java 17, it complains about version incompatibilities. Plus, I think it is generally a good idea to build/run your tests at the save language level as your production code.

Note that this is just a temporary measure. Once we build Java 17-level production code, we can start using Java 9+ level language features in tests again.
___
*[`core/src/main/java/google/registry/beam/spec11/SafeBrowsingTransforms.java` line 228 at r1](https://reviewable.io/reviews/google/nomulus/2255#gh-1426878819:-NldmS-h5BzKh-V7l7QH:b-iurftj) ([raw file](https://github.com/google/nomulus/blob/547400ce357f40ebf2bb6faa007facfe4f52f4fc/core/src/main/java/google/registry/beam/spec11/SafeBrowsingTransforms.java#L228)):*

## Log Injection

This log entry depends on a [user-provided value](1).

[Show more details](https://github.com/google/nomulus/security/code-scanning/1634)
</blockquote></details>

False positive. It is OK to log the response from safebrowsing.
___
*[`prober/src/main/java/google/registry/monitoring/blackbox/handler/WebWhoisActionHandler.java` line 90 at r1](https://reviewable.io/reviews/google/nomulus/2255#gh-1426878820:-NlhkQlK7b-ClHSWey8Y:b-7eu3fh) ([raw file](https://github.com/google/nomulus/blob/547400ce357f40ebf2bb6faa007facfe4f52f4fc/prober/src/main/java/google/registry/monitoring/blackbox/handler/WebWhoisActionHandler.java#L90)):*

## Log Injection

This log entry depends on a [user-provided value](1).

[Show more details](https://github.com/google/nomulus/security/code-scanning/1635)
</blockquote></details>

Proof-of-concept code, not used.

weiminyu (commented):
I remember situations where java 9+ dependency was leaked into the nomulus jar when we had the release=8 option set.
Maybe I remembered wrong, but it worth checking if a known 9+ dependency  can be linked with nomulus.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-Nlxc8oa3fdV6teLXjoE:b-h5me98)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

weiminyu (commented):
Direct dependencies are checked by the compiler so they are ok.

However, if an indirect dependency is resolved by gradle to a  version at java 9+ language level, 
the compiler cannot tell that, and the error only shows up on AppEngine.

I once saw this happening in Crash with a dependency used by the secret-manager.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-Nm83kntDAGqunSHy6qg:b-xd5sqd)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

jianglai:
> Direct dependencies are checked by the compiler so they are ok.
> 
> However, if an indirect dependency is resolved by gradle to a version at java 9+ language level, the compiler cannot tell that, and the error only shows up on AppEngine.
> 
> I once saw this happening in Crash with a dependency used by the secret-manager.
> 
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-Nm83kntDAGqunSHy6qg:b-xd5sqd)_ status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

So this is not an issue specific to App Engine? In theory any Java application can declare a target version but have an indirect dependency that requires a higher version of JRE? That makes the entire Gradle build system very brittle and unreliable... Does that mean everyone who uses Gradle needs to do this check?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

jianglai (commented):
> Reviewed 9 of 9 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

jianglai (commented):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

jianglai (commented):
> Reviewed 20 of 20 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-advanced-security[bot], and @ptkach)

weiminyu (commented):
In theory, yes. In practice, if you are reasonably up to date with your JVM version, you don't have  to.
Look how long it takes for libs like guava or hibernate to move past java 8. So once we move off AppEngine, it will be safe to delete this subproject.

Also, it is not that a lib specifically requires a higher version of jvm. The higher version slips in due to open-ended dependency ranges.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-Nn9zRbl85X8sCP_L71p:b-m1mg9s)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-advanced-security[bot], and @ptkach)

jianglai (commented):
This PR is a prerequisite to enabling Java 17 runtime on App Engine. At this point I think it is not worth the effort making the java8compatiblity package work (it uses some internal APIs not available in Java 17 anymore), given that after it's merged we will quickly turn on Java 17 and make the risk of accidentally pulling in a Java 8 only dependency in the meantime minimal. It is prudent that we merge this PR sooner as it touches many parts of the code base and is likely to be in merge conflict the longer we wait on it.

> Reviewed 13 of 13 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-NnjaI836JJhjttQLAxM:bjgqdu3)* status: all files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-advanced-security[bot], and @ptkach)

weiminyu (approved):
Approved. Once you submit it, we should probably forbid any dependency change until the AppEngine runtime environment is updated to Java 17 also.  Just last week Sarah ran into this problem briefly that was caught by the com[atibility check.

> Reviewed 1 of 128 files at r1, 1 of 13 files at r2, 1 of 20 files at r14, 1 of 20 files at r16.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-Nnjj0Je0VGMrIWwLz7b:bd90rv1)* status: 131 of 150 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-advanced-security[bot], @jianglai, and @ptkach)

jianglai:
> Approved. Once you submit it, we should probably forbid any dependency change until the AppEngine runtime environment is updated to Java 17 also. Just last week Sarah ran into this problem briefly that was caught by the com[atibility check.
> 
> > Reviewed 1 of 128 files at r1, 1 of 13 files at r2, 1 of 20 files at r14, 1 of 20 files at r16.
> > _[Reviewable](https://reviewable.io/reviews/google/nomulus/2255#-:-Nnjj0Je0VGMrIWwLz7b:bd90rv1)_ status: 131 of 150 files reviewed, 2 unresolved discussions (waiting on @gbrodman, @github-advanced-security[bot], @jianglai, and @ptkach)

Sure. I'll also take another crack at making the compatiblity test work after it's merged. I just want this to go through sooner rather than later.

PR #2258, opened 2023-12-19
Add type conversion to TimedTransitionProperty<Money> deserializer to handle JPY currency

sarahcaseybot:
JPY prices do not have cent values, whereas USD prices do. Therefore the Jackson deserializer reads in the prices from the YAML file as a Double object if it is in USD and as an Integer object if it is in JPY. The `MoneyDeserializer` class already handled this issue cleanly because it parses the object into a `MoneyJson` object which indicates that the amount should be parsed as a BigDecimal object. However, the `TimedTransitionProperty<Money>` deserializer was failing since it was attempting to cast the money value to a double which would fail when the value was an Integer for JPY TLDs. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2258)

### DIFF
--- a/core/src/main/java/google/registry/model/EntityYamlUtils.java
+++ b/core/src/main/java/google/registry/model/EntityYamlUtils.java
@@ -359,7 +359,7 @@ public class EntityYamlUtils {
                       key ->
                           Money.of(
                               CurrencyUnit.of(valueMap.get(key).get("currency").toString()),
-                              (double) valueMap.get(key).get("amount")))));
+                              new BigDecimal(String.valueOf(valueMap.get(key).get("amount")))))));
     }
   }
 
--- a/core/src/main/java/google/registry/model/EntityYamlUtils.java
+++ b/core/src/main/java/google/registry/model/EntityYamlUtils.java
@@ -349,7 +349,7 @@ public class EntityYamlUtils {
     @Override
     public TimedTransitionProperty<Money> deserialize(JsonParser jp, DeserializationContext context)
         throws IOException {
-      SortedMap<String, LinkedHashMap> valueMap = jp.readValueAs(SortedMap.class);
+      SortedMap<String, LinkedHashMap<String, Object>> valueMap = jp.readValueAs(SortedMap.class);
       return TimedTransitionProperty.fromValueMap(
           valueMap.keySet().stream()
               .collect(
--- a/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java
+++ b/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java
@@ -99,10 +99,8 @@ public class ConfigureTldCommandTest extends CommandTestCase<ConfigureTldCommand
     runCommandForced("--input=" + tldFile);
     Tld tld = Tld.get("jpy");
     assertThat(tld).isNotNull();
-    assertThat(tld.getDriveFolderId()).isEqualTo("driveFolder");
     assertThat(tld.getCreateBillingCost()).isEqualTo(Money.of(JPY, new BigDecimal("250")));
     testTldConfiguredSuccessfully(tld, "jpy.yaml");
-    assertThat(tld.getBreakglassMode()).isFalse();
   }
 
   @Test
--- a/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java
+++ b/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java
@@ -29,6 +29,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.logging.Level.INFO;
 import static org.joda.money.CurrencyUnit.JPY;
 import static org.joda.money.CurrencyUnit.USD;
+import static org.joda.time.DateTimeZone.UTC;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -51,7 +52,6 @@ import java.util.Optional;
 import java.util.logging.Logger;
 import org.joda.money.Money;
 import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
@@ -100,7 +100,7 @@ public class ConfigureTldCommandTest extends CommandTestCase<ConfigureTldCommand
     Tld tld = Tld.get("jpy");
     assertThat(tld).isNotNull();
     assertThat(tld.getCreateBillingCost()).isEqualTo(Money.of(JPY, new BigDecimal("250")));
-    assertThat(tld.getEapFeeFor(DateTime.now()).getCost()).isEqualTo(new BigDecimal(0));
+    assertThat(tld.getEapFeeFor(DateTime.now(UTC)).getCost()).isEqualTo(new BigDecimal(0));
     testTldConfiguredSuccessfully(tld, "jpy.yaml");
   }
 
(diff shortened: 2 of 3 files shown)


jianglai (changes requested):
Does it only affect TLDs that use JPY as its currency? I think all the prices are `BigDecimal`s, aren't they?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258#-:-Nm2rEGb9MDTuq-Bip8j:bcbsz2n)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 102 at r1](https://reviewable.io/reviews/google/nomulus/2258#-Nm2rBKHE7epvphcbL3E:-Nm2rBKI5FJzfQfJDUft:b8hplfs) ([raw file](https://github.com/google/nomulus/blob/cf3f4a02d288bdc198469c60c6e0f82ff452a2ae/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L102)):*
> ```Java
>     Tld tld = Tld.get("jpy");
>     assertThat(tld).isNotNull();
>     assertThat(tld.getDriveFolderId()).isEqualTo("driveFolder");
> ```

This is probably not needed.

sarahcaseybot (commented):
JPY prices are stored as BigIntegers, USD prices are stored as BigDecimal

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258#-:-Nm7SlQj0M9mcZZqPSj8:b-lpjq9q)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 102 at r1](https://reviewable.io/reviews/google/nomulus/2258#-Nm2rBKHE7epvphcbL3E:-Nm7TuguCQiyl3HuxUM-:b-896fix) ([raw file](https://github.com/google/nomulus/blob/cf3f4a02d288bdc198469c60c6e0f82ff452a2ae/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L102)):*

This is probably not needed.
</blockquote></details>

Done.

jianglai (changes requested):
I guess what makes me a bit confused is that the only change you made was to how `TimedTransitionProperty<Money>` is parsed. How could it affect the create price (the one you are testing), which is not a `TimedTransitionProperty`?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258#-:-Nm7f07i8cxNCjmMI83B:b-d90dku)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

sarahcaseybot (commented):
This was a failing test before I added the change to `TimedTransitionProperty<Money>` that failed on the line `runCommandForced` with the error message 

``` Caused by: java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Double ```

I added the asserts to this tests just to check that now with the change, that the TLD was created successfully and using the correct currency. For clarity, I have now also added a check that the EAP schedule was created.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258#-:-Nm7iL7-3EB2Fdp7UZp4:b-o81nwb)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @jianglai)

jianglai (approved):
The PR description is misleading as it talks about adding `BigInt` conversion when in fact it was converting to `BigDecimal`.

I suspect the test failure that you see has nothing to do with the create price as it was simply a `Money` which should have been handled correctly before this PR by `MoneyDeserializer`. I'm actually a bit curious why you need a special deserializer in the first place, given that you didn't have a special serializer. I'd assume if the built-in serializer can handle `TimeTransitionProperty<T>`, the built-in deserializer should be able to do the same. Is it just so that the resulting map is sorted?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258#-:-Nm8-g-tDx5WS4h-mNxn:b-am8dzs)* status: all files reviewed, 2 unresolved discussions (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/EntityYamlUtils.java` line 352 at r3](https://reviewable.io/reviews/google/nomulus/2258#-Nm83M6S35GAQOgRIePx:-Nm83M6S35GAQOgRIePy:bq8lyw5) ([raw file](https://github.com/google/nomulus/blob/dfae3918e0c74c6755851814fb542f4270cf262e/core/src/main/java/google/registry/model/EntityYamlUtils.java#L352)):*
> ```Java
>     public TimedTransitionProperty<Money> deserialize(JsonParser jp, DeserializationContext context)
>         throws IOException {
>       SortedMap<String, LinkedHashMap> valueMap = jp.readValueAs(SortedMap.class);
> ```

We should not use the raw type `LinkedHashMap` here, given that you are already doing an unchecked cast from `SortedMap` to `SortedMap<String, LinkedHashMap>` already. I think it should be `LinkedHashMap<String, Object>`?
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 102 at r3](https://reviewable.io/reviews/google/nomulus/2258#-Nm84ey375w2XKamMXQW:-Nm84ey375w2XKamMXQX:b-n9o3vm) ([raw file](https://github.com/google/nomulus/blob/dfae3918e0c74c6755851814fb542f4270cf262e/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L102)):*
> ```Java
>     Tld tld = Tld.get("jpy");
>     assertThat(tld).isNotNull();
>     assertThat(tld.getCreateBillingCost()).isEqualTo(Money.of(JPY, new BigDecimal("250")));
> ```

There is no need to use `BigDecimal` here. You can just provide a double to create the Money object.

sarahcaseybot (commented):
Agreed the failure was not on the create price, the failure was on the EapFeeSchedule. The use of the createfee check was just an extra gut check. 

Here is the link to the initial failing test: https://fusion2.corp.google.com/ci/kokoro/prod:domain_registry%2Fubuntu%2Fgob-pr/activity/8e2988ca-04b2-4685-a05a-05650e0a4fc0/log

I am unclear why the custom `TimedTransitionProperty<T>` serializer is not necessary, but the custom deserializer definitely is. I just confirmed once again and default deserializer is unable to deserialize `ImmutableSortedMap`.

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258#-:-Nm8DtzZ3t1sdJIgHdeP:b8j7ift)* status: all files reviewed, 2 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/model/EntityYamlUtils.java` line 352 at r3](https://reviewable.io/reviews/google/nomulus/2258#-Nm83M6S35GAQOgRIePx:-Nm8F5edD7kPQn5uhZT9:b-896fix) ([raw file](https://github.com/google/nomulus/blob/dfae3918e0c74c6755851814fb542f4270cf262e/core/src/main/java/google/registry/model/EntityYamlUtils.java#L352)):*

We should not use the raw type `LinkedHashMap` here, given that you are already doing an unchecked cast from `SortedMap` to `SortedMap<String, LinkedHashMap>` already. I think it should be `LinkedHashMap<String, Object>`?
</blockquote></details>

Done.
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 102 at r3](https://reviewable.io/reviews/google/nomulus/2258#-Nm84ey375w2XKamMXQW:-Nm8G58QCsRC7fexbKIc:b-z6tmus) ([raw file](https://github.com/google/nomulus/blob/dfae3918e0c74c6755851814fb542f4270cf262e/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L102)):*

There is no need to use `BigDecimal` here. You can just provide a double to create the Money object.
</blockquote></details>

Not when the currency is JPY. Error message is "Scale of amount 250.0 is greater than the scale of the currency JPY"

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2258)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #2261, opened 2023-12-20
Remove create_tld and update_tld commands

sarahcaseybot:
These commands are no longer necessary now that configure_tld command is available. However, the configure_tld command should only be used for crash, QA, and alpha environments. TLDs in production and sandbox must be modified using modifications to their config files in Gerrit unless using the configure_tld command in breakglass mode. Check the "How to configure TLDs" procedure doc for more info.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2261)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2261)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/google/nomulus/2261#-NmbaS5aCnsBJqNbNvH0):*
We should probably check in the ConfigureTldCommand that we're either not in prod/sandbox or that the breakglass flag is set right?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2261)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @CydeWeys)

___
*[`-- commits` line 4 at r1](https://reviewable.io/reviews/google/nomulus/2261#-NmbaS5aCnsBJqNbNvH0):*

We should probably check in the ConfigureTldCommand that we're either not in prod/sandbox or that the breakglass flag is set right?
</blockquote></details>

Just discussed this in 1:1 with @CydeWeys. Unfortunately, we cannot enforce forcing the breakglass flag in sandbox/production since Kokoro and Cloud Build run the command in those environments without the breakglass flag to actually sync TLDs. 

Decided to add a new flag to only be used by the build environment to indicate that the configureTldCommand is being run by the build environment and will make clear in documentation that human users should never use that flag. Then will modify the command to require either the build-environment flag or the breakglass flag to run in sandbox or production. I will add these changes in a follow-up PR.

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2261)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #2262, opened 2023-12-21
Bypass SCRYPT hashing in tests

jianglai:
SCRYPT is much computationally heavier than SHA265 (by design), which
resulted in test run time doubling due to most tests initializing canned
data that uses hashing.

Since out tests are not verifying the correctness of a specific hashing
algorithm anyway, this PR makes it so that simple concatenation is used
in tests.

Also moved RegistryEnvironment to the util subproject so it can be called by
PasswordUtils, which makes sense as it is a utility class.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2262)


jianglai (commented):
> Reviewed 39 of 40 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2262)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (approved):
> Reviewed 38 of 40 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2262)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

jianglai:
Note that I will have to temporarily disable presubmits because the internal codebase also has a dependency on `RegistryEnvironment`, which is being moved (see https://team-review.git.corp.google.com/c/domain-registry-eng/nomulus-internal/+/2118858). I think as long as the FOSS presubmit succeeds, we should be fairly confident that this PR works. It's not worth the effort making several changes just to make the internal presubmit not break during the migration, in my opinion.

PR #2271, opened 2024-01-04
Check BSA block status in CheckApi

weiminyu:
Checks for and reports BSA block status if the name is not registered or reserved.

Also moves CheckApiActionTest to standardTest. Whatever problem forcing it to another suite has apparently disappeared.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -38,7 +38,6 @@ def jsDir = "${project.projectDir}/src/main/javascript"
 def outcastTestPatterns = [
     // Problem seems to lie with AppEngine TaskQueue for test.
     "google/registry/batch/RefreshDnsOnHostRenameActionTest.*",
-    "google/registry/flows/CheckApiActionTest.*",
     "google/registry/flows/EppLifecycleHostTest.*",
     "google/registry/flows/domain/DomainCreateFlowTest.*",
     "google/registry/flows/domain/DomainUpdateFlowTest.*",
--- a/core/src/main/java/google/registry/flows/CheckApiAction.java
+++ b/core/src/main/java/google/registry/flows/CheckApiAction.java
@@ -16,12 +16,14 @@ package google.registry.flows;
 
 import static com.google.common.base.Strings.nullToEmpty;
 import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN;
+import static google.registry.flows.domain.DomainFlowUtils.isBlockedByBsa;
 import static google.registry.flows.domain.DomainFlowUtils.validateDomainName;
 import static google.registry.flows.domain.DomainFlowUtils.validateDomainNameWithIdnTables;
 import static google.registry.flows.domain.DomainFlowUtils.verifyNotInPredelegation;
 import static google.registry.model.tld.label.ReservationType.getTypeOfHighestSeverity;
 import static google.registry.model.tld.label.ReservedList.getReservationTypes;
 import static google.registry.monitoring.whitebox.CheckApiMetric.Availability.AVAILABLE;
+import static google.registry.monitoring.whitebox.CheckApiMetric.Availability.BSA_BLOCKED;
 import static google.registry.monitoring.whitebox.CheckApiMetric.Availability.REGISTERED;
 import static google.registry.monitoring.whitebox.CheckApiMetric.Availability.RESERVED;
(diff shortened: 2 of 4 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2271)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/CheckApiAction.java` line 129 at r1](https://reviewable.io/reviews/google/nomulus/2271#-NnLQTu46CUHgvxH8but:-NnLQTu46CUHgvxH8buu:b9yqyo4) ([raw file](https://github.com/google/nomulus/blob/56eeb4bdd4b4251a5079f0bcb0c70b731fef615a/core/src/main/java/google/registry/flows/CheckApiAction.java#L129)):*
> ```Java
>       }
> 
>       boolean isRegistered = checkExists(domainString, now);
> ```

I think we want an outer wrapping `tm().transact() { ... }` around both the `checkExists()` call and the `isBlockedbyBsa()` call, as both methods will load from the DB in most circumstances.

Might as well include the `Tld.get()` call inside the transaction as well, as it's rare that the check will short circuit out on invalid registry phase so it will usually need to go onto the subsequent DB loads anyway, plus getting a Tld does sometimes need to hit the DB (when that cache needs to be refreshed).

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2271)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @weiminyu)

___
*[`core/src/main/java/google/registry/flows/CheckApiAction.java` line 138 at r1](https://reviewable.io/reviews/google/nomulus/2271#-NnLRDVW8UNLL9KyeKtc:-NnLRDVW8UNLL9KyeKtd:b-5oa31e) ([raw file](https://github.com/google/nomulus/blob/56eeb4bdd4b4251a5079f0bcb0c70b731fef615a/core/src/main/java/google/registry/flows/CheckApiAction.java#L138)):*
> ```Java
>       }
>       if (!isRegistered && !isReserved) {
>         isBsaBlocked = isBlockedByBsa(domainName.parts().get(0), tld, clock.nowUtc());
> ```

This should use the transaction time once it's enrolled in a transaction as per above.

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2271)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/CheckApiAction.java` line 129 at r1](https://reviewable.io/reviews/google/nomulus/2271#-NnLQTu46CUHgvxH8but:-NnecRgNBJUGz7EpU_oF:b-u579us) ([raw file](https://github.com/google/nomulus/blob/56eeb4bdd4b4251a5079f0bcb0c70b731fef615a/core/src/main/java/google/registry/flows/CheckApiAction.java#L129)):*

I think we want an outer wrapping `tm().transact() { ... }` around both the `checkExists()` call and the `isBlockedbyBsa()` call, as both methods will load from the DB in most circumstances.

Might as well include the `Tld.get()` call inside the transaction as well, as it's rare that the check will short circuit out on invalid registry phase so it will usually need to go onto the subsequent DB loads anyway, plus getting a Tld does sometimes need to hit the DB (when that cache needs to be refreshed).
</blockquote></details>

There is no easy way to wrap just those few statements in a transaction, since the logic is based on exception thrown and we don't have a systematic way to pass exceptions across transaction boundary.

I just put most of the `doCheck` method into another one and transact on it. 
___
*[`core/src/main/java/google/registry/flows/CheckApiAction.java` line 138 at r1](https://reviewable.io/reviews/google/nomulus/2271#-NnLRDVW8UNLL9KyeKtc:-NnecQsG-sui3C5rLbap:b-896fix) ([raw file](https://github.com/google/nomulus/blob/56eeb4bdd4b4251a5079f0bcb0c70b731fef615a/core/src/main/java/google/registry/flows/CheckApiAction.java#L138)):*

This should use the transaction time once it's enrolled in a transaction as per above.
</blockquote></details>

Done.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2271#-:-NnjRLEbE_TOHm23pH0r:bnfp4nl)* status: 0 of 4 files reviewed, all discussions resolved

PR #2277, opened 2024-01-09
Define the --build_environment flag and change --break_glass flag to a Boolean type

sarahcaseybot:
It is necessary to define these flags in a deployment before merging go/r3pr/2273 in order to prevent breaking the exisitng TLD syncing and entity presubmit testing that has already been enabled.  

The --break_glass flag is now a Boolean parameter instead of a boolean to allow it to be used to also turn off break glass mode when it is set to false.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2277)

### DIFF
--- a/core/src/main/java/google/registry/tools/ConfigureTldCommand.java
+++ b/core/src/main/java/google/registry/tools/ConfigureTldCommand.java
@@ -64,22 +64,22 @@ public class ConfigureTldCommand extends MutatingCommand {
   Path inputFile;
 
   @Parameter(
-      names = {"-b", "--breakglass"},
+      names = {"-b", "--break_glass"},
       description =
-          "Sets the breakglass field on the TLD to true, preventing Cloud Build from overwriting"
+          "Sets the breakGlass field on the TLD to true, preventing Cloud Build from overwriting"
               + " these new changes until the TLD configuration file stored internally matches the"
               + " configuration in the database.")
-  boolean breakglass;
+  boolean breakGlass;
 
   @Parameter(
-      names = {"--end_breakglass"},
+      names = {"--end_break_glass"},
       description =
           "Sets the breakglass field on the TLD to false, indicating that whatever changes made to"
               + " the TLD should be overwritten by the internal config file on the next TLD sync."
-              + " If this flag is not used to end breakglass mode, breakglass mode will also end"
+              + " If this flag is not used to end break glass mode, break glass mode will also end"
               + " automatically if the configuration file stored internally is updated to match"
               + " the current state of the TLD.")
-  boolean endBreakglass;
+  boolean endBreakGlass;
 
   @Parameter(
       names = {"--build_environment"},
@@ -104,10 +104,10 @@ public class ConfigureTldCommand extends MutatingCommand {
   boolean newDiff = true;
 
   /**
-   * Indicates if the existing TLD is currently in breakglass mode and should not be modified unless
-   * the breakglass flag is used
+   * Indicates if the existing TLD is currently in break glass mode and should not be modified
+   * unless the --break_glass flag is used
    */
-  boolean oldTldInBreakglass = false;
+  boolean oldTldInBreakGlass = false;
 
   @Override
   protected void init() throws Exception {
@@ -118,23 +118,23 @@ public class ConfigureTldCommand extends MutatingCommand {
     Tld oldTld = getTlds().contains(name) ? Tld.get(name) : null;
     Tld newTld = mapper.readValue(inputFile.toFile(), Tld.class);
     if (oldTld != null) {
-      oldTldInBreakglass = oldTld.getBreakglassMode();
+      oldTldInBreakGlass = oldTld.getBreakglassMode();
       newDiff = !oldTld.equalYaml(newTld);
     }
 
-    if (!newDiff && !oldTldInBreakglass) {
+    if (!newDiff && !oldTldInBreakGlass) {
       // Don't construct a new object if there is no new diff
       return;
     }
 
-    if (oldTldInBreakglass && !breakglass) {
+    if (oldTldInBreakGlass && !breakGlass) {
       checkArgument(
           !newDiff,
-          "Changes can not be applied since TLD is in breakglass mode but the breakglass flag was"
-              + " not used");
+          "Changes can not be applied since TLD is in break glass mode but the --break_glass flag"
+              + " was not used");
       // if there are no new diffs, then the YAML file has caught up to the database and the
-      // breakglass mode should be removed
-      logger.atInfo().log("Breakglass mode removed from TLD: %s", name);
+      // break glass mode should be removed
+      logger.atInfo().log("Break glass mode removed from TLD: %s", name);
     }
 
     checkPremiumList(newTld);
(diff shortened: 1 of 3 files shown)


CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 67 at r1](https://reviewable.io/reviews/google/nomulus/2277#-NnoZC6C85WcDkWgsvUv:-NnoZC6C85WcDkWgsvUw:b-l5o1vt) ([raw file](https://github.com/google/nomulus/blob/50eea914dc7bb898021614fb35119c68553b24c8/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L67)):*
> ```Java
> 
>   @Parameter(
>       names = {"-b", "--breakglass"},
> ```

It might be cleaner to make this take a `boolean`, so that it will look like either:

```
nomulus update_tld dev --breakglass=true
nomulus update_tld dev --breakglass=false
```

What do you think?

Also "break glass" is two words so this should all be `breakGlass` for variables and `break_glass` for command line flags.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 67 at r1](https://reviewable.io/reviews/google/nomulus/2277#-NnoZC6C85WcDkWgsvUv:-NnuIJsH7cZpURuhJ3n1:b-4jhcm0) ([raw file](https://github.com/google/nomulus/blob/50eea914dc7bb898021614fb35119c68553b24c8/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L67)):*

I don't think the logic will work for making this take a boolean. The idea of go/r3pr/2273 was to require either a break glass or a build environment flag in the production environment. In this case, we would need to make this flag a required flag to pass the newly added check, but we cannot do that because the build environment by design does not know whether or not the TLD is in break glass mode and would not know which value to give the flag.
</blockquote></details>

Updated the flag name and variable names, note that the field name on the TLD, however,  is still breakglassMode.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 67 at r1](https://reviewable.io/reviews/google/nomulus/2277#-NnoZC6C85WcDkWgsvUv:-NnzEucR33e418UxSJtr:bbyqi16) ([raw file](https://github.com/google/nomulus/blob/50eea914dc7bb898021614fb35119c68553b24c8/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L67)):*

When a TLD is set to break glass mode, that means that we want to prevent the build system from overwriting any changes made via the command line with whatever is saved to the configuration files. The TLD sync is done using Cloud Build with this script https://cs.opensource.google/nomulus/nomulus/+/master:release/tld-updater/sync_tlds.sh and once the --build_environment tag is added, we will add that tag to the configure_tld command call in that script. 

Your suggestion to just have --break_glass=false indicate that the break glass mode should end would automatically end break glass mode on all Cloud Build calls of that script since without specifying a --break_glass flag, it would default to false. This would mean that the local changes would be overwritten by the internal configuration file.
</blockquote></details>

`without specifying a --break\_glass flag, it would default to false`

It would be an optional parameter that takes three values -- either not specified at all (retain existing value), false, or true.  So it'd need to be a `Boolean` instead of a `boolean` and have `null` as the third state.  We have plenty of other examples in our codebase of optional boolean nomulus parameters, e.g.: 

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java;l=65-75;drc=68750569db14e79fd258b26c40745add84c9dab1

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java;l=330-336;drc=68750569db14e79fd258b26c40745add84c9dab1

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 540 at r2](https://reviewable.io/reviews/google/nomulus/2277#-Nnz42-h7qXk6Z4ZYeWM:-Nnz42-h7qXk6Z4ZYeWN:b-p89gye) ([raw file](https://github.com/google/nomulus/blob/68a5ec8b9520920d67b86502e17c1f406a62fa34/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L540)):*
> ```Java
> 
>   @Test
>   void testSuccess_breaklassFlag_continuesBreakGlassMode() throws Exception {
> ```

nit: missing a "g"

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @gbrodman)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 67 at r1](https://reviewable.io/reviews/google/nomulus/2277#-NnoZC6C85WcDkWgsvUv:-NoO1hL3CfHwKs9n-nom:b-hjxsoj) ([raw file](https://github.com/google/nomulus/blob/50eea914dc7bb898021614fb35119c68553b24c8/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L67)):*

`without specifying a --break\_glass flag, it would default to false`

It would be an optional parameter that takes three values -- either not specified at all (retain existing value), false, or true.  So it'd need to be a `Boolean` instead of a `boolean` and have `null` as the third state.  We have plenty of other examples in our codebase of optional boolean nomulus parameters, e.g.: 

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java;l=65-75;drc=68750569db14e79fd258b26c40745add84c9dab1

https://cs.opensource.google/nomulus/nomulus/+/master:core/src/main/java/google/registry/tools/CreateOrUpdateTldCommand.java;l=330-336;drc=68750569db14e79fd258b26c40745add84c9dab1
</blockquote></details>

Done. Changed this flag to be a Boolean type. This now required most of the changes that were in go/r3pr/2273 to be moved to this PR to add the logic for ending break glass mode. I will update that PR.

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 151 at r4](https://reviewable.io/reviews/google/nomulus/2277#-NoTjG9x2X2AGnhYHNBl:-NoTjG9x2X2AGnhYHNBm:bhs59qj) ([raw file](https://github.com/google/nomulus/blob/e7bb4807a43ead42f577a567f93ebcdc68fe1354/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L151)):*
> ```Java
>     }
>     // Set the new TLD to break glass mode if break glass flag was used
>     if (Boolean.TRUE.equals(breakGlass)) {
> ```

Shouldn't these three lines just be: `Optional.ofNullable(breakGlass).ifPresent(...);`

And by the way, shouldn't setting this new boolean to `false` turn off break glass mode? I "m not seeing where that would actually happen though?

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2277#-:-NoY241S8dA1A7-D1BzN:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @gbrodman)

PR #2279, opened 2024-01-10
Add support for Ubuntu20 on kokoro

ptkach:
We've had 3 blockers for kokoro ubuntu update which are now resolved and we can move forward with kokoro os update:

1. missing dependency - fixed separately in kokoro config
2. DumpGoldenSchemaCommandTest failure - this apparently is not new and has been happening in GCP Cloud Build docker where it'd been simply disabled (See https://cs.opensource.google/nomulus/nomulus/+/master:core/build.gradle;l=49) . I've found a way to make it work, which I will later enable in CB as well after this runs well
3. PathParameterTest failures - Pretty much the same story as 2 - it's been failing on GCP Cloud Build docker where the entire test suite has been disabled. My change disables only 2 tests affected test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2279)

### DIFF
--- a/core/src/nonprod/java/google/registry/tools/DumpGoldenSchemaCommand.java
+++ b/core/src/nonprod/java/google/registry/tools/DumpGoldenSchemaCommand.java
@@ -16,11 +16,9 @@ package google.registry.tools;
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Path;
 import org.flywaydb.core.Flyway;
-import org.testcontainers.containers.BindMode;
 import org.testcontainers.containers.Container;
 
 /**
@@ -35,9 +33,6 @@ public class DumpGoldenSchemaCommand extends PostgresqlCommand {
 
   // The mount point in the container.
   private static final String CONTAINER_MOUNT_POINT = "/tmp/pg_dump.out";
-
-  // Temporary workaround to fix permission issues on certain Linux distro (e. g. Arch Linux).
-  private static final String CONTAINER_MOUNT_POINT_TMP = "/tmp/pg_dump.tmp";
 
   @Parameter(
       names = {"--output", "-o"},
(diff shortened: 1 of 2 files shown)


ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2279)* status: 0 of 2 files reviewed, 3 unresolved discussions

___
*[`core/src/nonprod/java/google/registry/tools/DumpGoldenSchemaCommand.java` line 40 at r1](https://reviewable.io/reviews/google/nomulus/2279#-NnoYtazBF54x8UKByrQ:-NnoYtazBF54x8UKByrR:b4otcai) ([raw file](https://github.com/google/nomulus/blob/265d69051bf97390ad1791082c649f1cd838110e/core/src/nonprod/java/google/registry/tools/DumpGoldenSchemaCommand.java#L40)):*
> ```Java
> 
>   // Temporary workaround to fix permission issues on certain Linux distro (e. g. Arch Linux).
>   private static final String CONTAINER_MOUNT_POINT_TMP = "/tmp/pg_dump.tmp";
> ```

With copying approach I used, instead of mounting, we're getting rid of this "extra" variable and mounting that seem to have caused failures if used from within docker container.
___
*[`core/src/test/java/google/registry/tools/params/PathParameterTest.java` line 110 at r1](https://reviewable.io/reviews/google/nomulus/2279#-NnoZFaM7IoUcKtJKRRN:-NnoZFaM7IoUcKtJKRRO:b-ixds4g) ([raw file](https://github.com/google/nomulus/blob/66cf42eae46f677395ea15720f48ee2cda8c27f7/core/src/test/java/google/registry/tools/params/PathParameterTest.java#L110)):*
> ```Java
>     Path file = Files.createFile(tmpDir.resolve("tmpfile.txt"));
>     Files.setPosixFilePermissions(file, PosixFilePermissions.fromString("-w-------"));
>     // This test doesn't take into account the fact that root user has access to read/write the
> ```

Same as below. [Existing problem](https://cs.opensource.google/nomulus/nomulus/+/master:core/build.gradle;l=57) that doesn't present any easy way to overcome, thus I've at least aimed to minimize affected parts - instead of excluding all tests, this will exclude (or rather not assert) only 2 affected.
___
*[`core/src/test/java/google/registry/tools/params/PathParameterTest.java` line 156 at r1](https://reviewable.io/reviews/google/nomulus/2279#-NnoZp_2Czcbl2qrp1nt:-NnoZp_2Czcbl2qrp1nu:b-34vmyd) ([raw file](https://github.com/google/nomulus/blob/66cf42eae46f677395ea15720f48ee2cda8c27f7/core/src/test/java/google/registry/tools/params/PathParameterTest.java#L156)):*
> ```Java
>     // file even if posix permissions disallow it.
>     // For the environmnent that run strictly under root, we will ignore the test results.
>     if (!isRoot) {
> ```

See above

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2279#-:-Nnob2Tt1UvqSkDOGfrT:bnfp4nl)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @jianglai)

PR #2300, opened 2024-01-24
Don't override existing registrar email address when setting referral email

CydeWeys:
The fallback should only apply on creates, not on updates, otherwise it can override an existing value for the email address when only the referral email should be what's updated.

This fixes a bug introduced back in commit in 0ead4f8d9d19b938a7f959a70df41cafd9b5c966.

BUG= http://b/322026165

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2300)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2300)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #2305, opened 2024-01-25
Make the formatting tasks work with Java 17 (take 2)

jianglai:
We should not assume org.gradle.java.home to exist on kokoro or GCB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2305)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -504,7 +504,13 @@ def createGetBuildSrcDirectDepsTask(outputFileName) {
 rootProject.ext {
   invokeJavaDiffFormatScript = { action ->
     def javaHome = project.findProperty('org.gradle.java.home')
-    println("Running the formatting tool with JDK at $javaHome")
+    def javaBin
+    if (javaHome != null) {
+      javaBin = "$javaHome/bin/java"
+    } else {
+      javaBin = ext.execInBash("which java", rootDir)
+    }
+    println("Running the formatting tool with $javaBin")
     def scriptDir = rootDir.path.endsWith('buildSrc')
             ? "${rootDir}/../java-format"
             : "${rootDir}/java-format"
@@ -515,7 +521,7 @@ rootProject.ext {
     def pythonExe = getPythonExecutable()
 
     return ext.execInBash(
-        "JAVA=${javaHome}/bin/java PYTHON=${pythonExe} ${formatDiffScript} ${action}", "${workingDir}")
+        "JAVA=${javaBin} PYTHON=${pythonExe} ${formatDiffScript} ${action}", "${workingDir}")
   }
 }
 


ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2305)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`build.gradle` line 507 at r2](https://reviewable.io/reviews/google/nomulus/2305#-Np0GtNK0uRUnl8BkyCv:-Np0GtNK0uRUnl8BkyCw:b210qvm) ([raw file](https://github.com/google/nomulus/blob/dfbaf3cd28fdfa63f02efc524fc614cdaebe2bd0/build.gradle#L507)):*
> ```Gradle
>   invokeJavaDiffFormatScript = { action ->
>     def javaHome = project.findProperty('org.gradle.java.home')
>     if (javaHome == null) {
> ```

Hmm, I'm not sure we should define it here. It adds one more place where java17 is hardcoded in some way, which will make any future updates a little more cumbersome. 
In fact I don't think we should even define Java env variable at the bottom in `execInBash`. Can we get rid of it and then just check `which java` to see where the java is located inside the `google-java-format-git-diff.sh` and if it's null then throw? That would make it fail fast and easy to find in the future.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2305)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ptkach)

___
*[`build.gradle` line 507 at r2](https://reviewable.io/reviews/google/nomulus/2305#-Np0GtNK0uRUnl8BkyCv:-Np0OoVE-TzFJA8KDpN7:bvycuuo) ([raw file](https://github.com/google/nomulus/blob/dfbaf3cd28fdfa63f02efc524fc614cdaebe2bd0/build.gradle#L507)):*

Hmm, I'm not sure we should define it here. It adds one more place where java17 is hardcoded in some way, which will make any future updates a little more cumbersome. 
In fact I don't think we should even define Java env variable at the bottom in `execInBash`. Can we get rid of it and then just check `which java` to see where the java is located inside the `google-java-format-git-diff.sh` and if it's null then throw? That would make it fail fast and easy to find in the future.
</blockquote></details>

Well, I guess in this particular instance, we do NOT have to use Java 17 here (as it is just calling the .jar file). The reason it failed before is that the shell script has a hardcoded check on the Java version to be 11 and kokoro is upgraded to use Java 17 by default.

Still, I would recommend everyone to set `org.gradle.java.home` in their home directory as it is the most clean way to specify the Java version for Gradle without interfering with the default Java set up on your machine.

ptkach (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2305)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jianglai)

PR #2314, opened 2024-01-30
Add java changes for createBillingCostTransitions

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2314)

### DIFF
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -470,10 +470,11 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
       })
   Money createBillingCost = DEFAULT_CREATE_BILLING_COST;
 
+  // TODO(sarahbot@): Make this field not null and add a default value once field is populated on
+  // all existing TLDs
   /** A property that transitions to different create billing costs at different times. */
   @JsonDeserialize(using = TimedTransitionPropertyMoneyDeserializer.class)
-  TimedTransitionProperty<Money> createBillingCostTransitions =
-      TimedTransitionProperty.withInitialValue(DEFAULT_CREATE_BILLING_COST);
+  TimedTransitionProperty<Money> createBillingCostTransitions;
 
   /** The one-time billing cost for restoring a domain name from the redemption grace period. */
   @Type(type = JodaMoneyType.TYPE_NAME)
@@ -1157,7 +1158,10 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
       // here to catch cases where we loaded an invalid TimedTransitionProperty from the database
       // and cloned it into a new builder, to block re-building a Tld in an invalid state.
       instance.tldStateTransitions.checkValidity();
-      instance.createBillingCostTransitions.checkValidity();
+      // TODO(sarahbot@): Remove null check when createBillingCostTransitions field is made not-null
+      if (instance.createBillingCostTransitions != null) {
+        instance.createBillingCostTransitions.checkValidity();
+      }
       instance.renewBillingCostTransitions.checkValidity();
       instance.eapFeeSchedule.checkValidity();
       // All costs must be in the expected currency.
--- a/core/src/main/java/google/registry/model/tld/Tld.java
+++ b/core/src/main/java/google/registry/model/tld/Tld.java
@@ -74,6 +74,7 @@ import google.registry.tldconfig.idn.IdnTableEnum;
 import google.registry.util.Idn;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
@@ -458,6 +459,8 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
   @JsonDeserialize(using = CurrencyDeserializer.class)
   CurrencyUnit currency = DEFAULT_CURRENCY;
 
+  // TODO(sarahbot@): Remove this field and make createBillingCostTransitions not-null once all TLDs
+  // are populated with a create cost transition map
   /** The per-year billing cost for registering a new domain name. */
   @Type(type = JodaMoneyType.TYPE_NAME)
   @Columns(
@@ -466,6 +469,11 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
         @Column(name = "create_billing_cost_currency")
       })
   Money createBillingCost = DEFAULT_CREATE_BILLING_COST;
+
+  /** A property that transitions to different create billing costs at different times. */
+  @JsonDeserialize(using = TimedTransitionPropertyMoneyDeserializer.class)
+  TimedTransitionProperty<Money> createBillingCostTransitions =
+      TimedTransitionProperty.withInitialValue(DEFAULT_CREATE_BILLING_COST);
 
   /** The one-time billing cost for restoring a domain name from the redemption grace period. */
   @Type(type = JodaMoneyType.TYPE_NAME)
@@ -674,6 +682,13 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
   @VisibleForTesting
   public Money getCreateBillingCost() {
     return createBillingCost;
+  }
+
+  public ImmutableSortedMap<DateTime, Money> getCreateBillingCostTransitions() {
+    return Objects.requireNonNullElseGet(
+            createBillingCostTransitions,
+            () -> TimedTransitionProperty.withInitialValue(getCreateBillingCost()))
+        .toValueMap();
   }
 
   /**
@@ -959,6 +974,17 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
       return this;
     }
 
+    public Builder setCreateBillingCostTransitions(
+        ImmutableSortedMap<DateTime, Money> createCostsMap) {
+      checkArgumentNotNull(createCostsMap, "Create billing costs map cannot be null");
+      checkArgument(
+          createCostsMap.values().stream().allMatch(Money::isPositiveOrZero),
+          "Create billing cost cannot be negative");
+      getInstance().createBillingCostTransitions =
+          TimedTransitionProperty.fromValueMap(createCostsMap);
+      return this;
+    }
+
     public Builder setReservedListsByName(Set<String> reservedListNames) {
       // TODO(b/309175133): forbid if enrolled with BSA
       checkArgument(reservedListNames != null, "reservedListNames must not be null");
@@ -1131,6 +1157,7 @@ public class Tld extends ImmutableObject implements Buildable, UnsafeSerializabl
       // here to catch cases where we loaded an invalid TimedTransitionProperty from the database
       // and cloned it into a new builder, to block re-building a Tld in an invalid state.
       instance.tldStateTransitions.checkValidity();
+      instance.createBillingCostTransitions.checkValidity();
       instance.renewBillingCostTransitions.checkValidity();
       instance.eapFeeSchedule.checkValidity();
       // All costs must be in the expected currency.
--- a/core/src/main/java/google/registry/tools/ConfigureTldCommand.java
+++ b/core/src/main/java/google/registry/tools/ConfigureTldCommand.java
@@ -42,6 +42,7 @@ import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -155,6 +156,14 @@ public class ConfigureTldCommand extends MutatingCommand {
     checkPremiumList(newTld);
     checkDnsWriters(newTld);
     checkCurrency(newTld);
+    // TODO(sarahbot@): Remove this once the createBillingCost field is removed
+    checkArgument(
+        Objects.equals(
+            TimedTransitionProperty.fromValueMap(newTld.getCreateBillingCostTransitions())
+                .getValueAtTime(DateTime.now()),
+            newTld.getCreateBillingCost()),
+        "The createBillingCostTransitions map must have the same current cost as the"
+            + " createBillingCost field");
     // bsaEnrollStartTime only exists in DB. Need to carry it over to the updated copy. See Tld.java
     // for more information.
     Optional<DateTime> bsaEnrollTime =
--- a/core/src/main/java/google/registry/tools/ConfigureTldCommand.java
+++ b/core/src/main/java/google/registry/tools/ConfigureTldCommand.java
@@ -15,6 +15,7 @@ package google.registry.tools;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static google.registry.model.tld.Tld.Builder.ROID_SUFFIX_PATTERN;
+import static google.registry.model.tld.Tld.DEFAULT_CREATE_BILLING_COST;
 import static google.registry.model.tld.Tlds.getTlds;
 import static google.registry.util.ListNamingUtils.convertFilePathToName;
 
@@ -28,6 +29,7 @@ import com.google.common.collect.ImmutableSortedMap;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Sets.SetView;
 import com.google.common.flogger.FluentLogger;
+import google.registry.model.common.TimedTransitionProperty;
 import google.registry.model.tld.Tld;
 import google.registry.model.tld.label.PremiumList;
 import google.registry.model.tld.label.PremiumListDao;
@@ -114,8 +116,15 @@ public class ConfigureTldCommand extends MutatingCommand {
     String name = convertFilePathToName(inputFile);
     Map<String, Object> tldData = new Yaml().load(Files.newBufferedReader(inputFile));
     checkName(name, tldData);
-    checkForMissingFields(tldData);
     Tld oldTld = getTlds().contains(name) ? Tld.get(name) : null;
+    if (!tldData.containsKey("createBillingCostTransitions")) {
+      tldData.put(
+          "createBillingCostTransitions",
+          oldTld != null
+              ? oldTld.getCreateBillingCostTransitions()
+              : TimedTransitionProperty.withInitialValue(DEFAULT_CREATE_BILLING_COST).toValueMap());
+    }
+    checkForMissingFields(tldData);
     Tld newTld = mapper.readValue(inputFile.toFile(), Tld.class);
     if (oldTld != null) {
       oldTldInBreakGlass = oldTld.getBreakglassMode();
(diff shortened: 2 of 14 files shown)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 476 at r1](https://reviewable.io/reviews/google/nomulus/2314#-NpfX3hv4VG_BwCQpyxB:-NpfX3hv4VG_BwCQpyxC:b-33sexy) ([raw file](https://github.com/google/nomulus/blob/b2fbcc0a0951fb289c4f07ea834f6fd18358a113/core/src/main/java/google/registry/model/tld/Tld.java#L476)):*
> ```Java
>   @JsonDeserialize(using = TimedTransitionPropertyMoneyDeserializer.class)
>   TimedTransitionProperty<Money> createBillingCostTransitions =
>       TimedTransitionProperty.withInitialValue(DEFAULT_CREATE_BILLING_COST);
> ```

this doesn't seem right since it means that any TLD that we load will have the transitions map populated to 8 USD, even TLDs that we've already launched with different amounts

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 476 at r1](https://reviewable.io/reviews/google/nomulus/2314#-NpfX3hv4VG_BwCQpyxB:-NpfinOC-bXHB28_XrFD:btc8u6a) ([raw file](https://github.com/google/nomulus/blob/b2fbcc0a0951fb289c4f07ea834f6fd18358a113/core/src/main/java/google/registry/model/tld/Tld.java#L476)):*

This new field does need to have a default value and it can't be based off of the createBillingCost since as I noted in the TODO, the plan is to remove that field. Yes, this will initialize the field to 8 USD, but once this PR is deployed, we still need to go and populate all the launched TLD's yaml files in the config/ directory with an accurate map, which I plan to write a simple script to construct using the existing createBillingCost value for each launched TLD.
</blockquote></details>

Why does it need to have a default value? The table in SQL is nullable so this can be too. And the getter already assumes that it can be null (and correctly returns the single create billing cost if so). 

It is better to have this have no value (then the getter returns the create cost) than to have the incorrect value.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 476 at r1](https://reviewable.io/reviews/google/nomulus/2314#-NpfX3hv4VG_BwCQpyxB:-NpvG-ss67h4ubUrXzgQ:b-be2s8l) ([raw file](https://github.com/google/nomulus/blob/b2fbcc0a0951fb289c4f07ea834f6fd18358a113/core/src/main/java/google/registry/model/tld/Tld.java#L476)):*

Why does it need to have a default value? The table in SQL is nullable so this can be too. And the getter already assumes that it can be null (and correctly returns the single create billing cost if so). 

It is better to have this have no value (then the getter returns the create cost) than to have the incorrect value.
</blockquote></details>

This is only temporarily nullable (See note in above TODO). I removed the default value for now, but will only have to add the default value back when I make this field not-null once it has been populated in all the YAML files.
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 180 at r2](https://reviewable.io/reviews/google/nomulus/2314#gh-1479035977:-NpvYngy2MEBa6aWI4cx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/9112977d9732a301ef9507165be76a2e7533228c/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L180)):*

## Unread local variable

Variable 'Tld tld' is never read.

[Show more details](https://github.com/google/nomulus/security/code-scanning/1823)
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 7 of 12 files at r1, 4 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: all files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 476 at r1](https://reviewable.io/reviews/google/nomulus/2314#-NpfX3hv4VG_BwCQpyxB:-NpzdNgl6SAOxVv-kZCc:b-6uut8g) ([raw file](https://github.com/google/nomulus/blob/b2fbcc0a0951fb289c4f07ea834f6fd18358a113/core/src/main/java/google/registry/model/tld/Tld.java#L476)):*

This is only temporarily nullable (See note in above TODO). I removed the default value for now, but will only have to add the default value back when I make this field not-null once it has been populated in all the YAML files.
</blockquote></details>

Right yeah it will eventually have values for every row, but until it does, it's better to have a null value (which is what we have here) than the incorrect value (which is what it had earlier in this PR). 

Using `getCreateBillingCost()` as a default in the getter, as you're doing below, is def the right way to do it. 
___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 125 at r3](https://reviewable.io/reviews/google/nomulus/2314#-NpzebbP2uzIoIZB2g7t:-NpzebbP2uzIoIZB2g7u:b-1zjl2r) ([raw file](https://github.com/google/nomulus/blob/bd01996713691689c9e7264b9bd6fdad24a7eef3/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L125)):*
> ```Java
>           oldTld != null
>               ? oldTld.getCreateBillingCostTransitions()
>               : TimedTransitionProperty.withInitialValue(DEFAULT_CREATE_BILLING_COST).toValueMap());
> ```

Similar to the comments from the previous file, we should not use DEFAULT_CREATE_BILLING_COST here. 

If there's no createBillingCostTransitions specified then we should probably assume that the old TLD object didn't have it set either. Either way, with this code as written we could get in a situation where the single createBillingCost would not match the createBillingCostTransitions (if the cost is changed without specifying a transitions map). This would be pretty bad. We should probably have a test for that sort of situation too.

Elsewhere, we enforce that the user has a createBillingCost set, so we should just use that instead of whatever the old TLD provided.
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 176 at r3](https://reviewable.io/reviews/google/nomulus/2314#-Npzk6K22CfMaDx_B16u:-Npzk6K34lxninR7ukmU:b-werg49) ([raw file](https://github.com/google/nomulus/blob/bd01996713691689c9e7264b9bd6fdad24a7eef3/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L176)):*
> ```Java
>     // TLD's YAML will contain the fields in the correct order
>     assertThat(updatedTld.getCreateBillingCost()).isEqualTo(Money.of(USD, 25));
>   }
> ```

Yeah, like I mentioned above, we should test situations where the map and the createBillingCost don't line up. Either because we've changed the billing cost and never bothered with the createBillingCostTransitions map, or some sort of other user error.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 7 of 13 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 125 at r3](https://reviewable.io/reviews/google/nomulus/2314#-NpzebbP2uzIoIZB2g7t:-Nq-Esuv5lUiLbvu5nQV:b-uq2pw4) ([raw file](https://github.com/google/nomulus/blob/bd01996713691689c9e7264b9bd6fdad24a7eef3/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L125)):*

Similar to the comments from the previous file, we should not use DEFAULT_CREATE_BILLING_COST here. 

If there's no createBillingCostTransitions specified then we should probably assume that the old TLD object didn't have it set either. Either way, with this code as written we could get in a situation where the single createBillingCost would not match the createBillingCostTransitions (if the cost is changed without specifying a transitions map). This would be pretty bad. We should probably have a test for that sort of situation too.

Elsewhere, we enforce that the user has a createBillingCost set, so we should just use that instead of whatever the old TLD provided.
</blockquote></details>

This is only an issue in cases where we are creating a new TLD. Since there are no plans to create any new TLDs before removing the createBillingCost field, this is not really an issue. Any situation when a createBillingCostTransitions map is not specified... the oldTld will not be null and the getCreateBillingCost() method will be called.  

We actually don't know yet here that createBillingCost exists and it is not straightforward to construct a TimedTransitionMap from the tldData representation of a Money Object of varying currencies, but rather than further refactoring this command only for the couple weeks that the field is a non-null field I think it's fine to just use the default value. The added check will prevent the values from being different from each other.
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 176 at r3](https://reviewable.io/reviews/google/nomulus/2314#-Npzk6K22CfMaDx_B16u:-Nq-_iGX7f1BLhx5IHdK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/bd01996713691689c9e7264b9bd6fdad24a7eef3/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L176)):*

Yeah, like I mentioned above, we should test situations where the map and the createBillingCost don't line up. Either because we've changed the billing cost and never bothered with the createBillingCostTransitions map, or some sort of other user error.
</blockquote></details>

Done.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 7 of 13 files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 125 at r3](https://reviewable.io/reviews/google/nomulus/2314#-NpzebbP2uzIoIZB2g7t:-Nq36TefBK64BP3KDsqq:b-x7gzbh) ([raw file](https://github.com/google/nomulus/blob/bd01996713691689c9e7264b9bd6fdad24a7eef3/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L125)):*

This is only an issue in cases where we are creating a new TLD. Since there are no plans to create any new TLDs before removing the createBillingCost field, this is not really an issue. Any situation when a createBillingCostTransitions map is not specified... the oldTld will not be null and the getCreateBillingCost() method will be called.  

We actually don't know yet here that createBillingCost exists and it is not straightforward to construct a TimedTransitionMap from the tldData representation of a Money Object of varying currencies, but rather than further refactoring this command only for the couple weeks that the field is a non-null field I think it's fine to just use the default value. The added check will prevent the values from being different from each other.
</blockquote></details>

Here's what I need as a test to illustrate why this code is not correct:
- create a TLD with a non-negligible billing cost transitions map (like, with at least one change in the future) 
- Instantiate an update on said TLD where the new YAML file does not include a createBillingCostTransitions map, but *does* have the same current create cost 
- Check the new transitions map. We'd expect it to be unchanged. It won't be.

you have a test that covers a trivial update, but it's not sufficient because 1. it doesn't cover edge cases 2. it trivially succeeds -- it succeeds because it just nukes the entire map and sets it to the new createBillingCost.

Later on, this code deserializes from a JSON map into a TLD object, and that *does not work* if you use actual DateTime/Money objects. 

And we still should definitely not use the DEFAULT cost anywhere. This command will fail if we don't include the createBillingCost, so just use that. 
___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 163 at r4](https://reviewable.io/reviews/google/nomulus/2314#-Nq36J4x14VQ2fJ_lFij:-Nq36J4x14VQ2fJ_lFik:by49nw1) ([raw file](https://github.com/google/nomulus/blob/f38f28a04bccd326c788ec446ef49c6e98a1c793/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L163)):*
> ```Java
>         Objects.equals(
>             TimedTransitionProperty.fromValueMap(newTld.getCreateBillingCostTransitions())
>                 .getValueAtTime(DateTime.now()),
> ```

it'll be mad at using DateTime.now() here with no argument, either pass in UTC or inject a clock into the command (latter is probably better)

gbrodman (changes requested):
> Reviewed 3 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 9 of 14 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 125 at r3](https://reviewable.io/reviews/google/nomulus/2314#-NpzebbP2uzIoIZB2g7t:-Nq45ap01NHnX53-VkVb:b8haqmj) ([raw file](https://github.com/google/nomulus/blob/bd01996713691689c9e7264b9bd6fdad24a7eef3/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L125)):*

I don't think I agree that is a case we want to handle that way. The Yaml representation of a file should be the defacto representation of a TLD. It should not be possible to add a CreateBillingCostTransitionsMap to a TLD without it being represented in the YAML file. In this situation, it would not be possible since we run an hourly sync of the TLDs with the YAML files, if someone were to manually update a TLD (which they should not) and add a transition map, it would be overwritten by the YAML file that does not contain one and hence would remove it thus replacing it with the constructed map using the TLD's current createBillingCost. 

If there was a CreateBillingCostTransitions map in the YAML file and it is removed, the command should revert to using the constructed transitions map from the CreateBillingCost, not for the transitions map to remain unchanged.

Either way, this code is meant only to be a TEMPORARY place holder so that the current sync jobs continue to work and the Cloud Build job that syncs the YAML files does NOT FAIL while the YAML files do not yet have values in them for the CreateBillingCostTransitions map. Once those values have been added to the files, the field will become a required field and this wont be an edge case that needs to be handled since the command SHOULD FAIL if the field is missing from the file.

As far as the default value, I realized that as far as for the missing fields check, only the key needs to be added to the tldData and not the value, so I changed that here. 

The deserialization works the same as it does for renewBillingCostTransitions using the TimedTransitionPropertyMoneyDeserializer. I added a test to show this.
</blockquote></details>

discussed this offline -- what this PR does now, where it just ignores the transitions field when doing the "are all fields present?" check is definitely the better way of going about this, rather than worrying about trying to set possibly-bogus transitions on an object that are then ignored. 

Could also explicitly ignore that field in `checkForMissingFields` rather than adding it to the object here, but that's kinda six one / half dozen the other. Then you wouldn't have to do any Stream shenanigans. 
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 189 at r5](https://reviewable.io/reviews/google/nomulus/2314#-Nq4ioKiCxiXBOGMuYw_:-Nq4ioKiCxiXBOGMuYwa:b-76ff83) ([raw file](https://github.com/google/nomulus/blob/bb3421e47de38d62ec7d2b1bf764253019d42098/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L189)):*
> ```Java
>         .isEqualTo(ImmutableSortedMap.of(START_OF_TIME, Money.of(USD, 25)));
>   }
> ```

do we have a test for removing the transitions map if, for some reason, it's missing from the yaml file? we probably shouldn't want to do that in practice, but it should still work i think

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 9 of 14 files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 129 at r5](https://reviewable.io/reviews/google/nomulus/2314#-Nq4jKAt0Ut09H5CSuBW:-Nq4jKAt0Ut09H5CSuBX:b7twu6c) ([raw file](https://github.com/google/nomulus/blob/bb3421e47de38d62ec7d2b1bf764253019d42098/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L129)):*
> ```Java
>         Stream.concat(tldData.keySet().stream(), Stream.of("createBillingCostTransitions"))
>             .collect(toImmutableSet()));
>     Tld newTld = mapper.readValue(inputFile.toFile(), Tld.class);
> ```

can we just pass in the tldData object to avoid reading the file in twice?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: 8 of 14 files reviewed, 2 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 129 at r5](https://reviewable.io/reviews/google/nomulus/2314#-Nq4jKAt0Ut09H5CSuBW:-Nq9e0RSEliojOsTAPUQ:bh0i13d) ([raw file](https://github.com/google/nomulus/blob/bb3421e47de38d62ec7d2b1bf764253019d42098/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L129)):*

can we just pass in the tldData object to avoid reading the file in twice?
</blockquote></details>

Nope, unfortunately that won't work since the map is not serializable and will not deserialize properly into a TLD object.
___
*[`core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java` line 189 at r5](https://reviewable.io/reviews/google/nomulus/2314#-Nq4ioKiCxiXBOGMuYw_:-Nq9oaA33bIkT7fRDF-Z:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/bb3421e47de38d62ec7d2b1bf764253019d42098/core/src/test/java/google/registry/tools/ConfigureTldCommandTest.java#L189)):*

do we have a test for removing the transitions map if, for some reason, it's missing from the yaml file? we probably shouldn't want to do that in practice, but it should still work i think
</blockquote></details>

done

gbrodman (approved):
> Reviewed 4 of 6 files at r4, 1 of 4 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2314)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/tools/ConfigureTldCommand.java` line 129 at r5](https://reviewable.io/reviews/google/nomulus/2314#-Nq4jKAt0Ut09H5CSuBW:-NqDXZVz2a238ULBsMEQ:b1niz9x) ([raw file](https://github.com/google/nomulus/blob/bb3421e47de38d62ec7d2b1bf764253019d42098/core/src/main/java/google/registry/tools/ConfigureTldCommand.java#L129)):*

Nope, unfortunately that won't work since the map is not serializable and will not deserialize properly into a TLD object.
</blockquote></details>

womp thanks for trying

PR #2322, opened 2024-02-09
Add batching to DeleteProberDataAction

sarahcaseybot:
The previous implementation of this action using a scrollable query did not actually separate the deletions into separate transactions like it intended to which resulted in serialization failures. Additionally, the actual deletions of the prober domains were failing due to foreign key constraints that did not cascade on delete like we thought they did. Therefore the prober data has not actually been deleted in a long time resulting in ~9 million dead entries in the Domain table. This likely has an effect on our system's overall latency. This new batching implementation actually creates separate batched reads of the domain table and processes the deletions in separate transactions. It adds the additional queries for deletion from the domain's sub-tables. 

I tested this PR in crash where there are currently ~1.8 million domains that need to be deleted by this action and the action succeeded with no serialization failures, however it is still very slowly, only deleting a 1000 domain batch every ~11 minutes. Further investigation should be done to determine why these deletions are taking so long. At the current rate of deletion (which will likely be even slower on production since there is much more data in that environment) the action will be killed before it is able to successfully delete all the prober data. We should increase the frequency of how often this action is run in order to eventually delete all the built up old data that we currently have in production. 

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2322)


weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 101 at r1](https://reviewable.io/reviews/google/nomulus/2322#-NqYq8EQ3QovEuZaOnwW:-NqYq8EQ3QovEuZaOnwX:bxfbuxw) ([raw file](https://github.com/google/nomulus/blob/93c8565ffccbe3656edc15a92d09d5625553b84d/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L101)):*
> ```Java
>           + " :creationTimeCutoff AND ((d.creationTime <= :nowAutoTimestamp AND d.deletionTime >"
>           + " current_timestamp()) OR d.deletionTime < :nowMinusSoftDeleteDelay) %s ORDER BY"
>           + " d.repoId ASC";
> ```

I don't think the `%s ORDER BY d.repoId ASC` part is necessary.  In fact, the `ORDER BY` clause in the original query is not necessary either.

It doesn't matter in order the domains are processed; and since every processed domain will not be in the result set again, there is no need for
the `repoId > ...` condition. By removing the ORDER BY, the query should run a lot faster when there are many domains to delete.

The only thing that may go wrong is if there is a significant clock drift between Nomulus VMs and Cloud SQL server, to the extent that previous
transactionTime is greater than DB's `current_timestamp()`.  To guard against this, replace `current_timestamp()` in the template to `:now` and
set its value to current transaction's time  (the `now` variable, which also gives value to `:nowAutoTimestamp`.


If performance is an issue, another performance improvement is to use the same `now`, `nowAutoTimestamp` and `nowMinusSoftDeletedDelay` values for
all batched queries. This way, DB can reuse the cached query plan and intermediate results.



_Code quote:_
```Java
          + " current_timestamp()) OR d.deletionTime < :nowMinusSoftDeleteDelay) %s ORDER BY"
          + " d.repoId ASC";
```

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 101 at r1](https://reviewable.io/reviews/google/nomulus/2322#-NqYq8EQ3QovEuZaOnwW:-NqdjzGe-zY6KGKBGkhz:b-p040k1) ([raw file](https://github.com/google/nomulus/blob/93c8565ffccbe3656edc15a92d09d5625553b84d/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L101)):*

We cannot necessarily assume that these domains have already been deleted in later batch result sets due to the dryrun option of this action which does not actually delete the data. Unless... we decide we should remove the dryrun option.
</blockquote></details>

Separated to 2 separate query strings for dry run mode that kept the old query and regular runs that removes the order by and comparison to the previous batch.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 101 at r1](https://reviewable.io/reviews/google/nomulus/2322#-NqYq8EQ3QovEuZaOnwW:-Nqhdl0Q9Cm7Ri39-3B6:b-q7jg9k) ([raw file](https://github.com/google/nomulus/blob/93c8565ffccbe3656edc15a92d09d5625553b84d/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L101)):*

Separated to 2 separate query strings for dry run mode that kept the old query and regular runs that removes the order by and comparison to the previous batch.
</blockquote></details>

Keeping two queries is not ideal. I'd think that a dryrun should stop after one batch. We can't manually examine 1000+ domains.

Also, can you compare the performance with and without ordering when actually deleting domains in crash? Set a limit of say 100K for each run
and compare the processing times.

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 240 at r3](https://reviewable.io/reviews/google/nomulus/2322#-NrbOMlSE-pWwehX514V:-NrbOMlSE-pWwehX514W:b-lhin8o) ([raw file](https://github.com/google/nomulus/blob/4c51e325f2a6820fb22ee413a8d963862d048533/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L240)):*
> ```Java
>     tm().query("DELETE FROM GracePeriod WHERE domainRepoId IN :repoIds")
>         .setParameter("repoIds", domainRepoIds)
>         .executeUpdate();
> ```

Are you sure these are necessary? GracePeriod is part of Domain and should be cascade-deleted by the JPL query `Delete from Domain`. If not, then our JPA annotations may be wrong.

Same question for the other additions: they are part of Domain or DomainHistory.

Also, shouldn't  HostHistory be deleted? 



_Code quote:_
```Java
    tm().query("DELETE FROM GracePeriod WHERE domainRepoId IN :repoIds")
        .setParameter("repoIds", domainRepoIds)
        .executeUpdate();
```

weiminyu (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @sarahcaseybot)

___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 101 at r1](https://reviewable.io/reviews/google/nomulus/2322#-NqYq8EQ3QovEuZaOnwW:-NrgSqrtBokDb1YuWQDI:b-z2gazz) ([raw file](https://github.com/google/nomulus/blob/93c8565ffccbe3656edc15a92d09d5625553b84d/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L101)):*

At this point this PR is very much still a work in progress since I have not even started writing any tests. However when I attempted to actually run this in crash performing actual deletes (before I was only using dry-run mode), I realized that this action did not actually properly delete the data at all and probably never has due to foreign key constraints. I have now modified the hardDeleteDomainsAndHosts() method so that it actually works, however there are a lot of issues. One issue is that there remains lots of database contention. The largest batch size that seems to work without causing serialization issues in my testing on crash is a batch size of 200. But an even larger issue is that the action as it is now seems to only be able to delete domains at a rate of ~1 name per second. This is not sustainable since there are over 8 million domains that need to be deleted in production. 

I'm not sure if it is safe to split up the read and deletes into separate transactions or if that would make this action faster. My thought, though, is that this may be safe to do since most of these names are junk names that are no longer being used.

As far as comparing the query with and without ordering. I saw significantly less serialization failures when using the query without ordering.
</blockquote></details>

Could you try setting the transaction isolation level to REPEATABLE_READ and see if there is a performance improvement?
___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 240 at r3](https://reviewable.io/reviews/google/nomulus/2322#-NrbOMlSE-pWwehX514V:-NrgTMwm1yZiNRYHf6Ic:b-7tnnur) ([raw file](https://github.com/google/nomulus/blob/4c51e325f2a6820fb22ee413a8d963862d048533/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L240)):*

Yes, I was seeing failures related to these foreign key constraints when trying to run this action in crash. See this query for some examples: https://pantheon.corp.google.com/logs/query;query=%22deleteProberData%22%20and%20%22violates%20foreign%20key%20constraint%20%22%0Aresource.type%3D%22gae_app%22%0Aresource.labels.module_id%3D%22backend%22%0A;summaryFields=:false:32:beginning;cursorTimestamp=2024-02-16T22:12:45.831119Z;startTime=2024-02-16T04:00:00.000Z;endTime=2024-02-17T10:20:00.000Z?project=domain-registry-crash&pli=1
</blockquote></details>

Turns out that JPQL queries do not cascade. 

I'd suggest that we make all queries on sub tables (GracePeriod and GracePeriodHistory etc) native to distinguish them from top-level entities.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @CydeWeys and @weiminyu)

___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 101 at r1](https://reviewable.io/reviews/google/nomulus/2322#-NqYq8EQ3QovEuZaOnwW:-Nu5fFfi1vv8APrx8YDo:b-896fix) ([raw file](https://github.com/google/nomulus/blob/93c8565ffccbe3656edc15a92d09d5625553b84d/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L101)):*

Could you try setting the transaction isolation level to REPEATABLE_READ and see if there is a performance improvement?
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/batch/DeleteProberDataAction.java` line 240 at r3](https://reviewable.io/reviews/google/nomulus/2322#-NrbOMlSE-pWwehX514V:-Nu5f9bPC6FxKvs7fou0:b-896fix) ([raw file](https://github.com/google/nomulus/blob/4c51e325f2a6820fb22ee413a8d963862d048533/core/src/main/java/google/registry/batch/DeleteProberDataAction.java#L240)):*

Turns out that JPQL queries do not cascade. 

I'd suggest that we make all queries on sub tables (GracePeriod and GracePeriodHistory etc) native to distinguish them from top-level entities.
</blockquote></details>

Done.

sarahcaseybot (commented):
> Dismissed @github-advanced-security[bot] from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2322)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @CydeWeys and @weiminyu)

PR #2323, opened 2024-02-14
Use standard JVM shutdown hook

jianglai:
This removes a dependency on the App Engine SDK. It also looks like
(from the logs at least) that shutdown hooks registered the old  way stopped
working after the runtime is upgraded to Java 17.

Also removed some random leftover dependencies on the App Engine SKD
that are not needed any more.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2323)


gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2323)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #2326, opened 2024-02-15
Remove Duplicate billing events from the invoicing pipeline

jianglai:
The Distinct transform removes duplicates based on the serialized format
of the elements. By providing a deterministic coder, we can guarantee
that no duplicates exist.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2326)

### DIFF
--- a/core/src/main/java/google/registry/beam/billing/BillingEvent.java
+++ b/core/src/main/java/google/registry/beam/billing/BillingEvent.java
@@ -21,26 +21,21 @@ import google.registry.reporting.billing.BillingModule;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.io.Serializable;
 import java.util.regex.Pattern;
 import org.apache.beam.sdk.coders.AtomicCoder;
 import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.DoubleCoder;
+import org.apache.beam.sdk.coders.NullableCoder;
 import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.coders.VarIntCoder;
+import org.apache.beam.sdk.coders.VarLongCoder;
 import org.joda.time.DateTime;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 
-/**
- * A POJO representing a single billable event, parsed from a {@code SchemaAndRecord}.
- *
- * <p>This is a trivially serializable class that allows Beam to transform the results of a Cloud
- * SQL query into a standard Java representation, giving us the type guarantees and ease of
- * manipulation Cloud SQL lacks.
- */
+/** A POJO representing a single billable event, parsed from a {@code SchemaAndRecord}. */
 @AutoValue
-public abstract class BillingEvent implements Serializable {
-
-  private static final long serialVersionUID = -3593088371541450077L;
+public abstract class BillingEvent {
 
   private static final DateTimeFormatter DATE_TIME_FORMATTER =
       DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss zzz");
@@ -85,7 +80,7 @@ public abstract class BillingEvent implements Serializable {
   /** Returns the tld this event was generated for. */
   abstract String tld();
 
-  /** Returns the billable action this event was generated for (i.e. RENEW, CREATE, TRANSFER...) */
+  /** Returns the billable action this event was generated for (i.e., RENEW, CREATE, TRANSFER...) */
   abstract String action();
 
   /** Returns the fully qualified domain name this event was generated for. */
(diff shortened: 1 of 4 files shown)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2326)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2326)* status: all files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/beam/billing/BillingEvent.java` line 310 at r1](https://reviewable.io/reviews/google/nomulus/2326#-Nqj7y390qvFYbTA1LUM:-Nqj7y390qvFYbTA1LUN:bjgid0a) ([raw file](https://github.com/google/nomulus/blob/8a218493dc856e04ae33ebe24775cdac6e989aca/core/src/main/java/google/registry/beam/billing/BillingEvent.java#L310)):*
> ```Java
>   }
> 
>   static class BillingEventCoder extends AtomicCoder<BillingEvent> {
> ```

Huh, it's weird that doing this so manually is necessary.
___
*[`core/src/main/java/google/registry/beam/billing/InvoicingPipeline.java` line 106 at r1](https://reviewable.io/reviews/google/nomulus/2326#-Nqj7gJ48DouZDAEPzEq:-Nqj7gJ48DouZDAEPzEr:b-bohzrx) ([raw file](https://github.com/google/nomulus/blob/8a218493dc856e04ae33ebe24775cdac6e989aca/core/src/main/java/google/registry/beam/billing/InvoicingPipeline.java#L106)):*
> ```Java
>         .apply("Read BillingEvents from Cloud SQL", read)
>         .apply("Remove null elements", Filter.by(Objects::nonNull))
>         .apply("Remove duplicates", Distinct.create());
> ```

See, I knew it had to be easy enough!
___
*[`core/src/test/java/google/registry/beam/billing/InvoicingPipelineTest.java` line 369 at r1](https://reviewable.io/reviews/google/nomulus/2326#-Nqj7owdD9OmPhQuIg_D:-Nqj7owdD9OmPhQuIg_E:bdiqjij) ([raw file](https://github.com/google/nomulus/blob/8a218493dc856e04ae33ebe24775cdac6e989aca/core/src/test/java/google/registry/beam/billing/InvoicingPipelineTest.java#L369)):*
> ```Java
>             """);
>   }
> ```

I don't see a test that shows duplicates being removed?

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2326)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/beam/billing/BillingEvent.java` line 310 at r1](https://reviewable.io/reviews/google/nomulus/2326#-Nqj7y390qvFYbTA1LUM:-NqnTEGb4e-CqGnXQCIW:b-8ywbn9) ([raw file](https://github.com/google/nomulus/blob/8a218493dc856e04ae33ebe24775cdac6e989aca/core/src/main/java/google/registry/beam/billing/BillingEvent.java#L310)):*

Huh, it's weird that doing this so manually is necessary.
</blockquote></details>

In theory we could take advantage of `AutoValueSchema`, but it only [works with `AutoValue` classes that have its getters named as `getXxx()`](https://stackoverflow.com/questions/62546191/how-do-i-use-an-autovalue-data-type-for-my-pcollection-in-apache-beam). I tried it and it doesn't seem worth it, given the amount of refactoring, and it also makes the code less legible than just using `xxx()` as getters.
___
*[`core/src/test/java/google/registry/beam/billing/InvoicingPipelineTest.java` line 369 at r1](https://reviewable.io/reviews/google/nomulus/2326#-Nqj7owdD9OmPhQuIg_D:-NqnTjAU0TrfcGDBp0JN:b-jr7w6o) ([raw file](https://github.com/google/nomulus/blob/8a218493dc856e04ae33ebe24775cdac6e989aca/core/src/test/java/google/registry/beam/billing/InvoicingPipelineTest.java#L369)):*

I don't see a test that shows duplicates being removed?
</blockquote></details>

It is not easy to write tests for because if we do the de-duplication `readFromCloudSql` function, we need to have SQL return duplicate results for testing. It is not possible as the primary ID is one of the fields tested for equality. I actually separated out the de-dup step to a separate function and wrote test for it, and it worked as expected. But I feel then we are just testing if `Distinct` works, which is outside the responsibility of the user of the Beam framework. Not to mention it adds an additional function just to call `Distinct`, which seems redundant. So I rolled back that part.

Most importantly, if we trust `Distinct` does [what it says it does](https://beam.apache.org/releases/javadoc/current/index.html?org/apache/beam/sdk/transforms/Distinct.html), we only need to make sure that the coder for `BillingEvent` is deterministic, and there is a test for that.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2326)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @CydeWeys)

CydeWeys (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2326)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @jianglai)

PR #2332, opened 2024-02-21
Update Fix NPM IP incorrectly identifies some private IP addresses as public

imhunterand:
The `isPublic()` function in the NPM package `ip` doesn't correctly identify certain private IP addresses in uncommon formats such as `0x7F.1` as private. Instead, it reports them as public by returning `true`. This can lead to security issues such as Server-Side Request Forgery (SSRF) if `isPublic()` is used to protect sensitive code paths when passed user input. Versions 1.1.9 and 2.0.1 fix the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2332)

### DIFF
--- a/console-webapp/package-lock.json
+++ b/console-webapp/package-lock.json
@@ -9678,9 +9678,9 @@
       }
     },
     "node_modules/ip": {
-      "version": "2.0.0",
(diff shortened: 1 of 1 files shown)


ptkach:
It was fixed by dependabot already, closing

PR #2275, opened 2024-01-09
Include a better error message to debug nomulus tool not working

CydeWeys:
Failures to initialize the tool transaction manager seem to often be caused by stale local credentials.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2275)

PR #2342, opened 2024-02-28
Revert "Include a better error message to debug nomulus tool not working (#2275)"

jianglai:
This reverts commit 64f59712758941a35b000e81642966dc7e6473ad.

The catch block is too broad and most of the times the errors caught is
because `command.run()` failed and it had nothing to do with getting
the transaction manager. The `runCommand` method is already wrapped in a try
block that checks for `LoginRequiredException` and gives the appropriate
error message.

We need to re-assess the situation when the next time we encounter a
login issue that did not trigger `LoginRequiredException`. A blanket try
catch block is not the solution and only makes the situation more
confusing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2342)

### DIFF
--- a/core/src/main/java/google/registry/tools/RegistryCli.java
+++ b/core/src/main/java/google/registry/tools/RegistryCli.java
@@ -23,7 +23,6 @@ import com.beust.jcommander.Parameter;
 import com.beust.jcommander.ParameterException;
 import com.beust.jcommander.Parameters;
 import com.beust.jcommander.ParametersDelegate;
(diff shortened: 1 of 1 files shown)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2342)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2350, opened 2024-03-05
Add Java classes for console history objects

gbrodman:
This also creates base classes for the objects contained within the history classes, e.g. RegistrarBase. This is the same way that objects stored in the HistoryEntry subclasses have base classes, e.g. DomainBase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2350)


jianglai (approved):
> Reviewed 79 of 81 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2350)* status: all files reviewed, 15 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/console/ConsoleUpdateHistory.java` line 61 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsQUpCu1h0zx5x56bZz:-NsQUpCu1h0zx5x56b_-:b-9guhfr) ([raw file](https://github.com/google/nomulus/blob/ee3ed11fb4f42a2c1ab24889c9b4aff5d5aa552c/core/src/main/java/google/registry/model/console/ConsoleUpdateHistory.java#L61)):*
> ```Java
> 
>   /** The user that performed the modification. */
>   @JoinColumn(name = "historyActingUser", referencedColumnName = "emailAddress")
> ```

nullable = false?
___
*[`core/src/main/java/google/registry/model/console/ConsoleUpdateHistory.java` line 78 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsQTjdq9Cmmgha-D621:-NsQTjdq9Cmmgha-D622:b-89u5mt) ([raw file](https://github.com/google/nomulus/blob/ee3ed11fb4f42a2c1ab24889c9b4aff5d5aa552c/core/src/main/java/google/registry/model/console/ConsoleUpdateHistory.java#L78)):*
> ```Java
> 
>   /** The time at which the modification was mode. */
>   @Column(nullable = false, name = "historyDateTime")
> ```

It might be a bit late (because the schema change has gone in), but I think we should probably keep consistency across history objects. In EPP history resources we use `historyModificationTime` I believe.  I'll leave it up to you, depending on how much trouble you are willing to go through (at least there's no data in the table yet) and how soon you want to commit these changes. I actually think it is OK to change schema and the code at the same time, as long as you are  not actively using the table. But our presubmits might prevent you from doing that.
___
*[`core/src/main/java/google/registry/model/console/ConsoleUpdateHistory.java` line 88 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsQUUoR5pDjWnRs-yxM:-NsQUUoR5pDjWnRs-yxN:b-8bzswl) ([raw file](https://github.com/google/nomulus/blob/ee3ed11fb4f42a2c1ab24889c9b4aff5d5aa552c/core/src/main/java/google/registry/model/console/ConsoleUpdateHistory.java#L88)):*
> ```Java
>   public long getRevisionId() {
>     // For some reason, Hibernate throws NPE during some initialization phases if we don't deal with
>     // the null case. Setting the id to 0L when it is null should be fine because 0L for primitive
> ```

Is it still true? Or did you just copy-and-pasted this part?
___
*[`core/src/main/java/google/registry/ui/server/RegistrarFormFields.java` line 93 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsPC0gb2mm6_pru2EvY:-NsPC0gc9a4Npi-Z7r56:b-1gtrlm) ([raw file](https://github.com/google/nomulus/blob/c896c022a6c530c055eb81d402dae784b0912d38/core/src/main/java/google/registry/ui/server/RegistrarFormFields.java#L93)):*
> ```Java
>           .build();
> 
>   public static final FormField<String, Registrar.State> STATE_FIELD =
> ```

These removals seem unrelated to the PR?
___
*[`core/src/test/java/google/registry/testing/DatabaseHelper.java` line 1018 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsPpfXO06OSevaTmFbq:-NsPpfXPErLrZw7khq_A:b-toxhk4) ([raw file](https://github.com/google/nomulus/blob/ee3ed11fb4f42a2c1ab24889c9b4aff5d5aa552c/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1018)):*
> ```Java
>   }
> 
>   /** Creates and persists an admin {@link User} with the given email address. */
> ```

if it does not already exist.
___
*[`core/src/test/java/google/registry/testing/DatabaseHelper.java` line 1031 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsPpnsIE4Yd22m_34Z9:-NsPpnsIE4Yd22m_34ZA:brkk0s3) ([raw file](https://github.com/google/nomulus/blob/ee3ed11fb4f42a2c1ab24889c9b4aff5d5aa552c/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1031)):*
> ```Java
>             .build();
>     tm().transact(() -> tm().put(user));
>     return UserDao.loadUser(emailAddress).get();
> ```

Any reason to load it back from the database? Are there any (significant) fields that only get set when the entity is persisted?

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2350)* status: all files reviewed, 10 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/test/java/google/registry/testing/DatabaseHelper.java` line 1031 at r2](https://reviewable.io/reviews/google/nomulus/2350#-NsPpnsIE4Yd22m_34Z9:-NsindCY6G1gpUnc_o_1:bwpmxgq) ([raw file](https://github.com/google/nomulus/blob/ee3ed11fb4f42a2c1ab24889c9b4aff5d5aa552c/core/src/test/java/google/registry/testing/DatabaseHelper.java#L1031)):*

the ID, unfortunately
</blockquote></details>

Ah, of course...

gbrodman:
(this will still fail because the other change has not been deployed)

gbrodman:
@jianglai just checking, are you still good with this? It's been a while since it was in review (due to me going on vacation) so I just wanted to double-check

jianglai:
> @jianglai just checking, are you still good with this? It's been a while since it was in review (due to me going on vacation) so I just wanted to double-check 

Yes, go ahead.

PR #2373, opened 2024-03-14
Remove java.util.Date

jianglai:
There is one remaining instance in JpaTransactionManagerImpl that cannot
be removed because DetachingTypedQuery is implementing TypedQuery, which has
a method that expectred java.util.Date.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2373)

### DIFF
--- a/config/presubmits.py
+++ b/config/presubmits.py
@@ -173,17 +173,24 @@ PRESUBMITS = {
     ):
         "JavaScript files should not include console logging.",
     PresubmitCheck(
-        r"org\.testcontainers\.shaded\.",
+        r".*org\.testcontainers\.shaded.*",
         "java",
         {"/node_modules/"},
     ):
         "Do not use shaded dependencies from testcontainers.",
     PresubmitCheck(
-        r"com\.google\.common\.truth\.Truth8",
+        r".*com\.google\.common\.truth\.Truth8.*",
         "java",
         {"/node_modules/"},
     ):
         "Truth8 is deprecated. Use Truth instead.",
+    PresubmitCheck(
+        r".*java\.util\.Date.*",
+        "java",
+        {"/node_modules/", "JpaTransactionManagerImpl.java"},
+    ):
+        "Do not use java.util.Date. Use classes in java.time package instead.",
+
 }
 
 # Note that this regex only works for one kind of Flyway file.  If we want to
--- a/core/src/main/java/google/registry/flows/certs/CertificateChecker.java
+++ b/core/src/main/java/google/registry/flows/certs/CertificateChecker.java
@@ -31,7 +31,6 @@ import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.security.interfaces.ECPublicKey;
 import java.security.interfaces.RSAPublicKey;
-import java.util.Date;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import org.bouncycastle.jcajce.provider.asymmetric.util.EC5Util;
@@ -106,12 +105,9 @@ public class CertificateChecker {
     // These 2 different instances of PublicKey need to be handled separately since their OIDs are
     // encoded differently. More details on this can be found at
     // https://stackoverflow.com/questions/49895713/how-to-find-the-matching-curve-name-from-an-ecpublickey.
-    if (key instanceof ECPublicKey) {
-      ECPublicKey ecKey = (ECPublicKey) key;
+    if (key instanceof ECPublicKey ecKey) {
       params = EC5Util.convertSpec(ecKey.getParams());
-    } else if (key instanceof org.bouncycastle.jce.interfaces.ECPublicKey) {
-      org.bouncycastle.jce.interfaces.ECPublicKey ecKey =
-          (org.bouncycastle.jce.interfaces.ECPublicKey) key;
+    } else if (key instanceof org.bouncycastle.jce.interfaces.ECPublicKey ecKey) {
       params = ecKey.getParameters();
     } else {
       throw new IllegalArgumentException("Unrecognized instance of PublicKey.");
@@ -148,7 +144,7 @@ public class CertificateChecker {
     if (!violations.isEmpty()) {
       String displayMessages =
           violations.stream()
-              .map(violation -> getViolationDisplayMessage(violation))
+              .map(this::getViolationDisplayMessage)
               .collect(Collectors.joining("\n"));
       throw new InsecureCertificateException(violations, displayMessages);
     }
@@ -162,7 +158,7 @@ public class CertificateChecker {
     ImmutableSet.Builder<CertificateViolation> violations = new ImmutableSet.Builder<>();
 
     // Check if currently in validity period
-    Date now = clock.nowUtc().toDate();
+    DateTime now = clock.nowUtc();
     if (DateTimeComparator.getInstance().compare(certificate.getNotAfter(), now) < 0) {
       violations.add(CertificateViolation.EXPIRED);
     } else if (DateTimeComparator.getInstance().compare(certificate.getNotBefore(), now) > 0) {
@@ -231,13 +227,13 @@ public class CertificateChecker {
       DateTime lastExpiringNotificationSentDate, String certificateStr) {
     X509Certificate certificate = getCertificate(certificateStr);
     DateTime now = clock.nowUtc();
-    // expiration date is one day after lastValidDate
+    // the expiration date is one day after lastValidDate
     DateTime lastValidDate = new DateTime(certificate.getNotAfter());
     if (lastValidDate.isBefore(now)) {
       return false;
     }
     /*
-     * Client should receive a notification if :
+     * Client should receive a notification if:
      *    1) client has never received notification (lastExpiringNotificationSentDate is initially
      *    set to START_OF_TIME) and the certificate has entered the expiring period, OR
      *    2) client has received notification but the interval between now and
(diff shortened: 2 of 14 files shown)


jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

config/presubmits.py
```diff
@@ -173,17 +173,24 @@ def fails(self, file):
...
         "java",
         {"/node_modules/"},
     ):
         "Truth8 is deprecated. Use Truth instead.",
+    PresubmitCheck(
+        r".*java\.util\.Date.*",
+        "java",
+        {"/node_modules/", "JpaTransactionManagerImpl.java"},
+    ):
+        "Do not use java.util.Date. Use classes in Java.date package instead.",
```
CydeWeys, line 192: What is the `Java.date` package?  That doesn't seem correct to me.

core/src/main/java/google/registry/flows/certs/CertificateChecker.java
```diff
@@ -162,7 +158,7 @@ public ImmutableSet<CertificateViolation> checkCertificate(X509Certificate certi
     ImmutableSet.Builder<CertificateViolation> violations = new ImmutableSet.Builder<>();
 
     // Check if currently in validity period
-    Date now = clock.nowUtc().toDate();
+    DateTime now = clock.nowUtc();
```
CydeWeys, line 161: This is a functionality change?  Are you sure this is even correct?

What's the motivation for removing `Date` everywhere anyway?  Sometimes you do legitimately need to process just a date, not a date and an attached time.

core/src/main/java/google/registry/tools/CreateOrUpdateBulkPricingPackageCommand.java
```diff
@@ -56,9 +56,11 @@ abstract class CreateOrUpdateBulkPricingPackageCommand extends MutatingCommand {
   @Nullable
   @Parameter(
       names = "--next_billing_date",
+      converter = DateTimeParameter.class,
+      validateWith = DateTimeParameter.class,
       description =
           "The next date that the bulk pricing package should be billed for its annual fee")
-  Date nextBillingDate;
+  DateTime nextBillingDate;
```
CydeWeys, line 63: This seems like a bad change to me?  Why are we going to make people specify an entire date time on the command line when the only thing that should be passed in is a date?

networking/src/test/java/google/registry/networking/handler/SslInitializerTestUtils.java
```diff
@@ -67,21 +66,26 @@ public static KeyPair getKeyPair() throws Exception {
...
     X500Name subjectDnName = new X500Name("CN=" + hostname);
     BigInteger serialNumber = BigInteger.valueOf(System.currentTimeMillis());
     X500Name issuerDnName = new X500Name(ssc.cert().getIssuerX500Principal().getName());
     ContentSigner sigGen = new JcaContentSignerBuilder("SHA256WithRSAEncryption").build(ssc.key());
     X509v3CertificateBuilder v3CertGen =
         new JcaX509v3CertificateBuilder(
-            issuerDnName, serialNumber, from, to, subjectDnName, keyPair.getPublic());
+            issuerDnName,
+            serialNumber,
+            from.toDate(),
```
CydeWeys, line 85: This is literally still using a `java.util.Date` then, only it's purposely being packaged into a DateTime, and then stripped out of it, for no reason!

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[`config/presubmits.py` line 192 at r2](https://reviewable.io/reviews/google/nomulus/2373#gh-1525338603:-NsyHN5F-kxf43TrOCme:brukr1b) ([raw file](https://github.com/google/nomulus/blob/c845f74cb70536fbcfa7225f6d36566edb26727a/config/presubmits.py#L192)):*

What is the `Java.date` package?  That doesn't seem correct to me.
</blockquote></details>

My bad. Should be `java.time`.
___
*[`core/src/main/java/google/registry/flows/certs/CertificateChecker.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/2373#gh-1525340727:-NsyHzjt2S8KbhUfIgZd:b-c65d2m) ([raw file](https://github.com/google/nomulus/blob/c845f74cb70536fbcfa7225f6d36566edb26727a/core/src/main/java/google/registry/flows/certs/CertificateChecker.java#L161)):*

This is a functionality change?  Are you sure this is even correct?

What's the motivation for removing `Date` everywhere anyway?  Sometimes you do legitimately need to process just a date, not a date and an attached time.
</blockquote></details>

`java.util.Date` is not just a date. It's [an instant in time with millisecond precision](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html). There are many problems with it(no timezone support, no calendar support, etc) and most of its methods are deprecated for decades (since JDK 1.1). That's the whole reason why people prefer Joda Time instead of Java.util.Date.
___
*[`core/src/main/java/google/registry/tools/CreateOrUpdateBulkPricingPackageCommand.java` line 63 at r2](https://reviewable.io/reviews/google/nomulus/2373#gh-1525341435:-NsyJFRh86XoJ5T2-SDe:bcqkxmm) ([raw file](https://github.com/google/nomulus/blob/c845f74cb70536fbcfa7225f6d36566edb26727a/core/src/main/java/google/registry/tools/CreateOrUpdateBulkPricingPackageCommand.java#L63)):*

This seems like a bad change to me?  Why are we going to make people specify an entire date time on the command line when the only thing that should be passed in is a date?
</blockquote></details>

Even with `Date` people are going to have to specify the time (see the other comment on `Data` being an instant in time). I don't know how jcommander handlers `Date` conversions, it may work if you don't specify one and it gives you some default time, but without timezone support that time is also ambiguous. We have been using `DateTime` in all other commands.

In general `java.util.Date` has such bad APIs that everyone should run from it unless they absolutely have to (like when their dependencies still expect it). Even then one should defer the conversion to `Date` until the last minute,  preferably manipulating Joda Time or `java.time` until they call `toDate()` to convert it and pass it to the dependencies.
___
*[`networking/src/test/java/google/registry/networking/handler/SslInitializerTestUtils.java` line 85 at r2](https://reviewable.io/reviews/google/nomulus/2373#gh-1525343457:-NsyGgg96900SvL6R20D:bwzcsa6) ([raw file](https://github.com/google/nomulus/blob/c845f74cb70536fbcfa7225f6d36566edb26727a/networking/src/test/java/google/registry/networking/handler/SslInitializerTestUtils.java#L85)):*

This is literally still using a `java.util.Date` then, only it's purposely being packaged into a DateTime, and then stripped out of it, for no reason!
</blockquote></details>

We have no control over how our dependencies' interfaces are defined, but we can avoid directly interacting with the deprecated `java.util.date` API as much as possible.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

CydeWeys (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/flows/certs/CertificateChecker.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/2373#gh-1525340727:-Nsyj9jxDouWqszTzkFO:b-pq4eat) ([raw file](https://github.com/google/nomulus/blob/c845f74cb70536fbcfa7225f6d36566edb26727a/core/src/main/java/google/registry/flows/certs/CertificateChecker.java#L161)):*

`java.util.Date` is not just a date. It's [an instant in time with millisecond precision](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html). There are many problems with it(no timezone support, no calendar support, etc) and most of its methods are deprecated for decades (since JDK 1.1). That's the whole reason why people prefer Joda Time instead of Java.util.Date.
</blockquote></details>

How about https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDate.html then?  That looks to do what we need it to do here, but without that baggage.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

___
*[`core/src/main/java/google/registry/flows/certs/CertificateChecker.java` line 161 at r2](https://reviewable.io/reviews/google/nomulus/2373#gh-1525340727:-NszDDjpB8Yfoj3dMznQ:b98ja26) ([raw file](https://github.com/google/nomulus/blob/c845f74cb70536fbcfa7225f6d36566edb26727a/core/src/main/java/google/registry/flows/certs/CertificateChecker.java#L161)):*

How about https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDate.html then?  That looks to do what we need it to do here, but without that baggage.
</blockquote></details>

In this case, we definitely want both a date and a time (as it pertains to checking the expiration date of certificates or CRLs).

In general, the recommendation is to use `java.time` for greenfield projects and `joda.time` for existing project. We have a mixture of both at the moment, though I think it might be a good idea to pick a lane and stick to it to avoid unnecessary conversions. Joda actually recommends actively migrating to `java.time`, but we have a lot of `joda.time` and the migration might not be worth the effort.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2373)* status: all files reviewed, 4 unresolved discussions (waiting on @CydeWeys)

PR #2391, opened 2024-03-27
Add console favicon

ptkach:
I took the one we use in the "old" console

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2391)


jianglai (approved):
lol I don't even know what it is (like a badge?). I think we might be better off with the Nomulus logo or just the Google "G". But we can always change it later.

PR #2398, opened 2024-04-10
Upgrade jcommander

jianglai:
Removed escaped quotes in tests. I don't know why they are necessary in the first place because the shell would have stripped the quotes anyway if the arguments are supplied from the stdin. The old version of jcommander probably does an implicit "unquoting" to extract the quoted string, but now it's just treating the quotes literally when supplied in tests. e. g. if one of the input token is `"--flag=\"foo bar\""`, the old version would assign `foo bar` to `flag`, but the new version assigns `"foo bar"` to it.

TESTED=run the jar locally, quoted strings are parsed as expected.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2398)


jianglai (commented):
> Reviewed 38 of 38 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2398)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (approved):
> Reviewed 38 of 38 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2398)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

PR #2399, opened 2024-04-10
Run BSA Validate without lock

weiminyu:
As a read-only action that tolerates inconsistency due to recent changes, locking is unnecessary. This should help with the lock contention we are observing.

Also reduces the number of VM instances provisioned for BSA and increase the idle timeout. This should reduce invocation delay. Longer delay may cause Cloud Scheduler to report a cron job as failure on its dashboard even though the job succeeds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2399)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2399)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @weiminyu)

PR #2403, opened 2024-04-16
Bind console users to the appropriate IAP roles upon creation

jianglai:
Console users need IAP to inject the necessary OIDC tokens into their
request headers and therefore need to be bound to appropriate roles.

Note that in environments managed by latchkey, changes made by the
nomulus tool are NOT being reverted by latchkey. It seems the IAP role
is not (exclusively) managed by latchkey, even though it looks like a
regular role (and there are indeed bindings for this role specified in
the latchkey config file).

TESTED=ran the nomulus command against alpha and verified that the
bindings are created/removed upon console user creation/deletion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2403)


jianglai (commented):
> Reviewed 19 of 21 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2403)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

gbrodman (changes requested):
> Reviewed 18 of 21 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2403)* status: all files reviewed, 4 unresolved discussions (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/tools/CreateUserCommand.java` line 29 at r3](https://reviewable.io/reviews/google/nomulus/2403#-Nvm_MRiCWSmyzgfTQVM:-Nvm_MRiCWSmyzgfTQVN:b-vdcrl3) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/CreateUserCommand.java#L29)):*
> ```Java
> public class CreateUserCommand extends CreateOrUpdateUserCommand {
> 
>   static final String IAP_SECURED_WEB_APP_USER_ROLE = "roles/iap.httpsResourceAccessor";
> ```

I think possibly better to move this into the IamClient class because it's used in multiple places? it can live near MEMBER_FORMAT 
___
*[`core/src/main/java/google/registry/tools/IamClient.java` line 48 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmcNBP0e48A_yxeoXc:-NvmcNBP0e48A_yxeoXd:bnf5ejc) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/IamClient.java#L48)):*
> ```Java
>                 credentialsBundle.getJsonFactory(),
>                 credentialsBundle.getHttpRequestInitializer())
>             .setApplicationName("nomulus")
> ```

What's the significance of the application name here? I'm guessing we want "domain-registry" (the project ID) instead of "nomulus" based off what we do elsewhere 
___
*[`core/src/main/java/google/registry/tools/IamClient.java` line 78 at r3](https://reviewable.io/reviews/google/nomulus/2403#-Nvmd_CgD8OEL3pthxtM:-Nvmd_CgD8OEL3pthxtN:binnvpz) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/IamClient.java#L78)):*
> ```Java
>                 .getBindings()
>                 .add(new Binding().setRole(role).setMembers(ImmutableList.of(member))));
>     setPolicy(policy);
> ```

seems like this still makes an unnecessary SetIamPolicyRequest in the case where the role in question already contains the user  (same note for removeBinding below)
___
*[`core/src/test/java/google/registry/tools/IamClientTest.java` line 66 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmhGZeEpyNvjSGo0PQ:-NvmhGZeEpyNvjSGo0PR:b46ar4j) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/test/java/google/registry/tools/IamClientTest.java#L66)):*
> ```Java
>   }
> 
>   void verifyResponse() throws Exception {
> ```

I think the name of this method is not quite precise.  It's verifying that a SetIamPolicyRequest was sent (and the contents of that), not the response received from the CloudResourceManager

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2403)* status: all files reviewed, 5 unresolved discussions (waiting on @jianglai and @ptkach)

___
*[`core/src/main/java/google/registry/tools/CreateUserCommand.java` line 28 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmnjKz2bc20W05IBBN:-NvmnjKz2bc20W05IBBO:bliswdx) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/CreateUserCommand.java#L28)):*
> ```Java
> @Parameters(separators = " =", commandDescription = "Update a user account")
> public class CreateUserCommand extends CreateOrUpdateUserCommand {
> ```

not directly related to this PR, but can we make the email field non-null in CreateOrUpdateUserCommand? It should be required

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2403)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/tools/CreateUserCommand.java` line 28 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmnjKz2bc20W05IBBN:-Nw4wOXBAggh8ozaL0gC:b-896fix) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/CreateUserCommand.java#L28)):*

not directly related to this PR, but can we make the email field non-null in CreateOrUpdateUserCommand? It should be required
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/tools/CreateUserCommand.java` line 29 at r3](https://reviewable.io/reviews/google/nomulus/2403#-Nvm_MRiCWSmyzgfTQVM:-Nw4wPe5CJgD77-kJ6G-:b-g7u613) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/CreateUserCommand.java#L29)):*

I think possibly better to move this into the IamClient class because it's used in multiple places? it can live near MEMBER_FORMAT
</blockquote></details>

I think `IamClient` should be general. The role used here is specific to this command and should belong to this class in my opinion.
___
*[`core/src/main/java/google/registry/tools/IamClient.java` line 48 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmcNBP0e48A_yxeoXc:-Nw4wh4R-BdGNg3IJ1u7:bgfryd5) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/IamClient.java#L48)):*

What's the significance of the application name here? I'm guessing we want "domain-registry" (the project ID) instead of "nomulus" based off what we do elsewhere
</blockquote></details>

It doesn't actually do much, maybe for logging purpose. If I don't set it, it will log a warning. I guess making it consistent with other usages makes sense.
___
*[`core/src/main/java/google/registry/tools/IamClient.java` line 78 at r3](https://reviewable.io/reviews/google/nomulus/2403#-Nvmd_CgD8OEL3pthxtM:-Nw4x9roB2p7cpKpGV2J:bbvdi48) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/IamClient.java#L78)):*

seems like this still makes an unnecessary SetIamPolicyRequest in the case where the role in question already contains the user  (same note for removeBinding below)
</blockquote></details>

Yes it does, but as far as I can tell the server will just discard it as a no-op. If we want to handle the special case we'd have to add more conditionals and maybe use a boolean to keep track of if the call is unnecessary. Doesn't seem to be worth the complication for little performance gain.
___
*[`core/src/test/java/google/registry/tools/IamClientTest.java` line 66 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmhGZeEpyNvjSGo0PQ:-Nw4yhA18YWD4f93DZhO:b-896fix) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/test/java/google/registry/tools/IamClientTest.java#L66)):*

I think the name of this method is not quite precise.  It's verifying that a SetIamPolicyRequest was sent (and the contents of that), not the response received from the CloudResourceManager
</blockquote></details>

Done.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2403)* status: all files reviewed, 5 unresolved discussions (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/tools/IamClient.java` line 48 at r3](https://reviewable.io/reviews/google/nomulus/2403#-NvmcNBP0e48A_yxeoXc:-Nw4ypxG1feFlG_h0F37:b-assufo) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/IamClient.java#L48)):*

It doesn't actually do much, maybe for logging purpose. If I don't set it, it will log a warning. I guess making it consistent with other usages makes sense.
</blockquote></details>

This is also how the example works, BTW:

https://cloud.google.com/iam/docs/write-policy-client-libraries#using\_the\_client\_library

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2403)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman and @ptkach)

___
*[`core/src/main/java/google/registry/tools/IamClient.java` line 78 at r3](https://reviewable.io/reviews/google/nomulus/2403#-Nvmd_CgD8OEL3pthxtM:-NwFChND4GHRMiJBKrIH:b-2k83tn) ([raw file](https://github.com/google/nomulus/blob/94b658da1c606a50dcdc24e14312e5e86761aeca/core/src/main/java/google/registry/tools/IamClient.java#L78)):*

I don't think we'd need to do anything that complicated. It's pretty similar to what you have to do if you have a Map of Lists but can't use a multimap for some reason 

```
Binding binding = optionalBinding.orElseGet(() -> {
  Binding newBinding = new Binding().setRole(role);
  policy.getBindings().add(newBinding);
  return newBinding;
}
if (findMember(binding, member).isEmpty()) {
  binding.getMembers().add(member);
  setPolicy(policy);
}
```

(and it reduces nesting which is often kinda nice) 

At a minimum, I think this is significantly easier to do in the deletion case (we can just move the setPolicy call into the innermost ifPresent block )
</blockquote></details>

I agree. This does look much more legible than the nested `ifPresent()` calls.

PR #2404, opened 2024-04-17
Handle bad header names in registrar sheet syncing action

CydeWeys:
The existing behavior was to ignore bad header names, in a way that was counter-intuitive as a user of the Google Sheet. If a header name was bad (which could just be someone accidentally changing it not realizing it needs to correspond exactly to the name of the field on the Java object), then all of the data in that column was just silently left as-is and never updated. This led to gradually worsening sync and offset shift errors over time.

Now, it will write out an error message into every single cell in the bad column, so it's clear that the column name is wrong and does not correspond to any actual data in the DB.

BUG=http://b/332336068

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2404)

### DIFF
--- a/core/src/main/java/google/registry/export/sheet/SheetSynchronizer.java
+++ b/core/src/main/java/google/registry/export/sheet/SheetSynchronizer.java
@@ -38,6 +38,8 @@ class SheetSynchronizer {
 
   private static final String SHEET_NAME = "Registrars";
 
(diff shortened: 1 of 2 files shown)


sarahcaseybot (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2404)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @CydeWeys)

PR #2405, opened 2024-04-17
Make createBillingCostTransitions not null

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2405)


gbrodman (changes requested):
Have you checked all environments to make sure that this is non-null everywhere? 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2405#-:-Nvmo7QQCuO5rHT4iUVq:b-s9o465)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2405#-NvmnylJBhOeBQLF11OZ:-NvmnylKCpjMYIMoiXzB:bn6lyqv) ([raw file](https://github.com/google/nomulus/blob/640c2e5c2780e55313ea6b41352887aa8ac1ca90/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*
> ```Java
>   CurrencyUnit currency = DEFAULT_CURRENCY;
> 
>   // TODO(sarahbot@): Remove this field
> ```

can we remove this field now? if not, can we note when we can (in the comment)?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2405)* status: 6 of 26 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @JsonIgnore)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2405#-NvmnylJBhOeBQLF11OZ:-NvsKoC688v3UaJXlFph:b-ojquoa) ([raw file](https://github.com/google/nomulus/blob/640c2e5c2780e55313ea6b41352887aa8ac1ca90/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*

can we remove this field now? if not, can we note when we can (in the comment)?
</blockquote></details>

We can't remove it yet because its still on all the config files in gerrit. I added the @JsonIgnore annotation so we can begin removing the field from the config files. Once those are removed, we can remove the field.

gbrodman (changes requested):
> Reviewed 18 of 20 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2405)* status: 24 of 26 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`core/src/test/java/google/registry/model/tld/TldTest.java` line 151 at r2](https://reviewable.io/reviews/google/nomulus/2405#-Nw5mNNW6LeeMM1_JbkL:-Nw5mNNW6LeeMM1_JbkM:b-j3thwl) ([raw file](https://github.com/google/nomulus/blob/5591ace7763db1c4cb8d6d8c1249f04fb651d46b/core/src/test/java/google/registry/model/tld/TldTest.java#L151)):*
> ```Java
>             .setDnsAPlusAaaaTtl(Duration.standardHours(1))
>             .setDnsWriters(ImmutableSet.of("baz", "bang"))
>             .setCreateBillingCost(Money.of(USD, 8))
> ```

Why change these? It seems like we're going back to setting the deprecated field which seems like it's moving backward, not forward.

gbrodman (approved):
> Reviewed 1 of 20 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2405)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

___
*[`core/src/test/java/google/registry/model/tld/TldTest.java` line 151 at r2](https://reviewable.io/reviews/google/nomulus/2405#-Nw5mNNW6LeeMM1_JbkL:-NwBAEJZ7qlb5zofcuZp:byisli2) ([raw file](https://github.com/google/nomulus/blob/5591ace7763db1c4cb8d6d8c1249f04fb651d46b/core/src/test/java/google/registry/model/tld/TldTest.java#L151)):*

This is because `DatabaseHelper.createTld` sets the createBillingCost to $13 however, when a TLD is constructed from a YAML file containing no createBillingCost field, it will use the default which is $8. I could change the Database helper to set it to the default, or prevent the builder from assigning it the default value, but that would break a lot of other tests and require a lot more refactoring in this PR that would only be removed soon anyways once we remove the createBillingCost field from Tld.java after the field has been removed from all the yaml files in gerrit.
</blockquote></details>

discussed offline, this is correct but it is annoying that we have to do it haha

PR #2407, opened 2024-04-19
Fix conflicts between locks

weiminyu:
Use REPEATABLE READ for lock acquire/release operations to avoid conlicts between locks.

Postgresql uses table scan on small tables, causing false sharing at the SERIALIZABLE isolation level.

See b/333537928 for details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2407)

### DIFF
--- a/core/src/main/java/google/registry/model/server/Lock.java
+++ b/core/src/main/java/google/registry/model/server/Lock.java
@@ -15,6 +15,7 @@
 package google.registry.model.server;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static google.registry.persistence.PersistenceModule.TransactionIsolationLevel.TRANSACTION_REPEATABLE_READ;
 import static google.registry.persistence.transaction.TransactionManagerFactory.tm;
 import static google.registry.util.DateTimeUtils.isAtOrAfter;
 import static google.registry.util.PreconditionsUtils.checkArgumentNotNull;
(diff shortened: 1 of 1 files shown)


CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2407)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

___
*[`core/src/main/java/google/registry/model/server/Lock.java` line 208 at r1](https://reviewable.io/reviews/google/nomulus/2407#-NvrYZdk9kGeROnlHYEV:-NvrYZdk9kGeROnlHYEW:b6k2jml) ([raw file](https://github.com/google/nomulus/blob/ed9ef31f4dc3da54c1b62994bd3fc062e1842929/core/src/main/java/google/registry/model/server/Lock.java#L208)):*
> ```Java
>      *
>      * <p>Using REPEATABLE READ is safe since Lock acquire/release only accesses one row. Note that
>      * passing the isolation level to the `transact` method requires that it is not a nested
> ```

It's probably not 100% necessary to mention all this, as this overload of `transact()` is itself already enforcing that it cannot be called from within an outer transaction.

CydeWeys (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2407#-:-NvrYlCX3UraVntZeSYc:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jianglai and @weiminyu)

weiminyu (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2407)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`core/src/main/java/google/registry/model/server/Lock.java` line 208 at r1](https://reviewable.io/reviews/google/nomulus/2407#-NvrYZdk9kGeROnlHYEV:-NvrZ2Vl9TV3CUod3ZRn:b2nhb2c) ([raw file](https://github.com/google/nomulus/blob/ed9ef31f4dc3da54c1b62994bd3fc062e1842929/core/src/main/java/google/registry/model/server/Lock.java#L208)):*

It's probably not 100% necessary to mention all this, as this overload of `transact()` is itself already enforcing that it cannot be called from within an outer transaction.
</blockquote></details>

Ack

PR #2415, opened 2024-04-25
Create a load testing EPP client

sarahcaseybot:
A lot of this code came from what was used for a past EPP load testing client written by @jianglai that can be found in Google3 at https://source.corp.google.com/piper///depot/google3/experimental/users/jianglai/proxy/java/google/registry/proxy/client/

I modified the old client to get it working again, to be open-source friendly and use Gradle.

For now, this only performs a login and logout command, I will further expand on this in later PRs to add other EPP commands so that we can truly load test the system.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2415)


jianglai (changes requested):
> Reviewed 5 of 11 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: all files reviewed, 13 unresolved discussions (waiting on @sarahcaseybot)

___
*[`build.gradle` line 307 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-Jk5H3o9G06moVSRD:-Nx-Jk5H3o9G06moVSRE:b-oz7n9k) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/build.gradle#L307)):*
> ```Gradle
>   afterEvaluate {
>     if (rootProject.enableDependencyLocking.toBoolean()
>         && (project.name != 'integration' && project.name != 'load-testing')) {
> ```

Any reason we don't want to do dependency looking for the `load-testing` project?
___
*[`config/presubmits.py` line 103 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-JseD6OuFv9nlUw6b:-Nx-JseD6OuFv9nlUw6c:b-k2a0er) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/config/presubmits.py#L103)):*
> ```Python
>         "Source files must end in a newline.",
> 
>     # System.(out|err).println should only appear in tools/ or load-testing/
> ```

I think we can probably use loggers in the load testing client? There is no need to directly use `System.(out|err)`.
___
*[`load-testing/build.gradle` line 36 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-KMcG9zqarjnxXQ7p:-Nx-KMcG9zqarjnxXQ7q:b9z3h7j) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/build.gradle#L36)):*
> ```Gradle
> 
> distZip.shouldRunAfter(build)
> ```

Not sure why this is necessary? I think if you use `createUberJar` (mentioned below), you should be able to skip this task.
___
*[`load-testing/build.gradle` line 38 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-K5IW1mdPDhB4AxvY:-Nx-K5IW1mdPDhB4AxvZ:bz73xc4) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/build.gradle#L38)):*
> ```Gradle
> 
> jar {
>     manifest {
> ```

You can use the [`createUberJar`](https://cs.opensource.google/search?q=createuberjar&ss=nomulus%2Fnomulus) function to create tasks that build the uber jar.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 75 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxD4ciY4x5RMJpOYWar:-NxD4ciY4x5RMJpOYWas:b52jjna) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L75)):*
> ```Java
> /** A simple EPP client that can be used for load testing. */
> @Parameters(separators = " =")
> @SuppressWarnings("FutureReturnValueIgnored")
> ```

We probably don't need this annotation (and all the unused return values) anymore. There were there because google3 has a linter that forces you to explicitly deal with return values.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 183 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxE0pTe0_x1dmHbtdzc:-NxE0pTe0_x1dmHbtdzd:bw4d8f5) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L183)):*
> ```Java
>   }
> 
>   static byte[] readBytesFromFile(String filename) {
> ```

Is this function used anywhere?

Also, all the static functions here can be `private`, right?
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 279 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEONP597w4keC8nSJt:-NxEONP597w4keC8nSJu:bvs593g) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L279)):*
> ```Java
>   public void run() {
>     String outputFolder = DateTime.now(UTC).toString();
>     int numberOfThreads = Runtime.getRuntime().availableProcessors();
> ```

In retrospect, it seems ill-advised to use **all** available cores for log-writing. We are probably better off using a thread pool here with a modest fixed number of cores here. We want to dedicate as many cores to network IO as possible (which `NioEventLoopGroup` does if you don't specify the thread pool size, I believe).
___
*[`load-testing/src/main/java/google/registry/client/EppClientHandler.java` line 82 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxENssi5Alw0wUQiKP8:-NxENssi5Alw0wUQiKP9:by3txix) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClientHandler.java#L82)):*
> ```Java
>     ctx.channel().attr(REQUEST_SENT).get().add(now);
>     ctx.channel().attr(LOGGING_REQUEST_COMPLETE).set(ctx.executor().newPromise());
>     if (ctx.channel().attr(LOGGING_REQUEST_COMPLETE).get() == null) {
> ```

I don't recall why I did this. Do you know...? We just set this channel attribute in the line above. It doesn't seem possible for it to be `null` now. Also if it is `null`, how can we call `setSuccess()` on it...?
___
*[`load-testing/src/main/java/google/registry/client/ProxyHeaderHandler.java` line 32 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxELzhN7cpyR47q94uj:-NxELzhN7cpyR47q94uk:b-wu3ax5) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/ProxyHeaderHandler.java#L32)):*
> ```Java
>  */
> @SuppressWarnings("FutureReturnValueIgnored")
> public class ProxyHeaderHandler extends ChannelInboundHandlerAdapter {
> ```

I don't think this is necessary any more. Have you tried not installing this handler in the channel pipeline?
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 58 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEPUFP-jZwWAsid2ZU:-NxEPUFP-jZwWAsid2ZV:b-r4bp0q) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/README.md#L58)):*
> ```Markdown
>     $ mkdir stage
>     $ cp -r load-testing/build/distributions/load-testing/ stage/
>     $ cp -r jdk-21_linux-x64_bin.tar.gz stage/
> ```

You are probably better off storing these files in GoB and writing Gradle tasks that take care off building an executable jar file that can be used as-is. Even better, maybe you want to create docker images instead of providing jars, which eliminate the need to configure Java on your client machines.

All I'm saying here is that if you want to productionize this client, it's worth it to do the initial heavy-lifting to make it as easy to use as possible, without the need to consult the `README` file and run a bunch of commands from it.
___
*[`load-testing/src/main/java/google/registry/client/run.sh` line 20 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEQPio-Jmti03t86D4:-Nx-MQJv9XLr23o-Ad5g:b49uw15) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/run.sh#L20)):*
> ```Shell
> tar -xvf jdk-21_linux-x64_bin.tar.gz
> 
> jdk-21.0.2/bin/java -jar load-testing/load-testing.jar --host epp.registry-sandbox.google --certificate certificate.pem -k key.pem -ft
> ```

We certainly do not want to leak the endpoints here. Also, having a re-usable shell script that extracts the JDK installation every time certainly looks sub-optimal. You should either just expect Java to exist on the machines (because you have pre-configured the machines), or use a docker image that brings all of its dependencies with it.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: 4 of 11 files reviewed, 13 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`build.gradle` line 307 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-Jk5H3o9G06moVSRD:-NxJxtN-A4p0ixjkHWRZ:b-1y2ax1) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/build.gradle#L307)):*

Any reason we don't want to do dependency looking for the `load-testing` project?
</blockquote></details>

Dependency locking did not work when making a jar how I was doing it previously. Now that I am using createUberJar, I was able to add back dependency locking. 
___
*[`config/presubmits.py` line 103 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-JseD6OuFv9nlUw6b:-NxTxR6r9K9utTqDX12T:b-xtljli) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/config/presubmits.py#L103)):*

I think we can probably use loggers in the load testing client? There is no need to directly use `System.(out|err)`.
</blockquote></details>

Agreed that we can use loggers for the output of the epp commands, but I think we need to print to system output for the summary at the end of the load test, otherwise the summary info would be logged separately on the logs of each individual instance.
___
*[`load-testing/build.gradle` line 36 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-KMcG9zqarjnxXQ7p:-NxJy9jC8tX7JF9o8DaS:bkxdvxg) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/build.gradle#L36)):*

Not sure why this is necessary? I think if you use `createUberJar` (mentioned below), you should be able to skip this task.
</blockquote></details>

Yes, now that I figured out how createUberJar works, I was able to remove this
___
*[`load-testing/build.gradle` line 38 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-K5IW1mdPDhB4AxvY:-NxJyKi-6U7zz82oI6me:b-896fix) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/build.gradle#L38)):*

You can use the [`createUberJar`](https://cs.opensource.google/search?q=createuberjar&ss=nomulus%2Fnomulus) function to create tasks that build the uber jar.
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 343 at r1](https://reviewable.io/reviews/google/nomulus/2415#gh-1579977210:-NxOYZuB28Qzi4pMJ97c:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0b300f0b3d252e8d48fc4bcd99e2dfebd41e27ad/load-testing/src/main/java/google/registry/client/EppClient.java#L343)):*

## Unread local variable

Variable 'Future<?> unusedFuture' is never read.

[Show more details](https://github.com/google/nomulus/security/code-scanning/7862)
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 75 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxD4ciY4x5RMJpOYWar:-NxTgO6F5OeRktjQxjPX:b-3gdwn0) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L75)):*

We probably don't need this annotation (and all the unused return values) anymore. There were there because google3 has a linter that forces you to explicitly deal with return values.
</blockquote></details>

Nope, this was not from Google3. I added this annotation since it is needed for the Gradle build to pass since we have -Werror specified
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 183 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxE0pTe0_x1dmHbtdzc:-NxOUNSHBiO7PTmc71fX:b-o6n6wf) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L183)):*

Is this function used anywhere?

Also, all the static functions here can be `private`, right?
</blockquote></details>

done and removed, this isn't necessary anymore now that I'm reading the key and certificate file from input
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 279 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEONP597w4keC8nSJt:-NxU22Qe70GLFnRAF4SW:b-fg4fdc) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L279)):*

In retrospect, it seems ill-advised to use **all** available cores for log-writing. We are probably better off using a thread pool here with a modest fixed number of cores here. We want to dedicate as many cores to network IO as possible (which `NioEventLoopGroup` does if you don't specify the thread pool size, I believe).
</blockquote></details>

is 5 a good number of threads?
___
*[`load-testing/src/main/java/google/registry/client/EppClientHandler.java` line 82 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxENssi5Alw0wUQiKP8:-NxTjlth9mOMcoEf-zSy:b-b9f5nx) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClientHandler.java#L82)):*

I don't recall why I did this. Do you know...? We just set this channel attribute in the line above. It doesn't seem possible for it to be `null` now. Also if it is `null`, how can we call `setSuccess()` on it...?
</blockquote></details>

hmm, this does look strange. Seems like it's just unnecessary lines, removed.
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 58 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEPUFP-jZwWAsid2ZU:-NxUAUju5FjWCXzSWUXK:btepzce) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/README.md#L58)):*

You are probably better off storing these files in GoB and writing Gradle tasks that take care off building an executable jar file that can be used as-is. Even better, maybe you want to create docker images instead of providing jars, which eliminate the need to configure Java on your client machines.

All I'm saying here is that if you want to productionize this client, it's worth it to do the initial heavy-lifting to make it as easy to use as possible, without the need to consult the `README` file and run a bunch of commands from it.
</blockquote></details>

Moved some of the files to GoB and created gradle tasks to run most of these commands. Updated the readme with the new instructions on how to run this.
___
*[`load-testing/src/main/java/google/registry/client/ProxyHeaderHandler.java` line 53 at r1](https://reviewable.io/reviews/google/nomulus/2415#gh-1579977211:-NxUDoqGAVUBuJzCUI-l:b-896fix) ([raw file](https://github.com/google/nomulus/blob/0b300f0b3d252e8d48fc4bcd99e2dfebd41e27ad/load-testing/src/main/java/google/registry/client/ProxyHeaderHandler.java#L53)):*

## Use of default toString()

Default toString(): ProxyHeaderHandler inherits toString() from Object, and so is not suitable for printing.

[Show more details](https://github.com/google/nomulus/security/code-scanning/7863)
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/ProxyHeaderHandler.java` line 32 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxELzhN7cpyR47q94uj:-NxOwqye4Vz7ms7nVvFd:b-ts98tm) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/ProxyHeaderHandler.java#L32)):*

I don't think this is necessary any more. Have you tried not installing this handler in the channel pipeline?
</blockquote></details>

tried removing and things seem to still work
___
*[`load-testing/src/main/java/google/registry/client/run.sh` line 20 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEQPio-Jmti03t86D4:-NxUAi9eCkaGzho9v0oi:b-apf3uy) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/run.sh#L20)):*

We certainly do not want to leak the endpoints here. Also, having a re-usable shell script that extracts the JDK installation every time certainly looks sub-optimal. You should either just expect Java to exist on the machines (because you have pre-configured the machines), or use a docker image that brings all of its dependencies with it.
</blockquote></details>

moved this script to GoB and moved the jdk extraction into a step in the gradle build so it only needs to be done once

jianglai (changes requested):
> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: all files reviewed, 14 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`build.gradle` line 307 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-Jk5H3o9G06moVSRD:-NxZ1rYK8-S_Oi7MkejF:b-z67nxt) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/build.gradle#L307)):*

Dependency locking did not work when making a jar how I was doing it previously. Now that I am using createUberJar, I was able to add back dependency locking.
</blockquote></details>

You can remove the extra whitespace added and revert this file now.
___
*[`load-testing/build.gradle` line 38 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZacLE7cbV-x3laOnB:-NxZacLF-70dBC74osX2:b-lo8u6k) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L38)):*
> ```Gradle
> 
> task makeStagingDirectory {
>     new File('stage').mkdirs()
> ```

better to make a staging directory under `build`, which will be cleaned up by the `clean` task.
___
*[`load-testing/build.gradle` line 41 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZam2D07HWj5rcrj6C:-NxZam2D07HWj5rcrj6D:b-rf9uak) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L41)):*
> ```Gradle
> }
> 
> task copyJarToStaging(dependsOn: makeStagingDirectory, type: Copy) {
> ```

You don't have to create a task to copy everything, just the include/exclude/filter pattern:

https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Copy.html
___
*[`load-testing/build.gradle` line 63 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZb3oN00VmXKeRsgEN:-NxZb3oODsoBOGAy2Inw:birpq0u) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L63)):*
> ```Gradle
> task copyJavaToStaging(dependsOn: copyKeyToStaging, type: Copy) {
>     from "${rootDir}/load-testing/jdk-21_linux-x64_bin.tar.gz"
>     into "${rootDir}/stage"
> ```

I still don't understand why you'd want to package your own JDK (at the very least a JRE should suffice).

If you look at the other uber jars that we make, we never package JDKs with them. I think it is a reasonable assumption that the runtime will be provided by the environment. If you really want to make your client platform-independent, make it a Docker image. See [`buildToolImage`](https://cs.opensource.google/nomulus/nomulus/+/master:core/build.gradle;l=825?q=build.gradle&ss=nomulus%2Fnomulus).
___
*[`load-testing/build.gradle` line 79 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZeMYlB2eTf4JE70PF:-NxZeMYm3DEQGgpDzcWx:b-74qf7t) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L79)):*
> ```Gradle
> 
> task deployLoadTest(dependsOn: deployStagingToInstances) {
>     doLast {println 'deploying load tests'}
> ```

Why the extra task? It doesn't do anything...?
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 343 at r1](https://reviewable.io/reviews/google/nomulus/2415#gh-1579977210:-NxZkH0TC3g7jhtVrnFF:bxiz3zt) ([raw file](https://github.com/google/nomulus/blob/0b300f0b3d252e8d48fc4bcd99e2dfebd41e27ad/load-testing/src/main/java/google/registry/client/EppClient.java#L343)):*

Done.
</blockquote></details>

Well, that's unfortunate. It looks like GitHub code analysis doesn't understand `unusedFuture`.... I guess a class-level suppression is the way to go if you want to get rid of all these warnings, or you can just ignore GitHub, as this is not a blocking violation.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 75 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxD4ciY4x5RMJpOYWar:-NxZ8_65-y2ca1KgoyaD:butcwzl) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L75)):*

Nope, this was not from Google3. I added this annotation since it is needed for the Gradle build to pass since we have -Werror specified
</blockquote></details>

OK so it's the same error prone rule that checks for this condition. You can just assign the return value to a variable called `unusedFuture` to signal that it is intentionally unused. It's better than having a class-level suppression.

https://errorprone.info/bugpattern/FutureReturnValueIgnored
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 279 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEONP597w4keC8nSJt:-NxZfY-G70tssZX71HYq:b-2hfv38) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L279)):*

is 5 a good number of threads?
</blockquote></details>

I'd imagine a small number like 5 would suffice as writing logs is mostly disk IO-bound. It also depends on the machines that you run on, as you want to leave enough threads to do the network IO work.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 287 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZgJPiAkWSTA1HP4a2:-NxZgJPiAkWSTA1HP4a3:b-hl616c) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/EppClient.java#L287)):*
> ```Java
>               .group(eventLoopGroup)
>               .channel(NioSocketChannel.class)
>               .handler(makeChannelInitializer(outputFolder, ImmutableList.of(loggingExecutors)));
> ```

You still need a list of single thread executors, not a thread pool executor with a certain number of threads. Have you looked at the logs after this change? You should have seen logs from different network activities interwoven with each other.

With a list of single thread executors and each channel getting mapped to the same executor, logging jobs from the same channel submitted to the same executor will be processed in order (as there is only one thread to work).

If you instead of a thread pool, even though network activities are processed in order (and therefore the submission of the corresponding logging jobs is also in order), different threads in the pool might pick up these jobs and write to the same file at the same time.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 316 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZfRZ7B3UTzFP_KmoC:-NxZfRZ7B3UTzFP_KmoD:br7qz2v) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/EppClient.java#L316)):*
> ```Java
>             .closeFuture()
>             .awaitUninterruptibly(
>                 Duration.ofSeconds(
> ```

What's the point of calling`Duraing.ofSeconds().getSeconds()?`
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 51 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZh9TMB6mwkB2qsGsi:-NxZh9TMB6mwkB2qsGsj:b-jg7fza) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/README.md#L51)):*
> ```Markdown
> 
> * Run the load test. Configurations of the load test can be made by configuring 
> this `run.sh` file locally and re-deploying.
> ```

I don't think you have a `run.sh` any more?
___
*[`load-testing/src/main/java/google/registry/client/deploy.sh` line 16 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZcYVEBwEruDmvGwhn:-NxZcYVEBwEruDmvGwho:biahqn) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/deploy.sh#L16)):*
> ```Shell
> # limitations under the License.
> 
> HOSTS=$(gcloud compute instances list | awk '/^loadtest/ { print $5 }')
> ```

The file should be under `load-testing`.
___
*[`load-testing/src/main/java/google/registry/client/deploy.sh` line 17 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZccJw1InPMKZ8LF5T:-NxZccJw1InPMKZ8LF5U:b-sbukd3) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/deploy.sh#L17)):*
> ```Shell
> 
> HOSTS=$(gcloud compute instances list | awk '/^loadtest/ { print $5 }')
> for host in $HOSTS; do ssh $host sudo apt-get -y install rsync; done
> ```

If you know you are going to use a set number of VMs, it is better to have a separate script to set up them **once** (like installing the JDK, installing rsync, etc), and leave this script to just do rsync.

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: all files reviewed, 14 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`load-testing/src/main/java/google/registry/client/README.md` line 51 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZh9TMB6mwkB2qsGsi:-NxZlKdr9BP8FThe5loB:b-k28tcv) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/README.md#L51)):*

I don't think you have a `run.sh` any more?
</blockquote></details>

Oh, I see it's in GoB now, you can probably just copy paste the content and redact the password and hostnames here, so the public README is still useful.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: 5 of 15 files reviewed, 14 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`load-testing/build.gradle` line 38 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZacLE7cbV-x3laOnB:-NxnQUkD1XcITMAQX5Tt:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L38)):*

better to make a staging directory under `build`, which will be cleaned up by the `clean` task.
</blockquote></details>

Done.
___
*[`load-testing/build.gradle` line 41 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZam2D07HWj5rcrj6C:-Nxng3wt21DHk0Ev5XjF:b965l14) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L41)):*

You don't have to create a task to copy everything, just the include/exclude/filter pattern:

https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Copy.html
</blockquote></details>

For some reason, the include/exclude pattern did not seem to work when I tried it. I was able to combine the copies into a single task though. Needed to include a specified base directory first https://discuss.gradle.org/t/multiple-copies-in-a-single-copy-task-works-a-bit-weird/23187
___
*[`load-testing/build.gradle` line 63 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZb3oN00VmXKeRsgEN:-NxsZ9woAB17uYyfOq0W:b-kq3j54) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L63)):*

I still don't understand why you'd want to package your own JDK (at the very least a JRE should suffice).

If you look at the other uber jars that we make, we never package JDKs with them. I think it is a reasonable assumption that the runtime will be provided by the environment. If you really want to make your client platform-independent, make it a Docker image. See [`buildToolImage`](https://cs.opensource.google/nomulus/nomulus/+/master:core/build.gradle;l=825?q=build.gradle&ss=nomulus%2Fnomulus).
</blockquote></details>

I modified this as suggested below to instead have a one time run set up script for the instances that configures java 21. 
___
*[`load-testing/build.gradle` line 79 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZeMYlB2eTf4JE70PF:-NxnhCT_1_-7Z5mA8RUi:bidokyg) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L79)):*

Why the extra task? It doesn't do anything...?
</blockquote></details>

I figured this would be a nice "trigger" task to trigger the deployment of the string of tasks, but if that's unusual I'll remove it. - removed
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 343 at r1](https://reviewable.io/reviews/google/nomulus/2415#gh-1579977210:-NxxkQwZBeAKFuNxcd1-:bcnzlyx) ([raw file](https://github.com/google/nomulus/blob/0b300f0b3d252e8d48fc4bcd99e2dfebd41e27ad/load-testing/src/main/java/google/registry/client/EppClient.java#L343)):*

Well, that's unfortunate. It looks like GitHub code analysis doesn't understand `unusedFuture`.... I guess a class-level suppression is the way to go if you want to get rid of all these warnings, or you can just ignore GitHub, as this is not a blocking violation.
</blockquote></details>

I think I would prefer to keep the class level supression to avoid these annoying github warnings in the future
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 75 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxD4ciY4x5RMJpOYWar:-NxxbDGWD6xGJa00JcCE:b-896fix) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L75)):*

OK so it's the same error prone rule that checks for this condition. You can just assign the return value to a variable called `unusedFuture` to signal that it is intentionally unused. It's better than having a class-level suppression.

https://errorprone.info/bugpattern/FutureReturnValueIgnored
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 279 at r2](https://reviewable.io/reviews/google/nomulus/2415#-NxEONP597w4keC8nSJt:-Nxt-z0W8h7798xd7Pnp:b-896fix) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/load-testing/src/main/java/google/registry/client/EppClient.java#L279)):*

I'd imagine a small number like 5 would suffice as writing logs is mostly disk IO-bound. It also depends on the machines that you run on, as you want to leave enough threads to do the network IO work.
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 287 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZgJPiAkWSTA1HP4a2:-NxxfVeP10911WDPCTNe:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/EppClient.java#L287)):*

You still need a list of single thread executors, not a thread pool executor with a certain number of threads. Have you looked at the logs after this change? You should have seen logs from different network activities interwoven with each other.

With a list of single thread executors and each channel getting mapped to the same executor, logging jobs from the same channel submitted to the same executor will be processed in order (as there is only one thread to work).

If you instead of a thread pool, even though network activities are processed in order (and therefore the submission of the corresponding logging jobs is also in order), different threads in the pool might pick up these jobs and write to the same file at the same time.
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 316 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZfRZ7B3UTzFP_KmoC:-NxxibC29MhquWueCtG5:bjeghjp) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/EppClient.java#L316)):*

What's the point of calling`Duraing.ofSeconds().getSeconds()?`
</blockquote></details>

It wasn't calling Duration.ofSecond().getSeconds(), it was calling Duration.ofSeconds(seconds - durution.getSeconds()). However, I changed it to multiply the TIMEOUT_SECONDS by 1000 to make it less confusing. 
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 51 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZh9TMB6mwkB2qsGsi:-NxxmvfPBTNaKag7MDXg:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/README.md#L51)):*

Oh, I see it's in GoB now, you can probably just copy paste the content and redact the password and hostnames here, so the public README is still useful.
</blockquote></details>

Done.
___
*[`build.gradle` line 307 at r2](https://reviewable.io/reviews/google/nomulus/2415#-Nx-Jk5H3o9G06moVSRD:-Nxy9pxN7S9BIbgoY3ov:b-896fix) ([raw file](https://github.com/google/nomulus/blob/46627311047798b4474d3b5fade11a541cf78d35/build.gradle#L307)):*

You can remove the extra whitespace added and revert this file now.
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/deploy.sh` line 16 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZcYVEBwEruDmvGwhn:-NxslrEV5Hr2sJkpvxt2:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/deploy.sh#L16)):*

The file should be under `load-testing`.
</blockquote></details>

Done.
___
*[`load-testing/src/main/java/google/registry/client/deploy.sh` line 17 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZccJw1InPMKZ8LF5T:-NxxUH-h2nGUymFuU44Y:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/deploy.sh#L17)):*

If you know you are going to use a set number of VMs, it is better to have a separate script to set up them **once** (like installing the JDK, installing rsync, etc), and leave this script to just do rsync.
</blockquote></details>

Done.

jianglai (changes requested):
> Reviewed 10 of 10 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: all files reviewed, 14 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`load-testing/build.gradle` line 38 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZacLE7cbV-x3laOnB:-NyLbeq81nHq3luPuDW0:bpo45h) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L38)):*

Done.
</blockquote></details>

I'd use a Gradle variable like `layout.buildDirectory` instead of hardcoding `build`.

https://docs.gradle.org/current/userguide/working_with_files.html
___
*[`load-testing/build.gradle` line 41 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZam2D07HWj5rcrj6C:-NyLcDvX0XREaCzssTpw:bdm1ckp) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L41)):*

For some reason, the include/exclude pattern did not seem to work when I tried it. I was able to combine the copies into a single task though. Needed to include a specified base directory first https://discuss.gradle.org/t/multiple-copies-in-a-single-copy-task-works-a-bit-weird/23187
</blockquote></details>

I think you can just do `{projectDir}`, instead of `${rootDir}/load-testing`, or use the `layout` class.

https://cs.opensource.google/search?q=projectdir&sq=&ss=nomulus%2Fnomulus

By hardcoding the directory name, you are making it harder to change it in the future, if you ever want to.
___
*[`load-testing/build.gradle` line 54 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjfOb9Ly76Sn5izK8:-NyLjfOb9Ly76Sn5izK9:b-7p6iyj) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/build.gradle#L54)):*
> ```Gradle
> task deployLoadTestsToInstances (dependsOn: copyFilesToStaging, type: Exec) {
>     executable "sh"
>     workingDir "${rootDir}/"
> ```

working dir can be the project dir, and you just call `./deploy.sh` from it.
___
*[`load-testing/instanceSetUp.sh` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjyRJA-W2l7BLhEHC:-NyLjyRJA-W2l7BLhEHD:bkjenv1) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/instanceSetUp.sh#L27)):*
> ```Shell
>   do
>     for i in {1..60}; do
>       if ssh $host sudo apt-get -y install rsync; then
> ```

For newly created VMs, you need to run `apt-get update && apt-get upgrade` first to make sure your apt database and the installed software is up-to-date, before calling `apt-get install`.

Also, I'd do everything in one ssh session, instead of connection every time to execute one command. Something like:

```
ssh $host "sudo apt-get -y update && sudo apt-get -y upgrade && sudo apt-get -y install rsync && mkdir -p test-client
```

Is it common to have to retry? What's the reason for the transient failure?
___
*[`load-testing/src/main/java/google/registry/client/EppClient.java` line 316 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZfRZ7B3UTzFP_KmoC:-NyLh4U63MLI2oobTkaf:bpdvf7t) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/src/main/java/google/registry/client/EppClient.java#L316)):*

It wasn't calling Duration.ofSecond().getSeconds(), it was calling Duration.ofSeconds(seconds - durution.getSeconds()). However, I changed it to multiply the TIMEOUT_SECONDS by 1000 to make it less confusing.
</blockquote></details>

Oh, duh.
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 8 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjGuNDjx-W6NbqAko:-NyLjGuNDjx-W6NbqAkp:b-573l1m) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L8)):*
> ```Markdown
> ### Setting up the test instances
> * Download the Java 21 JDK jdk-21_linux-x64_bin.tar.gz (note that if this version 
> is no longer available, you'll need to modify the instanceSetUp.sh script(it 
> ```

Just use apt-get to install OpenJDK 21. If the instance OS is reasonably up-to-date, there is no reason why this won't work. 
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 61 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLlPfr-Eh0K1rwFknC:-NyLlPfsC88f6nq9DI-e:b-ln0g05) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L61)):*
> ```Markdown
>     ```shell
>     $ HOSTS=$(gcloud compute instances list | awk '/^loadtest/ { print $5 }')
>     $ for host in $HOSTS; do ssh $host test-client/run.sh; done
> ```

Since all of your scripts are operating in batch mode (connecting to all the VMs and repeating the same operations on each of them), why is `run.sh` only operating in a single machine? Why not put the loop in `run.sh` as well?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: 10 of 15 files reviewed, 14 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`load-testing/build.gradle` line 38 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZacLE7cbV-x3laOnB:-NyMvflS-B1puD1Y18eQ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L38)):*

I'd use a Gradle variable like `layout.buildDirectory` instead of hardcoding `build`.

https://docs.gradle.org/current/userguide/working_with_files.html
</blockquote></details>

Done.
___
*[`load-testing/build.gradle` line 41 at r3](https://reviewable.io/reviews/google/nomulus/2415#-NxZam2D07HWj5rcrj6C:-NyMvf1iC5FKO4euczrK:b-896fix) ([raw file](https://github.com/google/nomulus/blob/731a28dbccefc459481d27c82a49632e3bd5c28b/load-testing/build.gradle#L41)):*

I think you can just do `{projectDir}`, instead of `${rootDir}/load-testing`, or use the `layout` class.

https://cs.opensource.google/search?q=projectdir&sq=&ss=nomulus%2Fnomulus

By hardcoding the directory name, you are making it harder to change it in the future, if you ever want to.
</blockquote></details>

Done.
___
*[`load-testing/instanceSetUp.sh` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjyRJA-W2l7BLhEHC:-NyN7DRC3W9WC3e0351l:b-2y5an7) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/instanceSetUp.sh#L27)):*

For newly created VMs, you need to run `apt-get update && apt-get upgrade` first to make sure your apt database and the installed software is up-to-date, before calling `apt-get install`.

Also, I'd do everything in one ssh session, instead of connection every time to execute one command. Something like:

```
ssh $host "sudo apt-get -y update && sudo apt-get -y upgrade && sudo apt-get -y install rsync && mkdir -p test-client
```

Is it common to have to retry? What's the reason for the transient failure?
</blockquote></details>

The retry was necessary since the instances created on line 18 take a while, so the ssh was failing when the instances were not done being created. Just adding an && to line 18 was still causing failures since I guess the command returns a success before the instances are actually done being created
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 8 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjGuNDjx-W6NbqAko:-NyRSqfi16nuZlW5zzDg:bi2ptvo) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L8)):*

Just use apt-get to install OpenJDK 21. If the instance OS is reasonably up-to-date, there is no reason why this won't work.
</blockquote></details>

apt-get install does not seem to work, error message: 'Package "openjdk-21-jdk" is not available, but is referred to by another package." Used Wget instead to download from the web and install.
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 61 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLlPfr-Eh0K1rwFknC:-NyRSoOYAulkuVsEAcps:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L61)):*

Since all of your scripts are operating in batch mode (connecting to all the VMs and repeating the same operations on each of them), why is `run.sh` only operating in a single machine? Why not put the loop in `run.sh` as well?
</blockquote></details>

Done.

jianglai (changes requested):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: all files reviewed, 5 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`load-testing/instanceSetUp.sh` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjyRJA-W2l7BLhEHC:-NyW3yyS2ulDiw__Fvn4:b-17dd1t) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/instanceSetUp.sh#L27)):*

The retry was necessary since the instances created on line 18 take a while, so the ssh was failing when the instances were not done being created. Just adding an && to line 18 was still causing failures since I guess the command returns a success before the instances are actually done being created
</blockquote></details>

That's too bad. I guess it's not too big of a deal, given that these scripts are only run once.

Perhaps you can sleep for a while before trying to ssh, if you know it's going to take some time for the machines to be provisioned after the `gcloud` command returns anyway?
___
*[`load-testing/instanceSetUp.sh` line 27 at r5](https://reviewable.io/reviews/google/nomulus/2415#-NyVvZ3Y9IYj0e7HZi_H:-NyVvZ3Y9IYj0e7HZi_I:b-l30fbz) ([raw file](https://github.com/google/nomulus/blob/97f54212b653fdfe74a12423e5bb02034291e548/load-testing/instanceSetUp.sh#L27)):*
> ```Shell
>   do
>     for i in {1..60}; do
>       if ssh $host 'sudo apt-get -y update &&
> ```

Does SSH work for you directly? I tried to create a test VM but it I can't connect to its external IP directly. I had to use `gcloud compute ssh` to connect.
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 8 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjGuNDjx-W6NbqAko:-NyW3BXu4G46vgPIbfTo:bld42eb) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L8)):*

apt-get install does not seem to work, error message: 'Package "openjdk-21-jdk" is not available, but is referred to by another package." Used Wget instead to download from the web and install.
</blockquote></details>

That's because the default OS image is Debian, and JDK 21 is still in Debian testing. If you use Ubuntu, you can install `openjdk-21-jdk`.

`gcloud compute instances create test-ubuntu --machine-type n4-standard-16 --image-family ubuntu-2204-lts --image-project ubuntu-os-cloud --zone us-east4-a`

Also, I'd recommend using a higher-speced machine. g1-small is really slow and running apt-get takes forever. For load testing you'd want to not be limited by your client.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: 11 of 15 files reviewed, 5 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`load-testing/instanceSetUp.sh` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjyRJA-W2l7BLhEHC:-NyX5ljXAtFJOk8apTvu:bj0f9ko) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/instanceSetUp.sh#L27)):*

That's too bad. I guess it's not too big of a deal, given that these scripts are only run once.

Perhaps you can sleep for a while before trying to ssh, if you know it's going to take some time for the machines to be provisioned after the `gcloud` command returns anyway?
</blockquote></details>

I don't think it's a good idea to rely on sleep since it could variable how long the instances take to be created. I think retry is the best option here. Added a comment for why retry is needed 
___
*[`load-testing/instanceSetUp.sh` line 27 at r5](https://reviewable.io/reviews/google/nomulus/2415#-NyVvZ3Y9IYj0e7HZi_H:-NyWTkPU3srRwkjJBM8Y:b-eyc086) ([raw file](https://github.com/google/nomulus/blob/97f54212b653fdfe74a12423e5bb02034291e548/load-testing/instanceSetUp.sh#L27)):*

Does SSH work for you directly? I tried to create a test VM but it I can't connect to its external IP directly. I had to use `gcloud compute ssh` to connect.
</blockquote></details>

ssh works fine for me
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 8 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjGuNDjx-W6NbqAko:-NyWzFpB-xxXmCh2J4XI:bor30ka) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L8)):*

That's because the default OS image is Debian, and JDK 21 is still in Debian testing. If you use Ubuntu, you can install `openjdk-21-jdk`.

`gcloud compute instances create test-ubuntu --machine-type n4-standard-16 --image-family ubuntu-2204-lts --image-project ubuntu-os-cloud --zone us-east4-a`

Also, I'd recommend using a higher-speced machine. g1-small is really slow and running apt-get takes forever. For load testing you'd want to not be limited by your client.
</blockquote></details>

When I tried changing the machine type, I was hit with this error saying I had exceeded the CPU limit. 

ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - Quota 'CPUS_PER_VM_FAMILY' exceeded.  Limit: 24.0 in region us-east4.
	metric name = compute.googleapis.com/cpus_per_vm_family
	limit name = CPUS-PER-VM-FAMILY-per-project-region
	limit = 24.0
	dimensions = region: us-east4
vm_family: N4
Try your request in another zone, or view documentation on how to increase quotas: https://cloud.google.com/compute/quotas.


Since the most pertinent issues we have been experiencing in production show a qps of only 30 requests/second. Hoepefully for now we can get enough information from these smaller instances. Later on, it looks like there may be a process for increasing our quota, or I was able to successfully create one VM, it failed on the second VM, so perhaps I could create multiple VMs each in a different region.

jianglai (approved):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: all files reviewed, 5 unresolved discussions (waiting on @github-advanced-security[bot] and @sarahcaseybot)

___
*[`load-testing/instanceSetUp.sh` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjyRJA-W2l7BLhEHC:-Ny_fTkw9DfQR7DiN9Au:b-1elfh0) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/instanceSetUp.sh#L27)):*

I don't think it's a good idea to rely on sleep since it could variable how long the instances take to be created. I think retry is the best option here. Added a comment for why retry is needed
</blockquote></details>

Oh, I meant to sleep before trying ssh, in addition to the retry. If you know it's going to take some time for the machines to be ready, there's no point trying it right away.
___
*[`load-testing/instanceSetUp.sh` line 27 at r5](https://reviewable.io/reviews/google/nomulus/2415#-NyVvZ3Y9IYj0e7HZi_H:-Ny_gJHp7AmHVEILwOvH:b6we5fk) ([raw file](https://github.com/google/nomulus/blob/97f54212b653fdfe74a12423e5bb02034291e548/load-testing/instanceSetUp.sh#L27)):*

ssh works fine for me
</blockquote></details>

It's likely that you have something configured already. There's no reason why it would work directly without configuring the keys or other authentication mechanisms. Can you try on a different machine, perhaps your cloudtop?
___
*[`load-testing/src/main/java/google/registry/client/README.md` line 8 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjGuNDjx-W6NbqAko:-Ny_hBHV9cXf2fiGAhNN:b-t3sgoh) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/src/main/java/google/registry/client/README.md#L8)):*

When I tried changing the machine type, I was hit with this error saying I had exceeded the CPU limit. 

ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - Quota 'CPUS_PER_VM_FAMILY' exceeded.  Limit: 24.0 in region us-east4.
	metric name = compute.googleapis.com/cpus_per_vm_family
	limit name = CPUS-PER-VM-FAMILY-per-project-region
	limit = 24.0
	dimensions = region: us-east4
vm_family: N4
Try your request in another zone, or view documentation on how to increase quotas: https://cloud.google.com/compute/quotas.


Since the most pertinent issues we have been experiencing in production show a qps of only 30 requests/second. Hoepefully for now we can get enough information from these smaller instances. Later on, it looks like there may be a process for increasing our quota, or I was able to successfully create one VM, it failed on the second VM, so perhaps I could create multiple VMs each in a different region.
</blockquote></details>

These annoying quotas are per-zone, not per machine. So it's in your interest to get machines with better CPUs that have more processing power per core. But I agree that 30 QPS is not a tall order.

The process to request a quota increase is quite opaque. You fill in a form and it either gets auto approved or rejected. In order to appeal you have to contact your "sales representative", which we don't have. I think your best bet is to use more zones, should you need more power than one zone can provide.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2415)* status: 13 of 15 files reviewed, 5 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`load-testing/instanceSetUp.sh` line 27 at r4](https://reviewable.io/reviews/google/nomulus/2415#-NyLjyRJA-W2l7BLhEHC:-Nyaovzr0pmyFR4ohese:b-896fix) ([raw file](https://github.com/google/nomulus/blob/e42f7a9b20968fc6a2c890d265960c68e922b7f6/load-testing/instanceSetUp.sh#L27)):*

Oh, I meant to sleep before trying ssh, in addition to the retry. If you know it's going to take some time for the machines to be ready, there's no point trying it right away.
</blockquote></details>

Done.
___
*[`load-testing/instanceSetUp.sh` line 27 at r5](https://reviewable.io/reviews/google/nomulus/2415#-NyVvZ3Y9IYj0e7HZi_H:-Nyb1jH6EuZOdBTsCmuD:bx7k1j5) ([raw file](https://github.com/google/nomulus/blob/97f54212b653fdfe74a12423e5bb02034291e548/load-testing/instanceSetUp.sh#L27)):*

It's likely that you have something configured already. There's no reason why it would work directly without configuring the keys or other authentication mechanisms. Can you try on a different machine, perhaps your cloudtop?
</blockquote></details>

Oh, now I remember, I needed to add my public key here: https://pantheon.corp.google.com/compute/metadata?resourceTab=sshkeys&project=domain-registry-sandbox&scopeTab=projectMetadata

added this to the readme

PR #2418, opened 2024-04-29
Add registryLockEmailAddress field to User object

gbrodman:
We've added the field in the database in a previous PR. This is only used in the old console for now because the new console does not have registry lock functionality yet

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2418)


gbrodman:
nb: this will fail the tests until this week's deployment finishes

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2418)* status: all files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 201 at r1](https://reviewable.io/reviews/google/nomulus/2418#-NwkyPRk8q2QDrl7INgP:-NwkyPRk8q2QDrl7INgQ:b9shhso) ([raw file](https://github.com/google/nomulus/blob/351c738afbbeae8eba7759373f10aaddd0aed7e6/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L201)):*
> ```Java
>         "Incorrect registry lock password for user");
>     return user.getRegistryLockEmailAddress() != null
>         ? user.getRegistryLockEmailAddress()
> ```

Should we do this? Do we want to fallback to the login email address?

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2418)* status: all files reviewed, 1 unresolved discussion (waiting on @jianglai)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 201 at r1](https://reviewable.io/reviews/google/nomulus/2418#-NwkyPRk8q2QDrl7INgP:-Nwos4Fi7xSo4lQeDjQa:b-pzvxf0) ([raw file](https://github.com/google/nomulus/blob/351c738afbbeae8eba7759373f10aaddd0aed7e6/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L201)):*

Should we do this? Do we want to fallback to the login email address?
</blockquote></details>

Ehhhhh yeah so I went back and looked at what we do for the existing RegistrarPoc objects and I'm just going to copy that behavior. Notably, this means that the registry lock email address *must* be set to use registry lock (it can be the same value as the email address if you really want it to be).

jianglai (approved):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2418)* status: all files reviewed, 1 unresolved discussion (waiting on @registry)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 201 at r1](https://reviewable.io/reviews/google/nomulus/2418#-NwkyPRk8q2QDrl7INgP:-Nwp-4L18LGFaBzBgbLk:bm72fz5) ([raw file](https://github.com/google/nomulus/blob/351c738afbbeae8eba7759373f10aaddd0aed7e6/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L201)):*

Ehhhhh yeah so I went back and looked at what we do for the existing RegistrarPoc objects and I'm just going to copy that behavior. Notably, this means that the registry lock email address *must* be set to use registry lock (it can be the same value as the email address if you really want it to be).
</blockquote></details>

My primary concern is that the registrars don't actually check emails sent to their login account (the @registry.google account we created for them). So if we silently send emails to that account, they might miss it.

But I guess if they are initiating a lock/unlock request, the UI is telling them to check for an email (along with the email address returned from this function displayed), so it might not be as big an issue.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2418)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @jianglai and @registry)

___
*[`core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java` line 201 at r1](https://reviewable.io/reviews/google/nomulus/2418#-NwkyPRk8q2QDrl7INgP:-Nwp8LbyEVWxInTPCPVN:bijyoqt) ([raw file](https://github.com/google/nomulus/blob/351c738afbbeae8eba7759373f10aaddd0aed7e6/core/src/main/java/google/registry/ui/server/registrar/RegistryLockPostAction.java#L201)):*

My primary concern is that the registrars don't actually check emails sent to their login account (the @registry.google account we created for them). So if we silently send emails to that account, they might miss it.

But I guess if they are initiating a lock/unlock request, the UI is telling them to check for an email (along with the email address returned from this function displayed), so it might not be as big an issue.
</blockquote></details>

Agreed that it would be bad to send emails to an address that they don't expect. So we should be good here then, if we just fail if they don't have a separate field set 

Rebased which required moving some code around, no functionality changes in the most recent push

jianglai (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2418)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #2421, opened 2024-04-29
Generate fake XSRF token in FakeConsoleApiParams for tests

gbrodman:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2421)

### DIFF
--- a/core/src/test/java/google/registry/testing/FakeConsoleApiParams.java
+++ b/core/src/test/java/google/registry/testing/FakeConsoleApiParams.java
@@ -15,12 +15,14 @@
 package google.registry.testing;
 
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 import com.google.appengine.api.users.UserService;
 import google.registry.request.auth.AuthResult;
 import google.registry.request.auth.UserAuthInfo;
 import google.registry.security.XsrfTokenManager;
 import google.registry.ui.server.registrar.ConsoleApiParams;
+import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletRequest;
 import java.util.Optional;
 import org.joda.time.DateTime;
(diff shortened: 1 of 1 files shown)


ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2421)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/testing/FakeConsoleApiParams.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/2421#-NwfIQ_W0GfdIkgUeeJE:-NwfIQ_W0GfdIkgUeeJF:bqva83x) ([raw file](https://github.com/google/nomulus/blob/69adcffb809f5f7b0126155f510163ce447febb5/core/src/test/java/google/registry/testing/FakeConsoleApiParams.java#L45)):*
> ```Java
>         new XsrfTokenManager(
>             new FakeClock(DateTime.parse("2020-02-02T01:23:45Z")), mock(UserService.class));
>     when(request.getCookies())
> ```

I know it's easier and cleaner to have it like this, but I don't think it's correct, as it always makes the request legit (signed with token), which prevents us from testing use-cases when it should fail. 
I suggest we updated it to have an overload to allow us have both with cookies and without.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2421)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @ptkach)

___
*[`core/src/test/java/google/registry/testing/FakeConsoleApiParams.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/2421#-NwfIQ_W0GfdIkgUeeJE:-NwfJdTICI_ADTcQZnRs:b-w577il) ([raw file](https://github.com/google/nomulus/blob/69adcffb809f5f7b0126155f510163ce447febb5/core/src/test/java/google/registry/testing/FakeConsoleApiParams.java#L45)):*

I know it's easier and cleaner to have it like this, but I don't think it's correct, as it always makes the request legit (signed with token), which prevents us from testing use-cases when it should fail. 
I suggest we updated it to have an overload to allow us have both with cookies and without.
</blockquote></details>

Hard disagree -- the cases where this should fail should be (and are?) tested separately and explicitly. This class is for generating simple params for use in tests where the more important part is everything else being written / tested.

ptkach (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2421)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/testing/FakeConsoleApiParams.java` line 45 at r1](https://reviewable.io/reviews/google/nomulus/2421#-NwfIQ_W0GfdIkgUeeJE:-Nwk6YZEDlQ-jRNDeiq-:bqy67xj) ([raw file](https://github.com/google/nomulus/blob/69adcffb809f5f7b0126155f510163ce447febb5/core/src/test/java/google/registry/testing/FakeConsoleApiParams.java#L45)):*

Hard disagree -- the cases where this should fail should be (and are?) tested separately and explicitly. This class is for generating simple params for use in tests where the more important part is everything else being written / tested.
</blockquote></details>

Fair enough

PR #2422, opened 2024-04-29
Bring codebase up to more recent Java standards

CydeWeys:
This includes using the new switch format (though IntelliJ does not yet understand patterns including default so those aren't used), multiline strings, replacing some unnecessary type declarations with <>, converting some classes to records, replacing some Guava predicates with native Java code, and some other miscellaneous Code Inspection fixes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2422)


gbrodman (changes requested):
> Reviewed 146 of 156 files at r1, 1 of 3 files at r2, 6 of 7 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 9 unresolved discussions (waiting on @CydeWeys and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 57 at r2](https://reviewable.io/reviews/google/nomulus/2422#-NwfaugGAMgCvYFjIDb7:-NwfaugGAMgCvYFjIDb8:b-146r02) ([raw file](https://github.com/google/nomulus/blob/764e1dce0995a97401eb2b7b43005f34a5122746/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L57)):*
> ```Java
>           case READ_ONLY_REPLICA ->
>               registryPipelineComponent.getReadOnlyReplicaJpaTransactionManager();
>           case REGULAR -> registryPipelineComponent.getJpaTransactionManager();
> ```

if there's only two choices ("READ_ONLY_REPLICA" vs "anything else") then we don't need a switch statement at all 
___
*[`core/src/main/java/google/registry/bsa/persistence/BsaDomainRefresh.java` line 99 at r2](https://reviewable.io/reviews/google/nomulus/2422#-NwfcaBD3brc3N1HQffR:-NwfcaBD3brc3N1HQffS:b-v4mq6c) ([raw file](https://github.com/google/nomulus/blob/764e1dce0995a97401eb2b7b43005f34a5122746/core/src/main/java/google/registry/bsa/persistence/BsaDomainRefresh.java#L99)):*
> ```Java
>       return true;
>     }
>     if (!(o instanceof BsaDomainRefresh that)) {
> ```

is this a special case re: instanceof where you can declare a variable in an if statement that's used later on outside of it? 
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwfmoee0qfstUF9vbn7:b-nt2jcc) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

## Inefficient empty string test

Inefficient comparison to empty string, check for zero length instead.

[Show more details](https://github.com/google/nomulus/security/code-scanning/7866)
</blockquote></details>

it's probably not that much more inefficient tbqh
___
*[`core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java` line 84 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoL5z7QWWv17rweUW:-NwfoL5z7QWWv17rweUX:bkiqyaj) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java#L84)):*
> ```Java
> 
>         // Note that RDE_SSH_CLIENT public / private keys are slightly different than other key
>         // pairs,
> ```

line breaks look a little wonky here
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java` line 117 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoXFEAvvGI9M2-7rY:-NwfoXFEAvvGI9M2-7rZ:b-kq1psj) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java#L117)):*
> ```Java
>         return;
>       }
>       default -> {
> ```

....we can just remove this block right? 
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java` line 139 at r5](https://reviewable.io/reviews/google/nomulus/2422#-Nwfo_JpAciQfNbo1ESB:-Nwfo_JpAciQfNbo1ESC:b-i24r8j) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java#L139)):*
> ```Java
>         return;
>       }
>       default -> {
> ```

we can just remove this block righ t? 
___
*[`core/src/test/java/google/registry/testing/DeterministicStringGenerator.java` line 70 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfpfEIDqK4Hu409ej6:-NwfpfEIDqK4Hu409ej7:b-kydcia) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/test/java/google/registry/testing/DeterministicStringGenerator.java#L70)):*
> ```Java
>       case PREPEND_COUNTER -> String.format("%04d_%s", counter++, password.toString());
>       case DEFAULT -> password.toString();
>       default -> password.toString();
> ```

Do we need the default case given that the only two possible options are already listed? I thought Java was now smart enough to know when you've covered everything (and thus don't need a default) 
___
*[`prober/src/test/java/google/registry/monitoring/blackbox/handler/EppActionHandlerTest.java` line 85 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqMGV2bXk5PFa3W7g:-NwfqMGV2bXk5PFa3W7h:b-snteqg) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/prober/src/test/java/google/registry/monitoring/blackbox/handler/EppActionHandlerTest.java#L85)):*
> ```Java
>         }
>       }
>       case "domainExists" -> {
> ```

out of curiosity do we need the braces here given that it's just a single line? 
___
*[`util/src/main/java/google/registry/util/Concurrent.java` line 72 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqvYV8nXrSm1qS3W4:-NwfqvYV8nXrSm1qS3W5:b-u5tf86) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/util/src/main/java/google/registry/util/Concurrent.java#L72)):*
> ```Java
>     }
>     ;
>     try (ExecutorService executor = newFixedThreadPool(threadCount)) {
> ```

This is technically not the same, since the auto-close for ExecutorService calls shutdown, not shutdownNow. Thoughts?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 9 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 57 at r2](https://reviewable.io/reviews/google/nomulus/2422#-NwfaugGAMgCvYFjIDb7:-NwjuIxx3Ks07xUjpBh6:bt81dso) ([raw file](https://github.com/google/nomulus/blob/764e1dce0995a97401eb2b7b43005f34a5122746/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L57)):*

if there's only two choices ("READ_ONLY_REPLICA" vs "anything else") then we don't need a switch statement at all
</blockquote></details>

It's written this way in case future transaction manager types are added.
___
*[`core/src/main/java/google/registry/bsa/persistence/BsaDomainRefresh.java` line 99 at r2](https://reviewable.io/reviews/google/nomulus/2422#-NwfcaBD3brc3N1HQffR:-NwjuWQh5H8Jcpryc2r7:b3vymcn) ([raw file](https://github.com/google/nomulus/blob/764e1dce0995a97401eb2b7b43005f34a5122746/core/src/main/java/google/registry/bsa/persistence/BsaDomainRefresh.java#L99)):*

is this a special case re: instanceof where you can declare a variable in an if statement that's used later on outside of it?
</blockquote></details>

It's a new Java 14 feature called "pattern matching" and it works like this: https://www.baeldung.com/java-pattern-matching-instanceof

It's pretty sick, and solves a long-running Java anti-pattern that I don't like (having to check `instanceof` and then do the exact same cast).
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwjv02l5kCTJr99H1AP:b-6ns7j4) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

it's probably not that much more inefficient tbqh
</blockquote></details>

This CodeQL warning is out of date.  `String.isEmpty()` is indeed efficient, and if you look at its implementation you'll see that all it's doing is checking for `length() == 0`.  I'm using this over raw length because it's clearer from the callsite.

And yes domains don't have many label parts to them so it's all academic anyway.
___
*[`core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java` line 84 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoL5z7QWWv17rweUW:-NwjvrNt3UxUYAUN-9Lz:b-896fix) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java#L84)):*

line breaks look a little wonky here
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java` line 117 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoXFEAvvGI9M2-7rY:-NwjwB8z9nA4gaUGon9C:b-upqj7k) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java#L117)):*

....we can just remove this block right?
</blockquote></details>

I made it throw an exception as this shouldn't be able to happen.
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java` line 139 at r5](https://reviewable.io/reviews/google/nomulus/2422#-Nwfo_JpAciQfNbo1ESB:-Nwjwfz567CGN72IByK-:bs4q9op) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java#L139)):*

we can just remove this block righ t?
</blockquote></details>

Ditto.
___
*[`core/src/test/java/google/registry/testing/DeterministicStringGenerator.java` line 70 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfpfEIDqK4Hu409ej6:-Nwjwjkk3WVW7PdWWmgp:b-8fdqq0) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/test/java/google/registry/testing/DeterministicStringGenerator.java#L70)):*

Do we need the default case given that the only two possible options are already listed? I thought Java was now smart enough to know when you've covered everything (and thus don't need a default)
</blockquote></details>

It was written this way so as to not break with runtime exceptions if a new enum value is added.
___
*[`prober/src/test/java/google/registry/monitoring/blackbox/handler/EppActionHandlerTest.java` line 85 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqMGV2bXk5PFa3W7g:-Nwjx4zz-sEsWvMWqBaV:bdrafjc) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/prober/src/test/java/google/registry/monitoring/blackbox/handler/EppActionHandlerTest.java#L85)):*

out of curiosity do we need the braces here given that it's just a single line?
</blockquote></details>

Yes.  It would work if it were `return switch(...) { ... }` but that's not how it's written.
___
*[`util/src/main/java/google/registry/util/Concurrent.java` line 72 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqvYV8nXrSm1qS3W4:-NwjxBOB0YSiqwDtswKb:b-13j3p6) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/util/src/main/java/google/registry/util/Concurrent.java#L72)):*

This is technically not the same, since the auto-close for ExecutorService calls shutdown, not shutdownNow. Thoughts?
</blockquote></details>

Yeah I looked into it and it's not super clear to me why we would specifically need to call `shutdownNow()` over `shutdown()`.  https://stackoverflow.com/questions/11520189/difference-between-shutdown-and-shutdownnow-of-executor-service

When the `finally` clause runs normally, all of the threads should already be cleared, so it's no different.  The only other way the `finally` can be hit is, I suppose, if an exception is thrown from within a thread, in which case this will terminate all other threads immediately, which I'm not clear is actually desired or expected?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 9 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`util/src/main/java/google/registry/util/Concurrent.java` line 72 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqvYV8nXrSm1qS3W4:-Nwjxt911b4LPMi8JB_j:b-lo73ga) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/util/src/main/java/google/registry/util/Concurrent.java#L72)):*

Yeah I looked into it and it's not super clear to me why we would specifically need to call `shutdownNow()` over `shutdown()`.  https://stackoverflow.com/questions/11520189/difference-between-shutdown-and-shutdownnow-of-executor-service

When the `finally` clause runs normally, all of the threads should already be cleared, so it's no different.  The only other way the `finally` can be hit is, I suppose, if an exception is thrown from within a thread, in which case this will terminate all other threads immediately, which I'm not clear is actually desired or expected?
</blockquote></details>

If we were specifically wanting to shut down all threads on a single thread throwing an exception, I would think you'd do that withing the `catch` block inside the futures for-loop.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 9 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`util/src/main/java/google/registry/util/Concurrent.java` line 72 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqvYV8nXrSm1qS3W4:-NwjyeIH2NW4hD0riqWA:b-kmr79c) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/util/src/main/java/google/registry/util/Concurrent.java#L72)):*

If we were specifically wanting to shut down all threads on a single thread throwing an exception, I would think you'd do that withing the `catch` block inside the futures for-loop.
</blockquote></details>

This was added in http://cl/79858546 and never subsequently altered, fwiw.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 6 unresolved discussions (waiting on @CydeWeys, @gbrodman, and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 57 at r2](https://reviewable.io/reviews/google/nomulus/2422#-NwfaugGAMgCvYFjIDb7:-Nwk464GBp88uqFDBbVk:b5shfp5) ([raw file](https://github.com/google/nomulus/blob/764e1dce0995a97401eb2b7b43005f34a5122746/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L57)):*

It's written this way in case future transaction manager types are added.
</blockquote></details>

We could add a switch statement at that point if it changes the behavior of "READ_ONLY_REPLICA vs everything else"

otherwise, by that logic, shouldn't we use switch statements in every spot where we check whether an enum object has a particular value? 
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwk36mTDtgXDVmnfhd-:b-pkbsht) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

This CodeQL warning is out of date.  `String.isEmpty()` is indeed efficient, and if you look at its implementation you'll see that all it's doing is checking for `length() == 0`.  I'm using this over raw length because it's clearer from the callsite.

And yes domains don't have many label parts to them so it's all academic anyway.
</blockquote></details>

I think its point is that you should use isEmpty() or length() instead the equals() comparison. So it's not out of date. 
___
*[`core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java` line 84 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoL5z7QWWv17rweUW:-Nwk58h_8FE2SJl4yRni:b21f6kq) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java#L84)):*

Done.
</blockquote></details>

no it isn't 
___
*[`core/src/test/java/google/registry/testing/DeterministicStringGenerator.java` line 70 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfpfEIDqK4Hu409ej6:-Nwk5GXf3g_bJfl5NQS6:bvz6bec) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/test/java/google/registry/testing/DeterministicStringGenerator.java#L70)):*

It was written this way so as to not break with runtime exceptions if a new enum value is added.
</blockquote></details>

But if a new enum value is added, we probably want to fix it here 
___
*[`util/src/main/java/google/registry/util/Concurrent.java` line 72 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfqvYV8nXrSm1qS3W4:-Nwk5jOpDnYnzuZc3eOp:b1e8qb2) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/util/src/main/java/google/registry/util/Concurrent.java#L72)):*

This was added in http://cl/79858546 and never subsequently altered, fwiw.
</blockquote></details>

gosh how dare you not remember the exact thought process that went into a CL nine and a half years ago

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 6 unresolved discussions (waiting on @CydeWeys and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java` line 117 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoXFEAvvGI9M2-7rY:-Nwk5BJ8-xE1tgmF7ZGI:b3kre) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/ui/server/registrar/ConsoleOteSetupAction.java#L117)):*

I made it throw an exception as this shouldn't be able to happen.
</blockquote></details>

?
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java` line 139 at r5](https://reviewable.io/reviews/google/nomulus/2422#-Nwfo_JpAciQfNbo1ESB:-Nwk5DEg9rvSOFq4U0XN:b3kre) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/ui/server/registrar/ConsoleRegistrarCreatorAction.java#L139)):*

Ditto.
</blockquote></details>

?

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: 155 of 158 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwk69yBCLZFBUbbo2zs:b-yxk03k) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

I think its point is that you should use isEmpty() or length() instead the equals() comparison. So it's not out of date.
</blockquote></details>

I am using `isEmpty()`.  Look at the code.

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: 155 of 158 files reviewed, 6 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java` line 84 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfoL5z7QWWv17rweUW:-Nwk6HMQ0R5pZuPKeZLD:blcg3i) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/main/java/google/registry/tools/UpdateKeyringSecretCommand.java#L84)):*

no it isn't
</blockquote></details>

I made the changes but I got interrupted by the page before I was able to push them.

gbrodman (commented):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwk6qENA3TCvZoSd85i:btmwihf) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

I am using `isEmpty()`.  Look at the code.
</blockquote></details>

I was looking at the code, I'm not dumb-- the CodeQL comment was when you were using `""::equals` which you fixed at some point, not sure if that's the commit you pushed like a minute ago

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 3 unresolved discussions (waiting on @CydeWeys and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwk87T88gY8z5DGge5P:b-egwv6c) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

I was looking at the code, I'm not dumb-- the CodeQL comment was when you were using `""::equals` which you fixed at some point, not sure if that's the commit you pushed like a minute ago
</blockquote></details>

Reviewable was only showing the old version of the code, even when I loaded the review fresh from the home page

CydeWeys (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 3 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java` line 57 at r2](https://reviewable.io/reviews/google/nomulus/2422#-NwfaugGAMgCvYFjIDb7:-NwkbNO08pH5OPClg8Zy:b235e8g) ([raw file](https://github.com/google/nomulus/blob/764e1dce0995a97401eb2b7b43005f34a5122746/core/src/main/java/google/registry/beam/common/RegistryPipelineWorkerInitializer.java#L57)):*

We could add a switch statement at that point if it changes the behavior of "READ_ONLY_REPLICA vs everything else"

otherwise, by that logic, shouldn't we use switch statements in every spot where we check whether an enum object has a particular value?
</blockquote></details>

I changed it to throw an exception on an unknown transaction manager value so that we're forced to address any added values in the future.

Note that `if (A) { ... } else { ... implicitly assume B ... }` isn't a safe way to code; it tends to break without people realizing it when new types are added in the future.  Better to be explicit with ` if (A) { ... } else if (B) { ... }`
___
*[`core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java` line 208 at r4](https://reviewable.io/reviews/google/nomulus/2422#gh-1583791362:-Nwkc4Lx3w4M5Bb1lGDS:baob1lw) ([raw file](https://github.com/google/nomulus/blob/2fb67341d2a2c0042ad591219782bfe1e0b87a27/core/src/main/java/google/registry/flows/domain/DomainFlowUtils.java#L208)):*

Reviewable was only showing the old version of the code, even when I loaded the review fresh from the home page
</blockquote></details>

You hadn't clicked "Show full diff" at the top.  It's been this way for awhile.
___
*[`core/src/test/java/google/registry/testing/DeterministicStringGenerator.java` line 70 at r5](https://reviewable.io/reviews/google/nomulus/2422#-NwfpfEIDqK4Hu409ej6:-NwkcTPt3V_BKQlSsQXK:b-6tbf5z) ([raw file](https://github.com/google/nomulus/blob/2dc59c271b7568fd8a9f16b3d72f46db5dd80091/core/src/test/java/google/registry/testing/DeterministicStringGenerator.java#L70)):*

But if a new enum value is added, we probably want to fix it here
</blockquote></details>

I made it throw an error.

CydeWeys (commented):
PTAL

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422#-:-NwkeVK2CvycgY5g4ZHo:bb71pro)* status: 153 of 158 files reviewed, 3 unresolved discussions (waiting on @gbrodman and @github-advanced-security[bot])

gbrodman (approved):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

gbrodman (approved):
> Reviewed 10 of 10 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

gbrodman (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

gbrodman (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2422)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

PR #2425, opened 2024-05-01
Remove the createBillingCost field from Tld

sarahcaseybot:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2425)


gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2425#-NwpYR7c-NE498VycISX:-NwpYR7c-NE498VycISY:b-rqptd2) ([raw file](https://github.com/google/nomulus/blob/98855f49407026efa6653aec06464dc1f8bb73b7/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*
> ```Java
>   CurrencyUnit currency = DEFAULT_CURRENCY;
> 
>   // TODO(sarahbot@): Rename this field to createBillingCost once the old createBillingCost columns
> ```

fix the TODO?

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2425#-NwpYR7c-NE498VycISX:-Nwpc_e24dDReF76sbPh:b3h819n) ([raw file](https://github.com/google/nomulus/blob/98855f49407026efa6653aec06464dc1f8bb73b7/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*

I figured I would wait to do this once the column is actually dropped from the database.
</blockquote></details>

can we drop that column, rename the existing column, and change the java name of this field all in one PR later on?

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: all files reviewed, 1 unresolved discussion (waiting on @sarahcaseybot)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2425#-NwpYR7c-NE498VycISX:-Nwpr7pu5Pm-t8Mto9Na:b-s1n1yb) ([raw file](https://github.com/google/nomulus/blob/98855f49407026efa6653aec06464dc1f8bb73b7/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*

We can definitely drop the column for the field being deleted in this PR and rename the new map field in the same future PR. Renaming the new column in the database is complicated I believe since I think it requires actually creating a new column and dropping the old one which would be a 2 step process.
</blockquote></details>

Right yeah renaming in the database is annoying. 

Since this PR is changing the Java code, couldn't we rename the Java field to `createBillingCost` in this PR with the `@Column(name = "createBillingCostTransitions")` that we'll eventually want?

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: 2 of 20 files reviewed, 1 unresolved discussion (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2425#-NwpYR7c-NE498VycISX:-NwzRmGyC0NJbCVCeS_K:bx0fv29) ([raw file](https://github.com/google/nomulus/blob/98855f49407026efa6653aec06464dc1f8bb73b7/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*

Right yeah renaming in the database is annoying. 

Since this PR is changing the Java code, couldn't we rename the Java field to `createBillingCost` in this PR with the `@Column(name = "createBillingCostTransitions")` that we'll eventually want?
</blockquote></details>

Yes, done. needed to add some handling in configureTldCommand for this to still work with the old name but also work when it is changed to the new name.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: 2 of 20 files reviewed, all discussions resolved

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2425#-NwpYR7c-NE498VycISX:-Nx-LS6v5JQ100pA399D:bqqw2cp) ([raw file](https://github.com/google/nomulus/blob/98855f49407026efa6653aec06464dc1f8bb73b7/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*

Yes, done. needed to add some handling in configureTldCommand for this to still work with the old name but also work when it is changed to the new name.
</blockquote></details>

oof yeah this is some awkward handing that we'd have to do here. 

Alternatively, what if we just kept the field named `createBillingCostTransitions`? We already have that for `renewBillingCostTransitions` and `tldStateTransitions` so tbh it kinda makes sense to use the same naming convention. 

We'd just remove the TODO then, in that case.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: 2 of 20 files reviewed, all discussions resolved (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/model/tld/Tld.java` line 461 at r1](https://reviewable.io/reviews/google/nomulus/2425#-NwpYR7c-NE498VycISX:-NxJUtp1AIMX8wsBokYD:bb7f2mf) ([raw file](https://github.com/google/nomulus/blob/98855f49407026efa6653aec06464dc1f8bb73b7/core/src/main/java/google/registry/model/tld/Tld.java#L461)):*

oof yeah this is some awkward handing that we'd have to do here. 

Alternatively, what if we just kept the field named `createBillingCostTransitions`? We already have that for `renewBillingCostTransitions` and `tldStateTransitions` so tbh it kinda makes sense to use the same naming convention. 

We'd just remove the TODO then, in that case.
</blockquote></details>

done

gbrodman (approved):
> Reviewed 1 of 18 files at r2, 17 of 17 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2425)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sarahcaseybot)

PR #2433, opened 2024-05-07
Create node.js console service

ptkach:
This removes the console from being packed with default service war and creates a new node.js service named.

Integrated with local deployment via `./gradlew deploy` as well as spinnaker. Gradlew deployment tested on alpha and qa, spinnaker deployment tested on crash env

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2433)

### DIFF
--- a/appengine_war.gradle
+++ b/appengine_war.gradle
@@ -47,10 +47,6 @@ war {
 
 if (project.path == ":services:default") {
     war {
-        from("${rootDir}/console-webapp/dist/console-webapp") {
-            include "**/*"
-            into("console")
-        }
         from("${coreResourcesDir}/google/registry/ui") {
             include "registrar_bin.js"
             if (environment != "production") {
@@ -103,10 +99,11 @@ explodeWar.doLast {
     file("${it.explodedAppDirectory}/WEB-INF/lib/tools.jar").setWritable(true)
 }
 
+appengineDeployAll.mustRunAfter ':console-webapp:deploy'
 appengineDeployAll.finalizedBy ':deployCloudSchedulerAndQueue'
 rootProject.deploy.dependsOn appengineDeployAll
+
 rootProject.stage.dependsOn appengineStage
-tasks['war'].dependsOn ':console-webapp:buildConsoleWebappProd'
 tasks['war'].dependsOn ':core:compileProdJS'
 tasks['war'].dependsOn ':core:processResources'
 tasks['war'].dependsOn ':core:jar'
--- a/build.gradle
+++ b/build.gradle
@@ -539,8 +539,6 @@ task coreDev {
   dependsOn 'javadoc'
   dependsOn 'checkDependenciesDotGradle'
   dependsOn 'checkLicense'
-  // TODO: @ptkach reenable after console design merged
-  // dependsOn ':console-webapp:runConsoleWebappUnitTests'
   dependsOn ':core:check'
   dependsOn 'assemble'
 }
--- a/console-webapp/.gitignore
+++ b/console-webapp/.gitignore
@@ -41,3 +41,6 @@ testem.log
 # System files
 .DS_Store
 Thumbs.db
+
+# Build artifact
+/staged/dist
--- a/console-webapp/build.gradle
+++ b/console-webapp/build.gradle
@@ -11,12 +11,12 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
 def consoleDir = "${rootDir}/console-webapp"
+def projectParam = "--project=${rootProject.gcpProject}"
 
 clean {
   delete "${consoleDir}/node_modules"
-  delete "${consoleDir}/dist"
+  delete "${consoleDir}/staged/dist"
 }
 
 task npmInstallDeps(type: Exec) {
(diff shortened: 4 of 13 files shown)


jianglai (approved):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2433)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`appengine_war.gradle` line 102 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYxow3DfRhtf1FdOyf:-NxYxow496tAu22avlis:bahn0du) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/appengine_war.gradle#L102)):*
> ```Gradle
> }
> 
> appengineDeployAll.mustRunAfter ':console-webapp:deploy'
> ```

Is there a reason why the app engine deployment must happen after console web deployment? If it's because of `dispatch.xml`, that only needs the service to exist, right? If is is because you want the `deploy` task in the root project to deploy everything, it is better to have `rootProject.deploy` depend on `console-webapp.deploy`.
___
*[`console-webapp/package.json` line 7 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYzUT418319yE0xmqG:-NxYzUT418319yE0xmqH:b-g2zvfh) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/console-webapp/package.json#L7)):*
> ```JSON
>     "ng": "ng",
>     "start": "ng serve --proxy-config dev-proxy.config.json",
>     "build": "ng build --base-href=/console/ --output-path=staged/dist",
> ```

I think it might be a better idea to put the staged assets under the respective `build` folder, so they can be properly cleaned up by `gradle clean` (the same goes for the main console artifacts that are currently stored under `console/dist`, but that's a separate issue). You also don't have to explicitly exclude them in `.gitignore`.
___
*[`console-webapp/staged/app.yaml` line 1 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYzuXg0jUr2_LiMMDV:-NxYzuXg0jUr2_LiMMDW:b-c66ter) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/console-webapp/staged/app.yaml#L1)):*
> ```YAML
> runtime: nodejs20
> ```

I'm not sure why you want to used `staged` the folder name, it makes it sounds like these files are temporary. Wouldn't it better for them to be under something like `node` or `node-server`?

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2433)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @jianglai)

___
*[`appengine_war.gradle` line 102 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYxow3DfRhtf1FdOyf:-NxZ7n6B0C-0RAvHg-eq:bmu587v) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/appengine_war.gradle#L102)):*

Is there a reason why the app engine deployment must happen after console web deployment? If it's because of `dispatch.xml`, that only needs the service to exist, right? If is is because you want the `deploy` task in the root project to deploy everything, it is better to have `rootProject.deploy` depend on `console-webapp.deploy`.
</blockquote></details>

Yeah, this was the only way to make it in the right order. I tried dependsOn but I got a parallel execution which triggered a fail. `mustRunAfter` took care of the order
___
*[`console-webapp/package.json` line 7 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYzUT418319yE0xmqG:-NxZ9U_B-anQSobKMkzp:b52uy4e) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/console-webapp/package.json#L7)):*

I think it might be a better idea to put the staged assets under the respective `build` folder, so they can be properly cleaned up by `gradle clean` (the same goes for the main console artifacts that are currently stored under `console/dist`, but that's a separate issue). You also don't have to explicitly exclude them in `.gitignore`.
</blockquote></details>

Oops, I actually forgot to update a gradle clean task, thanks. The reason it's done this way is so we have one place where GAE deployment-related stuff is, including scripts and yaml. Otherwise I'd need to have extra staging task to prepare it. 
___
*[`console-webapp/staged/app.yaml` line 1 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYzuXg0jUr2_LiMMDV:-NxZ8Nea8sHmra9DKlfx:b8visti) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/console-webapp/staged/app.yaml#L1)):*

I'm not sure why you want to used `staged` the folder name, it makes it sounds like these files are temporary. Wouldn't it better for them to be under something like `node` or `node-server`?
</blockquote></details>

well mainly because in current form with app.yaml, server.js and dist it serves a single purpose - staging for GAE deployment and yes I feel like it's temporary until we migrate to GKE, so for now lets keep it this way. When the GKE time comes I think we will come up with a better approach or naming

jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2433)* status: all files reviewed, 2 unresolved discussions (waiting on @ptkach)

___
*[`appengine_war.gradle` line 102 at r1](https://reviewable.io/reviews/google/nomulus/2433#-NxYxow3DfRhtf1FdOyf:-NxZVAgc47yZwi2JgmTs:bktaii1) ([raw file](https://github.com/google/nomulus/blob/cf75a2dc0c38895e06e3522c41c5e40f5b0045a7/appengine_war.gradle#L102)):*

Yeah, this was the only way to make it in the right order. I tried dependsOn but I got a parallel execution which triggered a fail. `mustRunAfter` took care of the order
</blockquote></details>

I meant that once the service is created, it doesn't have to be deployed **before** the dispatch.xml file is updated, right?
___
*[`console-webapp/build.gradle` line 19 at r2](https://reviewable.io/reviews/google/nomulus/2433#-NxZVpNB6gImyzHyZvwq:-NxZVpNB6gImyzHyZvwr:b-zaaf3c) ([raw file](https://github.com/google/nomulus/blob/f2d5713695d5d0ee60e63fb5ace2a238b438a48f/console-webapp/build.gradle#L19)):*
> ```Gradle
> clean {
>   delete "${consoleDir}/node_modules"
>   delete "${consoleDir}/staged/dist"
> ```

Also needs to update the Jetty job to copy from the updated location:

https://cs.opensource.google/nomulus/nomulus/+/master:jetty/build.gradle;l=62.

PR #2439, opened 2024-05-11
Update jetty's console build path

ptkach:
This is a follow up on a last comment in https://github.com/google/nomulus/pull/2433.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2439)

### DIFF
--- a/jetty/build.gradle
+++ b/jetty/build.gradle
@@ -59,7 +59,7 @@ dependencies {
 }
 
 tasks.register('copyConsole', Copy) {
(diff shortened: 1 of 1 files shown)


jianglai:
This is being served by vanilla Jetty, where you don't have the issue with the "expires" header. If you remove it you'd have to come up with a different way to serve the static content. If I recall correctly you didn't want the node server to be a permanent solution, right? For now you just need to change the task to copy from the correct location. Later we can work on fine tuning the parameters used by Jetty's default servlet.

jianglai (approved):

PR #2442, opened 2024-05-14
Make logging work correctly on Jetty

jianglai:
There are serveral issues with how logging is set up to work with the Nomulus Jetty image currently:

1. `-D` must come before `-jar`, otherwise it will be interpreted as arguments instead of options. To make things more confusing, when there is only one argument, and it is in the form of `-Dname=value`, Java will attempt to interpret it as system property option. But if there are more then one `-D` present, none of them will be attempted as system properties.

2. For reasons that I don't fully understand, the `GcpJsonFormatter` specified in `logging.properties` is __not__ picked up when Jetty runs, even tough the handler specified there __is__ picked up. This forces me to set up the formatter in `RegistryServlet`'s static initializer block.

3. We cannot use `com.google.cloud.logging.LoggingHandler` with `GcpJsonFormatter`. The structured logs (output from `GcpJsonFormatter`) can only be ingested by Stackdriver if they are written to STDOUT or STDERR verbatim. If we use `com.google.cloud.logging.LoggingHandler`, it treats the output as raw texts and wraps it in a `LogEntry` along with the metadata (some already present in the JSON output). The resulting `LogEntry` is then directly RPC'ed to Stackdriver, or converted to JSON (by the handler) again and written to STDOUT/STDERR. This resulted in doubly nested that cannot be ingested correctly by Stackdriver. As a result, I decided to use `ConsoleHandler` instead, akin to how the proxy writes it logs.

See: https://github.com/googleapis/java-logging/blob/main/google-cloud-logging/src/main/java/com/google/cloud/logging/LoggingHandler.java#L342

TESTED=deployed to alpha and verified that the metadata (including trace id, source location, severity, etc) is ingested correctly .

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2442)


jianglai (commented):
> Reviewed 23 of 24 files at r1, 10 of 11 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2442)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

weiminyu (commented):
The gcp LoggingHandler needs to be configured to write to STDOUT. Have you tried to load the properties using `LogManager.readConfiguration(InputStream)` in the static block?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2442#-:-NxrH-Pv1eW2W6S-hKIV:b-cjvdiu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai:
> The gcp LoggingHandler needs to be configured to write to STDOUT. Have you tried to load the properties using `LogManager.readConfiguration(InputStream)` in the static block?
> 
> > *[Reviewable](https://reviewable.io/reviews/google/nomulus/2442#-:-NxrH-Pv1eW2W6S-hKIV:b-cjvdiu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)
> 
> 
> 
> 
> 

I did try that. In fact if you deploy the master branch (which uses LoggingHandler and writes to STDOUT) you should see that the logs are not parsed correctly. Was it working as of your PR? If so, something must have changed since then.

weiminyu (approved):
I tried with a jetty container locally and it's not working now.

Jetty is pushing SLF4J so maybe they removed JavaLogging-related support?

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2442#-:-NxrRsho6kUfgGjcXSgz:b-av0z4c)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jianglai)

jianglai:
SLF4j is just a facade that Jetty uses, it should not mess with the logging implementation, or how we do logging in our own code.

PR #2443, opened 2024-05-14
Refactor common exception handling in ConsoleApiAction

gbrodman:
There are a bunch of cases where we want common exception handling and it's annoying to have to deal with the common "set failed response and make sure to return" a bunch of times.

This allows us to break up request methods more easily, since we can now often throw exceptions that will break all the way back up to ConsoleApiAction. Previously, any error handling had to exist in the primary handler method so it could return.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2443)


ptkach (approved):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2443)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @gbrodman)

PR #2450, opened 2024-05-22
Add the ability to add/remove console users from a Google Group

jianglai:
If a console user group is defined in the config file, we will add/remove the user to/from the group during user creation/deletion, instead of granting/removing the necessary IAP roles to/from the user.

This makes it easy when IAM roles are managed externally (e. g. under latchkey) where individual changes made to IAM will be rolled back, if not also made to latchkey. This way we can simply grant the role to the group via an one-time config change to latchkey, and rely on a user's membership in the group to determine if IAP would all or deny the user's access.

Also forbid mocking of `Response` in tests. We should use the fake implementation instead.

TESTED=tested both individual and group IAP behavior on crash. Verified that email addresses are added/removed to/from the group.

This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2450)

### DIFF
--- a/config/presubmits.py
+++ b/config/presubmits.py
@@ -196,6 +196,12 @@ PRESUBMITS = {
         {"/node_modules/"},
     ):
         "Use status code from jakarta.servlet.http.HttpServletResponse.",
+    PresubmitCheck(
+        r".*mock\(Response\.class\).*",
+        "java",
+        {"/node_modules/"},
+    ):
+        "Do not mock Response, use FakeResponse.",
 }
 
 # Note that this regex only works for one kind of Flyway file.  If we want to
--- a/core/src/main/java/google/registry/config/RegistryConfig.java
+++ b/core/src/main/java/google/registry/config/RegistryConfig.java
@@ -492,6 +492,18 @@ public final class RegistryConfig {
     }
 
     /**
+     * Returns the email address of the group containing emails of console users.
+     *
+     * <p>This group should be granted the {@code roles/iap.httpsResourceAccessor} role.
+     */
+    @Provides
+    @Config("gSuiteConsoleUserGroupEmailAddress")
+    public static Optional<String> provideGSuiteConsoleUserGroupEmailAddress(
+        RegistryConfigSettings config) {
+      return Optional.ofNullable(Strings.emptyToNull(config.gSuite.consoleUserGroupEmailAddress));
+    }
+
+    /**
      * Returns the email address(es) that notifications of registrar and/or registrar contact
      * updates should be sent to, or the empty list if updates should not be sent.
      *
--- a/core/src/main/java/google/registry/config/RegistryConfigSettings.java
+++ b/core/src/main/java/google/registry/config/RegistryConfigSettings.java
@@ -83,6 +83,7 @@ public class RegistryConfigSettings {
     public String outgoingEmailDisplayName;
     public String adminAccountEmailAddress;
     public String supportGroupEmailAddress;
+    public String consoleUserGroupEmailAddress;
   }
 
   /** Configuration options for registry policy. */
--- a/core/src/main/java/google/registry/config/files/default-config.yaml
+++ b/core/src/main/java/google/registry/config/files/default-config.yaml
@@ -47,6 +47,11 @@ gSuite:
   # given "ADMIN" role on the registrar console.
   supportGroupEmailAddress: support@example.com
 
+  # Group containing the emails of console users. This group should be granted
+  # roles/iap.httpsResourceAccessor out-of-band. If this field is empty, each
+  # console user will be granted to the role individually when they are created.
+  consoleUserGroupEmailAddress:
+
 registryPolicy:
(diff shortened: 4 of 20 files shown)


jianglai (commented):
> Reviewed 20 of 20 files at r1, 4 of 4 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2450)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/test/java/google/registry/tools/server/UpdateUserGroupActionTest.java` line 35 at r1](https://reviewable.io/reviews/google/nomulus/2450#gh-1610063866:-NyVeW605ujESPpRg_7e:b-9j5dzw) ([raw file](https://github.com/google/nomulus/blob/1e59b109f06c4b5c1a9b3dddf27c9acb34ea1168/core/src/test/java/google/registry/tools/server/UpdateUserGroupActionTest.java#L35)):*

Unused class: UpdateUserGroupActionTest is not referenced within this codebase. If not used as an external API it should be removed.

[Show more details](https://github.com/google/nomulus/security/code-scanning/9047)
</blockquote></details>

False positive. This is a test class.

ptkach (changes requested):
> Reviewed 5 of 20 files at r1, 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2450)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

___
*[`core/src/main/java/google/registry/tools/CreateUserCommand.java` line 56 at r4](https://reviewable.io/reviews/google/nomulus/2450#-NyWQ_z--YHSYX7CLfBH:-NyWQ_z--YHSYX7CLfBI:baumrqg) ([raw file](https://github.com/google/nomulus/blob/0c04d13303b2b8800cd32bce6a504d3572935838/core/src/main/java/google/registry/tools/CreateUserCommand.java#L56)):*
> ```Java
>   protected String execute() throws Exception {
>     String ret = super.execute();
>     if (maybeGroupEmailAddress.isPresent()) {
> ```

An assumption is made here that when group email is present we always add to the group, this locks us into this assumption. I'm guessing this might be confusing for someone not familiar with what we're trying to achieve and why, like when group is present or when it's missing, etc. 
Would it make more sense to expand the command and tools with param that would accept the groupname to add to? I'm thinking this might make things more straighforward and clear

jianglai (commented):
> Reviewed 6 of 7 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2450)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

___
*[`core/src/main/java/google/registry/tools/CreateUserCommand.java` line 56 at r4](https://reviewable.io/reviews/google/nomulus/2450#-NyWQ_z--YHSYX7CLfBH:-NyW_k2p0mR9i7Xsvq3g:bc9xbhn) ([raw file](https://github.com/google/nomulus/blob/0c04d13303b2b8800cd32bce6a504d3572935838/core/src/main/java/google/registry/tools/CreateUserCommand.java#L56)):*

An assumption is made here that when group email is present we always add to the group, this locks us into this assumption. I'm guessing this might be confusing for someone not familiar with what we're trying to achieve and why, like when group is present or when it's missing, etc. 
Would it make more sense to expand the command and tools with param that would accept the groupname to add to? I'm thinking this might make things more straighforward and clear
</blockquote></details>

The behavior is well documented in `default-config.yaml`. I don't think we want to make it possible to specify an arbitrary group email address on the command line, as that should be part of the configuration that is under version control.

I do think it make sense to send the group as an argument to the backend so the backend doesn't rely on hard-coded group email addresses.

jianglai (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2450)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @ptkach)

ptkach (commented):
Looking at the IAP ticket that's been bouncing around I'm not 100% that this approach will get to fly. Thus I'm waiting for the to approve the change first. Otherwise we will go with `allAuthorized` approach and that will make this PR obsolete
 

> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2450#-:-NyzgtCiEnNeKJQJQtu-:bwcncz8)* status: 19 of 20 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @jianglai)

ptkach (approved):
> Reviewed 6 of 20 files at r1, 6 of 7 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2450)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

PR #2457, opened 2024-05-28
Move domain-list search+download outside of loading bar

gbrodman:
This means that they'll stick around even while we're loading domains from the server.

This also adds a loading spinner.

https://b.corp.google.com/issues/343213150
https://b.corp.google.com/issues/343211773

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2457)

### DIFF
--- a/console-webapp/src/app/domains/domainList.component.html
+++ b/console-webapp/src/app/domains/domainList.component.html
@@ -1,18 +1,6 @@
 <app-selected-registrar-wrapper>
   <div class="console-domains">
     <h1 class="mat-headline-4">Domains</h1>
-    @if (totalResults === 0) {
-    <div class="console-app__empty-domains">
-      <h1>
-        <mat-icon class="console-app__empty-domains-icon secondary-text"
-          >apps_outage</mat-icon
-        >
-      </h1>
-      <h1>No domains found</h1>
-    </div>
-    } @else if(isLoading) {
-    <mat-progress-bar mode="indeterminate"></mat-progress-bar>
-    } @else {
     <a
       mat-stroked-button
       color="primary"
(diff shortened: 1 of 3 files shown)


ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2457)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/domains/domainList.component.html` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2457#-NzZh7BaAMwlGNrdMITJ:-NzZh7BaAMwlGNrdMITK:bz4nvr) ([raw file](https://github.com/google/nomulus/blob/e797af384e11dd570210ab51856c13c36d5093e5/console-webapp/src/app/domains/domainList.component.html#L28)):*
> ```HTML
>     @if (isLoading) {
>     <div class="console-app__domains-spinner">
>       <mat-spinner />
> ```

I don't think we use spinner anywhere else in the app. We use loading bar. Spinner works as well for when we search or navigate the domains table. However it should be positioned on top of the table, indicating that table is inactive / loading.

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2457)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/domains/domainList.component.html` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2457#-NzZh7BaAMwlGNrdMITJ:-Nz_499XB7xoYAGIsJ1A:bz59tz8) ([raw file](https://github.com/google/nomulus/blob/e797af384e11dd570210ab51856c13c36d5093e5/console-webapp/src/app/domains/domainList.component.html#L28)):*

I don't think we use spinner anywhere else in the app. We use loading bar. Spinner works as well for when we search or navigate the domains table. However it should be positioned on top of the table, indicating that table is inactive / loading.
</blockquote></details>

I used a spinner on purpose because on other pages we use the loading bar at the top of the page to indicate that the whole page is loading, whereas here we load most of the page immediately and the spinner only applies to the table element. It also looks really weird to have potentially two of the same type of loading bar on top of each other with the search bar separating them

ptkach (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2457)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @gbrodman)

___
*[`console-webapp/src/app/domains/domainList.component.html` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2457#-NzZh7BaAMwlGNrdMITJ:-Nz_5uFv3FguDcWi5N_G:b-ncv0ds) ([raw file](https://github.com/google/nomulus/blob/e797af384e11dd570210ab51856c13c36d5093e5/console-webapp/src/app/domains/domainList.component.html#L28)):*

I used a spinner on purpose because on other pages we use the loading bar at the top of the page to indicate that the whole page is loading, whereas here we load most of the page immediately and the spinner only applies to the table element. It also looks really weird to have potentially two of the same type of loading bar on top of each other with the search bar separating them
</blockquote></details>

I guess I wasn't clear. Decision to use spinner for table loading is good, it just that it needs to be on top of the table, we don't want to remove the table element entirely and build it over again. First of all it's complex element with lots of DOM nodes and secondly removing it entirely to show the spinner and than building it back again is not a good UX. Just put the spinner on top of the table, while keeping the table inactive. That should be good

gbrodman (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2457)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @ptkach)

___
*[`console-webapp/src/app/domains/domainList.component.html` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2457#-NzZh7BaAMwlGNrdMITJ:-Nzd0iMrAymbBLfkUifU:bw8ggs6) ([raw file](https://github.com/google/nomulus/blob/e797af384e11dd570210ab51856c13c36d5093e5/console-webapp/src/app/domains/domainList.component.html#L28)):*

I guess I wasn't clear. Decision to use spinner for table loading is good, it just that it needs to be on top of the table, we don't want to remove the table element entirely and build it over again. First of all it's complex element with lots of DOM nodes and secondly removing it entirely to show the spinner and than building it back again is not a good UX. Just put the spinner on top of the table, while keeping the table inactive. That should be good
</blockquote></details>

Gotcha. This new revision does that.

PR #2459, opened 2024-05-29
Unify email notifications for console updates

ptkach:
This change is [Reviewable](https://reviewable.io/reviews/google/nomulus/2459)

### DIFF
--- a/core/src/main/java/google/registry/ui/server/SendEmailUtils.java
+++ b/core/src/main/java/google/registry/ui/server/SendEmailUtils.java
@@ -35,7 +35,7 @@ import javax.mail.internet.InternetAddress;
 public class SendEmailUtils {
 
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
-  private final GmailClient gmailClient;
+  public final GmailClient gmailClient;
   private final ImmutableList<String> registrarChangesNotificationEmailAddresses;
 
   @Inject
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java
@@ -14,26 +14,44 @@
 
 package google.registry.ui.server.console;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static google.registry.request.Action.Method.GET;
 import static jakarta.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 import static jakarta.servlet.http.HttpServletResponse.SC_FORBIDDEN;
 import static jakarta.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static jakarta.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
 
+import com.google.common.base.Ascii;
 import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
 import com.google.common.flogger.FluentLogger;
+import google.registry.batch.CloudTasksUtils;
+import google.registry.export.sheet.SyncRegistrarsSheetAction;
 import google.registry.model.console.ConsolePermission;
 import google.registry.model.console.GlobalRole;
 import google.registry.model.console.User;
+import google.registry.model.registrar.Registrar;
+import google.registry.model.registrar.RegistrarPoc;
+import google.registry.model.registrar.RegistrarPocBase;
+import google.registry.request.Action.Service;
 import google.registry.request.HttpException;
 import google.registry.security.XsrfTokenManager;
 import google.registry.ui.server.registrar.ConsoleApiParams;
 import google.registry.ui.server.registrar.ConsoleUiAction;
+import google.registry.util.DiffUtils;
 import google.registry.util.RegistryEnvironment;
 import jakarta.servlet.http.Cookie;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Stream;
+import javax.inject.Inject;
 
 /** Base class for handling Console API requests */
 public abstract class ConsoleApiAction implements Runnable {
@@ -41,6 +59,8 @@ public abstract class ConsoleApiAction implements Runnable {
   private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
   protected ConsoleApiParams consoleApiParams;
+
+  @Inject CloudTasksUtils cloudTasksUtils;
 
   public ConsoleApiAction(ConsoleApiParams consoleApiParams) {
     this.consoleApiParams = consoleApiParams;
@@ -124,10 +144,115 @@ public abstract class ConsoleApiAction implements Runnable {
     return true;
   }
 
+  private Map<String, Object> expandRegistrarWithContacts(
+      ImmutableSet<RegistrarPoc> contacts, Registrar registrar) {
+
+    ImmutableSet<Map<String, Object>> expandedContacts =
+        contacts.stream()
+            .map(RegistrarPoc::toDiffableFieldMap)
+            // Note: per the javadoc, toDiffableFieldMap includes sensitive data, but we don't want
+            // to display it here
+            .peek(
+                map -> {
+                  map.remove("registryLockPasswordHash");
+                  map.remove("registryLockPasswordSalt");
+                })
+            .collect(toImmutableSet());
+
+    Map<String, Object> registrarDiffMap = registrar.toDiffableFieldMap();
+    Stream.of("passwordHash", "salt") // fields to remove from final diff
+        .forEach(fieldToBeRemoved -> registrarDiffMap.remove(fieldToBeRemoved));
+
+    // Use LinkedHashMap here to preserve ordering; null values mean we can't use ImmutableMap.
+    LinkedHashMap<String, Object> result = new LinkedHashMap<>(registrarDiffMap);
+    result.put("contacts", expandedContacts);
+    return result;
+  }
+
+  protected void sendExternalUpdates(
+      Map<?, ?> diffs, Registrar registrar, ImmutableSet<RegistrarPoc> contacts) {
+
+    if (!consoleApiParams.sendEmailUtils().hasRecipients() && contacts.isEmpty()) {
+      return;
+    }
+
+    if (!RegistryEnvironment.UNITTEST.equals(RegistryEnvironment.get())
+        && cloudTasksUtils != null) {
+      // Enqueues a sync registrar sheet task if enqueuing is not triggered by console tests and
+      // there's an update besides the lastUpdateTime
+      cloudTasksUtils.enqueue(
+          SyncRegistrarsSheetAction.QUEUE,
+          cloudTasksUtils.createGetTask(
+              SyncRegistrarsSheetAction.PATH, Service.BACKEND, ImmutableMultimap.of()));
+    }
+
+    String environment = Ascii.toLowerCase(String.valueOf(RegistryEnvironment.get()));
+    consoleApiParams
+        .sendEmailUtils()
+        .sendEmail(
+            String.format(
+                "Registrar %s (%s) updated in registry %s environment",
+                registrar.getRegistrarName(), registrar.getRegistrarId(), environment),
+            String.format(
+                """
+                  The following changes were made in registry %s environment to the registrar %s by\
+                   %s:
+
+                  %s""",
+                environment,
+                registrar.getRegistrarId(),
+                consoleApiParams.authResult().userIdForLogging(),
+                DiffUtils.prettyPrintDiffedMap(diffs, null)),
+            contacts.stream()
+                .filter(c -> c.getTypes().contains(RegistrarPocBase.Type.ADMIN))
+                .map(RegistrarPoc::getEmailAddress)
+                .collect(toImmutableList()));
+  }
+
+  /**
+   * Determines if any changes were made to the registrar besides the lastUpdateTime, and if so,
+   * sends an email with a diff of the changes to the configured notification email address and all
+   * contact addresses and enqueues a task to re-sync the registrar sheet.
+   */
+  protected void sendExternalUpdatesIfNecessary(EmailInfo emailInfo) {
+    ImmutableSet<RegistrarPoc> existingContacts = emailInfo.contacts();
+    Registrar existingRegistrar = emailInfo.registrar();
+
+    Map<?, ?> diffs =
+        DiffUtils.deepDiff(
+            expandRegistrarWithContacts(existingContacts, existingRegistrar),
+            expandRegistrarWithContacts(emailInfo.updatedContacts(), emailInfo.updatedRegistrar()),
+            true);
+
+    @SuppressWarnings("unchecked")
+    Set<String> changedKeys = (Set<String>) diffs.keySet();
+    if (Sets.difference(changedKeys, ImmutableSet.of("lastUpdateTime")).isEmpty()) {
+      return;
+    }
+
+    sendExternalUpdates(diffs, existingRegistrar, existingContacts);
+  }
+
+  protected record EmailInfo(
+      Registrar registrar,
+      Registrar updatedRegistrar,
+      ImmutableSet<RegistrarPoc> contacts,
+      ImmutableSet<RegistrarPoc> updatedContacts) {
+
+    public static EmailInfo create(
+        Registrar registrar,
+        Registrar updatedRegistrar,
+        ImmutableSet<RegistrarPoc> contacts,
+        ImmutableSet<RegistrarPoc> updatedContacts) {
+      return new EmailInfo(registrar, updatedRegistrar, contacts, updatedContacts);
+    }
+  }
+
   /** Specialized exception class used for failure when a user doesn't have the right permission. */
   private static class ConsolePermissionForbiddenException extends RuntimeException {
     private ConsolePermissionForbiddenException(String message) {
       super(message);
     }
   }
+
 }
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java
@@ -22,10 +22,11 @@ import static jakarta.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 import static jakarta.servlet.http.HttpServletResponse.SC_OK;
 
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.gson.annotations.Expose;
 import google.registry.flows.EppException.AuthenticationErrorException;
 import google.registry.flows.PasswordOnlyTransportCredentials;
-import google.registry.groups.GmailClient;
 import google.registry.model.console.User;
 import google.registry.model.registrar.Registrar;
 import google.registry.request.Action;
@@ -34,10 +35,9 @@ import google.registry.request.auth.Auth;
 import google.registry.request.auth.AuthenticatedRegistrarAccessor;
 import google.registry.request.auth.AuthenticatedRegistrarAccessor.RegistrarAccessDeniedException;
 import google.registry.ui.server.registrar.ConsoleApiParams;
-import google.registry.util.EmailMessage;
+import google.registry.util.DiffUtils;
 import java.util.Optional;
 import javax.inject.Inject;
-import javax.mail.internet.InternetAddress;
 
 @Action(
     service = Action.Service.DEFAULT,
@@ -45,16 +45,12 @@ import javax.mail.internet.InternetAddress;
     method = {POST},
     auth = Auth.AUTH_PUBLIC_LOGGED_IN)
 public class ConsoleEppPasswordAction extends ConsoleApiAction {
-  protected static final String EMAIL_SUBJ = "EPP password update confirmation";
-  protected static final String EMAIL_BODY =
-      "Dear %s,\n" + "This is to confirm that your account password has been changed.";
 
   public static final String PATH = "/console-api/eppPassword";
 
   private final PasswordOnlyTransportCredentials credentials =
       new PasswordOnlyTransportCredentials();
   private final AuthenticatedRegistrarAccessor registrarAccessor;
-  private final GmailClient gmailClient;
 
   private final Optional<EppPasswordData> eppPasswordChangeRequest;
 
@@ -62,11 +58,9 @@ public class ConsoleEppPasswordAction extends ConsoleApiAction {
   public ConsoleEppPasswordAction(
       ConsoleApiParams consoleApiParams,
       AuthenticatedRegistrarAccessor registrarAccessor,
-      GmailClient gmailClient,
       @Parameter("eppPasswordChangeRequest") Optional<EppPasswordData> eppPasswordChangeRequest) {
     super(consoleApiParams);
     this.registrarAccessor = registrarAccessor;
-    this.gmailClient = gmailClient;
     this.eppPasswordChangeRequest = eppPasswordChangeRequest;
   }
 
@@ -107,12 +101,13 @@ public class ConsoleEppPasswordAction extends ConsoleApiAction {
 
     tm().transact(
             () -> {
-              tm().put(registrar.asBuilder().setPassword(eppRequestBody.newPassword()).build());
-              this.gmailClient.sendEmail(
-                  EmailMessage.create(
-                      EMAIL_SUBJ,
-                      String.format(EMAIL_BODY, registrar.getRegistrarName()),
-                      new InternetAddress(registrar.getEmailAddress(), true)));
+              Registrar updatedRegistrar =
+                  registrar.asBuilder().setPassword(eppRequestBody.newPassword()).build();
+              tm().put(updatedRegistrar);
+              sendExternalUpdates(
+                  ImmutableMap.of("password", new DiffUtils.DiffPair("********", "••••••••")),
+                  registrar,
+                  ImmutableSet.of());
             });
 
     consoleApiParams.response().setStatus(SC_OK);
--- a/core/src/main/java/google/registry/ui/server/console/ConsoleUpdateRegistrarAction.java
+++ b/core/src/main/java/google/registry/ui/server/console/ConsoleUpdateRegistrarAction.java
@@ -21,7 +21,7 @@ import static google.registry.util.PreconditionsUtils.checkArgumentPresent;
 import static org.apache.http.HttpStatus.SC_OK;
 
 import com.google.common.base.Strings;
-import google.registry.groups.GmailClient;
+import com.google.common.collect.ImmutableSet;
 import google.registry.model.console.ConsolePermission;
 import google.registry.model.console.User;
 import google.registry.model.registrar.Registrar;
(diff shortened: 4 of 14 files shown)


gbrodman (changes requested):
> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: all files reviewed, 11 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 152 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz519Tn19pDKg_EZyBV:-Nz519Tn19pDKg_EZyBW:b-pqn3dy) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L152)):*
> ```Java
>       Iterable<RegistrarPoc> contacts, Registrar registrar) {
>     ImmutableSet<Map<String, Object>> expandedContacts =
>         Streams.stream(contacts)
> ```

i think `contacts` can be a set, so you can just stream it directly 

but even then do we need to keep track of the contacts separately? The registrar itself has a method that gets the contacts on it. Though that method is a database call, I don't think we'd need to call it multiple times. 

and even then, we're already calling `registrar.getContacts()` multiple times when creating the EmailInfo objects anyway, so removing the existingContacts field from that makes up for it.
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 164 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz53-7O9PBzdDGvxJkX:-Nz53-7O9PBzdDGvxJkY:b-f1vvki) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L164)):*
> ```Java
> 
>     Map<String, Object> registrarDiffMap = registrar.toDiffableFieldMap();
>     Set.of("passwordHash", "salt") // Set of fields to remove from final diff
> ```

seems like overkill to create a set here when there's just two fields? eh 
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 184 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz59wcK00XKue1YXmBw:-Nz59wcK00XKue1YXmBx:b-avuegb) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L184)):*
> ```Java
>     Registrar existingRegistrar = emailInfo.registrar();
> 
>     Map<?, ?> diffs =
> ```

We know that we'll always want to send the email if the override is specified and it seems odd to have the generic method include the diffs that are only specified once

So you can just have two methods 

`void sendExternalUpdatesIfNecessary(EmailInfo emailInfo);` that figures out if we need to do anything and then calls the method below

and

`void sendExternalUpdates(Registrar registrar, Map<?, ?> diffs);` that does the actual work
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 195 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz59lds7NLmmixTYMvk:-Nz59lds7NLmmixTYMvl:b8tgr8i) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L195)):*
> ```Java
>     @SuppressWarnings("unchecked")
>     Set<String> changedKeys = (Set<String>) diffs.keySet();
>     if (CollectionUtils.difference(changedKeys, "lastUpdateTime").isEmpty()) {
> ```

can just use Sets.difference
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 209 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz592mID3bU3nlWm4aS:-Nz592mID3bU3nlWm4aT:b-di5lh5) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L209)):*
> ```Java
> 
>     String environment = Ascii.toLowerCase(String.valueOf(RegistryEnvironment.get()));
>     consoleApiParams
> ```

This gets a bit simpler if you inject the SendEmailUtils which does some of this for us.
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 246 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz57hNr-50gSj6WG4U-:-Nz57hNr-50gSj6WG4U0:bpkn9v0) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L246)):*
> ```Java
>       Registrar registrar,
>       Registrar updatedRegistrar,
>       ImmutableSet<RegistrarPoc> contacts,
> ```

We can just call `getContacts` on the registrar when (if) we want the contacts, no point in requiring it here.
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz56rmy4oVwnqQQWAe1:-Nz56rmz1cQhcIMC6Eun:bik0dyi) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java#L113)):*
> ```Java
>                       registrar.getContacts()),
>                   Optional.of(
>                       ImmutableMap.of("password", new DiffUtils.DiffPair("********", "••••••••"))));
> ```

Maybe use something like "redacted for security" rather than the random-ish characters? That seems clearer, especially since the number of characters probably won't line up.
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5--M9B8w2nwYc5qvM:-Nz5--M9B8w2nwYc5qvN:b-rm70e3) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L113)):*
> ```Java
>                 contacts.get().stream().map(RegistrarPoc::toJsonMap).collect(toImmutableList())));
> 
>     if (!registrarAccessor.hasRoleOnRegistrar(Role.OWNER, registrar.getRegistrarId())) {
> ```

I don't think this is the right permission check since it's using the old role type
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 128 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5AtiU0K6j295JzTmS:-Nz5AtiV7MMYVzBDTa8S:bb4c3j8) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L128)):*
> ```Java
>             () -> {
>               RegistrarPoc.updateContacts(registrar, updatedContacts);
>               Registrar updatedRegistrar =
> ```

We should only do this if the contacts are actually different 
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleApiParams.java` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz58uizA34HujjslhI4:-Nz58uizA34HujjslhI5:b-ityzfo) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/registrar/ConsoleApiParams.java#L28)):*
> ```Java
>     Response response,
>     AuthResult authResult,
>     GmailClient gmailClient,
> ```

This doesn't fit with the rest of the fields, which have to do with the actual request and response themselves. Just inject the client into the action (or even better, inject the SendEmailUtils) 
___
*[`core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java` line 215 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5B7ViApPCDq-MZp9k:-Nz5B7ViApPCDq-MZp9l:b-nuu4l2) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java#L215)):*
> ```Java
>     assertThat(((FakeResponse) consoleApiParams.response()).getStatus()).isEqualTo(SC_FORBIDDEN);
>   }
> ```

We probably want to have a test with a no-op update, where nothing happens if the contacts are the same.

gbrodman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: all files reviewed, 11 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 152 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz519Tn19pDKg_EZyBV:-Nz9xB959ClEK3SJOl7w:b1htz2z) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L152)):*

i think `contacts` can be a set, so you can just stream it directly 

but even then do we need to keep track of the contacts separately? The registrar itself has a method that gets the contacts on it. Though that method is a database call, I don't think we'd need to call it multiple times. 

and even then, we're already calling `registrar.getContacts()` multiple times when creating the EmailInfo objects anyway, so removing the existingContacts field from that makes up for it.
</blockquote></details>

discussed offline, we still need to store the contacts in the EmailInfo object so as to keep the parallelism between expanding the old registrar and the new registrar 

however, i'm p sure we can still change contacts to be some type of set

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: 1 of 14 files reviewed, 11 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 152 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz519Tn19pDKg_EZyBV:-Nz98jR0-VK5XyxG_L6l:b-d00h8f) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L152)):*

discussed offline, we still need to store the contacts in the EmailInfo object so as to keep the parallelism between expanding the old registrar and the new registrar 

however, i'm p sure we can still change contacts to be some type of set
</blockquote></details>

Updated to be a Set
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 164 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz53-7O9PBzdDGvxJkX:-Nz99qLW-PdpYaOI23Vk:bqog80t) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L164)):*

seems like overkill to create a set here when there's just two fields? eh
</blockquote></details>

Agreed
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 184 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz59wcK00XKue1YXmBw:-Nz9QC878JJTI8WCLndJ:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L184)):*

We know that we'll always want to send the email if the override is specified and it seems odd to have the generic method include the diffs that are only specified once

So you can just have two methods 

`void sendExternalUpdatesIfNecessary(EmailInfo emailInfo);` that figures out if we need to do anything and then calls the method below

and

`void sendExternalUpdates(Registrar registrar, Map<?, ?> diffs);` that does the actual work
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 195 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz59lds7NLmmixTYMvk:-Nz9Qvdw0lPgKsB7bC_m:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L195)):*

can just use Sets.difference
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 209 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz592mID3bU3nlWm4aS:-NzA7B6M9RayfYVBAKyx:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L209)):*

This gets a bit simpler if you inject the SendEmailUtils which does some of this for us.
</blockquote></details>

Done.
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java` line 246 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz57hNr-50gSj6WG4U-:-NzA7DqSBYjMJASzn4tY:b255xay) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleApiAction.java#L246)):*

We can just call `getContacts` on the registrar when (if) we want the contacts, no point in requiring it here.
</blockquote></details>

As discussed offline there's a point
___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz56rmy4oVwnqQQWAe1:-NzA7Plt2bL1ssU-Cj4i:bvu3yd8) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java#L113)):*

Maybe use something like "redacted for security" rather than the random-ish characters? That seems clearer, especially since the number of characters probably won't line up.
</blockquote></details>

I don't know, having bother of them say "redacted for privacy" is confusing, but having one leaves the other one to be filled with starts or dots anyway, so why not have both? I believe it makes it more or less clear that the password has been updated.
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5--M9B8w2nwYc5qvM:-NzA7sa24v-QYaY498p6:bu3fqw4) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L113)):*

I don't think this is the right permission check since it's using the old role type
</blockquote></details>

Not sure what do you mean. This is what existing console has in place. 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 128 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5AtiU0K6j295JzTmS:-NzA8Zjl853H3hmMvl4i:bixilsa) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L128)):*

We should only do this if the contacts are actually different
</blockquote></details>

I think we can safely assume that if POST request with contacts is made there was a change. If that's not the case we can add a check to front-end to only submit a request (enable save button) when user makes a change
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleApiParams.java` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz58uizA34HujjslhI4:-NzA9D8tDH7wTrBmgS85:bg0mv2q) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/registrar/ConsoleApiParams.java#L28)):*

This doesn't fit with the rest of the fields, which have to do with the actual request and response themselves. Just inject the client into the action (or even better, inject the SendEmailUtils)
</blockquote></details>

Well this always served a purpose to unify common dependencies for Console API handlers. It just so happened that used to be Request/ Response related, but there's no reason it has to be. 

On a separate note - gmailClient or SendEmailUtils is required in a base class `ConsoleApiAction`, so injection and using it in tests is  rather difficult. This is the best way that works.

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: 1 of 14 files reviewed, 11 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java` line 215 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5B7ViApPCDq-MZp9k:-NzACmWt3dydCrSu77Dd:b-896fix) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java#L215)):*

We probably want to have a test with a no-op update, where nothing happens if the contacts are the same.
</blockquote></details>

Done.

gbrodman (changes requested):
> Reviewed 13 of 13 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: all files reviewed, 3 unresolved discussions (waiting on @ptkach)

___
*[`core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz56rmy4oVwnqQQWAe1:-NzAQ11E034ROfkujWXx:b4pgsj2) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/ConsoleEppPasswordAction.java#L113)):*

I don't know, having bother of them say "redacted for privacy" is confusing, but having one leaves the other one to be filled with starts or dots anyway, so why not have both? I believe it makes it more or less clear that the password has been updated.
</blockquote></details>

sure, i don't feel super strongly
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5--M9B8w2nwYc5qvM:-NzAFDc76fHHU02RtQUo:bowfc2z) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L113)):*

Not sure what do you mean. This is what existing console has in place.
</blockquote></details>

Right but this action isn't the existing console, it's the new one. The AuthenticatedRegistrarAccessor should be deleted once we move entirely to the new console, since it doesn't use the granular permissions model. 

In general ideally we move toward using the granular permission model so that we can delete AuthenticatedRegistrarAccessor sooner 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 128 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5AtiU0K6j295JzTmS:-NzAHWX54T_meVohxehp:b-fzkr3f) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L128)):*

I think we can safely assume that if POST request with contacts is made there was a change. If that's not the case we can add a check to front-end to only submit a request (enable save button) when user makes a change
</blockquote></details>

That is true about the UI, but we probably shouldn't be considering the UI when making sure we're doing the right thing on the backend. On the other hand, I'm not sure how easy that would be to implement. 

In any case, we should probably ignore the "contactsRequireSyncing" field in the registrar object when doing the comparison in "sendExternalUpdatesIfNecessary" right? 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 121 at r2](https://reviewable.io/reviews/google/nomulus/2459#-NzAO0A90pYRNUgCbx8X:-NzAO0A90pYRNUgCbx8Y:bwig6py) ([raw file](https://github.com/google/nomulus/blob/4a42ef7d4da1545bfd1e1f0a21af4658f22ab3ee/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L121)):*
> ```Java
>             savedRegistrar,
>             updatedRegistrar,
>             savedRegistrar.getContacts(),
> ```

FWIW, for this and for the WhoisRegistrarFieldsAction maybe we don't bother actually loading the contacts (especially since this, as written, would involve one database call for each of the "getContacts" calls)? We know they won't be changed, so maybe we just pass an empty set.
___
*[`core/src/main/java/google/registry/ui/server/registrar/ConsoleApiParams.java` line 28 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz58uizA34HujjslhI4:-NzAPrU-3l2sXnS6_sqO:b-7ajd9t) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/registrar/ConsoleApiParams.java#L28)):*

Well this always served a purpose to unify common dependencies for Console API handlers. It just so happened that used to be Request/ Response related, but there's no reason it has to be. 

On a separate note - gmailClient or SendEmailUtils is required in a base class `ConsoleApiAction`, so injection and using it in tests is  rather difficult. This is the best way that works.
</blockquote></details>

ugh yeah that's one downside to this injection structure
___
*[`core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java` line 49 at r2](https://reviewable.io/reviews/google/nomulus/2459#-NzAPxfV7FeGQGx-ffac:-NzAPxfWE0kDGehMmiJT:b-pj6byc) ([raw file](https://github.com/google/nomulus/blob/4a42ef7d4da1545bfd1e1f0a21af4658f22ab3ee/core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java#L49)):*
> ```Java
> import google.registry.request.auth.AuthenticatedRegistrarAccessor;
> import google.registry.testing.ConsoleApiParamsUtils;
> import goojohndoe@theregistrar.comgle.registry.testing.FakeResponse;
> ```

hey now

lol

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: 7 of 14 files reviewed, 4 unresolved discussions (waiting on @gbrodman)

___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 113 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5--M9B8w2nwYc5qvM:-NzEOtED1H7GZh-v9apV:b-26lca4) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L113)):*

Right but this action isn't the existing console, it's the new one. The AuthenticatedRegistrarAccessor should be deleted once we move entirely to the new console, since it doesn't use the granular permissions model. 

In general ideally we move toward using the granular permission model so that we can delete AuthenticatedRegistrarAccessor sooner
</blockquote></details>

As per offline discussion I removed this check. I wasn't sure how and when old roles had been migrated to the new roles and I thought that contacts update still relies on the old role verification. 
___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 128 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5AtiU0K6j295JzTmS:-NzEWRvA-L4imN9lx-5T:b-w0z7ju) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L128)):*

That is true about the UI, but we probably shouldn't be considering the UI when making sure we're doing the right thing on the backend. On the other hand, I'm not sure how easy that would be to implement. 

In any case, we should probably ignore the "contactsRequireSyncing" field in the registrar object when doing the comparison in "sendExternalUpdatesIfNecessary" right?
</blockquote></details>

Well yeah I agree that it might not be very easy to implement this check. However we made sure that no email is sent if there's no change, so other than a DB call to put a new contact there's downside. I'm not sure that it's worth looking into at the moment.

Re `contactsRequireSyncing` I thought so too, as I looked into existing implementation that doesn't have it excluded, but it doesn't affect comparison. So I did the same and it has no effect on a diff, test confirms it.
___
*[`core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java` line 121 at r2](https://reviewable.io/reviews/google/nomulus/2459#-NzAO0A90pYRNUgCbx8X:-NzEazeT95QPOtH01Sbw:b-orumjy) ([raw file](https://github.com/google/nomulus/blob/4a42ef7d4da1545bfd1e1f0a21af4658f22ab3ee/core/src/main/java/google/registry/ui/server/console/settings/SecurityAction.java#L121)):*

FWIW, for this and for the WhoisRegistrarFieldsAction maybe we don't bother actually loading the contacts (especially since this, as written, would involve one database call for each of the "getContacts" calls)? We know they won't be changed, so maybe we just pass an empty set.
</blockquote></details>

Good point

gbrodman (changes requested):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: all files reviewed, 1 unresolved discussion (waiting on @github-advanced-security[bot])

___
*[`core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java` line 128 at r1](https://reviewable.io/reviews/google/nomulus/2459#-Nz5AtiU0K6j295JzTmS:-NzEejAr4f3RVh5XjTj3:b9ss8zv) ([raw file](https://github.com/google/nomulus/blob/1336e64b85c2b0fe1e354879bb3a242330434a7d/core/src/main/java/google/registry/ui/server/console/settings/ContactAction.java#L128)):*

Well yeah I agree that it might not be very easy to implement this check. However we made sure that no email is sent if there's no change, so other than a DB call to put a new contact there's downside. I'm not sure that it's worth looking into at the moment.

Re `contactsRequireSyncing` I thought so too, as I looked into existing implementation that doesn't have it excluded, but it doesn't affect comparison. So I did the same and it has no effect on a diff, test confirms it.
</blockquote></details>

oh cool
___
*[`core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java` line 298 at r4](https://reviewable.io/reviews/google/nomulus/2459#gh-1622733045:-NzEepLWDX_qFLNmExSl:bgcbsj5) ([raw file](https://github.com/google/nomulus/blob/155e0cffa0d215ffba9bb6c37971feee3dacaee7/core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java#L298)):*

## Unread local variable

Variable 'AuthenticatedRegistrarAccessor authenticatedRegistrarAccessor' is never read.

[Show more details](https://github.com/google/nomulus/security/code-scanning/9416)
</blockquote></details>

yeah should probably remove it

ptkach (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2459)* status: 13 of 14 files reviewed, 1 unresolved discussion (waiting on @gbrodman and @github-advanced-security[bot])

___
*[`core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java` line 298 at r4](https://reviewable.io/reviews/google/nomulus/2459#gh-1622733045:-NzEfv9F2pCP0DeDU1Cq:b-896fix) ([raw file](https://github.com/google/nomulus/blob/155e0cffa0d215ffba9bb6c37971feee3dacaee7/core/src/test/java/google/registry/ui/server/console/settings/ContactActionTest.java#L298)):*

yeah should probably remove it
</blockquote></details>

Done.

PR #2463, opened 2024-05-31
Add FeatureFlag table to the database

sarahcaseybot:


jianglai (changes requested):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2463)* status: all files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @sarahcaseybot)

___
*[`db/src/main/resources/sql/flyway/V172__create_feature_flag_table.sql` line 17 at r1](https://reviewable.io/reviews/google/nomulus/2463#-NzUbzfiDxeRK3_-UPBQ:-NzUbzfiDxeRK3_-UPBR:b8606of) ([raw file](https://github.com/google/nomulus/blob/55268a2c2158831f6fac6f508ce459ef0d5bb060/db/src/main/resources/sql/flyway/V172__create_feature_flag_table.sql#L17)):*
> ```TSQL
> create table "FeatureFlag" (
>     feature_name text not null,
>     status bytea not null,
> ```

`bytea` is too generic. I thought we decided that we would just save a timestamp for when the feature is going to be active? Even if you want to store a state transition map, this should be an `hstore`, like `Tld.tld_state_transitions`.

sarahcaseybot (commented):
> *[Reviewable](https://reviewable.io/reviews/google/nomulus/2463)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @CydeWeys and @jianglai)

___
*[`db/src/main/resources/sql/flyway/V172__create_feature_flag_table.sql` line 17 at r1](https://reviewable.io/reviews/google/nomulus/2463#-NzUbzfiDxeRK3_-UPBQ:-NzdlwpwCjJ8m1B4pnpW:b8g1o0j) ([raw file](https://github.com/google/nomulus/blob/55268a2c2158831f6fac6f508ce459ef0d5bb060/db/src/main/resources/sql/flyway/V172__create_feature_flag_table.sql#L17)):*

`bytea` is too generic. I thought we decided that we would just save a timestamp for when the feature is going to be active? Even if you want to store a state transition map, this should be an `hstore`, like `Tld.tld_state_transitions`.
</blockquote></details>

Realized this was made into a bytea and not an hstore because I was missing the necessarry converter in go/r3pr/2464. Added the converter now and the script now made it an hstore type.

